PR #32, opened 2021-04-28
Update the KCP command line to CRD-tenancy changes, and include the cluster controller.

davidfestal:
... with the CRD tenancy implemented on the Kubernetes side.

This R should only be merged when PR https://github.com/kcp-dev/kubernetes/pull/4 is merged.

PR #40, opened 2021-05-04
Update the syncer with CRD tenancy changes

smarterclayton:
The syncer will fetch CRDs, install them into admin, and then
other logical clusters will be able to see them.

Includes updated go.mod from #32


smarterclayton:
Should be merged back into 32, but at least compiles.  I think the demo will need to be updated (primed) with two clusters (don't assume a particular type i think - just take two kubeconfigs?)

PR #42, opened 2021-05-05
More fixes on the Syncer

davidfestal:
- Make the syncer correctly point to the `to` cluster (with the option of passing a kubeconfig for it)
- Only try to synchronize the resource types that have effectively been been imported by the cluster controller
- Create the required rbac rules to enable the syncer service account to create the synced objects in the `to` cluster.
- Fix the syncer reconcile loop and add more detailed logs.

### DIFF
--- a/cmd/syncer/main.go
+++ b/cmd/syncer/main.go
@@ -28,8 +28,9 @@ const (
 )
 
 var (
-	kubeconfig = flag.String("kubeconfig", "", "Config file for -from cluster")
-	clusterID  = flag.String("cluster", "", "ID of this cluster")
+	kubeconfig   = flag.String("from", "", "Kubeconfig file for -from cluster")
+	toKubeconfig = flag.String("to", "", "Kubeconfig file for -to cluster. If not set, the InCluster configuration will be used")
+	clusterID    = flag.String("cluster", "", "ID of this cluster")
 )
 
 func main() {
@@ -46,11 +47,17 @@ func main() {
 		o.LabelSelector = fmt.Sprintf("cluster = %s", *clusterID)
 	})
 
-	// Create a client to modify "to".
-	toConfig, err := clientcmd.BuildConfigFromFlags("", *kubeconfig) // rest.InClusterConfig()
+	var toConfig *rest.Config
+	if *toKubeconfig != "" {
+		toConfig, err = clientcmd.BuildConfigFromFlags("", *kubeconfig) // rest.InClusterConfig()
+	} else {
+		toConfig, err = rest.InClusterConfig()
+	}
 	if err != nil {
 		klog.Fatal(err)
 	}
+
+	// Create a client to modify "to".
 	toClient := dynamic.NewForConfigOrDie(toConfig)
 
 	queue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
--- a/cmd/syncer/main.go
+++ b/cmd/syncer/main.go
@@ -28,9 +28,11 @@ const (
 )
 
 var (
-	kubeconfig   = flag.String("from", "", "Kubeconfig file for -from cluster")
-	toKubeconfig = flag.String("to", "", "Kubeconfig file for -to cluster. If not set, the InCluster configuration will be used")
-	clusterID    = flag.String("cluster", "", "ID of this cluster")
+	fromKubeconfig = flag.String("from_kubeconfig", "", "Kubeconfig file for -from cluster")
+	fromContext    = flag.String("from_context", "", "Context to use in the Kubeconfig file for -from cluster, instead of the current context")
+	toKubeconfig   = flag.String("to_kubeconfig", "", "Kubeconfig file for -to cluster. If not set, the InCluster configuration will be used")
+	toContext      = flag.String("to_context", "", "Context to use in the Kubeconfig file for -to cluster, instead of the current context")
+	clusterID      = flag.String("cluster", "", "ID of this cluster")
 )
 
 func main() {
@@ -38,10 +40,19 @@ func main() {
 	syncedResourceTypes := flag.Args()
 
 	// Create a client to dynamically watch "from".
-	fromConfig, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
+
+	var fromOverrides clientcmd.ConfigOverrides
+	if *fromContext != "" {
+		fromOverrides.CurrentContext = *fromContext
+	}
+
+	fromConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: *fromKubeconfig},
+		&fromOverrides).ClientConfig()
 	if err != nil {
 		klog.Fatal(err)
 	}
+
 	fromClient := dynamic.NewForConfigOrDie(fromConfig)
 	fromDSIF := dynamicinformer.NewFilteredDynamicSharedInformerFactory(fromClient, resyncPeriod, metav1.NamespaceAll, func(o *metav1.ListOptions) {
 		o.LabelSelector = fmt.Sprintf("cluster = %s", *clusterID)
@@ -49,7 +60,17 @@ func main() {
 
 	var toConfig *rest.Config
 	if *toKubeconfig != "" {
-		toConfig, err = clientcmd.BuildConfigFromFlags("", *kubeconfig) // rest.InClusterConfig()
+		var toOverrides clientcmd.ConfigOverrides
+		if *toContext != "" {
+			toOverrides.CurrentContext = *toContext
+		}
+
+		toConfig, err = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+			&clientcmd.ClientConfigLoadingRules{ExplicitPath: *toKubeconfig},
+			&toOverrides).ClientConfig()
+		if err != nil {
+			klog.Fatal(err)
+		}
 	} else {
 		toConfig, err = rest.InClusterConfig()
 	}
--- a/pkg/reconciler/cluster/syncer.go
+++ b/pkg/reconciler/cluster/syncer.go
@@ -6,6 +6,8 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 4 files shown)


cmd/syncer/main.go
```diff
@@ -28,8 +28,9 @@ const (
 )
 
 var (
-	kubeconfig = flag.String("kubeconfig", "", "Config file for -from cluster")
-	clusterID  = flag.String("cluster", "", "ID of this cluster")
+	kubeconfig   = flag.String("from", "", "Kubeconfig file for -from cluster")
```
imjasonh, line 31: Let's rename this variable to `fromKubeconfig` so there's never any notion of "the kubeconfig" in this context.

cmd/syncer/main.go
```diff
@@ -28,8 +28,9 @@ const (
...
+	toKubeconfig = flag.String("to", "", "Kubeconfig file for -to cluster. If not set, the InCluster configuration will be used")
```
imjasonh, line 32: What do you think about `-from_kubeconfig` and `-to_kubeconfig` as flag names? `-from` and `-to` on their own seem a bit too vague.

pkg/reconciler/cluster/syncer.go
```diff
@@ -76,7 +137,7 @@ func installSyncer(ctx context.Context, client kubernetes.Interface, syncerImage
 
 	args := []string{
 		"-cluster", clusterID,
-		"-kubeconfig", "/kcp/kubeconfig",
+		"-from", "/kcp/kubeconfig",
```
imjasonh, line 140: Can we pass the `to` kubeconfig here too? We'd just need to write its value to the ConfigMap too (from the Cluster's `.spec.kubeconfig`), and it'd let us not rely on `InClusterConfig` in real code paths.

pkg/syncer/syncer.go
```diff
@@ -67,32 +68,48 @@ func (c *Controller) handleErr(err error, i interface{}) {
...
 }
 
 func (c *Controller) process(gvr schema.GroupVersionResource, obj interface{}) error {
+	klog.Infof("Process object of type: %T : %v", obj, obj)
 	obj, exists, err := c.FromDSIF.ForResource(gvr).Informer().GetIndexer().Get(obj)
 	if err != nil {
+		klog.Error(err)
 		return err
 	}
+	/*
```
imjasonh, line 89: Is `interfaceToUnstructured` wrong in some way? If so we should delete it. If not we should at least have a TODO about how to fix it here.

pkg/syncer/syncer.go
```diff
@@ -67,32 +68,48 @@ func (c *Controller) handleErr(err error, i interface{}) {
...
+			return err
+		}
+	*/
+	unstrob, isUnstructured := obj.(*unstructured.Unstructured)
 
-	unstrob, err := interfaceToUnstructured(obj)
-	if err != nil {
+	if !isUnstructured {
+		err := errors.New("Object to synchronize is expected to be Unstructured")
+		klog.Error(err)
```
imjasonh, line 100: Instead of `klog.Error`ing before each `return err`, can we just rely on a single `klog.Error` in handling the error returned by `process`? It looks like we already have that above, so we shouldn't even need these, unless something is slipping through?

pkg/syncer/syncer.go
```diff
@@ -103,6 +120,28 @@ func (c *Controller) process(gvr schema.GroupVersionResource, obj interface{}) e
 	return err
 }
 
+func (c *Controller) ensureNamespaceExists(namespace string) error {
+	klog.Infof("Ensuring namespace %s exists", namespace)
```
imjasonh, line 124: This will probably end up being spammy, since we do this check before each new object is created or updated.

Can we perhaps only log when the namespace wasn't found?

Or, since if the namespace isn't found we'll find out anyway when we try to create-or-update the object, what's the value of this explicit namespace existence check?

pkg/syncer/syncer.go
```diff
@@ -120,37 +159,57 @@ func (c *Controller) delete(ctx context.Context, gvr schema.GroupVersionResource
...
-	unstrob.SetResourceVersion("")
+
 	unstrob.SetUID("")
-	if _, err := client.Update(ctx, unstrob, metav1.UpdateOptions{}); err != nil {
-		if k8serrors.IsNotFound(err) {
-			if _, err := client.Create(ctx, unstrob, metav1.CreateOptions{}); err != nil {
-				return err
-			}
-		} else {
+	existing, err := client.Get(ctx, unstrob.GetName(), metav1.GetOptions{})
```
imjasonh, line 168: This always-get-then-update-or-create results in us making 2x more API requests for each object change notification, vs just create-or-update, as before. Is there a reason this is more correct or robust?

pkg/syncer/syncer.go
```diff
@@ -120,37 +159,57 @@ func (c *Controller) delete(ctx context.Context, gvr schema.GroupVersionResource
...
+			klog.Error(err)
 			return err
 		}
+	} else {
+		klog.Error(err)
+		return err
+	}
+
+	klog.Infof("Update the status of object %s/%s", gvr.Resource, unstrob.GetName())
+	_, err = client.UpdateStatus(ctx, unstrob, metav1.UpdateOptions{})
```
imjasonh, line 190: Style nit: I tend to prefer the `if _, err := client.UpdateStatus(...); err != nil { ... }` form for things like this, since the value of `err` won't escape the scope of this small block.

Then you can either `return err` inside the block, or `return nil` outside of it.

Or, as before, just get the value of `err`, then return it (whether nil or not). This change seems to be in service of logging when `UpdateStatus` fails, which will already happen when `upsert` fails, when `process` fails, when `handleErr` sees it.

cmd/syncer/main.go
davidfestal: I agree, especially because we might want to also add `from_Context` and `to_Context` to override the context that we want to refer inside the KubeConfig file, which would be useful to specify in which logical cluster we want to watch on the KCP side.

I'll do that.

pkg/reconciler/cluster/syncer.go
davidfestal: Not sure I agree here. The cluster `.spec.kubeconfig` probably contains the kubeconfig for "external access" to your cluster.
When some code is running inside a cluster POD, it seems more robust to me to use the InCluster config : it uses internal Kubernetes service, the already injected POD service account, etc, etc... that will avoid a number of potential problems with external access, like proxies, etc ...

pkg/syncer/syncer.go
davidfestal: It is not necessary since the object is already of this type (at least in the case of CRDs). And before removing it I was having an error ("Kind should not be empty"), occurring during the conversion.

I'll remove the function then.

pkg/syncer/syncer.go
davidfestal: One big thing with `klog` is that you have the file / line in the code where the error occurred (well, not really... in fact when the  error was *reported*). And that's very handy to understand an error when you cannot debug and have to only read the code to find the error.
Initially, since error was only logged in `handlerErr()` at the end of the processing function, there was no way to understand where precisely the error was triggered.
I'd rather log errors (with the detailed message) as near as possible from where they occur.

Since we're doing a prototype here, it doesn't seem a problem to me if logs are a bit too wordy, especially if it's only to provide more details in cases of error.

pkg/syncer/syncer.go
davidfestal: Since namespace existence is a prerequisite for request creation, creating it (if not already there) seemed natural to me.
The log here is obviously a rest from my tests.
I'll change remove it and will complete the error message in case the namespace cannot be created though it didn't already exist.

pkg/syncer/syncer.go
davidfestal: I'll still `klog` here to have the precise error location in the log message, but will refactor it in your preferred form :-)

davidfestal: as much as possible :-)

davidfestal:
@imjasonh are the answers and changes OK for you ?

pkg/syncer/syncer.go
imjasonh: I understand. Can we move the klog change to a separate, smaller, focused PR, where it's applied evenly across the codebase?

A pattern I've seen is that when `if err != nil { return err }` is replaced with `if err != nil { log.Error(err); return err }` is that it's easy to forget to add the `log` part consistently, and that weakens its usefulness. This isn't `klog`'s fault or `zap`'s or anybody's, except maybe the language for not having better support for this.

An alternative (not my prescribed mandated answer, just an option) is to `return fmt.Errorf("failed trying to 'blah': %w", err)`, which adds debugging information and is harder to miss (and is a line shorter).

Anyway, having this discussion in a separate PR focused on how we want to return and bubble up error information would let us have this conversation separate from the rest of the PR, which we shouldn't otherwise block.

pkg/reconciler/cluster/syncer.go
imjasonh: That makes sense, I'm convinced.

In that case, when we move syncer out of the target cluster and into cluster-controller, will we even need these flags? The cluster-controller can just start a syncer controller with `syncer.Controller{queue, fromDSIF, toClient}` constructed from kubeconfigs, and call `go StartWorker()` on it.

cmd/syncer/main.go
davidfestal: Done

cmd/syncer/main.go
davidfestal: Done

pkg/reconciler/cluster/syncer.go
davidfestal: Well, I somehow think that being able to start the syncer as a standalone command line, and *not as part of the cluster controller* is useful.
That means that is could be changed / adapted by users and started locally, without impacting the cluster-manager, which still have to manage *api resource definitions*.

pkg/syncer/syncer.go
davidfestal: Would it be possible to merge this one and open an issue about making the log management homogeneous ?
I don't really see the point in removing more detailed logs when people might start trying it en encounter failures...

imjasonh (approved):

PR #44, opened 2021-05-06
Bring go.sum up to date with go.mod

iamwillbar:
It looks like the go.sum isn't up to date with the latest changes to go.mod.


smarterclayton:
Thanks

PR #45, opened 2021-05-06
Add getting started instructions to README.md

iamwillbar:


README.md
```diff
@@ -52,6 +52,15 @@ First off, this is a prototype, not a project. We're exploring these ideas here
   - **`deployment-splitter`**, which demonstrates a controller that can split a `Deployment` object into multiple "virtual Deployment" objects across multiple clusters.
   - **`crd-puller`** which demonstrates mirroring CRDs from a cluster back to `kcp`
 
+## How do I get started?
```
imjasonh, line 55: Love this! ❤️ 

Could we move it lower down, either before or after **This sounds cool and I want to help!** ?

Where it's at it feels like it sorta breaks the flow of "what is this; why do I want it; what's it for", and I think "how do I play with it" belongs somewhere closer to "how can I help"

Does that make sense?

iamwillbar: This makes sense. I put it just before "how can I help" because typically people will want to try something out before they know they want to help.

imjasonh (approved):

PR #46, opened 2021-05-06
Added a link for virtual cluster

gyliu513:


docs/investigations/logical-clusters.md
```diff
@@ -1,3 +1,25 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
```
smarterclayton, line 1: Were these generated by the tooling or by you?  If they were generated it'd be great to split that out into its own PR that has a make command  (if you'd like to contribute that)

imjasonh: FYI GitHub automatically generates ToCs for any README.md files. If we renamed this `docs/investigates/logical-clusters/README.md` we'd get the ToC for free, and have a directory to put pictures into.

gyliu513: @smarterclayton it was generated by https://github.com/thlorenz/doctoc , let me check how to enable this in Makefile, but this will request `doctoc` as a pre-requrest. For this PR, I will just add the link for virtual cluster.

@imjasonh with this, it may not good if there are multiple md files under one folder, we cannot name all of them as README.md :)

imjasonh: > @imjasonh with this, it may not good if there are multiple md files under one folder, we cannot name all of them as README.md :)

Right, my suggestion was to move logical-clusters.md into logical-clusters/README.md (separate directory for each doc).

I'd prefer not to generate ToCs since in my experience it leads to syncing issues, or contributor speedbumps, and if GitHub is able to generate them for us I think that's better for everyone.

imjasonh (approved):
Thanks!

Issue #43, opened 2021-05-05
Metadata content of the pod template inside Deployments is pruned on KCP clusters


davidfestal:
When creating a deployment in a KCP cluster that imported the `deployments` resource from a physical cluster, the deployment is created, but the content of the `spec.template.metadata` deployment field is pruned.

This is probably related to a problem in the way the Deploment schema is managed, that triggers the pruning while it is not expected.

This prevents the syncing of Deployments, since the deployment that is being synced s therefore invalid.

PR #47, opened 2021-05-06
Avoid pruning metadata fields that are inside spec (Fix for issue #43)

davidfestal:
Avoid pruning metadata fields that are inside spec

This fixes issue #43.
The problem was that the `ObjectMeta` hard-coded Schema was not defining the `x-preserve-unknown-fields` openapi extension.
So when creating a custom reosurce whose schema contains a field of this type inside its Spec, the content of the metadata map was systematically pruned.

### DIFF
--- a/pkg/crdpuller/discovery.go
+++ b/pkg/crdpuller/discovery.go
@@ -413,6 +413,12 @@ var knownPackages map[string]map[string]apiextensionsv1.JSONSchemaProps = map[st
 		// generate validation for it.
 		"ObjectMeta": apiextensionsv1.JSONSchemaProps{
 			Type: "object",
+
+			// CHANGE vs the OperatorSDK code:
+			// Add `x-preserve-unknown-fields: true` here since it is necessary when metadata is not at the top-level
+			// (for example in Deployment spec.template.metadata), in order not to prune unknown fields and avoid
+			// emptying metadata
+			XPreserveUnknownFields: boolPtr(true),
 		},
 		"Time": apiextensionsv1.JSONSchemaProps{
 			Type:   "string",


pkg/crdpuller/discovery.go
```diff
@@ -413,6 +413,12 @@ var knownPackages map[string]map[string]apiextensionsv1.JSONSchemaProps = map[st
 		// generate validation for it.
 		"ObjectMeta": apiextensionsv1.JSONSchemaProps{
 			Type: "object",
+
+			// CHANGE vs the OperatorSDK code:
```
imjasonh, line 417: This solves the issue for Deployments, and anything else in apps/v1, but does it help users' own CRDs? Or other k8s core types that happen to have a `metadata` field?

davidfestal: yes, it does. The problem would occur for any metadata field that is not at the top-level, and the fix fixes it in all cases, since it is in the CRD generation itself.

imjasonh: Got it, thanks for clarifying. 👍

imjasonh (approved):

imjasonh (approved):

pkg/crdpuller/discovery.go
```diff
@@ -413,6 +413,12 @@ var knownPackages map[string]map[string]apiextensionsv1.JSONSchemaProps = map[st
...
+			// Add `x-preserve-unknown-fields: true` here since it is necessary when metadata is not at the top-level
+			// (for example in Deployment spec.template.metadata), in order not to prune unknown fields and avoid
+			// emptying metadata
+			XPreserveUnknownFields: boolPtr(true),
```
smarterclayton, line 421: I'm really confused why ObjectMeta wouldn't be a struct like any other.  This is fine for now but we're actually going to have to *know* which metadata fields different objects have (they won't all have the same) and so CRD normalization has to have a concrete list.

davidfestal: I partly agree with you.
However currently metadata has quite complex context like the `managedFields` field that obviously are not really designed to be structural-schema-compliant: the CRD generated from this isn't compliant on the server-side.
I assume this might be what explains the choice made by the controller-tools team in this regard.
But sure, in the future we could decide to build the metadata schema ourselves for metadata that are not at the top-level.
FTR metadata at the top-level are expected to be an empty object Json Schema definition by the CRD validator on the apiServer side.

PR #52, opened 2021-05-07
fix(readme): Add min golang version

davidffrench:

PR #58, opened 2021-05-07
Bump to go 1.16 in go.mod

2uasimojo:
We declared go 1.16 as the minimum version via #52. Reflect this in go.mod.

Also `go mod tidy`.


imjasonh:
The `go` directive doesn't enforce a minimum version of Go used to build the module, just adds a hint about what semantics are expected when building the module. https://golang.org/ref/mod#go-mod-file-go

But this is otherwise a harmless change, and the `go mod tidy` is nice, so lgtm.

2uasimojo:
> The `go` directive doesn't enforce a minimum version of Go

True story; but it's the first place I look when trying to figure out what go version I should be using (usually when I try to build and get weird go-ish problems).

Issue #15, opened 2021-04-08
Flag to specify port that kcp should listen on


imjasonh:
kcp currently hard-codes 6443 as its port, it could be 6443 by default with a --port flag to override.

smarterclayton:
Fixed in #59

PR #59, opened 2021-05-07
adding --listen to configure host and port

jhjaggars:
Probably not the best way to do this, but it was a fairly straightforward way to address #15 

Signed-off-by: Jesse Jaggars <jhjaggars@gmail.com>

### DIFF
--- a/cmd/kcp/kcp.go
+++ b/cmd/kcp/kcp.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"net"
 	"net/url"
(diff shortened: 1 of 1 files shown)


cmd/kcp/kcp.go
```diff
@@ -104,6 +108,24 @@ func main() {
 				}
 
 				serverOptions := options.NewServerRunOptions()
+				parts := strings.SplitN(listen, ":", 2)
```
smarterclayton, line 111: you should be able to use ParseHostPort

PR #66, opened 2021-05-10
Make the CRD Puller more robust

davidfestal:
Make the CRD Puller more robust:
- correctly get the OpenAPI v2 model based on the GVK (instead of of guessing it erroneously)
- avoid pulling resource types that are part of the kcp control plane
- avoid stack overflow for recursive OpenAPI V2 schemas
- support resources with non structural OpenAPI schemas, but disable validation in imported resource type
- support protected community groups by adding the required annotation in imported CRD.

### DIFF
--- a/pkg/crdpuller/discovery.go
+++ b/pkg/crdpuller/discovery.go
@@ -2,19 +2,25 @@ package crdpuller
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
+	"k8s.io/apiextensions-apiserver/pkg/apihelpers"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	extensionsapiserver "k8s.io/apiextensions-apiserver/pkg/apiserver"
 	apiextensionsv1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apiserver/pkg/endpoints/openapi"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog"
 	"k8s.io/kube-openapi/pkg/util"
 	"k8s.io/kube-openapi/pkg/util/proto"
 	"k8s.io/kube-openapi/pkg/util/sets"
+	"k8s.io/kubernetes/pkg/api/controlplanescheme"
+	_ "k8s.io/kubernetes/pkg/controlplane/apis/install"
 )
 
 // SchemaPuller allows pulling the API resources as CRDs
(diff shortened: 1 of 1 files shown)


pkg/crdpuller/discovery.go
```diff
@@ -92,133 +107,161 @@ func (sp *schemaPuller) PullCRDs(context context.Context, resourceNames ...strin
 	for _, apiResourcesList := range apiResourcesLists {
 		gv, err := schema.ParseGroupVersion(apiResourcesList.GroupVersion)
 		if err != nil {
-			klog.Errorf("skipping discovery due to error parsing GroupVersion %s: %v", apiResourcesList.GroupVersion, err)
+			klog.Errorf("CRDPuller: Skipping discovery due to error parsing GroupVersion %s: %v", apiResourcesList.GroupVersion, err)
```
imjasonh, line 110: If klog is used here mainly to get line numbers, I think we could remove the CRDPuller prefix, wdyt?

pkg/crdpuller/discovery.go
```diff
@@ -92,133 +107,161 @@ func (sp *schemaPuller) PullCRDs(context context.Context, resourceNames ...strin
...
+				SpecReplicasPath:   ".spec.replicas",
+				StatusReplicasPath: ".status.replicas",
+			}
+			if !hasSubResource("scale") {
+				scaleSubResource = nil
+			}
+
+			crd = &apiextensionsv1.CustomResourceDefinition{
+				TypeMeta:   typeMeta,
+				ObjectMeta: objectMeta,
```
imjasonh, line 232: Can these just be inlined instead of referencing variables dozens of lines away?

pkg/crdpuller/discovery.go
```diff
@@ -92,133 +107,161 @@ func (sp *schemaPuller) PullCRDs(context context.Context, resourceNames ...strin
...
+						Categories: apiResource.Categories,
+						ShortNames: apiResource.ShortNames,
+						Singular:   apiResource.SingularName,
+					},
+				},
+			}
 
-				break
+			apiextensionsv1.SetDefaults_CustomResourceDefinition(crd)
+			if apihelpers.IsProtectedCommunityGroup(gv.Group) {
```
imjasonh, line 261: Add a comment describing why we're doing this here, and how it's related to that PR.

imjasonh (commented):
> avoid stack overflow for recursive OpenAPI V2 schemas

Do we have examples of recursive schemas like this? This seems like a useful test case.

pkg/crdpuller/discovery.go
```diff
@@ -92,133 +107,161 @@ func (sp *schemaPuller) PullCRDs(context context.Context, resourceNames ...strin
...
+			if !pullAllResources && !resourcesToPull.Has(apiResource.Name) {
+				klog.Infof("CRDPuller: Ignoring a resource that has no OpenAPI Schema: %s (%s)", apiResource.Name, gvk.String())
+				continue
+			}
 
-				typeMeta := metav1.TypeMeta{
-					Kind:       "CustomResourceDefinition",
-					APIVersion: "apiextensions.k8s.io/v1",
-				}
+			CRDName := apiResource.Name
```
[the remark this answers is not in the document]
davidfestal: Sure ! Thanks for catching that.

pkg/crdpuller/discovery.go
davidfestal: Sure. Initially they were used in 2 locations, which explains why they were there.
I'll change it.

davidfestal: Fixed

PR #71, opened 2021-05-11
Sync objects with kcp.dev/cluster label

### LINKED ISSUE #70: Prefix syncer `cluster` label with `kcp.dev/`
Created based on the conversation in https://github.com/kcp-dev/kcp/issues/67.

imjasonh:
Fixes #70 

This aligns code with docs at https://github.com/kcp-dev/kcp/blob/main/contrib/demo/README.md#object-syncer


davidfestal (commented):
What about the `owned-by` ? is it specific to KCP or a well-known annotation ?
Additionally what's the use of the `owned-by` annotation knowing that the owner reference it set ?

imjasonh:
> What about the `owned-by` ? is it specific to KCP or a well-known annotation ?
> Additionally what's the use of the `owned-by` annotation knowing that the owner reference it set ?

Good call, done.

PR #72, opened 2021-05-11
Don't sync status from kcp->cluster

imjasonh:
Based on discussion in Slack: https://kubernetes.slack.com/archives/C021U8WSAFK/p1620737930010800

Over time we might identify resources that _should_ have their status updated from kcp to the physical cluster, but in the general case this shouldn't be necessary; clients talking to kcp set or update the spec, kcp+syncer sync that spec to a physical cluster, the cluster updates the status, and syncer updates status back up to kcp.


davidfestal (approved):
OK for me.
FTR you should be able to remove the assignment back to `unstrob` in line [180](https://github.com/kcp-dev/kcp/pull/72/files#diff-b043590efbb8609f4eced1377afbe4f51358c5c628fb79c0fdd24e8f9d553551R180) since it will not be used in the next `UpdateStatus()` anymore.

PR #73, opened 2021-05-11
Support --push_mode, where syncer runs inside cluster controller

imjasonh:
This adds a mode to the cluster controller that runs the syncer code as a separate loop inside the cluster controller, which allows syncing to happen from outside the cluster.

This is an alternative to the two existing modes: pull mode, which runs syncer as a deployment inside the target cluster, and "none-mode" (name needs work) where the syncer is run as a separate binary started up by the person running the other kcp machinery.

This new mode allows for kcp+controllers to more easily run as one standalone binary responsible for watching and managing multiple clusters, without having to run things inside them.

### DIFF
--- a/cmd/cluster-controller/main.go
+++ b/cmd/cluster-controller/main.go
@@ -14,7 +14,8 @@ const numThreads = 2
 var (
 	kubeconfigPath = flag.String("kubeconfig", "", "Path to kubeconfig")
 	syncerImage    = flag.String("syncer_image", "", "Syncer image to install on clusters")
-	pullModel      = flag.Bool("pull_model", true, "Deploy the syncer in registered physical clusters in POD, and have it sync resources from KCP")
+	pullMode       = flag.Bool("pull_mode", true, "Deploy the syncer in registered physical clusters in POD, and have it sync resources from KCP")
+	pushMode       = flag.Bool("push_mode", false, "If true, run syncer for each cluster from inside cluster controller")
 )
 
 func main() {
@@ -39,5 +40,16 @@ func main() {
 		resourcesToSync = []string{"pods", "deployments"}
 	}
 
-	cluster.NewController(r, *syncerImage, kubeconfig, resourcesToSync, *pullModel).Start(numThreads)
+	if *pullMode && *pushMode {
+		log.Fatal("can't set --push_mode and --pull_mode")
+	}
+	syncerMode := cluster.SyncerModeNone
+	if *pullMode {
+		syncerMode = cluster.SyncerModePull
+	}
+	if *pushMode {
+		syncerMode = cluster.SyncerModePush
+	}
+
+	cluster.NewController(r, *syncerImage, kubeconfig, resourcesToSync, syncerMode).Start(numThreads)
 }
--- a/pkg/reconciler/cluster/cluster.go
+++ b/pkg/reconciler/cluster/cluster.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/kcp-dev/kcp/pkg/apis/cluster/v1alpha1"
 	"github.com/kcp-dev/kcp/pkg/crdpuller"
+	"github.com/kcp-dev/kcp/pkg/syncer"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -20,7 +21,10 @@ import (
 	"k8s.io/klog"
 )
 
-const pollInterval = time.Minute
+const (
+	pollInterval     = time.Minute
+	numSyncerThreads = 2
+)
 
 func clusterOriginLabel(clusterID string) string {
 	return "imported-from/" + clusterID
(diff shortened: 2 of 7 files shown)


pkg/reconciler/cluster/cluster.go
```diff
@@ -123,14 +132,48 @@ func (c *Controller) reconcile(ctx context.Context, cluster *v1alpha1.Cluster) e
...
+				cluster.Status.Conditions.SetReady(corev1.ConditionFalse,
+					"ErrorStartingSyncer",
+					fmt.Sprintf("Error starting syncer: %v", err))
+				return nil // Don't retry.
+			}
+
+			s := syncer.New(to, from, syncedResourceTypes, cluster.Name)
+			c.syncers[cluster.Name] = s
+			s.Start(numSyncerThreads)
+		case SyncerModeNone:
```
gyliu513, line 169: May I know why SyncerModeNone is needed? What is the use case for this mode?

imjasonh: Mostly just for flexibility, to be able to run the syncer in isolation as a separate process. You could even run it remotely, it just needs two kubeconfigs and connectivity to kcp and the target cluster.

Over time we might decide it's not worth keeping this mode, but for now I didn't want to remove any behavior while adding this new mode.

pkg/reconciler/cluster/cluster.go
```diff
@@ -123,14 +132,48 @@ func (c *Controller) reconcile(ctx context.Context, cluster *v1alpha1.Cluster) e
...
+			kubeConfig.CurrentContext = logicalCluster
+			bytes, err := clientcmd.Write(*kubeConfig)
+			if err != nil {
+				log.Printf("error writing kubeconfig for syncer: %v", err)
+				cluster.Status.Conditions.SetReady(corev1.ConditionFalse,
+					"ErrorInstallingSyncer",
+					fmt.Sprintf("Error installing syncer: %v", err))
+				return nil // Don't retry.
+			}
+			from, err := clientcmd.RESTConfigFromKubeConfig(bytes)
```
davidfestal, line 148: Instead of writing the `kubeconfig` and parsing it again, you should be able to use:

```suggestion
			from, err := clientcmd.NewNonInteractiveClientConfig(*kubeConfig, logicalCluster, &clientcmd.ConfigOverrides{}, nil).ClientConfig()
```

cmd/cluster-controller/main.go
```diff
@@ -15,6 +15,7 @@ var (
 	kubeconfigPath = flag.String("kubeconfig", "", "Path to kubeconfig")
 	syncerImage    = flag.String("syncer_image", "", "Syncer image to install on clusters")
 	pullModel      = flag.Bool("pull_model", true, "Deploy the syncer in registered physical clusters in POD, and have it sync resources from KCP")
+	pushModel      = flag.Bool("push_model", false, "If true, run syncer for each cluster from inside cluster controller")
```
davidfestal, line 18: Maybe `pull_mode` and `push_mode` would make more sense. I feel like the initial `pull_model` arg name was quite the wrong word when I introduced it.
WDYT ?

pkg/reconciler/cluster/cluster.go
```diff
@@ -123,14 +132,48 @@ func (c *Controller) reconcile(ctx context.Context, cluster *v1alpha1.Cluster) e
 			cluster.Status.Conditions.SetReady(corev1.ConditionUnknown,
 				"SyncerInstalling",
 				"Installing syncer on cluster")
-		} else {
+		case SyncerModePush:
+			// TODO: specify more types
+			syncedResourceTypes := []string{"pods", "deployments"}
```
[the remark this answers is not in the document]
imjasonh: Great suggestion, done.

pkg/reconciler/cluster/cluster.go
```diff
@@ -193,17 +236,28 @@ func (c *Controller) cleanup(ctx context.Context, deletedCluster *v1alpha1.Clust
...
 
 		uninstallSyncer(ctx, client, logicalCluster)
+	case SyncerModePush:
+		s, ok := c.syncers[deletedCluster.Name]
+		if !ok {
+			klog.Errorf("could not find syncer for cluster %q", deletedCluster.Name)
+			return
+		}
+		klog.Infof("stopping syncer for cluster %q", deletedCluster.Name)
+		s.Stop()
```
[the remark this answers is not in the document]
imjasonh: Yep, forgot to do that.

pkg/reconciler/cluster/cluster.go
imjasonh: Very nice, done.

cmd/cluster-controller/main.go
imjasonh: Yeah good call. I was thinking about doing it in a followup but it was easy enough to do in here.

pkg/reconciler/cluster/cluster.go
davidfestal: This would allow you keeping the logic of the cluster controller (that also manages the import of physical API resource definitions as CRDs in KCP) while replacing the default syncing with your own.

kind-east.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: cluster.example.dev/v1alpha1
```
davidfestal, line 1: Not sure these new yaml files make sense in this PR. I assume they were in your repo for testing and were added with the rest of the changes ?

imjasonh: Whoops, bitten by `git add .` again. Deleted.

davidfestal (approved):

CryptoGodfatherVA2 (commented):
var5

Issue #48, opened 2021-05-06
Even lighter weight CRD-only API Server?


vorburger:
@smarterclayton I love this project! Long been thinking this is really at the core of all of it.

Re. https://github.com/kcp-dev/kcp#this-sounds-cool-and-i-want-to-help "feedback can have a big impact", here's a thought:

What if one wanted to build an EVEN MORE lightweight "API server" than the resource types that are currently "built in" here? It's what got me interested in playing with and poking around in @detiber's https://github.com/thetirefire, see [his KubeCon presentation](https://static.sched.com/hosted_files/kccnceu2021/79/From%20Tweet%20to%20BadIdea.pdf) - before hearing about this today. That currently has ONLY `customresourcedefinitions` and `apiservices` built-in - for some scenarios, that may be all one wants.

For example, for something like what @andrewrynhard is building in https://github.com/cosi-project/runtime, see [his KubeCon presentation](https://static.sched.com/hosted_files/kccnceu2021/63/COSI_KubeconEU_2021.pdf), or in say a similar hypothetical agent for a basic KRM-inspired Machine Management CRD with a purely localhost controller running e.g. in a [Static Pod](https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/), having the types currently baked-in to kcp may be too much already?

On the fine day when (eventually...)  _[Kubernetes can be configured to run as kcp runs today](https://github.com/kcp-dev/kcp#is-kcp-a-fork-of-kubernetes-)_, perhaps the "minimal required base resources" could be configurable? Or (better..) there simply could be separate binaries with "nothing at all" baked in (~badidea), "some base types" (~kcp), "original fat one" ( ~apiserver)?

Full disclosure: I'm only half way through reading up on this project, and just starting to learn more about what's what in space. Perhaps this is already possible using the "base apiserver library" (?) - but even just better illustrating and documenting how to really do that IMHO could have value for the community (if that turns out to be all that's "technically" required).

imjasonh:
Thanks for your feedback! I haven't dug into those projects, but I'm definitely excited to learn more.

[...]

PR #74, opened 2021-05-11
docs/investigations: Provide a stub for the minimal API server

smarterclayton:
Just a stub since this has the most need of feedback from different
efforts. Also add one more principle (discussed in terms of how
kcp can help with minimal API server).

Starts #48

### DIFF
--- a/docs/investigations/minimal-api-server.md
+++ b/docs/investigations/minimal-api-server.md
@@ -1,0 +1,55 @@
+# Minimal API Server
+
+The Kubernetes API machinery provides a pattern for declarative config-driven API with a number of conventions that simplify building configuration loops and consolidating sources of truth. There have been many efforts to make that tooling more reusable and less dependent on the rest of the Kube concepts but without a strong use case driving separation and a design the tooling is still fairly coupled to Kube.
+
+## Goal: 
+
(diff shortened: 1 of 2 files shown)


docs/investigations/minimal-api-server.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Constraint: The abstraction for a minimal API server should not hinder Kubernetes development
+
+The primary consumer of the Kube API is Kubernetes - any abstraction that makes a standalone API server possible must not regress Kubernetes performance or overly complicate Kubernetes evolution. The abstraction *should* be an opportunity to improve interfaces within Kubernetes to decouple components and improve comprehension.
+
+### Constraint: Reusing the existing code base
+
+While it is certainly possible to rebuild all of Kube from scratch, a large amount of client tooling and benefit exists within patterns like declarative apply. This investigation is scoped to working within the context of improving the existing code and making the minimal changes within the bounds of API compatibility to broaden utility.
+
+It should be possible to include an arbitrary set of the existing Kube resource infra within the tooling so as to leverage things like `Namespaces` (for logical clusters) or `Secrets` (for benefitting from encryption at rest).
```
imjasonh, line 20: I'm not sure I understand what this is trying to say. Can you elaborate?

smarterclayton: yeah, in brief if you want full fidelity of a resource behavior crds don’t cut it.  so secret encryption at rest or stringData behavior.  will expand a bit although would prefer a higher level use case.

imjasonh (approved):

Issue #80, opened 2021-05-13
kube-root-ca.crt was not deleted under syncer-system after cluster was deleted


gyliu513:
After the cluster was deleted, check `cm` under `syncer-system`, found `kube-root-ca.crt` was not deleted.

```console
$ kubectl get cm -n syncer-system
NAME               DATA   AGE
kube-root-ca.crt   1      14m
```

But if we can delete the ns `syncer-system`, then all resources under this ns will be deleted.

imjasonh:
Yeah, let's just delete all of `syncer-system` when uninstalling the in-cluster syncer. A good e2e test scenario would be exercising install/uninstall/reinstall scenarios to make sure we don't leave anything behind. 👍

PR #81, opened 2021-05-13
Delete syncer-system namespace on cleanup

imjasonh:
Fixes https://github.com/kcp-dev/kcp/issues/80

### DIFF
--- a/pkg/reconciler/cluster/syncer.go
+++ b/pkg/reconciler/cluster/syncer.go
@@ -142,7 +142,7 @@ func installSyncer(ctx context.Context, client kubernetes.Interface, syncerImage
 	args = append(args, resourcesToSync...)
 
 	var one int32 = 1
-	// Create or Update Pod
+	// Create or Update Deployment
 	deployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: syncerNS,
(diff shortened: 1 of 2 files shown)


pkg/reconciler/cluster/syncer.go
```diff
@@ -211,14 +211,10 @@ func installSyncer(ctx context.Context, client kubernetes.Interface, syncerImage
...
 
-// uninstallSyncer uninstalls the syncer image from the target cluster.
-func uninstallSyncer(ctx context.Context, client kubernetes.Interface, logicalCluster string) {
-	if err := client.CoreV1().ConfigMaps(syncerNS).Delete(ctx, syncerConfigMapName(logicalCluster), metav1.DeleteOptions{}); err != nil {
-		klog.Error(err)
-	}
-
-	if err := client.AppsV1().Deployments(syncerNS).Delete(ctx, syncerWorkloadName(logicalCluster), metav1.DeleteOptions{}); err != nil {
-		klog.Error(err)
+// uninstallSyncer uninstalls the syncer from the target cluster by deleting the syncer namespace.
```
gyliu513, line 214: @imjasonh do we want to support the case of multiple upstream cluster share one downstream cluster? If want to support, then deleting the namespace may be a problem, as this will delete all syncer resources for upstream clusters.

As an alternative solution, can we add a pre-fix to the namespace, such as `<upstreamcluster>-syncer-manager`, this can enable each upstream cluster have its own resources in separate namespaces.

imjasonh: Good question. I think we should not expect that behavior, and in fact we should probably take more active steps in future changes to prevent it from happening.

I can't think of a use case for it, and changing the topology from a ~two-level tree into an N-level graph seems like it would make everything more complicated.

gyliu513: Thanks @imjasonh , Yes, we can take this case as low priority. 

Just minor comments, what do you think of changing the namespace to  `<upstreamcluster>-syncer-system`, this can give the end user a clear view for where is the upstream cluster from even if they do not check the resources in this namespace. Again, we can take this comment as low priority as well, as your current fix already fixed the original issue, and I'm good with this now.

imjasonh: For now at least I'd rather not make the namespace name dynamic, since it makes it a bit harder to tell users what to expect to see, and can make things like https://github.com/kcp-dev/kcp/issues/83 harder to implement.

davidfestal: @imjasonh Not sure I understand the underlying reasoning here, but do we really want to prevent a physical cluster to join several logical clusters of a single KCP ? The current code was supporting this by prefix names of workload elements and configmaps by the logical cluster name.
This change seems like a feature regression and a quite hard limitation to me, but I might be missing something...

imjasonh: My thinking was it's easier to add support for multi-ownership of a cluster than to limit it in the future.

The case for a cluster running many syncers seems dubious to me, and not captured in goals.md AFAIK.

If we decide to support it we can do so relatively easily with purposeful changes.

PR #88, opened 2021-05-18
In multiple cluster scenario, CRD sync is incorrect

vincent-pli:
In multiple `cluster` scenario, when CRD in new real `cluster` is exactly same as existed one, should append `imported-from/clustername` to the labels of CRD rather than replacement.

Current implements will cause no CRD in `kcp` when uninstall one of the `cluster`, until next regular reconcile (default 1 min).

### DIFF
--- a/pkg/reconciler/cluster/cluster.go
+++ b/pkg/reconciler/cluster/cluster.go
@@ -11,6 +11,7 @@ import (
 	"github.com/kcp-dev/kcp/pkg/syncer"
 	"github.com/kcp-dev/kcp/pkg/util/errors"
 	corev1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


davidfestal (approved):
That seems good to me. Thanks @vincent-pli  !

pkg/reconciler/cluster/cluster.go
```diff
@@ -80,18 +81,25 @@ func (c *Controller) reconcile(ctx context.Context, cluster *v1alpha1.Cluster) e
...
+
 					pulledCrd.ResourceVersion = clusterCrd.ResourceVersion
 					_, err = c.crdClient.CustomResourceDefinitions().Update(logicalClusterContext, pulledCrd, v1.UpdateOptions{})
 					if k8serrors.IsConflict(err) {
 						return errors.NewRetryableError(fmt.Errorf("Conflict when applying CRD pulled from cluster %s for resource %s: %v", cluster.Name, resourceName, err))
 					}
+				} else if !equality.Semantic.DeepEqual(pulledCrd.Labels, clusterCrd.Labels) {
+					if _, ok := clusterCrd.Labels[clusterOriginLabel(cluster.Name)]; !ok {
+						clusterCrd.Labels[clusterOriginLabel(cluster.Name)] = ""
+						_, err = c.crdClient.CustomResourceDefinitions().Update(logicalClusterContext, clusterCrd, v1.UpdateOptions{})
```
davidfestal, line 101: Please could you add the same check in the second `Update` as the check done in [line 95](https://github.com/kcp-dev/kcp/pull/88/files#diff-6bda1f49bbe6d47cbfdff2057c2e38ea3d12129b78e9c19837b7359f4bf0d06fR95) ?

davidfestal (changes requested):

pkg/reconciler/cluster/cluster.go
vincent-pli: Done

davidfestal (approved):
Thanks !

PR #95, opened 2021-05-26
Pull crd with its additionalPrinterColumns

vincent-pli:
Seems we missing `additionalPrinterColumns` when pull crd from downstream cluster. like this
```
NAME                                             SUCCEEDED   REASON   STARTTIME   COMPLETIONTIME
pipelinerun-with-taskspec-to-echo-good-morning
```

for now, only `NAME` and `AGE` is displayed by default.

### DIFF
--- a/pkg/crdpuller/discovery.go
+++ b/pkg/crdpuller/discovery.go
@@ -20,7 +20,7 @@ import (
 	"k8s.io/kube-openapi/pkg/util/proto"
 	"k8s.io/kube-openapi/pkg/util/sets"
 
-	// The following import is to provide access to the KCP control plane scheme, 
+	// The following import is to provide access to the KCP control plane scheme,
 	// that gathers a minimal set of Kubernetes APIs without any workload-related APIs.
 	//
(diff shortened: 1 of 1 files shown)


pkg/crdpuller/discovery.go
```diff
@@ -224,8 +226,9 @@ func (sp *schemaPuller) PullCRDs(context context.Context, resourceNames ...strin
 				scaleSubResource = nil
 			}
 
+			hasAdditonalPrinterColumn := len(additionalPrinterColumns) != 0
```
davidfestal, line 229: Is the `hasAdditonalPrinterColumn` variable really useful here? Can't it be inlined since it is used only once ?`

davidfestal (approved):
I just left a comment about possibly inlining a variable, but it's up to you.
Apart from that it's OK for me.
Thanks for catching that and fixing it !

pkg/crdpuller/discovery.go
vincent-pli: Make sense, thanks point it out @davidfestal

PR #105, opened 2021-06-15
Support rebalancing Deployments across Clusters

imjasonh:
Instead of only creating leaf Deployments, also update and delete
existing leafs. This means that as Clusters are added and become Ready,
Deployments' replicas will be rebalanced across ready clusters, and as
they get deleted or become un-Ready, they'll rebalance across ready
clusters.

This is incomplete: cluster creations/deletions/changes won't trigger
root deployment reconciliations as of this change.

This also adds unit tests and configures CI to run tests.

(Also fix some lint errors and style nits)

edit: This PR has also grown to include DaemonSet and Pod scheduling, _not hooked up to any controller_, to demonstrate how they'd work in unit tests. These are examples of different scheduling strategies we'd like to generalize and apply to other arbitrary types.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.16
 require (
 	github.com/MakeNowJust/heredoc v1.0.0
 	github.com/go-openapi/jsonpointer v0.19.5 // indirect
+	github.com/google/go-cmp v0.3.0
 	github.com/muesli/reflow v0.1.0
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/pflag v1.0.5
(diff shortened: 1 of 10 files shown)


davidfestal (approved):
I tested this with the kubecon demo (after rebasing on `main` branch and fixing conflicts) and it worked as expected.
So approving !

ncdc:
@imjasonh is this something we want to carry forward and eventually merge?

imjasonh:
> @imjasonh is this something we want to carry forward and eventually merge?

Nope, I think we can drop this. The namespace scheduler will take over as the way to move deployments (and anything) around, and eventually deployment-splitter can just be deleted.

PR #106, opened 2021-06-15
Standardize on klog for logging

imjasonh:
FWIW I'm not totally in love with klog either, but standardizing on _something_ should at least make it easier to change to something else if we find something we like.

### DIFF
--- a/cmd/kcp/kcp.go
+++ b/cmd/kcp/kcp.go
@@ -2,23 +2,21 @@ package main
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
-	"log"
 	"net"
 	"net/url"
 	"os"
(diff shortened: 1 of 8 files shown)


cmd/kcp/kcp.go
```diff
@@ -195,7 +193,7 @@ func main() {
 						}
 
 						if pullMode && pushMode {
-							log.Fatal("can't set --push_mode and --pull_mode")
+							klog.Fatal("can't set --push_mode and --pull_mode")
```
smarterclayton, line 196: heh, this shouldn’t even be fatal, just an err return (as any good cli command should)

smarterclayton:
This is fine for now - if you can correct the one fatal

PR #110, opened 2021-06-18
Add CLI flags to optionally connect to an external etcd.

dgoodwin:
Flags and TLS setup based on code from kube-apiserver.

Tested and should work with both TLS enabled etcd and plain http.

### DIFF
--- a/cmd/kcp/kcp.go
+++ b/cmd/kcp/kcp.go
@@ -2,9 +2,12 @@ package main
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"errors"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"net/url"
 	"os"
@@ -32,6 +35,7 @@ var (
 	installClusterController bool
 	pullMode, pushMode       bool
 	listen                   string
+	etcdClientInfo           etcd.ClientInfo
 )
 
 func main() {
@@ -88,7 +92,7 @@ func main() {
 			}
 			ctx := context.TODO()
 
-			return s.Run(func(cfg etcd.ClientInfo) error {
+			runFunc := func(cfg etcd.ClientInfo) error {
 				c, err := clientv3.New(clientv3.Config{
 					Endpoints: cfg.Endpoints,
 					TLS:       cfg.TLS,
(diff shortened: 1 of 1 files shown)


cmd/kcp/kcp.go
```diff
@@ -219,7 +223,37 @@ func main() {
...
+					etcdClientInfo.TLS.Certificates = []tls.Certificate{cert}
+				}
+
+				if len(etcdClientInfo.TrustedCAFile) > 0 {
+					if caCert, err := ioutil.ReadFile(etcdClientInfo.TrustedCAFile); err != nil {
+						return fmt.Errorf("failed to read ca file: %s", err)
+					} else {
+						caPool := x509.NewCertPool()
+						caPool.AppendCertsFromPEM(caCert)
+						etcdClientInfo.TLS.RootCAs = caPool
```
pires, line 250: At this point, wouldn't you want to set `etcdClientInfo.TLS.InsecureSkipVerify = false`?

cmd/kcp/kcp.go
```diff
@@ -229,8 +263,17 @@ func main() {
...
-	cmd.AddCommand(startCmd)
 
+	startCmd.Flags().StringSliceVar(&etcdClientInfo.Endpoints, "etcd-servers", etcdClientInfo.Endpoints,
+		"List of external etcd servers to connect with (scheme://ip:port), comma separated. If absent an in-process etcd will be created.")
+	startCmd.Flags().StringVar(&etcdClientInfo.KeyFile, "etcd-keyfile", etcdClientInfo.KeyFile,
+		"SSL key file used to secure etcd communication.")
+	startCmd.Flags().StringVar(&etcdClientInfo.CertFile, "etcd-certfile", etcdClientInfo.CertFile,
+		"SSL certification file used to secure etcd communication.")
+	startCmd.Flags().StringVar(&etcdClientInfo.TrustedCAFile, "etcd-cafile", etcdClientInfo.TrustedCAFile,
+		"SSL Certificate Authority file used to secure etcd communication.")
```
pires, line 274: Can we take this opportunity to replace _SSL_ with _TLS_? 👮🏻‍♂️

pires (commented):
Would be cool to actually implement e2e test that brings up an out-of-process etcd _a la_ [kubebuilder and envtest](https://book.kubebuilder.io/reference/envtest.html)?

cmd/kcp/kcp.go
dgoodwin: Good catch, I had that in there at one point and deleted it somehow. Thanks!

cmd/kcp/kcp.go
dgoodwin: SGTM.

davidfestal (approved):
That's nice. Thanks !

cmd/kcp/kcp.go
```diff
@@ -219,7 +223,38 @@ func main() {
 				prepared := server.PrepareRun()
 
 				return prepared.Run(ctx.Done())
-			})
+			}
+
+			if len(etcdClientInfo.Endpoints) == 0 {
+				// No etcd servers specified so create one in-process:
+				return s.Run(runFunc)
+			} else {
```
imjasonh, line 231: style nit: Since the previous `if` returns, you can skip the `else` here and unindent the rest of this.

imjasonh (commented):
Nice!

dgoodwin:
Updated for the style change, thanks!

imjasonh (approved):

imjasonh:
Thanks for adding this! 👍

Issue #93, opened 2021-05-25
Implement a schema diff for imported resource schemas. 


davidfestal:
As described in [this investigation doc](https://github.com/kcp-dev/kcp/blob/main/docs/investigations/logical-clusters.md#crd-virtualization-inheritance-and-normalization), API resources are imported inside a KCP logical cluster from external clusters through CRDs, that can be built from the discovery and openapiv2 information published by the external clusters.
When importing resources as CRDs, if the API resource has already been added as a CRD in the logical cluster, we should perform a diff between the OpenAPI V3 schema of the imported API resources and the schema of the already installed CRD.

We should also provide utilities to:
- Detect whether schemas are incompatible so that we can triggering an error when necessary
- Calculate the Lowest Common Denominator schema between 2 different schemas, when the schemas are compatible (we also have to defined more precisely the case when this LCD makes sense, and when it doesn't and the schema should be considered incompatible).

davidfestal:
This is related to issue https://github.com/kcp-dev/kcp/issues/89

PR #114, opened 2021-06-21
Structural Schema compatibility checker

davidfestal:
This PR provides a function that implements compatibility checks (and LCD generation) on structural schema (JSON schemas used in Kubernetes), and fixes issue https://github.com/kcp-dev/kcp/issues/93

the `EnsureStructuralSchemaCompatibility` function compares a new structural schema to an existing one, to ensure that the existing schema is a sub-schema of the new schema.
In other words that means that all the documents validated by the existing schema will also be validated by the new schema, so that the new schema can be considered backward-compatible with the existing schema.
If it's not the case, errors are reported for each incompatible schema change.

PLEASE NOTE that the implementation is incomplete (it's ongoing work), but still consistent: if some Json Schema elements are changed and the comparison of this type of element is not yet implemented, then an incompatible change error is triggered explaining that the compatison on this element type is not supported.
So there should never be any case when a schema is considered backward-compatible while in fact it is not.

- If the `narrowExisting` argument is `true`, then the LCD (Lowest-Common-Denominator) between existing schema and the new schema is built (when possible), and returned if no incompatible change was detected (like type change, etc...)
- If the `narrowExisting` argument is `false`, then the existing schema is untouched and no LCD schema is calculated. Only the schema compatibility check is done.

In either case, when no errors are reported, it is ensured that either the existing schema or the calculated LCD is a sub-schema of the new schema.


pkg/schemacompat/schemacompat.go
```diff
@@ -0,0 +1,374 @@
...
+	errs := lcdForStructural(fldPath, existingStructural, newStrucural, lcdStructural, narrowExisting)
+	if len(errs) > 0 {
+		return nil, errs.ToAggregate()
+	}
+	serialized, err := json.Marshal(lcdStructural.ToGoOpenAPI())
+	if err != nil {
+		return nil, utilerrors.NewAggregate([]error{err})
+	}
+	var jsonSchemaProps apiextensionsv1.JSONSchemaProps
+	json.Unmarshal(serialized, &jsonSchemaProps)
```
imjasonh, line 57:
```suggestion
	err := json.Unmarshal(serialized, &jsonSchemaProps)
```

pkg/schemacompat/schemacompat.go
```diff
@@ -0,0 +1,374 @@
...
+// then an incompatible change error is triggered explaining that the compatison on this element type is not supported.
+// So there should never be any case when a schema is considered backward-compatible while in fact it is not.
+//
+// If the narrowExisting argument is true, then the LCD (Lowest-Common-Denominator) between existing schema and the new schema
+// is built (when possible), and returned if no incompatible change was detected (like type change, etc...).
+// If the narrowExisting argument is false, the existing schema is untouched and no LCD schema is calculated.
+//
+// In either case, when no errors are reported, it is ensured that either the existing schema or the calculated LCD
+// is a sub-schema of the new schema.
+func EnsureStructuralSchemaCompatibility(fldPath *field.Path, existing, new *apiextensionsv1.JSONSchemaProps, narrowExisting bool) (lcd *apiextensionsv1.JSONSchemaProps, errors utilerrors.Aggregate) {
```
imjasonh, line 33: This should probably just return `error`, which `utilerrors.Aggregate` is one of. Then you can avoid returning `utilerrors.NewAggregate([]error{err})` when you only have a regular error, and can just `return err`.

pkg/schemacompat/schemacompat.go
```diff
@@ -0,0 +1,374 @@
...
+)
+
+// EnsureStructuralSchemaCompatibility compares a new structural schema to an existing one, to ensure that the existing
+// schema is a sub-schema of the new schema. In other words that means that all the documents validated by the existing schema
+// will also be validated by the new schema, so that the new schema can be considered backward-compatible with the existing schema.
+// If it's not the case, errors are reported for each incompatible schema change.
+//
+// PLEASE NOTE that the implementation is incomplete (it's ongoing work), but still consistent:
+// if some Json Schema elements are changed and the comparison of this type of element is not yet implemented,
+// then an incompatible change error is triggered explaining that the compatison on this element type is not supported.
```
imjasonh, line 24:
```suggestion
// then an incompatible change error is triggered explaining that the comparison on this element type is not supported.
```

pkg/schemacompat/schemacompat.go
```diff
@@ -0,0 +1,374 @@
...
+	}
+	var jsonSchemaProps apiextensionsv1.JSONSchemaProps
+	json.Unmarshal(serialized, &jsonSchemaProps)
+	if err != nil {
+		return nil, utilerrors.NewAggregate([]error{err})
+	}
+	return &jsonSchemaProps, nil
+}
+
+func checkTypesAreTheSame(fldPath *field.Path, existing, new *schema.Structural) (errorList field.ErrorList) {
```
imjasonh, line 64: Can this just `return err` too? It looks like you could just `return field.Invalid(...)`, which satisfies the `error` interface.

pkg/schemacompat/schemacompat.go
```diff
@@ -0,0 +1,374 @@
...
+
+func checkUnsupportedValidation(fldPath *field.Path, existing, new interface{}, validationName, typeName string) (errorList field.ErrorList) {
+	if !reflect.ValueOf(existing).IsZero() || !reflect.ValueOf(new).IsZero() {
+		return field.ErrorList{field.Forbidden(fldPath, fmt.Sprintf("The '%s' JSON Schema construct is not supported by the Schema negotiation for type '%s'", validationName, typeName))}
+	}
+	return nil
+}
+
+func checkUnsupportedValidationForNumerics(fldPath *field.Path, existing, new *schema.ValueValidation, typeName string) (errorList field.ErrorList) {
+	errorList = append(errorList, checkUnsupportedValidation(fldPath, existing.Not, new.Not, "not", typeName)...)
```
imjasonh, line 79: `append` is variadic, so you could do something like:

```
errorList = append(errorList,
    checkUnsupportedValidation(fldPath, existing.Not, new.Not, "not", typeName),
    checkUnsupportedValidation(fldPath, existing.AllOf, new.AllOf, "allOf", typeName),
    ...
```

pkg/schemacompat/schemacompat.go
```diff
@@ -0,0 +1,374 @@
...
+	errorList = append(errorList, checkUnsupportedValidation(fldPath, existing.AllOf, new.AllOf, "allOf", typeName)...)
+	errorList = append(errorList, checkUnsupportedValidation(fldPath, existing.AllOf, new.AllOf, "anytOf", typeName)...)
+	errorList = append(errorList, checkUnsupportedValidation(fldPath, existing.AllOf, new.AllOf, "oneOf", typeName)...)
```
imjasonh, line 82:
```suggestion
	errorList = append(errorList, checkUnsupportedValidation(fldPath, existing.AllOf, new.AllOf, "allOf", typeName)...)
	errorList = append(errorList, checkUnsupportedValidation(fldPath, existing.AnyOf, new.AnyOf, "anyOf", typeName)...)
	errorList = append(errorList, checkUnsupportedValidation(fldPath, existing.OneOf, new.OneOf, "oneOf", typeName)...)
```

pkg/schemacompat/schemacompat.go
```diff
@@ -0,0 +1,374 @@
...
+	errorList = append(errorList, checkUnsupportedValidation(fldPath, existing.Enum, new.Enum, "enum", typeName)...)
+	if !(new.Maximum == existing.Maximum && new.Minimum == existing.Minimum && new.ExclusiveMaximum == existing.ExclusiveMaximum && new.ExclusiveMinimum == existing.ExclusiveMinimum) {
```
imjasonh, line 84: Can we rephrase this for readability?

```
if existing.Maximum != new.Maximum ||
    existing.Minimum != new.Minimum ||
    existing.ExclusiveMaximum != new.ExclusiveMaximum ||
    existing.ExclusiveMinimum != new.ExclusiveMinimum {
  errorList = append(errorList,
    checkUnsupportedValidation(...),
    checkUnsupportedValidation(...))
}
```

pkg/schemacompat/schemacompat.go
```diff
@@ -0,0 +1,374 @@
...
+		errorList = append(errorList, checkUnsupportedValidation(fldPath, existing.Maximum, new.Maximum, "maximum", typeName)...)
+		errorList = append(errorList, checkUnsupportedValidation(fldPath, existing.Minimum, new.Minimum, "minimum", typeName)...)
+	}
+	if new.MultipleOf != existing.MultipleOf {
```
imjasonh, line 88: Is this `if` necessary? Won't `checkUnsupportedValidation` also perform this check?

pkg/schemacompat/schemacompat.go
```diff
@@ -0,0 +1,374 @@
...
+		errorList = append(errorList, checkUnsupportedValidation(fldPath, existing.MultipleOf, new.MultipleOf, "multipleOf", typeName)...)
+	}
+	return
+}
+
+func lcdForStructural(fldPath *field.Path, existing, new *schema.Structural, lcd *schema.Structural, narrowExisting bool) (errorList field.ErrorList) {
```
imjasonh, line 94: Same comment about returning `error` here.

pkg/schemacompat/schemacompat.go
```diff
@@ -0,0 +1,374 @@
...
+	if lcd == nil && narrowExisting {
+		return field.ErrorList{field.InternalError(fldPath, errors.New("lcd argument should be passed when narrowExisting is true"))}
+	}
+	if new == nil {
+		return field.ErrorList{field.Invalid(fldPath, nil, "new schema doesn't allow anything")}
+	}
+	if existing.XPreserveUnknownFields != new.XPreserveUnknownFields {
+		errorList = append(errorList, field.Invalid(fldPath.Child("x-preserve-unknown-fields"), new.XPreserveUnknownFields, "x-preserve-unknown-fields value has been changed in an incompatible way"))
```
imjasonh, line 102: Can this error message be more descriptive, like to tell me how it changed? Like, `x-preserve-unknown-fields value changed in an incompatible way from %t to %t`

pkg/schemacompat/schemacompat.go
```diff
@@ -0,0 +1,374 @@
...
+		if existing.XIntOrString {
+			return lcdForIntOrString(fldPath, existing, new, lcd, narrowExisting)
+		} else if existing.XPreserveUnknownFields {
+			return lcdForPreserveUnknownFields(fldPath, existing, new, lcd, narrowExisting)
+		}
+	}
+	return field.ErrorList{field.Invalid(field.NewPath(fldPath.String(), "type"), existing.Type, "Invalid type")}
+}
+
+func lcdForIntegerValidation(fldPath *field.Path, existing, new *schema.ValueValidation, lcd *schema.ValueValidation, narrowExisting bool) (errorList field.ErrorList) {
```
imjasonh, line 128: These should also just return `error` and you won't have to deal with appending to an `ErrorList`.

pkg/schemacompat/schemacompat_test.go
```diff
@@ -0,0 +1,217 @@
...
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+func TestSuccessNewHasMoreProperties(t *testing.T) {
+	existing := &apiextensionsv1.JSONSchemaProps{
+		Type: "object",
+		Properties: map[string]apiextensionsv1.JSONSchemaProps{
+			"existing": {
```
imjasonh, line 17: style nit: it can sometimes help readability to compactify these struct declarations into a single line, e.g.:

```
"existing": {Type: "string"},
"new":      {Type: "integer"},
```

pkg/schemacompat/schemacompat_test.go
```diff
@@ -0,0 +1,217 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+func TestSuccessNewHasMoreProperties(t *testing.T) {
```
imjasonh, line 13: These tests seem like a good opportunity for a [table-driven test](https://github.com/golang/go/wiki/TableDrivenTests).

The case struct looks like it would have `existing` and `new`, and a `want` result and (possibly `nil`) `wantErr` field, and the test body would check that `lcd` matches `c.want` and `err` matches `c.wantErr`.

pkg/schemacompat/schemacompat.go
davidfestal: Good catch ! Thanks !

pkg/schemacompat/schemacompat.go
davidfestal: Thanks for catching this copy/paste error !

pkg/schemacompat/schemacompat.go
davidfestal: TBF the meaning of the returned value is really a *list of errors*. I don't see how hiding this behind a generic `error` interface would make it better.
In most cases you would have to cast the returned error back to an `errors.Aggregate` (provided you first know it is an `Aggregate`), in order to manage errors individually (such as adding the error messages in a list whose format you control).
So if it's acceptable for you, I'd rather keep explicitly declaring the `Aggregate` return type, since you can still assign it to an `error` variable if you really need to.

pkg/schemacompat/schemacompat.go
davidfestal: Yes, I somehow agree, but these `checkUnsupportedValidation` calls (that return a single `field.Error` you're right) are expected to be replaced (in future contributions) by real validation methods that could very likely return a list of errors (because `oneOf`, `anyOf`, etc ... are themselves deep schemas).
So I'd rather let it like that, and have a consistent structure for this comparison logic, if you agree.

pkg/schemacompat/schemacompat.go
davidfestal: Well, in fact I'm not really interested at returning the `error` interface, but more a `*field.Error`.

The logic here is that top-level function (`EnsureStructuralSchemaCompatibility`) returns an Aggregate of errors which can be either `*field.Error` for every error related to the content of the schema, or a generic `error` for errors that occured outside of proper schema comparison (structural schema parse, etc ...).
That's why all schema comparison function share the same return value, which is a list of `field.*Error`. I agree that in some cases we know we only return one error (especially in the case that are not yet implemented), but in full generality comparison at a given level might return several errors. Let's even take the case of integers: as soon as they are implemented, the comparison function might return 2 incompatibility errors: on related to intervals, the other related to `multiple-of`.

So here I can remove the `ErrorList` and return only 1 `*field.Error`, but it's really one of the few case where we're sure we'll never return several errors.

pkg/schemacompat/schemacompat.go
davidfestal: It is necessary, since `checkUnsupportedValidation` is a generic function that returns an error message if at least one the 2 values in non-zero.
In addition to that basic check, here we're comparing the float values... Well, we should, but it seems I have a bug here since I'm comparing pointers...
I'll fix that.

pkg/schemacompat/schemacompat.go
davidfestal: Not sure I see what you mean here. The function will very likely return several `*field.Error`s.

pkg/schemacompat/schemacompat.go
davidfestal: Not sure we really need to duplicate this is the message, since we already add the incompatible new value in `BadValue` field of the `field.*Error`.
So it will be mentioned in the resulting error message.

pkg/schemacompat/schemacompat.go
davidfestal: I don't see why here. `checkUnsupportedValidationForNumerics` may return several incompatibility errors (for min/max interval constraints, and for the `multipleOf` constraint.

pkg/schemacompat/schemacompat_test.go
davidfestal: +1, but I'd rather let this for a next PR, to avoid postponing the merge of the whole API Negotiation work.

pkg/schemacompat/schemacompat_test.go
davidfestal: Sure, I'll fix it.

pkg/schemacompat/schemacompat.go
davidfestal: Quick follow up: when switching the return value here to a `*field.Error`, I'll have to test, at all call sites, whether the returned value is nil, to avoid adding a nil error in the error list.
That doesn't see to make the code cleaner to me.

pkg/schemacompat/schemacompat.go
imjasonh: Handling and reporting multiple errors in Go is unfortunately really painful :(

I think my main complaint might be with `utilerrors.Aggregate`, which seems to be mainly designed only to be used by k8s apimachinery, and doesn't seem to be intended for arbitrary external usage. Things in the code like [`In case of input error list contains nil`](https://github.com/kubernetes/apimachinery/blob/v0.21.2/pkg/util/errors/errors.go#L35) and [`This should never happen, really`](https://github.com/kubernetes/apimachinery/blob/v0.21.2/pkg/util/errors/errors.go#L71) lead me to believe the authors have a very specific usage in mind, and if we do append a nil error or create an `Aggregate` of zero errors, we might get unexpected bugs -- unexpected by us, and by the authors.

I wonder if using a package designed specifically to signal multiple errors might be better, something like [Uber's `multierr`](https://github.com/uber-go/multierr) or [Hashicorp's `go-multierror`](https://github.com/hashicorp/go-multierror). Both of these are a bit more consumer-friendly because they let methods return a standard `error` if they want, but they can also return a multi-error if callers want to crack open the error to filter or format internal errors.

If we expect to make this method available to arbitrary callers, I think thinking through the optimal experience will be worth it, even if it means a bit more work on our end.

I don't think it necessarily needs to block this PR, but we should stay open to changing the return type before proposing more people use this. We might also find more sharp corners as we use it more ourselves.

davidfestal: Thanks for the more detailed answer. I agree we might have to think about it more deeply when about to make this method available to arbitrary users. OTOH the comparison is limited to Kubernetes structural schemas, and take `*apiextensionsv1.JSONSchemaProps` arguments, which is also completely related to Kubernetes.
So while the goal is finally to share / contribute this method outside KCP, I'm not sure it makes sense outside of the Kubernetes ecosystem.

imjasonh (approved):

PR #115, opened 2021-06-21
Integrate API Negotiation into KCP

davidfestal:
This PR introduces a prototype of API Negotiation for KCP based on the Structural Schema compatibility checker introduced in PR https://github.com/kcp-dev/kcp/pull/114.

- It defines 2 new CRDs:
    - `APIResourceImport`s: they contain the description of API resources imported from external clusters (physical or logical), with the OpenAPIv3 schema built from what has been pulled from the external cluster by the CRDPuller. In a KCP logical cluster, there would be 1 APIResourceImport per API resource and per location (typically physical cluster) 
    - `NegociatedAPIResource`s: they define the result of the API Negotiation between all the `APIResourceImport`s for a given api resource, and contain the resource schema that is expected to be used inside the KCP logical cluster. In a logical cluster, there would be only 1 `NegociatedAPIResource` per API Resource. `APIResourceImport`s are checked against the `NegociatedAPIResource` for compatibility. `NegociatedAPIResource`s can be published to the KCP logical cluster, so that the API will be available and visible in the logical cluster. This is done by creating a CRD with the information, especially the schema, contained inside the `NegociatedAPIResource`.
    
    More information about each type is provided in the code comments.

- It implements the API negotiation logic in a dedicated controller that also watches CRDs, in addition to those 2 new CRDs.
- It integrates the API Negotiation inside KCP by creating `APIResourceImport`s for each API resource imported from physical clusters, instead of directly creating a CRD without any API compatibility check (as it was before)
- It adds the `apiNegotiation` demo script that showcases how the API negotiation works, with 1 kind cluster based on the latest Kubernetes 1.20 version, and the second cluster based on Kubernetes 1.5.

**NOTE:** This PR is opened against the branch of PR https://github.com/kcp-dev/kcp/pull/114
As soon as PR https://github.com/kcp-dev/kcp/pull/114 is reviewed and merged, this PR would be rebased against `main`

PR #116, opened 2021-06-23
Integrate API Negotiation into KCP

davidfestal:
This PR introduces a prototype of API Negotiation for KCP based on the Structural Schema compatibility checker introduced in PR https://github.com/kcp-dev/kcp/pull/114.

- It defines 2 new CRDs:
    - `APIResourceImport`s: they contain the description of API resources imported from external clusters (physical or logical), with the OpenAPIv3 schema built from what has been pulled from the external cluster by the CRDPuller. In a KCP logical cluster, there would be 1 APIResourceImport per API resource and per location (typically physical cluster) 
    - `NegociatedAPIResource`s: they define the result of the API Negotiation between all the `APIResourceImport`s for a given api resource, and contain the resource schema that is expected to be used inside the KCP logical cluster. In a logical cluster, there would be only 1 `NegociatedAPIResource` per API Resource. `APIResourceImport`s are checked against the `NegociatedAPIResource` for compatibility. `NegociatedAPIResource`s can be published to the KCP logical cluster, so that the API will be available and visible in the logical cluster. This is done by creating a CRD with the information, especially the schema, contained inside the `NegociatedAPIResource`.
    
    More information about each type is provided in the code comments.

- It implements the API negotiation logic in a dedicated controller that also watches CRDs, in addition to those 2 new CRDs.
- It integrates the API Negotiation inside KCP by creating `APIResourceImport`s for each API resource imported from physical clusters, instead of directly creating a CRD without any API compatibility check (as it was before)
- It adds the `apiNegotiation` demo script that showcases how the API negotiation works, with 1 kind cluster based on the latest Kubernetes 1.20 version, and the second cluster based on Kubernetes 1.5.


davidfestal:
This PR is indeed a duplicate of PR https://github.com/kcp-dev/kcp/pull/115 that was by mistake merged against branch [kcp:schema-diff-and-lcd-builder](https://github.com/kcp-dev/kcp/tree/schema-diff-and-lcd-builder) instead of the main main

imjasonh (approved):

PR #120, opened 2021-06-29
Make group not required, to support core types

imjasonh:
Also:
- gitignore `*.log` files created by demo scripts
- rewrite `make build` for brevity
- log resources that cluster controller will sync
- rebuild and apply CRDs in demo script


ianzhang366:
it works on my env, thanks a lot

PR #121, opened 2021-06-29
Use multierr instead of k8s aggregate error

imjasonh:
~~Also rewrite schemacompat_test.go to use table-driven tests~~

Based on non-blocking review feedback in #114


pkg/schemacompat/schemacompat.go
```diff
@@ -103,34 +103,36 @@ func stringPointersEqual(p1, p2 *string) bool {
...
-		return field.ErrorList{field.InternalError(fldPath, errors.New("lcd argument should be passed when narrowExisting is true"))}
+		return field.InternalError(fldPath, errors.New("lcd argument should be passed when narrowExisting is true"))
 	}
 	if new == nil {
-		return field.ErrorList{field.Invalid(fldPath, nil, "new schema doesn't allow anything")}
+		return field.Invalid(fldPath, nil, "new schema doesn't allow anything")
 	}
 	if existing.XPreserveUnknownFields != new.XPreserveUnknownFields {
-		errorList = append(errorList, field.Invalid(fldPath.Child("x-preserve-unknown-fields"), new.XPreserveUnknownFields, "x-preserve-unknown-fields value has been changed in an incompatible way"))
+		return field.Invalid(fldPath.Child("x-preserve-unknown-fields"), new.XPreserveUnknownFields, "x-preserve-unknown-fields value has been changed in an incompatible way")
```
imjasonh, line 135: This raises a question about the intent of the original code. Before this change, if `x-preserve-unknown-fields` didn't match, we'd append to `errorList`, but because of how named return values work, this would never actually get returned, because all the `return`s below would shadow and override it.

With this change, as it currently stands, if `x-preserve-unknown-fields` doesn't match, we'll return that error only, and not perform LCD for the remaining types.

I think the intention was that if `x-preserve-unknown-fields` didn't match, we should report that, _and_ any other errors from LCDing.

In any case, we should have a test that covers this behavior, whichever behavior we decide we want.

pkg/schemacompat/schemacompat.go
```diff
@@ -343,46 +347,49 @@ func lcdForObject(fldPath *field.Path, existing, new *schema.Structural, lcd *sc
...
 				lcd.AdditionalProperties.Bool = false
 				lcd.AdditionalProperties.Structural = new.AdditionalProperties.Structural
 			}
 		}
-	} else {
-		// Existing schema doesn't allow anything => new will always be a superset of existing
 	}
 
-	return
+	multierr.AppendInto(&err, lcdForObjectValidation(fldPath, existing.ValueValidation, new.ValueValidation, lcd.ValueValidation, narrowExisting))
```
imjasonh, line 381: This was found by linting, because `lcdForObjectValidation` was never called previously. Let me know if this matches your expectation for when/how this should be called.

davidfestal (changes requested):
I've just gave a quick look to the PR for now, but already have the following remarks:
- Do you think it could be possible to split the PR into 2:
    - one PR to setup table-drive tests (and possibly CI additions for linting, etc ...),
    - a distinct PR related to compatibility checks error management ?
   
- As for the changes related to error management, I think I'd rather completely switch to our own structured errors (similar to `field.Error` provides, but dedicated to the compatibility checks). It would be a wider change, but also IMO a more valuable change than hiding structured errors behind a generic multierr `error`.
Why not opening a dedicated issue for this, to discuss the design, and then link the resulting dedicated PR ?

imjasonh:
Rebasing this on top of https://github.com/kcp-dev/kcp/pull/122 (once it's merged) should fix all the lint issues (unless I introduced any new ones 😆 )

I'd be fine with introducing a new PR only to add table-driven tests, I can send that soon.

imjasonh:
This branch is now based on changes in https://github.com/kcp-dev/kcp/pull/126/

This change is now basically: https://github.com/imjasonh/kcp/compare/table-driven..multierr

PR #122, opened 2021-06-30
Fix lint errors and don't ignore new lint errors

imjasonh:


pkg/reconciler/cluster/controller.go
```diff
@@ -107,7 +107,9 @@ func NewController(cfg *rest.Config, syncerImage string, kubeconfig clientcmdapi
 			}
 			return []string{}, nil
 		},
-	})
+	}); err != nil {
+		klog.Fatalf("Failed to add indexer for APIResourceImport: %v", err)
```
davidfestal, line 111: Do we really want to **exit** the calling executable here ? Isn't `Errorf` sufficient, especially if we think that this method should be available as a library to be included into other pieces of software ?
I'd rather add an error return value to the `NewController` function I assume.

davidfestal (changes requested):

davidfestal (approved):

PR #124, opened 2021-07-06
Enable demos to be easily run in interactive mode or as non-interactive tests (possibly in CI)

davidfestal:
Enable demos to be easily run in interactive mode or as non-interactive tests (possibly in CI in a dedicated GH action).
For now as a GH action it can only be started on demand (with the Run Workflow button on GH actions).
In the future it could also be added to CI (as soon as the tests are stable enough).


contrib/demo/testDemos.sh
```diff
@@ -0,0 +1,48 @@
...
+
+source ${DEMO_ROOT}/.startUtils
+setupTraps $0
+
+cleanupDemoOutput() {
+  sed -E --posix -e '/^☸️ \$ /d' -e '/^ +certificate-authority-data: /d' -e '/^ +server: https:\/\/127\.0\.0\.1:[0-9]{5}$/d' -e 's/^(.*)( +[0-9]+s|[0-9]+m([0-9]+s)?)$/\1/' $1
+}
+
+error=false
+for demo in kubecon ; do
```
imjasonh, line 15: This for loop seems like a premature optimization.

contrib/demo/.startUtils
```diff
@@ -0,0 +1,35 @@
+wait_command() {
+  local command="$1"; shift
+  local wait_seconds="${1:-40}"; shift # 40 seconds as default timeout
```
imjasonh, line 3: `wait_command` is never called with a second argument, we can just make this a constant value.

contrib/demo/.startUtils
```diff
@@ -0,0 +1,35 @@
...
+
+  until [[ $((wait_seconds--)) -eq 0 ]] || eval "$command &> /dev/null" ; do sleep 1; done
+
+  ((++wait_seconds))
+}
+
+wait_file() {
```
imjasonh, line 10: `wait_file` appears to be unused in this change. Can we delete it?

contrib/demo/testDemos.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
imjasonh, line 1: This feels like an uncomfortable amount of Bash engineering for me. I _love_ that we'll have guards against demo regressions in CI, but I think we should consider some e2e tests in Go in the future.

contrib/demo/deployment.yaml
```diff
@@ -3,7 +3,11 @@ kind: Deployment
 metadata:
   name: my-deployment
 spec:
-  replicas: 15
+  replicas: 14
```
imjasonh, line 6: I assume the change to 14 is so the replicas balance reproducibly?

What's the motivation behind `RollingUpdate`?

.github/workflows/ci.yaml
```diff
@@ -41,7 +42,6 @@ jobs:
     - uses: actions/checkout@v1
     - uses: reviewdog/action-misspell@v1
```
imjasonh, line 43:
```suggestion
    - uses: reviewdog/action-misspell@v1
      with:
        exclude: .github/workflows/ci.yaml
```

contrib/demo/testDemos.sh
davidfestal: My plan is to add the already-existing second demo about API negotiation :-)

contrib/demo/.startUtils
davidfestal: I assume we can delete it, yes.

contrib/demo/testDemos.sh
davidfestal: +1 for this in the future.
Though being able to use, in tests, the exact same scenarios as those we define in demos we show seemed relevant to me.

contrib/demo/deployment.yaml
davidfestal: Default rolling update strategy has a non-zero `maxUnavailable`, so that  the Deployment is considered Available even when not all the replicas are there. This finally leads to the demo scenario not being reproducible.

.github/workflows/ci.yaml
davidfestal: I just made this change on my side.
FTR the PR is still a bit draft, but not a realt (Draft PR) since you need to have a real PR to test the GH actions properly.
So some additional fixes might come in.

imjasonh: Sounds good, let me know when you'd like me to take another look. 👍

contrib/demo/.startUtils
davidfestal: Done

contrib/demo/.startUtils
davidfestal: I now use the argument

contrib/demo/testDemos.sh
davidfestal: Done

contrib/demo/testDemos.sh
davidfestal: I reworked this script be also be an easy and clean way to run a demo (one of the 2 demos that already exist, or additional demos in the future) either interactively or as an automatic test that can be triggered in a GH action.
Of course it's completely clear that GO-based E2E tests are still mandatory.

PR #126, opened 2021-07-06
Table-driven tests for schemacompat

imjasonh:
Split out from https://github.com/kcp-dev/kcp/pull/121


davidfestal (approved):
Nice ! We could even go one step further, and use test fixtures, like what had been done in repo https://github.com/devfile/api/blob/main/pkg/utils/overriding/overriding_test.go#L180 (see [here](https://github.com/devfile/api/tree/main/pkg/utils/overriding/test-fixtures) to see the YAML files that define the tests. That's even more easy to read and contribute to.

PR #131, opened 2021-07-08
Generalize createKindClusters.sh to create multiple clusters

imjasonh:
When given no args, it creates a us-east1 and us-west1 cluster as
before. When given args, it creates clusters with those names. KinD
configs, kubeconfigs and Cluster CRD YAMLs are generated and gitignored
to avoid churn and potentially leaking sensitive information.

### DIFF
--- a/contrib/demo/clusters/kind/createKindClusters.sh
+++ b/contrib/demo/clusters/kind/createKindClusters.sh
@@ -3,6 +3,7 @@
 set -euxo pipefail
 
 DEMO_ROOT="$(dirname "${BASH_SOURCE}")/../.."
+TMPDIR=$(mktemp -d)
 
 clusters="$@"
 if [[ $# -eq 0 ]]; then
@@ -10,18 +11,16 @@ if [[ $# -eq 0 ]]; then
 fi
 
 for name in ${clusters}; do
-  if [[ ! -f ${DEMO_ROOT}/clusters/kind/${name}.config ]]; then
-    cat > ${DEMO_ROOT}/clusters/kind/${name}.config << EOF
+  cat > ${TMPDIR}/${name}.config << EOF
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 name: ${name}
 networking:
   apiServerAddress: "127.0.0.1"
 EOF
-  fi
 
   kind delete cluster --name=${name} || true
-  kind create cluster --config ${DEMO_ROOT}/clusters/kind/${name}.config --kubeconfig ${DEMO_ROOT}/clusters/kind/${name}.kubeconfig
+  kind create cluster --config ${TMPDIR}/${name}.config --kubeconfig ${TMPDIR}/${name}.kubeconfig
 
   cat > ${DEMO_ROOT}/clusters/${name}.yaml << EOF
 apiVersion: cluster.example.dev/v1alpha1
@@ -31,5 +30,5 @@ metadata:
 spec:
   kubeconfig: |
 EOF
-  sed -e 's/^/    /' ${DEMO_ROOT}/clusters/kind/${name}.kubeconfig >> ${DEMO_ROOT}/clusters/${name}.yaml
+  sed -e 's/^/    /' ${TMPDIR}/${name}.kubeconfig >> ${DEMO_ROOT}/clusters/${name}.yaml
 done
--- a/contrib/demo/clusters/kind/createKindClusters.sh
+++ b/contrib/demo/clusters/kind/createKindClusters.sh
@@ -1,9 +1,35 @@
-#!/bin/bash -e
+#!/usr/bin/env bash
+
+set -euxo pipefail
 
 DEMO_ROOT="$(dirname "${BASH_SOURCE}")/../.."
 
-kind create cluster --config ${DEMO_ROOT}/clusters/kind/us-east1.config --kubeconfig ${DEMO_ROOT}/clusters/kind/us-east1.kubeconfig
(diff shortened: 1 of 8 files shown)


contrib/demo/clusters/kind/createKindClusters.sh
```diff
@@ -1,9 +1,34 @@
...
+  clusters="us-east1 us-west1"
+fi
+
+for name in ${clusters}; do
+  cat > ${TMPDIR}/${name}.config << EOF
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+name: ${name}
+networking:
+  apiServerAddress: "127.0.0.1"
```
mangelajo, line 19: just curious, what's the reason for apiServerAddress 127.0.0.1 ?

mangelajo (approved):
One question (unrelated to this patch), but looks good to me.

mangelajo:
@imjasonh @davidfestal can we merge this? I wanted to build a PR on top of it and it looks good to me.

davidfestal:
I'm OK with it, apart from the fact that, in the default case for the west cluster this PR removes the overriden k8s version (1.15.x) which is necessary for the apiNegociation test.
@imjasonh was it on purpose ?

davidfestal (commented):
In the default case for the west cluster this PR removes the overriden k8s version (1.15.x) which is necessary for the apiNegociation test.
@imjasonh was it on purpose ?

mangelajo:
> In the default case for the west cluster this PR removes the overriden k8s version (1.15.x) which is necessary for the apiNegociation test.
> @imjasonh was it on purpose ?

Oh right, I guess it's accidental `  image: kindest/node:v1.15.12 `would be missing for the west cluster.

I guess we could have an array with the versions ?

davidfestal:
> Oh right, I guess it's accidental ` image: kindest/node:v1.15.12`would be missing for the west cluster.
> 
> I guess we could have an array with the versions ?

@imjasonh Do you agree this would be a good simple way to make this PR fully compatible with the previous behavior (especially for the `apiNegociation` demo) ?

imjasonh:
> > Oh right, I guess it's accidental ` image: kindest/node:v1.15.12`would be missing for the west cluster.
> > I guess we could have an array with the versions ?
> 
> @imjasonh Do you agree this would be a good simple way to make this PR fully compatible with the previous behavior (especially for the `apiNegociation` demo) ?

How about `createKindClusters.sh us-east us-west@1.15.2`, and the script just checks for `@` in the name?

The default would still be to create two clusters with the latest k8s version, but apiNegotiation would still be able to request an older version for one cluster.

mangelajo:
ping @imjasonh @davidfestal

mangelajo:
Hey sorry, I realized I can make my work independent of this, please ignore my comments, no urge in merging.

contrib/demo/clusters/kind/createKindClusters.sh
```diff
@@ -1,9 +1,34 @@
...
+EOF
 
-kind create cluster --config ${DEMO_ROOT}/clusters/kind/us-east1.config --kubeconfig ${DEMO_ROOT}/clusters/kind/us-east1.kubeconfig
-sed -e 's/^/    /' ${DEMO_ROOT}/clusters/kind/us-east1.kubeconfig | cat ${DEMO_ROOT}/clusters/us-east1.yaml - > ${DEMO_ROOT}/clusters/kind/us-east1.yaml
+  kind delete cluster --name=${name} || true
+  kind create cluster --config ${TMPDIR}/${name}.config --kubeconfig ${TMPDIR}/${name}.kubeconfig
```
mangelajo, line 23:
```suggestion
  kind create cluster --config ${TMPDIR}/${name}.config --kubeconfig ${DEMO_ROOT}/clusters/kind/${name}.kubeconfig
```

Could we keep it like this?

And then add those .kubeconfigs to .gitignore ? 

Otherwise for some work I'm doing later it's not possible to fund the kubeconfigs and use them

You can always rely on `kind get kubeconfig --name $cluster` but, if you have other kind cluster in the system we can't know if they belong to this demo or something else.

mangelajo (changes requested):
check the kubeconfig comments.

contrib/demo/clusters/kind/createKindClusters.sh
```diff
@@ -1,9 +1,34 @@
...
+  cat > ${DEMO_ROOT}/clusters/${name}.yaml << EOF
+apiVersion: cluster.example.dev/v1alpha1
+kind: Cluster
+metadata:
+  name: ${name}
+spec:
+  kubeconfig: |
+EOF
+  sed -e 's/^/    /' ${TMPDIR}/${name}.kubeconfig >> ${DEMO_ROOT}/clusters/${name}.yaml
+done
```
mangelajo, line 34: I say this, because this kubeconfig .yaml "Cluster" kind is something I can't consumer later down the road (I could strip it out again dynamically though).

davidfestal: I completely agree

ncdc:
Going through old-ish PRs - @imjasonh do you want to revive this & get it in? Want someone to take it over? Close it? Thanks!

stevekuznetsov:
I wrote Go code to do the `Cluster` creation with the kubeconfig for a cluster, would rather us have more generic Go than Bash TBH

ncdc:
SGTM. Would like to hear from Jason before closing, though.

ncdc:
@imjasonh based on yesterday's community meeting discussion about not wanting to maintain the demo framework, is it ok to close this?

PR #153, opened 2021-08-27
Make the API server an importable package

clarkmcc:
I sent a [Slack message](https://kubernetes.slack.com/archives/C021U8WSAFK/p1630021405002500) indicating that it would be useful if the KCP project could be structured in a way that we could easily combine both the KCP API server and our own custom controllers into a single binary. Currently, the KCP API server is constructed in the `cmd/kcp/kcp.go` file and is configured only with command line flags which is not conducive to importing into your own projects.

This PR aims to address a few things
* Moves the core construction of the API server to a new package (`pkg/server`) and replaces all the construction code in `cmd/kcp/kcp.go` with an implementation of the new API server package.
* Adds the ability to customize the API server by either passing a `*pflag.FlagSet` or a struct with all the configuration values you want.
* Adds default config constructor for quickly initializing an API server config based on the defaults this project uses currently
* Adds the ability to register custom `PostStartHookFunc`s and `PreShutdownHookFunc`s that get passed to the API server. This will be useful when registering custom controllers, where the controller will need access to the client config for connecting to the API server.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -1,0 +1,301 @@
+package server
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"github.com/kcp-dev/kcp/pkg/etcd"
+	"github.com/kcp-dev/kcp/pkg/reconciler/apiresource"
+	"github.com/kcp-dev/kcp/pkg/reconciler/cluster"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"io/ioutil"
+	genericapiserver "k8s.io/apiserver/pkg/server"
+	"k8s.io/apiserver/pkg/storage/storagebackend"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"k8s.io/kubernetes/pkg/genericcontrolplane"
+	"k8s.io/kubernetes/pkg/genericcontrolplane/clientutils"
+	"k8s.io/kubernetes/pkg/genericcontrolplane/options"
+	"net"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strconv"
+)
+
+// Server manages the configuration and kcp api-server. It allows callers to easily use kcp
+// as a library rather than as a single binary. Using its constructor function, you can easily
+// setup a new api-server and start it:
+//
+//   srv := server.NewServer(server.DefaultConfig())
+//   srv.Run(ctx)
+//
+// You may optionally provide PostStartHookFunc and PreShutdownHookFunc hooks before starting
+// the server that should be passed to the api-server itself. These hooks have access to a
+// restclient.Config which allows you to easily create a client.
+//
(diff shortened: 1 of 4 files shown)


pkg/server/server.go
```diff
@@ -0,0 +1,285 @@
...
+	"k8s.io/kubernetes/pkg/genericcontrolplane/clientutils"
+	"k8s.io/kubernetes/pkg/genericcontrolplane/options"
+	"net"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strconv"
+)
+
+type Server struct {
```
davidfestal, line 28: What about adding comments on the public declarations in this file: `Server`, `Server.Run` and `NewServer` ?

Having a minimal example of how to start a server from a config would be great (even though it's quite obvious).

davidfestal (approved):
Great change which will make it easier to use KCP as a library.
I'm approving, though it might be nice to add some documentation in the new `server` package (see my related review comments).
Another nice-to-have would be some documentation, in `DEVELOPMENT.md` or in a dedicated `.md` file, that would explain how to use this API to bundle your own controllers with KCP as a single binary. But this might also be added in a next PR.

pkg/server/server.go
clarkmcc: I added clarifying comments with usage examples in the comments. Feel free to add or suggest any corrections to them.

clarkmcc:
@davidfestal I added usage examples and documentation to the `DEVELOPMENT.md` file. It seems concise enough to fit well enough within the context of the `DEVELOPMENT.md` that it doesn't need it's own `.md` file (yet, imo).

pkg/server/server.go
```diff
@@ -0,0 +1,301 @@
...
+//	srv := server.NewServer(server.DefaultConfig())
+//  srv.Run(ctx)
+//
+// You may optionally provide PostStartHookFunc and PreShutdownHookFunc hooks before starting
+// the server that should be passed to the api-server itself. These hooks have access to a
+// restclient.Config which allows you to easily create a client.
+//
+//  srv.AddPostStartHook("my-hook", func(context genericapiserver.PostStartHookContext) error {
+//		client := clientset.NewForConfigOrDie(context.LoopbackClientConfig)
+//	})
```
davidfestal, line 41: Are there some formatting problems here ?

pkg/server/server.go
davidfestal: Thanks !

davidfestal: However it seems there are some formatting problems in the example code of the `Server` documentation

pkg/server/server.go
clarkmcc: Interesting, it looks fine in my IDE. I'll commit a few things and see if I can get it right.

clarkmcc: Alright, we should be good now

https://github.com/kcp-dev/kcp/pull/153/files#diff-104a11712684e520f0affa95e6a053faeb713306a10f69ee99b9729cc9dd1996R35-R41

pkg/server/server.go
davidfestal: Nice !

PR #155, opened 2021-09-06
Fixes "resources_to_sync" flag as it was type mismatched when parsing the flags.

jmprusi:
The var resources_to_sync was not being used as when trying to parse it, it was parsed as type `StringSlice` while being defined as `StringArray`.

``` 
	resourcesToSync, err := flags.GetStringSlice("resources_to_sync")
	if err == nil {
		cfg.ResourcesToSync = resourcesToSync
	}
```

I changed the var definition instead that the parsing as with StringSlice we can define multiple resources separated by commas and/or by adding extra `resouce_to_sync` flags.:

```
--resources_to_sync=deployments.apps --resources_to_sync=ingresses.networking.k8s.io,ingressclasses.networking.k8s.io
```


davidfestal (approved):
Thanks for spotting this and fixing this !

PR #164, opened 2021-09-15
hack: use an automated tool to ensure import order

stevekuznetsov:
Nobody likes reviews that ask for re-ordering of import statements.
Corey's tool will correctly order stdlib/external/internal and runs very
quickly, solving this forever.


Makefile
```diff
@@ -13,3 +13,7 @@ vendor:
 codegen:
 	./hack/update-codegen.sh
 .PHONY: codegen
+
+.PHONY: imports
+imports:
+	go run github.com/coreydaley/openshift-goimports/ -m github.com/kcp-dev/kcp
```
imjasonh, line 19: I'd love to be able to `go install` this at a given tag, instead of pulling from HEAD. If the semantics change someday we'll pick them up immediately, and worse, if someone compromises that repo and pushes malicious code, we'll also pick that up immediately.

(Not a blocker, just a vague wish for a better world)

stevekuznetsov: Don't we get that behavior implicitly as a specific hash is called out in `go.mod`?

imjasonh: I think you might only get that if you add it to tools.go (which is also a terrible hack we should get rid of; CI tooling shouldn't be included in go.mod)

We can be sure with:

```
go install github.com/coreydaley/openshift-goimports@b92214262c6ce8598aefdee87aae6b8cf1a9fc86 && \
    openshift-goimports -m github.com/kcp-dev/kcp
```

PR #166, opened 2021-09-20
Adding downward API for the namespace of the syncer iamge

### LINKED ISSUE #83: Don't sync anything in the syncer-system namespace
The in-cluster syncer runs in `syncer-system`, and today if you create something in that namespace in kcp, the syncer will apply it to its own local cluster, possibly even replacing/breaking its own deployment.

Let's just keep syncer from messing with its own namespace entirely, to avoid a whole bunch of problems.

shawn-hurley:
* Removing processing of items in the same namespace as the syncer.

fixes #83

### DIFF
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -3,6 +3,7 @@ package syncer
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
@@ -23,6 +24,7 @@ import (
 )
 
 const resyncPeriod = 10 * time.Hour
+const SyncerNamespaceKey = "SYNCER_NAMESPACE"
 
 type Syncer struct {
 	specSyncer   *Controller
@@ -52,6 +54,7 @@ func StartSyncer(upstream, downstream *rest.Config, resources sets.String, clust
 	}
 	specSyncer.Start(numSyncerThreads)
 	statusSyncer.Start(numSyncerThreads)
+
 	return &Syncer{
 		specSyncer:   specSyncer,
 		statusSyncer: statusSyncer,
@@ -76,6 +79,8 @@ type Controller struct {
 
 	upsertFn UpsertFunc
 	deleteFn DeleteFunc
+
+	namespace string
 }
 
 // New returns a new syncer Controller syncing spec from "from" to "to".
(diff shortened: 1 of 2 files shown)


shawn-hurley:
@imjasonh @davidfestal 

I think that this may be one solution. I think the other solution is to get the object meta and check the namespace in the `eventhandler`, this would I think keep workers from getting items but I didn't know if this was something that was desired, as I was wondering if the tombstone check could also be there?

Let me know if you would like this approach and I can change it for that.

pkg/syncer/syncer.go
```diff
@@ -23,6 +24,7 @@ import (
 )
 
 const resyncPeriod = 10 * time.Hour
+const SYNCER_NAMESPACE_KEY = "SYNCER_NAMESPACE"
```
imjasonh, line 27: style nit: typically this would just be `SyncerNamespaceKey`

pkg/syncer/syncer.go
```diff
@@ -91,8 +96,9 @@ func New(from, to *rest.Config, upsertFn UpsertFunc, deleteFn DeleteFunc, handle
 
 		stopCh: stopCh,
 
-		upsertFn: upsertFn,
-		deleteFn: deleteFn,
+		upsertFn:  upsertFn,
+		deleteFn:  deleteFn,
+		namespace: os.Getenv(SYNCER_NAMESPACE_KEY),
```
imjasonh, line 101: How should syncer react when this env var is unset? Should it panic, or fallback to some default?

pkg/syncer/syncer.go
```diff
@@ -337,3 +347,16 @@ func (c *Controller) getClient(gvr schema.GroupVersionResource, namespace string
 	}
 	return nri
 }
+
+func (c *Controller) inSyncerNamesapce(objectNamespace string) bool {
```
imjasonh, line 351: typo: `inSyncerNamespace`

imjasonh (commented):
This approach seems fine to me, at least for now. Other options definitely exist, but none stick out to me as noticeably better. Thanks for taking this on!

pkg/syncer/syncer.go
shawn-hurley: that should be I thought covered in `inSyncerNamespace`

davidfestal (approved):

imjasonh (approved):

PR #169, opened 2021-09-22
Include core kcp resources in syncing

davidfestal:
Include Core KCP resources into what can be synchronized with a physical cluster

And add fixes for some small bugs at the same time.

### DIFF
--- a/pkg/reconciler/cluster/controller.go
+++ b/pkg/reconciler/cluster/controller.go
@@ -13,13 +13,16 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	k8serorrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/discovery"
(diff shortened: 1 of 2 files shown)


pkg/reconciler/cluster/controller.go
```diff
@@ -280,15 +317,24 @@ func (c *Controller) deletedCluster(obj interface{}) {
...
+	ex, err := os.Executable()
+	if err != nil {
+		return err
+	}
+	if path.Base(ex) == "__debug_bin" {
+		ex = path.Dir(path.Dir(ex))
+	}
+
+	configDir := path.Join(path.Dir(ex), "config")
+	files, err := ioutil.ReadDir(configDir)
```
imjasonh, line 329: Can you add a comment describing what this is doing? It seems like this will only work if the binary is being `go run`?

If we want to embed all these yaml files in here, we can either use `//go:embed` or take advantage of [`ko`'s build-time resource injection](https://github.com/google/ko#static-assets), which will make the files available at `$KO_DATA_PATH`.

davidfestal: Yeah; it was just a way to make this a bit more robust than what was previously (it used to only work when the current directory was the KCP root directory).
In fact I added this to debug my PR main implementation more easily :-)
But you're right, so I'll remove it from the current PR, and would create a dedicated PR with a clean implementation based on `go:embed`

davidfestal: Removed from this PR

imjasonh (approved):

PR #170, opened 2021-09-22
Use `go:embed` to embed KCP CRDs

davidfestal:
This allows registering KCP CRDs (located inside the `config` folder) at KCP start, whatever be the current directory in which KCP is started.

### DIFF
--- a/embed.go
+++ b/embed.go
@@ -1,0 +1,12 @@
+package kcp
+
+import "embed"
+
+// This file should ONLY be used to define variables with the content of data files
+// accessible from the root of the source code, that should be embedded inside
+// GO code through go:embed annotations
+// It is required to define this file at the root of the module since
+// go:embed cannot reference files in parent directories relatively to the current package.
+
+//go:embed config
+var ConfigDir embed.FS
--- a/pkg/reconciler/cluster/controller.go
+++ b/pkg/reconciler/cluster/controller.go
@@ -3,7 +3,7 @@ package cluster
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
+	"path"
 	"reflect"
 	"strings"
 	"time"
@@ -25,6 +25,7 @@ import (
 	"k8s.io/kubernetes/pkg/api/genericcontrolplanescheme"
 	"sigs.k8s.io/yaml"
 
+	kcp "github.com/kcp-dev/kcp"
 	apiresourcev1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apiresource/v1alpha1"
 	clusterv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/cluster/v1alpha1"
 	versionedclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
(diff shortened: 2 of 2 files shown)


embed.go
```diff
@@ -0,0 +1,61 @@
+package kcp
```
imjasonh, line 1: I don't think we want this at the top level. Maybe just keep it in pkg/reconciler/cluster?

embed.go
```diff
@@ -0,0 +1,61 @@
...
+		if err != nil {
+			return err
+		}
+		crd := &apiextensionsv1.CustomResourceDefinition{}
+		err = yaml.Unmarshal(bytes, crd)
+		if err != nil {
+			return err
+		}
+
+		_, err = crdClient.CustomResourceDefinitions().Create(context.TODO(), crd, metav1.CreateOptions{})
```
imjasonh, line 45: style nit:

```
if _, err := crdClient...Create(); k8serrors.IsAlreadyExists(err) {
  ...
} else if err != nil {
  return nil
}
```

This makes it easier to tell at a glance that the error doesn't escape the scope, and won't shadow a previous error.

Same with `err = yaml.Unmarshal` above:

```
if err := yaml.Unmarshal(...); err != nil {
  return err
}
```

This is mainly just hygiene and a good practice to follow, neither change will have a functional effect on this code.

embed.go
davidfestal: Well, that the easiest way to be able to embed the content of the `/config` directory, since you cannot embed files at a lower level than the current package (`../` is not supported). And I'm not sure moving `/config` at a package level is a good idea either.

embed.go
davidfestal: Just forced-pushed to fix this.

embed.go
imjasonh: Hmm, this is really unfortunate. 😞 

Maybe we could symlink `config/` to `pkg/reconciler/cluster/config/`? Does //go:embed` chase symlinks? 🤞

davidfestal: I'm not a fan of the symlink solution TBF.
I forced pushed (in commit 80e1b86d5fef7c95a23de2bf7c29f0ec6038c2df) a fix that seems quite a compromise: just have an `embed.go` file at the root of the module, that only defines global variables, and use those variables in code placed in the regular packages.

pkg/reconciler/cluster/controller.go
```diff
@@ -315,39 +316,36 @@ func (c *Controller) deletedCluster(obj interface{}) {
...
+			if !strings.HasSuffix(file.Name(), "yaml") {
+				continue
 			}
-			crd.ResourceVersion = existingCRD.ResourceVersion
-			if _, err := crdClient.CustomResourceDefinitions().Update(context.TODO(), crd, metav1.UpdateOptions{}); err != nil {
+			if bytes, err := kcp.ConfigDir.ReadFile(path.Join("config", file.Name())); err != nil {
 				return err
+			} else {
+				crd := &apiextensionsv1.CustomResourceDefinition{}
+				if err := yaml.Unmarshal(bytes, crd); err != nil {
```
imjasonh, line 330: Not blocking the PR, but how does this behave when a non-CRD otherwise-valid-YAML file sneaks into the `config/` directory?

imjasonh (approved):

embed.go
imjasonh: This is a good compromise, thanks for thinking of it. 👍

pkg/reconciler/cluster/controller.go
davidfestal: Very good remark, that we'll have to think about. Maybe we should open a new issue for it.

PR #171, opened 2021-09-22
Adding namespace controller to minimal kube controllers

### LINKED ISSUE #165: Namespace deletion (via kubectl) hangs
**Repro steps**
```shell
make build
./bin/kcp start

# In a separate shell:
export KUBECONFIG=.kcp/data/admin.kubeconfig
kubectl create namespace some-test
kubectl delete namespace some-test
```

**Expected**
The namespace should be deleted immediately (there is nothing in it)

**Actual**
`kubectl` says `namespace some-test deleted`, but then it never returns.
If `kubectl` is interrupted (`^C`) and re-run, the same thing happens.

shawn-hurley:
* Adding this to the Server hooks because I don't think you want this
  running without it

fixes #165

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -12,13 +12,20 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"time"
 
 	clientv3 "go.etcd.io/etcd/client/v3"
 
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	"k8s.io/apiserver/pkg/storage/storagebackend"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/metadata"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -300,5 +310,34 @@ func (s *Server) AddPreShutdownHook(name string, hook genericapiserver.PreShutdo
...
+func (s *Server) startNamespaceController(hookContext genericapiserver.PostStartHookContext) error {
+	kubeClient, err := kubernetes.NewForConfig(hookContext.LoopbackClientConfig)
+	if err != nil {
+		return err
+	}
+	metadata, err := metadata.NewForConfig(hookContext.LoopbackClientConfig)
+	if err != nil {
+		return err
+	}
+	versionedInformer := informers.NewSharedInformerFactory(kubeClient, resyncPeriod)
```
shawn-hurley, line 331: @davidfestal I am not sure if this is the correct thing to do with the informer. I assume that we will eventually add more controllers that should use the versionedInformer and we may want to manage it at the `server` struct level?

davidfestal (approved):
Thanks !

imjasonh (approved):

PR #175, opened 2021-09-24
Shard-aware transparent LIST/WATCH

stevekuznetsov:
This commit implements a shard-aware transparent LIST and WATCH calls. Lots of
hacks in here:

- we are still running authn in the kcp handler chain, but for
  multi-shard kcp deployments we will need an external authn solution
  that is trusted by all ... for now the shard-aware delegator sits
  after auth{n,z} and uses impersonation to hack in something usable
- the encoded continue tokens do not enforce a specific ordering between
  shards but do (by side-effect of the impl) fully list one shard before
  moving on to another
- shards are identified by DN/host but each individual shard will
  disagree about what to call itself as all shards see a loopback
  address as their own
- no index of apiVersion to shards containing resources at that version
  is implemented
- no index of workspace to shard(s) containing that workspace exists and
  all sharded requests look at all workspaces by default


pkg/sharding/handler.go
```diff
@@ -0,0 +1,202 @@
...
+		Groups:   userInfo.GetGroups(),
+		Extra:    userInfo.GetExtra(),
+	}
+	transport, err := rest.TransportFor(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("could not get transport for config: %w", err)
+	}
+
+	var httpClient *http.Client
+	if transport != http.DefaultTransport {
```
imjasonh, line 169: I think we just always want to do this, both to enforce the timeout even in this rare case, and so that `httpClient.Do` doesn't nil-panic on line 242.

pkg/sharding/handler.go
```diff
@@ -0,0 +1,202 @@
...
+			for _, item := range t.Items {
+				item.SetClusterName(fmt.Sprintf("%s/clusters/%s", shard, item.GetClusterName()))
+			}
+			body := bytes.Buffer{}
+			if err := unstructured.UnstructuredJSONScheme.Encode(t, &body); err != nil {
+				klog.Warningf("could not encode sharded response body: %v", err)
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+			resp.Header.Set("Content-Length", strconv.Itoa(body.Len()))
```
imjasonh, line 140: This shouldn't be necessary, it should already be set to the length of the contents.

pkg/server/server.go
```diff
@@ -125,6 +129,18 @@ func (s *Server) Run(ctx context.Context) error {
 			KeyFile:       cfg.KeyFile,
 			TrustedCAFile: cfg.TrustedCAFile,
 		}
+
+		injector := make(chan *rest.Config)
+		clientLoader, err := sharding.New(s.cfg.ShardKubeconfigFile, injector)
```
davidfestal, line 134: How does it behave when `shard-kubeconfig-file` is empty ? Are we compatible with the default, non-sharding behavior ?

pkg/server/config.go
```diff
@@ -109,4 +130,10 @@ func AddConfigFlags(flags *pflag.FlagSet) {
 		"TLS certification file used to secure etcd communication.")
 	flags.String("etcd-cafile", "",
 		"TLS Certificate Authority file used to secure etcd communication.")
+	flags.String("shard-kubeconfig-file", "",
+		"Kubeconfig holding admin(!) credentials to peer kcp shards.")
+	flags.String("root_directory", ".kcp",
```
sttts, line 135: odd mix of _ and - in here. Any rule when to use what?

pkg/sharding/apiserver/apiserver.go
```diff
@@ -0,0 +1,109 @@
...
+		c.GenericConfig.Version = &version.Info{
+			Major: "0",
+			Minor: "1",
+		}
+	}
+
+	return CompletedConfig{&c}
+}
+
+// New returns a new instance of CustomResourceDefinitions from the given config.
```
sttts, line 94: left-over comment from apiextensions-apiserver

pkg/sharding/apiserver/handler.go
```diff
@@ -0,0 +1,359 @@
...
+	default:
+		responsewriters.ErrorNegotiated(
+			apierrors.NewMethodNotSupported(schema.GroupResource{Group: requestInfo.APIGroup, Resource: requestInfo.Resource}, requestInfo.Verb),
+			Codecs, schema.GroupVersion{Group: requestInfo.APIGroup, Version: requestInfo.APIVersion}, w, req,
+		)
+		handlerFunc = nil
+	}
+	if handlerFunc != nil {
+		handlerFunc = metrics.InstrumentHandlerFunc(requestInfo.Verb, requestInfo.APIGroup, requestInfo.APIVersion, requestInfo.Resource, requestInfo.Subresource, metrics.CleanScope(requestInfo), metrics.APIServerComponent, false, "", handlerFunc)
+		handler := genericfilters.WithWaitGroup(handlerFunc, longRunningFilter, h.wg)
```
sttts, line 218: for what is that needed? It's something to synchronize audit with the request chain. I don't think you care here.

pkg/sharding/apiserver/handler.go
```diff
@@ -0,0 +1,359 @@
...
+		return
+	}
+	equivalentResourceRegistry := runtime.NewEquivalentResourceRegistry()
+	groupVersion := schema.GroupVersion{Group: requestInfo.APIGroup, Version: requestInfo.APIVersion}
+	resource := groupVersion.WithResource(requestInfo.Resource)
+
+	SchemeLocker.Lock()
+	Scheme.AddUnversionedTypes(groupVersion, unversionedTypes...)
+	SchemeLocker.Unlock()
+	SchemeLocker.RLock()
```
sttts, line 71: ouch. is this to dynamically extend the scheme?

pkg/sharding/apiserver/handler.go
```diff
@@ -0,0 +1,359 @@
...
+		return rows, nil
+	}); err != nil {
+		utilruntime.HandleError(err)
+		responsewriters.ErrorNegotiated(
+			apierrors.NewInternalError(err),
+			Codecs, schema.GroupVersion{Group: requestInfo.APIGroup, Version: requestInfo.APIVersion}, w, req,
+		)
+		return
+	}
+	tableConverter := storage.TableConvertor{TableGenerator: tableGenerator}
```
sttts, line 129: why having this in the critical path ?

pkg/sharding/apiserver/storage.go
```diff
@@ -0,0 +1,423 @@
...
+		return nil, "", fmt.Errorf("failed to create sharded request: %w", err)
+	}
+	request.SetHeader("X-Kubernetes-Cluster", m.GetClusterName())
+	request.OverwriteParam("resourceVersion", strconv.FormatInt(resourceVersion, 10))
+	request = request.Body(obj)
+	return request, identifier, nil
+}
+
+func (s *shardedStorage) Get(ctx context.Context, name string, options *metav1.GetOptions) (runtime.Object, error) {
+	// TODO: figure out how the client is going to choose a shard here
```
sttts, line 118: why should the client choose a shard?

pkg/sharding/apiserver/storage.go
```diff
@@ -0,0 +1,423 @@
...
+		case *unstructured.Unstructured:
+			var conversionErr error
+			list, conversionErr = r.ToList()
+			if conversionErr != nil {
+				return nil, fmt.Errorf("could not convert sharded response from %T to a list: %w", result, conversionErr)
+			}
+		default:
+			return nil, fmt.Errorf("could not parse sharded response as a list, got %T", result)
+		}
+		if err := state.UpdateWith(shard, list); err != nil {
```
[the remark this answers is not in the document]
p0lyn0mial: oh, sorry, this is a state from the query param.

pkg/sharding/apiserver/storage.go
```diff
@@ -0,0 +1,423 @@
...
+		}
+		request, err := s.requestFor(client)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create sharded request: %w", err)
+		}
+		request.OverwriteParam("limit", "500")
+		request.OverwriteParam("continue", continueToken)
+		result, err := request.Do(ctx).Get()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get data from shard: %w", err)
```
p0lyn0mial, line 317: would it make sense to log the shard identifier?

stevekuznetsov:
Some thoughts:

- we need a better method for hacking on the dependencies here, or merging becomes a dance of "get this in first, re-hack the go.mod while  you work" etc 
- do we want to add e2e at all to presubmits? I am currently starting kcp and testing it like so:

```
(kcp) $ make build && ./contrib/demo/startKcpAndClusterController.sh -auto_publish_apis=true .configmaps
(kubernetes) $ make WHAT=test/e2e/e2e.test && _output/local/bin/linux/amd64/e2e.test --provider local --kubeconfig .kcp/data/admin.kubeconfig --ginkgo.focus "Watchers" --e2e-verify-service-account=false
```

pkg/sharding/apiserver/state.go
```diff
@@ -0,0 +1,306 @@
...
+	// ShardResourceVersion is the version at which the list of shards
+	// that needed to be queried was resolved.
+	ShardResourceVersion int64 `json:"srv"`
+	// ResourceVersions hold versions for individual shards being queried.
+	ResourceVersions []ShardedResourceVersion `json:"rvs"`
+}
+
+type ShardedResourceVersion struct {
+	// Identifier is used to locate credentials for this shard
+	Identifier      string `json:"id"`
```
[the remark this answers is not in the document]
sttts: ack, RV size certainly matters. Sparse makes sense.

pkg/sharding/apiserver/storage.go
```diff
@@ -0,0 +1,423 @@
...
+		delegates: []stopper{},
+		events:    make(chan watch.Event),
+		wg:        &sync.WaitGroup{},
+		state:     state,
+		lock:      &sync.Mutex{},
+	}
+	for identifier := range delegates {
+		w.wg.Add(1)
+		go func(identifier string, events <-chan watch.Event) {
+			defer w.wg.Done()
```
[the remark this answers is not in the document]
p0lyn0mial: https://github.com/kubernetes/kubernetes/blob/af7cf4abc6bfeb0d2cfaca76097cf7a0603c4495/staging/src/k8s.io/apimachinery/pkg/util/runtime/runtime.go#L45

pkg/sharding/apiserver/storage.go
```diff
@@ -0,0 +1,423 @@
...
+		return
+	}
+	a.lock.Lock()
+	defer a.lock.Unlock()
+	if err := a.state.UpdateWith(identifier, obj); err != nil {
+		a.events <- watch.Event{
+			Type:   watch.Error,
+			Object: &errors.NewInternalError(fmt.Errorf("failed to update resource version vector clock: %v", err)).ErrStatus,
+		}
+		return
```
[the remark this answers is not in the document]
p0lyn0mial: Closing the watch would be preferred but it would also complicate the code. I think that client-go stops on error not sure about other libraries.

pkg/sharding/apiserver/storage.go
```diff
@@ -0,0 +1,432 @@
...
+			return nil, fmt.Errorf("could not format new resource version: %w", err)
+		}
+		list.SetResourceVersion(updatedResourceVersion)
+		updatedContinueToken, err := state.Encode()
+		if err != nil {
+			return nil, fmt.Errorf("could not format new continue token: %w", err)
+		}
+		list.SetContinue(updatedContinueToken)
+		for i := range list.Items {
+			list.Items[i].SetResourceVersion(updatedResourceVersion)
```
p0lyn0mial, line 355: Is this correct? Shouldn't we get an RV of the item and change it instead?

pkg/sharding/apiserver/storage.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+type storageMux struct {
+	storageBase
+	sharded listWatcher
+	delegating rest.StandardStorage
+}
+
+func (s *storageMux) Get(ctx context.Context, name string, options *metav1.GetOptions) (runtime.Object, error) {
+	if c := request.ClusterFrom(ctx); c.Wildcard {
```
p0lyn0mial, line 55: It seems that the master branch points to Kube 1.22 but I didn't find the `ClusterFrom` method in `k8s.io/apiserver/pkg/endpoints/request` package.

pkg/sharding/apiserver/handler.go
```diff
@@ -0,0 +1,451 @@
...
+			)
+			return
+		}
+		req.Body = io.NopCloser(&buf)
+		cloned.Body = io.NopCloser(bytes.NewReader(buf.Bytes()))
+	}
+
+	klog.Info("v=== INCOMING ===v")
+	v, e := httputil.DumpRequest(cloned, true)
+	klog.Info(string(v))
```
p0lyn0mial, line 215: might be too verbose

pkg/sharding/handler.go
```diff
@@ -0,0 +1,35 @@
...
+		}
+		if sharded := req.Header.Get("X-Kubernetes-Sharded-Request"); sharded == "true" {
+			// we're being asked for our own data by some other shard, no need to fan out
+			apiHandler.ServeHTTP(w, req)
+			return
+		}
+		loader.RLock()
+		clients := loader.Clients
+		loader.RUnlock()
+		handler := apiserver.NewShardedHandler(clients, 0, 10*time.Minute)
```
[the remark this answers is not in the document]
p0lyn0mial: 10 minutes is quite a large timeout.

pkg/sharding/apiserver/storage.go
p0lyn0mial: i suspect the method might be defined in the fork `github.com/kcp-dev/kubernetes/staging/`

pkg/sharding/apiserver/request.go
```diff
@@ -0,0 +1,74 @@
...
+			if err := unstructured.UnstructuredJSONScheme.Encode(obj, encoded); err != nil {
+				return nil, fmt.Errorf("could not encode mutated body: %w", err)
+			}
+			body = encoded.Bytes()
+		}
+		request := rest.NewRequest(client).
+			Verb(req.Method).
+			RequestURI(req.URL.RequestURI()).
+			SetHeader("X-Kubernetes-Sharded-Request", "true").
+			Body(body).Debug()
```
p0lyn0mial, line 46: Instead of a custom `Debug` method added to the forked repo, we could provide a custom RoundTripper defined in this repo.

pkg/sharding/apiserver/sharded_storage.go
```diff
@@ -0,0 +1,231 @@
...
+	for i := range state.ResourceVersions {
+		client, err := s.clientFor(s.shards[state.ResourceVersions[i].Identifier])
+		if err != nil {
+			return nil, fmt.Errorf("failed to create sharded client: %w", err)
+		}
+		request, err := s.requestFor(client)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create sharded request: %w", err)
+		}
+		request.OverwriteParam("limit", "500")
```
p0lyn0mial, line 48: the default client-go limit is `500` not sure why we have to overwrite it here?

pkg/sharding/clients.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+type IdentifiedConfig struct {
+	Identifier string
+	Config     *rest.Config
+}
+
+type ClientLoader struct {
+	*sync.RWMutex
+	Clients map[string]*rest.Config
```
ncdc, line 19: I think if you're embedding a mutex in this struct and using it to guard the clients map, you wouldn't want to export Clients to users, as that would allow them to ignore the mutex.

I'd do Clients -> clients, and optionally decide if you want ClientLoader.Lock/Unlock to be a thing, or if you'd rather present a finer-grained API to consumers (e.g. GetClient or whatever you'd like to call it).

pkg/sharding/handler.go
```diff
@@ -0,0 +1,35 @@
...
+			apiHandler.ServeHTTP(w, req)
+			return
+		}
+		if sharded := req.Header.Get("X-Kubernetes-Sharded-Request"); sharded == "true" {
+			// we're being asked for our own data by some other shard, no need to fan out
+			apiHandler.ServeHTTP(w, req)
+			return
+		}
+		loader.RLock()
+		clients := loader.Clients
```
ncdc, line 29: AFAIK this is a shallow copy, so it'll potentially be subject to data races.

ncdc:
@stevekuznetsov what is your thinking (long term) for how client kubeconfigs are propagated among all the shards, as membership changes, client certs are rotated, etc.?

pkg/sharding/clients.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func New(delegates string, injector <-chan IdentifiedConfig) (*ClientLoader, error) {
```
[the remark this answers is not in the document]
stevekuznetsov: Why? I need to synchronize the producer and consumer anyway

pkg/sharding/apiserver/sharded_storage.go
```diff
@@ -0,0 +1,231 @@
...
+			return nil, fmt.Errorf("could not update sharded state: %w", err)
+		}
+		updatedResourceVersion, err := state.ToResourceVersions().Encode()
+		if err != nil {
+			return nil, fmt.Errorf("could not format new resource version: %w", err)
+		}
+		list.SetResourceVersion(updatedResourceVersion)
+		for i := range list.Items {
+			uniqueVersion := state.ToResourceVersions()
+			if err := uniqueVersion.UpdateWith(shard, &list.Items[i]); err != nil {
```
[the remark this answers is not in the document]
stevekuznetsov: Covers what? Issuing an UPDATE with a RV precondition for an item from a LIST? Yes, that's in the workflow test.

pkg/sharding/apiserver/request.go
stevekuznetsov: Pros/cons?

pkg/sharding/apiserver/handler.go
stevekuznetsov: Yeah, I have a commit that pushes these to V(10) I will un-revert before merge

pkg/sharding/apiserver/storage.go
stevekuznetsov: All I know is it's compiling ... uh... ???

stevekuznetsov: (and this predates these changes)

pkg/sharding/apiserver/request.go
```diff
@@ -0,0 +1,74 @@
...
+		cloned := rest.CopyConfig(cfg)
+		cloned.Impersonate = rest.ImpersonationConfig{
+			UserName: userInfo.GetName(),
+			Groups:   userInfo.GetGroups(),
+			Extra:    userInfo.GetExtra(),
+		}
+		cloned.ContentConfig.NegotiatedSerializer = negotiatedSerializer
+		// we handle conversions on our end for the client
+		cloned.ContentConfig.ContentType = "application/json"
+		return rest.UnversionedRESTClientFor(cloned)
```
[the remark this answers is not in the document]
stevekuznetsov: If this becomes a scaling problem we can tackle it. I would prefer to use typed, well-known options rather than going underneath that layer to add headers manually.

pkg/sharding/handler.go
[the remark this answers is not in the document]
stevekuznetsov: If we hit issues with scaling, we can handle it.

pkg/sharding/clients.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to load: %w", err)
+	}
+	for context := range cfg.Contexts {
+		contextCfg, err := clientcmd.NewNonInteractiveClientConfig(*cfg, context, &clientcmd.ConfigOverrides{}, loader).ClientConfig()
+		if err != nil {
+			return nil, fmt.Errorf("create %s client: %w", context, err)
+		}
+		contextCfg.ContentType = "application/json"
+		l.Lock()
```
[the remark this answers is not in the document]
stevekuznetsov: This guards the reader

pkg/sharding/clients.go
```diff
@@ -0,0 +1,53 @@
...
+		l.Clients[genericcontrolplane.SanitizeClusterId(context)] = contextCfg
+		l.Unlock()
+	}
+
+	l.Lock()
```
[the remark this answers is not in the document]
stevekuznetsov: This guards the reader

go.mod
```diff
@@ -19,37 +21,39 @@ require (
...
-	k8s.io/kubectl => github.com/kcp-dev/kubernetes/staging/src/k8s.io/kubectl v0.0.0-20210812085835-84c07cbd39b4
-	k8s.io/kubelet => github.com/kcp-dev/kubernetes/staging/src/k8s.io/kubelet v0.0.0-20210812085835-84c07cbd39b4
-	k8s.io/kubernetes => github.com/kcp-dev/kubernetes v0.0.0-20210812085835-84c07cbd39b4
-	k8s.io/legacy-cloud-providers => github.com/kcp-dev/kubernetes/staging/src/k8s.io/legacy-cloud-providers v0.0.0-20210812085835-84c07cbd39b4
-	k8s.io/metrics => github.com/kcp-dev/kubernetes/staging/src/k8s.io/metrics v0.0.0-20210812085835-84c07cbd39b4
-	k8s.io/mount-utils => github.com/kcp-dev/kubernetes/staging/src/k8s.io/mount-utils v0.0.0-20210812085835-84c07cbd39b4
-	k8s.io/pod-security-admission => github.com/kcp-dev/kubernetes/staging/src/k8s.io/pod-security-admission v0.0.0-20210812085835-84c07cbd39b4
-	k8s.io/sample-apiserver => github.com/kcp-dev/kubernetes/staging/src/k8s.io/sample-apiserver v0.0.0-20210812085835-84c07cbd39b4
+	k8s.io/api => /home/stevekuznetsov/code/kcp-dev/kubernetes/src/github.com/kcp-dev/kubernetes/staging/src/k8s.io/api
```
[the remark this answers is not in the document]
stevekuznetsov: Will do before merge

pkg/sharding/handler.go
stevekuznetsov: Good catch, should be full copy once we expect the membership to change

stevekuznetsov:
@ncdc it depends on whether we deploy within a kcp or alongside them, I imagine a inotify-based directory watcher for kubeconfigs is simple enough, lots of prior art for that when we need it.

ncdc:
Gotcha. And so there would be some other mechanism to distribute the kubeconfigs? (If you want to defer discussing this and add it to the list of TODOs, that's fine too)

stevekuznetsov:
Other than through the filesystem? WDYM?

ncdc:
I'm thinking about the operational & management side of things. I'm an admin for the shard fleet, I need to add a new shard... what are all the things I have to do to fully "register" that new shard and make all the existing shards be able to talk to it? Who/what/how does the new shard's kubeconfig make it onto all the other shard's file systems?

pkg/sharding/clients.go
ncdc: @stevekuznetsov we can merge this for now, but can you add a FIXME here?

pkg/sharding/handler.go
ncdc: + FIXME please 😄

PR #182, opened 2021-10-07
WIP: First sketch of `Workspace` and `WorkspaceShard` APIs

davidfestal:


davidfestal:
This is a draft PR as a basis for discussion. First points to discuss:
- The `group` is named `tenancy.kcp.dev`. Probably the wrong name. Proposals welcomed !
- The `Workspace` still seems somehow too much "general purpose" and would complete with meaning like IDE workspace, for example. Why not something like `cloudspace` ou `clusterspace` ?

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -0,0 +1,142 @@
...
+
+// WorkspaceShardSpec holds the desired state of the WorkspaceShard.
+type WorkspaceShardSpec struct {
+}
+
+// WorkspaceShardStatus communicates the observed state of the WorkspaceShard.
+type WorkspaceShardStatus struct {
+	// Set of integer resources that workspaces can be scheduled into
+	// +optional
+	Capacity []int `json:"capacity,omitempty"`
```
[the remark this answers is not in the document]
sttts: kube type for capacity: https://github.com/kubernetes/api/blob/master/core/v1/types.go#L5009

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -0,0 +1,142 @@
...
+	// +optional
+	ReadOnly bool `json:"readOnly,omitempty"`
+
+	Shard string `json:"shard"`
+
+	// +optional
+	TargetShard string `json:"targetShard,omitempty"`
+}
+
+// SharedStatus contains details for the current status of a workspace shard.
```
stevekuznetsov, line 55:
```suggestion
// ShardStatus contains details for the current status of a workspace shard.
```

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -0,0 +1,142 @@
...
+// +crd
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster,categories=kcp
+type Workspace struct {
+	metav1.TypeMeta `json:",inline"`
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
```
stevekuznetsov, line 34: How do I go from an object to the name of the workspace it resides in?

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// +optional
+	Spec WorkspaceSpec `json:"spec,omitempty"`
```
sttts, line 37: Make it required. An object without spec is odd. If kubebuilder wasn't broken, we could add a default `{}`.

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -0,0 +1,142 @@
...
+	Status WorkspaceStatus `json:"status,omitempty"`
+}
+
+// WorkspaceSpec holds the desired state of the Workspace.
+type WorkspaceSpec struct {
+
+	// +optional
+	ReadOnly bool `json:"readOnly,omitempty"`
+
+	Shard string `json:"shard"`
```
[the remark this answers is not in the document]
sttts: define some minimal length. Otherwise, `""` is a valid value.

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -0,0 +1,142 @@
...
+	Spec WorkspaceSpec `json:"spec,omitempty"`
+
+	// +optional
+	Status WorkspaceStatus `json:"status,omitempty"`
+}
+
+// WorkspaceSpec holds the desired state of the Workspace.
+type WorkspaceSpec struct {
+
+	// +optional
```
sttts, line 46: assuming you use kubebuilder's controller-gen, you probably have to use the required+optional markers from kubebuilder, not these from kube. They are ignored.

pkg/apis/tenancy/v1alpha1/types.go
davidfestal: Right. I'll change it

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -0,0 +1,142 @@
...
+//
+// +crd
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster,categories=kcp
+type Workspace struct {
+	metav1.TypeMeta `json:",inline"`
+	// +optional
```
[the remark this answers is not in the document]
davidfestal: This is mainly aligned on what is in Kubernetes native types:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/apps/v1/types.go#L51 for example.

pkg/apis/tenancy/v1alpha1/types.go
davidfestal: I assume the name of a workspace would be the name of the logical cluster that is used under the cover (or at least could be deduced from it). So you would be able to find the workspace an object belongs to by searching for the workspace resource whose name is equal to the `ClusterName` field of the object.

config/tenancy.kcp.dev_workspaceshards.yaml
```diff
@@ -0,0 +1,56 @@
...
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.5.0
+  creationTimestamp: null
+  name: workspaceshards.tenancy.kcp.dev
+spec:
+  group: tenancy.kcp.dev
+  names:
+    categories:
+    - kcp
+    kind: WorkspaceShard
```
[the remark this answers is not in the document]
davidfestal: I'm more concerned about the `Workspace` name itself TBF.

pkg/apis/tenancy/v1alpha1/types.go
ncdc: @sttts is it ok to resolve this?

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -0,0 +1,144 @@
...
+	// Phase of the workspace  (Initializing / Active / Terminating)
+	Phase WorkspacePhaseType `json:"phase,omitempty"`
+
+	// List of shards related to this workspace.
+	// The first shard in this list with LiveAfterResourceVersion is set is the active shard
+	//
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	Shards []ShardStatus `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"name"`
```
ncdc, line 91:
```suggestion
	Shards []ShardStatus `json:"shards,omitempty" patchStrategy:"merge" patchMergeKey:"name"`
```

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -0,0 +1,144 @@
...
+// WorkspaceShard describes a Shard (== KCP instance) on which a number of
+// workspaces will live
+//
+// +crd
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster,categories=kcp
+type WorkspaceShard struct {
```
ncdc, line 113: Do you have any more details on how this type will be used?

ncdc (changes requested):

davidfestal:
> Naming is hard. Can we possibly name this something other `Workspace`. There are so many neighbouring areas where Workspace is used that this is going to just cause confusion. Perhaps we can use `Space` as an alternate?

@jwforres @smarterclayton Do we really want to merge this PR with the `Workspace` name ? Are alternatives like `Space`, `CloudSpace`, or any equivalent still considered ?

pkg/apis/tenancy/v1alpha1/types.go
davidfestal: I guess it will mainly be used to identify a workspace shard with a name, and have a way to associate a capacity (by a controller I assume).
I also assume that we might have additional fields to add in the spec in the future.
@smarterclayton Would be great to have you complete the answer and correct me if I was wrong.

pkg/apis/tenancy/v1alpha1/types.go
davidfestal: Fixed in commit 522952de99a16b33b22df878b1a809011843e8ca

config/tenancy.kcp.dev_workspaces.yaml
```diff
@@ -0,0 +1,86 @@
...
+                type: boolean
+              shard:
+                minLength: 1
+                type: string
+              targetShard:
+                type: string
+            required:
+            - shard
+            type: object
+          status:
```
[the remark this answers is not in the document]
davidfestal: Fixed in commit 522952de99a16b33b22df878b1a809011843e8ca

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -0,0 +1,144 @@
...
+type WorkspaceSpec struct {
+
+	// +optional
+	ReadOnly bool `json:"readOnly,omitempty"`
+
+	// +kubebuilder:validation:MinLength=1
+	Shard string `json:"shard"`
+
+	// +optional
+	TargetShard string `json:"targetShard,omitempty"`
```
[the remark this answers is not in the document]
davidfestal: I added comments in commit 522952de99a16b33b22df878b1a809011843e8ca

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+// ShardStatus contains details for the current status of a workspace shard.
+type ShardStatus struct {
+	// Name of an active shard.
+	//
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
```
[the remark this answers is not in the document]
davidfestal: I assume it is. I made it clearer in the comment.

pkg/apis/tenancy/v1alpha1/types.go
davidfestal: @ncdc did my answer make sense to you, and can we resolve this, or do you want more details from @smarterclayton first ?

pkg/apis/tenancy/v1alpha1/types.go
sttts: if kubebuilder learned that by now, sure.

pkg/apis/tenancy/v1alpha1/types.go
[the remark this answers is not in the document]
sttts: not important. Technically it is wrong, but Kube also has it wrong. But there are no consequences of it, maybe only in OpenAPI spec.

pkg/apis/tenancy/v1alpha1/types.go
ncdc: I'm ok for now, so we can iterate quickly. I was mainly asking because the spec struct is currently empty.

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -0,0 +1,155 @@
...
+	// Name of an active WorkspaceShard.
+	//
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
+
+	// Resource version at which writes to this shard should not be accepted.
+	LiveBeforeResourceVersion string `json:"liveBeforeResourceVersion,omitempty"`
+
+	// Resource version after which writes can be accepted on this shard.
+	LiveAfterResourceVersion string `json:"liveAfterResourceVersion,omitempty"`
```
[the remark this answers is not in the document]
davidfestal: @smarterclayton Your insight would help here ...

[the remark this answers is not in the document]
sttts: I think the key is `Shards []ShardStatus`, i.e. a slice. There can be multiple of them, partitioning the RV space. So the point for understanding is probably that `LiveAfterResourceVersion < LiveBeforeResourceVersion` should hold, maybe worth an API comment.

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -0,0 +1,155 @@
...
+	Status WorkspaceStatus `json:"status,omitempty"`
+}
+
+// WorkspaceSpec holds the desired state of the Workspace.
+type WorkspaceSpec struct {
+
+	// +optional
+	ReadOnly bool `json:"readOnly,omitempty"`
+
+	// Name of the WorkspaceShard this Workspace lives on.
```
stevekuznetsov, line 50: Why are we not modelling this intent/reality with spec/status? e.g. Why do we need both the current and future shard in spec?

ncdc:
We had a [discussion](https://kubernetes.slack.com/archives/C021U8WSAFK/p1636476695212400) today in Slack about desired and current shard assignment, spec vs. status, failure/recovery modes. We've decided to merge this as-is, and follow up with changes from the discussion - move shard assignment out of spec and into status.

Issue #128, opened 2021-07-08
Increase coverage for pkg/schemacompat


imjasonh:
```
$ go test -cover ./pkg/schemacompat
ok  	github.com/kcp-dev/kcp/pkg/schemacompat	0.469s	coverage: 34.0% of statements
```

To see which lines aren't covered (GitHub won't let me attach it):

```
go test -coverprofile=coverage.out ./pkg/schemacompat && go tool cover -html=coverage.out
```

More coverage of this code should give us more confidence in the implementation, and it should be fairly easy to cover. We can maintain high coverage using [codecov](https://github.com/codecov/codecov-action), to note coverage deltas in PR comments.

(As always, the goal isn't 100% coverage, it's confidence in the code, whatever coverage %age is reported)

Once coverage is higher we can also consider adding mutation tests, using e.g., https://github.com/zimmski/go-mutesting, which can help us find even more behavior that isn't tested, for example cases where the input to a method doesn't affect test failure.

imjasonh:
After #129 we have shareable coverage reports: https://codecov.io/gh/kcp-dev/kcp/src/main/pkg/schemacompat/schemacompat.go

dgoodwin:
I'll take a look at this one for hack day.

/assign

PR #186, opened 2021-10-11
Add unit tests for schemacompat

ardaguclu:
This PR adds unit tests for schemacompat. After newly added unit tests,
code coverage increases to 68%.

fixes https://github.com/kcp-dev/kcp/issues/128 partially.

### DIFF
--- a/pkg/schemacompat/schemacompat_test.go
+++ b/pkg/schemacompat/schemacompat_test.go
@@ -244,6 +244,7 @@ func TestCompatibility(t *testing.T) {
 			XIntOrString: true,
 			Properties: map[string]apiextensionsv1.JSONSchemaProps{
 				"existing": {Type: "integer"},
+				"new":      {Type: "number"},
 			},
 		},
 		// LCD is the same as existing.
@@ -268,6 +269,7 @@ func TestCompatibility(t *testing.T) {
 			XPreserveUnknownFields: boolPtr(true),
 			Properties: map[string]apiextensionsv1.JSONSchemaProps{
 				"existing": {Type: "integer"},
+				"new":      {Type: "number"},
 			},
 		},
 		// LCD is the same as existing.
--- a/pkg/schemacompat/schemacompat_test.go
+++ b/pkg/schemacompat/schemacompat_test.go
@@ -364,7 +364,6 @@ func TestCompatibility(t *testing.T) {
 				},
 			},
 		},
-		// LCD is the same as existing.
 		wantLCD: &apiextensionsv1.JSONSchemaProps{
 			Type: "object",
 			AdditionalProperties: &apiextensionsv1.JSONSchemaPropsOrBool{
@@ -378,7 +377,7 @@ func TestCompatibility(t *testing.T) {
 			},
 		},
 	}, {
-		desc: "new has more properties, existing contains additional properties boolean",
+		desc: "new has additional properties boolean, existing contains additional properties",
 		existing: &apiextensionsv1.JSONSchemaProps{
 			Type: "object",
 			AdditionalProperties: &apiextensionsv1.JSONSchemaPropsOrBool{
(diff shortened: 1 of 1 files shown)


pkg/schemacompat/schemacompat_test.go
```diff
@@ -186,6 +190,244 @@ func TestCompatibility(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	}, {
+		desc: "new has more properties, existing contains additional properties boolean",
+		existing: &apiextensionsv1.JSONSchemaProps{
+			Type: "object",
+			AdditionalProperties: &apiextensionsv1.JSONSchemaPropsOrBool{
+				Allows: true,
```
sttts, line 383: do we allow that? I don't we do. That would mean polymorphic string maps, i.e. `map[string]interface{}`.

ardaguclu: It is allowed in here https://github.com/kcp-dev/kcp/blob/f1145a6ec820b66420d651c0bce83a3ccf30769a/pkg/schemacompat/schemacompat.go#L372.

While adding unit tests, I assumed that there is no bug. What do you think?

sttts: Note we have a number of restrictions that are not visible in the types, but are done in the API validation phase. The structural schema types are much more restricted in this regard.

sttts:
The `EnsureStructuralSchemaCompatibility` func lacks a call to structural validation:

  https://github.com/kcp-dev/kcp/blob/main/pkg/schemacompat/schemacompat.go#L42

i.e. it should call `ValidateStructural` after `NewStructural`. Only when the former succeeds, it is a type you can expect from CRDs. This also influences the test cases here in the PR.

cc @davidfestal

pkg/schemacompat/schemacompat_test.go
```diff
@@ -186,6 +190,244 @@ func TestCompatibility(t *testing.T) {
...
+			Properties: map[string]apiextensionsv1.JSONSchemaProps{
+				"existing": {Type: "boolean"},
+				"new":      {Type: "number"},
+			},
+		},
+		// LCD is the same as existing.
+		wantLCD: &apiextensionsv1.JSONSchemaProps{
+			Type: "object",
+			Properties: map[string]apiextensionsv1.JSONSchemaProps{
+				"existing": {Type: "boolean"},
```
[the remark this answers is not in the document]
ardaguclu: I agree with @imjasonh. My first attempt was to wrapping up this repetitive tests into one function, but it hurts the readibility.

pkg/schemacompat/schemacompat_test.go
```diff
@@ -186,6 +186,244 @@ func TestCompatibility(t *testing.T) {
...
+			XIntOrString: true,
+			Properties: map[string]apiextensionsv1.JSONSchemaProps{
+				"existing": {Type: "integer"},
+			},
+		},
+		new: &apiextensionsv1.JSONSchemaProps{
+			Type:         "",
+			XIntOrString: true,
+			Properties: map[string]apiextensionsv1.JSONSchemaProps{
+				"existing": {Type: "integer"},
```
davidfestal, line 246: I don't see the new properties

pkg/schemacompat/schemacompat_test.go
```diff
@@ -186,6 +186,244 @@ func TestCompatibility(t *testing.T) {
...
+			Type:                   "",
+			XPreserveUnknownFields: boolPtr(true),
+			Properties: map[string]apiextensionsv1.JSONSchemaProps{
+				"existing": {Type: "integer"},
+			},
+		},
+		new: &apiextensionsv1.JSONSchemaProps{
+			Type:                   "",
+			XPreserveUnknownFields: boolPtr(true),
+			Properties: map[string]apiextensionsv1.JSONSchemaProps{
```
davidfestal, line 269: I don't see the new properties

davidfestal:
> > It was not the purpose of this library to ensure that the schemas are structural schemas. Such a validation is already done in other places of the K8S codebase (in CRD controllers for example afaict).
> 
> That's my point. Just saying that we don't have to test more than we can see in the field with real clusters. Some test-cases here were of that kind.

Maybe we could open a distinct issue for this ?

pkg/schemacompat/schemacompat_test.go
ardaguclu: Thanks @davidfestal, I added new properties.

davidfestal (approved):

PR #188, opened 2021-10-20
Add UID in DeleteOption to avoid races in downstream

ardaguclu:
This PR adds UID into DeleteOptions when deployment is deleted in
downstream to avoid possible races.


ardaguclu:
Most probably syncher will have dramatic changes.

/close

PR #199, opened 2021-11-07
Do not try to sync status when status field is empty or nil in both objects

erwinvaneyk:
Currently `deepEqualStatus` returns true when the status is empty/nil in both objects. This is causing reconciling to happen needlessly for objects with an empty status, and triggers errors in objects that do not have a status (e.g., a ConfigMap).

### DIFF
--- a/pkg/syncer/statussyncer.go
+++ b/pkg/syncer/statussyncer.go
@@ -19,11 +19,10 @@ func deepEqualStatus(oldObj, newObj interface{}) bool {
 		return false
 	}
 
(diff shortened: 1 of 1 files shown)


stevekuznetsov (approved):
@davidfestal how does this look?

imjasonh:
This looks great. I'd love to have a unit test showing a new case where this wasn't handled correctly and now is, but since there aren't tests for this today I won't make you write them (I won't be sad if you did though 😉).

imjasonh (approved):

davidfestal (approved):

PR #201, opened 2021-11-08
Add unit test for pkg/syncer/statussyncer.go

imjasonh:
This behavior was changed in https://github.com/kcp-dev/kcp/pull/199 without a corresponding unit test, because there were no unit tests.

Pre-#199, the fourth unit test case would have returned `false`, and now returns `true`.


sttts:
lgtm

PR #202, opened 2021-11-08
Schedule namespaces and all resources in them to clusters

imjasonh:
This does not yet include:

- syncer transforming namespace names when applying to physical clusters (probably in another more syncer-focused PR)

### DIFF
--- a/pkg/gvk/gvk.go
+++ b/pkg/gvk/gvk.go
@@ -1,0 +1,50 @@
+/*
+Copyright 2021 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package gvk
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/discovery/cached/memory"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/restmapper"
+)
+
+// GVKTranslator translates a given GroupVersionKind into a
+// GroupVersionResource, using a RESTMapper.
+type GVKTranslator struct {
+	m *restmapper.DeferredDiscoveryRESTMapper
+}
+
+// NewGVKTranslator returns a GVKTranslator that can translate a given
+// GroupVersionKind into the corresponding GroupVersionResource, using a
+// RESTMapper.
+func NewGVKTranslator(cfg *rest.Config) *GVKTranslator {
+	return &GVKTranslator{m: restmapper.NewDeferredDiscoveryRESTMapper(
+		memory.NewMemCacheClient(
+			discovery.NewDiscoveryClientForConfigOrDie(cfg)))}
+}
+
+// Get returns the corresponding GroupVersionResource for the given
+// GroupVersionKind.
+func (t *GVKTranslator) Get(gvk *schema.GroupVersionKind) (*schema.GroupVersionResource, error) {
+	m, err := t.m.RESTMapping(gvk.GroupKind(), gvk.Version)
+	if err != nil {
+		return nil, err
+	}
+	return &m.Resource, nil
+}
--- a/pkg/informer/informer.go
+++ b/pkg/informer/informer.go
@@ -1,0 +1,217 @@
+/*
+Copyright 2021 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package informer
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	"go.uber.org/multierr"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+)
+
+const (
+	resyncPeriod   = 10 * time.Hour
+	pollTypesEvery = time.Minute
+)
+
+// DynamicDiscoverySharedInformerFactory is a SharedInformerFactory that
+// dynamically discovers new types and begins informing on them.
+type DynamicDiscoverySharedInformerFactory struct {
+	disco      *discovery.DiscoveryClient
+	dsif       dynamicinformer.DynamicSharedInformerFactory
+	handler    GVREventHandler
+	filterFunc func(interface{}) bool
+
+	mu   sync.Mutex // guards gvrs
+	gvrs sets.String
+}
+
+// IndexerFor returns the indexer for the given type GVR.
+func (d *DynamicDiscoverySharedInformerFactory) IndexerFor(gvr schema.GroupVersionResource) cache.Indexer {
+	return d.dsif.ForResource(gvr).Informer().GetIndexer()
+}
+
+// Listers returns a map of per-resource-type listers for all types that are
+// known by this informer factory, and that are synced.
+//
+// If any informers aren't synced, their GVRs are returned so that they can be
+// checked and processed later.
+func (d *DynamicDiscoverySharedInformerFactory) Listers() (listers map[schema.GroupVersionResource]cache.GenericLister, notSynced []schema.GroupVersionResource) {
+	listers = map[schema.GroupVersionResource]cache.GenericLister{}
+
+	for _, gvrstr := range d.gvrs.UnsortedList() {
+		gvr, _ := schema.ParseResourceArg(gvrstr)
+		if gvr == nil {
+			klog.Errorf("parsing GVR string %q", gvrstr)
+			continue
+		}
+
+		if !d.dsif.ForResource(*gvr).Informer().HasSynced() {
+			notSynced = append(notSynced, *gvr)
+			continue
+		}
+
+		listers[*gvr] = d.dsif.ForResource(*gvr).Lister()
+	}
+	return listers, notSynced
+}
+
+// NewDynamicDiscoverySharedInformerFactory returns a factory for shared
+// informers that discovers new types and informs on updates to resources of
+// those types.
+func NewDynamicDiscoverySharedInformerFactory(
+	disco *discovery.DiscoveryClient,
+	dynClient dynamic.Interface,
+	filterFunc func(obj interface{}) bool,
+	handler GVREventHandler) DynamicDiscoverySharedInformerFactory {
+	dsif := dynamicinformer.NewDynamicSharedInformerFactory(dynClient, resyncPeriod)
+	return DynamicDiscoverySharedInformerFactory{
+		disco:      disco,
+		dsif:       dsif,
+		handler:    handler,
+		filterFunc: filterFunc,
+		gvrs:       sets.NewString(),
+	}
+}
+
+// GVREventHandler is an event handler that includes the GroupVersionResource
+// of the resource being handled.
+type GVREventHandler interface {
+	OnAdd(gvr schema.GroupVersionResource, obj interface{})
+	OnUpdate(gvr schema.GroupVersionResource, oldObj, newObj interface{})
+	OnDelete(gvr schema.GroupVersionResource, obj interface{})
+}
+
+type GVREventHandlerFuncs struct {
+	AddFunc    func(gvr schema.GroupVersionResource, obj interface{})
+	UpdateFunc func(gvr schema.GroupVersionResource, oldObj, newObj interface{})
+	DeleteFunc func(gvr schema.GroupVersionResource, obj interface{})
+}
+
+func (g GVREventHandlerFuncs) OnAdd(gvr schema.GroupVersionResource, obj interface{}) {
+	if g.AddFunc != nil {
+		g.AddFunc(gvr, obj)
+	}
+}
+func (g GVREventHandlerFuncs) OnUpdate(gvr schema.GroupVersionResource, oldObj, newObj interface{}) {
+	if g.UpdateFunc != nil {
+		g.UpdateFunc(gvr, oldObj, newObj)
+	}
+}
+func (g GVREventHandlerFuncs) OnDelete(gvr schema.GroupVersionResource, obj interface{}) {
+	if g.DeleteFunc != nil {
+		g.DeleteFunc(gvr, obj)
+	}
+}
+
+func (d *DynamicDiscoverySharedInformerFactory) Start(ctx context.Context) {
+	// Immediately discover types and start informing.
+	// TODO: Feed any failure to discover types into /readyz, instead of
+	// panicking.
+	if err := d.discoverTypes(ctx); err != nil {
+		klog.Fatalf("Error discovering initial types: %v", err)
+	}
+
+	// Poll for new types in the background.
+	ticker := time.NewTicker(pollTypesEvery)
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				ticker.Stop()
+				return
+			case <-ticker.C:
+				if err := d.discoverTypes(ctx); err != nil {
+					klog.Errorf("Error discovering types: %v", err)
+				}
+			}
+		}
+	}()
+}
+
+func (d *DynamicDiscoverySharedInformerFactory) discoverTypes(ctx context.Context) error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	latest := sets.NewString()
+	rs, err := d.disco.ServerPreferredResources()
+	if err != nil {
+		return err
+	}
+	for _, r := range rs {
+		gv, err := schema.ParseGroupVersion(r.GroupVersion)
+		if err != nil {
+			return err
+		}
+		for _, ai := range r.APIResources {
+			if strings.Contains(ai.Name, "/") {
+				// foo/status, pods/exec, namespace/finalize, etc.
+				continue
+			}
+			if !ai.Namespaced {
+				// Ignore cluster-scoped things.
+				continue
+			}
+			if !sets.NewString([]string(ai.Verbs)...).HasAll("list", "watch") {
+				klog.V(2).Infof("resource %s.%s %s is not list+watchable: %v", gv.Group, gv.Version, ai.Name, ai.Verbs)
+				continue
+			}
+
+			latest.Insert(strings.Join([]string{ai.Name, gv.Version, gv.Group}, "."))
+		}
+	}
+
+	// Set up informers for any new types that have been discovered.
+	// TODO: Stop informers for types we no longer have.
+	newGVRs := latest.Difference(d.gvrs)
+	var merr error
+	for _, gvrstr := range newGVRs.UnsortedList() {
+		gvr, _ := schema.ParseResourceArg(gvrstr)
+		if gvr == nil {
+			multierr.AppendInto(&merr, fmt.Errorf("parsing GVR string %q: %w", gvrstr, err))
+			continue
+		}
+
+		inf := d.dsif.ForResource(*gvr).Informer()
+		inf.AddEventHandler(cache.FilteringResourceEventHandler{
+			FilterFunc: d.filterFunc,
+			Handler: cache.ResourceEventHandlerFuncs{
+				AddFunc:    func(obj interface{}) { d.handler.OnAdd(*gvr, obj) },
+				UpdateFunc: func(oldObj, newObj interface{}) { d.handler.OnUpdate(*gvr, oldObj, newObj) },
+				DeleteFunc: func(obj interface{}) { d.handler.OnDelete(*gvr, obj) },
+			},
+		})
+		go inf.Run(ctx.Done())
+	}
+	if merr != nil {
+		return merr
+	}
+
+	d.gvrs = latest
+	return nil
+}
--- a/pkg/reconciler/namespace/controller.go
+++ b/pkg/reconciler/namespace/controller.go
@@ -1,0 +1,415 @@
+/*
+Copyright 2021 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package namespace
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
+	coreinformers "k8s.io/client-go/informers/core/v1"
+	coreclient "k8s.io/client-go/kubernetes/typed/core/v1"
+	corelisters "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clusters"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+
+	clusterinformer "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/cluster/v1alpha1"
+	clusterlisters "github.com/kcp-dev/kcp/pkg/client/listers/cluster/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/gvk"
+	"github.com/kcp-dev/kcp/pkg/informer"
+)
+
+const controllerName = "namespace-scheduler"
+
+// NewController returns a new Controller which schedules namespaced resources to a Cluster.
+func NewController(
+	dynClient dynamic.Interface,
+	disco *discovery.DiscoveryClient,
+	clusterInformer clusterinformer.ClusterInformer,
+	clusterLister clusterlisters.ClusterLister,
+	namespaceInformer coreinformers.NamespaceInformer,
+	namespaceLister corelisters.NamespaceLister,
+	namespaceClient coreclient.NamespaceInterface,
+	gvkTrans *gvk.GVKTranslator,
+) *Controller {
+
+	resourceQueue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+	gvrQueue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+	namespaceQueue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+	clusterQueue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+
+	c := &Controller{
+		resourceQueue:  resourceQueue,
+		gvrQueue:       gvrQueue,
+		namespaceQueue: namespaceQueue,
+		clusterQueue:   clusterQueue,
+
+		dynClient:       dynClient,
+		clusterLister:   clusterLister,
+		namespaceLister: namespaceLister,
+		namespaceClient: namespaceClient,
+		gvkTrans:        gvkTrans,
+
+		syncChecks: []cache.InformerSynced{
+			namespaceInformer.Informer().HasSynced,
+			clusterInformer.Informer().HasSynced,
+		},
+	}
+	clusterInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) { c.enqueueCluster(obj) },
+		UpdateFunc: func(_, obj interface{}) { c.enqueueCluster(obj) },
+		DeleteFunc: func(obj interface{}) { c.enqueueCluster(obj) },
+	})
+	namespaceInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: filterNamespace,
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc:    func(obj interface{}) { c.enqueueNamespace(obj) },
+			UpdateFunc: func(_, obj interface{}) { c.enqueueNamespace(obj) },
+			DeleteFunc: nil, // Nothing to do.
+		},
+	})
+	c.ddsif = informer.NewDynamicDiscoverySharedInformerFactory(disco, dynClient,
+		filterResource,
+		informer.GVREventHandlerFuncs{
+			AddFunc:    func(gvr schema.GroupVersionResource, obj interface{}) { c.enqueueResource(gvr, obj) },
+			UpdateFunc: func(gvr schema.GroupVersionResource, _, obj interface{}) { c.enqueueResource(gvr, obj) },
+			DeleteFunc: nil, // Nothing to do.
+		})
+
+	return c
+}
+
+type Controller struct {
+	resourceQueue  workqueue.RateLimitingInterface
+	gvrQueue       workqueue.RateLimitingInterface
+	namespaceQueue workqueue.RateLimitingInterface
+	clusterQueue   workqueue.RateLimitingInterface
+
+	dynClient       dynamic.Interface
+	clusterLister   clusterlisters.ClusterLister
+	namespaceLister corelisters.NamespaceLister
+	namespaceClient coreclient.NamespaceInterface
+	ddsif           informer.DynamicDiscoverySharedInformerFactory
+	gvkTrans        *gvk.GVKTranslator
+
+	syncChecks []cache.InformerSynced
+}
+
+func filterResource(obj interface{}) bool {
+	key, err := cache.MetaNamespaceKeyFunc(obj)
+	if err != nil {
+		runtime.HandleError(err)
+		return false
+	}
+	_, clusterAwareName, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		runtime.HandleError(err)
+		return false
+	}
+	clusterName, _ := clusters.SplitClusterAwareKey(clusterAwareName)
+	if clusterName != "admin" {
+		klog.V(2).Infof("Skipping update for non-admin cluster %q", clusterName)
+		return false
+	}
+
+	current, ok := obj.(*unstructured.Unstructured)
+	if !ok {
+		klog.V(2).Infof("Object was not Unstructured: %T", obj)
+		return false
+	}
+
+	if namespaceBlocklist.Has(current.GetNamespace()) {
+		klog.V(2).Infof("Skipping syncing namespace %q", current.GetNamespace())
+		return false
+	}
+	return true
+}
+
+func filterNamespace(obj interface{}) bool {
+	key, err := cache.MetaNamespaceKeyFunc(obj)
+	if err != nil {
+		runtime.HandleError(err)
+		return false
+	}
+	_, clusterAwareName, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		runtime.HandleError(err)
+		return false
+	}
+	clusterName, name := clusters.SplitClusterAwareKey(clusterAwareName)
+	if clusterName != "admin" {
+		klog.Infof("Skipping update for non-admin cluster %q", clusterName)
+		return false
+	}
+	if namespaceBlocklist.Has(name) {
+		klog.V(2).Infof("Skipping syncing namespace %q", name)
+		return false
+	}
+	return true
+}
+
+func (c *Controller) enqueueResource(gvr schema.GroupVersionResource, obj interface{}) {
+	key, err := cache.MetaNamespaceKeyFunc(obj)
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+	gvrstr := strings.Join([]string{gvr.Resource, gvr.Version, gvr.Group}, ".")
+	c.resourceQueue.Add(gvrstr + "::" + key)
+}
+
+func (c *Controller) enqueueGVR(gvr schema.GroupVersionResource) {
+	gvrstr := strings.Join([]string{gvr.Resource, gvr.Version, gvr.Group}, ".")
+	c.gvrQueue.Add(gvrstr)
+}
+
+func (c *Controller) enqueueNamespace(obj interface{}) {
+	key, err := cache.MetaNamespaceKeyFunc(obj)
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+	c.namespaceQueue.Add(key)
+}
+
+func (c *Controller) enqueueCluster(obj interface{}) {
+	key, err := cache.MetaNamespaceKeyFunc(obj)
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+	c.clusterQueue.Add(key)
+}
+
+func (c *Controller) Start(ctx context.Context, numThreads int) {
+	defer runtime.HandleCrash()
+	defer c.resourceQueue.ShutDown()
+	defer c.gvrQueue.ShutDown()
+	defer c.namespaceQueue.ShutDown()
+	defer c.clusterQueue.ShutDown()
+
+	klog.Info("Starting Namespace scheduler")
+	defer klog.Info("Shutting down Namespace scheduler")
+
+	if !cache.WaitForNamedCacheSync("namespace-scheduler", ctx.Done(), c.syncChecks...) {
+		klog.Warning("Failed to wait for caches to sync")
+		return
+	}
+
+	c.ddsif.Start(ctx)
+
+	for i := 0; i < numThreads; i++ {
+		go wait.Until(func() { c.startResourceWorker(ctx) }, time.Second, ctx.Done())
+		go wait.Until(func() { c.startGVRWorker(ctx) }, time.Second, ctx.Done())
+		go wait.Until(func() { c.startNamespaceWorker(ctx) }, time.Second, ctx.Done())
+		go wait.Until(func() { c.startClusterWorker(ctx) }, time.Second, ctx.Done())
+	}
+	<-ctx.Done()
+}
+
+func (c *Controller) startResourceWorker(ctx context.Context) {
+	for c.processNextResource(ctx) {
+	}
+}
+func (c *Controller) startGVRWorker(ctx context.Context) {
+	for c.processNextGVR(ctx) {
+	}
+}
+func (c *Controller) startNamespaceWorker(ctx context.Context) {
+	for c.processNextNamespace(ctx) {
+	}
+}
+func (c *Controller) startClusterWorker(ctx context.Context) {
+	for c.processNextCluster(ctx) {
+	}
+}
+
+func (c *Controller) processNextResource(ctx context.Context) bool {
+	// Wait until there is a new item in the working queue
+	k, quit := c.resourceQueue.Get()
+	if quit {
+		return false
+	}
+	key := k.(string)
+
+	// No matter what, tell the queue we're done with this key, to unblock
+	// other workers.
+	defer c.resourceQueue.Done(key)
+
+	if err := c.processResource(ctx, key); err != nil {
+		runtime.HandleError(fmt.Errorf("%q controller failed to sync %q, err: %w", controllerName, key, err))
+		c.resourceQueue.AddRateLimited(key)
+		return true
+	}
+	c.resourceQueue.Forget(key)
+	return true
+}
+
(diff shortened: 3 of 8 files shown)


cmd/namespace-scheduler/README.md
```diff
@@ -0,0 +1,38 @@
+# Namespace Scheulder
+
+The Namespace Scheduler is responsible for watching `kcp` for namespaced resources and scheduling them to a physical cluster location.
+
+The Namespace Scheduler will not split resources like the Deployment Splitter, only schedule all resources in the same namespace to one cluster, and move it to another cluster if the cluster becomes unavailable.
+
+The underlying real clusters will react to the these resources by syncing them, updating status, and syncing that status back to `kcp`.
+
+## Running
```
stevekuznetsov, line 9: I remember starting out was hard since there were no real common well-worn runners. Can we put this into a script or set of scripts that we all use when we're hacking?

stevekuznetsov: (Or a set of scripts that set the env up to allow an end-to-end test)

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,259 @@
...
+
+const resyncPeriod = 10 * time.Hour
+
+// NewController returns a new Controller which schedules namespaced resources to a Cluster.
+func NewController(cfg *rest.Config) *Controller {
+	dynClient := dynamic.NewForConfigOrDie(cfg)
+	kubeClient := kubernetes.NewForConfigOrDie(cfg)
+	clusterClient := clusterclient.NewForConfigOrDie(cfg)
+	queue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+	stopCh := make(chan struct{}) // TODO: hook this up to SIGTERM/SIGINT
```
stevekuznetsov, line 42: @ncdc will not be happy with this one - are we missing some normal controller flow, why do we have this in all of our constructors? (I've only ever written with controller-runtime)

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,259 @@
...
+
+	// Give up and report error elsewhere.
+	c.queue.Forget(key)
+	runtime.HandleError(err)
+	klog.Infof("Dropping key %q after failed retries: %v", key, err)
+}
+
+// key is gvr::namespace/name
+func (c *Controller) process(key string) error {
+	parts := strings.SplitN(key, "::", 2)
```
stevekuznetsov, line 187: This is not multi-cluster/workspace aware. Should it be?

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	if !exists {
+		klog.Infof("Object with key %q was deleted", key)
+		return nil
+	}
+	current := obj.(*unstructured.Unstructured)
+	previous := current.DeepCopy()
+
+	ctx := context.TODO()
```
stevekuznetsov, line 210: Don't we have one of these we can pass in to the constructor?

stevekuznetsov (commented):
Might need to pull this down to read it, the Codecov spam is really annoying

pkg/reconciler/namespace/controller.go
imjasonh: Yeah I think I'm going to wait for #203 and follow that pattern, which is much better. I think that will get merged before this one anyway.

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,256 @@
...
+	queue workqueue.RateLimitingInterface
+
+	dynClient  dynamic.Interface
+	kubeClient kubernetes.Interface
+
+	clusterLister   clusterlisters.ClusterLister
+	namespaceLister corelister.NamespaceLister
+	namespaceClient coreclient.NamespaceInterface
+	stopCh          chan struct{}
+	indexer         cache.Indexer
```
ncdc, line 93: Does this ever get initialized?

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,256 @@
...
+	// If the object being reconciled changed as a result, update it.
+	if !equality.Semantic.DeepEqual(previous, current) {
+		_, uerr := c.dynClient.Resource(*gvr).Namespace(current.GetNamespace()).Update(ctx, current, metav1.UpdateOptions{})
+		return uerr
+	}
+	return nil
+}
+
+func (c *Controller) processNamespace(ctx context.Context, key string) error {
+	obj, exists, err := c.indexer.GetByKey(key)
```
ncdc, line 223: Can you use the namespaceLister here?

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,256 @@
...
+	current := obj.(*corev1.Namespace)
+	previous := current.DeepCopy()
+
+	if err := c.reconcileNamespace(ctx, current); err != nil {
+		return err
+	}
+
+	// If the namespace changed as a result, update it.
+	if !equality.Semantic.DeepEqual(previous, current) {
+		_, uerr := c.kubeClient.CoreV1().Namespaces().Update(ctx, current, metav1.UpdateOptions{})
```
ncdc, line 239: I'd recommend Patch over Update to avoid accidental field pruning (this client vendors the go struct for Namespace @ v1.22.x, server upgrades to v1.23+ which adds a new field to Namespace -> this client doing an update will result in that newer field being dropped).

ncdc: Would you agree @smarterclayton ?

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,256 @@
...
+		return uerr
+	}
+	return nil
+}
+
+func (c *Controller) processCluster(ctx context.Context, key string) error {
+	obj, exists, err := c.indexer.GetByKey(key)
```
ncdc, line 246: clusterLister?

pkg/reconciler/namespace/namespace.go
```diff
@@ -0,0 +1,175 @@
...
+
+	if lbls[clusterLabel] == ns.Labels[clusterLabel] {
+		// Already assigned to the right cluster.
+		return nil
+	}
+
+	// Update the resource's assignment. This will be applied when
+	// reconcile is done.
+	klog.Infof("updating %s %s/%s cluster %q -> %q", gvr.String(), unstr.GetNamespace(), unstr.GetName(), lbls[clusterLabel], ns.Labels[clusterLabel])
+	lbls[clusterLabel] = ns.Labels[clusterLabel]
```
ncdc, line 57: What if the namespace doesn't have the clusterLabel set?

pkg/reconciler/namespace/namespace.go
```diff
@@ -0,0 +1,175 @@
...
+		if len(cls) == 0 {
+			// TODO set status, on namespace and all resources in namespace.
+			return fmt.Errorf("no clusters")
+		}
+		// Select a cluster at random.
+		cl := cls[rand.Intn(len(cls))]
+		ns.Labels[clusterLabel] = cl.Name
+
+		// Update the namespace.
+		if _, err = c.namespaceClient.Update(ctx, ns, metav1.UpdateOptions{}); err != nil {
```
ncdc, line 91: Update -> Patch?

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,256 @@
...
+
+	dynClient  dynamic.Interface
+	kubeClient kubernetes.Interface
+
+	clusterLister   clusterlisters.ClusterLister
+	namespaceLister corelister.NamespaceLister
+	namespaceClient coreclient.NamespaceInterface
+	stopCh          chan struct{}
+	indexer         cache.Indexer
+	lister          cache.GenericLister
```
ncdc, line 94: Does this ever get initialized?

pkg/reconciler/namespace/namespace.go
```diff
@@ -0,0 +1,175 @@
...
+		sch := unstructured.UnstructuredJSONScheme
+		var buf bytes.Buffer
+		if err := sch.Encode(obj, &buf); err != nil {
+			return err
+		}
+		var unstr unstructured.Unstructured
+		_, gvk, err := sch.Decode(buf.Bytes(), nil, &unstr)
+		if err != nil {
+			return err
+		}
```
ncdc, line 113: Not sure if this is any better per se, but here's what we have done in the past in cluster-api:
```go
rawMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
if err != nil {
	return err
}
unstr := unstructured.Unstructured{Object: rawMap}
```

@smarterclayton any suggestions on which is better?

pkg/reconciler/namespace/namespace.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		return err
+	}
+	for _, ns := range nss {
+		if ns.Labels == nil {
+			// this shouldn't happen.
+			continue
+		}
+		delete(ns.Labels, clusterLabel)
+		if _, err := c.namespaceClient.Update(ctx, ns, metav1.UpdateOptions{}); err != nil {
```
ncdc, line 170: Update->Patch?

pkg/reconciler/namespace/namespace.go
imjasonh: I hadn't seen this before, thanks for sharing it.

I think I still need to use the schema to get the gvk out of it, so I can translate it to a gvr for the dynamic client. I'd absolutely believe there's a better way to do all this than what I've done though.

pkg/reconciler/namespace/namespace.go
imjasonh: Then the resources will also have their clusterLabel unset, if they weren't already. I believe this is desired behavior, since it means a resource belonging to an unassigned namespace can be ensured to also be unassigned, in case something outside of this code set the label for some reason.

pkg/reconciler/namespace/controller.go
imjasonh: Good call, and done in the latest revision. It turned out to be quite a bit simpler too since the patch bytes to describe setting the label is the same whether it's a resource or a namespace.

I'm now actually doing the Patch inside `reconcileResource`, instead of doing it implicitly outside, since updating the label is the only expected modification of the object, and this now matches `reconcileNamespace` and `reconcileCluster` semantics.

The only downside is that testing this is a bit painful since I have to get a fake clientset and a fake dynamic client to agree about what fake objects I've put into both of them, and I can't seem to figure out how to do that. I'll keep trying though.

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	key = parts[1]
+
+	obj, exists, err := c.indexer.GetByKey(key)
+	if err != nil {
+		return err
+	}
+
+	if !exists {
+		klog.Infof("Object with key %q was deleted", key)
```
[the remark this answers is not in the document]
imjasonh: Done,

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,259 @@
...
+// key is gvr::namespace/name
+func (c *Controller) process(key string) error {
+	parts := strings.SplitN(key, "::", 2)
+	if len(parts) != 2 {
+		return fmt.Errorf("invalid key: %s", key)
+	}
+	gvrstr := parts[0]
+	gvr, _ := schema.ParseResourceArg(gvrstr)
+	if gvr == nil {
+		return fmt.Errorf("Error parsing GVR %q", gvrstr)
```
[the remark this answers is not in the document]
imjasonh: Done.

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,259 @@
...
+	c.queue.Forget(key)
+	runtime.HandleError(err)
+	klog.Infof("Dropping key %q after failed retries: %v", key, err)
+}
+
+// key is gvr::namespace/name
+func (c *Controller) process(key string) error {
+	parts := strings.SplitN(key, "::", 2)
+	if len(parts) != 2 {
+		return fmt.Errorf("invalid key: %s", key)
```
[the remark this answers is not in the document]
imjasonh: Done.

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	// Give up and report error elsewhere.
+	c.queue.Forget(key)
+	runtime.HandleError(err)
+	klog.Infof("Dropping key %q after failed retries: %v", key, err)
+}
+
+// key is gvr::namespace/name
+func (c *Controller) process(key string) error {
```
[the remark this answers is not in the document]
imjasonh: Made it `processResource` to match `reconcileResource`, but I'm open to renaming both if you feel strongly.

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func (c *Controller) handleErr(err error, key string) {
+	// Reconcile worked, nothing else to do for this workqueue item.
+	if err == nil {
+		c.queue.Forget(key)
+		return
+	}
+
+	// Re-enqueue up to 5 times.
```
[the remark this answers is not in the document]
imjasonh: I think I got this bit from Knative, which Tekton also reuses. My main reason for wanting to retry is if a lister cache doesn't have a namespace or cluster yet, we shouldn't fail after one try, but we also should also give up eventually.

If you have a better pattern for this I'm open to ideas, we should do the same in cluster controller and deployment splitter too.

pkg/informer/informer.go
```diff
@@ -0,0 +1,157 @@
...
+			gvr := schema.GroupVersionResource{
+				Group:    gv.Group,
+				Version:  gv.Version,
+				Resource: ai.Name,
+			}
+			latest.Insert(gvr.String())
+		}
+	}
+
+	// Set up informers for any new types that have been discovered.
```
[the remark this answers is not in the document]
imjasonh: Added a TODO in the latest revision.

pkg/informer/informer.go
```diff
@@ -0,0 +1,157 @@
...
+			if strings.Contains(ai.Name, "/") {
+				// foo/status, pods/exec, namespace/finalize, etc.
+				continue
+			}
+			if !ai.Namespaced {
+				// Ignore cluster-scoped things.
+				continue
+			}
+			if !sets.NewString([]string(ai.Verbs)...).Has("watch") {
+				klog.Infof("resource %s %s is not watchable: %v", vr, ai.Name, ai.Verbs)
```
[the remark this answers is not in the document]
imjasonh: Done.

pkg/informer/informer.go
```diff
@@ -0,0 +1,157 @@
...
+	rs, err := d.disco.ServerPreferredResources()
+	if err != nil {
+		return err
+	}
+	for _, r := range rs {
+		gv, err := schema.ParseGroupVersion(r.GroupVersion)
+		if err != nil {
+			return err
+		}
+		vr := gv.Version + "." + gv.Group
```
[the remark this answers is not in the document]
imjasonh: Done, and since it was only used to log, I moved the whole thing into the logging format.

pkg/reconciler/namespace/controller.go
imjasonh: Done.

pkg/reconciler/namespace/controller.go
imjasonh: This is done now.

pkg/reconciler/namespace/controller.go
imjasonh: IOU'd.

pkg/informer/informer.go
```diff
@@ -0,0 +1,157 @@
...
+			latest.Insert(gvr.String())
+		}
+	}
+
+	// Set up informers for any new types that have been discovered.
+	newGVRs := latest.Difference(d.gvrs)
+	for _, gvrstr := range newGVRs.UnsortedList() {
+		gvr, _ := schema.ParseResourceArg(gvrstr)
+		if gvr == nil {
+			return fmt.Errorf("parsing GVR string %q: %w", gvrstr, err)
```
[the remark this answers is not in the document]
imjasonh: Done, using `multierr`.

pkg/reconciler/namespace/controller.go
imjasonh: It wasn't. I ended up replacing the `lister` and `indexer` both with somewhat terrible hacks to expose the per-type listers and indexers from the dynamic-discovery SIF. I'll play with this some more today and see what breaks.

pkg/reconciler/namespace/controller.go
imjasonh: Good call, done.

pkg/reconciler/namespace/controller.go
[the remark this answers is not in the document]
stevekuznetsov: Won't you re-try on re-list?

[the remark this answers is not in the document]
davidfestal: > If you have a better pattern for this I'm open to ideas, we should do the same in cluster controller and deployment splitter too.

+1 I've never been very comfortable with the idea of stopping after a fix number of retries. IIRC we already had to bypass this behavior with a specific `RetryableError` type in some places in the cluster controller to avoid loosing an event after a number of retries.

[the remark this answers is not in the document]
imjasonh: I'm sold. I'll use RetryableError and not give up after 5 retries. I've also cleaned up RetryableError usage in the other packages.

pkg/informer/informer.go
```diff
@@ -0,0 +1,179 @@
...
+func (g GVREventHandlerFuncs) OnDelete(gvr schema.GroupVersionResource, obj interface{}) {
+	if g.DeleteFunc != nil {
+		g.DeleteFunc(gvr, obj)
+	}
+}
+
+func (d *DynamicDiscoverySharedInformerFactory) Start(ctx context.Context) {
+	// Immediately discover types and start informing.
+	if err := d.discoverTypes(ctx); err != nil {
+		klog.Errorf("Error discovering initial types: %v", err)
```
ncdc, line 101: Should we return early here, if discovery fails the very first time?

pkg/informer/informer.go
```diff
@@ -0,0 +1,179 @@
...
+		for _, ai := range r.APIResources {
+			if strings.Contains(ai.Name, "/") {
+				// foo/status, pods/exec, namespace/finalize, etc.
+				continue
+			}
+			if !ai.Namespaced {
+				// Ignore cluster-scoped things.
+				continue
+			}
+			if !sets.NewString([]string(ai.Verbs)...).Has("watch") {
```
ncdc, line 141: Check for list too?

cmd/namespace-scheduler/main.go
```diff
@@ -0,0 +1,38 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"os"
+	"os/signal"
+
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog"
```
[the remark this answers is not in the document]
imjasonh: Thanks, I fixed this through this PR after presubmit complained.

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,268 @@
...
+)
+
+const resyncPeriod = 10 * time.Hour
+
+// NewController returns a new Controller which schedules namespaced resources to a Cluster.
+func NewController(ctx context.Context, cfg *rest.Config) *Controller {
+	disco := discovery.NewDiscoveryClientForConfigOrDie(cfg)
+	dynClient := dynamic.NewForConfigOrDie(cfg)
+	kubeClient := kubernetes.NewForConfigOrDie(cfg)
+	clusterClient := clusterclient.NewForConfigOrDie(cfg)
```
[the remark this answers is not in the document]
imjasonh: Yeah, I'll add a TODO and take care of this before un-WIPing.

cmd/namespace-scheduler/main.go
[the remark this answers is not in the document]
imjasonh: (whoops, actually had missed this one, thanks for catching it)

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,268 @@
...
+	// resources across all clusters. This means making all the workqueue
+	// keys below cluster-aware, and all Patches performed during
+	// reconciliation cluster-aware.
+
+	ddsif := informer.NewDynamicDiscoverySharedInformerFactory(disco, dynClient, informer.GVREventHandlerFuncs{
+		AddFunc:    func(gvr schema.GroupVersionResource, obj interface{}) { c.enqueueResource(gvr, obj) },
+		UpdateFunc: func(gvr schema.GroupVersionResource, _, obj interface{}) { c.enqueueResource(gvr, obj) },
+		DeleteFunc: nil, // Nothing to do.
+	})
+	ddsif.WaitForCacheSync(ctx)
```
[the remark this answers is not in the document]
imjasonh: Thinking about this some more, `processResource` calls `c.ddsif.IndexerFor(*gvr).GetByKey(key)` and drops the key if `!exists`, which means we'll want that cache to be up-to-date before hitting that, or retry on `!exists` and expect not to get notified of deleted things. The former sounds more reliable than the latter.

pkg/reconciler/namespace/namespace.go
```diff
@@ -0,0 +1,194 @@
...
+
+		// Patch the namespace.
+		klog.Infof("Patching to update cluster assignment for namespace %s: %q -> %q", ns.Name, old, new)
+		if _, err := c.namespaceClient.Patch(ctx, ns.Name, types.MergePatchType, clusterLabelPatchBytes(new), metav1.PatchOptions{}); err != nil {
+			return kerrors.NewRetryableError(err)
+		}
+	}
+
+	// Update all resources in the namespace with the cluster assignment.
+	for _, lister := range c.ddsif.Listers() {
```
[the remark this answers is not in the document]
imjasonh: Good call, TODO'd.

pkg/informer/informer.go
imjasonh: I've changed it to a `Fatal`, but I don't love it. We could retry in case it's flaky, but that's effectively what logging and starting the ticker did before.

pkg/informer/informer.go
imjasonh: Done.

pkg/reconciler/namespace/namespace_test.go
```diff
@@ -0,0 +1,95 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "my-namespace",
+				Labels: map[string]string{
+					clusterLabel: "my-cluster",
+				},
+			},
+		},
+	)
+	nsif := informers.NewSharedInformerFactory(clientset, time.Second)
+	nsif.Core().V1().Namespaces().Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{}) // This is needed to get the informer to start listening.
```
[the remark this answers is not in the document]
imjasonh: Nice, done.

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	clusterName, name := clusters.SplitClusterAwareKey(clusterAwareName)
+	if clusterName != "admin" {
+		klog.Infof("Skipping update for non-admin cluster %q", clusterName)
+		return nil
+	}
+	if namespaceBlocklist.Has(name) {
+		klog.V(2).Infof("Skipping syncing namespace %q", name)
+		return nil
+	}
```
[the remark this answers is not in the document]
imjasonh: Good idea. I don't know if it practically affects performance much, but we might as well filter as early as we can, rather than knowingly putting no-ops in the queue.

pkg/gvk/gvk.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+type GVKTranslator struct {
+	m *restmapper.DeferredDiscoveryRESTMapper
+}
+
+func NewGVKTranslator(cfg *rest.Config) *GVKTranslator {
+	return &GVKTranslator{m: restmapper.NewDeferredDiscoveryRESTMapper(
+		memory.NewMemCacheClient(
+			discovery.NewDiscoveryClientForConfigOrDie(cfg)))}
```
stevekuznetsov, line 18: Doesn't this assume a logical cluster?

imjasonh: Yeah, it does for now. This PR only works against a single lcluster for now.

pkg/reconciler/cluster/controller.go
```diff
@@ -25,15 +26,15 @@ import (
...
+	"github.com/kcp-dev/kcp"
 	apiresourcev1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apiresource/v1alpha1"
 	clusterv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/cluster/v1alpha1"
 	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
 	apiresourceinformer "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/apiresource/v1alpha1"
 	clusterinformer "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/cluster/v1alpha1"
 	"github.com/kcp-dev/kcp/pkg/reconciler/apiresource"
 	"github.com/kcp-dev/kcp/pkg/syncer"
-	"github.com/kcp-dev/kcp/pkg/util/errors"
+	kerrors "github.com/kcp-dev/kcp/pkg/util/errors"
```
stevekuznetsov, line 37: I think this package got :bomb:

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,303 @@
...
+	key := k.(string)
+
+	// No matter what, tell the queue we're done with this key, to unblock
+	// other workers.
+	defer c.queue.Done(key)
+
+	var err error
+	if strings.HasPrefix(key, "namespace::") {
+		err = c.processNamespace(ctx, strings.TrimPrefix(key, "namespace::"))
```
stevekuznetsov, line 213: Speaking for my interpretation of their comments on the Workspace PR, I think Andy and Stefan would have expected a queue to be for one type and one type only

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,303 @@
...
+}
+
+func (c *Controller) handleErr(err error, key string) {
+	// Reconcile worked, nothing else to do for this workqueue item.
+	if err == nil {
+		c.queue.Forget(key)
+		return
+	}
+
+	if _, ok := err.(*kerrors.RetryableError); ok {
```
stevekuznetsov: We deleted this code

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,303 @@
...
+	unstr, ok := obj.(*unstructured.Unstructured)
+	if !ok {
+		klog.Errorf("object was not Unstructured: %T", obj)
+		return err
+	}
+
+	// Get logical cluster name.
+	_, clusterAwareName, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		return err
```
stevekuznetsov, line 260: Returning this error will retry this key later, but if you can't split the key, likely that's a waste of time. Log+return nil?

stevekuznetsov: (ditto to parsing the gvr)

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+	gvrstr := parts[0]
+	gvr, _ := schema.ParseResourceArg(gvrstr)
+	if gvr == nil {
+		klog.Errorf("Error parsing GVR %q; dropping", gvrstr)
+		return nil
+	}
+	key = parts[1]
+
+	obj, exists, err := c.ddsif.IndexerFor(*gvr).GetByKey(key)
```
stevekuznetsov, line 243: Do you ever deepcopy this obj or otherwise enforce read-only?

pkg/reconciler/namespace/namespace.go
```diff
@@ -0,0 +1,162 @@
...
+	clusterName := ns.Labels[clusterLabel]
+	if clusterName == "" {
+		// Namespace is not assigned to a cluster; assign one.
+		// First, list all clusters.
+		cls, err := c.clusterLister.List(labels.Everything())
+		if err != nil {
+			return kerrors.NewRetryableError(err)
+		}
+
+		// TODO: Filter out un-Ready clusters so a namespace doesn't
```
stevekuznetsov, line 82: Label selector on your List() call?

imjasonh: Do we also apply a label on ready clusters? This might be a good followup.

pkg/reconciler/namespace/controller.go
imjasonh: It's good advice. I don't think we'd be able to have a queue per-dynamic-resource-type, but we could definitely split out namespace and cluster items from arbitrary-resource updates. I'll do that now.

pkg/informer/informer.go
```diff
@@ -0,0 +1,197 @@
...
+				DeleteFunc: func(obj interface{}) { d.handler.OnDelete(*gvr, obj) },
+			},
+		})
+		go inf.Run(ctx.Done())
+	}
+	if merr != nil {
+		return merr
+	}
+
+	d.gvrs = latest
```
stevekuznetsov, line 195: Do we need to lock on these accesses to members of `d`?

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,353 @@
...
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+	c.clusterQueue.AddRateLimited(key)
+}
+
+func (c *Controller) Start(ctx context.Context, numThreads int) {
+	c.ddsif.Start(ctx)
+
```
stevekuznetsov, line 197: nit: see the other controllers for the HasSynced waiting pattern here

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,353 @@
...
+	return c.reconcileNamespace(ctx, lclusterName, ns.DeepCopy())
+}
+
+func (c *Controller) processCluster(ctx context.Context, key string) error {
+	cluster, err := c.clusterLister.Get(key)
+	if err != nil {
+		return err
+	}
+	return c.observeCluster(ctx, cluster)
```
stevekuznetsov, line 352: The other two `process*()` funcs deep copy the item from the lister before passing on, can do that here as well, just for ease of review?

pkg/reconciler/namespace/namespace.go
```diff
@@ -0,0 +1,177 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/clusters"
+	"k8s.io/klog/v2"
+
+	"github.com/kcp-dev/kcp/pkg/apis/cluster/v1alpha1"
+)
+
+const clusterLabel = "kcp.dev/cluster"
```
stevekuznetsov, line 36: Can we put this into the API somewhere?

pkg/reconciler/namespace/namespace.go
```diff
@@ -0,0 +1,177 @@
...
+		if err != nil {
+			return err
+		}
+
+		// TODO: Filter out un-Ready clusters so a namespace doesn't
+		// get assigned to an un-Ready cluster.
+
+		if len(cls) == 0 {
+			// TODO set status, on namespace and all resources in namespace.
+			return fmt.Errorf("no clusters")
```
stevekuznetsov, line 103: Is this an error or should we just bubble up a Condition?

stevekuznetsov: ah yikes - did not read the TODO comment

pkg/server/server.go
```diff
@@ -394,6 +397,53 @@ func (s *Server) Run(ctx context.Context) error {
...
+		if err != nil {
+			return err
+		}
+		crossClusterClient := kcpClient.Cluster(clusterAll)
+
+		kubeClient := kubernetes.NewForConfigOrDie(adminConfig)
+		disco := discovery.NewDiscoveryClientForConfigOrDie(adminConfig)
+		dynClient := dynamic.NewForConfigOrDie(adminConfig)
+
+		csif := kcpexternalversions.NewSharedInformerFactory(crossClusterClient, resyncPeriod)
```
stevekuznetsov, line 417: nit: share these across controllers?

stevekuznetsov (commented):
Looking cool!

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+	_, clusterAwareName, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		runtime.HandleError(err)
+		return false
+	}
+	clusterName, _ := clusters.SplitClusterAwareKey(clusterAwareName)
+	if clusterName != "admin" {
+		klog.V(2).Infof("Skipping update for non-admin cluster %q", clusterName)
+		// TODO: Actually skip update.
```
ncdc, line 126: ?

pkg/reconciler/namespace/namespace.go
```diff
@@ -0,0 +1,177 @@
...
+	"k8s.io/client-go/tools/clusters"
+	"k8s.io/klog/v2"
+
+	"github.com/kcp-dev/kcp/pkg/apis/cluster/v1alpha1"
+)
+
+const clusterLabel = "kcp.dev/cluster"
+
+// reconcileResource is responsible for setting the cluster for a resource of
+// any type, to match the cluster where it's namespace is assigned.
```
ncdc, line 39:
```suggestion
// any type, to match the cluster where its namespace is assigned.
```

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,353 @@
...
+	return c.reconcileResource(ctx, lclusterName, unstr, gvr)
+}
+
+// namespaceBlocklist holds a set of namespaces that should never be synced from kcp to physical clusters.
+var namespaceBlocklist = sets.NewString("kube-system", "kube-public", "kube-node-lease")
+
+func (c *Controller) processNamespace(ctx context.Context, key string) error {
+	ns, err := c.namespaceLister.Get(key)
+	if err != nil {
+		return err
```
ncdc, line 333: If not found error, return nil (namespace was deleted)

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+	lclusterName, _ := clusters.SplitClusterAwareKey(clusterAwareName)
+
+	return c.reconcileNamespace(ctx, lclusterName, ns.DeepCopy())
+}
+
+func (c *Controller) processCluster(ctx context.Context, key string) error {
+	cluster, err := c.clusterLister.Get(key)
+	if err != nil {
+		return err
```
ncdc, line 350: If not found error, return nil (cluster was deleted)

pkg/server/server.go
ncdc: I'm going to file an issue and tackle it after we get this and #302 merged, if that's ok?

pkg/server/server.go
```diff
@@ -394,6 +397,53 @@ func (s *Server) Run(ctx context.Context) error {
...
+			csif.Cluster().V1alpha1().Clusters(),
+			csif.Cluster().V1alpha1().Clusters().Lister(),
+			nsif.Core().V1().Namespaces(),
+			nsif.Core().V1().Namespaces().Lister(),
+			kubeClient.CoreV1().Namespaces(),
+			gvkTrans,
+		)
+
+		if err := server.AddPostStartHook("install-namespace-scheduler", func(context genericapiserver.PostStartHookContext) error {
+			csif.WaitForCacheSync(context.StopCh)
```
ncdc, line 434: This needs to be called after Start()

pkg/server/server.go
```diff
@@ -394,6 +397,53 @@ func (s *Server) Run(ctx context.Context) error {
...
+			csif.Start(context.StopCh)
+			nsif.WaitForCacheSync(context.StopCh)
```
ncdc, line 436: This needs to be called after Start()

pkg/reconciler/namespace/controller.go
imjasonh: Will that be a problem if the lister's cache is just out of date? It might return not-found when it actually does exist, and we'll drop this update.

pkg/reconciler/namespace/controller.go
imjasonh: Same comment here.

pkg/informer/informer.go
```diff
@@ -0,0 +1,197 @@
...
+func (g GVREventHandlerFuncs) OnDelete(gvr schema.GroupVersionResource, obj interface{}) {
+	if g.DeleteFunc != nil {
+		g.DeleteFunc(gvr, obj)
+	}
+}
+
+func (d *DynamicDiscoverySharedInformerFactory) Start(ctx context.Context) {
+	// Immediately discover types and start informing.
+	if err := d.discoverTypes(ctx); err != nil {
+		klog.Fatalf("Error discovering initial types: %v", err)
```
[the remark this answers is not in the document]
imjasonh: Added a TODO

pkg/reconciler/namespace/controller.go
ncdc: The lister's cache is guaranteed to be at least as fresh as the event that triggered the shared informer's add/update/delete event handler.

This is more of a problem if you're doing a write operation and then you quickly/immediately expect that a lister's list/get invocation would show the changes from your write - those can be delayed.

Does this help?

imjasonh: Very helpful! Thanks, dropping `IsNotFound` now. 👍

pkg/informer/informer.go
```diff
@@ -0,0 +1,211 @@
...
+	var merr error
+	for _, gvrstr := range newGVRs.UnsortedList() {
+		gvr, _ := schema.ParseResourceArg(gvrstr)
+		if gvr == nil {
+			multierr.AppendInto(&merr, fmt.Errorf("parsing GVR string %q: %w", gvrstr, err))
+			continue
+		}
+
+		inf := d.dsif.ForResource(*gvr).Informer()
+		if !cache.WaitForNamedCacheSync(gvr.String(), ctx.Done(), inf.HasSynced) {
```
ncdc, line 192: FYI, this will block forever if the cache never syncs. Could be bad (?)

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,371 @@
...
+	lclusterName, _ := clusters.SplitClusterAwareKey(clusterAwareName)
+	return c.reconcileResource(ctx, lclusterName, unstr, gvr)
+}
+
+// namespaceBlocklist holds a set of namespaces that should never be synced from kcp to physical clusters.
+var namespaceBlocklist = sets.NewString("kube-system", "kube-public", "kube-node-lease")
+
+func (c *Controller) processNamespace(ctx context.Context, key string) error {
+	ns, err := c.namespaceLister.Get(key)
+	if !k8serrors.IsNotFound(err) {
```
ncdc, line 346:
```suggestion
	if k8serrors.IsNotFound(err) {
```

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,371 @@
...
+		return nil
+	}
+	lclusterName, _ := clusters.SplitClusterAwareKey(clusterAwareName)
+
+	return c.reconcileNamespace(ctx, lclusterName, ns.DeepCopy())
+}
+
+func (c *Controller) processCluster(ctx context.Context, key string) error {
+	cluster, err := c.clusterLister.Get(key)
+	if !k8serrors.IsNotFound(err) {
```
ncdc, line 365:
```suggestion
	if k8serrors.IsNotFound(err) {
```

ncdc (changes requested):

pkg/informer/informer.go
imjasonh: This was repurposed from @stevekuznetsov 's suggestion to do these checks. I can add a timeout if you'd like, but I'd like to add them everywhere consistently.

stevekuznetsov: We do this in the controllers and use the server's context for it, and we *do* see the server block forever on startup :)

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,371 @@
...
+}
+
+func (c *Controller) enqueueResource(gvr schema.GroupVersionResource, obj interface{}) {
+	key, err := cache.MetaNamespaceKeyFunc(obj)
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+	gvrstr := strings.Join([]string{gvr.Resource, gvr.Version, gvr.Group}, ".")
+	c.resourceQueue.AddRateLimited(gvrstr + "::" + key)
```
ncdc, line 180:
```suggestion
	c.resourceQueue.Add(gvrstr + "::" + key)
```

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,371 @@
...
+}
+
+func (c *Controller) enqueueNamespace(obj interface{}) {
+	key, err := cache.MetaNamespaceKeyFunc(obj)
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+	c.namespaceQueue.AddRateLimited(key)
```
ncdc, line 189:
```suggestion
	c.namespaceQueue.Add(key)
```

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,371 @@
...
+}
+
+func (c *Controller) enqueueCluster(obj interface{}) {
+	key, err := cache.MetaNamespaceKeyFunc(obj)
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+	c.clusterQueue.AddRateLimited(key)
```
ncdc, line 198:
```suggestion
	c.clusterQueue.Add(key)
```

test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -0,0 +1,350 @@
...
+
+	clusterv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/cluster/v1alpha1"
+	clientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	"github.com/kcp-dev/kcp/test/e2e/framework"
+)
+
+const clusterLabel = "kcp.dev/cluster"
+
+func TestNamespaceScheduler(t *testing.T) {
+	// TODO(jasonhall): Make tests pass.
```
stevekuznetsov, line 44: :rofl:

pkg/informer/informer.go
ncdc: Yeah, this is a bit of a special/different case because DynamicDiscoverySharedInformerFactory acts both as a shared informer factory and a "controller" that handles watch events from informers.

And now that I'm looking at this a second time, we have to move this down after `go inf.Run(ctx.Done())`. Otherwise, if we leave it here, `inf.HasSynced` will immediately return true, because that's the (perhaps unfortunate) behavior for an unstarted informer.

Which makes me think more about the relationship between an informer, HasSynced, and someone checking if an informer has synced. In the "traditional" informer/controller split, we have this sequence:

1. Create shared informer factory
2. Create controller
3. Controller adds event handlers for an informer
4. (goroutine) Start controller
    1. Wait for all informers to sync
    2. Start workers
5. (goroutine) Start all informers
6. Informer does initial list, iterating:
    1. Each item is added to informer's store
    2. Event handler is called for each item (usually to enqueue into a work queue)
9. Controller is still waiting for the full list to be processed
10. Once full list is processed, informer.HasSynced() returns true, controller can start workers
11. Workers pop keys one at a time from the work queue

Here, though, we don't have a work queue, and instead we're passing the watch events straight through to `handler`. And the call to `cache.WaitForNamedCacheSync` is done is this goroutine, while the add/update/delete handler funcs are all called async from a different goroutine. Thus, I don't think there is any easy way to wait for the informer sync here that prevents the handler funcs from being called.

I think we may need to consider adding a workqueue here too. Happy to chat on Meet if a higher bandwidth discussion medium is helpful!

pkg/informer/informer.go
```diff
@@ -0,0 +1,215 @@
...
+// checked and processed later.
+func (d *DynamicDiscoverySharedInformerFactory) Listers() (listers map[schema.GroupVersionResource]cache.GenericLister, notSynced []schema.GroupVersionResource) {
+	for _, gvrstr := range d.gvrs.UnsortedList() {
+		gvr, _ := schema.ParseResourceArg(gvrstr)
+		if gvr == nil {
+			klog.Errorf("parsing GVR string %q", gvrstr)
+			continue
+		}
+
+		if d.dsif.ForResource(*gvr).Informer().HasSynced() {
```
ncdc, line 72:
```suggestion
		if !d.dsif.ForResource(*gvr).Informer().HasSynced() {
```

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,414 @@
...
+		runtime.HandleError(err)
+		return
+	}
+	gvrstr := strings.Join([]string{gvr.Resource, gvr.Version, gvr.Group}, ".")
+	c.resourceQueue.Add(gvrstr + "::" + key)
+}
+
+func (c *Controller) enqueueGVR(gvr schema.GroupVersionResource) {
+	gvrstr := strings.Join([]string{gvr.Resource, gvr.Version, gvr.Group}, ".")
+	c.gvrQueue.AddRateLimited(gvrstr)
```
ncdc, line 188: Do you want Add or AddRateLimited here?

pkg/reconciler/namespace/controller.go
```diff
@@ -0,0 +1,414 @@
...
+		return
+	}
+	c.clusterQueue.Add(key)
+}
+
+func (c *Controller) Start(ctx context.Context, numThreads int) {
+	defer runtime.HandleCrash()
+	defer c.resourceQueue.ShutDown()
+	defer c.namespaceQueue.ShutDown()
+	defer c.clusterQueue.ShutDown()
```
ncdc, line 213: Need to defer shutting down c.gvrQueue?

pkg/reconciler/namespace/namespace.go
```diff
@@ -0,0 +1,210 @@
...
+	return nil
+}
+
+func (c *Controller) reconcileGVR(ctx context.Context, gvr schema.GroupVersionResource) error {
+	// Update all resources in the namespace with the cluster assignment.
+	listers, _ := c.ddsif.Listers()
+	lister, found := listers[gvr]
+	if !found {
+		klog.Infof("Informer for %s is not synced; re-enqueueing", gvr)
+		c.enqueueGVR(gvr)
```
ncdc, line 83: Maybe return err here instead? - it will bubble up and get requeued w/rate limiting

imjasonh: Good call, done.

ncdc (approved):

PR #203, opened 2021-11-09
Refactor server run

ncdc:
Simplify etcd Server Run() so it returns the client info instead of taking in a run func.

Replace the runFunc with inline code.

Create clients and shared informer factories higher in the stack so they can be shared by multiple controllers.

Plumb the context's Done channel to as many places as possible.

### DIFF
--- a/cmd/cluster-controller/main.go
+++ b/cmd/cluster-controller/main.go
@@ -114,8 +114,8 @@ func main() {
 	crdSharedInformerFactory.Start(ctx.Done())
 	crdSharedInformerFactory.WaitForCacheSync(ctx.Done())
 
-	clusterController.Start(numThreads, ctx.Done())
-	apiresourceController.Start(2, ctx.Done())
+	clusterController.Start(ctx, numThreads)
+	apiresourceController.Start(ctx, 2)
 
 	<-ctx.Done()
 }
--- a/cmd/cluster-controller/main.go
+++ b/cmd/cluster-controller/main.go
@@ -1,20 +1,30 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"os"
 	"os/signal"
-	"syscall"
+	"time"
 
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	typedapiextensions "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
+	crdexternalversions "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
+	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog"
 	"k8s.io/kubernetes/pkg/genericcontrolplane/clientutils"
 
+	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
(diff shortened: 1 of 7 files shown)


imjasonh (approved):

stevekuznetsov (approved):
Seems reasonable to me

ncdc:
Just pushed a new commit that further refactors by creating clients and shared informers higher so they can be shared by multiple controllers. There's probably still some more cleanup that could be done, but this seems like a good start. I'm also plumbing through context's Done channel to more places.

pkg/reconciler/apiresource/controller.go
```diff
@@ -21,56 +21,59 @@ import (
...
 const clusterNameAndGVRIndexName = "clusterNameAndGVR"
 
 func GetClusterNameAndGVRIndexKey(clusterName string, gvr metav1.GroupVersionResource) string {
 	return clusterName + "$" + gvr.String()
 }
 
-func NewController(cfg *rest.Config, autoPublishNegotiatedAPIResource bool) (*Controller, error) {
-	apiresourceClient := typedapiresource.NewForConfigOrDie(cfg)
+func NewController(
+	apiResourceClient typedapiresource.ApiresourceV1alpha1Interface,
```
stevekuznetsov, line 50: If we keep passing in the `*rest.Config` it would help my refactor be much smaller, since cluster-aware clientsets do not exist in your version of `kubernetes`, but they do in mine, and we need to use those in the controller

cmd/cluster-controller/main.go
```diff
@@ -61,27 +72,50 @@ func main() {
...
-		apiresourceController.Stop()
-	}()
+	kcpSharedInformerFactory.Start(ctx.Done())
+	kcpSharedInformerFactory.WaitForCacheSync(ctx.Done())
+
+	crdSharedInformerFactory.Start(ctx.Done())
+	crdSharedInformerFactory.WaitForCacheSync(ctx.Done())
+
+	clusterController.Start(numThreads, ctx.Done())
+	apiresourceController.Start(2, ctx.Done())
```
imjasonh, line 118: If these took a ctx as well we could pipe that down through to [`process`](https://github.com/kcp-dev/kcp/blob/8522a6b6b42151540f54dfb16c3014445d7f798c/pkg/reconciler/cluster/controller.go#L265) and into [`reconcile`](https://github.com/kcp-dev/kcp/blob/8522a6b6b42151540f54dfb16c3014445d7f798c/pkg/reconciler/cluster/cluster.go#L26), instead of having [`context.TODO()`s](https://github.com/kcp-dev/kcp/blob/8522a6b6b42151540f54dfb16c3014445d7f798c/pkg/reconciler/cluster/controller.go#L278). WDYT?

ncdc: Yes yes yes! Thanks, I'll make that change too.

pkg/reconciler/apiresource/controller.go
ncdc: Will adjust to passing in a clientset instead of typed clients. Large discussion here: https://kubernetes.slack.com/archives/C021U8WSAFK/p1636461810203200. Can adjust later if needed.

cmd/cluster-controller/main.go
ncdc: Implemented everywhere I could do it easily. There's still a couple context.TODOs left, but can do in a follow-up.

imjasonh (approved):

PR #210, opened 2021-11-17
Use the correct ctx for MemberList

ncdc:
Use the passed in ctx for MemberList instead of context.Background() so
the MemberList call respects when the ctx is done (e.g. via ctrl-c).


stevekuznetsov:
LGTM, will merge on green tests

PR #212, opened 2021-11-17
Workspace Scheduling Controller

stevekuznetsov:

### DIFF
--- a/pkg/reconciler/workspace/controller.go
+++ b/pkg/reconciler/workspace/controller.go
@@ -1,0 +1,349 @@
+/*
+Copyright 2021 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package workspace
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	jsonpatch "github.com/evanphx/json-patch"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clusters"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+
+	"github.com/kcp-dev/kcp/pkg/apis/tenancy/helpers/conditions"
+	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
+	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	tenancyinformer "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/tenancy/v1alpha1"
+	tenancylister "github.com/kcp-dev/kcp/pkg/client/listers/tenancy/v1alpha1"
+)
+
+const (
+	currentShardIndex  = "shard"
+	unschedulableIndex = "unschedulable"
+	controllerName     = "workspace"
+)
+
+func NewController(
+	kcpClient kcpclient.ClusterInterface,
+	workspaceInformer tenancyinformer.WorkspaceInformer,
+	workspaceShardInformer tenancyinformer.WorkspaceShardInformer,
+) (*Controller, error) {
+	queue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+
+	c := &Controller{
+		queue:                 queue,
+		kcpClient:             kcpClient,
+		workspaceIndexer:      workspaceInformer.Informer().GetIndexer(),
+		workspaceLister:       workspaceInformer.Lister(),
+		workspaceShardIndexer: workspaceShardInformer.Informer().GetIndexer(),
+		workspaceShardLister:  workspaceShardInformer.Lister(),
+		syncChecks: []cache.InformerSynced{
+			workspaceInformer.Informer().HasSynced,
+			workspaceShardInformer.Informer().HasSynced,
+		},
+	}
+
+	workspaceInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) { c.enqueue(obj) },
+		UpdateFunc: func(_, obj interface{}) { c.enqueue(obj) },
+	})
+	if err := c.workspaceIndexer.AddIndexers(map[string]cache.IndexFunc{
+		currentShardIndex: func(obj interface{}) ([]string, error) {
+			if workspace, ok := obj.(*tenancyv1alpha1.Workspace); ok {
+				return []string{workspace.Status.Location.Current}, nil
+			}
+			return []string{}, nil
+		},
+		unschedulableIndex: func(obj interface{}) ([]string, error) {
+			if workspace, ok := obj.(*tenancyv1alpha1.Workspace); ok {
+				if conditions.IsWorkspaceUnschedulable(workspace) {
+					return []string{"true"}, nil
+				}
+			}
+			return []string{}, nil
+		},
+	}); err != nil {
+		return nil, fmt.Errorf("failed to add indexer for Workspace: %w", err)
+	}
+
+	workspaceShardInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) { c.enqueueAddedShard(obj) },
+		DeleteFunc: func(obj interface{}) { c.enqueueDeletedShard(obj) },
+	})
+
+	return c, nil
+}
+
+// Controller watches Workspaces and WorkspaceShards in order to make sure every Workspace
+// is scheduled to a valid WorkspaceShard.
+type Controller struct {
+	queue workqueue.RateLimitingInterface
+
+	kcpClient        kcpclient.ClusterInterface
+	workspaceIndexer cache.Indexer
+	workspaceLister  tenancylister.WorkspaceLister
+
+	workspaceShardIndexer cache.Indexer
+	workspaceShardLister  tenancylister.WorkspaceShardLister
+
+	syncChecks []cache.InformerSynced
+}
+
+func (c *Controller) enqueue(obj interface{}) {
+	key, err := cache.MetaNamespaceKeyFunc(obj)
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+	klog.Infof("queueing workspace %q", key)
+	c.queue.Add(key)
+}
+
+func (c *Controller) enqueueAddedShard(obj interface{}) {
+	shard, ok := obj.(*tenancyv1alpha1.WorkspaceShard)
+	if !ok {
+		runtime.HandleError(fmt.Errorf("got %T when handling added WorkspaceShard", obj))
+		return
+	}
+	klog.Infof("handling added shard %q", shard.Name)
+	workspaces, err := c.workspaceIndexer.ByIndex(unschedulableIndex, "true")
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+	for _, workspace := range workspaces {
+		key, err := cache.MetaNamespaceKeyFunc(workspace)
+		if err != nil {
+			runtime.HandleError(err)
+			return
+		}
+		klog.Infof("queuing unschedulable workspace %q", key)
+		c.queue.Add(key)
+	}
+}
+
+func (c *Controller) enqueueDeletedShard(obj interface{}) {
+	shard, ok := obj.(*tenancyv1alpha1.WorkspaceShard)
+	if !ok {
+		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
+		if !ok {
+			klog.V(2).Infof("Couldn't get object from tombstone %#v", obj)
+			return
+		}
+		shard, ok = tombstone.Obj.(*tenancyv1alpha1.WorkspaceShard)
+		if !ok {
+			klog.V(2).Infof("Tombstone contained object that is not a WorkspaceShard: %#v", obj)
+			return
+		}
+	}
+	klog.Infof("handling removed shard %q", shard.Name)
+	workspaces, err := c.workspaceIndexer.ByIndex(currentShardIndex, shard.Name)
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+	for _, workspace := range workspaces {
+		key, err := cache.MetaNamespaceKeyFunc(workspace)
+		if err != nil {
+			runtime.HandleError(err)
+			return
+		}
+		klog.Infof("queuing orphaned workspace %q", key)
+		c.queue.Add(key)
+	}
+}
+
+func (c *Controller) Start(ctx context.Context, numThreads int) {
+	defer runtime.HandleCrash()
+	defer c.queue.ShutDown()
+
+	klog.Info("Starting Workspace controller")
+	defer klog.Info("Shutting down Workspace controller")
+
+	if !cache.WaitForNamedCacheSync(controllerName, ctx.Done(), c.syncChecks...) {
+		klog.Warning("Failed to wait for caches to sync")
+		return
+	}
+
+	for i := 0; i < numThreads; i++ {
+		go wait.Until(func() { c.startWorker(ctx) }, time.Second, ctx.Done())
+	}
+
+	<-ctx.Done()
+}
+
+func (c *Controller) startWorker(ctx context.Context) {
+	for c.processNextWorkItem(ctx) {
+	}
+}
+
+func (c *Controller) processNextWorkItem(ctx context.Context) bool {
+	// Wait until there is a new item in the working queue
+	k, quit := c.queue.Get()
+	if quit {
+		return false
+	}
+	key := k.(string)
+
+	klog.Infof("processing key %q", key)
+
+	// No matter what, tell the queue we're done with this key, to unblock
+	// other workers.
+	defer c.queue.Done(key)
+
+	if err := c.process(ctx, key); err != nil {
+		runtime.HandleError(fmt.Errorf("%q controller failed to sync %q, err: %w", controllerName, key, err))
+		c.queue.AddRateLimited(key)
+		return true
+	}
+	c.queue.Forget(key)
+	return true
+}
+
+func (c *Controller) process(ctx context.Context, key string) error {
+	namespace, clusterAwareName, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		klog.Errorf("invalid key: %q: %v", key, err)
+		return nil
+	}
+	if namespace != "" {
+		klog.Errorf("namespace %q found in key for cluster-wide Workspace object", namespace)
+		return nil
+	}
+	clusterName, name := clusters.SplitClusterAwareKey(clusterAwareName)
+
+	obj, err := c.workspaceLister.Get(key) // TODO: clients need a way to scope down the lister per-cluster
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil // object deleted before we handled it
+		}
+		return err
+	}
+	previous := obj
+	obj = obj.DeepCopy()
+
+	if err := c.reconcile(ctx, obj); err != nil {
+		return err
+	}
+
+	// If the object being reconciled changed as a result, update it.
+	if !equality.Semantic.DeepEqual(previous.Status, obj.Status) {
+		oldData, err := json.Marshal(tenancyv1alpha1.Workspace{
+			Status: previous.Status,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to Marshal old data for workspace %q|%q/%q: %w", clusterName, namespace, name, err)
+		}
+
+		newData, err := json.Marshal(tenancyv1alpha1.Workspace{
+			ObjectMeta: metav1.ObjectMeta{
+				UID:             previous.UID,
+				ResourceVersion: previous.ResourceVersion,
+			}, // to ensure they appear in the patch as preconditions
+			Status: obj.Status,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to Marshal new data for workspace %q|%q/%q: %w", clusterName, namespace, name, err)
+		}
+
+		patchBytes, err := jsonpatch.CreateMergePatch(oldData, newData)
+		if err != nil {
+			return fmt.Errorf("failed to create patch for workspace %q|%q/%q: %w", clusterName, namespace, name, err)
+		}
+		_, uerr := c.kcpClient.Cluster(clusterName).TenancyV1alpha1().Workspaces().Patch(ctx, obj.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{}, "status")
+		return uerr
+	}
+
+	return nil
+}
+
+func (c *Controller) reconcile(ctx context.Context, workspace *tenancyv1alpha1.Workspace) error {
+	if workspace.Status.BaseURL == "" {
+		// TODO: something sensible when we have some use for this field
+		workspace.Status.BaseURL = fmt.Sprintf("%s.workspaces.kcp.dev", workspace.Name)
+	}
+	var shouldReSchedule bool
+	if currentShard := workspace.Status.Location.Current; currentShard != "" {
+		// make sure current shard still exists
+		_, err := c.workspaceShardLister.Get(clusters.ToClusterAwareKey(workspace.ClusterName, currentShard))
+		if errors.IsNotFound(err) {
+			klog.Infof("de-scheduling workspace %q from nonexistent shard %q", workspace.Name, currentShard)
+			shouldReSchedule = true
+		} else if err != nil {
+			return err
+		}
+	}
+	if workspace.Status.Location.Current == "" || shouldReSchedule {
+		// find a shard for this workspace
+		shards, err := c.workspaceShardLister.List(labels.Everything())
+		if err != nil {
+			return err
+		}
+		if len(shards) != 0 {
+			target := shards[rand.Intn(len(shards))].Name
+			workspace.Status.Location.Target = target
+			klog.Infof("scheduling workspace %q to %q", workspace.Name, target)
+		}
+	}
+	if workspace.Status.Location.Current != workspace.Status.Location.Target {
+		klog.Infof("moving workspace %q from %q to %q", workspace.Name, workspace.Status.Location.Current, workspace.Status.Location.Target)
+		workspace.Status.Location.Current = workspace.Status.Location.Target
+		workspace.Status.Location.Target = ""
+	}
+	now := time.Now()
+	if workspace.Status.Location.Current == "" {
+		workspace.Status.Phase = tenancyv1alpha1.WorkspacePhaseInitializing
+		if !conditions.IsWorkspaceUnschedulable(workspace) {
+			klog.Infof("marking workspace %q unschedulable", workspace.Name)
+			conditions.SetWorkspaceCondition(workspace, tenancyv1alpha1.WorkspaceCondition{
+				Type:               tenancyv1alpha1.WorkspaceScheduled,
+				Status:             metav1.ConditionFalse,
+				LastProbeTime:      metav1.Time{Time: now},
+				LastTransitionTime: metav1.Time{Time: now},
+				Reason:             tenancyv1alpha1.WorkspaceReasonUnschedulable,
+				Message:            "No shards are available to schedule Workspaces to.",
+			})
+		}
+	} else {
+		workspace.Status.Phase = tenancyv1alpha1.WorkspacePhaseActive
+		if conditions.IsWorkspaceUnschedulable(workspace) {
+			klog.Infof("marking workspace %q scheduled", workspace.Name)
+			conditions.SetWorkspaceCondition(workspace, tenancyv1alpha1.WorkspaceCondition{
+				Type:               tenancyv1alpha1.WorkspaceScheduled,
+				Status:             metav1.ConditionTrue,
+				LastProbeTime:      metav1.Time{Time: now},
+				LastTransitionTime: metav1.Time{Time: now},
+				Reason:             "Scheduled",
+				Message:            "Workspace scheduled to a shard.",
+			})
+		}
+	}
+	return nil
+}
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -19,6 +19,7 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	apiextensionsv1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
 	crdexternalversions "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	genericapiserver "k8s.io/apiserver/pkg/server"
@@ -34,9 +35,12 @@ import (
 	"k8s.io/kubernetes/pkg/genericcontrolplane"
 	"k8s.io/kubernetes/pkg/genericcontrolplane/options"
 
+	"github.com/kcp-dev/kcp/config"
+	tenancyapi "github.com/kcp-dev/kcp/pkg/apis/tenancy"
 	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
 	kcpexternalversions "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
 	"github.com/kcp-dev/kcp/pkg/etcd"
+	"github.com/kcp-dev/kcp/pkg/reconciler/workspace"
 	"github.com/kcp-dev/kcp/pkg/sharding"
 )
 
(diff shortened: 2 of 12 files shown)


stevekuznetsov:
Depends on 
- https://github.com/kcp-dev/kcp/pull/215
- https://github.com/kcp-dev/kcp/pull/218

@ncdc @sttts @davidfestal @imjasonh @smarterclayton ready for review - the actual logic of the controller is not super well defined, so mostly looking for feedback on the e2e and overall approach

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+				}
+			}
+			return []string{}, nil
+		},
+	}); err != nil {
+		return nil, fmt.Errorf("failed to add indexer for Workspace: %w", err)
+	}
+
+	workspaceShardInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) { c.enqueueAddedShard(obj) },
```
ncdc, line 85: Do you need an UpdateFunc?

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+		return
+	}
+	klog.Infof("queueing workspace %q", key)
+	c.queue.AddRateLimited(key)
+}
+
+func (c *Controller) enqueueAddedShard(obj interface{}) {
+	shard, ok := obj.(*tenancyv1alpha1.WorkspaceShard)
+	if !ok {
+		runtime.HandleError(fmt.Errorf("got %T when handling deleted WorkspaceShard", obj))
```
ncdc, line 116:
```suggestion
		runtime.HandleError(fmt.Errorf("got %T when handling added WorkspaceShard", obj))
```

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+		return
+	}
+	for _, workspace := range workspaces {
+		key, err := cache.MetaNamespaceKeyFunc(workspace)
+		if err != nil {
+			runtime.HandleError(err)
+			return
+		}
+		klog.Infof("queuing unschedulable workspace %q", key)
+		c.queue.AddRateLimited(key)
```
ncdc, line 132: Usually you call AddRateLimited when requeuing after an error, and Add() in an enqueue func.

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+		return
+	}
+	for _, workspace := range workspaces {
+		key, err := cache.MetaNamespaceKeyFunc(workspace)
+		if err != nil {
+			runtime.HandleError(err)
+			return
+		}
+		klog.Infof("queuing orphaned workspace %q", key)
+		c.queue.AddRateLimited(key)
```
ncdc, line 155: ditto re Add

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+	klog.Errorf("Dropping key %q after failed retries: %v", key, err)
+}
+
+func (c *Controller) process(ctx context.Context, key string) error {
+	namespace, clusterAwareName, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		return err
+	}
+	if namespace != "" {
+		return fmt.Errorf("namespace %q found in key for cluster-wide Workspace object", namespace)
```
ncdc, line 225: Returning an error will cause this key to be requeued. Since it's unprocessable, you may want to consider logging and returning nil, to avoid the requeue.

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+	// Reconcile worked, nothing else to do for this workqueue item.
+	if err == nil {
+		klog.Infof("Successfully reconciled %q", key)
+		c.queue.Forget(key)
+		return
+	}
+
+	// Re-enqueue up to 5 times.
+	num := c.queue.NumRequeues(key)
+	if errors.IsRetryable(err) || num < 5 {
```
sttts, line 207: why even a limit? It's an exponential backoff. Do we care when there is a retry after n minutes?

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+	}
+	previous := obj.DeepCopy()
+
+	if err := c.reconcile(ctx, obj); err != nil {
+		return err
+	}
+
+	// If the object being reconciled changed as a result, update it.
+	if !equality.Semantic.DeepEqual(previous.Spec, obj.Spec) {
+		_, uerr := c.kcpClient.Cluster(clusterName).TenancyV1alpha1().Workspaces().Update(ctx, obj, metav1.UpdateOptions{})
```
ncdc, line 241: Update is ok for now, but this will need to be a Patch call to avoid and older client accidentally dropping fields added in a new revision of the CRD (same apiversion, fields added).

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+		return uerr
+	}
+	if !equality.Semantic.DeepEqual(previous.Status, obj.Status) {
+		_, uerr := c.kcpClient.Cluster(clusterName).TenancyV1alpha1().Workspaces().UpdateStatus(ctx, obj, metav1.UpdateOptions{})
```
ncdc, line 245: UpdateStatus is ok for now, but this will need to be a PatchStatus call to avoid and older client accidentally dropping fields added in a new revision of the CRD (same apiversion, fields added).

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+	}
+	clusterName, _ := clusters.SplitClusterAwareKey(clusterAwareName)
+
+	obj, err := c.workspaceLister.Get(key) // TODO: clients need a way to scope down the lister per-cluster
+	if err != nil {
+		return err
+	}
+	previous := obj.DeepCopy()
+
+	if err := c.reconcile(ctx, obj); err != nil {
```
sttts, line 235: you mutate obj. It is owned by the lister. Bad things happen. *always* deepcopy when getting something from a lister (a client is different, there the caller owns the object).

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+				Reason:             "Scheduled",
+				Message:            "Workspace scheduled to a shard.",
+			})
+		}
+	}
+	return nil
+}
+
+// SetWorkspaceCondition sets the status condition. It either overwrites the existing one or creates a new one.
+func SetWorkspaceCondition(workspace *tenancyv1alpha1.Workspace, newCondition tenancyv1alpha1.WorkspaceCondition) {
```
sttts, line 320: many public function. Why?

pkg/reconciler/workspace/controller.go
ncdc, line 235: If you're going to be modifying `obj`, you have to make sure you DeepCopy() it first (otherwise you mutate what's in the informer/lister's cache)

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+		return err
+	}
+	previous := obj.DeepCopy()
+
+	if err := c.reconcile(ctx, obj); err != nil {
+		return err
+	}
+
+	// If the object being reconciled changed as a result, update it.
+	if !equality.Semantic.DeepEqual(previous.Spec, obj.Spec) {
```
sttts, line 240: why does the controller change the spec. This looks like a bad API design.

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+	}
+	if !equality.Semantic.DeepEqual(previous.Status, obj.Status) {
+		_, uerr := c.kcpClient.Cluster(clusterName).TenancyV1alpha1().Workspaces().UpdateStatus(ctx, obj, metav1.UpdateOptions{})
+		return uerr
+	}
+
+	return nil
+}
+
+func (c *Controller) reconcile(ctx context.Context, obj *tenancyv1alpha1.Workspace) error {
```
ncdc, line 252: Would you be ok to rename `obj` to `workspace`?

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+	if currentShard := obj.Status.Location.Current; currentShard != "" {
+		// make sure current shard still exists
+		shards, err := c.workspaceShardLister.List(labels.Everything())
```
ncdc, line 255: Can you do a Get() here?

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+		var found bool
+		for _, shard := range shards {
+			if shard.Name == currentShard {
+				found = true
+				break
+			}
+		}
+		if !found {
+			klog.Infof("removing spec.targetShard for workspace %q", obj.Name)
+			obj.Spec.TargetShard = ""
```
sttts, line 268: a controller should not do this. It is the intent of the creator of the object. Don't override that. If it is not possible to "apply" that, tell the user in a condition.

sttts: I can see that you set the target if it has been empty before. It's also what kube-scheduler does. But already that is a questionable behaviour, but one can argue that it is like defaulting basically.

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+		}
+	} else {
+		obj.Status.Phase = tenancyv1alpha1.WorkspacePhaseAtive
+		if IsWorkspaceUnschedulable(obj) {
+			klog.Infof("marking workspace %q scheduled", obj.Name)
+			SetWorkspaceCondition(obj, tenancyv1alpha1.WorkspaceCondition{
+				Type:               tenancyv1alpha1.WorkspaceScheduled,
+				Status:             metav1.ConditionTrue,
+				LastProbeTime:      metav1.Time{Time: time.Now()},
+				LastTransitionTime: metav1.Time{Time: time.Now()},
```
sttts, line 310: this is bad style. You logic might avoid hotlooping, but better have a helper func that sets the condition, comparing the old to the new status and only then set LastTransitionTime.

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+		}
+	}
+	if obj.Spec.TargetShard != obj.Status.Location.Current {
+		klog.Infof("updating workspace %q status", obj.Name)
+		obj.Status.Location.Current = obj.Spec.TargetShard
+		obj.Status.Location.Target = ""
+	}
+	if obj.Status.Location.Current == "" {
+		obj.Status.Phase = tenancyv1alpha1.WorkspacePhaseInitializing
+		if !IsWorkspaceUnschedulable(obj) {
```
ncdc, line 291: If it's not unschedulable, then it's schedulable? Help me understand this flow?

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+				Type:               tenancyv1alpha1.WorkspaceScheduled,
+				Status:             metav1.ConditionFalse,
+				LastProbeTime:      metav1.Time{Time: time.Now()},
+				LastTransitionTime: metav1.Time{Time: time.Now()},
+				Reason:             tenancyv1alpha1.WorkspaceReasonUnschedulable,
+				Message:            "No shards are available to schedule Workspaces to.",
+			})
+		}
+	} else {
+		obj.Status.Phase = tenancyv1alpha1.WorkspacePhaseAtive
```
sttts, line 303: typo

ncdc, line 303: typo: Ative

pkg/reconciler/workspace/controller.go
sttts: all these helpers should probable go to pkg/apis/.../helpers

sttts: or rather pkg/apis/.../helpers/conditions

pkg/server/server.go
```diff
@@ -392,6 +393,55 @@ func (s *Server) Run(ctx context.Context) error {
 		}
 	}
 
+	if s.cfg.InstallWorkspaceController {
+		kubeconfig := clientConfig.DeepCopy()
```
sttts, line 397: why deepcopy?

config/tenancy.kcp.dev_workspaces.yaml
```diff
@@ -24,10 +24,14 @@ spec:
...
+            description: 'APIVersion defines the versioned schema of this representation
+              of an object. Servers should convert recognized schemas to the latest
+              internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
             type: string
           kind:
-            description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
+            description: 'Kind is a string value representing the REST resource this
+              object represents. Servers may infer this from the endpoint the client
+              submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
```
sttts, line 34: can we split generated code into its own commit? Makes reviews hard otherwise.

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+
+	workspaceShardInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) { c.enqueueAddedShard(obj) },
+		DeleteFunc: func(obj interface{}) { c.enqueueDeletedShard(obj) },
+	})
+
+	return c, nil
+}
+
+type Controller struct {
```
sttts, line 92: a 2 paragraph descriptions would be nice. Makes it much easier once we have 17 controllers and you lose the overview.

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,385 @@
...
+	workspaceShardLister  tenancylister.WorkspaceShardLister
+}
+
+func (c *Controller) enqueue(obj interface{}) {
+	key, err := cache.MetaNamespaceKeyFunc(obj)
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+	klog.Infof("queueing workspace %q", key)
```
sttts, line 109: general question: shouldn't we move to structured logging right away?

test/e2e/framework/kcp.go
```diff
@@ -0,0 +1,223 @@
...
+		artifactDir: artifactDir,
+		ctx:         ctx,
+		t:           t,
+		lock:        &sync.Mutex{},
+	}, nil
+}
+
+// Run runs the kcp server while the parent context is active. This call is not blocking,
+// callers should ensure that the server is Ready() before using it.
+func (c *kcpServer) Run(parentCtx context.Context) error {
```
sttts, line 82: :+1: good framework.

Btw, I notice that your out-of-process design has a nice side effect: it is hard to work around the flag interface. We are a small group now, but in upstream you have to carefully watch and review every PR against such a package because people like to add their own special tweak to the test server.

test/e2e/framework/kcp.go
```diff
@@ -0,0 +1,223 @@
...
+		cfg.NegotiatedSerializer = scheme.Codecs.WithoutConversion()
+	}
+	client, err := rest.UnversionedRESTClientFor(cfg)
+	if err != nil {
+		return fmt.Errorf("failed to create unversioned client: %w", err)
+	}
+
+	wg := sync.WaitGroup{}
+	wg.Add(3)
+	for _, endpoint := range []string{"/livez", "/readyz", "/healthz"} {
```
sttts, line 153: one includes the other. readyz green => healthz green. Not sure about livez.

test/e2e/framework/test.go
```diff
@@ -0,0 +1,98 @@
...
+// executed for each test case. We must control the scope in which each test runs, to
+// be able to defer() actions that cancel background processes when a test finishes
+// execution, as child goroutines are not by default killed when a test finishes and
+// interaction with the parent testing.T after the test is finished causes a panic.
+// Furthermore, we could not simply use the central testing.T for managing control
+// flow as it is not allowed to call testing.T.FailNow (via Fatal, etc) in anything
+// other than the main testing goroutine. Therefore, when more than one routine needs
+// to be able to influence the execution flow (e.g. preempt other routines) we must
+// have the central routine watch for incoming errors from delegate routines.
+func Run(top *testing.T, name string, f TestFunc, args ...[]string) {
```
ncdc, line 45: I think this is worthy of a live walkthrough 😄

pkg/reconciler/workspace/controller.go
stevekuznetsov: I'll admit this was copy-pasta from some other controller we have in the codebase. You suggest removing the limit for all?

pkg/reconciler/workspace/controller.go
stevekuznetsov: How would you expect the scheduler to do its job without changing the spec?

pkg/reconciler/workspace/controller.go
stevekuznetsov: We need a different design (`WorkspaceBinding`?) if we don't want to mutate spec.

pkg/server/server.go
stevekuznetsov: Also copy-pasta, no clue

pkg/reconciler/workspace/controller.go
stevekuznetsov: I'm all for it - how, using `klog`, is that supported?

pkg/reconciler/workspace/controller.go
stevekuznetsov: There's currently nothing on a shard that can change and affect workspaces (no capacity impl), so no.

test/e2e/framework/util.go
```diff
@@ -0,0 +1,85 @@
...
+		lockDir := filepath.Join(os.TempDir(), "kcp-e2e-ports")
+		if err := os.MkdirAll(lockDir, 0777); err != nil {
+			return "", fmt.Errorf("could not create port lockfile dir: %v", err)
+		}
+		lockFile := filepath.Join(lockDir, port)
+		if _, err := os.Stat(lockFile); os.IsNotExist(err) {
+			// we've never seen this before, we can use it
+			f, err := os.Create(lockFile)
+			if err != nil {
+				return "", fmt.Errorf("could not record port lockfile: %w", err)
```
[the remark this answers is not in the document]
stevekuznetsov: No, because we're using the kernel-provided connection open as a "lock" for this critical section. Nobody else can listen on that port before we're done.

test/e2e/framework/test.go
stevekuznetsov: Yes!

pkg/reconciler/workspace/controller.go
sttts: I would. Set the backoff to 10 min (might be default) and we are good.

pkg/reconciler/workspace/controller.go
sttts: see my comment about unsetting spec.target. I think the converse is fine (like defaulting). But unsetting because some precondition (existence of that shard) are not given, looks very strange.

pkg/reconciler/workspace/controller.go
sttts: I reviewed some PRs in kube, klog.InfoS I believe. Never done that before. But if we start fresh anyway, do it right and enjoy beautiful json log in the future.

pkg/reconciler/workspace/controller.go
sttts: setting it to "" is strange. Setting "" to some value is more natural. Why do we need this "unschedule" case?

test/e2e/framework/kcp.go
```diff
@@ -0,0 +1,223 @@
...
+
+// Config exposes a copy of the client config for this server.
+func (c *kcpServer) Config() *rest.Config {
+	c.lock.Lock()
+	cfg := c.cfg
+	c.lock.Unlock()
+	if cfg == nil {
+		return nil
+	}
+	return rest.CopyConfig(c.cfg)
```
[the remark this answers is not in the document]
stevekuznetsov: The config is read by a different goroutine than it's written by, so I synchronized access to it.

pkg/reconciler/workspace/controller.go
stevekuznetsov: If it's not already unschedulable, make it so. I borrowed the conditions from Pod, and I also do not like the pattern but at least it has precedent? Open to other ideas for sur

pkg/reconciler/workspace/controller.go
stevekuznetsov: I just wanted something sensible to handle shard movement for now, is all

pkg/reconciler/workspace/controller.go
stevekuznetsov: We'd need to create the cluster-aware key for the name, which would mean we assume the shard is in the same lcluster as the workspace... probably fine? WDYT?

pkg/reconciler/workspace/controller.go
stevekuznetsov: It's honestly not even that clear to me what this controller should be doing at the moment or what flows we want/need/can support without more machinery ... hence the confusion.

stevekuznetsov:
@ncdc @sttts addressed the simpler comments, broke into more commits that can be split out of this PR and/or for review

pkg/reconciler/workspace/controller.go
stevekuznetsov: https://github.com/kcp-dev/kcp/pull/212#discussion_r752604413 will handle this

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,318 @@
...
+
+func (c *Controller) handleErr(err error, key string) {
+	// Reconcile worked, nothing else to do for this workqueue item.
+	if err == nil {
+		klog.Infof("Successfully reconciled %q", key)
+		c.queue.Forget(key)
+		return
+	}
+
+	if errors.IsRetryable(err) {
```
sttts, line 206: compare https://github.com/kcp-dev/kcp/pull/219/files#r752915735. I think this logic is wrong (and kind of dangerous if we continue this and spread the pattern).

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -93,13 +75,71 @@ type WorkspaceStatus struct {
...
+type WorkspaceLocation struct {
+	// Current workspace placement (shard).
+	//
+	// +optional
+	Current string `json:"current,omitempty"`
+
+	// Target workspace placement (shard).
+	//
+	// +optional
+	Target string `json:"target,omitempty"`
```
sttts, line 121: how does this differ from the spec target?

pkg/reconciler/workspace/controller.go
sttts: Something like: watches workspaces without spec.target set and tries to find a shard this workspace can be assigned to. The shard then is written to ....

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,318 @@
...
+		return nil
+	}
+	clusterName, _ := clusters.SplitClusterAwareKey(clusterAwareName)
+
+	obj, err := c.workspaceLister.Get(key) // TODO: clients need a way to scope down the lister per-cluster
+	if err != nil {
+		return err
+	}
+	obj = obj.DeepCopy()
+	previous := obj.DeepCopy()
```
sttts, line 234: nit: previous could be the old obj before deep copy

pkg/reconciler/workspace/controller.go
sttts: also this UpdateStatus will definitely fail if the spec is changed too above because of new RV. Not good. It also shows why changing spec and status in one controller is flawed.

pkg/reconciler/workspace/controller.go
ncdc: +1

pkg/reconciler/workspace/controller.go
ncdc: I'd say it's probably fine, for now at least. We'll know if it breaks later 😄

test/e2e/framework/kcp.go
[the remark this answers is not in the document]
ncdc: You're able to determine if c.cfg is nil without any race, but what about this CopyConfig call - it's using c.cfg directly. Could that race (would any other goroutine be setting c.cfg async)?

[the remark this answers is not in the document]
stevekuznetsov: Oh, yikes, sorry - that is an oversight :(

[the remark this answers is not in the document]
ncdc: And cfg will be a shallow copy of c.cfg - is it safe to call rest.CopyConfig outside of the lock?

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,318 @@
...
+	// Give up and report error elsewhere.
+	c.queue.Forget(key)
+	runtime.HandleError(err)
+	klog.Errorf("Dropping key %q after failed retries: %v", key, err)
+}
+
+func (c *Controller) process(ctx context.Context, key string) error {
+	namespace, clusterAwareName, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
ncdc: Bump?

pkg/apis/tenancy/helpers/conditions/conditions.go
```diff
@@ -0,0 +1,77 @@
...
+			return condition.Status == status
+		}
+	}
+	return false
+}
+
+func IsWorkspaceUnschedulable(workspace *v1alpha1.Workspace) bool {
+	if IsWorkspaceConditionFalse(workspace, v1alpha1.WorkspaceScheduled) {
+		return FindWorkspaceCondition(workspace, v1alpha1.WorkspaceScheduled).Reason == v1alpha1.WorkspaceReasonUnschedulable
+	}
```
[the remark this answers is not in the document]
stevekuznetsov: https://github.com/kcp-dev/kcp/issues/246

pkg/reconciler/workspace/controller.go
stevekuznetsov: https://github.com/kcp-dev/kcp/issues/246

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	// If the object being reconciled changed as a result, update it.
+	if !equality.Semantic.DeepEqual(previous.Status, obj.Status) {
+		oldData, err := json.Marshal(tenancyv1alpha1.Workspace{
+			Status: previous.Status,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to Marshal old data for workspace %q|%q/%q: %v", clusterName, namespace, name, err)
```
ncdc, line 251: Are we wrapping errors (%w) or does it not matter?

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,337 @@
...
+			return err
+		}
+		if len(shards) != 0 {
+			target := shards[rand.Intn(len(shards))].Name
+			workspace.Status.Location.Target = target
+			klog.Infof("scheduling workspace %q to %q", workspace.Name, target)
+		}
+	}
+	if workspace.Status.Location.Current != workspace.Status.Location.Target {
+		klog.Infof("moving workspace %q from %q to %q", workspace.Name, workspace.Status.Location.Current, workspace.Status.Location.Target)
```
ncdc, line 305: Thinking through at least the log messages: if we have a new unscheduled workspace (current=""), we'll see "scheduling workspace Foo to Bar" from L301, then I'm assuming because current is still empty and we just set target, we'll see "moving workspace Foo from '' to 'Bar'"?

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+	if workspace.Status.Location.Current == "" {
+		workspace.Status.Phase = tenancyv1alpha1.WorkspacePhaseInitializing
+		if !conditions.IsWorkspaceUnschedulable(workspace) {
+			klog.Infof("marking workspace %q unschedulable", workspace.Name)
+			conditions.SetWorkspaceCondition(workspace, tenancyv1alpha1.WorkspaceCondition{
+				Type:               tenancyv1alpha1.WorkspaceScheduled,
+				Status:             metav1.ConditionFalse,
+				LastProbeTime:      metav1.Time{Time: time.Now()},
+				LastTransitionTime: metav1.Time{Time: time.Now()},
```
ncdc, line 317: Do you want to use a single `time.Now()` call here instead of 2?

pkg/reconciler/workspace/controller.go
```diff
@@ -0,0 +1,337 @@
...
+		}
+	} else {
+		workspace.Status.Phase = tenancyv1alpha1.WorkspacePhaseActive
+		if conditions.IsWorkspaceUnschedulable(workspace) {
+			klog.Infof("marking workspace %q scheduled", workspace.Name)
+			conditions.SetWorkspaceCondition(workspace, tenancyv1alpha1.WorkspaceCondition{
+				Type:               tenancyv1alpha1.WorkspaceScheduled,
+				Status:             metav1.ConditionTrue,
+				LastProbeTime:      metav1.Time{Time: time.Now()},
+				LastTransitionTime: metav1.Time{Time: time.Now()},
```
ncdc, line 330: Do you want to use a single `time.Now()` call here instead of 2?

test/e2e/framework/kcp.go
```diff
@@ -0,0 +1,274 @@
+package framework
```
ncdc, line 1: Add boilerplate?

test/e2e/framework/t.go
```diff
@@ -0,0 +1,80 @@
+package framework
```
ncdc, line 1: Add boilerplate?

test/e2e/framework/test.go
```diff
@@ -0,0 +1,100 @@
+package framework
```
ncdc, line 1: Add boilerplate?

test/e2e/framework/testhelper.go
```diff
@@ -0,0 +1,140 @@
+package framework
```
ncdc, line 1: Add boilerplate?

test/e2e/framework/testhelper_test.go
```diff
@@ -0,0 +1,42 @@
+package framework
```
ncdc, line 1: Add boilerplate?

test/e2e/framework/util.go
```diff
@@ -0,0 +1,85 @@
+package framework
```
ncdc, line 1: Add boilerplate?

test/e2e/reconciler/workspace/controller_test.go
```diff
@@ -0,0 +1,306 @@
+package workspace
```
ncdc, line 1: Add boilerplate?

test/e2e/framework/kcp.go
```diff
@@ -0,0 +1,274 @@
...
+		c.t.Errorf(lastMsg)
+	}
+}
+
+func (c *kcpServer) monitorEndpoint(client *rest.RESTClient, endpoint string) {
+	// we need a shorter deadline than the server, or else:
+	// timeout.go:135] post-timeout activity - time-elapsed: 23.784917ms, GET "/livez" result: Header called after Handler finished
+	ctx := c.ctx
+	if deadline, ok := c.t.Deadline(); ok {
+		deadlinedCtx, deadlinedCancel := context.WithDeadline(c.ctx, deadline.Add(-20*time.Second))
```
ncdc, line 264: What happens if the test's deadline is < 20 seconds from now when this line is called - won't deadlinedCtx immediately be canceled?

test/e2e/framework/test.go
```diff
@@ -0,0 +1,100 @@
...
+// be able to defer() actions that cancel background processes when a test finishes
+// execution, as child goroutines are not by default killed when a test finishes and
+// interaction with the parent testing.T after the test is finished causes a panic.
+// Furthermore, we could not simply use the central testing.T for managing control
+// flow as it is not allowed to call testing.T.FailNow (via Fatal, etc) in anything
+// other than the main testing goroutine. Therefore, when more than one routine needs
+// to be able to influence the execution flow (e.g. preempt other routines) we must
+// have the central routine watch for incoming errors from delegate routines.
+func Run(top *testing.T, name string, f TestFunc, args ...[]string) {
+	if _, previouslyCalled := seen.LoadOrStore(fmt.Sprintf("%p", top), nil); !previouslyCalled {
```
ncdc, line 46: When would `top` be seen multiple times - multiple Test funcs in the same package that call `framework.Run`?

stevekuznetsov: Yep, exactly in that case

test/e2e/framework/kcp.go
stevekuznetsov: We'd have to intentionally do this to ourselves, since the default timeout is 10 minutes, so that would be silly. Our simplest e2e already need to wait for the server to be ready and do some HTTP calls, so we're getting close to 20s as it is for normal runtime. I can add a guard in the code.

pkg/reconciler/workspace/controller.go
stevekuznetsov: I will add a commit to enforce this.

test/e2e/framework/kcp.go
stevekuznetsov: I will add a commit to enforce this

pkg/reconciler/workspace/controller.go
stevekuznetsov: Likely below our bar for worrying about it right now? The flow of this controller is 100% guaranteed to change in the short term when we have actual scheduling concepts in

ncdc: Yes, totally fine, thanks

pkg/reconciler/workspace/controller.go
stevekuznetsov: https://github.com/kcp-dev/kcp/pull/255

test/e2e/framework/kcp.go
stevekuznetsov: https://github.com/kcp-dev/kcp/pull/258

pkg/reconciler/workspace/controller.go
stevekuznetsov: Hilariously, the `golangci-lint` `errorlint` check did not fail on this :facepalm:

stevekuznetsov:
Needs https://github.com/kcp-dev/kcp/pull/258 for real verification but otherwise this addressed all comments

davidfestal (approved):
@ncdc  @sttts Is there anything that prevents moving forward and merging this PR ?

ncdc:
Giving it one more quick look

ncdc:
Going to merge. If we need to fix anything, can do in a follow-up.

PR #214, opened 2021-11-17
techdebt: use klog/v2

stevekuznetsov:
The rest of k8s core is already on v2, there's no reason for us to use
the older library and it cuases some pain when using both at once.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	k8s.io/apiserver v0.0.0
 	k8s.io/client-go v0.0.0
 	k8s.io/code-generator v0.0.0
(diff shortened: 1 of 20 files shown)


go.mod
```diff
@@ -19,7 +19,7 @@ require (
 	k8s.io/apiserver v0.0.0
 	k8s.io/client-go v0.0.0
 	k8s.io/code-generator v0.0.0
-	k8s.io/klog v1.0.0
+	k8s.io/klog/v2 v2.9.0
```
stevekuznetsov, line 22: This is the version used in our fork of k8s

ncdc (approved):

PR #215, opened 2021-11-18
Misc fixes

stevekuznetsov:
config: allow configuring kubeconfig path

---

server: allow absolute data directories

---

server: start controllers in their own goroutines

The code today blocks on the first Start() call, ensuring that the
second controller never starts and the server is never ready.

---

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -88,13 +88,18 @@ func (s *Server) Run(ctx context.Context) error {
 		go http.ListenAndServe(s.cfg.ProfilerAddress, nil)
 	}
 
-	dir := filepath.Join(".", s.cfg.RootDirectory)
-	if len(s.cfg.RootDirectory) != 0 {
+	var dir string
+	if filepath.IsAbs(s.cfg.RootDirectory) {
+		dir = s.cfg.RootDirectory
+	} else {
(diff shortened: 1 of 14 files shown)


pkg/server/server.go
```diff
@@ -88,13 +87,18 @@ func (s *Server) Run(ctx context.Context) error {
...
 	}
 
-	dir := filepath.Join(".", s.cfg.RootDirectory)
+	var dir string
+	if filepath.IsAbs(s.cfg.RootDirectory) {
+		dir = s.cfg.RootDirectory
+	} else {
+		dir = filepath.Join(".", s.cfg.RootDirectory)
+	}
 	if len(s.cfg.RootDirectory) != 0 {
```
ncdc, line 96: Should this be checking `dir` instead?

ncdc: If not, feel free to ignore 😄

stevekuznetsov: /shrug

I don't see a semantic difference between the two and was happy with the smaller diff. We default the value anyway so defending against `len() == 0` is on the silly side.

stevekuznetsov:
@ncdc this is updated for your comments :)

ncdc (approved):

PR #219, opened 2021-11-18
controllers: cleanup from other comments

stevekuznetsov:

/assign @ncdc @sttts 

These are issues that you found in my PR but were cargo-culted from other controllers, so I expanded the review to them, too.


pkg/reconciler/apiresource/controller.go
```diff
@@ -331,10 +331,8 @@ func (c *Controller) handleErr(err error, key queueElement) {
 		return
 	}
 
-	// Re-enqueue up to 5 times.
-	num := c.queue.NumRequeues(key)
-	if errors.IsRetryable(err) || num < 5 {
-		klog.Infof("Error reconciling key %q, retrying... (#%d): %v", key, num, err)
+	if errors.IsRetryable(err) {
```
sttts, line 334: I wondered what this IsRetryable actually is. It's from our utils package. Why do we need it? Why do we forget the key with every other error? I think that's wrong. We should blindly retry on error, and exponential backoff will do its job to keep retries sane.

cc @davidfestal

stevekuznetsov: OK, I can remove the util and simply retry blindly.

stevekuznetsov:
@sttts updated, removed the entire `Retryable()` concept

ncdc (approved):

PR #224, opened 2021-11-19
config: add a simple bootstrapping routine for CRDs

stevekuznetsov:
Today, we have a helper to apply some YAML to the default logical
cluster. However, we would like to be able to apply specific CRDs, as
not every controller needs all possible CRDs, we would like to direct
the application to specific logical clusters, and we would like to wait
for the system to establish the CRDs before moving on, or client calls
to these APIs will fail if made quickly after they are introduced. This
patch adds all of that functionality, as well as using Go's embedding
for the data so that the routine is portable.

### DIFF
--- a/config/bootstrap.go
+++ b/config/bootstrap.go
@@ -1,0 +1,107 @@
+package config
+
+import (
+	"context"
+	"embed"
+	"fmt"
+	"sync"
+	"time"
+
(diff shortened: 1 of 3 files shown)


stevekuznetsov:
Fixes https://github.com/kcp-dev/kcp/issues/225

ncdc (approved):

davidfestal (commented):
Thanks @stevekuznetsov for this fix

PR #226, opened 2021-11-22
Start controllers in its goroutines

vincent-pli:
When try to start `cluster-controller` by calling
```
go run ./cmd/cluster-controller \
    --syncer_image=$(ko publish ./cmd/syncer) \
    --kubeconfig=.kcp/admin.kubeconfig
```

It's not work, since the `apiresourceController` is not started...


ncdc:
Thanks for this @vincent-pli!

ncdc (approved):

PR #228, opened 2021-11-22
DRY out server startup

stevekuznetsov:
config: use a more common pattern for binding flags

---

server: refactor cluster controller to DRY it out

The server options loading and startup code is identical between the
standalone binary and the post-start-hook. This commit makes the code
actually shared.

---

### DIFF
--- a/cmd/cluster-controller/main.go
+++ b/cmd/cluster-controller/main.go
@@ -2,112 +2,78 @@ package main
 
 import (
 	"context"
-	"flag"
+	"errors"
 	"os"
 	"os/signal"
 	"time"
+
+	"github.com/spf13/pflag"
 
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	crdexternalversions "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
-	"k8s.io/kubernetes/pkg/genericcontrolplane/clientutils"
 
 	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
 	kcpexternalversions "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
-	"github.com/kcp-dev/kcp/pkg/reconciler/apiresource"
 	"github.com/kcp-dev/kcp/pkg/reconciler/cluster"
 )
 
-const numThreads = 2
 const resyncPeriod = 10 * time.Hour
 
-var (
-	kubeconfigPath  = flag.String("kubeconfig", "", "Path to kubeconfig")
-	syncerImage     = flag.String("syncer_image", "", "Syncer image to install on clusters")
-	pullMode        = flag.Bool("pull_mode", true, "Deploy the syncer in registered physical clusters in POD, and have it sync resources from KCP")
-	pushMode        = flag.Bool("push_mode", false, "If true, run syncer for each cluster from inside cluster controller")
-	autoPublishAPIs = flag.Bool("auto_publish_apis", false, "If true, the APIs imported from physical clusters will be published automatically as CRDs")
-)
+func bindOptions(fs *pflag.FlagSet) *options {
+	o := options{
+		Options: cluster.BindOptions(fs),
+	}
+	fs.StringVar(&o.kubeconfigPath, "kubeconfig", "", "Path to kubeconfig")
+	return &o
+}
+
+type options struct {
+	// in the all-in-one startup, client credentials already exist; in this
+	// standalone startup, we need to load credentials ourselves
+	kubeconfigPath string
+	*cluster.Options
+}
+
+func (o *options) Validate() error {
+	if o.kubeconfigPath == "" {
+		return errors.New("--kubeconfig is required")
+	}
+	return o.Options.Validate()
+}
 
 func main() {
 	// Setup signal handler for a cleaner shutdown
 	ctx, cancel := signal.NotifyContext(context.Background(), os.Kill, os.Interrupt)
 	defer cancel()
 
-	flag.Parse()
+	fs := pflag.NewFlagSet("cluster-controller", pflag.ExitOnError)
+	options := bindOptions(fs)
+	if err := fs.Parse(fs.Args()[1:]); err != nil {
+		klog.Fatal(err)
+	}
+	if err := options.Validate(); err != nil {
+		klog.Fatal(err)
+	}
 
 	configLoader := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
-		&clientcmd.ClientConfigLoadingRules{ExplicitPath: *kubeconfigPath},
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: options.kubeconfigPath},
 		&clientcmd.ConfigOverrides{})
 
 	r, err := configLoader.ClientConfig()
 	if err != nil {
 		klog.Fatal(err)
 	}
-	clientutils.EnableMultiCluster(r, nil, true, "clusters", "customresourcedefinitions", "apiresourceimports", "negotiatedapiresources")
 	kubeconfig, err := configLoader.RawConfig()
 	if err != nil {
 		klog.Fatal(err)
 	}
-
-	resourcesToSync := flag.Args()
-	if len(resourcesToSync) == 0 {
-		resourcesToSync = []string{"deployments.apps"}
-	}
-	klog.Infof("Syncing resources: %v", resourcesToSync)
-
-	if *pullMode && *pushMode {
-		klog.Fatal("can't set --push_mode and --pull_mode")
-	}
-	syncerMode := cluster.SyncerModeNone
-	if *pullMode {
-		syncerMode = cluster.SyncerModePull
-	}
-	if *pushMode {
-		syncerMode = cluster.SyncerModePush
-	}
-
 	kcpSharedInformerFactory := kcpexternalversions.NewSharedInformerFactoryWithOptions(kcpclient.NewForConfigOrDie(r), resyncPeriod)
 	crdSharedInformerFactory := crdexternalversions.NewSharedInformerFactoryWithOptions(apiextensionsclient.NewForConfigOrDie(r), resyncPeriod)
-
-	apiExtensionsClient := apiextensionsclient.NewForConfigOrDie(r)
-	kcpClient := kcpclient.NewForConfigOrDie(r)
-
-	clusterController, err := cluster.NewController(
-		apiExtensionsClient,
-		kcpClient,
-		kcpSharedInformerFactory.Cluster().V1alpha1().Clusters(),
-		kcpSharedInformerFactory.Apiresource().V1alpha1().APIResourceImports(),
-		*syncerImage,
-		kubeconfig,
-		resourcesToSync,
-		syncerMode,
-	)
-	if err != nil {
+	if err := options.Options.Complete(kubeconfig, kcpSharedInformerFactory, crdSharedInformerFactory).Start(ctx); err != nil {
 		klog.Fatal(err)
 	}
-
-	apiresourceController, err := apiresource.NewController(
-		apiExtensionsClient,
-		kcpClient,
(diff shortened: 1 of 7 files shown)


pkg/reconciler/cluster/startup.go
```diff
@@ -0,0 +1,128 @@
...
+		kcpSharedInformerFactory.Apiresource().V1alpha1().APIResourceImports(),
+		crdSharedInformerFactory.Apiextensions().V1().CustomResourceDefinitions(),
+	)
+	if err != nil {
+		return err
+	}
+
+	kcpSharedInformerFactory.Start(ctx.Done())
+	crdSharedInformerFactory.Start(ctx.Done())
+	kcpSharedInformerFactory.WaitForCacheSync(ctx.Done())
```
ncdc, line 121: And if we don't start the factories here, we can't wait for cache sync directly here either. We can either put both the waiting and the controller start in a goroutine, or we can pass individual informer.Synced pointers into each controller, and wait on those inside each controller's Start func.

pkg/reconciler/cluster/startup.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (o *Options) Validate() error {
+	if o.PullMode && o.PushMode {
+		return errors.New("can't set both --push_mode and --pull_mode")
+	}
+	return nil
+}
+
+func (o *Options) Start(ctx context.Context, kubeconfig clientcmdapi.Config, kcpSharedInformerFactory kcpexternalversions.SharedInformerFactory, crdSharedInformerFactory crdexternalversions.SharedInformerFactory) error {
```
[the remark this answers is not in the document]
stevekuznetsov: I can add a `Complete()` or sth to take flag-options and create server-options, yes

pkg/reconciler/cluster/startup.go
```diff
@@ -0,0 +1,128 @@
...
+		o.AutoPublishAPIs,
+		kcpSharedInformerFactory.Apiresource().V1alpha1().NegotiatedAPIResources(),
+		kcpSharedInformerFactory.Apiresource().V1alpha1().APIResourceImports(),
+		crdSharedInformerFactory.Apiextensions().V1().CustomResourceDefinitions(),
+	)
+	if err != nil {
+		return err
+	}
+
+	kcpSharedInformerFactory.Start(ctx.Done())
```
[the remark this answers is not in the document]
stevekuznetsov: Would that work, though? I thought in each post-start-hook we'd be starting the server, which waits for the informers to sync, which requires them to be started, which ... ?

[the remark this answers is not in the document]
ncdc: Yes, my comment below talks about it. This is probably what I'd do:

1) 
```go
type MyController struct {
  someInformerSynced cache.InformerSynced
}

func NewMyController(
  // ...,
  someInformer informers.SomeInformer,
  // ...,
) *MyController {
  mc := &MyController{
    someInformerSynced: someInformer.Informer().HasSynced
  }
  return mc
}

func (mc *MyController) Run(ctx context.Context, workers int) {
  defer utilruntime.HandleCrash()
  defer mc.queue.ShutDown()
  
  klog.Infof("Starting my controller")
  defer klog.Infof("Shutting down my controller")
  
  if !cache.WaitForNamedCacheSync("my controller", ctx.Done(), mc.informerSynced) {
    return
  }
  
  klog.V(5).Info("Starting workers of my controller")
  for i := 0; i < workers; i++ {
    go wait.Until(mc.worker, time.Second, ctx.Done())
  }
  <-ctx.Done()
}
```

It lets each controller set up which informers it needs to be fully synced before starting work.

pkg/reconciler/cluster/startup.go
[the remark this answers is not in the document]
ncdc: Thanks, I think that would look better

pkg/reconciler/cluster/startup.go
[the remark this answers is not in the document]
stevekuznetsov: Ah, I see

stevekuznetsov:
@ncdc updated to address the comments, but likely we want to wait for #231 to test the result :)

stevekuznetsov:
@ncdc so it turns out for all I can see, post-start-hooks run in parallel. Therefore, we cannot start the informers in one. I'm not really sure why we would need to, anyway. We need to do the following in a serial thread, in order:

 - create the informer factories
 - use factories to create informers
 - start factories

We can start the controllers in a post-start hook.

ncdc:
We can start the informers in a post start hook. Will discuss next week 😀

stevekuznetsov:
@ncdc ok, I'm happy to see how - maybe the catch is just to *use* the factories in the serial thread, before launching the hook to star them ... ? I will look at kcm later, this startup ordering is pretty tricky and I'm sure there's a nice pattern for us to copy from there...

cmd/cluster-controller/main.go
```diff
@@ -2,112 +2,78 @@ package main
...
 
 func main() {
 	// Setup signal handler for a cleaner shutdown
 	ctx, cancel := signal.NotifyContext(context.Background(), os.Kill, os.Interrupt)
 	defer cancel()
 
-	flag.Parse()
+	fs := pflag.NewFlagSet("cluster-controller", pflag.ExitOnError)
+	options := bindOptions(fs)
+	if err := fs.Parse(fs.Args()[1:]); err != nil {
```
ncdc, line 53:
```suggestion
	if err := fs.Parse(os.Args[1:]); err != nil {
```

fs.Args() is unset until after we call `fs.Parse()` - have to pass in command line args.

cmd/cluster-controller/main.go
```diff
@@ -2,112 +2,78 @@ package main
...
+		klog.Fatal(err)
+	}
+	if err := options.Validate(); err != nil {
+		klog.Fatal(err)
```
ncdc, line 57: The `klog.Fatal` call prints a full stack trace - might want to do something like this instead?

```go
fmt.Fprintln(os.Stderr, err)
os.Exit(1)
```

cmd/cluster-controller/main.go
```diff
@@ -2,112 +2,78 @@ package main
...
 const resyncPeriod = 10 * time.Hour
 
-var (
-	kubeconfigPath  = flag.String("kubeconfig", "", "Path to kubeconfig")
-	syncerImage     = flag.String("syncer_image", "", "Syncer image to install on clusters")
-	pullMode        = flag.Bool("pull_mode", true, "Deploy the syncer in registered physical clusters in POD, and have it sync resources from KCP")
-	pushMode        = flag.Bool("push_mode", false, "If true, run syncer for each cluster from inside cluster controller")
-	autoPublishAPIs = flag.Bool("auto_publish_apis", false, "If true, the APIs imported from physical clusters will be published automatically as CRDs")
-)
+func bindOptions(fs *pflag.FlagSet) *options {
```
ncdc, line 24: Thoughts on making this a method on options and having it create & return a flagset? (This is what I've seen elsewhere). I don't know that I have strong feelings one way or another - mostly just comparing w/other art I'm familiar with...

cmd/cluster-controller/main.go
stevekuznetsov: ew. another -1 to `klog`

cmd/cluster-controller/main.go
stevekuznetsov: Passing in the flagset allows for composition, so it's useful in the helper library for e.g. the `cluster` config. I just used the same approach here so it would be parallel & simple, but yeah I guess it doesn't matter one way or the other in `main`

stevekuznetsov: I think I might prefer it this way so that `bind` has one meaning across the codebase?

pkg/server/config.go
```diff
@@ -6,156 +6,63 @@ import (
...
+	fs.StringVar(&c.EtcdClientInfo.KeyFile, "etcd-keyfile", "", "TLS key file used to secure etcd communication.")
+	fs.StringVar(&c.EtcdClientInfo.CertFile, "etcd-certfile", "", "TLS certification file used to secure etcd communication.")
+	fs.StringVar(&c.EtcdClientInfo.TrustedCAFile, "etcd-cafile", "", "TLS Certificate Authority file used to secure etcd communication.")
+	fs.StringVar(&c.ProfilerAddress, "profiler-address", "", "[Address]:port to bind the profiler to.")
+	fs.StringVar(&c.ShardKubeconfigFile, "shard-kubeconfig-file", "", "Kubeconfig holding admin(!) credentials to peer kcp shards.")
+	fs.BoolVar(&c.EnableSharding, "enable-sharding", false, "Enable delegating to peer kcp shards.")
+	fs.StringVar(&c.RootDirectory, "root_directory", ".kcp", "Root directory.")
+	fs.StringVar(&c.EtcdPeerPort, "etcd_peer_port", "2380", "Port for etcd peer communication.")
+	fs.StringVar(&c.EtcdClientPort, "etcd_client_port", "2379", "Port for etcd client communication.")
+	fs.StringVar(&c.KubeConfigPath, "kubeconfig_path", "admin.kubeconfig", "Path to which the administrative kubeconfig should be written at startup.")
```
ncdc, line 63: We want to update the defaults to come from `c`, right?

pkg/server/server.go
```diff
@@ -87,6 +79,10 @@ type preShutdownHookEntry struct {
 
 // Run starts the KCP api-server. This function blocks until the api-server stops or an error.
 func (s *Server) Run(ctx context.Context) error {
+	if err := s.cfg.ClusterControllerOptions.Validate(); err != nil {
```
ncdc, line 82: Do you want to do this here, or only if we're installing the controller (L302)?

pkg/reconciler/cluster/startup.go
```diff
@@ -0,0 +1,152 @@
...
+	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	kcpexternalversions "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
+	"github.com/kcp-dev/kcp/pkg/reconciler/apiresource"
+)
+
+// DefaultOptions are the default options for the cluster controller.
+func DefaultOptions() *Options {
+	return &Options{
+		SyncerImage:     "",
+		PullMode:        false,
```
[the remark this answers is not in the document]
stevekuznetsov: Only in one of the two places, and I *really* don't like that default, since you need to turn off pull mode to use pull mode? I'd rather not.

pkg/server/config.go
stevekuznetsov: :facepalm:

pkg/reconciler/cluster/startup.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+		crdClient := apiextensionsv1client.NewForConfigOrDie(logicalClusterConfig).CustomResourceDefinitions()
+		if err := config.BootstrapCustomResourceDefinitions(ctx, crdClient, requiredCrds); err != nil {
+			return err
+		}
+	}
+
+	if c.PullMode && c.PushMode {
+		return errors.New("can't set --push_mode and --pull_mode")
+	}
```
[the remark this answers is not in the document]
stevekuznetsov: I was hoping to just do a minimal copy-paste of the existing code, to DRY it out, not to fix everything. Ideally we wouldn't have two flags for this anyway?

pkg/server/server.go
stevekuznetsov: I was preferring quick feedback for broken flags, but happy to move it down

stevekuznetsov:
@ncdc udpated for review comments

pkg/reconciler/cluster/startup.go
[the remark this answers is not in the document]
ncdc: Would it make more sense to switch this to a single `SyncMode` field and then we can default it to whatever we'd like, and not have to enable/disable the 2 bools in lock-step?

ncdc:
LGTM

ncdc:
LMK if you're ready for merge

stevekuznetsov:
@ncdc yeah, let's do it

ncdc (approved):

PR #251, opened 2021-11-29
add new flag: resourcesToSync to let sync/import crds when cluster-controller run standalone

vincent-pli:
When run `cluster controller` run standalone, it need the `resourcesToSync`

### DIFF
--- a/cmd/cluster-controller/main.go
+++ b/cmd/cluster-controller/main.go
@@ -2,11 +2,11 @@ package main
 
 import (
 	"context"
-	"flag"
 	"os"
 	"os/signal"
 	"time"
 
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
I do this in https://github.com/kcp-dev/kcp/pull/228 - do you have a strong preference for which lands? You can choose the CRDs today if you pass them as args.

ncdc:
@stevekuznetsov this is in now that #228 has merged, right? i.e. we can close this?

stevekuznetsov:
@vincent-pli I'm going to close this as the aforementioned PR landed and the flag is in! Thanks for this, though :)

PR #253, opened 2021-11-29
Support OIDC authentication flags

### LINKED ISSUE #190: Allow passing flags for OIDC
I've seen an attempt to use OIDC with KCP in https://www.youtube.com/watch?v=FD_kY3Ey2pI (incomplete). I'm thinking to create a demo based on that feature, so it would be nice if kcp could accept oidc/rbac related flags on startup without having to hack on code.

Is this something planned? I could also help.

imjasonh:
Fixes #190

### DIFF
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -21,10 +21,10 @@ import (
 
 	"github.com/spf13/pflag"
 
+	kubeoptions "k8s.io/kubernetes/pkg/kubeapiserver/options"
+
 	"github.com/kcp-dev/kcp/pkg/etcd"
 	"github.com/kcp-dev/kcp/pkg/reconciler/cluster"
-
-	kubeoptions "k8s.io/kubernetes/pkg/kubeapiserver/options"
 )
 
 // DefaultConfig is the default behavior of the KCP server.
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -43,8 +43,7 @@ func DefaultConfig() *Config {
 		ProfilerAddress:            "",
 		ShardKubeconfigFile:        "",
 		EnableSharding:             false,
-		// Currently only adding OIDC flags.
-		Authentication: kubeoptions.NewBuiltInAuthenticationOptions().WithOIDC(),
+		Authentication:             kubeoptions.NewBuiltInAuthenticationOptions().WithAll(),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/server/config.go
```diff
@@ -2,9 +2,13 @@ package server
 
 import (
 	"flag"
+	"log"
+	"strings"
 
 	"github.com/spf13/pflag"
 
+	kubeoptions "k8s.io/kubernetes/pkg/kubeapiserver/options"
```
stevekuznetsov, line 10: Isn't this going to be a disallowed import?

pkg/server/config.go
```diff
@@ -45,87 +50,107 @@ func DefaultConfig() *Config {
...
 	}
 }
 
 // ConfigFromFlags returns a default configuration with config values set from the provided flag set
 func ConfigFromFlags(flags *pflag.FlagSet) *Config {
 	cfg := DefaultConfig()
-	listen, err := flags.GetString("listen")
-	if err == nil {
+
+	if listen, err := flags.GetString("listen"); err == nil {
```
stevekuznetsov, line 61: This code gets removed in #228 so I wouldn't try *too* hard to refactor if you have anything else planned ;)

imjasonh: Excellent. I'll wait for that then. This was all pretty annoying because this seemed to want to be set up using [`AddFlags`](https://github.com/kubernetes/kubernetes/blob/63e7ee43bb5394abd6dd5010be47ebcb22ec0ef8/pkg/kubeapiserver/options/authentication.go#L245) but the way we added and parsed flags as two separate steps made that harder. I bet your change will make this all easier 👍

pkg/server/config.go
ncdc: It's fine for now, given that all of kcp is in the prototyping phase. But if we want to do this long-term/"for real", we need to move anything we need in k/k/pkg into k/k/staging & publish it as a standalone repo (or move it into a preexisting staging dir/repo).

stevekuznetsov:
Given Andy's comment, this LGTM - after a rebase we can merge

stevekuznetsov:
@imjasonh based on the e2e, the `kcp` binary does not start

imjasonh:
> @imjasonh based on the e2e, the `kcp` binary does not start

This was caused by nil clientcert config because I was only populating `WithOIDC`. I've temporarily switched to `WithAll` and parsing alllllll the authentication flags, but I'm not sure we want them all. We can probably just only define/parse the OIDC flags manually, if we feel strongly about it.

stevekuznetsov (approved):

davidfestal:
Does anyone know how to trigger the run of the missing checks that seem to prevent the merging ?

stevekuznetsov:
Wow, "re-run" does not run the updated ones .... @imjasonh mind re-pushing?

PR #258, opened 2021-11-29
ci: verify our boilerplate

stevekuznetsov:
This is a non-standard invocation as we don't want to pull in Bazel
right now.


Makefile
```diff
@@ -1,3 +1,17 @@
+# Copyright 2021 The Kubernetes Authors.
```
imjasonh, line 1: Ianal, is this what we want here? `The KCP Authors`?

imjasonh:
Another option that Knative uses: https://github.com/mattmoor/boilerplate-check

stevekuznetsov:
I think this needs? https://github.com/kubernetes/repo-infra/pull/246

Makefile
stevekuznetsov: /shrug

Open to ideas, no clue

smarterclayton:
generally rh legal allows license file at the top for apache projects and no boilerplate

ncdc:
That would certainly make things easier (less process). Does that work for you @stevekuznetsov (I know I was the one yesterday asking about missing boilerplate in one of your PRs, but ... 😄 )

smarterclayton:
but!  we want to remain generally aligned to kube specifically and cncf convention generally - to continue with existing codebase as a kube sub project will require us to establish ownership and some of what we’re doing might have to get rewritten.  i think it’s likely anything minimal api server “like” needs to be very much associated with the kube fork (at the point we suggest merges based on that is when ownership is transferred to kube and cncf)

imjasonh (approved):

PR #260, opened 2021-11-30
ci: turn on more linters

stevekuznetsov:
Looks like we catch instances of this in review, so there's nothing to
fix up in the codebase.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -50,7 +50,7 @@ jobs:
     - uses: reviewdog/action-misspell@v1
       with:
         exclude: ./.github/workflows/*
-        ignore: creater
+        ignore: creater,importas
 
   build-and-test:
     name: Build and Test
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
     - uses: reviewdog/action-misspell@v1
       with:
         exclude: ./.github/workflows/*
-        ignore: creater
+        ignore: creater,importas
```
imjasonh, line 53: We can also just replace `misspell`: https://golangci-lint.run/usage/linters/#misspell -- does golangci-lint try to lint its own config? 🤔

imjasonh (approved):

.github/workflows/ci.yaml
stevekuznetsov: I had that thought ...

PR #267, opened 2021-12-02
Adds missing flag to demo files

jmprusi:
Due to the latest changes on how flags are parsed, demos were broken.

Demos just requiring only "deployments.app" to be synced, were working out of pure luck as it matched the set defaults for the unset flag.

Signed-off-by: Joaquim Moreno <joaquim@redhat.com>


davidfestal (approved):

davidfestal:
Thanks !

Issue #268, opened 2021-12-02
Possible flake in TestWorkspaceController


ncdc:
Seen on main after merging #212 in https://github.com/kcp-dev/kcp/actions/runs/1531127337

```
--- FAIL: TestWorkspaceController (0.00s)
    --- FAIL: TestWorkspaceController/delete_all_shards,_expect_workspace_to_be_unschedulable (15.28s)
        t.go:74: Saving artifacts to /tmp/TestWorkspaceController_delete_all_shards,_expect_workspace_to_be_unschedulable90468456/001/TestWorkspaceController_delete_all_shards,_expect_workspace_to_be_unschedulable/38467.
        t.go:74: running: kcp start --root_directory=/tmp/TestWorkspaceController_delete_all_shards,_expect_workspace_to_be_unschedulable90468456/001/TestWorkspaceController_delete_all_shards,_expect_workspace_to_be_unschedulable/38467 --listen=:38467 --etcd_client_port=41263 --etcd_peer_port=36607 --kubeconfig_path=admin.kubeconfig --install_workspace_controller
        t.go:74: success contacting /livez
        t.go:74: success contacting /readyz
        test.go:114: did not see workspace updated: expected a scheduled workspace, got status.conditions: []v1alpha1.WorkspaceCondition{v1alpha1.WorkspaceCondition{Type:"WorkspaceScheduled", Status:"False", LastProbeTime:v1.Time{Time:time.Time{wall:0x0, ext:63774054437, loc:(*time.Location)(0x2bdfda0)}}, LastTransitionTime:v1.Time{Time:time.Time{wall:0x0, ext:63774054437, loc:(*time.Location)(0x2bdfda0)}}, Reason:"Unschedulable", Message:"No shards are available to schedule Workspaces to."}}
        t.go:69: cleanup: ending kcp server
FAIL
coverage: [no statements]
FAIL	github.com/kcp-dev/kcp/test/e2e/reconciler/workspace	41.286s
FAIL
Error: Process completed with exit code 1.
```

PR #269, opened 2021-12-02
reconciler: workspace: deflake movement

stevekuznetsov:
We never set target to be an empty string, so this logic was bugged and
coudl race.

fixes https://github.com/kcp-dev/kcp/issues/268

### DIFF
--- a/pkg/reconciler/workspace/controller.go
+++ b/pkg/reconciler/workspace/controller.go
@@ -312,7 +312,7 @@ func (c *Controller) reconcile(ctx context.Context, workspace *tenancyv1alpha1.W
 			klog.Infof("scheduling workspace %q to %q", workspace.Name, target)
 		}
 	}
-	if workspace.Status.Location.Current != workspace.Status.Location.Target {
+	if workspace.Status.Location.Target != "" && workspace.Status.Location.Current != workspace.Status.Location.Target {
 		klog.Infof("moving workspace %q from %q to %q", workspace.Name, workspace.Status.Location.Current, workspace.Status.Location.Target)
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
Hm, now to reproduce *that* locally ... ???

pkg/reconciler/workspace/controller.go
```diff
@@ -289,18 +289,17 @@ func (c *Controller) reconcile(ctx context.Context, workspace *tenancyv1alpha1.W
...
 		workspace.Status.BaseURL = fmt.Sprintf("%s.workspaces.kcp.dev", workspace.Name)
 	}
-	var shouldReSchedule bool
 	if currentShard := workspace.Status.Location.Current; currentShard != "" {
 		// make sure current shard still exists
 		_, err := c.workspaceShardLister.Get(clusters.ToClusterAwareKey(workspace.ClusterName, currentShard))
 		if errors.IsNotFound(err) {
 			klog.Infof("de-scheduling workspace %q from nonexistent shard %q", workspace.Name, currentShard)
-			shouldReSchedule = true
+			workspace.Status.Location.Current = ""
```
ncdc, line 297: If for some reason we've gone from 1 shard to 0 shards in the system, will doing this result in the workspace getting patched and us losing the current (i.e. "old") placement? L308-312 would not be exectued because len=0 in this scenario.

stevekuznetsov: Yes, and as written this is intended behavior and has a test to that effect.

ncdc (approved):

PR #273, opened 2021-12-02
Roll in CAPI Conditions, use them

stevekuznetsov:
util: conditions: seed from cluster-api

Pulled from version at https://github.com/kubernetes-sigs/cluster-api/commit/fc52fe3d5cfe05d436a7d2f0ba3cf8a1399aad59

---

util: conditions: update boilerplate

---

util: conditions: import our copied types

---

util: conditions: fixup copied deepcopies

---

util: conditions: excise controller-runtime

---

util: conditions: make test self-contained

---

util: conditions: fixup imports

---

util: conditions: remove unstructured helpers

These require that a type we have actually implements the Getter and
Setter, can be added in later.

---

util: conditions: use stdlib errors, update vendors

---

api: workspaces: use conditions library

---

/assign @ncdc


stevekuznetsov:
This works as-is but I still need to bring back the `Unstructured` stuff, if we need it

stevekuznetsov:
In this review I am mostly looking for:

 - are the edits post-copy valid?
 - should we be doing something more intense to leverage the patch generating capabilities of the library? what if we want to update more than just conditions in status in one Patch call?

pkg/apis/conditions/register.go
```diff
@@ -0,0 +1,21 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package conditions
+
+const (
+	GroupName = "conditions.kcp.dev"
```
ncdc, line 20: We likely won't need a proper group name unless we're planning on registering these as standalone types (i.e. as CRDs).

test/e2e/reconciler/workspace/controller_test.go
```diff
@@ -261,19 +261,23 @@ func exactMatcher(expected *tenancyv1alpha1.Workspace) matcher {
...
+func unschedulable(workspace *tenancyv1alpha1.Workspace) bool {
+	return utilconditions.IsFalse(workspace, tenancyv1alpha1.WorkspaceScheduled) && utilconditions.GetReason(workspace, tenancyv1alpha1.WorkspaceScheduled) == tenancyv1alpha1.WorkspaceReasonUnschedulable
+}
+
 func unschedulableMatcher() matcher {
 	return func(object *tenancyv1alpha1.Workspace) error {
-		if !conditions.IsWorkspaceUnschedulable(object) {
-			return fmt.Errorf("expected an unschedulable workspace, got status.conditions: %#v", object.Status.Conditions)
+		if !unschedulable(object) {
+			return fmt.Errorf("expected an unschedulable workspace, got status.utilconditions: %#v", object.Status.Conditions)
```
ncdc, line 271: Over-aggressive find/replace 😄 here & below

pkg/util/conditions/getter.go
```diff
@@ -0,0 +1,260 @@
...
+*/
+
+package conditions
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	clusterv1 "github.com/kcp-dev/kcp/pkg/apis/conditions/v1alpha1"
```
ncdc, line 24: Global comment: replace all `clusterv1` import aliases with something like `conditionsapi`

pkg/util/conditions/getter.go
```diff
@@ -0,0 +1,260 @@
...
+)
+
+// Getter interface defines methods that a Cluster API object should implement in order to
```
ncdc, line 27: Global comment: replace any code comment references to Cluster API appropriately

ncdc:
> should we be doing something more intense to leverage the patch generating capabilities of the library? what if we want to update more than just conditions in status in one Patch call?

I don't _think_ what you've imported interacts with the client in any way to do a patch. Cluster API has an extensive patch helper as well (https://github.com/kubernetes-sigs/cluster-api/blob/faffa2c609dc399a6c06596a64db179709f88f8c/util/patch/patch.go) that completely manages patching an object's spec and status (including conditions), but it would be much more work to use that (as it uses the controller-runtime Client).

Correctly generating patches is probably outside the scope of this PR - would you agree?

pkg/apis/conditions/v1alpha1/constants.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The Kubernetes Authors.
+Copyright 2021 The KCP Authors.
```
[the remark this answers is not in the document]
stevekuznetsov: Was wondering about that - boilerplate verification can't handle such a thing. What to do?

test/e2e/reconciler/workspace/controller_test.go
stevekuznetsov: thanks, goland :facepalm:

stevekuznetsov:
@ncdc this code brought in e.g.

```go
func NewPatch(before Getter, after Getter) Patch {}
```

but as long as we make *some* forward progress to make setting, checking and updating conditions easy + simple for controller authors in `kcp`, I think the PR is a success without having to use every possible part of the library.

pkg/apis/conditions/v1alpha1/constants.go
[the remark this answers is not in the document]
ncdc: The one thing that immediately comes to mind is to put this in a `third_party` directory and `--skip` it when calling `verify_boilerplate.py`

ncdc:
@smarterclayton do you have advice for how to retain the original copyright notice in the generated code (e.g. deepcopy)? Or is that not needed (perhaps because it's generated code)?

stevekuznetsov:
I've just added a separate generator step with a separate boilerplate template for that one

pkg/apis/conditions/v1alpha1/constants.go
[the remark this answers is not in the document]
stevekuznetsov: Did that!

stevekuznetsov:
I can squash this one down - the intermediate commits are only really useful for review.

third_party/conditions/apis/conditions/v1alpha1/constants.go
```diff
@@ -0,0 +1,253 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+// ANCHOR: CommonConditions
```
ncdc, line 19: Do we use anything from this file?

third_party/conditions/util/conditions/getter.go
```diff
@@ -0,0 +1,260 @@
...
+	conditionsapi "github.com/kcp-dev/kcp/third_party/conditions/apis/conditions/v1alpha1"
+)
+
+// Getter interface defines methods that an object should implement in order to
+// use the conditions package for getting conditions.
+type Getter interface {
+	metav1.Object
+	runtime.Object
+
+	// GetConditions returns the list of conditions for a cluster API object.
```
ncdc, line 33:
```suggestion
	// GetConditions returns the list of conditions for the object.
```

WDYT?

third_party/conditions/apis/conditions/v1alpha1/constants.go
stevekuznetsov: Some of the helpers in getter.go do. I wanted to do as little surgery as possible so that a future common library would be a drop-in replacement.

stevekuznetsov:
@ncdc updated for the one lowercase CAPI reference I missed !

ncdc:
From Clayton:
> generally would not update the headers and the copyright would fall through to the kubernetes authors

ncdc (approved):

Issue #286, opened 2021-12-03
Workspace E2E: Flakes When Workspace Remains Unschedulable


stevekuznetsov:
Example logs:

```
--- FAIL: TestWorkspaceController (0.00s)
    --- FAIL: TestWorkspaceController/delete_a_shard_that_a_workspace_is_scheduled_to,_expect_it_to_move_to_another_shard (14.11s)
        t.go:74: Saving artifacts to /tmp/kcp-e2e/TestWorkspaceController_delete_a_shard_that_a_workspace_is_scheduled_to,_expect_it_to_move_to_another_shard/kcp/main.
        t.go:69: Starting kcp servers...
        t.go:74: running: kcp start --root_directory=/tmp/kcp-e2e/TestWorkspaceController_delete_a_shard_that_a_workspace_is_scheduled_to,_expect_it_to_move_to_another_shard/kcp/main --listen=:34585 --etcd_client_port=46229 --etcd_peer_port=43351 --kubeconfig_path=admin.kubeconfig --install_workspace_controller
        t.go:74: success contacting /livez
        t.go:74: success contacting /readyz
        t.go:74: Started kcp servers after 12.063357612s
        test.go:126: did not see workspace updated: expected a scheduled workspace, got status.conditions: []v1alpha1.WorkspaceCondition{v1alpha1.WorkspaceCondition{Type:"WorkspaceScheduled", Status:"False", LastProbeTime:v1.Time{Time:time.Time{wall:0x0, ext:63774093840, loc:(*time.Location)(0x2be1de0)}}, LastTransitionTime:v1.Time{Time:time.Time{wall:0x0, ext:63774093840, loc:(*time.Location)(0x2be1de0)}}, Reason:"Unschedulable", Message:"No shards are available to schedule Workspaces to."}}
        t.go:69: cleanup: ending kcp server
```

Need to add artifacts to the run so we can look at logs for this one.

stevekuznetsov:
This hasn't been seen in a month, so likely fixed.

PR #287, opened 2021-12-03
ci: upload artifacts from e2e

stevekuznetsov:

Useful for #286

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -78,7 +78,7 @@ jobs:
     - name: Check codegen
       run: make verify-codegen
     - run: make build
-    - run: PATH="${PATH}:$(pwd)/bin/" go test -race -coverprofile=coverage.txt -covermode=atomic ./...
+    - run: ARTIFACT_DIR=/tmp/e2e PATH="${PATH}:$(pwd)/bin/" go test -race -coverprofile=coverage.txt -covermode=atomic ./...
     # Because verify-codegen updates imports under the covers, and it affects
     # everything in the working directory, we have to check out kubernetes
     # after verifying codegen; otherwise, kubernetes code will get modified,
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -92,4 +92,8 @@ jobs:
     - run: make WHAT=test/e2e/e2e.test
       working-directory: ./kubernetes
     - run: WORKDIR=/tmp/e2e hack/run-sharded-kcp.sh
-    - run: WORKDIR=/tmp/cluster hack/run-cluster-kcp.sh
\ No newline at end of file
+    - run: WORKDIR=/tmp/cluster hack/run-cluster-kcp.sh
+    - uses: actions/upload-artifact@v2
+      with:
+        name: e2e-artifacts
+        path: /tmp/e2e/
```
sttts, line 99: what will land there? binaries? logs?

stevekuznetsov: Logs and other test output goes here. Do we have a strong preference for "what" ?

stevekuznetsov: I guess right now it also includes the a couple directories which we likely don't need:

```
$ tree /tmp/e2e/
/tmp/e2e/
├── TestWorkspaceController_add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled
│   └── kcp
│       └── main
│           ├── admin.kubeconfig
│           ├── apiserver.crt
│           ├── apiserver.key
│           ├── kcp.log
│           ├── member
│           │   ├── snap
│           │   │   └── db
│           │   └── wal
│           │       ├── 0000000000000000-0000000000000000.wal
│           │       └── 0.tmp
│           └── secrets
│               ├── ca
│               │   ├── cert.pem
│               │   └── key.pem
│               ├── client
│               │   ├── cert.pem
│               │   └── key.pem
│               └── peer
│                   ├── cert.pem
│                   └── key.pem
```

stevekuznetsov: Right now I think the only real useful item is going to be the `*.log` items - I can filter down in a follow-up to direct test-driven data to a specific subset of this directory which we upload, while making sure that on a laptop the UX for looking at test artifacts like kubeconfig while a test runs is good

stevekuznetsov:
Updated to be more intelligent in https://github.com/kcp-dev/kcp/pull/289

PR #289, opened 2021-12-03
Improvements to E2E framework, artifact capture

stevekuznetsov:
e2e: framework: allow naming kcp servers

For better UX when poking around inside of the artifact directory for
tests with more than one kcp server, we allow naming them.

---

makefile: add a helper for running e2e tests

Since the e2e runs kcp servers via the binary, we need to make sure that
a) the built binaries are in $PATH and b) we re-build when running.

---

e2e: workspace: scope testCase variable

Without this scoping step, the reference leaks out to each test
goroutine.

---

test: e2e: improve artifact capture, run more tests

We now distinguish between the data directories for kcp servers we start
and the artifacts that a test produces, as we only want to capture the
latter in CI for review when a test fails. This commit also adds the
artifact publishing to CI.

---

Supersedes https://github.com/kcp-dev/kcp/pull/287
/assign @sttts

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -78,7 +78,7 @@ jobs:
     - name: Check codegen
       run: make verify-codegen
     - run: make build
-    - run: PATH="${PATH}:$(pwd)/bin/" go test -race -coverprofile=coverage.txt -covermode=atomic ./...
+    - run: ARTIFACT_DIR=/tmp/e2e PATH="${PATH}:$(pwd)/bin/" go test -race -coverprofile=coverage.txt -covermode=atomic ./...
     # Because verify-codegen updates imports under the covers, and it affects
     # everything in the working directory, we have to check out kubernetes
     # after verifying codegen; otherwise, kubernetes code will get modified,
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -92,4 +92,8 @@ jobs:
     - run: make WHAT=test/e2e/e2e.test
       working-directory: ./kubernetes
     - run: WORKDIR=/tmp/e2e hack/run-sharded-kcp.sh
-    - run: WORKDIR=/tmp/cluster hack/run-cluster-kcp.sh
\ No newline at end of file
+    - run: WORKDIR=/tmp/cluster hack/run-cluster-kcp.sh
+    - uses: actions/upload-artifact@v2
```
imjasonh, line 96: If you add:

```
if: ${{ always() }}
```

...this will always upload the artifact even if previous steps failed.

stevekuznetsov: Good catch!

stevekuznetsov:
@imjasonh updated

imjasonh (approved):

PR #302, opened 2021-12-07
Workspace API inheritance

ncdc:
Depends on https://github.com/kcp-dev/kubernetes/pull/26 (will update go.mod/go.sum after that's reviewed & merged).

Add new field workspace.spec.inheritFrom. Setting this lets a target workspace transparently gain the APIs provided by a CRD in the inheritFrom workspace. Storage for any CRs in the target workspace stays in the target workspace's partition of the etcd key space.

Given source and target workspaces as described above:

- A user listing CRDs in the target workspace will **not** see CRDs from the source workspace; they will only see CRDs that live directly in the target workspace
- A user doing API discovery in the target workspace will see APIs from the generic control plane (rbac, certs, core v1, etc.), the source workspace (from its CRDs), and the target workspace (from its CRDs)
- OpenAPI does **not** currently display inherited APIs for a workspace (this is a lot harder)

Additionally, any CRDs that are in core v1 (/api/v1) continue to be aggregated into discovery for /api/v1 alongside secrets, configmaps, and so on. The code to do this moved from kubernetes to kcp.


pkg/server/server.go
```diff
@@ -222,15 +246,161 @@ func (s *Server) Run(ctx context.Context) error {
...
+			chain.ProcessFilter(req, res)
+			return
+		}
+
+		// Special handling of the core ("") group
+
+		if requestInfo.IsResourceRequest {
+			// This is a CRUD request for somethig like pods. Try to see if there is a CRD for the resource. If so, let the CRD
+			// server handle it.
+			crdName := requestInfo.Resource + ".core"
```
ncdc, line 334: Not sure if we want to force this, or instead list all and match on name?

pkg/server/server.go
```diff
@@ -222,15 +246,161 @@ func (s *Server) Run(ctx context.Context) error {
...
+			_, err := serverChain.CustomResourceDefinitions.Informers.Apiextensions().V1().CustomResourceDefinitions().Lister().GetWithContext(ctx, crdName)
+			if err == nil {
```
ncdc, line 336: No error handling here (yet)... does it just make sense to log and still return a response?

pkg/server/server.go
```diff
@@ -222,15 +246,161 @@ func (s *Server) Run(ctx context.Context) error {
...
+			const passthroughHeader = "X-Kcp-Api-V1-Discovery-Passthrough"
+			if _, passthrough := req.Request.Header[passthroughHeader]; passthrough {
+				chain.ProcessFilter(req, res)
+				return
+			}
+
+			// Get all the CRDs (in the context's logical cluster) to see if any of them are in v1
+			crds, err := serverChain.CustomResourceDefinitions.Informers.Apiextensions().V1().CustomResourceDefinitions().Lister().ListWithContext(ctx, labels.Everything())
+			if err != nil {
+				// FIXME
```
ncdc, line 354: No error handling here (yet)... does it just make sense to log and still return a response (the built in /api/v1 discovery)?

pkg/server/server.go
```diff
@@ -222,15 +246,161 @@ func (s *Server) Run(ctx context.Context) error {
...
+
+			// v1 CRDs present - need to clone the request, add our passthrough header, and get /api/v1 discovery from
+			// the GenericControlPlane's server.
+			cr := utilnet.CloneRequest(req.Request)
+			cr.Header.Add(passthroughHeader, "1")
+
+			writer := newInMemoryResponseWriter()
+			serverChain.GenericControlPlane.GenericAPIServer.Handler.GoRestfulContainer.ServeHTTP(writer, cr)
+			if writer.respCode != http.StatusOK {
+				// FIXME
```
ncdc, line 371: No error handling here (yet)... this means the attempt to get /api/v1 for built-in types failed. Probably just need to echo the response back?

pkg/server/server.go
```diff
@@ -222,15 +246,161 @@ func (s *Server) Run(ctx context.Context) error {
...
+			}
+
+			// Decode the response
+			into := &metav1.APIResourceList{}
+			obj, _, err := genericcontrolplane.Codecs.UniversalDeserializer().Decode(writer.data, nil, into)
+			if err != nil {
+				// FIXME
```
ncdc, line 378: Not sure how to handle this error & would love suggestions

pkg/server/server.go
```diff
@@ -222,15 +246,161 @@ func (s *Server) Run(ctx context.Context) error {
...
+			}
+			v1ResourceList, ok := obj.(*metav1.APIResourceList)
+			if !ok {
+				// FIXME
```
ncdc, line 382: Not sure how to handle this error & would love suggestions

pkg/server/server.go
```diff
@@ -476,3 +646,264 @@ func adaptContext(parent genericapiserver.PostStartHookContext) context.Context
 	}(parent.StopCh)
 	return ctx
 }
+
+type kcpCRDLister struct {
```
ncdc, line 650: I should probably comment why this exists/what it does.

pkg/server/server.go
```diff
@@ -476,3 +646,264 @@ func adaptContext(parent genericapiserver.PostStartHookContext) context.Context
...
+			if !equality.Semantic.DeepEqual(crd.Spec, aCRD.Spec) {
+				return crd, false
+			}
+		}
+	}
+
+	return crd, true
+}
+
+type kcpAPIExtensionsSharedInformerFactory struct {
```
ncdc, line 817: I should probably comment on this chain of structs - it exists solely to be able to override the CRD lister you get from the CRD informer with our own.

pkg/server/server.go
```diff
@@ -476,3 +646,264 @@ func adaptContext(parent genericapiserver.PostStartHookContext) context.Context
...
+func (i *kcpAPIExtensionsApiextensionsV1CustomResourceDefinitionInformer) Lister() apiextensionslisters.CustomResourceDefinitionLister {
+	originalLister := i.CustomResourceDefinitionInformer.Lister()
+	l := &kcpCRDLister{
+		crdLister:       originalLister,
+		workspaceLister: i.workspaceLister,
+	}
+	return l
+}
+
+// COPIED FROM kube-aggregator
```
ncdc, line 870: @smarterclayton what's the best way to do attribution here or what should I do?

pkg/server/server.go
```diff
@@ -29,15 +29,32 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
+	"github.com/emicklei/go-restful"
+
+	// Enable Profiling
```
stevekuznetsov, line 38: This doesn't seem to be in the right spot?

pkg/server/server.go
```diff
@@ -222,15 +242,90 @@ func (s *Server) Run(ctx context.Context) error {
 		KeyFile:       s.cfg.EtcdClientInfo.KeyFile,
 		TrustedCAFile: s.cfg.EtcdClientInfo.TrustedCAFile,
 	}
+
+	// TODO(ncdc): I thought I was going to need this, but it turns out this breaks the CRD controllers because they
+	// try to issue Update() calls using the * client, which ends up with the cluster name being set to the default
```
stevekuznetsov, line 251: Update() with `*` is forbidden ... I have that code in the shardinator but likely it needs to be in kcp proper?

pkg/server/server.go
```diff
@@ -222,15 +242,90 @@ func (s *Server) Run(ctx context.Context) error {
...
+	}
+	serverOptions.SecureServing.LoopbackCertPEM = loopbackClientCert
+	serverOptions.SecureServing.LoopbackKeyPEM = loopbackClientCertKey
+	serverOptions.SecureServing.LoopbackToken = uuid.New().String()
+	loopbackClientConfig, err := serverOptions.SecureServing.NewLoopbackClientConfig(serverOptions.SecureServing.LoopbackToken, serverOptions.SecureServing.LoopbackCertPEM)
+	if err != nil {
+		return err
+	}
+
+	// FIXME: switch to a single set of shared informers
```
stevekuznetsov, line 275: We already have one of these for the workspace controller - can we at least use the same factory for cross-cluster bits?

pkg/server/server.go
```diff
@@ -476,3 +571,223 @@ func adaptContext(parent genericapiserver.PostStartHookContext) context.Context
...
+func (c *kcpCRDLister) ListWithContext(ctx context.Context, selector labels.Selector) ([]*apiextensionsv1.CustomResourceDefinition, error) {
+	cluster, err := genericapirequest.ValidClusterFrom(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Check for API inheritance
+	inheriting := false
+	inheritFrom := ""
+	const adminClusterName = "admin" // FIXME
```
stevekuznetsov, line 670: You could look to see which lcluster the workspace CRD exists in

stevekuznetsov (commented):
Could use a set of e2e to show off the behavior

pkg/server/server.go
ncdc: Thank you openshift-goimports 🤦

pkg/server/server.go
ncdc: I plan on unifying factories in a follow-up PR 👍

pkg/server/server.go
ncdc: Couldn't it exist in multiple? e.g. if we have multiple "org workspaces" == "lclusters that have the workspace apis"?

pkg/server/server.go
sttts: yes, godoc please :)

pkg/server/server.go
```diff
@@ -222,15 +242,90 @@ func (s *Server) Run(ctx context.Context) error {
...
+
+	workspaceLister := kcpSharedInformerFactory.Tenancy().V1alpha1().Workspaces().Lister()
+
+	serverOptions.APIExtensionsNewSharedInformerFactoryFunc = func(client apiextensionsclient.Interface, resyncPeriod time.Duration) crdexternalversions.SharedInformerFactory {
+		f := apiextensionsinformers.NewSharedInformerFactory(client, resyncPeriod)
+		return &kcpAPIExtensionsSharedInformerFactory{
+			SharedInformerFactory: f,
+			workspaceLister:       workspaceLister,
+		}
+	}
```
sttts, line 296: how do you make sure now that the apiextensions controllers don't follow the inheritence implemented here in that informer?

pkg/server/server.go
```diff
@@ -310,20 +314,91 @@ func (s *Server) Run(ctx context.Context) error {
...
 				serverChain.CustomResourceDefinitions.GenericAPIServer.Handler.NonGoRestfulMux.ServeHTTP(res.ResponseWriter, req.Request)
 				return
 			}
+		} else if req.Request.URL.Path == "/api/v1" || req.Request.URL.Path == "/api/v1/" {
+			// This is a discovery request. We may need to combine discovery from the GenericControlPlane (which has built-in v1 types)
+			// and CRDs, if there are any v1 CRDs.
+
+			// If we are retrieving the GenericControlPlane's v1 APIResources, let that request through.
+			const passthroughHeader = "X-Kcp-Api-V1-Discovery-Passthrough"
+			if _, passthrough := req.Request.Header[passthroughHeader]; passthrough {
```
sttts, line 344: who sets passthroughHeader ? could we share the constant with the code that sets this header?

sttts: just 20 lines below I guess. Am still not sure why we get a recursion here (and have to protect against with this header).

pkg/server/server.go
sttts: would leave it as simple as this until we have a better idea about workspace location and identity

pkg/server/server.go
ncdc: Right now via 2 different ways of "cheating":

1. Instead of specifying our own new client func (which would create a `*` client), I use what we had previously - the client is "enabled for multi-cluster", which means it uses our custom multi-cluster round tripper, which attempts to set the cluster on the request based on the ObjectMeta from the request body in a write operation.

2. I didn't undo a previously-written hack in the NamingConditionController that makes it lcluster-aware (it excludes CRDs from other lclusters when listing & looping).

We probably need a proper plan here...

pkg/server/server.go
ncdc: It was overflowing stack w/o this. Let me debug & figure out what's going on.

sttts: am curious to see a backtrace of the recursion.

pkg/server/server.go
sttts: Reconciling CRDs in all lclusters is fine and correct. Just the inheritence should not be applied here. I understand from (1) that we don't do that :+1:

pkg/server/server.go
sttts: simple enough code, no? Not complex enough to fall under attribution requirements ;-)

pkg/server/server.go
sttts: can this even fail?

pkg/server/server.go
sttts: that's a "should never happen" case, right? Would fail hard here, i.e. pass up that error to the real request.

pkg/server/server.go
sttts: as above, fail hard

pkg/server/server.go
sttts: wasn't it possible to pass nil as into?

pkg/server/server.go
ncdc: Agreed

pkg/server/server.go
ncdc: Agreed

pkg/server/server.go
ncdc: I tried that originally but was getting some weird behavior. Will try again & see what happens.

pkg/server/server.go
ncdc: Agreed, just checking

pkg/server/server.go
ncdc: The recursion happens because this entire function is a filter on the GoRestfulContainer, so this same filter function is invoked below when we call `serverChain.GenericControlPlane.GenericAPIServer.Handler.GoRestfulContainer.ServeHTTP`.

ncdc: And this required to be a filter function because the initial /api/v1 request is ultimately handled by:

https://github.com/kubernetes/kubernetes/blob/c16b2afc1d3c32462f068ea08cdc4791bd97b947/staging/src/k8s.io/apiserver/pkg/server/handler.go#L136

test/e2e/api_inheritance/api_inheritance_test.go
```diff
@@ -0,0 +1,271 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/kcp-dev/kcp/config"
+	"github.com/kcp-dev/kcp/pkg/apis/tenancy"
+	tenancyapi "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
+	clientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	"github.com/kcp-dev/kcp/test/e2e/framework"
+)
+
+func TestFoo(t *testing.T) {
```
stevekuznetsov, line 36: nit: needs a test name :)

test/e2e/api_inheritance/api_inheritance_test.go
```diff
@@ -0,0 +1,271 @@
...
+		if err != nil {
+			t.Errorf("error retrieving source workspace tenancy API discovery: %v", err)
+			return
+		}
+		if !resourceExists(resources, "workspaces") {
+			t.Errorf("source workspace discovery is missing workspaces resource")
+			return
+		}
+
+		// This workspace will be created in the source workspace. This is to ensure that it doesn't
```
stevekuznetsov, line 137: super-nit: might be nice to use something other than Workspace as the type you're inheriting just to keep mind-melting levels down in this :)

stevekuznetsov (commented):
Seems pretty reasonable to me

test/e2e/api_inheritance/api_inheritance_test.go
ncdc: Switched to clusters

test/e2e/api_inheritance/api_inheritance_test.go
ncdc: Fixed

stevekuznetsov (commented):
Now that e2e artifacts are in, it would be nice to see all of the objects that your test creates in there. Other than that and the go.mod changes, what do you think we should expect from this PR before we merge it in?

ncdc:
I can't think of anything else

stevekuznetsov:
k8s compile error in the version you pull in:

```
 # k8s.io/kubernetes/pkg/controller/replication
pkg/controller/replication/conversion.go:65:25: cannot use conversionLister{...} (type conversionLister) as type "k8s.io/kubernetes/vendor/k8s.io/client-go/listers/apps/v1".ReplicaSetLister in return argument:
	conversionLister does not implement "k8s.io/kubernetes/vendor/k8s.io/client-go/listers/apps/v1".ReplicaSetLister (missing ListWithContext method)
```

ncdc:
Bumped kube again w/the new compilation fixes. Fingers crossed

PR #305, opened 2021-12-08
Complete bootstrap for CRDs that already exist

kylape:
When a KCP is restarted without wiping out the storage beforehand, bootstrapping CRDs will hang because the watch never gets any events since the CRDs already exist.

This will get the CRDs if they already exist and complete bootstrapping if they are already established.

### DIFF
--- a/config/bootstrap.go
+++ b/config/bootstrap.go
@@ -102,7 +102,7 @@ func BootstrapCustomResourceDefinitionFromFS(ctx context.Context, client apiexte
 				return fmt.Errorf("Error creating CRD %s: %w", gk.String(), err)
 			}
 		} else {
-			return fmt.Errorf("Error fetching CRD 1 %s: %w", gk.String(), err)
+			return fmt.Errorf("Error fetching CRD %s: %w", gk.String(), err)
 		}
 	} else {
 		rawCrd.ResourceVersion = crdResource.ResourceVersion
@@ -118,7 +118,7 @@ func BootstrapCustomResourceDefinitionFromFS(ctx context.Context, client apiexte
 			if apierrors.IsNotFound(err) {
 				return false, fmt.Errorf("CRD %s was deleted before being established", gk.String())
 			}
-			return false, fmt.Errorf("Error fetching CRD 2 %s: %w", gk.String(), err)
+			return false, fmt.Errorf("Error fetching CRD %s: %w", gk.String(), err)
 		}
 
 		return crdhelpers.IsCRDConditionTrue(crd, apiextensionsv1.Established), nil
--- a/config/bootstrap.go
+++ b/config/bootstrap.go
@@ -94,36 +94,35 @@ func BootstrapCustomResourceDefinitionFromFS(ctx context.Context, client apiexte
 		return fmt.Errorf("decoded CRD %s into incorrect type, got %T, wanted %T", gk.String(), rawCrd, &apiextensionsv1.CustomResourceDefinition{})
 	}
 
-	crd, err := client.Get(ctx, rawCrd.Name, metav1.GetOptions{})
-	if err != nil && !apierrors.IsNotFound(err) {
-		return fmt.Errorf("Error fetching CRD %s: %w", gk.String(), err)
-	}
-
-	if crd != nil && crdhelpers.IsCRDConditionTrue(crd, apiextensionsv1.Established) {
-		return nil
-	}
-
-	crd, err = client.Create(ctx, rawCrd, metav1.CreateOptions{})
+	crdResource, err := client.Get(ctx, rawCrd.Name, metav1.GetOptions{})
 	if err != nil {
-		return fmt.Errorf("could not create CRD %s: %w", gk.String(), err)
+		if apierrors.IsNotFound(err) {
+			crdResource, err = client.Create(ctx, rawCrd, metav1.CreateOptions{})
+			if err != nil {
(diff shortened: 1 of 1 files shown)


config/bootstrap.go
```diff
@@ -96,8 +96,21 @@ func BootstrapCustomResourceDefinitionFromFS(ctx context.Context, client apiexte
...
+				return fmt.Errorf("Error fetching CRD %s: %w", gk.String(), err)
+			}
+
+			if crdhelpers.IsCRDConditionTrue(crd, apiextensionsv1.Established) {
+				return nil
+			}
+
+		} else {
+			return fmt.Errorf("could not create CRD %s: %w", gk.String(), err)
+		}
```
sttts, line 113: this is not the right solution IMO. We should rather remove the watch below. It is not safe. There is no guarantee the watch is not interrupted. We should rather just use wait.Until. Then this case here is included.

config/bootstrap.go
```diff
@@ -95,39 +94,36 @@ func BootstrapCustomResourceDefinitionFromFS(ctx context.Context, client apiexte
...
-				if crdhelpers.IsCRDConditionTrue(updated, apiextensionsv1.Established) {
-					return nil
-				}
-			case watch.Deleted:
-				return fmt.Errorf("CRD %s was deleted before being established", gk.String())
-			case watch.Error:
-				return fmt.Errorf("encountered error while watching CRD %s: %#v", gk.String(), event.Object)
+
+	done := make(chan struct{})
+	wait.Until(func() {
```
stevekuznetsov, line 112: Why poll instead of watching?

stevekuznetsov (changes requested):
Couldn't we do a more minimal change to just allow `AlreadyExist` on the original  create? If we do this, we also have to ensure that the content of the CRD is identical to that which we would be putting down.

config/bootstrap.go
ncdc, line 112: I think the general logic we'd like is:

1. If the CRD already exists, poll until it's established
2. If the CRD doesn't exist, try to create it, then poll until it's established

WDYT?

Also, I'd use wait.PollImmediateInfiniteWithContext - that'll save you having to create a done channel.

stevekuznetsov:
@sttts @ncdc the apiserver folks pilloried CI for polling in the past and had us change over to watching in all cases where we waited. Please help me reconcile that with your two suggestions here :)

ncdc:
@stevekuznetsov watching is fine too. But you'd need to use a reflector/informer instead of setting up your own watch (to avoid dropped/closed watches as @sttts commented earlier).

kylape:
Happy to implement this with a reflector/informer.  Is that the consensus?

stevekuznetsov:
If it's simple, let's do that instead of putting poll loops in.

stevekuznetsov:
I think the important bit is making sure the CRD did not change in between - or someone's local kcp gets into an unintended state.

ncdc:
We could, for the time being, blindly Update() whatever already exists in etcd to match the state of the CRDs in the repo?

kylape:
If there is some urgency to get this issue fixed, I can patch up the poll-based change with the Update() and create a follow up PR with the informer-based solution.  Otherwise I can just work on the informer version straight away.

ncdc:
I believe @stevekuznetsov is interested in fleshing out a more powerful informer-based solution that has broader uses (e.g. in tests).

Maybe for now we can put the small bandaid on (polling/update) while we all work together to design the informer approach? Thoughts @stevekuznetsov @kylape?

kylape:
I went ahead and made the "bandaid" changes.

config/bootstrap.go
```diff
@@ -95,39 +94,35 @@ func BootstrapCustomResourceDefinitionFromFS(ctx context.Context, client apiexte
...
-				return fmt.Errorf("CRD %s was deleted before being established", gk.String())
-			case watch.Error:
-				return fmt.Errorf("encountered error while watching CRD %s: %#v", gk.String(), event.Object)
+		if apierrors.IsNotFound(err) {
+			crdResource, err = client.Create(ctx, rawCrd, metav1.CreateOptions{})
+			if err != nil {
+				return fmt.Errorf("Error creating CRD %s: %w", gk.String(), err)
 			}
+		} else {
+			return fmt.Errorf("Error fetching CRD 1 %s: %w", gk.String(), err)
```
stevekuznetsov, line 105:
```suggestion
			return fmt.Errorf("Error fetching CRD %s: %w", gk.String(), err)
```

config/bootstrap.go
```diff
@@ -95,39 +94,35 @@ func BootstrapCustomResourceDefinitionFromFS(ctx context.Context, client apiexte
...
 		}
 	}
+
+	wait.PollImmediateInfiniteWithContext(ctx, 100*time.Millisecond, func(ctx context.Context) (bool, error) {
+		crd, err := client.Get(ctx, rawCrd.Name, metav1.GetOptions{})
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return false, fmt.Errorf("CRD %s was deleted before being established", gk.String())
+			}
+			return false, fmt.Errorf("Error fetching CRD 2 %s: %w", gk.String(), err)
```
stevekuznetsov, line 121:
```suggestion
			return false, fmt.Errorf("Error fetching CRD %s: %w", gk.String(), err)
```

config/bootstrap.go
kylape: Gah, left these in by accident. Thanks for catching.

PR #300, opened 2021-12-07
test: e2e: framework: add a simple artifact saver

stevekuznetsov:
Deferring the new framework.Artifact() will write a YAML file with the
object that you're asking to be saved. The file will be placed under
directories as necessary for the logical cluster and namespace it's in,
and named by kind.group_name.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

fixes https://github.com/kcp-dev/kcp/issues/271

PR #312, opened 2021-12-08
test: e2e: an informer-driven expectation helper

stevekuznetsov:
test: e2e: an informer-driven expectation helper

In tests, we want to focus on describing the state we expect the system
to fall into, not worrying about handling watch events and the like.
This commit adds a simple informer-driven expectation helper that allows
callers to provide some expectation and have it evaluated when the state
of the world changes.

---

I'm pretty sure this fixes flakes like #268 #286 
Currently based on top of #300 
cc @ncdc @davidfestal @imjasonh


stevekuznetsov:
Rebased, ready for review

test/e2e/framework/expect.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/google/uuid"
+
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/client-go/tools/cache"
+)
+
+// Expectation closes over a statement of intent, allowing the caller
+// to accumulate errors and determine when the expectation should cease
+// to be evaluated.
+type Expectation func() (done bool, err error)
```
ncdc, line 34: Just an FYI, this is the same as wait.ConditionFunc.

ncdc, line 34:
```suggestion
type Expectation func(ctx context.Context) (done bool, err error)
```

test/e2e/framework/expect.go
```diff
@@ -0,0 +1,129 @@
...
+
+// Expecter allows callers to register expectations
+type Expecter interface {
+	// ExpectBefore will result in the Expectation being evaluated whenever
+	// state changes, up until the desired timeout is reached.
+	ExpectBefore(Expectation, time.Duration)
```
ncdc, line 40:
```suggestion
	ExpectBefore(ctx context.Context, expectation Expectation, timeout time.Duration)
```

test/e2e/framework/expect.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func NewExpecter(informer cache.SharedIndexInformer) *expectationController {
```
ncdc, line 43: Godoc please

test/e2e/framework/expect.go
```diff
@@ -0,0 +1,129 @@
...
+		},
+	})
+
+	return &controller
+}
+
+type expectationController struct {
+	informer     cache.SharedIndexInformer
+	expectations map[uuid.UUID]func()
+	lock         *sync.RWMutex
```
ncdc, line 68: If you don't make this a pointer, you don't need to initialize it... do you have a specific need for it to be a pointer - do you plan to pass it around?

test/e2e/framework/expect.go
```diff
@@ -0,0 +1,129 @@
...
+			controller.triggerExpectations()
+		},
+	})
+
+	return &controller
+}
+
+type expectationController struct {
+	informer     cache.SharedIndexInformer
+	expectations map[uuid.UUID]func()
```
ncdc, line 67: I'd suggest adding a comment on what this map is holding

test/e2e/framework/expect.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+}
+
+func (c *expectationController) ExpectBefore(expectation Expectation, duration time.Duration) error {
+	type result struct {
+		done bool
+		err  error
+	}
+	results := make(chan result)
+	wrapped := func() {
```
ncdc, line 86: Mind adding a newline before this to improve readability?

test/e2e/framework/expect.go
```diff
@@ -0,0 +1,129 @@
...
+		done, err := expectation()
+		results <- result{
+			done: done,
+			err:  err,
+		}
+	}
+	id := uuid.New()
```
ncdc, line 93: Mind adding a newline before this to improve readability?

test/e2e/framework/expect.go
```diff
@@ -0,0 +1,129 @@
...
+	c.lock.Lock()
+	c.expectations[id] = wrapped
+	c.lock.Unlock()
+	defer func() {
```
ncdc, line 97: Mind adding a newline before this to improve readability?

test/e2e/framework/expect.go
```diff
@@ -0,0 +1,129 @@
...
+		c.lock.Lock()
+		delete(c.expectations, id)
+		c.lock.Unlock()
+	}()
+
+	// evaluate once to get the current state once we're registered to see future events
+	go func() {
+		wrapped()
+	}()
```
ncdc, line 106: You can just do `go wrapped()` here, can't you?

test/e2e/framework/expect.go
```diff
@@ -0,0 +1,129 @@
...
+
+	var errors []error
+	ctx, cancel := context.WithTimeout(context.Background(), duration)
+	defer func() {
+		cancel()
+	}()
```
ncdc, line 112: You can just do `defer cancel()` here, can't you?

test/e2e/framework/expect.go
```diff
@@ -0,0 +1,129 @@
...
+func (c *expectationController) triggerExpectations() {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	for _, expectation := range c.expectations {
+		expectation()
+	}
+}
+
+func (c *expectationController) ExpectBefore(expectation Expectation, duration time.Duration) error {
```
ncdc, line 80:
```suggestion
func (c *expectationController) ExpectBefore(ctx context.Context, expectation Expectation, duration time.Duration) error {
```

When you're calling this method from test cases, they already have a context. I think it would make sense to take in that context so this code can react to the passed in context being done, if needed. WDYT?

test/e2e/framework/expect.go
ncdc: Also, maybe we could find a more descriptive name for the function?

test/e2e/framework/expect.go
```diff
@@ -0,0 +1,129 @@
...
+		c.lock.Unlock()
+	}()
+
+	// evaluate once to get the current state once we're registered to see future events
+	go func() {
+		wrapped()
+	}()
+
+	var errors []error
+	ctx, cancel := context.WithTimeout(context.Background(), duration)
```
ncdc, line 109: Assuming you agree w/my suggestion to take in a context, wrap that here instead of context.Background()

test/e2e/framework/expect.go
```diff
@@ -0,0 +1,129 @@
...
+	defer func() {
+		cancel()
+	}()
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("expected state not found: %w, errors encountered: %v", ctx.Err(), kerrors.NewAggregate(errors))
+		case result := <-results:
+			if result.err != nil {
+				errors = append(errors, result.err)
```
ncdc, line 119: Ah, so unlike k/k wait.Poll*, here if we get an error, we store it but it's not fatal. And then later on if the expectation func eventually returns done w/no error, we ignore any/all previously stored errors?

test/e2e/reconciler/workspace/controller_test.go
```diff
@@ -23,26 +23,25 @@ import (
...
 	"github.com/kcp-dev/kcp/test/e2e/framework"
 )
 
 func TestWorkspaceController(t *testing.T) {
 	type runningServer struct {
 		framework.RunningServer
-		client  clientset.Interface
-		watcher watch.Interface
+		client clientset.Interface
+		expect func(*tenancyv1alpha1.Workspace, func(workspace *tenancyv1alpha1.Workspace) error) error
```
ncdc, line 44: At first glance, this signature is confusing. Would like to brainstorm to see if we can make it easier to use

test/e2e/reconciler/workspace/controller_test.go
```diff
@@ -252,15 +237,31 @@ func TestWorkspaceController(t *testing.T) {
...
 			testCase.work(ctx, t, runningServer{
 				RunningServer: server,
 				client:        client,
-				watcher:       watcher,
+				expect: func(workspace *tenancyv1alpha1.Workspace, f func(*tenancyv1alpha1.Workspace) error) error {
+					return expecter.ExpectBefore(func() (done bool, err error) {
+						key, err := cache.MetaNamespaceKeyFunc(workspace)
+						if err != nil {
+							return true, err
+						}
```
ncdc, line 256: If you wanted to, you could move this before `return expecter.ExpectBefore`

test/e2e/reconciler/workspace/controller_test.go
ncdc: Maybe
```go
type workspaceConditionFunc func(*Workspace) error
// ...
expect func(*Workspace, workspaceConditionFunc) error
```

test/e2e/reconciler/workspace/controller_test.go
```diff
@@ -252,15 +237,31 @@ func TestWorkspaceController(t *testing.T) {
 				return
 			}
 			client := clients.Cluster(clusterName)
-			watcher, err := client.TenancyV1alpha1().Workspaces().Watch(ctx, metav1.ListOptions{})
-			if err != nil {
-				t.Errorf("failed to watch workspaces: %v", err)
+			kcpSharedInformerFactory := kcpexternalversions.NewSharedInformerFactoryWithOptions(client, 10*time.Minute)
```
ncdc, line 240: Can set resync to 0 (never resync)?

test/e2e/framework/expect.go
stevekuznetsov: I make a habit of not doing the one-liner for this or goroutine creation since if you're passing an arg:

```go
defer whatever(produceVal())
```

The `produceVal()` gets called at the time you register the `defer` IIRC and I can't be bothered to remember which funcs and and which cannot be one-lined. Happy to change it though, this one is safe

test/e2e/framework/expect.go
stevekuznetsov: I figured generally you want to see *why* your expectation func was not passing in the moments leading up to your timeout - if you see "error: context deadline exceeded" the first thing you'll want to know is, did my expected state differ a bit, or did I never get events?

test/e2e/reconciler/workspace/controller_test.go
stevekuznetsov: Yeah, I wanted opinions on this one... and to boot there's no great way to better encapsulate the server's ownership of this process. That is, the server knows how to divulge credentials, but we expect people to pass their own in, since getting from `*rest.Config` is yucky. Furthremore, we will likely always have the lister preamble in expectation funcs, but since those are typed I can't see a way to DRY that out simply.

test/e2e/framework/expect.go
ncdc: 👍 yeah sometimes I get tripped up by that too. But for no-arg funcs like cancel(), I'd prefer it one-lined. Thanks.

test/e2e/framework/expect.go
ncdc: 👍

stevekuznetsov:
Something is goofy with the timeouts ... looking

ncdc (approved):

PR #315, opened 2021-12-10
Fix verify-codegen

ncdc:
Bump go CI version to 1.17.

Fix verify-codegen when running in CI. Because this repo is not cloned into GOPATH, we have to set up a symlink so the code generators that currently only support outputting in GOPATH will work correctly. Otherwise, it will look like code generation always passes verification because it's not checking the right files.

Regenerate clients to match generators in current Kubernetes version.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,12 +2,11 @@ name: CI
 
 on:
   push:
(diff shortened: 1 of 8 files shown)


stevekuznetsov (approved):
Approving so you can push the merge button when you're ready

PR #319, opened 2021-12-10
server: allow configuring etcd WAL size

stevekuznetsov:
Running tests does not require hundreds of MB of WAL, so allow
configuring it to be smaller.

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -35,6 +35,7 @@ import (
 	"time"
 
 	"go.etcd.io/etcd/server/v3/embed"
+	"go.etcd.io/etcd/server/v3/wal"
 
 	"k8s.io/klog/v2"
 )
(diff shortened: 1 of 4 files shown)


stevekuznetsov:
I have no idea how small we can make the WAL before things start to get goofy, but at the very least this size is small enough that even if you're in a tmpfs things should be fine

test/e2e/framework/kcp.go
```diff
@@ -95,6 +96,7 @@ func newKcpServer(t *T, cfg KcpConfig, artifactDir, dataDir string) (*kcpServer,
 			"--listen=:" + kcpListenPort,
 			"--etcd_client_port=" + etcdClientPort,
 			"--etcd_peer_port=" + etcdPeerPort,
+			"--etcd_wal_size_bytes=" + strconv.Itoa(5*1000), // 5KB
```
stevekuznetsov, line 99: I tested with `-count 100` and this seemed to work just fine, no impact on performance and no errors. Factor of 12800 less disk impact!

stevekuznetsov:
Test failure is fixed in https://github.com/kcp-dev/kcp/pull/320

pkg/etcd/etcd.go
```diff
@@ -52,8 +53,11 @@ type ClientInfo struct {
 	TrustedCAFile string
 }
 
-func (s *Server) Run(ctx context.Context, peerPort, clientPort string) (ClientInfo, error) {
+func (s *Server) Run(ctx context.Context, peerPort, clientPort string, walSizeBytes int64) (ClientInfo, error) {
```
ncdc, line 56: We're starting to get to the point where I think I'd like to see us consider either functional server options, or a config struct. Doesn't have to be in this PR though.

PR #323, opened 2021-12-13
Fix linting

ncdc:


stevekuznetsov:
I noticed this ... how did it pass CI????

stevekuznetsov (approved):

ncdc:
My only guess is that golint with go 1.16 didn't catch it (I think the PR was before I bumped to 1.17), but 1.17 does?

Issue #297, opened 2021-12-06
Cross-Cluster List Flake: cannot watch across logical clusters for a resource type with several distinct schemas


stevekuznetsov:
Rarely, we see the following when starting informers based on the cross-cluster client for the workspace controller. We install the CRD into every logical cluster using an identical definition we load from YAML in this repo, so it is not possible for there to be any difference in schema. Ideally, the error would be improved to show what the perceived difference was (or logged at the k8s level).

```
E1206 08:25:57.535530   42874 reflector.go:138] k8s.io/client-go@v0.0.0-20211203153731-1f4695370a63/tools/cache/reflector.go:167: Failed to watch *v1alpha1.WorkspaceShard: failed to list *v1alpha1.WorkspaceShard: Internal error occurred: error resolving resource: cannot watch across logical clusters for a resource type with several distinct schemas
```

jmprusi:
@stevekuznetsov looks like the code producing that error has changed substantially (How the comparison is handled, etc.), are you still able to reproduce it ?

ncdc:
@jmprusi while the code has moved around as part of #302, I don't think the comparison code has changed?

jmprusi:
@ncdc true! couldn't find it. thanks!

stevekuznetsov:
That being said, I have also not been able to reproduce again :)

ncdc:
I had it happen to me once or twice locally, but I haven't been able to reproduce ☹️

ncdc:
I haven't seen this in a long time. Closing. Will reopen if needed.

PR #328, opened 2021-12-15
Adds debug logging if specs mismatch in findCRD

jmprusi:
Helpful for debugging #297

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -35,6 +35,7 @@ import (
 	"time"
 
 	"github.com/emicklei/go-restful"
+	"github.com/google/go-cmp/cmp"
 	"github.com/google/uuid"
 	clientv3 "go.etcd.io/etcd/client/v3"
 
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -934,6 +936,7 @@ func findCRD(crdName string, crds []*apiextensionsv1.CustomResourceDefinition) (
 			crd = aCRD
 		} else {
 			if !equality.Semantic.DeepEqual(crd.Spec, aCRD.Spec) {
+				klog.V(10).Infof("Found multiple CRDs with the same name %q, but different specs: %v", crdName, cmp.Diff(crd.Spec, aCRD.Spec))
```
stevekuznetsov, line 939: V(10) means that when we have a rare flake, nobody will see this log, and you won't be able to debug :)

jmprusi: `klog.Errorf()` then? I didn't want it to be too spammy, but perhaps in this case it should

ncdc: Info (no v specified) or Error is fine for now. Let's put in a TODO to revisit the logging and/or level after we catch the flake, maybe.

ncdc:
FYI I removed the "closes" comment in the description because this won't fix the issue, just help debug it.

stevekuznetsov (approved):

PR #330, opened 2021-12-16
e2e: guard against 0-byte kubeconfig

### LINKED ISSUE #301: E2E Framework: Loading Admin Credentials Flake
```
        test.go:115: failed to load admin kubeconfig: v1.Config.Clusters: []v1.NamedCluster: v1.NamedCluster.Cluster: v1.Cluster.CertificateAuthorityData: decode base64: illegal base64 data at input byte 732, error found in #10 byte of ...|lBYQorQzl"}}]}|..., bigger context ...|Qrd0JTQjdPdjRuaEc5K05kYVVBeXZyV1M5Z0lHYklBYQorQzl"}}]}|...
```

We just load the `admin.kubeconfig` from disk where `kcp` places it ... so ... ?

ncdc:
If we try to read the kubeconfig and the file is 0 bytes, we get an
empty/default Config struct back. Check for this condition, and keep
waiting for a non-empty Config if that's the case.

I _think_ this will fix #301 but I'm not making this auto-close the issue until we've had several e2e iterations in CI that don't flake on this.

### DIFF
--- a/test/e2e/framework/kcp.go
+++ b/test/e2e/framework/kcp.go
@@ -265,22 +265,31 @@ func (c *kcpServer) loadCfg() error {
 	loadCtx, cancel := context.WithTimeout(c.ctx, 1*time.Minute)
 	wait.UntilWithContext(loadCtx, func(ctx context.Context) {
 		kubeconfigPath := filepath.Join(c.dataDir, "admin.kubeconfig")
-		if _, err := os.Stat(kubeconfigPath); os.IsNotExist(err) {
+		if fs, err := os.Stat(kubeconfigPath); os.IsNotExist(err) {
 			return // try again
 		} else if err != nil {
 			loadError = fmt.Errorf("failed to read admin kubeconfig after kcp start: %w", err)
 			return
+		} else if fs.Size() == 0 {
+			return // try again
 		}
+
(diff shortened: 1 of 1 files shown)


test/e2e/framework/kcp.go
```diff
@@ -271,16 +272,26 @@ func (c *kcpServer) loadCfg() error {
 			loadError = fmt.Errorf("failed to read admin kubeconfig after kcp start: %w", err)
 			return
 		}
+
 		rawConfig, err := clientcmd.LoadFromFile(kubeconfigPath)
 		if err != nil {
 			loadError = fmt.Errorf("failed to load admin kubeconfig: %w", err)
 			return
 		}
+		// LoadFromFile can return an empty default config if it read an empty file, so
```
stevekuznetsov, line 281: On L266-274 we have an explicit wait with a `os.Stat` - could we instead use the file info there to wait until the file exists *and* it has a non-zero size?

ncdc: I considered that after I wrote it this way. Would you prefer that?

stevekuznetsov: I think it would be cleaner as to your intent and would not require us to assume the behavior of LoadFromFile / be resilient to changes there

stevekuznetsov (approved):

PR #331, opened 2021-12-16
Update to our Kube 1.23

ncdc:


pkg/server/config.go
```diff
@@ -69,7 +67,6 @@ type Config struct {
 }
 
 func BindOptions(c *Config, fs *pflag.FlagSet) *Config {
-	fs.AddFlag(pflag.PFlagFromGoFlag(flag.CommandLine.Lookup("v")))
```
ncdc, line 72: This comes out because of the switch to component-base's `cli.Run()` helper.

cmd/kcp/kcp.go
```diff
@@ -70,8 +72,8 @@ func main() {
 		},
 	}
 	cfg = server.BindOptions(server.DefaultConfig(), startCmd.Flags())
-	cmd.AddCommand(startCmd)
-	if err := cmd.Execute(); err != nil {
-		fmt.Fprintf(os.Stderr, "error: %v\n", err)
-	}
+
+	kcpCmd.AddCommand(startCmd)
+	code := cli.Run(kcpCmd)
```
ncdc, line 77: I pulled over the cli.Run helper, but not the help/usage func code, because I think that generally assumes there are no subcommands. We can undo this if y'all don't like it...

sttts:
Still ingress test issues, cc @jmprusi

ncdc:
I don't think this is necessarily an ingress e2e issue. I think with all our tests running in parallel, things are really slow. For example, we are hitting the 34-second timeout in the rest createHandler in the server side.

sttts:
from upstream:
```
// For any test of the style:
//   ...
//   <- time.After(timeout):
//      t.Errorf("Timed out")
// The value for timeout should effectively be "forever." Obviously we don't want our tests to truly lock up forever, but 30s
// is long enough that it is effectively forever for the things that can slow down a run on a heavily contended machine
// (GC, seeks, etc), but not so long as to make a developer ctrl-c a test run if they do happen to break that test.
var ForeverTestTimeout = time.Second * 30
```

PR #335, opened 2021-12-16
workspaceshard: record the version of loaded credentials

stevekuznetsov:
Downstream consumers of shard credentials need to know when the
credentials are rotated. For instance, any client that first determines
that data they care about is on a shard, then begins some process to
monitor that data, must know when the credentials used to access that
shard change. In order to not require that all of these clients watch
Secrets, record the version of the credentials that are loaded.

### DIFF
--- a/pkg/apis/tenancy/v1alpha1/types.go
+++ b/pkg/apis/tenancy/v1alpha1/types.go
@@ -210,6 +210,10 @@ type WorkspaceShardStatus struct {
 	// Connection information for the WorkspaceShard.
 	// +optional
 	ConnectionInfo *ConnectionInfo `json:"connectionInfo,omitempty"`
+
+	// Version of credentials last successfully loaded.
+	// +optional
+	CredentialsHash string `json:"credentialsHash,omitempty"`
 }
 
 // ConnectionInfo holds the information necessary to connect to a shard.
--- a/pkg/reconciler/workspaceshard/controller.go
+++ b/pkg/reconciler/workspaceshard/controller.go
@@ -18,6 +18,8 @@ package workspaceshard
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"time"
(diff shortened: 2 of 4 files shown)


pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -210,6 +210,10 @@ type WorkspaceShardStatus struct {
 	// Connection information for the WorkspaceShard.
 	// +optional
 	ConnectionInfo *ConnectionInfo `json:"connectionInfo,omitempty"`
+
+	// Version of credentials last successfully loaded.
+	// +optional
+	CredentialVersion string `json:"credentialVersion,omitempty"`
```
stevekuznetsov, line 216: I wanted something less sensitive than the resource version of the secret - really detect when the underlying kubeconfig changed.

pkg/reconciler/workspaceshard/controller.go
```diff
@@ -289,6 +291,14 @@ func (c *Controller) reconcile(ctx context.Context, workspaceShard *tenancyv1alp
...
 	}
 
+	hash := sha256.New()
+	if _, err := hash.Write(data); err != nil {
+		// TODO: our hash cannot ever return an error on Write, but the interface makes it possible, do we care to do something better?
+		conditions.MarkFalse(workspaceShard, tenancyv1alpha1.WorkspaceShardCredentialsValid, tenancyv1alpha1.WorkspaceShardCredentialsReasonInvalid, "Referenced credentials could not be hashed: %v.", err.Error())
+		return nil
+	}
+
+	workspaceShard.Status.CredentialVersion = base32.StdEncoding.WithPadding(base32.NoPadding).EncodeToString(hash.Sum(nil))
```
ncdc, line 301: Why base32?

test/e2e/reconciler/workspaceshard/controller_test.go
```diff
@@ -258,6 +258,101 @@ func TestWorkspaceShardController(t *testing.T) {
...
+					if diff := cmp.Diff(workspaceShard.Status.ConnectionInfo, &tenancyv1alpha1.ConnectionInfo{
+						Host:    cfg.Host,
+						APIPath: cfg.APIPath,
+					}); diff != "" {
+						return fmt.Errorf("got incorrect connection info: %v", diff)
+					}
+					if workspaceShard.Status.CredentialVersion == "" {
+						return errors.New("no credential version encoded")
+					}
+					version = workspaceShard.Status.CredentialVersion
```
ncdc, line 325: You get a data race here

pkg/apis/tenancy/v1alpha1/types.go
ncdc: A hash makes sense 👍

ncdc: Might want to call it CredentialsHash instead of Version

test/e2e/reconciler/workspaceshard/controller_test.go
ncdc: The race occurs because this anonymous function is called via 2 paths (and therefore 2 goroutines), both of which hold a read lock:

1. Shared informer event -> OnUpdate -> triggerExpectations
2. The one-off execution of the expectation in ExpectBefore (`evaluate once to get the current state once we're registered to see future events`)

Both goroutines have a read lock, and both are allowed to proceed to call the expectation, creating the race.

Remind me why we evaluate the expectation once in a goroutine, before starting the for/select loop?

stevekuznetsov: If you race with the API server and your expectation is registered after the change you're looking for happens in the API server, then you'd never see your func executed

pkg/apis/tenancy/v1alpha1/types.go
stevekuznetsov: I kind of wanted it to be opaque, not sure it's important to know that the credentials were hashed to create it. WDYT?

ncdc (approved):

PR #345, opened 2022-01-03
Update copyright boilerplate to 2022

jmprusi:
Replace 2021 for 2022 to prevent codegen-verify from failing. 

Signed-off-by: Joaquim Moreno <joaquim@redhat.com>


sttts:
In kube we don't do this (anymore), too many PR conflicts. We just use a current year for new files.

jmprusi:
@sttts Ok that makes sense, I will drop this.

PR #353, opened 2022-01-06
cmd: add a shard proxy server

stevekuznetsov:
e2e: framework: expose artifact and data dirs to tests

When we're running many tests in parallel, random directories are chosen
for every invocation. In order to ensure that test processes are placing
artifacts and data in the correct place, we need to pass that
information down.

---

e2e: framework: isolate the accessory process code

---

e2e: rework how artifacts are deferred

Now, you can call server.Artifact() from anywhere and it should
correctly be run before the test is finished and kcp has been
terminated.

---

cmd: add a shard proxy server

Today, this server knows how to find all of the organizational
workspaces, the per-organization workspaces, and map them to the shards
they are placed on through history.

---

### DIFF
--- a/pkg/apis/tenancy/v1alpha1/helper/helper.go
+++ b/pkg/apis/tenancy/v1alpha1/helper/helper.go
@@ -1,0 +1,57 @@
+/*
+Copyright 2021 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package helper
+
+import (
+	"fmt"
+	"strings"
+
+	tenancyapi "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
+)
+
+const (
+	separator = "_"
+
+	// OrganizationCluster is the name of the logical cluster we expect to find
+	// organization inside.
+	OrganizationCluster = "admin"
+)
+
+// EncodeLogicalClusterName determines the logical cluster name for a workspace.
+// We assume that the organization that this workspace resides in is the cluster
+// it lives in.
+func EncodeLogicalClusterName(workspace *tenancyapi.Workspace) (string, error) {
+	orgName := workspace.ClusterName
+	if workspace.ClusterName != OrganizationCluster {
+		_, name, err := ParseLogicalClusterName(workspace.ClusterName)
+		if err != nil {
+			return "", err
+		}
+		orgName = name
+	}
+	return orgName + separator + workspace.Name, nil
+}
+
+// ParseLogicalClusterName determines the organization and workspace name from a
+// logical cluster name.
+func ParseLogicalClusterName(name string) (string, string, error) {
+	parts := strings.Split(name, separator)
+	if len(parts) != 2 {
+		return "", "", fmt.Errorf("expected logical cluster name to be in org_name format, got %s", name)
+	}
+	return parts[0], parts[1], nil
+}
--- a/pkg/reconciler/workspace/controller.go
+++ b/pkg/reconciler/workspace/controller.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net/url"
 	"path"
+	"strconv"
 	"time"
 
 	jsonpatch "github.com/evanphx/json-patch"
@@ -40,6 +41,7 @@ import (
 	"k8s.io/klog/v2"
 
 	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
+	tenancyhelper "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1/helper"
 	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
 	tenancyinformer "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/tenancy/v1alpha1"
 	tenancylister "github.com/kcp-dev/kcp/pkg/client/listers/tenancy/v1alpha1"
@@ -293,7 +295,7 @@ func (c *Controller) reconcile(ctx context.Context, workspace *tenancyv1alpha1.W
 		// make sure current shard still exists
 		currentShard, err := c.workspaceShardLister.Get(clusters.ToClusterAwareKey(workspace.ClusterName, currentShardName))
 		if errors.IsNotFound(err) {
-			klog.Infof("de-scheduling workspace %q from nonexistent shard %q", workspace.Name, currentShardName)
+			klog.Infof("de-scheduling workspace %q|%q from nonexistent shard %q", workspace.ClusterName, workspace.Name, currentShardName)
 			workspace.Status.Location.Current = ""
 		} else if err != nil {
 			return err
@@ -306,17 +308,45 @@ func (c *Controller) reconcile(ctx context.Context, workspace *tenancyv1alpha1.W
 		if err != nil {
 			return err
 		}
-		if len(shards) != 0 {
-			targetShard := shards[rand.Intn(len(shards))]
+		// TODO: do we need to handle cross-cluster shard assignment? if so, we need shard name for the shard objects...
+		var filtered []*tenancyv1alpha1.WorkspaceShard
+		for i := range shards {
+			if shards[i].ClusterName == workspace.ClusterName {
+				filtered = append(filtered, shards[i])
+			}
+		}
+		if len(filtered) != 0 {
+			targetShard := filtered[rand.Intn(len(filtered))]
 			workspace.Status.Location.Target = targetShard.Name
 			shard = targetShard
-			klog.Infof("scheduling workspace %q to %q", workspace.Name, targetShard.Name)
+			klog.Infof("scheduling workspace %q|%q to %q|%q", workspace.ClusterName, workspace.Name, targetShard.ClusterName, targetShard.Name)
 		}
 	}
 	if workspace.Status.Location.Target != "" && workspace.Status.Location.Current != workspace.Status.Location.Target {
-		klog.Infof("moving workspace %q from to %q", workspace.Name, workspace.Status.Location.Target)
+		klog.Infof("moving workspace %q to %q", workspace.Name, workspace.Status.Location.Target)
 		workspace.Status.Location.Current = workspace.Status.Location.Target
 		workspace.Status.Location.Target = ""
+		// TODO: actually handle the RV resolution and double-sided accept we need for movement
+		if len(workspace.Status.Location.History) == 0 {
+			workspace.Status.Location.History = []tenancyv1alpha1.ShardStatus{
+				{Name: workspace.Status.Location.Current, LiveAfterResourceVersion: "1"},
+			}
+		} else {
+			previous := workspace.Status.Location.History[len(workspace.Status.Location.History)-1]
+			startingRV, err := strconv.ParseInt(previous.LiveAfterResourceVersion, 10, 64)
+			if err != nil {
+				conditions.MarkFalse(workspace, tenancyv1alpha1.WorkspaceScheduled, tenancyv1alpha1.WorkspaceReasonUnschedulable, conditionsv1alpha1.ConditionSeverityError, "Invalid status.location.history[%d].liveAfterResourceVersion: %v.", len(workspace.Status.Location.History)-1, err)
+				return nil
+			}
+			endingRV := strconv.FormatInt(startingRV+10, 10)
+			previous.LiveBeforeResourceVersion = endingRV
+			current := tenancyv1alpha1.ShardStatus{
+				Name:                     workspace.Status.Location.Current,
+				LiveAfterResourceVersion: endingRV,
+			}
+			workspace.Status.Location.History[len(workspace.Status.Location.History)-1] = previous
+			workspace.Status.Location.History = append(workspace.Status.Location.History, current)
+		}
 	}
 	if workspace.Status.Location.Current == "" {
 		workspace.Status.Phase = tenancyv1alpha1.WorkspacePhaseInitializing
@@ -334,7 +364,12 @@ func (c *Controller) reconcile(ctx context.Context, workspace *tenancyv1alpha1.W
 			conditions.MarkFalse(workspace, tenancyv1alpha1.WorkspaceURLValid, tenancyv1alpha1.WorkspaceURLReasonInvalid, conditionsv1alpha1.ConditionSeverityError, "Invalid connection information on target WorkspaceShard: %v.", err)
 			return nil
 		}
-		u.Path = path.Join(u.Path, shard.Status.ConnectionInfo.APIPath, "clusters", workspace.Name)
+		logicalCluster, err := tenancyhelper.EncodeLogicalClusterName(workspace)
+		if err != nil {
+			conditions.MarkFalse(workspace, tenancyv1alpha1.WorkspaceURLValid, tenancyv1alpha1.WorkspaceURLReasonInvalid, conditionsv1alpha1.ConditionSeverityError, "Invalid Workspace location: %v.", err)
+			return nil
+		}
+		u.Path = path.Join(u.Path, shard.Status.ConnectionInfo.APIPath, "clusters", logicalCluster)
 		workspace.Status.BaseURL = u.String()
 		conditions.MarkTrue(workspace, tenancyv1alpha1.WorkspaceURLValid)
 	}
--- a/pkg/reconciler/workspaceindex/controller.go
+++ b/pkg/reconciler/workspaceindex/controller.go
@@ -1,0 +1,472 @@
+/*
+Copyright 2021 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package workspaceindex
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+
+	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1/helper"
+	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	kcpexternalversions "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
+	tenancyinformer "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/tenancy/v1alpha1"
+	tenancylister "github.com/kcp-dev/kcp/pkg/client/listers/tenancy/v1alpha1"
+	"github.com/kcp-dev/kcp/third_party/conditions/util/conditions"
+)
+
+const (
+	rootShard         = "root"
+	currentShardIndex = "shard"
+	controllerName    = "workspaceindex-controller"
+)
+
+const resyncPeriod = 10 * time.Hour
+
+func NewController(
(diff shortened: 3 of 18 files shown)


stevekuznetsov:
Depends on https://github.com/kcp-dev/kubernetes/pull/32

pkg/apis/tenancy/v1alpha1/helper/helper.go
```diff
@@ -0,0 +1,57 @@
...
+	separator = "_"
+
+	// OrganizationCluster is the name of the logical cluster we expect to find
+	// organization inside.
+	OrganizationCluster = "admin"
+)
+
+// EncodeLogicalClusterName determines the logical cluster name for a workspace.
+// We assume that the organization that this workspace resides in is the cluster
+// it lives in.
```
sttts, line 36: "living in" is confusing here. You want to say that we prefix the workspace name with the org cluster. It does not mean that the workspace must be on the same shard as the org workspace, does it?

pkg/reconciler/workspaceindex/orchestrator.go
```diff
@@ -0,0 +1,506 @@
...
+
+const (
+	rootShard         = "root"
+	currentShardIndex = "shard"
+	controllerName    = "workspaceindex-orchestrator"
+)
+
+const resyncPeriod = 10 * time.Hour
+
+func NewOrchestrator(
```
sttts, line 55: is this something else than a controller?

pkg/reconciler/workspaceindex/orchestrator.go
```diff
@@ -0,0 +1,506 @@
...
+	workspaceInformer tenancyinformer.WorkspaceInformer,
+	workspaceShardInformer tenancyinformer.WorkspaceShardInformer,
+) (*Orchestrator, error) {
+	orgQueue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+	workspaceQueue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+
+	c := &Orchestrator{
+		orgQueue:              orgQueue,
+		workspaceQueue:        workspaceQueue,
+		activeWorkersLock:     sync.RWMutex{},
```
sttts, line 66: this is implicit as the zero value

pkg/reconciler/workspaceindex/orchestrator.go
```diff
@@ -0,0 +1,506 @@
...
+		informerLock:     sync.RWMutex{},
+		shardInformers:   map[string]organizationInformer{},
+		mappingLock:      sync.RWMutex{},
+		workspaceMapping: map[string]map[string][]parsedShardStatus{},
+	}
+
+	workspaceInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			c.enqueueOrg(obj)
+			c.enqueueWorkspace(rootShard, obj)
```
sttts, line 85: usually there is a type-cast here. You know the type. Would make the whole code easier to reason about.

pkg/reconciler/workspaceindex/orchestrator.go
```diff
@@ -0,0 +1,506 @@
...
+// Organizations that need to be watched as well as the credentials with which that
+// should be done. For each Organization, Workspaces are watched in order to build a
+// mapping of logical cluster name and resource version to the kcp shard on which the
+// data will be found.
+type Orchestrator struct {
+	orgQueue                    workqueue.RateLimitingInterface
+	orgQueueActiveWorkers       uint64
+	workspaceQueue              workqueue.RateLimitingInterface
+	workspaceQueueActiveWorkers uint64
+	activeWorkersLock           sync.RWMutex
```
sttts, line 124: what is this protecting? By convention it is above those fields it protects.

pkg/reconciler/workspaceindex/orchestrator.go
```diff
@@ -0,0 +1,506 @@
...
+	workspaceIndexer cache.Indexer
+	workspaceLister  tenancylister.WorkspaceLister
+
+	workspaceShardIndexer cache.Indexer
+	workspaceShardLister  tenancylister.WorkspaceShardLister
+
+	syncChecks []cache.InformerSynced
+
+	// shardInformers are the set of informers we manage for each shard that orgs are on
+	informerLock   sync.RWMutex
```
sttts, line 138: would be systematic here: shardInformersLock, same below for workspace mapping.

pkg/reconciler/workspaceindex/server.go
```diff
@@ -0,0 +1,159 @@
...
+
+	"github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1/helper"
+)
+
+func (o *Orchestrator) stable() bool {
+	o.activeWorkersLock.RLock()
+	defer o.activeWorkersLock.RUnlock()
+	orgLen, orgWorkers, workspaceLen, workspaceWorkers := o.orgQueue.Len(), o.orgQueueActiveWorkers, o.workspaceQueue.Len(), o.workspaceQueueActiveWorkers
+	klog.Infof("workspaceQueue.Len()=%d, workspaceQueueActiveWorkers=%d, orgQueue.Len()=%d, orgQueueActiveWorkers=%d", workspaceLen, workspaceWorkers, orgLen, orgWorkers)
+	return workspaceWorkers == 0 && orgWorkers == 0 && orgLen == 0 && workspaceLen == 0
```
sttts, line 39: hairy logic. As discussed yesterday: wait for listers to be complete and iterate over the objects once. Much simpler.

pkg/reconciler/workspaceindex/server.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+type cacheSyncWaiter interface {
+	WaitForCacheSync(stopCh <-chan struct{}) map[reflect.Type]bool
+}
+
+func (o *Orchestrator) ListenAndServe(port int, waiter cacheSyncWaiter) {
```
sttts, line 46: what is this for?

sttts: assuming this is the proxy?

pkg/reconciler/workspaceindex/orchestrator.go
```diff
@@ -0,0 +1,506 @@
...
+
+	const clusterAll = "*"
+	kcpClient, err := kcpclient.NewClusterForConfig(cfg)
+	if err != nil {
+		klog.Errorf("error creating client with workspace shard credentials: %v", err)
+		return nil
+	}
+
+	octx, cancel := context.WithCancel(ctx)
+	crossClusterKcpClient := kcpClient.Cluster(clusterAll)
```
sttts, line 332: clusterAll is surprising here. I would have expected to watch just the new org workspace.

sttts: ... whose lcluster we know.

pkg/apis/tenancy/v1alpha1/helper/helper.go
stevekuznetsov: Yikes. Naming. The `Workspace` object for a workspace has a `metadata.clusterName`. This is the organizational workspace which owns the workspace. That's what this comment means.

pkg/reconciler/workspaceindex/server.go
stevekuznetsov: Can't do that, if we know we have outstanding work ... why bother serving? :)

stevekuznetsov: Plus, the logic for "list root once, list all orgs once, list all workspaces in an org once ..." is hard to reason about as well (and totally useless past startup)

pkg/reconciler/workspaceindex/orchestrator.go
stevekuznetsov: Why do we need to cast? We just need to get the key for it

stevekuznetsov:
Note to reviewers - yes, there is a lot of space here to write unit tests etc for the actual mapping of user-facing RV to etcd-local RV but as of now we don't even have the right types or process in place to migrate data, so I would love to hold off on locking into an impl until the basics are in and we can do something more intelligent while actually landing a real migration workflow.

pkg/reconciler/workspaceindex/server.go
```diff
@@ -0,0 +1,164 @@
...
+		})
+		if idx == len(history) {
+			// this should never happen, since the current shard should have no liveBefore set
+			http.Error(w, fmt.Sprintf("cluster %q has no shard history containing resourceVersion %d", clusterName, resourceVersion), http.StatusInternalServerError)
+			return
+		}
+		shardName = history[idx].Name
+	}
+	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
+	fmt.Fprint(w, shardName)
```
ncdc, line 148: Do we want to return json here?

pkg/reconciler/workspaceindex/server.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+type cacheSyncWaiter interface {
+	WaitForCacheSync(stopCh <-chan struct{}) map[reflect.Type]bool
+}
+
+func (s *server) ListenAndServe(ctx context.Context) {
+	mux := http.NewServeMux()
+	mux.Handle("/shard", http.HandlerFunc(s.handleShard))
+	mux.Handle("/data", http.HandlerFunc(s.handleData))
```
ncdc, line 63: Do we want to include an API version in the path, thinking about the future?

pkg/server/handler.go
```diff
@@ -30,7 +30,7 @@ import (
 	"k8s.io/kubernetes/pkg/genericcontrolplane"
 )
 
-var reClusterName = regexp.MustCompile(`^[a-z0-9][a-z0-9-]{0,78}[a-z0-9]$`)
+var reClusterName = regexp.MustCompile(`^([a-z0-9][a-z0-9-]{0,78}[a-z0-9]_)?[a-z0-9][a-z0-9-]{0,78}[a-z0-9]$`)
```
ncdc, line 33: Would you be able to write a comment describing the format?

test/e2e/reconciler/workspaceindex/controller_test.go
```diff
@@ -0,0 +1,452 @@
...
+	kcpclientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	"github.com/kcp-dev/kcp/test/e2e/framework"
+	utilconditions "github.com/kcp-dev/kcp/third_party/conditions/util/conditions"
+)
+
+type runningServer struct {
+	framework.RunningServer
+	client      clientset.Interface
+	kubeClient  kubernetesclientset.Interface
+	expect      framework.RegisterWorkspaceExpectation
```
ncdc, line 53: Want to call this expectWorkspace?

test/e2e/reconciler/workspaceindex/controller_test.go
```diff
@@ -0,0 +1,452 @@
...
+						initErrorChan <- fmt.Errorf("failed to resolve raw credentials: %w", err)
+						return
+					}
+					if err := initializeShard(ctx, t, mainServer, serverName, rawCfg); err != nil {
+						initErrorChan <- err
+						return
+					}
+				}(serverName)
+			}
+			close(initErrorChan)
```
ncdc, line 241: Could this result in panic (send on closed channel) in the goroutines above? Does it need to be moved after wg.Wait()?

pkg/reconciler/workspace/controller.go
```diff
@@ -306,17 +308,45 @@ func (c *Controller) reconcile(ctx context.Context, workspace *tenancyv1alpha1.W
...
+				{Name: workspace.Status.Location.Current, LiveAfterResourceVersion: "1"},
+			}
+		} else {
+			previous := workspace.Status.Location.History[len(workspace.Status.Location.History)-1]
+			startingRV, err := strconv.ParseInt(previous.LiveAfterResourceVersion, 10, 64)
+			if err != nil {
+				conditions.MarkFalse(workspace, tenancyv1alpha1.WorkspaceScheduled, tenancyv1alpha1.WorkspaceReasonUnschedulable, conditionsv1alpha1.ConditionSeverityError, "Invalid status.location.history[%d].liveAfterResourceVersion: %v.", len(workspace.Status.Location.History)-1, err)
+				return nil
+			}
+			endingRV := strconv.FormatInt(startingRV+10, 10)
```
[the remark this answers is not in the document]
stevekuznetsov: Yeah, I basically just wanted some hack for not having them be overlapping for now.

pkg/reconciler/workspaceindex/server.go
```diff
@@ -0,0 +1,164 @@
...
+	healthz.InstallReadyzHandler(mux, healthz.NamedCheck("workspaces-synced", func(r *http.Request) error {
+		if !s.stable() {
+			return errors.New("work queues have outstanding keys")
+		}
+		return nil
+	}), healthz.NewInformerSyncHealthz(s.waiter))
+	httpServer := http.Server{Addr: ":" + strconv.Itoa(s.port), Handler: mux}
+	go func() {
+		<-ctx.Done()
+		if err := httpServer.Shutdown(context.Background()); err != nil {
```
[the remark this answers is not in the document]
stevekuznetsov: Yeah, I wrote an `interrupts` library to do graceful termination for Prow components, wasn't sure how much this needed to be flushed out and/or if k8s had a similar concept? I think controller-runtime does.

pkg/reconciler/workspaceindex/server.go
stevekuznetsov: I don't think necessarily that we need this API to even exist in the future, but it's a convenient way to test this in a black-box e2e.

pkg/reconciler/workspaceindex/server.go
stevekuznetsov: I don't think necessarily that we need this API to even exist in the future, but it's a convenient way to test this in a black-box e2e.

pkg/server/handler.go
stevekuznetsov: I don't actually know what the format *is* - it looks like a domain name segment or something similar? I can't tell why there's a length limit, etc - all I did was allow for two of these such things with a separator that's not in the allowed set of characters, to make decoding simple.

stevekuznetsov: @smarterclayton @imjasonh @davidfestal any of you know where this regex came from?

test/e2e/reconciler/workspaceindex/controller_test.go
stevekuznetsov: Good catch1

pkg/reconciler/workspaceindex/controller.go
```diff
@@ -0,0 +1,450 @@
...
+
+	kcpClient, err := kcpclient.NewClusterForConfig(cfg)
+	if err != nil {
+		klog.Errorf("error creating client with workspace shard credentials: %v", err)
+		return nil
+	}
+
+	octx, cancel := context.WithCancel(ctx)
+	crossClusterKcpClient := kcpClient.Cluster(logicalCluster)
+	kcpSharedInformerFactory := kcpexternalversions.NewSharedInformerFactoryWithOptions(crossClusterKcpClient, resyncPeriod)
```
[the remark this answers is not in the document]
ncdc: Makes sense. We can always reevaluate if we run into performance issues w/multiple informers.

pkg/reconciler/workspaceindex/index.go
```diff
@@ -0,0 +1,150 @@
...
+	if !workspaceExists {
+		return nil, fmt.Errorf("workspace %q not found", workspace)
+	}
+	return history, nil
+}
+
+func (i *index) List() map[string]map[string][]ShardAssignment {
+	i.RLock()
+	defer i.RUnlock()
+	return i.workspaceMapping
```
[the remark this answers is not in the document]
ncdc: Was I correct that this is only used to write out json?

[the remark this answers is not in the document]
stevekuznetsov: Yes!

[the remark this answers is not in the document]
ncdc: I'd say for now, even though this code exists for the benefit of e2e testing, we can still be somewhat defensive and implement json.Marshaler instead of handing out the map to callers. WDYT?

pkg/reconciler/workspaceindex/controller.go
```diff
@@ -0,0 +1,472 @@
...
+		}
+		return err
+	}
+	obj = obj.DeepCopy()
+
+	return c.index.Record(obj)
+}
+
+// Stable exposes whether - to the best of our knowledge - the controller has stabilized.
+func (c *Controller) Stable() bool {
```
sttts, line 466: am still not sure this is a viable approach. Stability of workspaces might never happen in every non-toy kcp.

Instead synchronously get the root org workspace once. Then start the org informers, synchronously wait for them to start up, but with some upper bound on waiting time (e.g. 60s). For every incoming org request (we know the org from the lcluster, right?) that is not in-sync return 50x with retry-after 10s or so.

sttts: Maybe, we don't have to do that right away. Follow up is fine, but put a big TODO here.

pkg/reconciler/workspaceindex/index.go
```diff
@@ -0,0 +1,149 @@
...
+		return nil, fmt.Errorf("organization %q not found", organization)
+	}
+	history, workspaceExists := historyByWorkspace[workspace]
+	if !workspaceExists {
+		return nil, fmt.Errorf("workspace %q not found", workspace)
+	}
+	return history, nil
+}
+
+func (i *index) List() map[string]map[string][]ShardAssignment {
```
sttts, line 145: I want some doc of this data structure :-) Or maybe a good type name.

pkg/reconciler/workspaceindex/index.go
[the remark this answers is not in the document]
sttts: > This gives back the underlying map which the caller could then destructively mutate

A go doc about the "this is not owned by the caller" would help.

pkg/reconciler/workspace/controller.go
```diff
@@ -306,17 +308,45 @@ func (c *Controller) reconcile(ctx context.Context, workspace *tenancyv1alpha1.W
 		if err != nil {
 			return err
 		}
-		if len(shards) != 0 {
-			targetShard := shards[rand.Intn(len(shards))]
+		// TODO: do we need to handle cross-cluster shard assignment? if so, we need shard name for the shard objects...
+		var filtered []*tenancyv1alpha1.WorkspaceShard
+		for i := range shards {
+			if shards[i].ClusterName == workspace.ClusterName {
```
sttts, line 314: not sure I understand this filtering. It's saying to schedule a Workspace to only those shards that are defined on the same shard the Workspace as object lives on. Why?

pkg/reconciler/workspaceindex/index.go
```diff
@@ -0,0 +1,152 @@
...
+	sync.RWMutex
+	// workspaceMapping holds resourceVersion-bound locations for workspaces
+	// through history. The list of locations is sorted. The mapping is from
+	// org to workspace to history.
+	workspaceMapping map[string]map[string][]ShardAssignment
+}
+
+func (i *index) Record(workspace *tenancyv1alpha1.Workspace) error {
+	if !conditions.IsTrue(workspace, tenancyv1alpha1.WorkspaceScheduled) {
+		klog.Infof("workspace %s/%s not scheduled, skipping...", workspace.ClusterName, workspace.Name)
```
sttts, line 66: V(4)

pkg/reconciler/workspaceindex/index.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+		org = workspaceName
+	}
+
+	i.Lock()
+	if _, ok := i.workspaceMapping[org]; !ok {
+		i.workspaceMapping[org] = map[string][]ShardAssignment{}
+	}
+	i.workspaceMapping[org][workspace.Name] = history
+	klog.Infof("added history for %s->%s:%v", org, workspace.Name, history)
```
sttts, line 93: V(2)

pkg/reconciler/workspaceindex/index.go
```diff
@@ -0,0 +1,152 @@
...
+		return nil, fmt.Errorf("organization %q not found", organization)
+	}
+	history, workspaceExists := historyByWorkspace[workspace]
+	if !workspaceExists {
+		return nil, fmt.Errorf("workspace %q not found", workspace)
+	}
+	return history, nil
+}
+
+func (i *index) MarshalJSON() ([]byte, error) {
```
sttts, line 146: this is just for debugging, right? Maybe worth a godoc.

pkg/reconciler/workspaceindex/server.go
sttts: `/debug/data` is a better endpoint name, and used in many kube components.

pkg/reconciler/workspaceindex/server.go
sttts: so in other words: the actual proxy handler is still missing. It would take `/clusters/<org>_<worksapce>` to the very same path on the shard, i.e. `io.Copy` the connection. Would like to see this being implemented. Seems to be a slight variation of the upper code.

pkg/reconciler/workspaceindex/server.go
sttts: same for `/debug/shard`

pkg/reconciler/workspaceindex/index.go
stevekuznetsov: I imagine it will be removed in the next PR ... trying not to over-bake this so we have something to actually show (proxying) for the work :)

pkg/reconciler/workspace/controller.go
stevekuznetsov: See the TODO above this - if we want cross-workspace shard assignment OR more than one logical cluster in which shard objects live (maybe it will be just one admin one on the root shard?), then we need to qualify the cluster in the shard assignment. Today, we don't do this - the object only has one string to refer to a shard.

sttts: Shards should live in the root cluster. If there don't yet, we have to move them there.

stevekuznetsov: We can't. It's all written out ... we cannot do that until we have a cross-shard scheduling controller. I need to get this work merged before we can have a virtual workspace that shows one controller all workspaces in need of scheduling.

sttts:
Comments to be addressed in follow-ups.

sttts (approved):

Issue #285, opened 2021-12-03
Transparent Multicluster: pcluster health checks


pweil-:
Use Case: Transparent Multicluster

Objectives:

- User can add a second location
- Deployed application moves between locations when service is disrupted (cluster down)
- Ingress follows application

Details

KCP should health-check a pcluster to notice when it's unhealthy and its assignments should be scheduled to another pcluster

sttts:
Which of the other tasks does this depend on?

jmprusi:
@sttts the kcp-ingress controller already handles changes on the ingress location.  I guess it is the scheduler's responsibility to switch the Ingress location when a pcluster is down.

sttts:
@jmprusi are you working on this or is it rather @imjasonh's domain?

jmprusi:
@sttts I think @imjasonh is on it, but I can take it if he has no bandwidth for this or prefers to focus on other stuff.

sttts:
Some questions:

1. How do we know that a cluster is healthy or not?
2. For a second location, what is missing? An API type?

jmprusi:
> 1. How do we know that a cluster is healthy or not?

That feels like a "Syncer" job, as we should avoid reaching the pCluster apiserver directly from kcp due to possible networking issues / requiring more permissions. 

The Syncer can report its status (running), adding some extra info like the pCluster Readyz endpoint and updating the proper Cluster Status conditions. With that information, and the last update timestamp, we can infer (minus edge cases :) ) the cluster condition:

- Cluster reported healthy and inside the time threshold? Nothing to do
- Last status report is above a threshold? Trigger a failover
- Cluster reported as unhealthy: Trigger a failover

This is a really basic health check. In the future, we can extend the health conditions to consider capacity calculations to decide whether a cluster is ok or not or other types of metrics.

sttts:
Actually that resembles kubelet heartbeating in the node status. If there is no update, a controller in the control plane will mark the node as unready.

eddycharly:
Would be happy to give it a try, will be a good opportunity to learn more about kcp if it's ok.

sttts:
First step would be an API draft in https://docs.google.com/document/d/1k72fgGBNJzbY1Rkxj9_1WksHZR0cNT5OeadCpWuc-zM/edit#.

ncdc:
This is how cluster-api does a remote cluster apiserver health check:

[...]

PR #356, opened 2022-01-10
Add lease controller to syncer

eddycharly:
This PR adds one lease controller to syncer (part of #285).

Syncer (in pull mode) will renew the lease periodically and KCP should use the lease infos to health check physical clusters.

NOTE: ideally lease should be owned by cluster resource in KCP so that deleting a cluster also removes the lease.

---

```shell
$ k get cluster,lease -A -o wide
NAME                            LOCATION   READY   SYNCED API RESOURCES
cluster.cluster.example.dev/a   a          True    ["statefulsets.apps"]
cluster.cluster.example.dev/b   b          True    ["statefulsets.apps"]
cluster.cluster.example.dev/c   c          True    ["statefulsets.apps"]

NAMESPACE       NAME        HOLDER      AGE
syncer-system   cluster-a   cluster-a   9s
syncer-system   cluster-b   cluster-b   11s
syncer-system   cluster-c   cluster-c   11s
```

### DIFF
--- a/cmd/syncer/main.go
+++ b/cmd/syncer/main.go
@@ -36,11 +36,12 @@ import (
 const numThreads = 2
 
 var (
-	fromKubeconfig = flag.String("from_kubeconfig", "", "Kubeconfig file for -from cluster.")
-	fromCluster    = flag.String("from_cluster", "", "Name of the -from logical cluster.")
-	toKubeconfig   = flag.String("to_kubeconfig", "", "Kubeconfig file for -to cluster. If not set, the InCluster configuration will be used.")
-	toContext      = flag.String("to_context", "", "Context to use in the Kubeconfig file for -to cluster, instead of the current context.")
-	clusterID      = flag.String("cluster", "", "ID of the -to cluster. Resources with this ID set in the 'kcp.dev/cluster' label will be synced.")
+	fromKubeconfig       = flag.String("from_kubeconfig", "", "Kubeconfig file for -from cluster.")
+	fromCluster          = flag.String("from_cluster", "", "Name of the -from logical cluster.")
+	toKubeconfig         = flag.String("to_kubeconfig", "", "Kubeconfig file for -to cluster. If not set, the InCluster configuration will be used.")
+	toContext            = flag.String("to_context", "", "Context to use in the Kubeconfig file for -to cluster, instead of the current context.")
+	clusterID            = flag.String("cluster", "", "ID of the -to cluster. Resources with this ID set in the 'kcp.dev/cluster' label will be synced.")
+	leaseDurationSeconds = flag.Int("lease-duration-seconds", 0, "Cluster lease duration seconds, set to 0 to disable lease controller (default).")
 )
 
 func main() {
@@ -91,7 +92,7 @@ func main() {
 	defer cancel()
 
 	klog.Infoln("Starting workers")
-	syncer, err := syncer.StartSyncer(fromConfig, toConfig, sets.NewString(syncedResourceTypes...), *clusterID, *fromCluster, numThreads)
+	syncer, err := syncer.StartSyncer(fromConfig, toConfig, sets.NewString(syncedResourceTypes...), *clusterID, *fromCluster, numThreads, int32(*leaseDurationSeconds))
 	if err != nil {
 		klog.Fatal(err)
 	}
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -23,41 +23,56 @@ import (
 	"strings"
 	"time"
 
+	coordinationv1 "k8s.io/api/coordination/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/clock"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/dynamicinformer"
+	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/component-helpers/apimachinery/lease"
 	"k8s.io/klog/v2"
 )
 
 const resyncPeriod = 10 * time.Hour
 const SyncerNamespaceKey = "SYNCER_NAMESPACE"
 
+// leaseRenewIntervalFraction is the fraction of lease duration to renew the lease
+const leaseRenewIntervalFraction = 0.25
+
+// syncerNamespace is the syncer namespace where cluster leases are created
+const syncerNamespace = "syncer-system"
+
 type Syncer struct {
+	stopCh       chan struct{}
 	specSyncer   *Controller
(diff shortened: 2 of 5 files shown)


cmd/syncer/main.go
```diff
@@ -33,11 +33,13 @@ import (
...
-	toContext      = flag.String("to_context", "", "Context to use in the Kubeconfig file for -to cluster, instead of the current context.")
-	clusterID      = flag.String("cluster", "", "ID of the -to cluster. Resources with this ID set in the 'kcp.dev/cluster' label will be synced.")
+	fromKubeconfig         = flag.String("from_kubeconfig", "", "Kubeconfig file for -from cluster.")
+	fromCluster            = flag.String("from_cluster", "", "Name of the -from logical cluster.")
+	toKubeconfig           = flag.String("to_kubeconfig", "", "Kubeconfig file for -to cluster. If not set, the InCluster configuration will be used.")
+	toContext              = flag.String("to_context", "", "Context to use in the Kubeconfig file for -to cluster, instead of the current context.")
+	clusterID              = flag.String("cluster", "", "ID of the -to cluster. Resources with this ID set in the 'kcp.dev/cluster' label will be synced.")
+	leaseControllerEnabled = flag.Bool("lease-controller-enabled", false, "Enable lease controller (should be set to true only in pull mode).")
```
imjasonh, line 41: The syncer called from `main.go` is always going to be in pull-mode. It's only when `syncer.StartSyncer` etc is wrapped in cluster controller code that it's push mode. So this comment can just drop the parenthetical about pull mode.

cmd/syncer/main.go
```diff
@@ -33,11 +33,13 @@ import (
...
+	leaseDurationSeconds   = flag.Int("lease-duration-seconds", 40, "Cluster lease duration seconds.")
```
imjasonh, line 42: How about having only one flag, where `--lease-duration-seconds=0` indicates there should be no lease?

pkg/syncer/syncer.go
```diff
@@ -67,13 +75,34 @@ func StartSyncer(upstream, downstream *rest.Config, resources sets.String, clust
...
+	leaseController := lease.NewController(
+		clock.RealClock{},
+		leaseClient,
+		fmt.Sprintf("cluster-%s", cluster),
+		int32(leaseDurationSeconds),
+		// TODO: onRepeatedHeartbeatFailure ? closeAllConns ?
+		nil,
+		time.Duration(float64(time.Duration(leaseDurationSeconds))*leaseRenewIntervalFraction),
+		// TODO: namespace ?
+		v1.NamespaceNodeLease,
```
imjasonh, line 92: How about `syncer-system`? This is the namespace the syncer gets installed into in physical clusters, so this keeps us from having to think about two different namespace names.

pkg/syncer/syncer.go
```diff
@@ -67,13 +75,34 @@ func StartSyncer(upstream, downstream *rest.Config, resources sets.String, clust
...
+	if err != nil {
+		specSyncer.Stop()
+		return nil, err
+	}
+	leaseController := lease.NewController(
+		clock.RealClock{},
+		leaseClient,
+		fmt.Sprintf("cluster-%s", cluster),
+		int32(leaseDurationSeconds),
+		// TODO: onRepeatedHeartbeatFailure ? closeAllConns ?
```
imjasonh, line 88: At least log a warning for now, and we can add more when we figure out what else we should do.

pkg/syncer/syncer.go
```diff
@@ -67,13 +75,34 @@ func StartSyncer(upstream, downstream *rest.Config, resources sets.String, clust
 		specSyncer.Stop()
 		return nil, err
 	}
+	leaseClient, err := clientset.NewForConfig(upstream)
+	if err != nil {
+		specSyncer.Stop()
```
imjasonh, line 80: Should this also stop the `statusSyncer`?

pkg/syncer/syncer.go
```diff
@@ -67,13 +75,34 @@ func StartSyncer(upstream, downstream *rest.Config, resources sets.String, clust
...
+		clock.RealClock{},
+		leaseClient,
+		fmt.Sprintf("cluster-%s", cluster),
+		int32(leaseDurationSeconds),
+		// TODO: onRepeatedHeartbeatFailure ? closeAllConns ?
+		nil,
+		time.Duration(float64(time.Duration(leaseDurationSeconds))*leaseRenewIntervalFraction),
+		// TODO: namespace ?
+		v1.NamespaceNodeLease,
+		// TODO: newLeasePostProcessFunc ? set owner ?
```
imjasonh, line 93: Again, probably at least log something here and we can figure out what else to add later.

imjasonh (commented):
Nice!!

cmd/syncer/main.go
eddycharly: ✅

cmd/syncer/main.go
eddycharly: I wonder if the lease controller needs to be part of the `Syncer` struct or just an additional controller stated in `main`.
Looks like they serve different purposes.

pkg/syncer/syncer.go
eddycharly: ✅

pkg/syncer/syncer.go
eddycharly: ✅ 
What surprises me though is that leases are created in a namespace that does not exist 🤔 
```shell
$ k get lease -A -o wide
NAMESPACE       NAME        HOLDER      AGE
syncer-system   cluster-a   cluster-a   9s
syncer-system   cluster-b   cluster-b   11s
syncer-system   cluster-c   cluster-c   11s
```
```shell
$ k get ns
No resources found
```

pkg/syncer/syncer.go
eddycharly: ✅

pkg/syncer/syncer.go
eddycharly: ✅

cmd/syncer/main.go
imjasonh: That could make sense. Do we think the push-mode syncer will ever use a lease? If not, the lease controller could only be started by `cmd/syncer/main.go`.

Should the lease controller have any hooks into syncer logic? For example, if we fail to renew the lease, should the syncer stop syncing? If those are totally separate responsibilities they make more sense being totally separate in `main.go`, but if they interact with each other they might benefit from being more aware of each other.

eddycharly: Good point, i will let it here for now.

I'll see if it makes sense to move it when implementing health check in the cluster controller.

ncdc:
Hi @eddycharly, sorry we haven't gotten back to this. If you'd still like to get it merged, would you please rebase? Thanks!

pkg/syncer/syncer.go
```diff
@@ -23,41 +23,56 @@ import (
...
 )
 
 const resyncPeriod = 10 * time.Hour
 const SyncerNamespaceKey = "SYNCER_NAMESPACE"
 
+// leaseRenewIntervalFraction is the fraction of lease duration to renew the lease
+const leaseRenewIntervalFraction = 0.25
+
+// syncerNamespace is the syncer namespace where cluster leases are created
+const syncerNamespace = "syncer-system"
```
ncdc, line 52: Maybe kcp-system? cc @sttts

pkg/syncer/syncer.go
```diff
@@ -23,41 +23,56 @@ import (
...
+
 type Syncer struct {
+	stopCh       chan struct{}
```
ncdc, line 55: This has been refactored in main to use the ctx that's passed in to StartSyncer to control when to stop - please adjust accordingly. Thanks!

marun:
@eddycharly Given the lack of activity on this PR and our need for this functionality for P3, I'm going to pick it up in a new PR. I'll ensure that your co-authorship of the eventual result is maintained.

PR #357, opened 2022-01-11
Improve syncer install/uninstall process

eddycharly:
This PR improves `syncer` install/uninstall process.

Basically, it takes care of uninstalling `syncer` from target clusters when necessary (config changed, going from pull to push mode).

### DIFF
--- a/pkg/reconciler/cluster/cluster.go
+++ b/pkg/reconciler/cluster/cluster.go
@@ -62,6 +62,7 @@ func (c *Controller) reconcile(ctx context.Context, cluster *clusterv1alpha1.Clu
 		return nil // Don't retry.
 	}
 
+	// Start api importer
 	if c.apiImporters[cluster.Name] == nil {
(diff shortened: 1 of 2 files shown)


stevekuznetsov (commented):
Yikes. I know this predates the changes you're making here but the cyclomatic complexity of these methods is really, really high. I certainly don't feel like I can provide a good review (or even keep all the if/then logic in my head).

Perhaps @imjasonh  or @davidfestal want to look at?

eddycharly:
@stevekuznetsov I agree this is not the most straightforward piece of code to review. I didn't change the code that much but maybe it could have been implemented in a simpler way.

Not sure if changing the cluster controller mode is a use case we want to support by now. If there's no review i will close the PR and will eventually come back to it later.

ncdc:
We are planning on removing push mode from the code base. Are there other parts of this PR that are worth extracting & keeping, or should we close? cc @marun @imjasonh

ncdc:
I'm going to close this, given the plans to remove the functionality from the cluster controller to manage syncers.

PR #361, opened 2022-01-12
Add openapi support to virtual workspaces

davidfestal:
Add Openapi support in virtual workspaces.

This implies:
- generating openapi v2 definitions for the KCP API, so that Workspace OpenAPIDefinitions for workspaces would be available for the `Workspace` virtual workspace.
- so cleaning a bit the KCP GO APIs to avoid kube-openapi generation errors
- logic to manage OpenAPI merging in the FixedGVs VirtualWorkspaces.
- adding e2e tests that check that a merged OpenAPI Spec is produced at the right virtual workspace path from all the resources of all the GVs.


hack/update-codegen-clients.sh
```diff
@@ -39,3 +39,12 @@ bash "${CODEGEN_PKG}"/generate-groups.sh "deepcopy,client,informer,lister" \
...
 
+go install "${CODEGEN_PKG}"/cmd/openapi-gen
+
+"$GOPATH"/bin/openapi-gen  --input-dirs github.com/kcp-dev/kcp/pkg/apis/cluster/v1alpha1 \
+--input-dirs github.com/kcp-dev/kcp/pkg/apis/apiresource/v1alpha1 \
+--input-dirs github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1 \
+--input-dirs github.com/kcp-dev/kcp/third_party/conditions/apis/conditions/v1alpha1 \
+--input-dirs k8s.io/apimachinery/pkg/apis/meta/v1,k8s.io/apimachinery/pkg/runtime,k8s.io/apimachinery/pkg/version \
+--output-package github.com/kcp-dev/kcp/pkg/openapi -O zz_generated.openapi \
+--go-header-file ./hack/../hack/boilerplate/boilerplate.go.txt --output-base /home/dfestal/go/src
```
sttts: /home/dfestal/go/src not found

pkg/virtual/framework/fixedgvs/register.go
```diff
@@ -36,14 +46,28 @@ func (vw *FixedGroupVersionsVirtualWorkspace) Register(rootAPIServerConfig gener
 			GenericConfig: &genericapiserver.RecommendedConfig{Config: *rootAPIServerConfig.Config, SharedInformerFactory: rootAPIServerConfig.SharedInformerFactory},
 			ExtraConfig: apiserver.ExtraConfig{
 				GroupVersion:    groupVersionAPISet.GroupVersion,
-				AddToScheme:     groupVersionAPISet.AddToScheme,
 				StorageBuilders: make(map[string]func(apiGroupAPIServerConfig genericapiserver.CompletedConfig) (restStorage.Storage, error)),
 			},
 		}
 		for resourceName, builder := range restStorageBuilders {
+			builder := builder
```
sttts, line 53: is this needed? The builder is a func, isn't it?

pkg/virtual/framework/fixedgvs/register.go
```diff
@@ -36,14 +46,28 @@ func (vw *FixedGroupVersionsVirtualWorkspace) Register(rootAPIServerConfig gener
...
 			cfg.ExtraConfig.StorageBuilders[resourceName] = builder
 		}
 
+		scheme := runtime.NewScheme()
+		if err := kubernetesscheme.AddToScheme(scheme); err != nil {
+			return nil, err
+		}
+		if err := groupVersionAPISet.AddToScheme(scheme); err != nil {
+			return nil, err
+		}
```
sttts, line 63: why do we need that here? Not sure I like this, a scheme that is shared more than it should be.

pkg/virtual/framework/rootapiserver/root_apiserver.go
```diff
@@ -161,7 +161,10 @@ func (c completedConfig) getRootHandlerChain(delegateAPIServer genericapiserver.
 			if accepted, prefixToStrip, context := c.resolveRootPaths(req.URL.Path, req.Context()); accepted {
 				req.URL.Path = strings.TrimPrefix(req.URL.Path, prefixToStrip)
 				req.URL.RawPath = strings.TrimPrefix(req.URL.RawPath, prefixToStrip)
-				req = req.WithContext(genericapirequest.WithCluster(context, genericapirequest.Cluster{Name: "virtual"}))
+				if req.URL.Path != "/openapi/v2" {
```
sttts, line 164: why?

hack/update-codegen-clients.sh
davidfestal: This should have been fixed by commit 212d233fbabd08864c6f30e65b52623002e20922.
Did you review commit-by-commit ?

pkg/virtual/framework/fixedgvs/register.go
davidfestal: You're right, it doesn't seem to be needed.

pkg/virtual/framework/fixedgvs/register.go
davidfestal: Not sure I understand you concern here. It's mainly doing the same thing as it used to do, but outside from (== before) the `New` function call, in order to be able to use this scheme in the `DefaultOpenAPIConfig` creation.

pkg/virtual/framework/rootapiserver/root_apiserver.go
davidfestal: Add a comment in the rebased commits

pkg/virtual/framework/fixedgvs/register.go
davidfestal: Fixed in rebased commits

sttts (approved):

PR #366, opened 2022-01-18
When a pull resource is added, be sure to add it only once

guillaumerose:
When sync’ing pods in an OpenShift cluster, kcp is confused by pods and pods.metrics. 
It imports both as CRDs but the sync fails later on. 

This change ensures only one CRD is loaded per name. 
Hopefully, core CRDs will always be loaded first.


pkg/crdpuller/discovery.go
```diff
@@ -142,6 +142,8 @@ func (sp *schemaPuller) PullCRDs(context context.Context, resourceNames ...strin
 				continue
 			}
 
+			resourcesToPull.Delete(groupResource.String(), apiResource.Name)
```
ncdc, line 145: Could you please add a comment why we're doing this here? Thanks!

stevekuznetsov:
Could we write a test for this? Would anything with a subresource break today?

pkg/crdpuller/discovery.go
guillaumerose: I added a comment, yes 👍

guillaumerose:
> Could we write a test for this? 

Yes, I can try. 

> Would anything with a subresource break today?

I don't know :/

ncdc:
@guillaumerose how are you invoking this (to configure the resources to pull)?

guillaumerose:
I do:

```
./kcp/bin/kcp start \
  --push-mode=true \
  --pull-mode=false \
  --install-cluster-controller \
  --install-workspace-controller \
  --auto-publish-apis \
  --resources-to-sync="deployments.apps,pods,services"
```
 
Usually I have a kind cluster but this time I wanted to use OpenShift.
The syncer failed with the error: `The following resource types were requested to be synced, but were not found in the KCP logical cluster: admin#$pods.metrics.k8s.io`

ncdc:
Thank you!

pkg/crdpuller/discovery.go
```diff
@@ -142,6 +142,9 @@ func (sp *schemaPuller) PullCRDs(context context.Context, resourceNames ...strin
 				continue
 			}
 
+			// pull only the first resource when there is a name conflict (example: pods can be pods or pods.metrics)
+			resourcesToPull.Delete(groupResource.String(), apiResource.Name)
```
ncdc, line 146: I don't think this is the correct fix for the issue. In your example, pods.metrics.k8s.io is a resource that is distinct from pods (from core/v1).

ncdc: One alternative approach is to use a RESTMapper to resolve a partially-specified resource name (e.g. "pods") to the match having the highest priority - see https://github.com/kcp-dev/kcp/pull/369/files?w=1 for an example (note, this is not an efficient implementation, but rather the quickest thing I could put together to demonstrate). Hope this helps!

guillaumerose: Yes it works, thanks. It is better like this. I updated this PR so that the pullAllResources flag is still running. 

If we follow this route, there is an other place where we need to fix this: https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/cluster/cluster.go#L95

I am not sure I should add a discovery client here and almost the same code. WDYT?

Also, what should be an efficient implementation? Something like using an informer for SchemaGroupVersion ?

ncdc:
cc @sttts for 👀 re RESTMapper usage

pkg/crdpuller/discovery.go
```diff
@@ -105,8 +106,23 @@ func (sp *schemaPuller) PullCRDs(context context.Context, resourceNames ...strin
 		return nil, err
 	}
 
+	groupResources, err := restmapper.GetAPIGroupResources(sp.discoveryClient)
+	if err != nil {
+		return nil, err
+	}
+	mapper := restmapper.NewDiscoveryRESTMapper(groupResources)
```
stevekuznetsov, line 113: Why not `restmapper.NewDeferredDiscoveryRESTMapper` so we don't make the entire discovery data pull part of the critical path here?

pkg/crdpuller/discovery.go
```diff
@@ -105,8 +106,23 @@ func (sp *schemaPuller) PullCRDs(context context.Context, resourceNames ...strin
...
+
 	pullAllResources := len(resourceNames) == 0
-	resourcesToPull := sets.NewString(resourceNames...)
+	resourcesToPull := sets.NewString()
+	for _, resourceToPull := range resourceNames {
+		gr := schema.ParseGroupResource(resourceToPull)
+		gvr, err := mapper.ResourceFor(gr.WithVersion(""))
```
stevekuznetsov, line 119: If we have versions in discovery, why will this be valid?

pkg/crdpuller/discovery.go
ncdc: SGTM

pkg/crdpuller/discovery.go
ncdc: RESTMappers perform matches against partial data. This is e.g. how kubectl works when you do `kubectl get po` or `kubectl get deployments` - it uses a RESTMapper to translate a partial group+resource specification (po=pods (.core), deployments = deployments.apps) to the highest priority/preferred GVR, as specified by the server, assuming it can find a match.

pkg/crdpuller/discovery.go
guillaumerose: done 👍

guillaumerose: Although, I am not sure it will be effective. Few lines after there is a call for ServerPreferredResources which triggers lot of api calls.

guillaumerose:
I added a unit test. It tests half of the method and at least what I want to fix.
Do you know a way to construct a openapi2.Document easily for testing purpose?

I had to create my own discovery type because ServerPreferredResources returns nil, nil in the fake package.

pkg/crdpuller/discovery.go
sttts: I think for the time being this is fine. It originates from the over-simplified flag `--resources-to-sync` which takes `pods`, `configmaps`, etc. In the future we will likely drive this through some API and then we will fully qualify the list of resources.

pkg/crdpuller/discovery.go
sttts: This is run every minute. Good enough for now.

sttts: +1 that this optimization does not help.

sttts:
Lgtm.

/assign @davidfestal

pkg/crdpuller/discovery.go
davidfestal: And I would add that in line [137](https://github.com/kcp-dev/kcp/pull/366/files#diff-2846812f7811365a5002800bbdde45978a15c09070f64b90eafcd5f72bb5e4d0R137) we only get the preferred version for each `GroupResource` by calling `ServerPreferredResources`.

davidfestal (approved):
Thanks for this fix @guillaumerose !

PR #379, opened 2022-01-25
test/e2e: use a build tag for e2e tests

stevekuznetsov:
With the new 'e2e' build tag, end-to-end tests do not compile by
default. This means that tools like `go test` and `golangci-lint` won't
run anything from those files by default, you need to ask for it. The
`make test-e2e` target does that for tests, and the `make lint` target
makes sure you do it for static analysis as well.

/assign @ncdc


ncdc (approved):

ncdc:
@sttts any objections?

PR #380, opened 2022-01-26
Map namespaces in kcp to unique names in pclusters

### LINKED ISSUE #283: Transparent Multicluster: Logical/Physical Cluster Namespaces
Use Case: Transparent Multicluster

Objectives:

- [ ] User can add a second location
- [ ] Deployed application moves between locations when service is disrupted (cluster down)
- [ ] Ingress follows application

Details

Namespaces should be addressable by a name human readable name even when conflicts may exists at the physical cluster level.  Map namespace requests between the name given in a workspace to the physical cluster's actual namespace name

kylape:
A namespace gets translated to a template that includes the kcp workspace name.

Example: syncer-test -> kcp--admin--syncer-test

This change required that the syncer start processing namespaces (the only non-namespaced resource currently processed by the syncer).  The `process` method was modified to handle both namespaces and namespaced
resources.

This includes an e2e test that creates a namespace and a custom resource in that namespace and validates that it gets sync'd to a downstream cluster using the translated namespace.

There is (at least on my local box) a mysterious ~60s delay in the namespace scheduler in reconciling and applying the correct label to the custom resource in the e2e test.

Fixes #283


pkg/syncer/syncer.go
```diff
@@ -299,6 +324,35 @@ func (c *Controller) handleErr(err error, i interface{}) {
...
 }
 
+const nsDelimiter = "--"
+const nsTemplate = "kcp" + nsDelimiter + "%s" + nsDelimiter + "%s"
+
+func transformForCluster(gvr schema.GroupVersionResource, meta metav1.Object) (string, string, error) {
+	if gvr.Group == "" && gvr.Resource == "namespaces" {
+		return fmt.Sprintf(nsTemplate, meta.GetClusterName(), meta.GetName()), "", nil
+	}
+	return meta.GetName(), fmt.Sprintf(nsTemplate, meta.GetClusterName(), meta.GetNamespace()), nil
```
stevekuznetsov, line 334: Namespaces have a length limit, we likely want to do something else here. Perhaps hash the inputs and prefix the output, if that gives us enough collision resistance?

pkg/syncer/syncer.go
```diff
@@ -299,6 +324,35 @@ func (c *Controller) handleErr(err error, i interface{}) {
...
+}
+
+func transformForKcp(gvr schema.GroupVersionResource, meta metav1.Object) (string, string, error) {
+	if gvr.Group == "" && gvr.Resource == "namespaces" {
+		segments := strings.Split(meta.GetName(), nsDelimiter)
```
stevekuznetsov, line 339: Can we encode this in a structured for in e.g. annotations?

test/e2e/syncer/wildwest.dev_cowboys.yaml
```diff
@@ -0,0 +1,58 @@
+
+---
```
stevekuznetsov, line 2: Why did this need to get copied?

kylape: It doesn't, I can refer to the under `reconciler/cluster`.

pkg/syncer/syncer.go
kylape: We need to be able to reverse the mapping so we can sync the status up to kcp.  So we can either do that by having a reversible transformation (like this PR) or by storing the mapping somewhere (an annotation?).

I was thinking we'd do the simplest thing for now and follow up with a better transformation strategy, but I am happy to do the better transformation in this PR.

test/e2e/syncer/wildwest.dev_cowboys.yaml
kylape: ah, well it looks like [I can't use `..` in a `go:embed` directive](https://pkg.go.dev/embed):

> Patterns may not contain ‘.’ or ‘..’ or empty path elements, nor may they begin or end with a slash.

test/e2e/syncer/spec_test.go
```diff
@@ -0,0 +1,254 @@
...
+					return sinkNsClient.Get(ctx, fmt.Sprintf("kcp--admin--%s", ns.Name), metav1.GetOptions{})
+				})
+				defer servers[sinkClusterName].Artifact(t, func() (runtime.Object, error) {
+					return sinkWildwestClient.Cowboys(fmt.Sprintf("kcp--admin--%s", cowboy.Name)).Get(ctx, cowboy.GetName(), metav1.GetOptions{})
+				})
+
+				if err := nsExpecter.ExpectBefore(ctx, func(ctx context.Context) (bool, error) {
+					sinkNsList, err := sinkNsLister.List(labels.Everything())
+					if err != nil {
+						return !apierrors.IsNotFound(err), err
```
ncdc, line 135: I don't believe you can get a NotFound error with a List call. I think you can safely return `true, err` because it's really hard (close to impossible) to get an error from a lister's List() call, and if we do get one, I doubt repeated attempts will yield different results.

test/e2e/syncer/spec_test.go
```diff
@@ -0,0 +1,254 @@
...
+					return false, fmt.Errorf("No namespace in sink cluster: %v", names)
+				}, 30*time.Second); err != nil {
+					t.Errorf("did not see namespace spec updated on sink cluster: %v", err)
+					return
+				}
+
+				if err := cowboyExpecter.ExpectBefore(ctx, func(ctx context.Context) (bool, error) {
+					sinkCowboyList, err := sinkCowboyLister.List(labels.Everything())
+					if err != nil {
+						return !apierrors.IsNotFound(err), err
```
ncdc, line 153: Ditto re NotFound

test/e2e/syncer/spec_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+					if len(sinkCowboyList) < 1 {
+						return false, fmt.Errorf("No cowboys found in target ns in sink cluster")
+					} else if len(sinkCowboyList) > 1 {
+						return false, fmt.Errorf("Too many cowboys found in target ns in sink cluster: %d", len(sinkCowboyList))
+					}
+
+					return true, nil
+				}, 30*time.Second); err != nil {
+					t.Errorf("did not see namespace spec updated on sink cluster: %v", err)
```
ncdc, line 164: Is this error message correct?

pkg/syncer/specsyncer.go
```diff
@@ -77,58 +76,21 @@ func NewSpecSyncer(from, to *rest.Config, syncedResourceTypes []string, clusterI
...
-	}
-	return nil
+	return New(fromDiscovery, fromClient, toClient, KcpToPcluster, syncedResourceTypes, clusterID)
 }
 
 func deleteFromDownstream(c *Controller, ctx context.Context, gvr schema.GroupVersionResource, namespace, name string) error {
 	// TODO: get UID of just-deleted object and pass it as a precondition on this delete.
 	// This would avoid races where an object is deleted and another object with the same name is created immediately after.
 
+	// TODO: check to see if ns is deleting/deleted
```
ncdc, line 86: Is this something you need to resolve in this PR? What is the effect here if a namespace is being deleted?

ncdc (commented):
I still have a few files left to review but it's my EOD & I wanted to submit what I currently have. Will pick it back up tomorrow.

pkg/syncer/syncer.go
```diff
@@ -299,6 +324,55 @@ func (c *Controller) handleErr(err error, i interface{}) {
 	klog.Errorf("Dropping key %q after failed retries: %v", i, err)
 }
 
+const nsDelimiter = "--"
+const nsTemplate = "kcp" + nsDelimiter + "%s" + nsDelimiter + "%s"
```
stevekuznetsov, line 328: Yeah I don't think it would be that hard to store the data in an annotation and write something that would not overflow the namespace name length limit. I don't really think the slightly higher coding cost is worth having a system where if you have a long enough cluster/namespace name, something panics somewhere or the system falls over.

pkg/syncer/syncer.go
```diff
@@ -41,6 +41,11 @@ import (
 const resyncPeriod = 10 * time.Hour
 const SyncerNamespaceKey = "SYNCER_NAMESPACE"
 
+type Direction string
```
ncdc, line 44: Because this is an exported type, please add godoc for it.

pkg/syncer/syncer.go
```diff
@@ -41,6 +41,11 @@ import (
...
+
+const KcpToPcluster Direction = "kcpToPcluster"
```
ncdc, line 46: Please add godoc.

I would also consider naming the constant KCPToPhysicalCluster.

pkg/syncer/syncer.go
```diff
@@ -41,6 +41,11 @@ import (
...
+const PclusterToKcp Direction = "pclusterToKcp"
```
ncdc, line 47: Please add godoc.

I would also consider naming the constant PhysicalClusterToKCP.

pkg/syncer/syncer.go
```diff
@@ -316,12 +390,28 @@ func (c *Controller) process(gvr schema.GroupVersionResource, obj interface{}) e
...
+	var err error
+	if c.direction == KcpToPcluster {
+		targetName, targetNamespace, err = transformForCluster(gvr, meta)
+	} else {
+		targetName, targetNamespace, err = transformForKcp(gvr, meta)
+	}
+
+	if err != nil {
+		klog.Error(err)
+		return err
```
ncdc, line 412: If you return an error here, the controller logic will requeue the same key (with backoff). Presumably we'd encounter the same error again. It's probably better to return nil here to avoid that.

test/e2e/syncer/spec_test.go
```diff
@@ -0,0 +1,254 @@
...
+						return !apierrors.IsNotFound(err), err
+					}
+					var names []string
+					for _, ns := range sinkNsList {
+						names = append(names, ns.Name)
+						if ns.Name == expectedSinkNsName {
+							return true, nil
+						}
+					}
+					return false, fmt.Errorf("No namespace in sink cluster: %v", names)
```
marun, line 144: I was initially wondering why it made sense to use a list instead of a get, and the answer appears to be so that this error message can contain a list of all namespaces in the target cluster. Why is that useful in this case vs indicating something like 'namespace %s not present'?

test/e2e/syncer/spec_test.go
```diff
@@ -0,0 +1,254 @@
...
+	}{
+		{
+			name: "create a resource in new namespace, expect ns name to contain lcluster + ns name and for resource to get synced to correct ns",
+			work: func(ctx context.Context, t framework.TestingTInterface, servers map[string]runningServer) {
+				const testNamespace = "syncer-test"
+				expectedSinkNsName := fmt.Sprintf("kcp--admin--%s", testNamespace)
+
+				sourceNsClient := servers[sourceClusterName].k8sClientSet.CoreV1().Namespaces()
+				sourceCowboyClient := servers[sourceClusterName].wildwestClientSet.WildwestV1alpha1().Cowboys(testNamespace)
+				sinkNsClient := servers[sinkClusterName].k8sClientSet.CoreV1().Namespaces()
```
marun, line 73: (No action required) What is the motivation for the source/sink var naming? vs something like source/dest or <other>? sink makes sense to me in the context of event producer/consumer but resource syncing between pcluster and kcp is necessarily bidirectional (spec one way, status another. maybe metadata both ways).

cc: @stevekuznetsov

test/e2e/syncer/spec_test.go
```diff
@@ -0,0 +1,254 @@
...
+
+				if err := cowboyExpecter.ExpectBefore(ctx, func(ctx context.Context) (bool, error) {
+					sinkCowboyList, err := sinkCowboyLister.List(labels.Everything())
+					if err != nil {
+						return !apierrors.IsNotFound(err), err
+					}
+
+					if len(sinkCowboyList) < 1 {
+						return false, fmt.Errorf("No cowboys found in target ns in sink cluster")
+					} else if len(sinkCowboyList) > 1 {
```
marun, line 158: What could result in multiple resources being created?

test/e2e/syncer/spec_test.go
```diff
@@ -0,0 +1,254 @@
...
+		testCase := testCases[i]
+		framework.Run(t, testCase.name, func(t framework.TestingTInterface, servers map[string]framework.RunningServer, artifactDir, dataDir string) {
+			start := time.Now()
+			ctx := context.Background()
+			if deadline, ok := t.Deadline(); ok {
+				withDeadline, cancel := context.WithDeadline(ctx, deadline)
+				t.Cleanup(cancel)
+				ctx = withDeadline
+			}
+			if len(servers) != 2 {
```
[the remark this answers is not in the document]
marun: Is this check unnecessary? The server configurations are hard-coded as parameters to the same `Run` invocation that the enclosing function is provided to.

test/e2e/syncer/spec_test.go
```diff
@@ -0,0 +1,254 @@
...
+					t.Errorf("did not see namespace spec updated on sink cluster: %v", err)
+					return
+				}
+			},
+		},
+	}
+	for i := range testCases {
+		testCase := testCases[i]
+		framework.Run(t, testCase.name, func(t framework.TestingTInterface, servers map[string]framework.RunningServer, artifactDir, dataDir string) {
+			start := time.Now()
```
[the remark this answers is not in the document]
marun: Was this line added before `InstallCRD` but not removed from here?

test/e2e/syncer/spec_test.go
```diff
@@ -0,0 +1,254 @@
...
+				if err := nsExpecter.ExpectBefore(ctx, func(ctx context.Context) (bool, error) {
+					sinkNsList, err := sinkNsLister.List(labels.Everything())
+					if err != nil {
+						return !apierrors.IsNotFound(err), err
+					}
+					var names []string
+					for _, ns := range sinkNsList {
+						names = append(names, ns.Name)
+						if ns.Name == expectedSinkNsName {
+							return true, nil
```
marun, line 141: (No action required) Maybe log that the namespace was successfully synced to the pcluster?

test/e2e/syncer/spec_test.go
```diff
@@ -0,0 +1,254 @@
...
+						}
+					}
+					return false, fmt.Errorf("No namespace in sink cluster: %v", names)
+				}, 30*time.Second); err != nil {
+					t.Errorf("did not see namespace spec updated on sink cluster: %v", err)
+					return
+				}
+
+				if err := cowboyExpecter.ExpectBefore(ctx, func(ctx context.Context) (bool, error) {
```
marun, line 150: The change proposed in this PR involves the following:

1. transform `metadata.name` of synced namespaces
1. transform `metadata.namespace` of synced resources

Maybe the second change suggests checking that the cowboy resource has the expected transformed namespace?

pkg/syncer/syncer.go
```diff
@@ -299,6 +324,55 @@ func (c *Controller) handleErr(err error, i interface{}) {
...
+const nsDelimiter = "--"
+const nsTemplate = "kcp" + nsDelimiter + "%s" + nsDelimiter + "%s"
+
+type nameTransformer struct {
+	clusterName string
+	namespace   string
+}
+
+func (n *nameTransformer) toCluster() string {
+	return fmt.Sprintf(nsTemplate, n.clusterName, n.namespace)
```
marun, line 336: (No action required) Why is the cluster name what we want to encode, vs org-workspace-namespace? Does cluster name map to an org+workspace pair?

cc: @ncdc

test/e2e/syncer/spec_test.go
```diff
@@ -0,0 +1,254 @@
...
+				coreSharedInformerFactory := coreexternalversions.NewSharedInformerFactoryWithOptions(servers[sinkClusterName].k8sClientSet, 0)
+				nsInformer := coreSharedInformerFactory.Core().V1().Namespaces()
+				nsExpecter := framework.NewExpecter(nsInformer.Informer())
+				sinkNsLister := nsInformer.Lister()
+				coreSharedInformerFactory.Start(ctx.Done())
+
+				wildwestSharedInformerFactory := wildwestexternalversions.NewSharedInformerFactoryWithOptions(servers[sinkClusterName].wildwestClientSet, 0)
+				cowboyInformer := wildwestSharedInformerFactory.Wildwest().V1alpha1().Cowboys()
+				cowboyExpecter := framework.NewExpecter(cowboyInformer.Informer())
+				sinkCowboyLister := cowboyInformer.Lister().Cowboys("")
```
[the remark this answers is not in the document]
kylape: I honestly don't know why I'm using `""`.  I'm gonna change it to `expectedSinkNsName`.

test/e2e/syncer/spec_test.go
kylape: nope

pkg/syncer/syncer.go
kylape: ok cool, I'll do that.

pkg/server/config.go
```diff
@@ -45,6 +46,7 @@ func DefaultConfig() *Config {
 		ShardKubeconfigFile:        "",
 		EnableSharding:             false,
 		Authentication:             kubeoptions.NewBuiltInAuthenticationOptions().WithAll(),
+		DiscoveryPollInterval:      time.Duration(60) * time.Second,
```
imjasonh, line 49: nit: You can just do `60*time.Second`

pkg/syncer/syncer.go
```diff
@@ -299,6 +324,55 @@ func (c *Controller) handleErr(err error, i interface{}) {
...
+		return nil, fmt.Errorf("Invalid name: %s", namespace)
+	}
+
+	return &nameTransformer{
+		clusterName: segments[1],
+		namespace:   segments[2],
+	}, nil
+}
+
+func transformForCluster(gvr schema.GroupVersionResource, meta metav1.Object) (string, string, error) {
```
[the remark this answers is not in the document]
imjasonh: I'm a bit worried about making that a load-bearing assumption, in case we add another cluster-scoped resource in the future (not sure what, but you never know), and this subtly stops being true.

pkg/syncer/syncer.go
imjasonh: +1, if you do that, consider logging the error with an indication that we're dropping as a result. It's tough to tell from logs sometimes whether an error means "an error! (don't worry, I'll retry)" vs "an error! (you'll never see this again)"

test/e2e/syncer/spec_test.go
kylape: The original reason is that I copy/pasted from a different e2e test, and that test was listing.  But I will say that having the list of existing namespaces did help me debug the test.  I should probably at least add the expected name to the log message.

test/e2e/syncer/spec_test.go
kylape: I can't think of anything, probably another thing that was copy/pasted. maybe it could be simplified to

```suggestion
					} else if len(sinkCowboyList) != 1 {
					    return false, fmt.Errorf("Expected 1 cowboy, found %d", len(sinkCowboyList))
```

test/e2e/syncer/spec_test.go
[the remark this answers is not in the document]
kylape: It's not _really_ necessary, but I could see it helping if you shoot yourself in the foot by messing with the servers down the road.  Admittedly not a likely scenario.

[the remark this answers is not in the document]
marun: I would err on the side of not checking an error if it could be easily discoverable otherwise.

pkg/syncer/syncer.go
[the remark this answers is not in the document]
marun: On reflection, I realize that my original comment focused too much on how namespaces should be treated specially. An empty namespace signifies a cluster-scoped resource, and keying on that instead of gvr would provide the desired name transformation for all cluster-scoped resources (not just namespaces).

test/e2e/syncer/spec_test.go
marun: I can see the list of existing namespaces helping if the intent is to validate more than just the creation of the namespace. But if the goal is validating the metadata transformation, maybe that could tested with less cost in a unit test?

test/e2e/syncer/spec_test.go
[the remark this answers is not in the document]
kylape: `start` is set multiple times to track several independent operations: Installing the CRDs, installing the `Cluster`, and setting up clients.

test/e2e/syncer/spec_test.go
kylape: Yes this was the intention, but @ncdc pointed out that I wasn't setting the correct namespace when creating the lister used in this expecter.  So I'm going to update the lister to use the target namespace (on the pcluster).

pkg/syncer/syncer.go
kylape: `clusterName` here refers to the logical cluster, and workspaces are a type of logical cluster.  My understanding of orgs is that they'll be another logical cluster that contains `Workspace` resources but that this hierarchy doesn't really matter to a syncer.  A logical cluster/workspace name should be unique enough without the org included.  The only way I think this wouldn't be true is if workspace names are not unique between shards.

kylape:
I've addressed all of the very simple fixes and marked those conversations as resolved.  I'll make another pass at the slightly-less-trivial suggestions.

test/e2e/syncer/wildwest.dev_cowboys.yaml
kylape: yes that makes sense.  Mind if I make this a follow up issue? I'm thinking we should move the CRD file and gen'd packages up under something like `test/e2e/crds`, update `hack/update-codegen.sh`, and expose some kind of interface for pulling in CRDs into a test.

pkg/syncer/syncer.go
kylape: I saw the slack thread about this, so please disregard my comment :smile:

test/e2e/syncer/spec_test.go
stevekuznetsov: Most of them are doing a Get, not a List

test/e2e/syncer/spec_test.go
stevekuznetsov: If you want to have the list of namespaces after the test you can also certainly add them as an artifact

test/e2e/syncer/spec_test.go
stevekuznetsov: I think this was how the syncer code labeled these concepts? I created the original test as a port of a shell script

marun:
LGTM

test/e2e/syncer/spec_test.go
marun: You are correct, my statement of 'all invocations' was in error. Still, the pattern appears to be metastasizing outside of the expect framework:

Get:
 - https://github.com/kcp-dev/kcp/blob/main/test/e2e/framework/expect.go#L238
 - https://github.com/kcp-dev/kcp/blob/main/test/e2e/framework/expect.go#L290

List:
 - https://github.com/kcp-dev/kcp/blob/main/test/e2e/reconciler/cluster/controller_test.go#L255
 - https://github.com/kcp-dev/kcp/blob/main/test/e2e/reconciler/ingress/controller_test.go#L343
 - https://github.com/kcp-dev/kcp/blob/main/test/e2e/framework/expect.go#L259
 - + the one in this PR

pkg/syncer/specsyncer.go
```diff
@@ -89,6 +89,34 @@ func deleteFromDownstream(c *Controller, ctx context.Context, gvr schema.GroupVe
 }
 
 func upsertIntoDownstream(c *Controller, ctx context.Context, gvr schema.GroupVersionResource, namespace string, unstrob *unstructured.Unstructured) error {
+	// See if we need to create the target namespace
```
imjasonh, line 92: (not blocking) over time this should all move up into the kcp layer to reduce the permissions the in-cluster syncer has to have: https://github.com/kcp-dev/kcp/issues/387

imjasonh (approved):

pkg/syncer/specsyncer.go
```diff
@@ -77,58 +76,49 @@ func NewSpecSyncer(from, to *rest.Config, syncedResourceTypes []string, clusterI
...
-			},
-			DeleteFunc: func(obj interface{}) { c.AddToQueue(gvr, obj) },
-		}
-	}, syncedResourceTypes, clusterID)
-}
-
-// TODO:
-// This function is there as a quick and dirty implementation of namespace creation.
-// In fact We should also be getting notifications about namespaces created upstream and be creating downstream equivalents.
-func (c *Controller) ensureNamespaceExists(namespace string) error {
```
marun, line 96: @ncdc Maybe just revert this removal vs reimplementing?

ncdc: K

ncdc:
I'm starting to wonder if the test failures are legit and not flakes

kylape:
I agree and am working through the test failures locally.

test/e2e/syncer/wildwest.dev_cowboys.yaml
kylape: This is no longer needed since I removed the syncer e2e tests.  The cluster controller e2e tests were already testing everything the syncer e2e tests were.

pkg/syncer/specsyncer.go
kylape: This comment is removed now, but I did file an issue around namespace deletion: https://github.com/kcp-dev/kcp/issues/398

pkg/reconciler/namespace/namespace.go
```diff
@@ -38,6 +38,12 @@ const clusterLabel = "kcp.dev/cluster"
 // reconcileResource is responsible for setting the cluster for a resource of
 // any type, to match the cluster where its namespace is assigned.
 func (c *Controller) reconcileResource(ctx context.Context, lclusterName string, unstr *unstructured.Unstructured, gvr *schema.GroupVersionResource) error {
+
+	if gvr.Group == "networking.k8s.io" && gvr.Version == "v1" && gvr.Resource == "ingresses" {
```
ncdc, line 42:
```suggestion
	if gvr.Group == "networking.k8s.io" && gvr.Resource == "ingresses" {
```

ncdc: It's safer to avoid checking the version

pkg/syncer/specsyncer.go
```diff
@@ -77,58 +76,65 @@ func NewSpecSyncer(from, to *rest.Config, syncedResourceTypes []string, clusterI
...
 		return err
 	}
 
 	client := c.getClient(gvr, namespace)
 
-	unstrob = unstrob.DeepCopy()
-
 	// Attempt to create the object; if the object already exists, update it.
 	unstrob.SetUID("")
 	unstrob.SetResourceVersion("")
```
sttts, line 140: is this the only change we do to the object when copying from up to downstream?

It's not enough. We have to clean much more. Finalizers come to mind, but also annotations that control aspects. IMO we should strip all annotations, all owner refs too. Maybe more. cc @marun

sttts: Also note line 160. We cannot do that. An OpenShift will use annotation, maybe other actors as well. We cannot just wipe all changes. That will also hotloop.

pkg/syncer/syncer.go
```diff
@@ -41,6 +41,15 @@ import (
 const resyncPeriod = 10 * time.Hour
 const SyncerNamespaceKey = "SYNCER_NAMESPACE"
 
+// Direction indicates which direction data is flowing for this particular syncer
+type Direction string
```
sttts, line 45: why is this a string and not just a `iota` based enumeration?

pkg/syncer/syncer.go
```diff
@@ -299,6 +328,46 @@ func (c *Controller) handleErr(err error, i interface{}) {
...
+
+func (n *nameTransformer) toCluster() string {
+	return fmt.Sprintf(nsTemplate, n.clusterName, n.namespace)
+}
+
+func toKcp(namespace string) (*nameTransformer, error) {
+	segments := strings.Split(namespace, nsDelimiter)
+
+	if len(segments) != 3 {
+		return nil, fmt.Errorf("Failed to transform name: %s", namespace)
```
sttts, line 347: always lower-case errors

pkg/syncer/syncer.go
```diff
@@ -299,6 +328,46 @@ func (c *Controller) handleErr(err error, i interface{}) {
 	klog.Errorf("Dropping key %q after failed retries: %v", i, err)
 }
 
+const nsDelimiter = "--"
+const nsTemplate = "kcp" + nsDelimiter + "%s" + nsDelimiter + "%s"
+
+type nameTransformer struct {
+	clusterName string
+	namespace   string
+}
```
sttts, line 337: where is the use of this struct? Just have the two funcs directly. There is no added value.

pkg/syncer/syncer.go
```diff
@@ -299,6 +328,46 @@ func (c *Controller) handleErr(err error, i interface{}) {
...
+	}
+
+	return meta.GetName(), t.toCluster(), nil
+}
+
+func transformForKcp(gvr schema.GroupVersionResource, meta metav1.Object) (string, string, error) {
+	name := meta.GetNamespace()
+	transformer, err := toKcp(name)
+	return meta.GetName(), transformer.namespace, err
+}
```
sttts, line 369: gvr is not used.

And I don't see the added value of the func.

pkg/syncer/syncer.go
```diff
@@ -316,12 +385,33 @@ func (c *Controller) process(gvr schema.GroupVersionResource, obj interface{}) e
 			return err
 		}
 	}
-	namespace, name := meta.GetNamespace(), meta.GetName()
-	if c.inSyncerNamespace(namespace) {
+
+	namespace := meta.GetNamespace()
+
+	if namespace == "" {
+		klog.V(2).Infof("Skipping cluster-level resource: %T : %v", obj, obj)
```
sttts, line 392: no need to log this

pkg/syncer/syncer.go
```diff
@@ -316,12 +385,33 @@ func (c *Controller) process(gvr schema.GroupVersionResource, obj interface{}) e
...
+		return nil
+	}
+
+	// This must be checked before the namespace is transformed
+	if c.inSyncerNamespace(meta) {
 		klog.V(2).Infof("Skipping object of type: %T : %v in syncer namespace", obj, obj)
```
sttts, line 398: no need to log this at this level.

pkg/syncer/syncer.go
```diff
@@ -358,15 +450,19 @@ func (c *Controller) getClient(gvr schema.GroupVersionResource, namespace string
 	return nri
 }
 
-func (c *Controller) inSyncerNamespace(objectNamespace string) bool {
+func (c *Controller) inSyncerNamespace(meta metav1.Object) bool {
```
sttts, line 453: why this change? It makes the signature more complex. Leave it as it is.

pkg/syncer/syncer.go
```diff
@@ -358,15 +450,19 @@ func (c *Controller) getClient(gvr schema.GroupVersionResource, namespace string
...
 	if c.namespace == "" {
 		return false
 	}
-	if c.namespace == objectNamespace {
-		return true
+	var err error
+	objectNamespace := meta.GetNamespace()
+
+	if c.direction == KcpToPhysicalCluster {
+		_, objectNamespace, err = transformForCluster(schema.GroupVersionResource{}, meta)
```
sttts, line 464: this block is odd. The whole func only makes sense in the `KcpToPhysicalCluster` case. Don't make it generic if it isn't. Do the logic where it is called, not deep in no-op funcs.

sttts (changes requested):
The deepcopy removals are a blocker. We cannot merge this.

pkg/syncer/specsyncer.go
```diff
@@ -77,58 +76,65 @@ func NewSpecSyncer(from, to *rest.Config, syncedResourceTypes []string, clusterI
...
+		// Already exists - all good
+		return nil
+	}
+
+	if !k8serrors.IsNotFound(err) {
+		// Got some error besides not-found - stop & return.
+		// TODO bubble this up as a condition somewhere.
+		klog.Errorf("Error checking if namespace %q exists: %v", namespace, err)
+		return err
+	}
```
sttts, line 111: all this is not needed. Create is enough.

sttts:
Comments addressed in follow-up https://github.com/kcp-dev/kcp/pull/405. Merging.

sttts (approved):

PR #383, opened 2022-01-27
Make OpenAPIV3Schema atomic

ncdc:
Set the x-kubernetes-map-type to atomic for the OpenAPIV3Schema field in
apiresourceimports and negotiatedapiresources. This causes server-side
apply to treat the schema as an atomic unit, instead of trying to track
each individual field/element of the schema (which causes 34-second
timeouts in the apiserver trying to process the data).

### DIFF
--- a/config/apiresource.kcp.dev_apiresourceimports.yaml
+++ b/config/apiresource.kcp.dev_apiresourceimports.yaml
@@ -139,6 +139,7 @@ spec:
                 type: string
               openAPIV3Schema:
                 type: object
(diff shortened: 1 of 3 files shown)


davidfestal (approved):
SGTM

sttts:
For reference: https://github.com/kubernetes/kubernetes/issues/102749. This was supposed to be fixed in 1.21, but seem to have regressed in 1.23 (or was not completely fixed). Upstream is working on it. But for our use-case `atomic` is more correct anyway.

PR #384, opened 2022-01-27
Better support for running the demos on Mac and with podman 3.x

ncdc:
Don't pass --posix to sed (breaks on Mac's BSD-based sed).

For podman 3.x, work around an issue where podman isn't able to
correctly port forward to a service bound to localhost in the podman vm
(see podman issue 11528). The workaround is to manually set up SSH port
forwarding.

### DIFF
--- a/contrib/demo/clusters/kind/createKindClusters.sh
+++ b/contrib/demo/clusters/kind/createKindClusters.sh
@@ -15,11 +15,76 @@
 # limitations under the License.
 
 set -o errexit
+set -o nounset
+set -o pipefail
+set -o xtrace
 
-DEMO_ROOT="$(dirname "${BASH_SOURCE}")/../.."
+if [[ -z "${CLUSTERS_DIR:-}" ]]; then
+    echo CLUSTERS_DIR is required
+    eixt 1
+fi
 
-kind create cluster --config ${DEMO_ROOT}/clusters/kind/us-east1.config --kubeconfig ${DEMO_ROOT}/clusters/kind/us-east1.kubeconfig
-sed -e 's/^/    /' ${DEMO_ROOT}/clusters/kind/us-east1.kubeconfig | cat ${DEMO_ROOT}/clusters/us-east1.yaml - > ${DEMO_ROOT}/clusters/kind/us-east1.yaml
+DEMO_ROOT="$(dirname "${BASH_SOURCE[0]}")/../.."
 
-kind create cluster --config ${DEMO_ROOT}/clusters/kind/us-west1.config --kubeconfig ${DEMO_ROOT}/clusters/kind/us-west1.kubeconfig
-sed -e 's/^/    /' ${DEMO_ROOT}/clusters/kind/us-west1.kubeconfig | cat ${DEMO_ROOT}/clusters/us-west1.yaml - > ${DEMO_ROOT}/clusters/kind/us-west1.yaml
+get_cluster_port() {
+    local cluster=kind-$1
+    local clusterPort
+    clusterPort=$(kubectl --kubeconfig "${CLUSTERS_DIR}/$1.kubeconfig" config view -o json | jq -r ".clusters[] | select(.name==\"${cluster}\").cluster.server" | cut -d ':' -f 3)
+    echo "${clusterPort}"
+}
+
+podman_ssh() {
+    local podmanMachine
+    podmanMachine="${PODMAN_MACHINE:-podman-machine-default*}"
+
+    local sshURL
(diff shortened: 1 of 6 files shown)


contrib/demo/runDemoScripts.sh
```diff
@@ -42,7 +46,7 @@ source ${DEMOS_DIR}/.startUtils
 setupTraps $0
 
 cleanupDemoOutput() {
-  sed -E --posix -e 's/\x1b\[(3J|H|2J)//g' -e '/^☸️ \$ /d' -e '/^ +certificate-authority-data: /d' -e '/^ +server: https:\/\/127\.0\.0\.1:[0-9]{5}$/d' -e 's/[0-9]+(s|m)/10s/g' -e 's/^(.*)( +[0-9]+s|[0-9]+m([0-9]+s)?)$/\1/' $1
+  sed -E -e 's/\x1b\[(3J|H|2J)//g' -e '/^☸️ \$ /d' -e '/^ +certificate-authority-data: /d' -e '/^ +server: https:\/\/127\.0\.0\.1:[0-9]{5}$/d' -e 's/[0-9]+(s|m)/10s/g' -e 's/^(.*)( +[0-9]+s|[0-9]+m([0-9]+s)?)$/\1/' $1
```
ncdc, line 49: Will need someone on Linux to verify the removal of `--posix` doesn't break things here.

stevekuznetsov:
Crazy thought: why is this all (fragile) shell? Is there no cross-platform demo software we can use that's going to give us better returns than what at least to me seems like a more home-grown shell library?

ncdc:
Are you referring to runDemoScripts.sh + everything that goes with it, or just my podman hacks in this PR?

davidfestal:
> Crazy thought: why is this all (fragile) shell? Is there no cross-platform demo software we can use that's going to give us better returns than what at least to me seems like a more home-grown shell library?

If this question is related to the whole demo thing: quick answer is probably "history". Because it has been based on, and extended from, the very first demo prepared for the first kubecon, and we never had time, according to priorities, to invest into refactoring the whole more that just restructuring it step-by-step when needed, as it has been done until now.

Not sure it's really worth investing more time in it now since for real CI testing we now have e2e tests.

However IMHO it still seems quite useful to be able to easily setup demos that can be either interactive or not, and whose main script file is quite self-explanatory about the CLI commands used.

stevekuznetsov:
I'm not saying this should be "testable" but just - and, I am ignorant of this, so please correct me - but my intuition would say that surely there is something a bit more structured and feature-rich specifically for writing reproducible demos in a cross-platform manner, which would perhaps also help with the recording?

contrib/demo/clusters/kind/createKindClusters.sh
```diff
@@ -15,11 +15,56 @@
...
+}
 
-kind create cluster --config ${DEMO_ROOT}/clusters/kind/us-west1.config --kubeconfig ${DEMO_ROOT}/clusters/kind/us-west1.kubeconfig
-sed -e 's/^/    /' ${DEMO_ROOT}/clusters/kind/us-west1.kubeconfig | cat ${DEMO_ROOT}/clusters/us-west1.yaml - > ${DEMO_ROOT}/clusters/kind/us-west1.yaml
+podman_ssh() {
+    local podmanMachine
+    podmanMachine="${PODMAN_MACHINE:-podman-machine-default*}"
+
+    local sshURL
+    sshURL=$(podman system connection ls --format=json | jq -r ".[]|select(.Name==\"${podmanMachine}\").URI" | grep -E -o 'ssh://[^:]+:\d+')
```
davidfestal, line 36: On my RHEL linux, `sshURL` is empty, obviously because `podman system connection ls` is also empty.
So the script stops here, and I cannot go further in the tests.

Do we miss some condition here, in order to fallback to the previous behavior ?

davidfestal (changes requested):

contrib/demo/clusters/kind/createKindClusters.sh
ncdc: Oh, so you have both podman & docker installed?

davidfestal: Seems yes :-)

ncdc: K, will fix

davidfestal: This works now.

contrib/demo/runDemoScripts.sh
davidfestal: tested it and it doesn't break anything.

PR #385, opened 2022-01-28
server: use composable options-config-server plumbing of genericcontrolplane

sttts:
/hold

Needs https://github.com/kcp-dev/kubernetes/pull/35.

PR #390, opened 2022-01-28
Consolidate shared informer factories to 1 each

### LINKED ISSUE #314: Consolidate shared informer factory creation
We have a few places where we duplicate starting shared informer factories in server.go. Let's truly consolidate their creation!

Happy to get started on this now, or wait until after #202 and #302 and any other open PRs that create factories are merged.

ncdc:
Fixes #314

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -65,7 +65,6 @@ import (
 	"k8s.io/apiserver/pkg/util/webhook"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
-	"k8s.io/client-go/informers"
 	coreexternalversions "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/metadata"
@@ -116,6 +115,10 @@ type Server struct {
 	preShutdownHooks []preShutdownHookEntry
 
 	crdServerReady chan struct{}
+
+	kcpSharedInformerFactory           kcpexternalversions.SharedInformerFactory
+	kubeSharedInformerFactory          coreexternalversions.SharedInformerFactory
+	apiextensionsSharedInformerFactory apiextensionsexternalversions.SharedInformerFactory
 }
 
 // postStartHookEntry groups a PostStartHookFunc with a name. We're not storing these hooks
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -277,28 +280,52 @@ func (s *Server) Run(ctx context.Context) error {
...
+
+	s.AddPostStartHook("start-informers", func(context genericapiserver.PostStartHookContext) error {
 		if err := s.waitForCRDServer(context.StopCh); err != nil {
 			return err
 		}
-		kcpSharedInformerFactory.Start(context.StopCh)
-		kcpSharedInformerFactory.WaitForCacheSync(context.StopCh)
+
+		s.kcpSharedInformerFactory.Start(context.StopCh)
+		s.kcpSharedInformerFactory.WaitForCacheSync(context.StopCh)
```
marun, line 315: Maybe wait for cache sync only after starting all informer factories to minimize startup time?

pkg/server/server.go
```diff
@@ -335,6 +362,7 @@ func (s *Server) Run(ctx context.Context) error {
 	}
 
 	apiExtensionsConfig.ExtraConfig.NewInformerFactoryFunc = func(client apiextensionsclient.Interface, resyncPeriod time.Duration) apiextensionsexternalversions.SharedInformerFactory {
+		// TODO could we use s.apiextensionsSharedInformerFactory (ignoring client & resyncPeriod) instead of creating a 2nd factory here?
```
marun, line 365: (No action required) Seems like a win to not have duplicative informer caches in the same process. What would be the drawback?

pkg/server/server.go
ncdc: Is this what you're suggesting?

```go
s.kcpSharedInformerFactory.Start(context.StopCh)
		s.kubeSharedInformerFactory.Start(context.StopCh)
		s.apiextensionsSharedInformerFactory.Start(context.StopCh)
		
		s.kcpSharedInformerFactory.WaitForCacheSync(context.StopCh)
		s.kubeSharedInformerFactory.WaitForCacheSync(context.StopCh)
		s.apiextensionsSharedInformerFactory.WaitForCacheSync(context.StopCh)
```

pkg/server/server.go
ncdc: The `NewInformerFactoryFunc` was something I added in our Kube fork so we could override the shared informer factory the apiextensions apiserver uses (so we can do workspace API inheritance). At the time, I had to do it this way because there was no way for me to inject the shared informer factory (or the client) before the server chain was started. But now that #385 is in, we can probably switch to passing in a client & shared informer factory and get rid of `NewInformerFactoryFunc`.

cc @sttts WDYT?

pkg/server/server.go
marun: Yes, that would ensue the startup time was bound by the longest wait duration vs the summation of all wait durations.

pkg/server/server.go
ncdc: Hmm looks like I'll need to do a small PR to our Kube fork to be able to pass in a client, but I could do that

ncdc: Good candidate for a follow-up PR (as it'll require a Kube PR+merge, then bumping Kube in kcp go.mod)

pkg/server/server.go
ncdc: Updated!

marun:
LGTM

pkg/server/server.go
```diff
@@ -277,28 +280,51 @@ func (s *Server) Run(ctx context.Context) error {
...
 			return err
 		}
-		kcpSharedInformerFactory.Start(context.StopCh)
-		kcpSharedInformerFactory.WaitForCacheSync(context.StopCh)
+
+		s.kcpSharedInformerFactory.Start(context.StopCh)
+		s.kubeSharedInformerFactory.Start(context.StopCh)
+		s.apiextensionsSharedInformerFactory.Start(context.StopCh)
+
+		s.kcpSharedInformerFactory.WaitForCacheSync(context.StopCh)
```
sttts, line 318: handle errors. This is not enough.

sttts: Or do we know these run indefinitely?

ncdc: They should go indefinitely. They call cache.WaitForCacheSync underneath

sttts:
Remaining work as follow-up.

sttts (approved):

PR #394, opened 2022-01-31
Cluster conditions and refactor

jmprusi:
- Using the third_party conditions for Clusters. 
- Added new Cluster conditions
- Added an optional "LastHeartbeatTime" for handling health checks in pclusters


third_party/conditions/apis/conditions/v1alpha1/types.go
```diff
@@ -67,6 +67,11 @@ type Condition struct {
 	// +optional
 	Severity ConditionSeverity `json:"severity,omitempty"`
 
+	// Last time the condition got an update.
+	// Can be used by the system to determine if the ConditionStatus is Unknown in certain cases.
+	// +optional
+	LastHeartbeatTime metav1.Time `json:"lastHeartbeatTime,omitempty"`
```
stevekuznetsov, line 73: Hm, even upstream has no concept of this. Why do we need it? If heartbeats can be arbitrarily delayed, why is this valuable?

jmprusi: @stevekuznetsov perhaps is better to create a clusterCondition struct for this use case?

Used to monitor when the kubelet (syncer in our case) stops reporting, based on:
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L4334

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -91,3 +86,29 @@ type ClusterList struct {
...
+	ClusterReadyCondition conditionsv1alpha1.ConditionType = "Ready"
+
+	// ClusterUnknownReason documents a Cluster which readyness is unknown.
+	ClusterUnknownReason = "ClusterStatusUnknown"
+
+	// ClusterReadyReason documents a Cluster that is ready.
+	ClusterReadyReason = "ClusterReady"
+
+	// ClusterNotReadyReason documents a Cluster is not ready, when the "readyz" check returns false.
+	ClusterNotReadyReason = "ClusterNotReady"
```
sttts, line 102: I don't think a constant reason is the intention of the reason field. This should basically classify your error message. E.g. SyncerInstallFailed, SyncerNotReady (this only one of many), SyncerStartFailed

third_party/conditions/apis/conditions/v1alpha1/types.go
stevekuznetsov: I don't think that link is what you wanted - what did you mean to link to?

jmprusi: @stevekuznetsov true, sorry, this is the correct line: https://github.com/kubernetes/kubernetes/blob/f95c70080f21fe9aa6f14b72b75078d79ef68614/pkg/apis/core/types.go#L4391

pkg/apis/cluster/v1alpha1/types.go
jmprusi: I was updating those conditions thinking on the "next steps", the agent-based approach, where the syncer will update those fields. But let me add the current cases. We can simplify it later!

third_party/conditions/apis/conditions/v1alpha1/types.go
stevekuznetsov: I see. I guess having a union type here of NodeCondition and Condition is ok? @ncdc you see any challenge with upstreaming this?

jmprusi: So, I've just created a new struct with the missing field and embedded the Condition struct into it, so we can avoid having to upstream any changes. I guess this could work.. wdyt @ncdc ?

pkg/apis/cluster/v1alpha1/types.go
jmprusi: Updated! Please check again.

test/e2e/framework/util.go
```diff
@@ -326,11 +327,8 @@ func InstallCluster(t TestingTInterface, ctx context.Context, source, server Run
 					continue
 				}
 				var ready bool
-				for _, condition := range updated.Status.Conditions {
-					if condition.Type == clusterv1alpha1.ClusterConditionReady && condition.Status == corev1.ConditionTrue {
-						ready = true
-						break
-					}
+				if conditions.IsTrue(updated, clusterv1alpha1.ClusterReadyCondition) {
```
marun, line 330: Now that there is no longer a loop, maybe just test the conditional directly for indication of success?

```golang
if conditions.IsTrue(updated, clusterv1alpha1.ClusterReadyCondition) {
  return nil
}
```

pkg/reconciler/namespace/namespace.go
```diff
@@ -128,8 +128,7 @@ func (c *Controller) updateSchedulableCondition(ctx context.Context, ns *corev1.
 		})
 	}
 	return c.patchStatus(ctx, ns, func(conditionsSetter conditions.Setter) {
-		conditions.MarkFalse(conditionsSetter, NamespaceScheduled, NamespaceReasonUnschedulable,
-			conditionsv1alpha1.ConditionSeverityNone, // NamespaceCondition doesn't support severity
+		conditions.MarkFalse(conditionsSetter, NamespaceScheduled, NamespaceReasonUnschedulable, conditionsv1alpha1.ConditionSeverityNone, // NamespaceCondition doesn't support severity
```
marun, line 131: Please keep the ConditionSeverityNone line and its comment on the same line for readability.

test/e2e/framework/util.go
jmprusi: 👍🏼  Thanks.

marun:
LGTM

stevekuznetsov (approved):

PR #399, opened 2022-02-01
cmd: wire genericapiserver.SetupSignalContext() into servers for proper shutdown behaviour

sttts:
First ctrl-c closed context. Second ctrl-c exits.

Also post-start-hook should not return errors, but either return nil or keep going (otherwise you can a pretty `klog.Fatalf` stacktrace). `WaitForSync` only returns when the context is closed. So no need to put a loop around.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -27,7 +27,6 @@ import (
 	_ "net/http/pprof"
 	"os"
 	"path/filepath"
-	"reflect"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/server/server.go
```diff
@@ -282,16 +281,6 @@ func (s *Server) Run(ctx context.Context) error {
 		s.kubeSharedInformerFactory.Start(context.StopCh)
 		s.apiextensionsSharedInformerFactory.Start(context.StopCh)
 
-		if err := AllInSync(s.kcpSharedInformerFactory.WaitForCacheSync(context.StopCh)); err != nil {
```
ncdc, line 285: We still need the waits, though?

sttts: of course

ncdc (approved):

PR #411, opened 2022-02-02
Fix CRD inheritance

ncdc:
Update the inheritance-enabled CRD lister to support the new naming
scheme for workspaces.

### DIFF
--- a/pkg/apis/tenancy/v1alpha1/helper/helper_test.go
+++ b/pkg/apis/tenancy/v1alpha1/helper/helper_test.go
@@ -55,7 +55,7 @@ func TestEncodeLogicalClusterName(t *testing.T) {
 			name: "organization workspace in wrong root cluster",
 			input: &tenancyv1alpha1.Workspace{
 				ObjectMeta: metav1.ObjectMeta{
-					ClusterName: "somethingwrong",
+					ClusterName: "too_many_parts",
 					Name:        "organization",
 				},
 			},
@@ -86,14 +86,26 @@ func TestParseLogicalClusterName(t *testing.T) {
 		expectedErr  bool
 	}{
 		{
-			name:         "valid name",
+			name:         "request for /clusters/admin",
+			input:        "admin",
+			expectedOrg:  "admin",
+			expectedName: "admin",
+		},
+		{
+			name:         "valid name for organization workspace",
+			input:        "admin_organization",
+			expectedOrg:  "admin",
+			expectedName: "organization",
+		},
+		{
+			name:         "valid name for org and workspace",
 			input:        "organization_workspace",
 			expectedOrg:  "organization",
 			expectedName: "workspace",
 		},
 		{
 			name:        "invalid name",
-			input:       "whoathere",
+			input:       "foobar",
 			expectedErr: true,
 		},
 	} {
@@ -110,6 +122,35 @@ func TestParseLogicalClusterName(t *testing.T) {
 			}
 			if actualName != testCase.expectedName {
 				t.Errorf("%s: got incorrect logical cluster name, expected %s got %s", testCase.name, testCase.expectedName, actualName)
+			}
+		})
+	}
+}
+
+func TestWorkspaceKey(t *testing.T) {
+	tests := []struct {
+		name string
+		org  string
+		ws   string
+		want string
+	}{
+		{
+			name: "org ws",
+			org:  OrganizationCluster,
+			ws:   "myws",
+			want: "admin#$#myws",
+		},
+		{
+			name: "normal ws",
+			org:  "myorg",
+			ws:   "myws",
+			want: "admin_myorg#$#myws",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := WorkspaceKey(tt.org, tt.ws); got != tt.want {
+				t.Errorf("WorkspaceKey() = %v, want %v", got, tt.want)
 			}
 		})
 	}
--- a/pkg/server/apiextensions.go
+++ b/pkg/server/apiextensions.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/client-go/tools/clusters"
 	"k8s.io/klog/v2"
 
+	"github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1/helper"
 	tenancylisters "github.com/kcp-dev/kcp/pkg/client/listers/tenancy/v1alpha1"
 )
 
@@ -67,16 +68,17 @@ func (c *inheritanceCRDLister) ListWithContext(ctx context.Context, selector lab
 	// Check for API inheritance
 	inheriting := false
 	inheritFrom := ""
-	// TODO(ncdc): for now, we are hard-coding "admin" as the logical cluster where all Workspace
-	// resources must live, at least for API inheritance to work.
-	const adminClusterName = "admin"
+	org, ws, err := helper.ParseLogicalClusterName(cluster.Name)
+	if err != nil {
+		return nil, err
+	}
 
 	// Chicken-and-egg: we need the Workspace CRD to be created in the default admin logical cluster
 	// before we can try to get said Workspace, but if we fail listing because the Workspace doesn't
 	// exist, we'll never be able to create it. Only check if the target workspace exists for
 	// non-default keys.
-	if cluster.Name != adminClusterName && c.workspaceLister != nil {
-		targetWorkspaceKey := clusters.ToClusterAwareKey(adminClusterName, cluster.Name)
+	if cluster.Name != helper.OrganizationCluster && c.workspaceLister != nil {
+		targetWorkspaceKey := helper.WorkspaceKey(org, ws)
 		workspace, err := c.workspaceLister.Get(targetWorkspaceKey)
 		if err != nil && !apierrors.IsNotFound(err) {
 			// Only return errors other than not-found. If we couldn't find the workspace, let's continue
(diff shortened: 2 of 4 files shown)


pkg/server/apiextensions.go
```diff
@@ -91,12 +99,13 @@ func (c *inheritanceCRDLister) ListWithContext(ctx context.Context, selector lab
...
 		if workspace != nil && workspace.Spec.InheritFrom != "" {
 			// Make sure the source workspace exists
-			sourceWorkspaceKey := clusters.ToClusterAwareKey(adminClusterName, workspace.Spec.InheritFrom)
+			sourceWorkspaceKey := clusters.ToClusterAwareKey(org, workspace.Spec.InheritFrom)
 			_, err := c.workspaceLister.Get(sourceWorkspaceKey)
 			switch {
 			case err == nil:
 				inheriting = true
-				inheritFrom = workspace.Spec.InheritFrom
+				// FIXME use a helper
```
marun, line 107: Should this be addressed before merge?

pkg/server/apiextensions.go
```diff
@@ -110,6 +119,7 @@ func (c *inheritanceCRDLister) ListWithContext(ctx context.Context, selector lab
 	crds, err := c.crdLister.List(labels.Everything())
 	if err != nil {
 		return nil, err
+
```
marun, line 122: nit: Unrelated non-functional change?

pkg/server/apiextensions.go
```diff
@@ -65,18 +66,25 @@ func (c *inheritanceCRDLister) ListWithContext(ctx context.Context, selector lab
 	}
 
 	// Check for API inheritance
+	const adminClusterName = "admin"
```
marun, line 69: (No action required) For future, maybe reading the cluster/clusterName/workspace from context could be factored out into a common method?

marun (approved):
Changes LGTM

pkg/server/apiextensions.go
```diff
@@ -137,6 +147,17 @@ func (c *inheritanceCRDLister) GetWithContext(ctx context.Context, name string)
 		return nil, err
 	}
 
+	const adminClusterName = "admin"
+	org := adminClusterName
+	ws := cluster.Name
+
+	if cluster.Name != adminClusterName {
```
stevekuznetsov, line 154: Somehow this seems like it should not leak outside of the helper lib?

pkg/server/apiextensions.go
stevekuznetsov, line 107: Somehow this seems like it should not leak outside of the helper lib?

pkg/server/apiextensions.go
```diff
@@ -65,18 +66,25 @@ func (c *inheritanceCRDLister) ListWithContext(ctx context.Context, selector lab
...
+	const adminClusterName = "admin"
 	inheriting := false
 	inheritFrom := ""
-	// TODO(ncdc): for now, we are hard-coding "admin" as the logical cluster where all Workspace
-	// resources must live, at least for API inheritance to work.
-	const adminClusterName = "admin"
+	org := adminClusterName
+	ws := cluster.Name
+
+	if cluster.Name != adminClusterName {
```
stevekuznetsov, line 75: Somehow this seems like it should not leak outside of the helper lib?

pkg/server/apiextensions.go
ncdc: Fixed

pkg/server/apiextensions.go
ncdc: Fixed

pkg/server/apiextensions.go
ncdc: Fixed

pkg/server/apiextensions.go
```diff
@@ -67,16 +68,22 @@ func (c *inheritanceCRDLister) ListWithContext(ctx context.Context, selector lab
 	// Check for API inheritance
 	inheriting := false
 	inheritFrom := ""
-	// TODO(ncdc): for now, we are hard-coding "admin" as the logical cluster where all Workspace
-	// resources must live, at least for API inheritance to work.
-	const adminClusterName = "admin"
+	org := helper.OrganizationCluster
+	ws := cluster.Name
+
+	if cluster.Name != helper.OrganizationCluster {
```
stevekuznetsov, line 74: This logic seems like it should live in the helper - it's pretty domain specific (and something we do in the parse function as well)

stevekuznetsov (approved):

pkg/apis/tenancy/v1alpha1/helper/helper_test.go
```diff
@@ -55,7 +55,7 @@ func TestEncodeLogicalClusterName(t *testing.T) {
 			name: "organization workspace in wrong root cluster",
 			input: &tenancyv1alpha1.Workspace{
 				ObjectMeta: metav1.ObjectMeta{
-					ClusterName: "somethingwrong",
+					ClusterName: "too_many_parts",
```
ncdc, line 58: @stevekuznetsov PTAL at the changes to the test cases here - thanks!

pkg/apis/tenancy/v1alpha1/helper/helper_test.go
```diff
@@ -86,14 +86,20 @@ func TestParseLogicalClusterName(t *testing.T) {
 		expectedErr  bool
 	}{
 		{
-			name:         "valid name",
+			name:         "valid name for organization workspace",
```
stevekuznetsov, line 89: I don't think this is right - an organization would be `admin_organization`, not `organization`. The only logical cluster without `_` is `admin` or whatever we choose the root to be.

test/e2e/api_inheritance/api_inheritance_test.go
```diff
@@ -65,69 +65,62 @@ func TestAPIInheritance(t *testing.T) {
...
 		kcpClients, err := clientset.NewClusterForConfig(cfg)
 		if err != nil {
 			t.Errorf("failed to construct kcp client for server: %v", err)
 			return
 		}
 
-		// For now, API inheritance only supports workspaces in the admin logical cluster, so let's create
-		// a couple there.
-		kcpAdminClient := kcpClients.Cluster("admin")
+		kcpOrganizationClient := kcpClients.Cluster(helper.OrganizationCluster)
```
stevekuznetsov, line 75: So ultimately I think this is the crux of it - what do we actually want to happen with this feature? If two `Workspace`s can support inheritance at one "level" ... e.g. within an org, one can inherit from the other, we actually want this to be some organizational workspace we're watching. As written, this (poorly named) constant has us watching the root logical cluster where we find organizational workspaces.

In the far future we expect to have N logical clusters on the kcp shard that mirror N organizational workspaces, right? If a shard holds workspaces from N orgs. So we would want to watch all of those for inheritance.

So perhaps we could still do a `*` client here, but the actual logical clusters in which we create `Workspace` objects would be e.g. `admin_org-a`, `admin_org-b` etc? The actual `Workspace`s would be e.g. `org-b_ws-1`...

pkg/apis/tenancy/v1alpha1/helper/helper_test.go
ncdc: Ok, so we could have

- /clusters/admin - org=?, ws=?
- /clusters/foobar - invalid (?)
- /clusters/admin_my-org - org=admin, ws=my-org
- /clusters/my-org_my-ws - org=my-org, ws=my-ws

It sounds like you're saying /clusters/foobar is invalid. If we all agree, I can update to check for that.

For /clusters/admin, what should this parsing helper return for org & ws?

sttts: I assume admin = root in the future. So root_org makes sense technically, but is guly. And yes, foobar is invalid.

Admin aka root will be a special case forever. Only hack we could do is root_root.

sttts: And we have to think about the local system lclusters. They might be different, i.e. some more exception. Or we find a naming scheme for them that fits into what we have here: e.g. local_foo.

ncdc: How does this look?
```go
{
			name:         "request for /clusters/admin",
			input:        "admin",
			expectedOrg:  "admin",
			expectedName: "admin",
		},
		{
			name:         "valid name for organization workspace",
			input:        "admin_organization",
			expectedOrg:  "admin",
			expectedName: "organization",
		},
		{
			name:         "valid name for org and workspace",
			input:        "organization_workspace",
			expectedOrg:  "organization",
			expectedName: "workspace",
		},
		{
			name:        "invalid name",
			input:       "foobar",
			expectedErr: true,
		},
```

test/e2e/api_inheritance/api_inheritance_test.go
stevekuznetsov: Yes, I guess two organizational workspaces can have inheritance? I think a `*` client is still necessary then, so it would be able to see `Workspace` objects regardless of if they were mirrored from the root logical cluster, or any organizational workspace. It may be easier to grok it all if you do `acme_team-abc` in the test (or that _and_ two org workspaces)

pkg/server/apiextensions.go
```diff
@@ -67,16 +68,28 @@ func (c *inheritanceCRDLister) ListWithContext(ctx context.Context, selector lab
...
+	org, ws, err := helper.ParseLogicalClusterName(cluster.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	/* org could possibly be:
+		- admin (helper.OrganizationCluster)
+		- someorg
+
+	   If it's someorg, when looking for Workspace instances, we need to look in the logical cluster
```
stevekuznetsov, line 80: This will inform what we require as a virtual workspace - can we somehow hide it ... ?

stevekuznetsov (approved):

pkg/server/apiextensions.go
ncdc: Virtual workspace, or just workspace in general?

sttts (approved):

ncdc:
Hold on merging for a few

pkg/server/apiextensions.go
stevekuznetsov: I was thinking virtual, since the question we want to answer is "show me the workspaces I need to know about"

stevekuznetsov: Client library with filtering would be fine too I guess

stevekuznetsov (approved):

PR #414, opened 2022-02-03
syncer: Fix informers and workers leak

### LINKED ISSUE #409: High CPU consumption
**Describe the bug**

I'm running KCP on my local machine (MacBook Pro 2019, 2,6 GHz 6-Core i7, 32 GB memory), and the KCP process steadily consumes > 1000% CPU.

Some client requests time out, with the KCP logs containing:

```
{"level":"warn","ts":"2022-02-02T14:26:11.184+0100","caller":"etcdserver/util.go:166","msg":"apply request took too long","took":"1.201838934s","expected-duration":"100ms","prefix":"read-only range ","request":"key:\"/registry/networking.k8s.io/ingresses/\" range_end:\"/registry/networking.k8s.io/ingresses0\" count_only:true ","response":"range_response_count:0 size:5"}
```

**To Reproduce**

```
$ start --push-mode --install-cluster-controller --resources-to-sync=deployments --resources-to-sync=services --resources-to-sync=ingresses.networking.k8s.io --auto-publish-apis
```

And to KinD clusters configured:

```
$ kubectl get clusters
NAME
kcp-cluster-a
kcp-cluster-b
```

**Expected behavior**

I've mitigated the issue by having only one physical cluster. Still the CPU consumption is > 1000% CPU, but at least the client requests do not time out.

The KCP process should not consume that much CPU, and scale well with the number of physical clusters.

**Additional context**

This happens with the `kcp-ingress` project setup: https://github.com/jmprusi/kcp-ingress#getting-started.

astefanutti:
Syncers currently leak informers and workers after being stopped, so that events from informers are still processed by the workers, while the underlying workqueue is shut down, which leads to infinite loops that starve CPU. 

The Done channel from the cancellable context is passed to the syncer informers and workers, instead of the parent context one, so that the syncer stop is properly propagated.

Fixes #409.

### DIFF
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -258,11 +258,11 @@ func (c *Controller) AddToQueue(gvr schema.GroupVersionResource, obj interface{}
 func (c *Controller) Start(ctx context.Context, numThreads int) {
 	c.ctx, c.cancelFn = context.WithCancel(ctx)
 
-	c.fromInformers.Start(ctx.Done())
-	c.fromInformers.WaitForCacheSync(ctx.Done())
+	c.fromInformers.Start(c.ctx.Done())
+	c.fromInformers.WaitForCacheSync(c.ctx.Done())
 
 	for i := 0; i < numThreads; i++ {
-		go c.startWorker(ctx)
+		go c.startWorker(c.ctx)
 	}
 }
 


ncdc (approved):

ncdc:
@astefanutti thanks for catching this!

PR #420, opened 2022-02-03
authorization: associate edit verb with cluster-admin

s-urbaniak:
cc @ncdc @sttts

### DIFF
--- a/pkg/authorization/bootstrap/policy.go
+++ b/pkg/authorization/bootstrap/policy.go
@@ -25,7 +25,7 @@ import (
 // ClusterRoleBindings return default rolebindings to the default roles
 func clusterRoleBindings() []rbacv1.ClusterRoleBinding {
 	return []rbacv1.ClusterRoleBinding{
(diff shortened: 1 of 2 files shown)


pkg/authorization/bootstrap/policy.go
```diff
@@ -25,7 +25,7 @@ import (
 // ClusterRoleBindings return default rolebindings to the default roles
 func clusterRoleBindings() []rbacv1.ClusterRoleBinding {
 	return []rbacv1.ClusterRoleBinding{
-		clusterRoleBindingCustomName(rbacv1helpers.NewClusterBinding("edit").Groups("system:kcp:workspace:edit").BindingOrDie(), "system:kcp:workspace:edit"),
+		clusterRoleBindingCustomName(rbacv1helpers.NewClusterBinding("cluster-admin").Groups("system:kcp:workspace:edit").BindingOrDie(), "system:kcp:workspace:edit"),
```
ncdc, line 28: Pretty sure this needs to be system:masters or we need to create a new cluster-admin group and add it to the cluster-admin role binding

ncdc (approved):

PR #423, opened 2022-02-03
Kubeconfig for the virtual workspace APIserver

davidfestal:
We now generate a `virtual.kubeconfig` file at start of the virtual workspaces APIServer, exactly the same
way as it is done for the KCP APIServer.

This allows easily connecting to the virtual workspace from `kubectl| without having certificate problems, or complex commands, but simply with:
```
kubectl --kubeconfig .kcp/virtual.kubeconfig get workpaces
```

This will greatly simplify/shorten a number of workspace-related commands in the prototype 2 demo.

### DIFF
--- a/pkg/virtual/framework/cmd/kubeconfig.go
+++ b/pkg/virtual/framework/cmd/kubeconfig.go
@@ -21,14 +21,13 @@ import (
 	"path/filepath"
 
 	"k8s.io/apimachinery/pkg/util/sets"
-	genericapiserver "k8s.io/apiserver/pkg/server"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 
 	"github.com/kcp-dev/kcp/pkg/virtual/framework"
 )
 
-func (o *APIServerOptions) writeKubeConfig(server *genericapiserver.GenericAPIServer, virtualWorkspaces []framework.VirtualWorkspace) error {
+func (o *APIServerOptions) writeKubeConfig(serverHost, TLSServerName string, CAData []byte, virtualWorkspaces []framework.VirtualWorkspace) error {
 	var clientConfig clientcmdapi.Config
 
 	contextNames := sets.NewString()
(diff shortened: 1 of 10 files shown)


pkg/virtual/framework/types.go
```diff
@@ -45,6 +49,9 @@ type ReadyFunc func() error
 type VirtualWorkspace interface {
 	GetName() string
 	ResolveRootPath(urlPath string, context context.Context) (accepted bool, prefixToStrip string, completedContext context.Context)
+	// GetKubeContextPaths returns a map in which keys are the kubeconfig context names of published endpoints
+	// provided by this virtual workspace, and values are associated endpoint path prefix.
+	GetKubeContextPaths() map[string]string
```
stevekuznetsov, line 54: Seems like a weird factoring - why does the server have an opinion on what some client calls these endpoints in their local kubeconfig?

pkg/virtual/framework/cmd/kubeconfig.go
```diff
@@ -0,0 +1,58 @@
...
+func (o *APIServerOptions) writeKubeConfig(server *genericapiserver.GenericAPIServer, virtualWorkspaces []framework.VirtualWorkspace) error {
+	var clientConfig clientcmdapi.Config
+
+	contextNames := sets.NewString()
+	clientConfig.Clusters = map[string]*clientcmdapi.Cluster{}
+	clientConfig.Contexts = map[string]*clientcmdapi.Context{}
+
+	for _, vw := range virtualWorkspaces {
+		for contextName, path := range vw.GetKubeContextPaths() {
+			if contextNames.Has(contextName) {
```
stevekuznetsov, line 40: You only ever populate this set with keys in a map - there cannot be any overlap

pkg/virtual/workspaces/cmd/cmd.go
```diff
@@ -148,3 +149,7 @@ func (o *WorkspacesSubCommandOptions) PrepareVirtualWorkspaces() ([]rootapiserve
 	}
 	return informerStarts, virtualWorkspaces, nil
 }
+
+func (o *WorkspacesSubCommandOptions) GetCurrentKubeContext() string {
```
stevekuznetsov, line 153: Why does _this_ part of the command get to choose what the current context is? When can this vary?

pkg/virtual/framework/cmd/kubeconfig.go
davidfestal: There's one map per virtual workspace, so due to the double loop it seems there could be duplicates.

pkg/virtual/framework/types.go
davidfestal: It's not on their local kubeconfig. These are the contexts that the virtual workspace wants to be published in the kubeconfig that will be created by the virtual workspaces Root API Server command.

pkg/virtual/workspaces/cmd/cmd.go
davidfestal: Typically, the *personal workspaces* command (`virtual-workspaces workspaces`) publishes 2 contexts:
- one (with the ./.../personal URL suffix) for the personal workspaces view
- one (with the ./.../organization URL suffix) for the organization workspaces view

But typically the `personal` view is the preferred one, or the default one, so it will be made the current context in the generated kubeconfig.

pkg/virtual/framework/types.go
stevekuznetsov: I understand. Why does the _server_ care about kubeconfig contexts? These are separate, unrelated domains.

sttts:
I am unsure about this PR, mostly because it looks like it solves a problem that shouldn't be there.

To my understanding, the issue is that the virtual apiserver has a different CA used to create the serving certs. Hence, one cannot just reuse the kcp CA cert in the kubeconfig.

IMO the right solution in a production setup is to have a common root CA cc @kylape @csams 

I am worried that we now add logic here that is only useful in a development setup. I would rather prefer that we enable the reuse of the certs in `.kcp` directory. What is stopping us from that?

ncdc:
Given the questions on this PR, I am converting it to a draft, and we'll follow up next week.

ncdc:
Replaced by #486

PR #424, opened 2022-02-03
server: switch to options structure of upstream

sttts:
- use composable options+config structs from upstream (especially apiserver)
- with that make use of flag validation
- really test pkg/genericcontrollerplane in kcp-dev/k
- switch to `--run-controllers` enabling all controllers locally
- open up all the features from upstream components we will need for production anyway and don't want to rebuild
- add whitelist+blacklist of flags, secured against drift through unit tests
```
$ kcp start --help
Start the control plane process

The server process listens on port 6443 and will act like a Kubernetes API server. It will initialize any necessary data to the provided start location or as a '.kcp' directory in the current directory. An admin kubeconfig file will be generated at initialization time that may be used to access the control plane.


Usage:
  kcp start [flags]

Most important flags:

      --etcd-servers strings
                List of etcd servers to connect with (scheme://ip:port), comma separated. By default an embedded etcd
                server is started. (default [embedded])
      --run-controllers
                Run the controllers in-process (default true)

Use "kcp start options" for a list of all flags available.
```

https://docs.google.com/document/d/17SiAbgkj5o2lp9x9lzIqaSo8D65oP1sRz4mYCGNq1b4/edit with all the flags. Lots to be filtered out.


pkg/server/options/genericcontrolplane.go
```diff
@@ -0,0 +1,112 @@
...
+
+// Validate checks Options and return a slice of found errs.
+func (s *ServerRunOptions) Validate() []error {
+	var errs []error
+	errs = append(errs, s.Etcd.Validate()...)
+	errs = append(errs, s.SecureServing.Validate()...)
+	errs = append(errs, s.Authentication.Validate()...)
+	errs = append(errs, s.Audit.Validate()...)
+	errs = append(errs, s.Admission.Validate()...)
+	//errs = append(errs, s.APIEnablement.Validate(legacyscheme.Scheme, apiextensionsapiserver.Scheme, aggregatorscheme.Scheme)...)
```
stevekuznetsov, line 106: ?

pkg/server/server.go
```diff
@@ -135,113 +121,50 @@ func (s *Server) Run(ctx context.Context) error {
...
 		}
-		serverOptions.SecureServing.BindPort = p
+		s.options.GenericControlPlane.Etcd.StorageConfig.Transport.ServerList = embeddedClientInfo.Endpoints
+		s.options.GenericControlPlane.Etcd.StorageConfig.Transport.KeyFile = embeddedClientInfo.KeyFile
+		s.options.GenericControlPlane.Etcd.StorageConfig.Transport.CertFile = embeddedClientInfo.CertFile
+		s.options.GenericControlPlane.Etcd.StorageConfig.Transport.TrustedCAFile = embeddedClientInfo.TrustedCAFile
 	}
 
+	// patch the handler chain. We should do this after creating the generic apiserver config, but before CreateKubeAPIServerConfig. But
+	// this needs surgery in k/k. So we do it here no, changing the already completed options. Not nice, but must be here because
```
stevekuznetsov, line 140:
```suggestion
	// this needs surgery in k/k. So we do it here now, changing the already completed options. Not nice, but must be here because
```

stevekuznetsov (commented):
Seems reasonable - is the `genericcontrolplane` change in k8s a follow-up?

pkg/server/options/embeddedetcd.go
```diff
@@ -0,0 +1,64 @@
...
+	return &EmbeddedEtcd{
+		Enable:     true,
+		Directory:  "",
+		PeerPort:   "2380",
+		ClientPort: "2379",
+	}
+}
+
+func (e *EmbeddedEtcd) AddFlags(fs *pflag.FlagSet) {
+	fs.BoolVar(&e.Enable, "enable-embedded-etcd", e.Enable, "Enable embedded etcd. Will be ignored if --etcd-servers is specified")
```
smarterclayton, line 44: This name seems fine, although I'm a bit worried about adding flags that we don't directly and clearly need today (commented out or disabled here).  If WAL size is needed for integration I guess I can live with it.

pkg/server/options/genericcontrolplane.go
```diff
@@ -0,0 +1,112 @@
...
+//
+
+type ServerRunOptions struct {
+	options.ServerRunOptions
+}
+
+// Flags returns flags for a specific APIServer by section name
+func (s *ServerRunOptions) Flags() (fss cliflag.NamedFlagSets) {
+	// TODO: move all these back into genericcontrolplane
+	s.GenericServerRunOptions.AddUniversalFlags(fss.FlagSet("generic"))
```
smarterclayton, line 40: Would prefer fewer flags rather than more being exposed to users right now.  150+ flags is too many - I might prefer an allowlist mindset (if that hasn't already been thought of and i just missed it in review)

smarterclayton:
To summarize, totally agree with the approach, but specifically the public api of number of flags has to be constrained and I would prefer only a very small allowlist of flags is exposed, and each one/set-of-similar-flags needs a use case justification

sttts:
@smarterclayton please see https://docs.google.com/document/d/17SiAbgkj5o2lp9x9lzIqaSo8D65oP1sRz4mYCGNq1b4/edit and mark those you don't want to have. Maybe mark in some color.

pkg/server/options/genericcontrolplane.go
sttts: legacyscheme is evil.

sttts: The whole set of flags is now disabled.

pkg/server/options/genericcontrolplane.go
sttts: agree, especially for these generic one. Most look like random one-offs.

pkg/server/options/embeddedetcd.go
sttts: Have removed it. Now `--etcd-servers=embedded` is the default.

```
Start the control plane process

The server process listens on port 6443 and will act like a Kubernetes API server. It will initialize any necessary data to the provided start location or as a '.kcp' directory in the current directory. An admin kubeconfig file will be generated at initialization time that may be used to access the control plane.


Usage:
  kcp start [flags]

Most important flags:

      --etcd-servers strings
                List of etcd servers to connect with (scheme://ip:port), comma separated. By default an embedded etcd
                server is started. (default [embedded])
      --run-controllers
                Run the controllers in-process (default true)

Use "kcp start options" for a list of all flags available.
```

sttts:
> Seems reasonable - is the genericcontrolplane change in k8s a follow-up?

@stevekuznetsov yes, have an unfinished branch for that. Then I think the injector will be history 🥳

sttts:
The failing tests seem to be timeouts. Locally I can run them without concurrency. Waiting for the CI PRs to land.

pkg/server/options/authorization.go
```diff
@@ -0,0 +1,112 @@
...
+func (s *Authorization) ApplyTo(config *genericapiserver.Config, informer coreexternalversions.SharedInformerFactory, workspaceLister v1alpha1.WorkspaceLister) error {
+	var authorizers []authorizer.Authorizer
+
+	// group authorizer
+	if len(s.AlwaysAllowGroups) > 0 {
+		authorizers = append(authorizers, authorizerfactory.NewPrivilegedGroups(s.AlwaysAllowGroups...))
+	}
+
+	// path authorizer
+	if len(s.AlwaysAllowPaths) > 0 {
```
sttts, line 92: @s-urbaniak why was this not needed before? Without `/readyz` only works authenticated.

test/e2e/api_inheritance/api_inheritance_test.go
```diff
@@ -75,25 +76,32 @@ func TestAPIInheritance(t *testing.T) {
 				return
 			}
 
+			t.Logf("Bootstrapping workspace CRD in org workspace %q", testCase.logicalClusterName)
 			apiExtensionsClients, err := apiextensionsclient.NewClusterForConfig(cfg)
 			if err != nil {
 				t.Errorf("failed to construct apiextensions client for server: %v", err)
 				return
 			}
+			if err := config.BootstrapCustomResourceDefinitions(ctx, apiExtensionsClients.Cluster(testCase.logicalClusterName).ApiextensionsV1().CustomResourceDefinitions(), []metav1.GroupResource{
```
sttts, line 85: @ncdc how did this work before? The non-"admin" workspace never had the workspace CRD installed (and should not until we have a controller doing that for org workspaces).

pkg/server/options/authorization.go
s-urbaniak: it seems it was not wired before:
```
% curl -k https://127.0.0.1:6443/readyz
Internal Server Error: "/readyz": expected logical cluster name to be in org_name format, got admin
```

test/e2e/api_inheritance/api_inheritance_test.go
ncdc: This test used to only test Workspaces in admin.

davidfestal (approved):
Approving with just a very minimal typo suggestion.

cmd/kcp/help.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"k8s.io/apimachinery/pkg/util/sets"
+	cliflag "k8s.io/component-base/cli/flag"
+)
+
+const (
+	usageFmt = "Usage:\n  %s\n"
+)
+
+// setUsageAndHelpFunc set both usage and help function.
```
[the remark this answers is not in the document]
sttts: odd that lint is not unhappy here.

pkg/server/options/flags_test.go
```diff
@@ -0,0 +1,75 @@
...
+func TestFlagWhitelist(t *testing.T) {
+	o := NewOptions()
+	fss := o.rawFlags()
+
+	missing := map[string][]*pflag.Flag{}
+
+	for section, fs := range fss.FlagSets {
+		fs.VisitAll(func(f *pflag.Flag) {
+			if !allowedFlags.Has(f.Name) && !disallowedFlags.Has(f.Name) {
+				t.Errorf("--%s in section \"%s flags\" is neither whitelisted nor blacklisted", f.Name, section)
```
ncdc, line 36:
```suggestion
				t.Errorf("--%s in section \"%s flags\" is neither allowed nor disallowed", f.Name, section)
```

ncdc (changes requested):

ncdc (approved):

PR #428, opened 2022-02-04
Ensure namespace scheduling satisfies prototype 2 requirements

marun:
- Ensure e2e testing validates desired scenarios
- Ensure scheduling status is communicated via namespace conditions
- Ensure cluster removal prompts rescheduling

### DIFF
--- a/pkg/reconciler/namespace/conditions.go
+++ b/pkg/reconciler/namespace/conditions.go
@@ -1,0 +1,122 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package namespace
+
+import (
+	"encoding/json"
+	"fmt"
+
+	jsonpatch "github.com/evanphx/json-patch"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	conditionsapi "github.com/kcp-dev/kcp/third_party/conditions/apis/conditions/v1alpha1"
+	"github.com/kcp-dev/kcp/third_party/conditions/util/conditions"
+)
+
+const (
+	// NamespaceScheduled represents status of the scheduling process for this namespace.
+	NamespaceScheduled conditionsapi.ConditionType = "NamespaceScheduled"
+	// NamespaceReasonUnschedulable reason in NamespaceScheduled Namespace Condition
+	// means that the scheduler can't schedule the namespace right now, e.g. due to a
+	// lack of ready clusters being available.
+	NamespaceReasonUnschedulable = "Unschedulable"
+)
+
+// NamespaceConditionsAdapter enables the use of the conditions helper
+// library with Namespaces.
+type NamespaceConditionsAdapter struct {
+	*corev1.Namespace
+}
+
+func (ca *NamespaceConditionsAdapter) GetConditions() conditionsapi.Conditions {
+	conditions := conditionsapi.Conditions{}
+	for _, c := range ca.Status.Conditions {
+		conditions = append(conditions, conditionsapi.Condition{
+			Type:   conditionsapi.ConditionType(c.Type),
+			Status: c.Status,
+			// Default to None because NamespaceCondition lacks a Severity field
+			Severity:           conditionsapi.ConditionSeverityNone,
+			LastTransitionTime: c.LastTransitionTime,
+			Reason:             c.Reason,
+			Message:            c.Message,
+		})
+	}
+	return conditions
+}
+
+func (ca *NamespaceConditionsAdapter) SetConditions(conditions conditionsapi.Conditions) {
+	nsConditions := []corev1.NamespaceCondition{}
+	for _, c := range conditions {
+		nsConditions = append(nsConditions, corev1.NamespaceCondition{
+			Type:   corev1.NamespaceConditionType(c.Type),
+			Status: c.Status,
+			// Severity is ignored
+			LastTransitionTime: c.LastTransitionTime,
+			Reason:             c.Reason,
+			Message:            c.Message,
+		})
+	}
+	ca.Status.Conditions = nsConditions
+}
+
+type updateConditionsFunc func(conditionsSetter conditions.Setter)
+
+// statusPatchBytes returns the bytes required to patch status for the
+// provided namespace from its original state to the state after it
+// has been mutated by the provided function.
+func statusPatchBytes(ns *corev1.Namespace, updateConditions updateConditionsFunc) ([]byte, error) {
+	oldData, err := json.Marshal(corev1.Namespace{
+		Status: ns.Status,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal existing status for namespace %q: %w", ns.Name, err)
+	}
+
+	// Mutation is assumed safe
+	conditionsAdapter := &NamespaceConditionsAdapter{ns}
+	updateConditions(conditionsAdapter)
+
+	newData, err := json.Marshal(corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			UID:             ns.UID,
+			ResourceVersion: ns.ResourceVersion,
+		}, // to ensure they appear in the patch as preconditions
+		Status: ns.Status,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal new status for namespace %q: %w", ns.Name, err)
+	}
+
+	patchBytes, err := jsonpatch.CreateMergePatch(oldData, newData)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create status patch for namespace %q: %w", ns.Name, err)
+	}
+	return patchBytes, nil
+}
+
+// HasScheduledStatus returns whether the given namespace's status
+// indicates it is scheduled.
+func HasScheduledStatus(ns *corev1.Namespace) bool {
+	for _, condition := range ns.Status.Conditions {
+		if condition.Type == corev1.NamespaceConditionType(NamespaceScheduled) {
+			return condition.Status == corev1.ConditionTrue
+		}
+	}
+	return false
+}
--- a/pkg/reconciler/namespace/conditions.go
+++ b/pkg/reconciler/namespace/conditions.go
@@ -85,7 +85,7 @@ func statusPatchBytes(ns *corev1.Namespace, updateConditions updateConditionsFun
 		Status: ns.Status,
 	})
 	if err != nil {
-		return nil, fmt.Errorf("failed to marshal existing status for namespace %q: %w", ns.Name, err)
+		return nil, fmt.Errorf("failed to marshal existing status for namespace %q|%q: %w", ns.ClusterName, ns.Name, err)
 	}
 
 	// Mutation is assumed safe
@@ -100,19 +100,19 @@ func statusPatchBytes(ns *corev1.Namespace, updateConditions updateConditionsFun
 		Status: ns.Status,
 	})
 	if err != nil {
-		return nil, fmt.Errorf("failed to marshal new status for namespace %q: %w", ns.Name, err)
+		return nil, fmt.Errorf("failed to marshal new status for namespace %q|%q: %w", ns.ClusterName, ns.Name, err)
 	}
 
 	patchBytes, err := jsonpatch.CreateMergePatch(oldData, newData)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create status patch for namespace %q: %w", ns.Name, err)
+		return nil, fmt.Errorf("failed to create status patch for namespace %q|%q: %w", ns.ClusterName, ns.Name, err)
 	}
 	return patchBytes, nil
 }
 
-// HasScheduledStatus returns whether the given namespace's status
-// indicates it is scheduled.
-func HasScheduledStatus(ns *corev1.Namespace) bool {
+// IsScheduled returns whether the given namespace's status indicates
+// it is scheduled.
+func IsScheduled(ns *corev1.Namespace) bool {
 	for _, condition := range ns.Status.Conditions {
 		if condition.Type == corev1.NamespaceConditionType(NamespaceScheduled) {
 			return condition.Status == corev1.ConditionTrue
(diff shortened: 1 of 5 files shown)


pkg/reconciler/namespace/namespace.go
```diff
@@ -100,55 +103,156 @@ func (c *Controller) reconcileGVR(ctx context.Context, gvr schema.GroupVersionRe
...
-func (c *Controller) reconcileNamespace(ctx context.Context, lclusterName string, ns *corev1.Namespace) error {
-	klog.Infof("Reconciling Namespace %s", ns.Name)
+func (c *Controller) patchStatus(ctx context.Context, ns *corev1.Namespace, updateConditions updateConditionsFunc) error {
+	patchBytes, err := statusPatchBytes(ns, updateConditions)
+	if err != nil {
+		return err
+	}
+	_, err = c.namespaceClient.Patch(ctx, ns.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{}, "status")
+	if err != nil {
+		return fmt.Errorf("failed to patch status on namespace %q: %w", ns.Name, err)
```
marun, line 113: I'm guessing all errors and logging should be qualifying resource names with the org+workspace?

pkg/reconciler/namespace/conditions.go
```diff
@@ -0,0 +1,122 @@
...
+
+// statusPatchBytes returns the bytes required to patch status for the
+// provided namespace from its original state to the state after it
+// has been mutated by the provided function.
+func statusPatchBytes(ns *corev1.Namespace, updateConditions updateConditionsFunc) ([]byte, error) {
+	oldData, err := json.Marshal(corev1.Namespace{
+		Status: ns.Status,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal existing status for namespace %q: %w", ns.Name, err)
```
marun, line 88: I'm guessing all errors and logging should be qualifying resource names with the org+workspace?

marun:
What's the general strategy for managing status for kcp resources?  Is status set by kcp to be stripped by the syncer in resources copied to pclusters? And should status synced back from pclusters be merged with status maintained in kcp?

pkg/reconciler/namespace/namespace.go
```diff
@@ -100,55 +103,156 @@ func (c *Controller) reconcileGVR(ctx context.Context, gvr schema.GroupVersionRe
...
+}
 
-	if ns.Labels == nil {
-		ns.Labels = map[string]string{}
+// Set the schedulable condition on namespaces and not the resources it
+// contains. Not all native types support status conditions (e.g. configmaps
+// and secrets) and there is no guarantee that arbitrary CRDs will support
+// conditions.
+func (c *Controller) updateSchedulableCondition(ctx context.Context, ns *corev1.Namespace, schedulable bool) error {
+	klog.Infof("Patching namespace %q status to indicate schedulable=%s", ns.Name, schedulable)
```
sttts, line 123: do we know here that there is a diff? Otherwise, this is noisy.

pkg/reconciler/namespace/namespace.go
```diff
@@ -100,55 +103,156 @@ func (c *Controller) reconcileGVR(ctx context.Context, gvr schema.GroupVersionRe
...
-	cl, err := c.clusterLister.Get(clusters.ToClusterAwareKey(lclusterName, clusterName))
+	return nil
+}
+
+// isValidCluster checks whether the given cluster name exists and is valid for the
+// purposes of scheduling a namespace to it.
+func (c *Controller) isValidCluster(ctx context.Context, lclusterName, clusterName string) (bool, error) {
+	cluster, err := c.clusterLister.Get(clusters.ToClusterAwareKey(lclusterName, clusterName))
+	if apierrors.IsNotFound(err) {
+		klog.Infof("Cluster %q does not exist", clusterName)
```
sttts, line 173: this seems to be noisy. I assume this func can be called pretty often.

pkg/reconciler/namespace/conditions.go
sttts: yes please

pkg/reconciler/namespace/conditions.go
```diff
@@ -0,0 +1,122 @@
...
+	patchBytes, err := jsonpatch.CreateMergePatch(oldData, newData)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create status patch for namespace %q: %w", ns.Name, err)
+	}
+	return patchBytes, nil
+}
+
+// HasScheduledStatus returns whether the given namespace's status
+// indicates it is scheduled.
+func HasScheduledStatus(ns *corev1.Namespace) bool {
```
imjasonh, line 115: (not blocking) `HasScheduledStatus` might also be interpreted to mean "has a status condition of type == scheduled (whatever its value/truthiness)"

Maybe `IsScheduled` would match the logic better?

pkg/reconciler/namespace/namespace.go
```diff
@@ -100,55 +103,156 @@ func (c *Controller) reconcileGVR(ctx context.Context, gvr schema.GroupVersionRe
...
+		return false, nil
+	}
 	if err != nil {
+		return false, err
+	}
+	isReady := cluster.Status.Conditions.HasReady()
```
imjasonh, line 179: (not blocking) This `isReady := ...HasReady()` sounds like more reason to prefer `is` phrasing over `has` phrasing.

imjasonh (approved):

imjasonh:
> What's the general strategy for managing status for kcp resources? Is status set by kcp to be stripped by the syncer in resources copied to pclusters?

I don't think it _has_ to be stripped, but I also don't think it should tend to have any effect down there either (i.e., syncer doesn't care about `Scheduleable` conditions in either direction).

Stripping and re-merging going upstream might be cumbersome for little benefit, sos I think it'd be fine to keep it. Unless there's some problem anyone foresees with that approach?

> And should status synced back from pclusters be merged with status maintained in kcp?

Yeah.

pkg/reconciler/namespace/conditions.go
stevekuznetsov: Logical cluster (`.metadata.clusterName`) is a global identifier, so that would be easy to use

pkg/reconciler/namespace/namespace.go
```diff
@@ -100,55 +103,156 @@ func (c *Controller) reconcileGVR(ctx context.Context, gvr schema.GroupVersionRe
...
+// ensureScheduledStatus ensures the status of the given namespace reflects whether the
+// namespace has been assigned to a cluster.
+func (c *Controller) ensureScheduledStatus(ctx context.Context, ns *corev1.Namespace, oldClusterName string) error {
+	// Set status on cluster change to bump the transition time
+	newClusterName := ns.Labels[clusterLabel]
+	clusterChanged := oldClusterName != newClusterName
+
+	// Set status if the scheduled state of the namespace has changed
+	scheduled := (newClusterName != "")
+	statusChangeRequired := scheduled != HasScheduledStatus(ns)
```
stevekuznetsov, line 222: So one of the benefits of the library I *think* is that attempting to set a condition to a value that's already set will be a no-op, so we should not have to do this sort of check?

test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -275,56 +187,29 @@ func scheduledMatcher(target string) matcher {
...
+		key, err := cache.MetaNamespaceKeyFunc(seed)
+		if err != nil {
+			return err
 		}
-	}
-	return false
+		return expecter.ExpectBefore(ctx, func(ctx context.Context) (done bool, err error) {
+			current, err := informer.Lister().Get(key)
+			if err != nil {
+				// Retry on all errors
```
stevekuznetsov, line 208: Depends on your intent but 403 -> 502 -> 200 would IMHO be a valid failure for the test, but this code would swallow the first broken responses as long as it converges?

stevekuznetsov (commented):
Mostly LGTM

pkg/reconciler/namespace/namespace.go
marun: As per `ensureScheduledStatus`, `updateScheduledCondition` is only called if a) there was a change in scheduled cluster or b) the current condition doesn't match the state of the namespace label.  I'm not sure I see your concern?

pkg/reconciler/namespace/namespace.go
marun: Any given cluster is likely to be assigned to multiple namespaces, so that if a cluster is deleted or goes unready many namespaces will be affected and hit these log lines. Isn't that desirable though? The log could show 'cluster foo does not exist' and then 'scheduling namespace foo to other cluster'.

pkg/reconciler/namespace/namespace.go
marun: My thinking is isReady[Cluster] vs HasReady[Condition]. In this context, my intended focus is on the cluster.

pkg/reconciler/namespace/namespace.go
```diff
@@ -100,55 +103,156 @@ func (c *Controller) reconcileGVR(ctx context.Context, gvr schema.GroupVersionRe
...
+			}
+		}
+	}
+	return nil
+}
+
+// ensureScheduledStatus ensures the status of the given namespace reflects whether the
+// namespace has been assigned to a cluster.
+func (c *Controller) ensureScheduledStatus(ctx context.Context, ns *corev1.Namespace, oldClusterName string) error {
+	// Set status on cluster change to bump the transition time
```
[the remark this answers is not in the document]
marun: I was intending to document each conditional that was being checked separately. Apparently my intent wasn't clear, so I'll revisit.

pkg/reconciler/namespace/namespace.go
marun: afaict there is nothing preventing us making a call to the server without these checks. What am I missing?

test/e2e/reconciler/namespace/controller_test.go
marun: I'm not sure I follow. How would this test be able to differentiate hard failure vs retryable failure from those codes? And what would be the value of doing so?

pkg/reconciler/namespace/namespace.go
stevekuznetsov: @ncdc might know -- but since you're building your own conditions adapter anyway it may not matter. I was under the impression that one of the big wins for the library was that it would handle deduping/re-setting.

test/e2e/reconciler/namespace/controller_test.go
stevekuznetsov: We're checking to response from the lister periodically (after events are received by the watcher) - if the intent of the test is to say "an object eventually exists with some state", ignoring 404 is strictly correct. Ignoring *all* errors may be more of a stretch, or it may not be - depends on what you intend.

imjasonh:
This change still LGTM, no serious concerns that should block it from merging and unblocking @ncdc's work on #439 at least. 👍

test/e2e/reconciler/namespace/controller_test.go
marun: I'm afraid I'm still missing your point. How do I determine whether an arbitrary error returned by the lister is fatal or not? And is there harm in assuming retry given that any pathological failure will hit a timeout and be reported as a test failure?

pkg/reconciler/namespace/namespace.go
```diff
@@ -100,55 +103,156 @@ func (c *Controller) reconcileGVR(ctx context.Context, gvr schema.GroupVersionRe
...
-	clusterName := ns.Labels[clusterLabel]
-	if clusterName == "" {
-		// Namespace is not assigned to a cluster; assign one.
-		// First, list all clusters.
-		cls, err := c.clusterLister.List(labels.Everything())
-		if err != nil {
-			return err
-		}
+// assignCluster attempts to schedule a namespace to a random cluster. If no cluster is
+// not available, any existing cluster assignment will be cleared.
```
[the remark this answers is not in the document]
marun: Done

test/e2e/reconciler/namespace/controller_test.go
ncdc: Fortunately for lister.Get, the only error you can get (today, at least) is a NotFound error.

I see what Steve is asking, but I don't think it's anything we can deal with (Maru's point about not knowing the severity of an error), and I think eventual convergence + hitting a timeout if we don't is good enough. WDYT?

pkg/reconciler/namespace/namespace.go
marun: Maybe you're thinking about the library setting the condition on a resource locally? Persisting that change still requires a server call (e.g. via `Patch`) and afaict there's no way to avoid that call without the kind of check you see here.

pkg/reconciler/namespace/conditions.go
marun: Done

pkg/reconciler/namespace/conditions.go
marun: Done

pkg/reconciler/namespace/namespace.go
marun: The check now ensures status is only updated if the namespace goes from scheduled -> not scheduled or vise versa.

pkg/reconciler/namespace/namespace.go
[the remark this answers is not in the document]
marun: This method now only checks for a change in scheduled state, PTAL.

marun:
Rebased and addressed review comments in a fixup commit to simplify review. Intend to squash before merge.

pkg/reconciler/namespace/namespace.go
marun: I apologize, I didn't read carefully. I've renamed `HasScheduledStatus` to `IsScheduled` as per your suggestion.

pkg/reconciler/namespace/namespace.go
stevekuznetsov: I assumed that the double-JSON-serialization and JSON Patch logic would be able to detect the no-op and short-circuit the call, so the code here (and elsewhere) can focus on declaration of intent - not imperative update-if-not-set logic

test/e2e/reconciler/namespace/controller_test.go
stevekuznetsov: I think my point is that "how you determine" is the code you write in this line - the other examples explicitly choose to ignore 404 and let state converge, but it is with your err check here that you state your preferred intent for the test - whether it is or is not fatal. It might be expected that a 403 or a 502 would be unexpected and fatal, or it might not - I was mostly making sure that we were clear in the fact that your choice here says that a string of requests 403 -> 502 -> 200 was OK for the client to see. Good point on what the lister can ever return though, it's clearly moot here

stevekuznetsov (approved):

marun:
Woops, neglected to merge. Next time...

pkg/reconciler/namespace/namespace.go
marun: @sttts Before I dig into code, does `Patch` detect no-op and thus avoid making unnecessary api calls?

PR #430, opened 2022-02-04
Speed-up CI

sttts:
- split test workflow into test and test-e2e
- limit concurrency to 2 for e2e

e2e tests with this run in 5:30min, instead of 7:50min before.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -80,6 +80,22 @@ jobs:
       - run: make build
       - run: PATH="${PATH}:$(pwd)/bin/" make test
 
+  e2e-single-run:
+    name: e2e-single-run
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: v1.17
+      - run: make build
+      - run: ARTIFACT_DIR=/tmp/e2e PATH="${PATH}:$(pwd)/bin/" COUNT=1 E2E_PARALLELISM=2 make test-e2e
+      - uses: actions/upload-artifact@v2
+        if: ${{ always() }}
+        with:
+          name: e2e-artifacts
+          path: /tmp/e2e/**/artifacts/
+
   e2e:
     name: e2e
     runs-on: ubuntu-latest
@@ -89,7 +105,7 @@ jobs:
         with:
           go-version: v1.17
       - run: make build
-      - run: ARTIFACT_DIR=/tmp/e2e PATH="${PATH}:$(pwd)/bin/" make test-e2e
+      - run: ARTIFACT_DIR=/tmp/e2e PATH="${PATH}:$(pwd)/bin/" COUNT=5 E2E_PARALLELISM=2 make test-e2e
       - uses: actions/upload-artifact@v2
         if: ${{ always() }}
         with:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -78,7 +78,18 @@ jobs:
         with:
           go-version: v1.17
       - run: make build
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -98,10 +98,11 @@ imports: $(OPENSHIFT_GOIMPORTS)
 	$(OPENSHIFT_GOIMPORTS) -m github.com/kcp-dev/kcp
 
 COUNT ?= 5
+E2E_PARALLELISM ?= 1
 
 .PHONY: test-e2e
 test-e2e: install
-	go test -tags e2e -race -count $(COUNT) ./test/e2e... $(WHAT)
+	go test -tags e2e -race -count $(COUNT) -p $(E2E_PARALLELISM) ./test/e2e... $(WHAT)
```
ncdc, line 105: Do you want to set `-parallel` too?

> -parallel n
    Allow parallel execution of test functions that call t.Parallel.
    The value of this flag is the maximum number of tests to run
    simultaneously; by default, it is set to the value of GOMAXPROCS.
    Note that -parallel only applies within a single test binary.
    The 'go test' command may run tests for different packages
    in parallel as well, according to the setting of the -p flag
    (see 'go help build').

sttts: that's `-p`, no?

Impressive how bad the go help is. I don't see `-p` there. Go the hint from some googling. So I assume it is `-parallel`.

ncdc: There's both `-p` and `-parallel`. `-p` is documented in `go help build`.

.github/workflows/ci.yaml
```diff
@@ -78,7 +78,18 @@ jobs:
...
+  e2e:
+    name: e2e
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: v1.17
+      - run: make build
+      - run: ARTIFACT_DIR=/tmp/e2e PATH="${PATH}:$(pwd)/bin/" COUNT=1 E2E_PARALLELISM=2 make test-e2e
```
stevekuznetsov, line 92: Why COUNT=1 ? NACK - we've already found flakes with COUNT=5

Makefile
sttts:
```
-p n
		the number of programs, such as build commands or
		test binaries, that can be run in parallel.
		The default is GOMAXPROCS, normally the number of CPUs available.
```
versus
-parallel n
	    Allow parallel execution of test functions that call t.Parallel, and
	    fuzz targets that call t.Parallel when running the seed corpus.
	    The value of this flag is the maximum number of tests to run
	    simultaneously.
	    While fuzzing, the value of this flag is the maximum number of
	    subprocesses that may call the fuzz function simultaneously, regardless of
	    whether T.Parallel is called.
	    By default, -parallel is set to the value of GOMAXPROCS.
	    Setting -parallel to values higher than GOMAXPROCS may cause degraded
	    performance due to CPU contention, especially when fuzzing.
	    Note that -parallel only applies within a single test binary.
	    The 'go test' command may run tests for different packages
	    in parallel as well, according to the setting of the -p flag
	    (see 'go help build').
```
So yes, `--parallel` needs to be set too. Wondering what applies then, the product of both?

.github/workflows/ci.yaml
sttts: running tests 5 times?

Can we at least split the e2e into one that does COUNT=1 and returns early, and another with COUNT=5 ?

I hate to wait.

sttts: e2e and e2e-quick

stevekuznetsov: How long does `-count 5` take and what is the comparison to unit tests? If it makes you wait more than 5 extra minutes, I'm more keen to try fewer counts...

sttts: Added benchmark commit. Let's see.

stevekuznetsov (approved):

PR #431, opened 2022-02-04
e2e: fix races in e2e framework

sttts:


test/e2e/framework/t.go
```diff
@@ -98,5 +100,24 @@ func (t *T) Wait() {
...
+			t.T.Helper()
+			t.T.Error(err)
+		default:
+			break loop
+		}
+	}
+}
+
+func (t *T) Failed() bool {
+	t.flush()
```
ncdc, line 118: Why is this needed?

test/e2e/framework/t.go
```diff
@@ -98,5 +100,24 @@ func (t *T) Wait() {
 	}
 }
 
+func (t *T) flush() {
+	t.T.Helper()
+loop:
+	for {
+		select {
+		case err := <-t.errors:
+			t.T.Helper()
```
ncdc, line 109: Does this do something in addition to the Helper call on L104?

test/e2e/framework/t.go
sttts: some go routine does `t.Errorf()` (e.g. the kcp starter in the framework), and calls `wg.Done`. The outer code waits for the wg and then checks `t.Failed()`. Today false is returned, although there is a clear order in time.

test/e2e/framework/t.go
sttts: Cargo culting the old code.

test/e2e/framework/t.go
sttts: This code:
```golang
			for _, srv := range servers {
				// binding the server to ctx ensures its lifetime is only
				// as long as the test we are running in this specific case
				if err := srv.Run(ctx); err != nil {
					t.Error(err)
					wg.Done()
				} else {
					go func(s *kcpServer) {
						defer wg.Done()
						if err := s.Ready(); err != nil {
							t.Errorf("kcp server %s never became ready: %v", s.name, err)
						}
					}(srv)
				}
				defer srv.GatherArtifacts()
			}
			wg.Wait()

			// if we've failed during startup, don't bother running the test
			if t.Failed() {
				return
			}
			t.Logf("Started kcp servers after %s", time.Since(start))
```

test/e2e/framework/t.go
sttts: fixed

test/e2e/framework/t.go
ncdc: Thanks, and what is the race you saw?

sttts: on timeout of the server startup the test func is called. It should not be called then. I.e.
```
			if t.Failed() {
				return
			}
```

test/e2e/framework/t.go
```diff
@@ -98,5 +100,23 @@ func (t *T) Wait() {
 	}
 }
 
+func (t *T) flush() {
+	t.T.Helper()
+loop:
```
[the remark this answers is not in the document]
sttts: your code blocks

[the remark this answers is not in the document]
stevekuznetsov: Whereas yours hot-loops on `loop:` and the `for{}`? I'm confused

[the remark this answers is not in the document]
ncdc: A for range on an open channel will only exit when the channel is closed, right?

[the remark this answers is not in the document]
stevekuznetsov: Right, but the code Stefan wrote either reads from the channel with a `select` (the same behavior) or breaks to the `loop` point ... which is before an infinite loop ... so it never exits? Am I crazy?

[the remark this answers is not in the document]
stevekuznetsov: Ah - can we just do a `return` instead of this stuff? Way easier to reason about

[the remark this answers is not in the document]
sttts: oh instead of break. Sure.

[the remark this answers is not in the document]
sttts: Break is evil. Has something of old GOTO days in Basic.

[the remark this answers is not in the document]
sttts: done

stevekuznetsov:
Based on the e2e failures here it seems we may have regressed on actually printing the error message :)

sttts:
@stevekuznetsov see second commit. I blindly added that flush to the cleanup funcs in the hope it might catch some similar issues on shutdown. Maybe it's not that obvious there.

stevekuznetsov:
Workspace shard test failures seem like bona-fide issues / reproducible

test/e2e/framework/t.go
ncdc: Got it, thanks!

ncdc:
Something about adding t.Cleanup(wrapped.flush) in NewT is causing the tests/contexts to get torn down prematurely and is making the tests fail. If you comment out that line, the tests pass.

ncdc:
Oh I see what it is. The goroutines we have monitoring livez and readyz are still running in the background, and when we shutdown the kcp server, they report failures to get livez and readyz using t.Errorf. Those errors were sent to the channel, we had already canceled the context by then, so the t.Wait call ignored them (because it honors a done context first). But adding flush as a test cleanup func now means we're checking for them, and seeing them.

sttts:
> But adding flush as a test cleanup func now means we're checking for them, and seeing them.

nice. So we were lucky before :)

ncdc (approved):

PR #434, opened 2022-02-04
syncer: Use Server-Side Apply to sync resources into downstream

astefanutti:
The syncer currently fails to update resources into downstream, when some fields are initialised by downstream controllers or admission webhooks, because it currently PUTs the whole upstream resources spec.  

One example is Service resources, for which unsetting `clusterIP` or `ipFamilies` is not allowed, and that cannot be synced downstream, once created:

```
E0204 16:41:06.117204   47047 specsyncer.go:155] Updating resource kcp--admin--myns/httpecho:
Service "httpecho" is invalid: [spec.clusterIPs[0]: Invalid value: []string(nil):
primary clusterIP can not be unset, spec.ipFamilies[0]:
Invalid value: []core.IPFamily(nil): primary ipFamily can not be unset]
E0204 16:41:06.907285   47047 syncer.go:318] Error reconciling key {{"" "v1" "services"}
%!q(*unstructured.Unstructured=&{map[apiVersion:v1 kind:Service
spec:map[ports:[map[name:http-port port:80 protocol:TCP targetPort:http-port]]
selector:map[app:echo-server]] status:map[loadBalancer:map[]]]})},
retrying... (#1): Service "httpecho" is invalid: [spec.clusterIPs[0]:
Invalid value: []string(nil):
primary clusterIP can not be unset,
spec.ipFamilies[0]: Invalid value: []core.IPFamily(nil): primary ipFamily can not be unset]
```

This PR changes the syncer to using SSA to apply upstream spec, without conflicting with downstream admission / validation.

### DIFF
--- a/pkg/syncer/specsyncer.go
+++ b/pkg/syncer/specsyncer.go
@@ -18,17 +18,20 @@ package syncer
 
 import (
 	"context"
+	"encoding/json"
 
 	"k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/pointer"
 )
 
 func deepEqualApartFromStatus(oldObj, newObj interface{}) bool {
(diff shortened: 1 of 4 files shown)


pkg/syncer/specsyncer.go
```diff
@@ -137,26 +140,16 @@ func (c *Controller) upsertIntoDownstream(ctx context.Context, gvr schema.GroupV
...
+	}
 
-		obj.SetResourceVersion(existing.GetResourceVersion())
-		if _, err := c.toClient.Resource(gvr).Namespace(namespace).Update(ctx, obj, metav1.UpdateOptions{}); err != nil {
-			klog.Errorf("Updating resource %s/%s: %v", namespace, obj.GetName(), err)
-			return err
-		}
-		return nil
+	if _, err := c.toClient.Resource(gvr).Namespace(namespace).Patch(ctx, obj.GetName(), types.ApplyPatchType, data, metav1.PatchOptions{FieldManager: syncerApplyManager, Force: pointer.Bool(true)}); err != nil {
+		klog.Errorf("Applying resource %s/%s: %v", namespace, obj.GetName(), err)
```
stevekuznetsov, line 149: nit: we should expose the metadata.clusterName for clarity here

stevekuznetsov (approved):
LGTM - @imjasonh PTAL?

pkg/syncer/specsyncer.go
astefanutti: `obj.GetClusterName()` is now added to the logging statements.

imjasonh (approved):

PR #435, opened 2022-02-04
Enable inheriting CRDs from helper.OrganizationCluster (admin)

ncdc:
CRD inheritance didn't support inheriting from helper.OrganizationCluster (admin) because it's not actually a Workspace
- it's only a logical cluster. Add in a hack that enables inheriting from helper.OrganizationCluster.

This PR also adds an environment variable to speed up e2e testing on Macs.

### DIFF
--- a/pkg/server/apiextensions.go
+++ b/pkg/server/apiextensions.go
@@ -92,18 +92,24 @@ func (c *inheritanceCRDLister) ListWithContext(ctx context.Context, selector lab
 		}
 
 		if workspace != nil && workspace.Spec.InheritFrom != "" {
-			// Make sure the source workspace exists
-			sourceWorkspaceKey := helper.WorkspaceKey(org, workspace.Spec.InheritFrom)
-			_, err := c.workspaceLister.Get(sourceWorkspaceKey)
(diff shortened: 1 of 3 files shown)


pkg/server/apiextensions.go
```diff
@@ -92,18 +92,24 @@ func (c *inheritanceCRDLister) ListWithContext(ctx context.Context, selector lab
...
+				switch {
+				case err == nil:
+					inheriting = true
+					inheritFrom = helper.EncodeOrganizationAndWorkspace(org, workspace.Spec.InheritFrom)
+				case apierrors.IsNotFound(err):
+					// A NotFound error is ok. It means we can't inherit but we should still proceed below to list.
+				default:
+					// Only error if there was a problem checking for workspace existence
+					return nil, err
+				}
```
sttts, line 112: ugly but 🤷‍♂️

sttts (approved):

pkg/server/apiextensions.go
ncdc: It's a quick hack....

PR #438, opened 2022-02-04
Makefile: fix WHAT not overriding the default package path

sttts:
`WHAT=./foo... make ...` should replace the default path

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -101,12 +101,14 @@ COUNT ?= 5
 E2E_PARALLELISM ?= 1
 
 .PHONY: test-e2e
+test-e2e: WHAT ?= ./test/e2e...
 test-e2e: install
-	go test -tags e2e -race -count $(COUNT) -p $(E2E_PARALLELISM) -parallel $(E2E_PARALLELISM) ./test/e2e... $(WHAT)
+	go test -tags e2e -race -count $(COUNT) -p $(E2E_PARALLELISM) -parallel $(E2E_PARALLELISM) $(WHAT)
 
 .PHONY: test
+test: WHAT ?= ./...
 test:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -100,12 +100,14 @@ imports: $(OPENSHIFT_GOIMPORTS)
 COUNT ?= 5
 
 .PHONY: test-e2e
+test-e2e: WHAT ?= ./test/e2e...
 test-e2e: install
-	go test -tags e2e -race -count $(COUNT) ./test/e2e... $(WHAT)
+	go test -tags e2e -race -count $(COUNT) $(WHAT)
 
 .PHONY: test
+test: WHAT ?= ./test/e2e...
```
stevekuznetsov, line 108: This should be `./...`

Makefile
```diff
@@ -100,12 +100,14 @@ imports: $(OPENSHIFT_GOIMPORTS)
 COUNT ?= 5
 
 .PHONY: test-e2e
+test-e2e: WHAT ?= ./test/e2e...
 test-e2e: install
-	go test -tags e2e -race -count $(COUNT) ./test/e2e... $(WHAT)
```
stevekuznetsov, line 104: Someone will be very surprised when they choose something other than e2e here and the build tags mean their `make` target does nothing :)

Makefile
sttts: indeed

sttts: fixed

Makefile
sttts: what do you suggest? Without this WHAT does not make sense either.

sttts: where exactly do we use the tag today?

PR #439, opened 2022-02-04
Enable multi-cluster for ns scheduler

ncdc:


pkg/reconciler/namespace/namespace.go
```diff
@@ -116,14 +115,21 @@ func (c *Controller) reconcileNamespace(ctx context.Context, lclusterName string
...
-		cls, err := c.clusterLister.List(labels.Everything())
+		allClusters, err := c.clusterLister.List(labels.Everything())
 		if err != nil {
 			return err
 		}
 
 		// TODO: Filter out un-Ready clusters so a namespace doesn't
 		// get assigned to an un-Ready cluster.
 
+		var cls []*v1alpha1.Cluster
```
stevekuznetsov, line 126: Why should we require that the `Cluster` is defined in the same logical cluster as the workloads that target it?

stevekuznetsov (commented):
Mostly LGTM - one question + you have compilation errors in test

pkg/reconciler/namespace/namespace.go
ncdc: I was assuming that because a `Cluster` lives in a `Workspace` (aka logical cluster), that you wouldn't want to assign something to a `Cluster` outside of the `Workspace` in question?

stevekuznetsov: I almost assumed the opposite - who is it that manages compute? If the use-case for a user being hermetically sealed into their own `Workspace` is that they get transparent multi-cluster and never think about it, I would expect some other external administrator to be the one determining `Cluster`s, `Location` membership, etc.

pkg/syncer/specsyncer.go
```diff
@@ -86,20 +87,40 @@ func (c *Controller) deleteFromDownstream(ctx context.Context, gvr schema.GroupV
...
+	b, err := json.Marshal(l)
+	if err != nil {
+		return err
+	}
+	newNamespace.SetAnnotations(map[string]string{
+		namespaceLocatorAnnotation: string(b),
+	})
+
+	if obj.GetLabels() != nil {
+		newNamespace.SetLabels(map[string]string{
```
imjasonh, line 119: `// TODO: this should be set once at syncer startup and propagated around everywhere.`

pkg/syncer/syncer.go
```diff
@@ -326,21 +328,18 @@ func (c *Controller) handleErr(err error, i interface{}) {
...
-		return "", "", fmt.Errorf("Failed to transform name: %s", namespace)
+func physicalClusterNamespaceName(l namespaceLocator) (string, error) {
+	b, err := json.Marshal(l)
+	if err != nil {
+		return "", err
 	}
-
-	return segments[1], segments[2], nil
+	hash := sha256.Sum224(b)
+	return fmt.Sprintf("kcp%x", hash), nil
```
imjasonh, line 342: Maybe comment why we do this prefixing?

imjasonh (approved):
LGTM, modulo rebasing on @marun 's change and fixing tests.

Thanks Andy! 👍

pkg/reconciler/namespace/namespace.go
ncdc: I think it's probably possible for someone/something (controller) to configure compute for a user in their Workspace, but I don't immediately see a path forward for assigning a resource from Workspace A to a Cluster from any random Workspace?

pkg/syncer/syncer.go
ncdc: I can remove unless you think it's useful?

pkg/syncer/specsyncer.go
```diff
@@ -86,20 +87,40 @@ func (c *Controller) deleteFromDownstream(ctx context.Context, gvr schema.GroupV
 	return c.toClient.Resource(gvr).Namespace(namespace).Delete(ctx, name, metav1.DeleteOptions{})
 }
 
+const namespaceLocatorAnnotation = "kcp.dev/namespace-locator"
+
 // TODO: This function is there as a quick and dirty implementation of namespace creation.
 //       In fact We should also be getting notifications about namespaces created upstream and be creating downstream equivalents.
-func (c *Controller) ensureNamespaceExists(ctx context.Context, namespace string) error {
+func (c *Controller) ensureNamespaceExists(ctx context.Context, namespace string, obj *unstructured.Unstructured) error {
```
marun, line 94: This function ensures that the namespace is created with the correct name and annotation, but does not maintain the annotation in the event that anything in the pcluster changes or removes it. If you don't intend to address that in this PR, maybe add a TODO?

pkg/informer/informer.go
```diff
@@ -161,30 +173,56 @@ func (d *DynamicDiscoverySharedInformerFactory) discoverTypes(ctx context.Contex
...
 	latest := sets.NewString()
-	rs, err := d.disco.ServerPreferredResources()
-	if err != nil {
-		return err
-	}
-	for _, r := range rs {
-		gv, err := schema.ParseGroupVersion(r.GroupVersion)
+	logicalClusterNames := sets.NewString()
+
+	// Assuming the workspaceLister is set, get a list of all the logical cluster names. We'll
```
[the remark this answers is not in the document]
ncdc: Valid questions! We definitely need to solidify our naming.

Given the admin logical cluster, when you create an org workspace, you have
- clusterName: admin
- name: myws

The API path prefix (URL segment) for resources in myws is `/clusters/admin_myws`. This is what I'm calling `logicalClusterName`. You'll notice that the name of the Workspace is `myws` but the "logical cluster name" for it is `admin_myws`. This represents both the URL prefix and the etcd key prefix for resources in this workspace.

I'm 100% open to coming up with a better name. But in this code, we're listing Workspaces and then collecting "logical cluster names" which are different from Workspace names.

pkg/informer/informer.go
```diff
@@ -161,30 +173,56 @@ func (d *DynamicDiscoverySharedInformerFactory) discoverTypes(ctx context.Contex
...
+			if err != nil {
+				return err
 			}
+			for _, ai := range r.APIResources {
+				if strings.Contains(ai.Name, "/") {
+					// foo/status, pods/exec, namespace/finalize, etc.
+					continue
+				}
+				if !ai.Namespaced {
+					// Ignore cluster-scoped things.
```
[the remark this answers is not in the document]
ncdc: This specific code was before I joined the project. I think `namespace/finalize` is excluded via L211 and `namespaces` and anything else cluster-scoped are excluded via L216 here.

If you want to see a smaller diff, take a look and have it hide whitespace differences - it's a lot smaller.

pkg/informer/informer.go
```diff
@@ -161,30 +173,56 @@ func (d *DynamicDiscoverySharedInformerFactory) discoverTypes(ctx context.Contex
...
+					continue
+				}
+				if !sets.NewString([]string(ai.Verbs)...).HasAll("list", "watch") {
+					klog.V(2).Infof("resource %s.%s %s is not list+watchable: %v", gv.Group, gv.Version, ai.Name, ai.Verbs)
+					continue
+				}
 
-			latest.Insert(strings.Join([]string{ai.Name, gv.Version, gv.Group}, "."))
+				latest.Insert(strings.Join([]string{ai.Name, gv.Version, gv.Group}, "."))
```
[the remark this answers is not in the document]
ncdc: Yes, because the dynamic shared informer factory is using a cross-logical cluster (`*`) client.

pkg/informer/informer.go
```diff
@@ -195,7 +233,7 @@ func (d *DynamicDiscoverySharedInformerFactory) discoverTypes(ctx context.Contex
 	for _, gvrstr := range newGVRs.UnsortedList() {
 		gvr, _ := schema.ParseResourceArg(gvrstr)
 		if gvr == nil {
-			multierr.AppendInto(&merr, fmt.Errorf("parsing GVR string %q: %w", gvrstr, err))
+			multierr.AppendInto(&merr, fmt.Errorf("Unable to parse %q to a GroupVersionResource", gvrstr))
```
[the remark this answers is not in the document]
ncdc: I don't think it's possible for an API to get into discovery and not be parseable, though. Also this code isn't tracking the GVR-lcluster relationship. Ok to add a TODO instead?

pkg/syncer/specsyncer.go
ncdc: Will add a TODO. There's a lot of code in here that could be cleaner.

pkg/reconciler/namespace/namespace.go
stevekuznetsov: ACK

stevekuznetsov:
```
# github.com/kcp-dev/kcp/pkg/reconciler/namespace [github.com/kcp-dev/kcp/pkg/reconciler/namespace.test]
Error: pkg/reconciler/namespace/namespace_test.go:119:5: cannot use dynClient (type *"k8s.io/client-go/dynamic/fake".FakeDynamicClient) as type dynamic.ClusterInterface in field value:
	*"k8s.io/client-go/dynamic/fake".FakeDynamicClient does not implement dynamic.ClusterInterface (missing Cluster method)
Error: pkg/reconciler/namespace/namespace_test.go:193:5: unknown field 'namespaceClient' in struct literal of type Controller
```

ncdc:
👍 re test compilation failures. I'm waiting to rebase on top of #428 before I tackle them.

pkg/informer/informer.go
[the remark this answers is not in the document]
marun: Ah, ok. So 'logical cluster name' is "[org name]_[workspace name]". Kind of like a resource name is `foo` but a namespace-qualified resource name is `namespace/foo`.  Maybe a parallel could be 'org-qualified workspace name'?

marun:
@ncdc #428 has merged.

ncdc:
Thanks! I'll maybe get to it tonight, or Monday

pkg/informer/informer.go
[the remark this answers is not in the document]
marun: Is a TODO still suggested here?

pkg/reconciler/namespace/namespace.go
```diff
@@ -137,26 +136,33 @@ func (c *Controller) updateSchedulableCondition(ctx context.Context, ns *corev1.
...
+	}
 
 	// TODO: Filter out un-Ready clusters so a namespace doesn't
 	// get assigned to an un-Ready cluster.
 
 	newClusterName := ""
 	if len(clusters) > 0 {
 		// Select a cluster at random.
-		cluster := clusters[rand.Intn(len(clusters))]
+		cluster := allClusters[rand.Intn(len(allClusters))]
```
marun, line 158: Is it your intention to filter the set of all clusters down to the set in the same logical cluster as the ns, but then assign from the set of all clusters?

ncdc:
@marun yes, thanks, will fix next!

pkg/reconciler/namespace/namespace.go
```diff
@@ -137,26 +136,33 @@ func (c *Controller) updateSchedulableCondition(ctx context.Context, ns *corev1.
...
 		newClusterName = cluster.Name
 	}
 
+	oldClusterName := ns.Labels[clusterLabel]
 	if oldClusterName != newClusterName {
 		klog.Infof("Patching to update cluster assignment for namespace %q|%q: %q -> %q", ns.ClusterName, ns.Name, oldClusterName, newClusterName)
-		if _, err := c.namespaceClient.Patch(ctx, ns.Name, types.MergePatchType, clusterLabelPatchBytes(newClusterName), metav1.PatchOptions{}); err != nil {
+		if _, err := c.kubeClient.Cluster(ns.ClusterName).CoreV1().Namespaces().Patch(ctx, ns.Name, types.MergePatchType, clusterLabelPatchBytes(newClusterName), metav1.PatchOptions{}); err != nil {
```
marun, line 165: (No action required) Nothing to do with this PR, but I realize that I'm a bit confused at the reuse of the concept of 'cluster name' for both 'the namespace exists in a workspace identified by clustername' and 'the namespace is scheduled to a physical cluster identified by clustername'. Given that there is no way around using `meta.ClusterName`, maybe differentiation of the second case is suggested.

pkg/reconciler/namespace/namespace.go
marun: @stevekuznetsov iirc we'll be introducing the concept of `ThingLocation` that is local to each workspace and hides the where and how of Cluster resources (i.e. a Cluster won't have to be in a given Workspace for resources in that Workspace to be scheduled to it).

pkg/reconciler/namespace/namespace.go
ncdc: "namespace is scheduled to a physical cluster identified by clustername" - clustername here should never be a meta.ClusterName, I don't think?

ncdc:
cluster/scheduler e2e updated (bonus: it found a bug in my code!)

imjasonh (approved):

pkg/syncer/syncer.go
marun: I think we want to retain this prefix so the provenance of a given namespace is obvious from its name (i.e. visible with `kubectl get <type>`) vs having to inspect individual objects.

In keeping with human-friendly UX, maybe insert a separator for readability (e.g. `kcp-<hash>`)?

marun:
@ncdc What (if any) change is suggested for the namespace scheduler e2e test to validate scheduling across more than one workspace?

marun:
LGTM

PR #440, opened 2022-02-04
Add kcp-config argument wrapper

kylape:
`usage: kcp-config [config file] [cmd] [args...]`

This allows kcp to be configured using a yaml file.  The yaml document needs to be a flat map (i.e. no hierarchy), and each key/value pair will be added as a command-line argument to kcp.

Example config.yaml:

    enable-controllers: true
    listen: localhost:6443

with this invocation:

    kcp-config config.yaml kcp start

will translate to:

    kcp start --enable-controllers=true --listen=localhost:6443

### DIFF
--- a/cmd/kcp-config/main.go
+++ b/cmd/kcp-config/main.go
@@ -1,0 +1,49 @@
+package main
+
+import (
+	"fmt"
+	"gopkg.in/yaml.v3"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"syscall"
+)
+
+func main() {
+	if len(os.Args) < 3 {
+		fmt.Printf("usage: kcp-config [config-file] [cmd] [args...]\n")
+		os.Exit(1)
+	}
+
+	fname := os.Args[1]
+
+	data, err := ioutil.ReadFile(fname)
+	if err != nil {
+		fmt.Printf("Error reading config file %s: %v\n", fname, err)
+		os.Exit(1)
+	}
+
+	contents := map[string]interface{}{}
+	err = yaml.Unmarshal([]byte(data), &contents)
+	if err != nil {
+		fmt.Printf("error: %v\n", err)
+		os.Exit(1)
+	}
+
+	args := os.Args[2:]
+	for k, v := range contents {
+		args = append(args, fmt.Sprintf("--%s=%v", k, v))
+	}
+
+	cmd, err := exec.LookPath(os.Args[2])
+
+	if err != nil {
+		fmt.Printf("kcp executable not found in $PATH\n")
+		os.Exit(1)
+	}
+
+	if err = syscall.Exec(cmd, args, os.Environ()); err != nil {
+		fmt.Printf("kcp exec() failed: %v\n", err)
+		os.Exit(1)
+	}
+}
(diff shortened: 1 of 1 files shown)


cmd/kcp-config/main.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+	"gopkg.in/yaml.v3"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"syscall"
+)
+
+func main() {
+	contents := map[string]interface{}{}
```
sttts, line 13: nit: move where used, just above Unmarshal

cmd/kcp-config/main.go
```diff
@@ -0,0 +1,49 @@
...
+		fmt.Printf("error: %v\n", err)
+		os.Exit(1)
+	}
+
+	args := os.Args[2:]
+	for k, v := range contents {
+		args = append(args, fmt.Sprintf("--%s=%v", k, v))
+	}
+
+	cmd, err := exec.LookPath(os.Args[2])
```
stevekuznetsov, line 38: When can this be anything other than `kcp`? Why is it user-configurable?

cmd/kcp-config/main.go
```diff
@@ -0,0 +1,49 @@
...
+
+	fname := os.Args[1]
+
+	data, err := ioutil.ReadFile(fname)
+	if err != nil {
+		fmt.Printf("Error reading config file %s: %v\n", fname, err)
+		os.Exit(1)
+	}
+
+	contents := map[string]interface{}{}
```
stevekuznetsov, line 26: We're passing it as a set of strings to the command, so why is this not `map[string]string`?

How would this allow passing a flag more than once?

stevekuznetsov (commented):
What sort of use-case is this supporting?

kylape:
I'm wanting to supply command line arguments from a `ConfigMap` [based on this conversation](https://kubernetes.slack.com/archives/C021U8WSAFK/p1643924314456659).

cmd/kcp-config/main.go
kylape: I'm thinking this wrapper could be used for any of the commands in this project that we'd want to run in a container (and have their args supplied in a `ConfigMap`).

cmd/kcp-config/main.go
kylape: The yaml document could have boolean or int types that get converted into strings.  We could probably add some type checking to ensure a flat hierarchy (i.e. no nested maps).

You currently can't pass more than one flag.  If we need that, perhaps we could consider a special delimiter for that?

stevekuznetsov:
I'm ambivalent. I am not sure I understand why passing flags is hard or why we are doing something different than existing k8s apiserver config (stored in etcd?)

kylape:
Passing flags isn't hard per se, but it seems intuitive to me to separate out the configuration of KCP from the more static pod spec.  I'm happy to use flags exclusively if that's the typical way k8s apiservers are managed.  If/when we write an operator to manage kcp, this will likely become a moot point anyways.

> why we are doing something different than existing k8s apiserver config (stored in etcd?)

I'm not sure exactly what you're referring to here.  Perhaps I'm not familiar with how k8s apiservers are normally configured.  Afaict it's just command-line arguments passed in via the container spec.

ncdc:
Discussed with @kylape on Slack - this is ok to close & we'll pass flags as needed in pod specs.

PR #442, opened 2022-02-07
e2e: flush error channel on termination

sttts:
Needs https://github.com/kcp-dev/kcp/pull/431.

### DIFF
--- a/test/e2e/framework/kcp.go
+++ b/test/e2e/framework/kcp.go
@@ -326,7 +326,7 @@ func (c *kcpServer) waitForEndpoint(client *rest.RESTClient, endpoint string) {
 			lastMsg = fmt.Sprintf("error contacting %s: %v", endpoint, err)
 		}
 	}, 100*time.Millisecond)
-	if !succeeded {
+	if !succeeded && c.ctx.Err() == nil {
 		c.t.Errorf(lastMsg)
 	}
 }
@@ -342,6 +342,9 @@ func (c *kcpServer) monitorEndpoint(client *rest.RESTClient, endpoint string) {
 	}
 	wait.UntilWithContext(ctx, func(ctx context.Context) {
 		_, err := rest.NewRequest(client).RequestURI(endpoint).Do(ctx).Raw()
+		if c.ctx.Err() != nil {
+			return
+		}
 		if err != nil {
 			c.t.Errorf("error contacting %s: %v", endpoint, err)
 		}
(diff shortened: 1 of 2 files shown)


test/e2e/framework/kcp.go
```diff
@@ -326,11 +326,20 @@ func (c *kcpServer) waitForEndpoint(client *rest.RESTClient, endpoint string) {
 			lastMsg = fmt.Sprintf("error contacting %s: %v", endpoint, err)
 		}
 	}, 100*time.Millisecond)
-	if !succeeded {
+	if !succeeded && !isDone(c.ctx) {
```
stevekuznetsov, line 329: Technically if we failed before the context was cancelled but executed this line after, this would eat the error. Not sure we can do better, though?

ncdc:
Turns out we have to fix both waitForEndpoint and monitorEndpoint. This fixes all the workspaceshard test failures at least.

```diff
diff --git a/test/e2e/framework/kcp.go b/test/e2e/framework/kcp.go
index 3be095a..9e8f769 100644
--- a/test/e2e/framework/kcp.go
+++ b/test/e2e/framework/kcp.go
@@ -351,6 +351,9 @@ func (c *kcpServer) monitorEndpoint(client *rest.RESTClient, endpoint string) {
 	}
 	wait.UntilWithContext(ctx, func(ctx context.Context) {
 		_, err := rest.NewRequest(client).RequestURI(endpoint).Do(ctx).Raw()
+		if isDone(ctx) {
+			return
+		}
 		if err != nil {
 			c.t.Errorf("error contacting %s: %v", endpoint, err)
 		}
```

test/e2e/framework/kcp.go
```diff
@@ -326,11 +326,20 @@ func (c *kcpServer) waitForEndpoint(client *rest.RESTClient, endpoint string) {
...
 		c.t.Errorf(lastMsg)
 	}
 }
 
+func isDone(ctx context.Context) bool {
```
stevekuznetsov, line 334: Nit: as per the Go language spec we should do `if ctx.Error() != nil`

stevekuznetsov (changes requested):

test/e2e/framework/kcp.go
sttts: fixed

ncdc:
LGTM. @stevekuznetsov?

stevekuznetsov (approved):

PR #444, opened 2022-02-07
Build and publish kcp image

csams:
Build a kcp image from main and publish it to ghcr if the CI workflow succeeds.

### DIFF
--- a/.github/workflows/kcp-image.yaml
+++ b/.github/workflows/kcp-image.yaml
@@ -22,13 +22,17 @@ jobs:
       with:
         go-version: v1.17
 
+    - name: Get the short sha
+      id: vars
+      run: echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+
     # Build and push a KCP image, tagged with latest and the commit SHA.
     - name: Build KCP Image
       id: build-image
       uses: redhat-actions/buildah-build@v2
       with:
         image: kcp
-        tags: latest ${{ github.sha }}
+        tags: latest ${{ steps.vars.outputs.sha_short }}
         containerfiles: |
           ./Dockerfile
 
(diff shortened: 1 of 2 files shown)


imjasonh:
Should we delete https://github.com/kcp-dev/kcp/blob/main/.github/workflows/syncer-image.yaml as well then? 

I don't have a strong opinion here, but using [`ko`](https://github.com/google/ko) here should be a bit faster, and easier to maintain than the equivalent Dockerfile and `docker build`.

`ko publish ./cmd/kcp --platform=all` builds on distroless/static:nonroot, for all five platforms it supports, without requiring a Dockerfile.

csams:
I'm fine with using `ko` later for the reasons you mention. We built on alpine for now so it's easier to exec into the container and dig around. I left syncer image as it is since it's the single binary and I didn't want to break anyone's current configs that depend on it. Do you want me to delete it and everyone can use the kcp image for syncer as well?

davidfestal:
BTW there will be the `virtual-workspaces` binary to build as well.

imjasonh:
> I'm fine with using `ko` later for the reasons you mention. We built on alpine for now so it's easier to exec into the container and dig around. I left syncer image as it is since it's the single binary and I didn't want to break anyone's current configs that depend on it. Do you want me to delete it and everyone can use the kcp image for syncer as well?

SGTM, whatever works best. I agree about not deleting syncer, at least for now.

FWIW, you can use `ko` to build on alpine with either `defaultBaseImage: alpine` in a `.ko.yaml` file, or the `KO_DEFAULTBASEIMAGE=alpine` env var at build time. It won't help if you want to build a single image with multiple executables, like you're doing now though.

imjasonh (approved):

PR #446, opened 2022-02-07
reconciler/cluster: fix startup races

sttts:
Informers and lister have to be created before shared informer factory start. This PR splits New() from Start().

### DIFF
--- a/cmd/cluster-controller/main.go
+++ b/cmd/cluster-controller/main.go
@@ -30,9 +30,14 @@ import (
 	crdexternalversions "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
 	"k8s.io/client-go/tools/clientcmd"
 
+	"github.com/kcp-dev/kcp/config"
+	apiresourceapi "github.com/kcp-dev/kcp/pkg/apis/apiresource"
+	clusterapi "github.com/kcp-dev/kcp/pkg/apis/cluster"
 	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
 	kcpexternalversions "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
 	"github.com/kcp-dev/kcp/pkg/reconciler/cluster"
+	apiextensionsv1client "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset/typed/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const resyncPeriod = 10 * time.Hour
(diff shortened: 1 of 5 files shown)


imjasonh (approved):

pkg/server/controllers.go
```diff
@@ -209,13 +211,41 @@ func (s *Server) installClusterController(clientConfig clientcmdapi.Config, serv
...
+	}
+
 	if err := server.AddPostStartHook("install-cluster-controller", func(context genericapiserver.PostStartHookContext) error {
 		if err := s.waitForCRDServer(context.StopCh); err != nil {
 			return err
 		}
 
-		ctx := goContext(context)
-		return s.cfg.ClusterControllerOptions.Complete(*kubeconfig, s.kcpSharedInformerFactory, s.apiextensionsSharedInformerFactory).Start(ctx)
+		// TODO(sttts): remove CRD creation from controller startup
```
stevekuznetsov, line 225: So the bizzarre thing is we're starting the informer before creating the types. Why exactly is this what we want?

pkg/server/controllers.go
```diff
@@ -209,13 +211,41 @@ func (s *Server) installClusterController(clientConfig clientcmdapi.Config, serv
...
+			crdClient := apiextensionsv1client.NewForConfigOrDie(logicalClusterConfig).CustomResourceDefinitions()
+			if err := config.BootstrapCustomResourceDefinitions(goContext(context), crdClient, requiredCrds); err != nil {
+				return err
+			}
+		}
+
+		s.kcpSharedInformerFactory.WaitForCacheSync(context.StopCh)
+		s.kubeSharedInformerFactory.WaitForCacheSync(context.StopCh)
+
+		go cluster.Start(goContext(context), c.NumThreads)
```
marun, line 245: (No action required) In the context of running these controllers in-process, would it make sense to package this controller initialization for reuse i.e. by test code vs just embedding it in the server? And maybe that suggests that responsibility for initialization should remain with the controller (i.e. separation of concerns)?

cmd/cluster-controller/main.go
```diff
@@ -91,10 +96,40 @@ func main() {
...
 	crdSharedInformerFactory := crdexternalversions.NewSharedInformerFactoryWithOptions(apiextensionsclient.NewForConfigOrDie(r), resyncPeriod)
-	if err := options.Options.Complete(kubeconfig, kcpSharedInformerFactory, crdSharedInformerFactory).Start(ctx); err != nil {
+	c := options.Options.Complete(kubeconfig, kcpSharedInformerFactory, crdSharedInformerFactory)
+	cluster, apiresource, err := c.New()
+	if err != nil {
 		fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
 
+	kcpSharedInformerFactory.Start(ctx.Done())
```
marun, line 106: This would appear to largely duplicate the init code in the server. Maybe factor out the common code for reuse?

pkg/server/controllers.go
ncdc: We do want the CRDs to exist first. But that's probably a follow-up to this? (It also applies to workspaces)

cmd/cluster-controller/main.go
sttts: shared informer factories IMO are primary responsibility of the caller, i.e. main.go or some other embedding code like the server. Would not overabstract this. There are only 3 factories anyway.

marun: ...and any test fixture that wants to start the controller in-process? Is there a good reason not to make the controller easy to start?

sttts: That asks for some framework-like thing to easily create factories and do the start+waitForSync right, rather than a specific helper.

PR #448, opened 2022-02-07
Remove e2e build tags

marun:
Requiring build tags for e2e tests complicates invoking tests via editor shortcuts. The same result of limiting the `make test` target to non-e2e files can be accomplished by path exclusion.


ncdc:
This seems ok too 😄

sttts (approved):

PR #453, opened 2022-02-07
Fix expression in kcp image workflow conditional

csams:


kylape (approved):
This workflow was opened in the Github Actions editor and it didn't have any validation errors.  Developing Github workflows against main is tricky since you can't run them unless you push them to main.

This fixes the issue reported in https://github.com/kcp-dev/kcp/actions/runs/1809276074/workflow

PR #454, opened 2022-02-08
e2e: Add kcp server fixture that allows t.Fatal

marun:
The current `RunParallel` implementation prevents the use of `t.Fatal` (and any library that depends on its use). It also couples test invocation to fixture use such that each test must use newly-instantiated fixture.

This PR proposes a simplified fixture that supports vanilla `*t.Testing` so that `t.Fatal` is permitted. This fixture allows tests to choose whether to instantiate fixture per-test (as per `RunParallel`) or to instantiate fixture once and reuse it across
multiple tests.

This PR also serves as a stepping-stone to enabling more expansive fixture reuse (e.g. across packages), up to and including unmanaged fixture (i.e. targeting a previously-deployed set of kcp servers).

The cluster controller e2e testing has been targeted for initial conversion to the new fixture.

### DIFF
--- a/test/e2e/framework/fixture.go
+++ b/test/e2e/framework/fixture.go
@@ -1,0 +1,93 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


test/e2e/framework/fixture.go
```diff
@@ -0,0 +1,93 @@
...
+	wg := sync.WaitGroup{}
+	wg.Add(len(f.rawServers))
+	for _, srv := range f.rawServers {
+		err := srv.Run(ctx)
+		require.NoError(t, err)
+
+		// Wait for the server to become ready
+		go func(s *kcpServer) {
+			err := s.Ready()
+			require.NoErrorf(t, err, "kcp server %s never became ready: %v", s.name, err)
```
sttts, line 75: why is this safe inside of the go routine?

sttts (approved):

test/e2e/framework/fixture.go
marun: It's not. Fix proposed in #460.

PR #457, opened 2022-02-08
syncer: cleanups

sttts:
- don't stop after 5 retries, avoid 10h outage: this is a time bomb :bomb:. We cannot stop retrying because we fail 5 times. If we don't like the load, there is a queue limiter to customize- e2e: actually test namespace creation in sink – no brainer
- retry when namespace is not yet in lister. Informers can have delay. We cannot assume we see everything immediately. Returning nil means to wait 10h.
- remove useless lister list check. Listing an unsynced lister has no sense.

Based on #454.

### DIFF
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -390,7 +390,7 @@ func (c *Controller) process(ctx context.Context, gvr schema.GroupVersionResourc
 		nsObj, err := nsInformer.Lister().Get(clusters.ToClusterAwareKey(meta.GetClusterName(), namespace))
 		if err != nil {
 			klog.Errorf("error listing namespace %q from the physical cluster: %v", namespace, err)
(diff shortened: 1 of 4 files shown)


pkg/syncer/syncer.go
```diff
@@ -316,14 +310,6 @@ func (c *Controller) handleErr(err error, i interface{}) {
...
 	}
 
-	// Re-enqueue up to 5 times.
-	num := c.queue.NumRequeues(i)
-	if num < 5 {
-		klog.Errorf("Error reconciling key %q, retrying... (#%d): %v", i, num, err)
-		c.queue.AddRateLimited(i)
-		return
-	}
-
```
sttts, line 326: this was a timebomb IMO

jmprusi (approved):

PR #459, opened 2022-02-08
test/e2e: add authz tests

s-urbaniak:
cc @sttts

### DIFF
--- a/test/e2e/authorizer/authorizer_test.go
+++ b/test/e2e/authorizer/authorizer_test.go
@@ -1,0 +1,240 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package authorizer
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"embed"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	kubeyaml "k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+
+	kcp "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	"github.com/kcp-dev/kcp/test/e2e/framework"
+)
+
+//go:embed *.yaml
+var embeddedResources embed.FS
+
+func createResources(ctx context.Context, filename string, client dynamic.Interface, decoder runtime.Decoder) error {
+	f, err := embeddedResources.Open(filename)
+	if err != nil {
+		return fmt.Errorf("failed to read resources: %w", err)
+	}
+
+	d := kubeyaml.NewYAMLReader(bufio.NewReader(f))
+	for {
+		doc, err := d.Read()
+		if errors.Is(err, io.EOF) {
+			break
+		} else if err != nil {
+			return err
+		}
+		if len(bytes.TrimSpace(doc)) == 0 {
+			continue
+		}
+		u := &unstructured.Unstructured{}
+		_, gvk, err := decoder.Decode(doc, nil, u)
+		if err != nil {
+			return fmt.Errorf("error decoding %s: %w", filename, err)
+		}
+
+		resource := gvk.Kind
+		resource = strings.ToLower(resource)
+		resource += "s"
+
+		gvr := schema.GroupVersionResource{
+			Group:    gvk.Group,
+			Version:  gvk.Version,
+			Resource: resource,
+		}
+
+		if _, err := client.Resource(gvr).Create(ctx, u, metav1.CreateOptions{}); err != nil {
+			return fmt.Errorf("error creating %s: %w", gvr, err)
+		}
+	}
+
+	return nil
+}
+
+type clients struct {
+	KubeClient kubernetes.Interface
+	KcpClient  kcp.Interface
+	Dynamic    dynamic.Interface
+}
+
+func newUserClient(t *testing.T, username, orgWorkspace, workspace string, cfg *rest.Config) *clients {
+	cfgCopy := rest.CopyConfig(cfg)
+	cfgCopy.Host = cfg.Host + "/clusters/" + orgWorkspace + "_" + workspace
+	cfgCopy.BearerToken = username + "-token"
+	kubeClient, err := kubernetes.NewForConfig(cfgCopy)
+	require.NoError(t, err)
+	kcpClient, err := kcp.NewForConfig(cfgCopy)
+	require.NoError(t, err)
+	dynamicClient, err := dynamic.NewForConfig(cfgCopy)
+	require.NoError(t, err)
+	return &clients{
+		KubeClient: kubeClient,
+		KcpClient:  kcpClient,
+		Dynamic:    dynamicClient,
+	}
+}
+
+func TestAuthorizer(t *testing.T) {
+	usersKCPArgs, err := framework.Users([]framework.User{
+		{
+			Name:   "user-1",
+			UID:    "1111-1111-1111-1111",
+			Token:  "user-1-token",
+			Groups: []string{"team-1"},
+		},
+		{
+			Name:   "user-2",
+			UID:    "1111-1111-1111-1111",
+			Token:  "user-2-token",
+			Groups: []string{"team-2"},
+		},
+		{
+			Name:   "user-3",
+			UID:    "1111-1111-1111-1111",
+			Token:  "user-3-token",
+			Groups: []string{"team-3"},
+		},
+	}).ArgsForKCP(t)
+	require.NoError(t, err)
+
+	f := framework.NewKCPFixture(framework.KcpConfig{
+		Name: "main",
+		Args: usersKCPArgs,
+	})
+	teardown := f.SetUp(t)
+	defer teardown()
+
+	ctx := context.Background()
+	if deadline, ok := t.Deadline(); ok {
+		withDeadline, cancel := context.WithDeadline(ctx, deadline)
+		t.Cleanup(cancel)
+		ctx = withDeadline
+	}
+	require.Equalf(t, len(f.Servers), 1, "incorrect number of servers")
+
+	server := f.Servers["main"]
+	kcpCfg, err := server.Config()
+	require.NoError(t, err)
+	clusterName, err := framework.DetectClusterName(kcpCfg, ctx, "workspaces.tenancy.kcp.dev")
+	require.NoError(t, err)
+	kubeClients, err := kubernetes.NewClusterForConfig(kcpCfg)
+	require.NoError(t, err)
+	kcpClients, err := kcp.NewClusterForConfig(kcpCfg)
+	require.NoError(t, err)
+	kcpClient := kcpClients.Cluster(clusterName)
+	require.NoError(t, err)
+	dynamicClient, err := dynamic.NewForConfig(kcpCfg)
+	require.NoError(t, err)
+
+	clients := map[string]*clients{
+		"admin": {
+			KubeClient: kubeClients.Cluster(clusterName),
+			KcpClient:  kcpClient,
+			Dynamic:    dynamicClient,
+		},
+		"user-1": newUserClient(t, "user-1", "admin", "workspace1", kcpCfg),
+		"user-2": newUserClient(t, "user-2", "admin", "workspace1", kcpCfg),
+		"user-3": newUserClient(t, "user-3", "admin", "workspace1", kcpCfg),
+	}
+
+	err = createResources(ctx, "resources.yaml", clients["admin"].Dynamic, serializer.NewCodecFactory(scheme.Scheme).UniversalDeserializer())
+	require.NoError(t, err)
+
+	_, err = clients["user-1"].KubeClient.CoreV1().Namespaces().Create(
+		ctx,
+		&v1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "default"}},
+		metav1.CreateOptions{},
+	)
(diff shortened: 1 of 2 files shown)


test/e2e/authorizer/authorizer_test.go
```diff
@@ -0,0 +1,417 @@
...
+		t.Error(err)
+		return
+	}
+
+	var testCases = []struct {
+		name string
+		work func(ctx context.Context, t framework.TestingTInterface, server testSetup)
+	}{
+		{
+			name: "create namespace and configmap as user-1",
```
marun, line 161: This test would appear to be a strict subset of the one that follows. Maybe remove?

test/e2e/authorizer/authorizer_test.go
```diff
@@ -0,0 +1,417 @@
...
+					metav1.GetOptions{},
+				)
+				if err != nil {
+					t.Errorf("failed to get configmap: %v", err)
+					return
+				}
+			},
+		},
+		{
+			name: "deny anything for user-3",
```
marun, line 251: Do the deny checks in this test case have to be executed in isolation from the success checks from the previous test case? Maybe add the deny checks to the previous test case?

marun (commented):
I've noticed a pattern in kcp e2e tests for a given controller  - many overlapping test cases whose need for isolation is questionable.  As per my inline comments, I think it would be preferable to aggregate related test operations into more monolithic 'golden-path' tests where isolation is not required.

Note also the recent introduction of a new test fixture that allows for simpler tests (`t.Fatal` and require.[assertion] rather than limited to `t.Error` and return) and fixture reuse:

https://github.com/kcp-dev/kcp/pull/454

s-urbaniak:
>  into more monolithic 'golden-path' tests where isolation is not required.

agreed, i recognized this pattern as well. These tests here don't need to run isolated, they definitely can run serially, sharing the same environment.

let me factor this using your recent changes :+1:

test/e2e/authorizer/authorizer_test.go
s-urbaniak: agreed, let's merge both tests.

test/e2e/authorizer/authorizer_test.go
s-urbaniak: no, let's merge into one test case.

stevekuznetsov:
I wanted a golden-path test but was explicitly asked not to do it by @sttts and @ncdc - perhaps they may want to chime in on why we went with the ones we did.

ncdc:
When we say golden-path, are we talking about 1 test case that tests a specific small/singular expected outcome (e.g. 1 field getting set correctly), or are we talking about the kcp server being configured to run a subset of controllers?

marun:
To me, 'golden-path' in the context of e2e means exercising functionality of a feature in a linear and representative way. The emphasis is on holistic testing the way a user would interact with the feature vs piecemeal isolated exercise of corner-cases.

ncdc (changes requested):
Here's a simplified way to load & decode yaml. @sttts can probably provide some better advice on how to go from GKV->GVR (I did a quick hack) and if there's an existing decoder we can use instead of creating our own scheme & codecs.

```diff
diff --git a/test/e2e/authorizer/authorizer_test.go b/test/e2e/authorizer/authorizer_test.go
index e2b5d7a..f22a0ff 100644
--- a/test/e2e/authorizer/authorizer_test.go
+++ b/test/e2e/authorizer/authorizer_test.go
@@ -1,6 +1,3 @@
-//go:build e2e
-// +build e2e
-
 /*
 Copyright 2022 The KCP Authors.
 
@@ -20,20 +17,26 @@ limitations under the License.
 package authorizer
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"embed"
 	"errors"
 	"fmt"
 	"io"
+	"strings"
 	"testing"
 
-	"gopkg.in/yaml.v3"
-
 	v1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
 	kubeyaml "k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
@@ -45,71 +48,44 @@ import (
 //go:embed *.yaml
 var embeddedResources embed.FS
 
-func createResources(ctx context.Context, filename string, kubeClient kubernetes.Interface, kcpClient kcp.Interface) error {
+func createResources(ctx context.Context, filename string, client dynamic.Interface, decoder runtime.Decoder) error {
 	f, err := embeddedResources.Open(filename)
 	if err != nil {
 		return fmt.Errorf("failed to read resources: %w", err)
 	}
 
-	d := yaml.NewDecoder(f)
+	d := kubeyaml.NewYAMLReader(bufio.NewReader(f))
 	for {
-		var node yaml.Node
-		err = d.Decode(&node)
-		if errors.Is(err, io.EOF) {
+		doc, err := d.Read()
+		if err == io.EOF {
 			break
+		} else if err != nil {
+			return err
 		}
-		if err != nil {
-			return fmt.Errorf("error decoding node %w", err)
+		if len(bytes.TrimSpace(doc)) == 0 {
+			continue
 		}
-
-		bytes, err := yaml.Marshal(&node)
+		u := &unstructured.Unstructured{}
+		// obj, gvk, err := unstructured.UnstructuredJSONScheme.Decode(bytes, nil, nil)
+		// obj, gvk, err := decoder.Decode(raw, nil, nil)
+		_, gvk, err := decoder.Decode(doc, nil, u)
 		if err != nil {
-			return fmt.Errorf("error marshaling node %w", err)
+			return fmt.Errorf("error decoding %s: %w", filename, err)
 		}
 
-		if len(node.Content) == 0 {
-			continue
+		resource := gvk.Kind
+		resource = strings.ToLower(resource)
+		resource += "s"
+
+		gvr := schema.GroupVersionResource{
+			Group:    gvk.Group,
+			Version:  gvk.Version,
+			Resource: resource,
 		}
+		resourceClient := client.Resource(gvr)
 
-		for _, c := range node.Content[0].Content {
-			switch c.Value {
-			case "WorkspaceShard":
-				var v tenancyv1alpha1.WorkspaceShard
-				if err := kubeyaml.Unmarshal(bytes, &v); err != nil {
-					return fmt.Errorf("failed to unmarshal WorkspaceShard: %w", err)
-				}
-				_, err := kcpClient.TenancyV1alpha1().WorkspaceShards().Create(ctx, &v, metav1.CreateOptions{})
-				if err != nil {
-					return fmt.Errorf("failed to create WorkspaceShard: %w", err)
-				}
-			case "Workspace":
-				var v tenancyv1alpha1.Workspace
-				if err := kubeyaml.Unmarshal(bytes, &v); err != nil {
-					return fmt.Errorf("failed to unmarshal Workspace: %w", err)
-				}
-				_, err := kcpClient.TenancyV1alpha1().Workspaces().Create(ctx, &v, metav1.CreateOptions{})
-				if err != nil {
-					return fmt.Errorf("failed to create WorkspaceShard: %w", err)
-				}
-			case "ClusterRole":
-				var v rbacv1.ClusterRole
-				if err := kubeyaml.Unmarshal(bytes, &v); err != nil {
-					return fmt.Errorf("failed to unmarshal ClusterRole: %w", err)
-				}
-				_, err := kubeClient.RbacV1().ClusterRoles().Create(ctx, &v, metav1.CreateOptions{})
-				if err != nil {
-					return fmt.Errorf("failed to create ClusterRole: %w", err)
-				}
-			case "ClusterRoleBinding":
-				var v rbacv1.ClusterRoleBinding
-				if err := kubeyaml.Unmarshal(bytes, &v); err != nil {
-					return fmt.Errorf("failed to unmarshal ClusterRoleBinding: %w", err)
-				}
-				_, err := kubeClient.RbacV1().ClusterRoleBindings().Create(ctx, &v, metav1.CreateOptions{})
-				if err != nil {
-					return fmt.Errorf("failed to create ClusterRoleBinding: %w", err)
-				}
-			}
+		if _, err := resourceClient.Create(ctx, u, metav1.CreateOptions{}); err != nil {
+			return fmt.Errorf("error creating %s: %w", filename, err)
 		}
 	}
 
@@ -120,6 +96,7 @@ func TestAuthorizer(t *testing.T) {
 	type clients struct {
 		KubeClient kubernetes.Interface
 		KcpClient  kcp.Interface
+		Dynamic    dynamic.Interface
 	}
 
 	type testSetup struct {
@@ -153,6 +130,17 @@ func TestAuthorizer(t *testing.T) {
 		return
 	}
 
+	scheme := runtime.NewScheme()
+	if err := rbacv1.AddToScheme(scheme); err != nil {
+		t.Error(err)
+		return
+	}
+	if err := tenancyv1alpha1.AddToScheme(scheme); err != nil {
+		t.Error(err)
+		return
+	}
+	codecs := serializer.NewCodecFactory(scheme)
+
 	var testCases = []struct {
 		name string
 		work func(ctx context.Context, t framework.TestingTInterface, server testSetup)
@@ -160,7 +148,7 @@ func TestAuthorizer(t *testing.T) {
 		{
 			name: "create namespace and configmap as user-1",
 			work: func(ctx context.Context, t framework.TestingTInterface, setup testSetup) {
-				createResources(ctx, "resources.yaml", setup.clients["admin"].KubeClient, setup.clients["admin"].KcpClient)
+				createResources(ctx, "resources.yaml", setup.clients["admin"].Dynamic, codecs.UniversalDeserializer())
 
 				_, err = setup.clients["user-1"].KubeClient.CoreV1().Namespaces().Create(
 					ctx,
@@ -186,7 +174,7 @@ func TestAuthorizer(t *testing.T) {
 		{
 			name: "allow view deny creation of resources for user-2",
 			work: func(ctx context.Context, t framework.TestingTInterface, server testSetup) {
-				createResources(ctx, "resources.yaml", server.clients["admin"].KubeClient, server.clients["admin"].KcpClient)
+				createResources(ctx, "resources.yaml", server.clients["admin"].Dynamic, codecs.UniversalDeserializer())
 
 				_, err = server.clients["user-1"].KubeClient.CoreV1().Namespaces().Create(
 					ctx,
@@ -250,7 +238,7 @@ func TestAuthorizer(t *testing.T) {
 		{
 			name: "deny anything for user-3",
 			work: func(ctx context.Context, t framework.TestingTInterface, server testSetup) {
-				createResources(ctx, "resources.yaml", server.clients["admin"].KubeClient, server.clients["admin"].KcpClient)
+				createResources(ctx, "resources.yaml", server.clients["admin"].Dynamic, codecs.UniversalDeserializer())
 
 				_, err = server.clients["user-1"].KubeClient.CoreV1().Namespaces().Create(
 					ctx,
@@ -359,6 +347,11 @@ func TestAuthorizer(t *testing.T) {
 				return
 			}
 
+			dynamicClient, err := dynamic.NewForConfig(kcpCfg)
+			if err != nil {
+				t.Errorf("failed to construct dynamic client for server: %w", err)
+			}
+
 			newUserClient := func(username, orgWorkspace, workspace string) *clients {
 				config := rest.CopyConfig(kcpCfg)
 				config.Host = kcpCfg.Host + "/clusters/" + orgWorkspace + "_" + workspace
@@ -372,7 +365,15 @@ func TestAuthorizer(t *testing.T) {
 				if err != nil {
 					t.Errorf("failed to construct kcp client for server: %w", err)
 				}
-				return &clients{kubeClient, kcpClient}
+				dynamicClient, err := dynamic.NewForConfig(config)
+				if err != nil {
+					t.Errorf("failed to construct dynamic client for server: %w", err)
+				}
+				return &clients{
+					KubeClient: kubeClient,
+					KcpClient:  kcpClient,
+					Dynamic:    dynamicClient,
+				}
 			}
 
 			testCase.work(ctx, t, testSetup{
@@ -382,6 +383,7 @@ func TestAuthorizer(t *testing.T) {
 					"admin": {
 						KubeClient: kubeClients.Cluster(clusterName),
 						KcpClient:  kcpClient,
+						Dynamic:    dynamicClient,
 					},
 					"user-1": newUserClient("user-1", "admin", "workspace1"),
 					"user-2": newUserClient("user-2", "admin", "workspace1"),
@@ -390,7 +392,7 @@ func TestAuthorizer(t *testing.T) {
 			})
 		}, framework.KcpConfig{
 			Name: serverName,
-			Args: append([]string{"--install-workspace-scheduler"}, usersKCPArgs...),
+			Args: usersKCPArgs,
 		})
 	}
 }
```

test/e2e/authorizer/authorizer_test.go
```diff
@@ -0,0 +1,417 @@
+//go:build e2e
+// +build e2e
```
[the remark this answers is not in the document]
s-urbaniak: ack, removed

s-urbaniak:
@ncdc wonderful, thank you 👏 using underlying `Decoder` is indeed a much better approach, i was brainstorming with @sttts already but the spark didn't fly at that moment, took over the suggestion 1:1.

s-urbaniak:
> I don't think you'll need to create a scheme and codecs since this is just decoding to Unstructured. I think there might be an unstructured decoder you can use. There also might be a simpler way to interact with yaml decoding - check the other types and functions in that package.

checking now 👍 

@marun do you mind to PTAL if this e2e now looks more idiomatic?

test/e2e/authorizer/authorizer_test.go
```diff
@@ -0,0 +1,249 @@
...
+	require.NoError(t, err)
+	dynamicClient, err := dynamic.NewForConfig(kcpCfg)
+	require.NoError(t, err)
+
+	scheme := runtime.NewScheme()
+	err = rbacv1.AddToScheme(scheme)
+	require.NoError(t, err)
+	err = tenancyv1alpha1.AddToScheme(scheme)
+	require.NoError(t, err)
+	codecs := serializer.NewCodecFactory(scheme)
```
ncdc, line 175: I think you can delete this and pass `unstructured.UnstructuredJSONScheme` below

test/e2e/authorizer/authorizer_test.go
```diff
@@ -0,0 +1,249 @@
...
+			Groups: []string{"team-3"},
+		},
+	}).ArgsForKCP(t)
+	require.NoError(t, err)
+
+	f := framework.NewKCPFixture(framework.KcpConfig{
+		Name: "main",
+		Args: usersKCPArgs,
+	})
+	teardown := f.SetUp(t)
```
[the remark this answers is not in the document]
s-urbaniak: ack, i was literally confused by the double invocation. separating it was just for readability.

test/e2e/authorizer/authorizer_test.go
s-urbaniak: hm, i need to recheck, but changing to this one fails e2e locally 🤔

s-urbaniak: we can omit scheme registration though and reference `k8s.io/client-go/kubernetes/scheme#Scheme` directly, as types register here, for us in https://github.com/kcp-dev/kcp/blob/5887e9b6d59dc5fcbc75fb22f964cfc45ba43a5a/test/e2e/framework/util.go#L96-L98.

(sideeffects, sideeffects, sideffects, everywhere 😬 )

sttts (approved):

sttts:
@ncdc dismissing your change request. @s-urbaniak says it's all addressed.

PR #460, opened 2022-02-08
Ensure new fixture fails properly if kcp servers do not become ready

marun:
#454 blithely ignored goroutine `t.Fatal` semantics, ensuring that any failure of a kcp server to start would stall the test run rather than failing fast. This PR ensures timely failure.

### DIFF
--- a/test/e2e/framework/fixture.go
+++ b/test/e2e/framework/fixture.go
@@ -71,12 +71,16 @@ func (f *KCPFixture) SetUp(t *testing.T) func() {
 
 		// Wait for the server to become ready
 		go func(s *kcpServer) {
+			defer wg.Done()
 			err := s.Ready()
 			require.NoErrorf(t, err, "kcp server %s never became ready: %v", s.name, err)
-			wg.Done()
 		}(srv)
 	}
 	wg.Wait()
+
+	if t.Failed() {
+		t.Fatal("Fixture setup failed: one or more servers did not become ready")
+	}
 
 	t.Logf("Started kcp servers after %s", time.Since(start))
 


test/e2e/framework/fixture.go
```diff
@@ -71,13 +71,17 @@ func (f *KCPFixture) SetUp(t *testing.T) func() {
 
 		// Wait for the server to become ready
 		go func(s *kcpServer) {
+			defer wg.Done()
 			err := s.Ready()
 			require.NoErrorf(t, err, "kcp server %s never became ready: %v", s.name, err)
-			wg.Done()
 		}(srv)
 	}
 	wg.Wait()
```
sttts, line 79: Ready() will look 60s for the admin kubeconfig. Can we move that logic into the Run func and terminate the test immediately when the server dies?

test/e2e/framework/fixture.go
```diff
@@ -71,13 +71,17 @@ func (f *KCPFixture) SetUp(t *testing.T) func() {
...
 
+	if t.Failed() {
+		t.Fatal("Fixture setup failed: one or more servers did not become ready")
```
sttts, line 82: do we see the error?

marun: Yes. The `require` call will log an error and mark the `testing.T` as failed. But since it's called in a goroutine it's panic won't propagate. Checking `t.Failed` here ensures that the parent can die in response to the assertion failure in the goroutine.

sttts: ack

sttts (approved):

PR #464, opened 2022-02-08
nss: Refactor namespace scheduler e2e to use new fixture

marun:


test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -39,126 +41,93 @@ const clusterLabel = "kcp.dev/cluster"
...
-			t.Errorf("incorrect number of servers: %d", len(servers))
-			return
-		}
-		server := servers[serverName]
-		cfg, err := server.Config()
-		if err != nil {
-			t.Error(err)
-			return
-		}
+	require.Equalf(t, len(f.Servers), 1, "incorrect number of servers")
```
stevekuznetsov, line 64: We've got a couple competing ideas about what to do here:

 - I use the Go standard library
 - you use testify
 - @davidfestal used some assertion library

Maybe let's find a common path and not simply introduce extra ones...?

imjasonh: FWIW, my personal vote is the Go standard library.

I find `require.Equal` is too easy to misuse, swapping `expected` and `actual` (as here).

marun: I'm using the same library as @davidfestal - it's testify in both cases.  And testify is a superset of what the go standard library does. Anyone that wants to laboriously check error codes and raise assertions is free to do so, arguably at the expense of readability.

ncdc: But we should agree as a community on a standard

marun:
Added back `t.Parallel` to ensure the single test can run in parallel with other packages.

marun:
Supplanted by #514, which updates all tests in a single PR.

PR #468, opened 2022-02-08
Simplify cluster controller e2e to run as a single test

marun:
This avoids the immediate problem of kcpServer fixture helpers expecting to be local to a test executed in parallel.

For tests that do want to execute in parallel, test setup for shared fixture will need to be performed outside of tests configured with `t.Parallel` to ensure that side-effects of setup (e.g. registration of functions that gather artifacts) are persistent across tests.

cc: @s-urbaniak

### DIFF
--- a/test/e2e/reconciler/cluster/controller_test.go
+++ b/test/e2e/reconciler/cluster/controller_test.go
@@ -62,78 +62,13 @@ const clusterName = "us-east1"
 const sourceClusterName, sinkClusterName = "source", "sink"
 
 func TestClusterController(t *testing.T) {
+	t.Parallel()
+
 	type runningServer struct {
 		framework.RunningServer
 		client     wildwestclient.WildwestV1alpha1Interface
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
This doesn't seem like a great way forward ... we certainly have tests running in parallel for other suites, and having them be disparate just because the fixture can't handle it (today) ... /shrug

marun:
> This doesn't seem like a great way forward ... we certainly have tests running in parallel for other suites, and having them be disparate just because the fixture can't handle it (today) ... /shrug

I am in no way suggesting that this PR represents the end-state for how we want to write tests. The current strategy of per-test fixture is also not sustainable given that deploying a set of kcp servers takes ~4s on a fast machine today and that duration is likely to increase over time. The goal here is evolving our tests to minimize the overhead represented by test fixture and that tests and their fixture become separable concerns that can evolve independently.

marun:
The reliability issues with the fixture are addressed instead by #501.

PR #478, opened 2022-02-09
Add ingress-controller to kcp.

jmprusi:
This PR adds the ingress-controller (https://github.com/jmprusi/kcp-ingress) into kcp repo.


pkg/envoy/cache.go
```diff
@@ -0,0 +1,106 @@
...
+	networkingv1 "k8s.io/api/networking/v1"
+	"k8s.io/klog/v2"
+)
+
+const (
+	defaultCleanupInterval = 1 * time.Minute
+	NodeID                 = "kcp-ingress"
+)
+
+type Cache struct {
```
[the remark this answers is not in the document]
jmprusi: It's a cache used for envoy configuration. Holds all the desired ingresses and then translates it into envoy config when desired (ToEnvoySnapshot).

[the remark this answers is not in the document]
sttts: used by whom? why public?

pkg/envoy/cache.go
```diff
@@ -0,0 +1,106 @@
+/*
```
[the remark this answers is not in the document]
sttts: would be more specific what this is. Is it the ingress controller plumbing code? Is it envoy launching code?

pkg/envoy-controlplane/controlplane.go
```diff
@@ -0,0 +1,51 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package envoycontrolplane
+
+import networkingv1 "k8s.io/api/networking/v1"
+
+type EnvoyControlPlane struct {
```
sttts, line 21: what is this from a high level? A server? Controllers?

pkg/reconciler/ingress/tracker.go
```diff
@@ -0,0 +1,87 @@
...
+import (
+	"sync"
+
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	"k8s.io/klog/v2"
+)
+
+// TODO(jmprusi): Generalize.
+type Tracker struct {
```
sttts, line 28: who needs tracker? Does it have to be public?

pkg/reconciler/ingress/tracker.go
```diff
@@ -0,0 +1,95 @@
...
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	klog.Infof("tracking service %q for ingress %q", s.Name, ingress.Name)
+	for _, ti := range t.trackedServices[serviceToKey(s)] {
+		if ingressToKey(&ti) == ingressToKey(ingress) {
+			return
+		}
+	}
+	t.trackedServices[serviceToKey(s)] = append(t.trackedServices[serviceToKey(s)], *ingress)
+	t.ingressToServices[ingressToKey(ingress)] = make(map[string]struct{})
```
ncdc, line 64: Can you ever have one ingress map to multiple services? If so, this appears to clear out the map each time this function is called?

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+)
+
+const (
+	clusterLabel = "kcp.dev/cluster"
+	ownedByLabel = "kcp.dev/owned-by"
+)
+
+// reconcile is triggered on every change to an ingress resource, or it's associated services (by tracker).
+func (c *Controller) reconcile(ctx context.Context, ingress *networkingv1.Ingress) error {
+	klog.Infof("reconciling Ingress %q", ingress.Name)
```
ncdc, line 41: I'd recommend logging ingress.ClusterName, ingress.Namespace, and ingress.Name. Can also use InfoS for structured logging if you'd like (I'd like us to be consistent, which we're not currently, but structured logging should be our preferred approach going forward).

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+
+	if ingress.Labels == nil || ingress.Labels[clusterLabel] == "" {
```
ncdc, line 43: You don't strictly need to check for a nil map - `ingress.Labels[clusterLabel] == ""` should suffice

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+		// This is a root Ingress; get its leafs.
+		sel, err := labels.Parse(fmt.Sprintf("%s=%s", ownedByLabel, ingress.Name))
+		if err != nil {
+			return err
```
ncdc, line 47: Log & return nil - this will be a permanent failure and will go into rate limited infinite backoff if you return the error.

Please add a TODO to set a user-visible condition describing the error

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+
+		// Get the current Leaves
+		currentLeaves, err := c.lister.List(sel)
+		if err != nil {
+			return err
```
ncdc, line 53: It's basically impossible to get an error from a lister, but if you ever did, you'd want to log, set a user-facing condition, and return nil

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+
+	return hashString(ingress.Name+ingress.Namespace+ingress.ClusterName) + "." + *domain
+}
+
+// getServices will parse the ingress object and return a list of the services.
+func (c *Controller) getServices(ctx context.Context, ingress *networkingv1.Ingress) ([]*corev1.Service, error) {
+	var services []*corev1.Service
+	for _, rule := range ingress.Spec.Rules {
+		for _, path := range rule.HTTP.Paths {
+			svc, err := c.client.CoreV1().Services(ingress.Namespace).Get(ctx, path.Backend.Service.Name, metav1.GetOptions{})
```
ncdc, line 252: Please use a serviceLister instead of a live client

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var clusterDests []string
+	for _, service := range services {
+		if service.Labels[clusterLabel] != "" {
+			clusterDests = append(clusterDests, service.Labels[clusterLabel])
+		} else {
+			klog.Infof("Skipping service %q because it is not assigned to any cluster", service.Name)
```
ncdc, line 182: Do you need a `continue` here? Or is it appropriate to add this service to the tracker even though it's not labeled?

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+// desiredLeaves returns a list of leaves (ingresses) to be created based on an ingress.
+func (c *Controller) desiredLeaves(ctx context.Context, root *networkingv1.Ingress) ([]*networkingv1.Ingress, error) {
+	// This will parse the ingresses and extract all the destination services,
+	// then create a new ingress leaf for each of them.
+	services, err := c.getServices(ctx, root)
+	if err != nil {
+		return nil, err
+	}
+
+	var clusterDests []string
```
ncdc, line 177: Do you want to use a `sets.String` here for uniqueness?

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+			Hostname: "",
+		}}
+		return nil, nil
+	}
+
+	desiredLeaves := make([]*networkingv1.Ingress, 0, len(clusterDests))
+	for _, cl := range clusterDests {
+		vd := root.DeepCopy()
+		// TODO: munge cluster name
+		vd.Name = fmt.Sprintf("%s--%s", root.Name, cl)
```
ncdc, line 202: Let's resolve this now if you don't mind. We have the cluster name in the labels. We can generate a random name (e.g. uuid), or we can base it off of root.Name by e.g. setting GenerateName to `<base>-` where `<base>` is root.Name trimmed (if needed) so it doesn't exceed <max allowed name length - generated suffix length>, if that makes sense?

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+
+		vd.Labels = map[string]string{}
+		vd.Labels[clusterLabel] = cl
+		vd.Labels[ownedByLabel] = root.Name
```
ncdc, line 206: I don't think root.Name is enough to guarantee uniqueness across namespaces and logical clusters. I think we need to set this to the output of cache.MetaNamespaceKeyFunc(root.Name), right?

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func findNonDesiredLeaves(current, desired []*networkingv1.Ingress) []*networkingv1.Ingress {
+	var missing []*networkingv1.Ingress
+
+	for _, c := range current {
+		found := false
+		for _, d := range desired {
+			if c.Name == d.Name {
+				found = true
```
ncdc, line 270:
```suggestion
				found = true
				break
```

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+		// Generate the desired leaves
+		desiredLeaves, err := c.desiredLeaves(ctx, ingress)
+		if err != nil {
+			return err
+		}
+
+		// Clean the leaves that are not desired anymore
+		for _, leaftoremove := range findNonDesiredLeaves(currentLeaves, desiredLeaves) {
+			klog.Infof("Deleting non desired leaf %q", leaftoremove.Name)
+			if err := c.client.NetworkingV1().Ingresses(leaftoremove.Namespace).Delete(ctx, leaftoremove.Name, metav1.DeleteOptions{}); err != nil {
```
ncdc, line 65: Will need to add in `.Cluster(leafToRemove.ClusterName)`

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+
+		// Generate the desired leaves
+		desiredLeaves, err := c.desiredLeaves(ctx, ingress)
+		if err != nil {
+			return err
+		}
+
+		// Clean the leaves that are not desired anymore
+		for _, leaftoremove := range findNonDesiredLeaves(currentLeaves, desiredLeaves) {
```
ncdc, line 63:
```suggestion
		for _, leafToRemove := range findNonDesiredLeaves(currentLeaves, desiredLeaves) {
```

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+			klog.Infof("Deleting non desired leaf %q", leaftoremove.Name)
```
ncdc, line 64: Please log clusterName + namespace + name

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+			if err := c.client.NetworkingV1().Ingresses(leaftoremove.Namespace).Delete(ctx, leaftoremove.Name, metav1.DeleteOptions{}); err != nil {
+				return err
```
ncdc, line 66: TODO set a condition on the root

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+			}
+		}
+
+		// TODO(jmprusi): ugly. fix. use indexer, etc.
```
ncdc, line 70: By this do you mean check the indexer to see if it exists & is up to date, and then update/create as needed? If so, 👍 please make the necessary changes

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+					existingLeaf, err := c.client.NetworkingV1().Ingresses(desiredleaf.Namespace).Get(ctx, desiredleaf.Name, metav1.GetOptions{})
+					if err != nil {
+						return err
+					}
+
+					// Set the resourceVersion and UID to update the desired leaf.
+					desiredleaf.ResourceVersion = existingLeaf.ResourceVersion
+					desiredleaf.UID = existingLeaf.UID
+
+					if _, err := c.client.NetworkingV1().Ingresses(desiredleaf.Namespace).Update(ctx, desiredleaf, metav1.UpdateOptions{}); err != nil {
```
ncdc, line 84: A Patch() call is safer. Can put it in as a TODO if you'd like.

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+		// If the ingress has the clusterLabel set, that means that it is a leaf and it's synced with
+		// a cluster.
+		//
+		// This update can come from the creation or because the syncer has update the status.
+
+		rootIngressName := ingress.Labels[ownedByLabel]
+		// A leaf Ingress was updated; get others and aggregate status.
+		sel, err := labels.Parse(fmt.Sprintf("%s=%s", ownedByLabel, rootIngressName))
+		if err != nil {
+			return err
```
ncdc, line 104: Log, TODO user-facing condition, return nil

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+		others, err := c.lister.List(sel)
+		if err != nil {
+			return err
```
ncdc, line 108: Log, TODO user-facing condition, return nil

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+
+		// Get the rootIngress based on the labels.
+		var rootIngress *networkingv1.Ingress
+
+		rootIf, exists, err := c.indexer.Get(&networkingv1.Ingress{
```
ncdc, line 114: If you make the change so the value of the ownedByLabel is the full key, you can do c.indexer.GetByKey(rootIngressName) here.

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+				ClusterName: ingress.GetClusterName(),
+			},
+		})
+		if err != nil {
+			return err
+		}
+
+		// TODO(jmprusi): A leaf without rootIngress?
+		if !exists {
+			return fmt.Errorf("Root Ingress not found: %s", rootIngressName)
```
ncdc, line 127: Log, TODO user-facing condition (on the leaf?), return nil

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+		}
+
+		rootIngress = rootIf.(*networkingv1.Ingress).DeepCopy()
+
+		// Clean the current status, and then recreate if from the other leafs.
```
ncdc, line 132:
```suggestion
		// Clean the current status, and then recreate it from the other leafs.
```

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+
+			statusHost := generateStatusHost(c.domain, rootIngress)
+			// Now overwrite the Status of the rootIngress with our desired LB
+			rootIngress.Status.LoadBalancer.Ingress = []corev1.LoadBalancerIngress{{
+				Hostname: statusHost,
+			}}
+		}
+
+		// Update the rootIngress status with our desired LB.
+		if _, err := c.client.NetworkingV1().Ingresses(rootIngress.Namespace).UpdateStatus(ctx, rootIngress, metav1.UpdateOptions{}); err != nil {
```
ncdc, line 153: .Cluster(rootIngress.ClusterName) and TODO use PatchStatus

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+			if errors.IsConflict(err) {
+				key, err := cache.MetaNamespaceKeyFunc(ingress)
+				if err != nil {
+					return err
+				}
+				c.queue.AddRateLimited(key)
+				return nil
+			}
+			return err
```
ncdc, line 162: You should just be able to `return err` here - in both flows, the key will be added with rate limiting.

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+	h.Write([]byte(s))
+	return fmt.Sprint(h.Sum32())
+}
+
+// generateStatusHost returns a string that represent the desired status hostname for the ingress.
+// If the host is part of the same domain, it will be preserved as the status hostname, if not
+// a new one will be generated based on a hash of the ingress name, namespace and clusterName.
+func generateStatusHost(domain *string, ingress *networkingv1.Ingress) string {
+
+	// TODO(jmprusi): using "contains" is a bad idea as it could be abused by crafting a malicious hostname, but for a PoC it should be good enough?
```
ncdc, line 230: What's the right thing to do here?

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+		vd.SetResourceVersion("")
+
+		desiredLeaves = append(desiredLeaves, vd)
+	}
+
+	return desiredLeaves, nil
+}
+
+func hashString(s string) string {
+	h := fnv.New32a()
```
ncdc, line 220: What made you choose this algorithm?

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,279 @@
...
+	h.Write([]byte(s))
+	return fmt.Sprint(h.Sum32())
+}
+
+// generateStatusHost returns a string that represent the desired status hostname for the ingress.
+// If the host is part of the same domain, it will be preserved as the status hostname, if not
+// a new one will be generated based on a hash of the ingress name, namespace and clusterName.
+func generateStatusHost(domain *string, ingress *networkingv1.Ingress) string {
+
```
ncdc, line 229:
```suggestion
```

pkg/envoy-controlplane/configcache.go
```diff
@@ -0,0 +1,109 @@
...
+type configCache struct {
+	mu         sync.Mutex
+	ingresses  *gocache.Cache
+	translator *translator
+}
+
+func newConfigCache(translator *translator) *configCache {
+	return &configCache{
+		mu:         sync.Mutex{},
+		ingresses:  gocache.New(gocache.NoExpiration, defaultCleanupInterval),
```
ncdc, line 52: Do you anticipate using a nonzero expiration ever? If not, can we use a normal `map`?

pkg/envoy-controlplane/translator.go
```diff
@@ -0,0 +1,226 @@
...
+	"google.golang.org/protobuf/types/known/anypb"
+	"google.golang.org/protobuf/types/known/durationpb"
+	"google.golang.org/protobuf/types/known/wrapperspb"
+
+	networkingv1 "k8s.io/api/networking/v1"
+)
+
+// translator takes care of translating the ingress objects into Envoy resources.
+type translator struct {
+	envoyListenPort *uint
```
ncdc, line 42: Why is this a pointer?

pkg/envoy-controlplane/translator.go
```diff
@@ -0,0 +1,226 @@
...
+func newTranslator(envoyListenPort *uint) *translator {
+	return &translator{
+		envoyListenPort: envoyListenPort,
+	}
+}
+
+// translateIngress has a "simple" implementation of a networkingv1.Ingress parser to translation to Envoy resources.
+// It traverses the ingress spec object and creates a list of Envoy resources.
+func (t *translator) translateIngress(ingress networkingv1.Ingress) ([]cachetypes.Resource, []*envoyroutev3.VirtualHost) {
+
```
ncdc, line 55:
```suggestion
```

pkg/envoy-controlplane/translator.go
```diff
@@ -0,0 +1,226 @@
...
+	cluster := t.newCluster(ingressToKey(ingress), 2*time.Second, endpoints, envoyclusterv3.Cluster_STRICT_DNS)
+	cluster.DnsLookupFamily = envoyclusterv3.Cluster_V4_ONLY
+
+	virtualHosts := make([]*envoyroutev3.VirtualHost, 0)
+	routes := make([]*envoyroutev3.Route, 0)
+	domains := make([]string, 0)
+
+	//TODO(jmprusi): We are ignoring the path type, we need to review this.
+	for i, rule := range ingress.Spec.Rules {
+
```
ncdc, line 78:
```suggestion
```

ncdc (changes requested):

ncdc:
@jmprusi this is a great start! Please make the controller multi-cluster aware - with the changes I've requested, it will be using cross-cluster informers & listers.

pkg/reconciler/ingress/ingress.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+func (c *Controller) updateLeafs(ctx context.Context, currentLeaves []*networkingv1.Ingress, desiredLeaves []*networkingv1.Ingress) ([]*networkingv1.Ingress, error) {
+	// Let's iterate over the current leaves and remove the ones that are not in the desired leaves, and update the ones that match.
+	for _, currentleaf := range currentLeaves {
+		found := false
+
+		// We will remove a leaf from the desiredLeaves if it's found already, so we don't create it later.
+		for i, rlen := 0, len(desiredLeaves); i < rlen; i++ {
+			j := i - (rlen - len(desiredLeaves))
```
ncdc, line 165: This is confusing. I also think that given this function is currently responsible for updating existing leaves and deleting nondesired ones, it would make sense to move the creation of "missing" leaves in here as well. If we want to split it up later, we can, but I think the current division of "update+delete, return missing" and "create missing" is not very clear.

pkg/reconciler/ingress/ingress.go
```diff
@@ -234,7 +235,8 @@ func (c *Controller) desiredLeaves(ctx context.Context, root *networkingv1.Ingre
 	}
 
 	// TODO(jmprusi): Calculate the max string length.
-	generateName := strings.Split(rootKey, "#$#")[1] + "-"
+	_, rootIngressName := clusters.SplitClusterAwareKey(rootKey)
```
ncdc, line 238: I think you need to call cache.SplitMetaNamespaceKey first, to get namespace, clusterAwareName, err. Then you can call SplitClusterAwareKey(clusterAwareName) to get cluster, name.

ncdc: Wait, you already have `root` that you can use, can't you? `root.Name`?

jmprusi: True! fixed

pkg/localenvoy/controllers/ingress/envoyingress.go
```diff
@@ -0,0 +1,105 @@
...
+	ownedByIngress   = "ingress.kcp.dev/owned-by-ingress"
+	ownedByNamespace = "ingress.kcp.dev/owned-by-namespace"
+)
+
+// reconcile is triggered on every change to an ingress resource, or it's associated services (by tracker).
+func (c *Controller) reconcile(ctx context.Context, ingress *networkingv1.Ingress) error {
+	klog.InfoS("reconciling Ingress", "ClusterName", ingress.ClusterName, "Namespace", ingress.Namespace, "Name", ingress.Name)
+
+	if ingress.Labels[clusterLabel] == "" {
+		// this is the root. Ignore.
```
ncdc, line 44: Am I interpreting the logic in here correctly?

- Only reconcile leaves
- Get the root
- Calculate the root LB hostname
- Label the leaf for Envoy (which will get updated in `process() after this function returns`)
- Update the root (to set LB hostname)

Would it make sense to split this into 1 part that reconciles a root (set hostname & update only if needed) and 1 part that reconciles a leaf?

jmprusi: yes.  There's only one small thing, i've added a new "check" to make sure that the leaf has a status set, If it doesn't, we shouldn't mark it ready for envoy, and neither set the root Ingress status field, as we don't know if there's another ingress leaf with an status endpoint.

Also, perhaps we should avoid updating the root Ingress status if it doesn't change?

ncdc (approved):

PR #483, opened 2022-02-09
Set up deploy pipeline using Github Actions

kylape:
Also includes the manifest of resources to apply every time an image is built.

### DIFF
--- a/.github/workflows/kcp-image.yaml
+++ b/.github/workflows/kcp-image.yaml
@@ -10,6 +10,7 @@ on:
 
 jobs:
   kcp-image:
+    if: github.repository_owner == 'kcp-dev'
     name: Build and Publish KCP Image
     runs-on: ubuntu-latest
 
@@ -42,3 +43,16 @@ jobs:
           registry: ghcr.io/${{ github.repository_owner }}
           username: ${{ github.actor }}
           password: ${{ github.token }}
+      - name: Set up kubectl and kustomize
+        run: |-
+          curl -sL "https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%2Fv4.5.1/kustomize_v4.5.1_linux_amd64.tar.gz" | tar xzf -
+          curl -sLO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+          chmod u+x ./kubectl
+          chmod u+x ./kustomize
+
+      - name: Deploy new image to CI
+        id: deploy-to-ci
+        run: |-
+          cd manifest
+          ../kustomize edit set image ghcr.io/kcp-dev/kcp=ghcr.io/${{ github.repository }}:${{ steps.vars.outputs.sha_short }}
+          ../kustomize build . | ../kubectl --server=https://${{ secrets.CICD_KUBE_HOST }} --token=${{ secrets.CICD_KUBE_TOKEN }} -n ${{ secrets.CICD_KUBE_NS }} apply -f -
--- a/.github/workflows/kcp-image.yaml
+++ b/.github/workflows/kcp-image.yaml
@@ -45,7 +45,7 @@ jobs:
           password: ${{ github.token }}
       - name: Set up kubectl and kustomize
         run: |-
-          curl -sL "https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%2Fv4.5.1/kustomize_v4.5.1_linux_amd64.tar.gz" | tar xzf -
+          curl -sL "https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%2Fv4.5.2/kustomize_v4.5.2_linux_amd64.tar.gz" | tar xzf -
           curl -sLO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
           chmod u+x ./kubectl
           chmod u+x ./kustomize
--- a/manifest/etcd.yaml
+++ b/manifest/etcd.yaml
@@ -4,7 +4,7 @@ kind: Certificate
 metadata:
   name: etcd
 spec:
-  secretName: etcd-secret
+  secretName: etcd-cert
   duration: 2160h # 90d
   renewBefore: 360h # 15d
   subject:
@@ -36,7 +36,7 @@ kind: Certificate
 metadata:
   name: etcd-peer
 spec:
-  secretName: etcd-peer-secret
+  secretName: etcd-peer-cert
   duration: 2160h # 90d
   renewBefore: 360h # 15d
   subject:
@@ -134,10 +134,10 @@ spec:
       volumes:
       - name: peer-certs
         secret:
-          secretName: etcd-peer-secret
+          secretName: etcd-peer-cert
       - name: server-certs
         secret:
-          secretName: etcd-secret
+          secretName: etcd-cert
   volumeClaimTemplates:
     - metadata:
         name: etcd-data
--- a/manifest/etcd.yaml
+++ b/manifest/etcd.yaml
@@ -1,0 +1,148 @@
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: etcd
+spec:
+  secretName: etcd-secret
+  duration: 2160h # 90d
+  renewBefore: 360h # 15d
+  subject:
+    organizations:
+      - redhat
+  isCA: false
+  privateKey:
+    algorithm: RSA
+    encoding: PKCS1
+    size: 2048
+  usages:
+    - server auth
+    - client auth
+  dnsNames:
+    - etcd
+    - etcd-0
+    - etcd-1
+    - etcd-2
+    - etcd-0.etcd
+    - etcd-1.etcd
+    - etcd-2.etcd
+  ipAddresses:
+    - 0.0.0.0
+  issuerRef:
+    name: kcp
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: etcd-peer
+spec:
+  secretName: etcd-peer-secret
+  duration: 2160h # 90d
+  renewBefore: 360h # 15d
+  subject:
+    organizations:
+      - redhat
+  isCA: false
+  privateKey:
+    algorithm: RSA
+    encoding: PKCS1
+    size: 2048
+  usages:
+    - server auth
+    - client auth
+  dnsNames:
+    - etcd-0.etcd
+    - etcd-1.etcd
+    - etcd-2.etcd
+    - etcd-0
+    - etcd-1
+    - etcd-2
+  ipAddresses:
+    - 0.0.0.0
+  issuerRef:
+    name: kcp
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: etcd
+spec:
+  clusterIP: None
+  ports:
+    - port: 2379
+      name: client
+    - port: 2380
+      name: peer
+  selector:
+    app: etcd
+---
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: etcd
+  labels:
+    app: etcd
+spec:
+  serviceName: etcd
+  selector:
+    matchLabels:
+      app: etcd
+  replicas: 3
+  template:
+    metadata:
+      name: etcd
+      labels:
+        app: etcd
+    spec:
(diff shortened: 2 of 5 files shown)


.github/workflows/kcp-image.yaml
```diff
@@ -42,3 +43,17 @@ jobs:
         registry: ghcr.io/${{ github.repository_owner }}
         username: ${{ github.actor }}
         password: ${{ github.token }}
+
+    - name: Set up kubectl and kustomize
+      run: |-
+        curl -sL "https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%2Fv4.5.1/kustomize_v4.5.1_linux_amd64.tar.gz" | tar xzf -
```
stevekuznetsov, line 49: No way to just pre-install these? Hate to see big downloads per-run

kylape: I looked at the kubectl github actions for a little bit, and they all either embedded the kubectl executable in a container image or they ran a container image that downloaded kubectl every time. So I figured I'd just cut out the middle man.

We could always create a custom runner that had all our binaries pre-installed.  This could also make caching image layers more straightforward.

manifest/kcp.yaml
```diff
@@ -0,0 +1,151 @@
...
+      containers:
+      - name: kcp
+        image: ghcr.io/kcp-dev/kcp:latest
+        ports:
+        - containerPort: 6443
+        command:
+        - /kcp
+        args:
+        - start
+        - --run-controllers
```
sttts, line 71: nit: this is actually defaulting to true

manifest/kcp.yaml
```diff
@@ -0,0 +1,151 @@
...
+        image: ghcr.io/kcp-dev/kcp:latest
+        ports:
+        - containerPort: 6444
+        command:
+        - /virtual-workspaces
+        args:
+        - workspaces
+        - --workspaces:kubeconfig=/tmp/kcp/admin.kubeconfig
+        - --authentication-kubeconfig=/tmp/kcp/admin.kubeconfig
+        - --authentication-skip-lookup
```
sttts, line 116: this disables delegated authentication I believe, i.e. the kubeconfig above is not used. cc @davidfestal

manifest/kcp.yaml
```diff
@@ -0,0 +1,151 @@
...
+        - --auto-publish-apis
+        - --push-mode
+        - --etcd-servers=https://etcd:2379
+        - --etcd-keyfile=/etc/etcd/tls/server/tls.key
+        - --etcd-certfile=/etc/etcd/tls/server/tls.crt
+        - --etcd-cafile=/etc/etcd/tls/server/ca.crt
+        - --tls-private-key-file=/etc/kcp/tls/server/tls.key
+        - --tls-cert-file=/etc/kcp/tls/server/tls.crt
+        - --root-directory=/tmp/kcp
+        - --oidc-issuer-url=https://sso.redhat.com/auth/realms/redhat-external
```
sttts, line 81: could we point the public instance to github SSO ?

manifest/kcp.yaml
```diff
@@ -0,0 +1,151 @@
...
+        - start
+        - --run-controllers
+        - --auto-publish-apis
+        - --push-mode
+        - --etcd-servers=https://etcd:2379
+        - --etcd-keyfile=/etc/etcd/tls/server/tls.key
+        - --etcd-certfile=/etc/etcd/tls/server/tls.crt
+        - --etcd-cafile=/etc/etcd/tls/server/ca.crt
+        - --tls-private-key-file=/etc/kcp/tls/server/tls.key
+        - --tls-cert-file=/etc/kcp/tls/server/tls.crt
```
sttts, line 79: while technically possible, can we split the two serving certs of kcp and virtual-workspaces and use a common CA ? (can be follow-up). Using the same for both looks like a non-production shortcut.

sttts: Also this means that we have to put the ca.crt, not the tls.crt into users' and admin's kubeconfig.

.github/workflows/kcp-image.yaml
sttts: for the record: kubectl is 46mb.

manifest/kcp.yaml
kylape: The virtual workspace service seems to be accepting and denying my requests as I would expect.

manifest/kcp.yaml
kylape: Sure, I can work on setting that up.  Why the preference for github SSO?

manifest/kcp.yaml
davidfestal: No, authentication-skip-lookup just disables the lookup of the authentication config map with auth settings in the delegated server. This one is intended, but I think I made it true by default now (just check with the --help arg on the cli)

manifest/kcp.yaml
kylape: I've updated the PR to have the virtual workspaces container use a different cert than kcp.

manifest/kcp.yaml
```diff
@@ -0,0 +1,151 @@
...
+        - --workspaces:kubeconfig=/tmp/kcp/admin.kubeconfig
+        - --authentication-kubeconfig=/tmp/kcp/admin.kubeconfig
+        - --authentication-skip-lookup
+        - --tls-private-key-file=/etc/kcp/tls/server/tls.key
+        - --tls-cert-file=/etc/kcp/tls/server/tls.crt
+        - --secure-port=6444
+        livenessProbe:
+          failureThreshold: 3
+          httpGet:
+            path: /services/applications/personal/livez
```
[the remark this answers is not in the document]
kylape: This was my experience with virtual workspace healthz endpoints:

```
$ curl -k https://localhost:6444/services/applications/healthz
404 page not found

$ curl -k https://localhost:6444/services/applications/personal/healthz
ok%
```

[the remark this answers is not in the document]
csams: Local testing shows `https://localhost:6444/livez` and `https://localhost:6444:readyz` also work. Switching to those.

manifest/kcp.yaml
sttts: Any public one is fine. Not everybody in the community has RH SSO access :)

manifest/etcd.yaml
```diff
@@ -0,0 +1,148 @@
...
+              mountPath: /etc/etcd/tls/server
+          command:
+            - /bin/sh
+            - -c
+            - |
+              PEERS="etcd-0=https://etcd-0.etcd:2380,etcd-1=https://etcd-1.etcd:2380,etcd-2=https://etcd-2.etcd:2380"
+              exec etcd --name ${HOSTNAME} \
+                --listen-peer-urls https://0.0.0.0:2380 \
+                --initial-advertise-peer-urls https://${HOSTNAME}:2380 \
+                --listen-client-urls https://0.0.0.0:2379 \
```
s-urbaniak, line 120: nit: can we restrict this somehow? `0.0.0.0` seems very open

manifest/kcp.yaml
```diff
@@ -0,0 +1,184 @@
...
+        - --etcd-servers=https://etcd:2379
+        - --etcd-keyfile=/etc/etcd/tls/server/tls.key
+        - --etcd-certfile=/etc/etcd/tls/server/tls.crt
+        - --etcd-cafile=/etc/etcd/tls/server/ca.crt
+        - --tls-private-key-file=/etc/kcp/tls/server/tls.key
+        - --tls-cert-file=/etc/kcp/tls/server/tls.crt
+        - --root-directory=/tmp/kcp
+        - --oidc-issuer-url=https://sso.redhat.com/auth/realms/redhat-external
+        - --oidc-client-id=rhoas-cli-prod
+        - --oidc-groups-claim=org_id
```
s-urbaniak, line 110: cc @slaskawi for verification

slaskawi: LGTM

manifest/kcp.yaml
slaskawi: We could actually do even better than that. Point to RHSSO and turn Github Social Login Provider in that specific Realm in Keycloak. This way, you could either login using RHSSO or just click "login with Github".

kylape: That would be great, but that assumes we can modify the realm, right?

.github/workflows/kcp-image.yaml
gorkem: RH does maintain an [action](https://github.com/marketplace/actions/openshift-tools-installer ) that does proper caching it supports several tools would that help?

manifest/kcp.yaml
slaskawi: Yes, that's right. I'm not sure if this is possible in our scenario but probably worth checking.

akoserwal: I doubt that will be allowed now in sso.rh (may be in future). Enabling Github IDP in sso.rh means all rh site supports the github login which isn't the case now.

.github/workflows/kcp-image.yaml
kylape: How about we create a follow up issue to improve on this later?

manifest/kcp.yaml
kylape: I believe we've landed on RH SSO for now and we can follow up later with a separate instance that uses Github IDP.

manifest/etcd.yaml
kylape: We tested setting all the hostnames to `${HOSTNAME}`, and it caused some binding conflicts.  Binding to `127.0.0.1` didn't expose the port to KCP.  FWIW the etcd running in our pcluster uses 0.0.0.0.

kylape:
@sttts @ncdc I have addressed the comments, and I believe this should be ready to merge.

.github/workflows/kcp-image.yaml
```diff
@@ -42,3 +43,16 @@ jobs:
           registry: ghcr.io/${{ github.repository_owner }}
           username: ${{ github.actor }}
           password: ${{ github.token }}
+      - name: Set up kubectl and kustomize
+        run: |-
+          curl -sL "https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%2Fv4.5.1/kustomize_v4.5.1_linux_amd64.tar.gz" | tar xzf -
```
ncdc, line 48: 4.5.2 is out if you want the latest. Ok to defer too.

manifest/etcd.yaml
```diff
@@ -0,0 +1,148 @@
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: etcd
+spec:
+  secretName: etcd-secret
```
ncdc, line 7: nit (global comment): try to avoid repeating the type of the resource in its name - e.g. use something like `etcd` or `etcd-certificate` instead of `etcd-secret`.

manifest/etcd.yaml
```diff
@@ -0,0 +1,148 @@
...
+kind: Service
+metadata:
+  name: etcd
+spec:
+  clusterIP: None
+  ports:
+    - port: 2379
+      name: client
+    - port: 2380
+      name: peer
```
ncdc, line 75: Do we need to specify the protocol for the ports, like you did for the kcp service?

csams: It's okay since the default protocol for services is TCP.

ncdc (approved):

manifest/kcp.yaml
```diff
@@ -0,0 +1,186 @@
...
+          mountPath: /tmp/kcp
+      - name: virtual-workspaces
+        image: ghcr.io/kcp-dev/kcp:latest
+        ports:
+        - containerPort: 6444
+        command:
+        - sh
+        - -c
+        - >
+          cat /tmp/kcp/admin.kubeconfig | sed -e 's|https://\([0-9]\{1,3\}\.\)\{3\}[0-9]\{1,3\}|https://localhost|' > /tmp/kcp/localhost.kubeconfig &&
```
sttts, line 145: follow-up: use advertised hostname in https://github.com/kcp-dev/kcp/blob/main/pkg/server/options/authentication.go#L138. @kylape @csams something one of you could do?

sttts: Oh, you do the opposite. Nvmd.

.github/workflows/kcp-image.yaml
kylape: https://github.com/kcp-dev/kcp/issues/621

PR #484, opened 2022-02-09
syncer: fix hotloop & minor cleanup

### LINKED ISSUE #482: Syncer has a hot-loop
It doesn't return after calling AddRateLimited, so it immediately calls Forget on the key, which resets the rate limiter back to 0.

ncdc:
Fix a bug in the syncer where, after it got an error, it failed to
return immediately after adding the key back to the queue with a rate
limit. Because it didn't return immediately, the key was subsequently
forgotten (i.e. rate limit reset), which broke rate limited backoff.

Refactor the holder struct not to store an object instance, which is
consistent with standard controller/queue best practices.

Update the process method code to (hopefully) correctly determine the
"from" and "to" namespaces and item keys.

Fixes #482

### DIFF
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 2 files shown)


pkg/syncer/syncer.go
```diff
@@ -297,95 +321,106 @@ func PhysicalClusterNamespaceName(l NamespaceLocator) (string, error) {
...
 			return nil
 		}
 	} else {
 		nsInformer := c.fromInformers.ForResource(schema.GroupVersionResource{Version: "v1", Resource: "namespaces"})
-		nsObj, err := nsInformer.Lister().Get(clusters.ToClusterAwareKey(meta.GetClusterName(), namespace))
+
+		nsKey := fromNamespace
+		if h.clusterName != "" {
+			// If our "physical" cluster is a kcp instance (e.g. for testing purposes), it will return resources
+			// with metadata.clusterName set, whicih means their keys are cluster-aware, so we need to do the same here.
```
stevekuznetsov, line 364:
```suggestion
			// with metadata.clusterName set, which means their keys are cluster-aware, so we need to do the same here.
```

imjasonh (approved):

pkg/syncer/syncer.go
ncdc: Thanks for catching, fixed!

stevekuznetsov (approved):

PR #486, opened 2022-02-10
Share KCP certs with the virtual workspaces APIServer

davidfestal:
This PR makes it so that the KCP server and virtual workspaces APIServer can work hand in hand with the same certs, and so that after restart either one or the other, things continue to work.

It also adapts the `workspace` `kubectl` plugin to match the new behavior and make it work by default with the KCP `admin.kubeconfig` when KCP and the virtual workspaces APIserver are started with the default options.

Individual commits comments should be self-explanatory.


pkg/server/kubeconfig.go
```diff
@@ -0,0 +1,162 @@
...
+	kubeConfig.Contexts = map[string]*clientcmdapi.Context{
+		"cross-cluster": {Cluster: "cross-cluster", AuthInfo: adminUserName},
+		"admin":         {Cluster: "admin", AuthInfo: adminUserName},
+		"user":          {Cluster: "user", AuthInfo: adminUserName},
+	}
+	kubeConfig.CurrentContext = "admin"
+	return &kubeConfig
+}
+
+func createKubeConfigs(server *server.GenericAPIServer, externalKubeConfigAdminToken string, externalKubeConfigAdminTokenHash []byte, externalKubeConfigPath string) (loopbackBasedKubeConfig *clientcmdapi.Config, externalKubeConfig *clientcmdapi.Config, err error) {
```
sttts, line 119: please split into one for loopback, one for admin

pkg/server/server.go
```diff
@@ -288,43 +293,17 @@ func (s *Server) Run(ctx context.Context) error {
...
-			CertificateAuthorityData: server.LoopbackClientConfig.CAData,
-			TLSServerName:            server.LoopbackClientConfig.TLSClientConfig.ServerName,
-		},
-	}
-	clientConfig.Contexts = map[string]*clientcmdapi.Context{
-		"cross-cluster": {Cluster: "cross-cluster", AuthInfo: "loopback"},
-		"admin":         {Cluster: "admin", AuthInfo: "loopback"},
-		"user":          {Cluster: "user", AuthInfo: "loopback"},
+	externalKubeConfigPath := filepath.Join(s.options.Extra.RootDirectory, s.options.Extra.KubeConfigPath)
+	loopbackClientConfig, externalClientConfig, err := createKubeConfigs(server, externalKubeConfigAdminToken, externalKubeConfigAdminTokenHash, externalKubeConfigPath)
```
sttts, line 297: loopbackKubeConfig, externalKubeConfig

pkg/server/server.go
```diff
@@ -288,43 +293,17 @@ func (s *Server) Run(ctx context.Context) error {
...
-			Server:                   server.LoopbackClientConfig.Host + "/clusters/user",
-			CertificateAuthorityData: server.LoopbackClientConfig.CAData,
-			TLSServerName:            server.LoopbackClientConfig.TLSClientConfig.ServerName,
-		},
-	}
-	clientConfig.Contexts = map[string]*clientcmdapi.Context{
-		"cross-cluster": {Cluster: "cross-cluster", AuthInfo: "loopback"},
-		"admin":         {Cluster: "admin", AuthInfo: "loopback"},
-		"user":          {Cluster: "user", AuthInfo: "loopback"},
+	externalKubeConfigPath := filepath.Join(s.options.Extra.RootDirectory, s.options.Extra.KubeConfigPath)
```
sttts, line 296: can s.options.Extra.KubeConfigPath be absolute?

pkg/virtual/framework/cmd/cmd.go
```diff
@@ -67,11 +70,15 @@ type SubCommandOptions interface {
...
-		SecureServing:     kubeoptions.NewSecureServingOptions().WithLoopback(),
+		SecureServing:     kubeoptions.NewSecureServingOptions(),
 		Authentication:    genericapiserveroptions.NewDelegatingAuthenticationOptions(),
 		SubCommandOptions: subCommandOptions,
 	}
 
+	options.SecureServing.ServerCert.CertKey.CertFile = filepath.Join(".", ".kcp", "apiserver.crt")
+	options.SecureServing.ServerCert.CertKey.KeyFile = filepath.Join(".", ".kcp", "apiserver.key")
+	options.SecureServing.BindPort = SecurePortDefault
+	options.Authentication.SkipInClusterLookup = true
```
sttts, line 81: we might want to start an options package here too, and move those default into options.NewSecureServing

pkg/virtual/framework/rootapiserver/root_apiserver.go
```diff
@@ -210,12 +212,36 @@ func (c completedConfig) NewRequestInfo(req *http.Request) (*genericapirequest.R
...
+			loopbackConfigCert, _ = sniCert.CurrentCertKeyContent()
+		}
+	}
+
+	if loopbackConfigCert == nil {
+		return nil, errors.New("No certs setup")
+	}
+
+	genericConfig.Config.LoopbackClientConfig, err = genericConfig.Config.SecureServing.NewLoopbackClientConfig(uuid.New().String(), loopbackConfigCert)
+	if err != nil {
```
sttts, line 244: are you sure this loopback client does not exist already? (I don't remember either, but could well be).

pkg/server/server.go
davidfestal: I believe that has been the assumption until now. I just extract what was already there: https://github.com/kcp-dev/kcp/pull/486/files/ed80c30f9205d422ff093fcad3dfaf84df8f0a2a#diff-104a11712684e520f0affa95e6a053faeb713306a10f69ee99b9729cc9dd1996L337

pkg/virtual/framework/cmd/cmd.go
davidfestal: Could we leave that for a next refactoring PR ?

davidfestal: Can we leave that for a broader refactoring task in a next PR ?

pkg/virtual/framework/rootapiserver/root_apiserver.go
davidfestal: I don't think it exists since we didn't use `SecureServingOptionsWithLoopback`.
In fact the loopback we create here is just because one is required, but the RootAPIServer wouldn't really need one since it doesn't manages REST objects on it own, and the virtual workspaces it hosts will generally have their own config that will point to an external delegate APIServer (typically KCP, or a proxy, or another virtual workspace).

pkg/server/kubeconfig.go
davidfestal: Fix in commit 7c3486cb3506a17f776e0a8fc6bcf5376c42d947

pkg/server/server.go
davidfestal: Fix in commit 7c3486cb3506a17f776e0a8fc6bcf5376c42d947

pkg/virtual/framework/cmd/cmd.go
davidfestal: Done after the rebase on your PR and the additional commits you provided

sttts: sure, was meant as follow-up.

pkg/virtual/framework/rootapiserver/root_apiserver.go
sttts: not sure I understand this change then. Why do we need the new code?

davidfestal: Because we should no run `SecureServingOptionsWithLoopback.ApplyTo`, because it creates and additional Lookpack config with its own SNIcerts, etc ... which we don't want.

davidfestal:
@sttts All you comments fixed.

pkg/server/options/flags.go
```diff
@@ -80,6 +80,9 @@ var (
 		// KCP Authorization flags
 		"authorization-always-allow-paths", // A list of HTTP paths to skip during authorization, i.e. these are authorized without contacting the 'core' kubernetes server.
 
+		// KCP Admin Authentication flags
+		"authentication-admin-token-store", // Path to which the administrative token hash should be written at startup. If this is relative, it is relative to --root-directory.
```
sttts, line 84: nit: `authentication-admin-token-path` would be little more in-line with other flags

sttts (approved):

pkg/server/options/flags.go
davidfestal: Fixed in commit f10ab04

PR #493, opened 2022-02-10
Fix syncer hotloop take 2

ncdc:


pkg/syncer/syncer.go
```diff
@@ -275,8 +275,7 @@ func (c *Controller) Start(ctx context.Context, numThreads int) {
 
 // startWorker processes work items until stopCh is closed.
 func (c *Controller) startWorker(ctx context.Context) {
-	for {
-		c.processNextWorkItem(ctx)
+	for c.processNextWorkItem(ctx) {
```
sttts, line 278: wow, nice find.

PR #499, opened 2022-02-11
Add simple cluster health check

ncdc:
- Add simple cluster health check
- Don't schedule to unready clusters
- Reevaluate unscheduled namespaces for ready clusters


imjasonh (approved):
Very nice! 🎉

PR #502, opened 2022-02-14
WIP e2e: Optionally run the cluster controller in-process

marun:
Passing `-in-process-controllers=true` as a test argument will enable running the cluster controller in the same process as the test to simplify development (change/build/test becomes change/test) and debugging ('start controller under debug and run test' becomes 'run test under debugger').

### DIFF
--- a/test/e2e/framework/fixture.go
+++ b/test/e2e/framework/fixture.go
@@ -18,12 +18,14 @@ package framework
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"strconv"
 	"sync"
 	"testing"
 	"time"
 
+	"github.com/spf13/pflag"
 	"github.com/stretchr/testify/require"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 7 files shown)


ncdc:
I think we already have `INPROCESS` to run kcp in-process? cc @sttts

marun:
> I think we already have `INPROCESS` to run kcp in-process? cc @sttts

Aha, I see that now. That doesn't obviate the need for this PR, since one of my goals is reducing iteration time and having to start the entire server in-process doesn't achieve that. My end-goal is running a persistent server and only running a specific controller in-process. I'll update this PR to enable a test arg for starting the server in-process.

marun:
Rebased

marun:
Rebased

marun:
Rebased

marun:
Rebased.

@sttts I had previously added support for configuring INPROCESS via flags, but I noticed you've enabled as a fixture field. Should I assume that you prefer to enable per-fixture vs globally via flags?

test/e2e/framework/fixture.go
```diff
@@ -185,3 +202,51 @@ func NewWorkspaceFixture(t *testing.T, server RunningServer, orgClusterName stri
...
+		return append(f.Args, f.InProcessServerArgs...)
+	}
+	return f.Args
+}
+
+func (f *ClusterControllerFixture) Run(t *testing.T, server RunningServer) {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	t.Cleanup(cancelFunc)
+
+	f.Args = append(f.Args, fmt.Sprintf("--kubeconfig=%s", server.KubeconfigPath()))
```
marun, line 241: @sttts From what I can tell, this kubeconfig is no longer sufficient to configure the cluster controller. The cluster controller test fails trying to install the cluster due to the cluster never becoming ready. When I debugged the in-process sync manager controller, it wasn't receiving events at all. 

Is there something about the configuration here that differs from what is supplied to the cluster controller installed by the kcp server?

marun:
Currently broken by bad configuration, and it's likely that a new controller manager that can run all the controllers in-process will obviate the need for this PR.

PR #507, opened 2022-02-15
Split logical Workspace resource from physical ClusterWorkspace resource

sttts:
Our physical ClusterWorkspace resource contains data about infrastructure like shard scheduling information, like shard movement, other status data possibly in the future. We are not able to keep this stable as our architecture is still in flux.

The Workspace type will be minimal and supportable:
```yaml
apiVersion: tenancy.kcp.dev/v1beta1
kind: Workspace
metadata:
  name: myapp
spec: {}
status:
  phase: Ready
  url: https://..../clusters/redhat_myapp
```

Also, this change allows us to introduce other backing types for workspace, and then repesent them as a unified type.

If we want to add more spec or status for specific backing types, we will use discriminated union fields (similar to storage in Kube).


ncdc:
I think this is LGTM modulo rebase & getting things green

sttts:
Merged through #508.

PR #517, opened 2022-02-15
e2e: Refactor Artifact to ensure tests fail when collection fails

marun:
This PR refactors `Artifact()` to fail rather than log if any error is encountered during artifact collection. A fix is included for a latent bug in one of the workspace e2e tests surfaced by the refactor.


marun:
Rebased and ready for review.

ncdc (approved):

Issue #518, opened 2022-02-15
Pods not synced back from physical cluster to kcp


lburgazzoli:
**Describe the bug**
Pods are not synced from the physical cluster back to kcp

**To Reproduce**
Started kcp from main with
```
kcp start \
    --auto-publish-apis=true \
    --pull-mode=false \
    --push-mode=true \
    --run-controllers=true \
    --resources-to-sync=deployments.apps,pods
```

kcp starts and reports:

```
syncer.go:141] "Set up informer" direction=kcpToPhysicalCluster clusterName="connectors" logicalCluster="admin" gvr="/v1, Resource=namespaces"
syncer.go:141] "Set up informer" direction=kcpToPhysicalCluster clusterName="connectors" logicalCluster="admin" gvr="/v1, Resource=pods"
syncer.go:141] "Set up informer" direction=kcpToPhysicalCluster clusterName="connectors" logicalCluster="admin" gvr="apps/v1, Resource=deployments"
syncer.go:141] "Set up informer" direction=physicalClusterToKcp clusterName="connectors" logicalCluster="admin" gvr="/v1, Resource=namespaces"
syncer.go:141] "Set up informer" direction=physicalClusterToKcp clusterName="connectors" logicalCluster="admin" gvr="/v1, Resource=pods"
syncer.go:141] "Set up informer" direction=physicalClusterToKcp clusterName="connectors" logicalCluster="admin" gvr="apps/v1, Resource=deployments"
```

Then register a physical cluster named `my-cluster` using kind
Then create a deployment

```
kubectl apply -f contrib/examples/deployment.yaml
kubectl label --overwrite deployment example kcp.dev/cluster=my-cluster
```

At this point, the deployment gets synchronized with the physical cluster to a namespace created by kcp and the related pods starts:

```
➜ kubectl get deployments.apps,pods -n kcpb0a2aacd123de54bf53be16aa7fdf1a9262aec753e1a27717df04e7b
NAME                      READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/example   1/1     1            1           15m

NAME                           READY   STATUS    RESTARTS   AGE
pod/example-5cbdcd89b7-5dwn5   1/1     Running   0          15m
```

However when looking at kcp, I see that the deployment is properly synced and reported as running but I don't see any pod being synced back:

```
➜ kubectl get deployments.apps,pods -n my-ns
NAME                      READY   UP-TO-DATE   AVAILABLE   AGE
[...]

PR #520, opened 2022-02-16
syncer: Fallback to creating object when updating status from pcluster

astefanutti:
This is a tentative fix for #518.

In the example of the[ Camel K operator](https://github.com/apache/camel-k), the operator reconciles Integration resources into Deployment resources, but also watches the Pods owned by these Deployments, to update Integration statuses, because the Deployment controller do not aggregate all the information that's required to update the Integration statuses.

Currently KCP assumes the resources exist already in the logical clusters, when syncing the statuses from the physical clusters.

In some setups, we would want a single Camel K operator instance, connected to the KCP control plane, and shard the Integration deployments into physical clusters, but still have that operator instance be able to watches for the Pods.

Generally, that would enable controllers and tools to use KCP as a multi-cluster control plane transparently.

### DIFF
--- a/pkg/syncer/statussyncer.go
+++ b/pkg/syncer/statussyncer.go
@@ -20,6 +20,7 @@ import (
 	"context"
 
 	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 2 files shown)


sttts:
IMO this needs more thought. I am very hesitant to add this in yet another syncer one-off fix. We need to think much more about an API to specify this behaviour. Something like a `ClusterSyncSet`.

astefanutti:
@sttts yes, I totally agree. That PR was more meant to share the use cases and raise awareness (and also unblock us in the meantime 😃).

lburgazzoli:
I guess we may need also to reason about what pods you want to sync back 
Maybe we may want to sync things based on the owner reference tree ?

PR #491, opened 2022-02-10
Flesh out more of the prototype 2 demo

ncdc:

PR #522, opened 2022-02-16
Prototype2

fgiloux:
Trying to run the prototype 2 demo as a learning exercise I have faced a few issues. This pull request aims at fixing them.


ncdc:
Hi, please take a look at #491 as that addresses some of the same problems and almost finishes the demo script. If you see anything I've missed, would you want to make a PR to my branch?

fgiloux:
Hi @ncdc thanks for the pointer. I gave it a try and it worked perfectly!

Some nitpicking:
- I had called contrib/demo/prototype2-script/startKcp.sh directly and when I hit ctrl-C it did not only killed the KCP processes but all my user processes and I lost my session. I understand that it was not supposed to be called directly but it would be safer to only kill the processes created (I saw that the PIDs were gathered) rather than killing everything children.
- In a similar fashion I feel uneasy when bash scripts do "rm -rf" in anything that is not in the system temp dir.

I will close this PR as you have already addressed the issues in yours. Thanks again!

PR #524, opened 2022-02-16
Add Unschedulable and Eviction fields to ClusterSpec

imjasonh:
When a cluster is Unschedulable, new namespaces won't be assigned to it.
When a cluster is Evicting, existing namespaces will also be reassigned.

Signed-off-by: Jason Hall <jasonhall@redhat.com>

Fixes https://github.com/kcp-dev/kcp/issues/556


pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -56,6 +56,16 @@ var _ conditions.Setter = &Cluster{}
...
+
+	// Unschedulable controls cluster schedulability of new workloads. By
+	// default, cluster is schedulable.
+	// +optional
+	Unschedulable bool `json:"unschedulable"`
+
+	// Evict controls cluster schedulability of new and existing workloads.
+	// If evict is true, unschedulable is ignored and assumed to be true.
+	// By default, cluster is not evicted.
+	Evict bool `json:"evict"`
```
ncdc, line 68: What if we made this EvictAfter so you could set a datetime for it?

pkg/reconciler/namespace/namespace.go
```diff
@@ -326,7 +339,10 @@ func clusterLabelPatchBytes(val string) []byte {
 func (c *Controller) observeCluster(ctx context.Context, cl *clusterv1alpha1.Cluster) error {
 	klog.Infof("Observing Cluster %s", cl.Name)
 
-	if conditions.IsTrue(cl, clusterv1alpha1.ClusterReadyCondition) {
+	// If cluster is Ready and not unschedulable or evicting, revisit any
+	// unscheduled namespaces.
+	if conditions.IsTrue(cl, clusterv1alpha1.ClusterReadyCondition) &&
+		!cl.Spec.Unschedulable && !cl.Spec.Evict {
```
ncdc, line 345: Mind making this 1 line? I think it would be easier to read.

pkg/apis/cluster/v1alpha1/types.go
imjasonh: Sure. That starts to diverge even more from how node eviction works, but we can still change how we want this all to work later.

Seeing an `EvictAfter` in the future on a cluster should enqueue a delayed workqueue item in the future with `AddAfter(duration)`, and `EvictAfter` in the present/past should evict workloads?

imjasonh: This is done now. As written currently, this might lead to a bad experience if you set `evictAfter: in ten minutes` without also setting `unschedulable: true` -- a workload might get scheduled to the cluster only to get evicted in ten minutes.

It's not a huge deal -- just cordon it before you schedule it to be drained -- but I wanted to call it out while I still remembered this.

ncdc: If it's easier to test the UX of a bool, happy to start there. @sttts WDYT?

sttts: You can set it to "time.Now()` to get the boolean behaviour, no? Time looks fine.

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -56,6 +56,17 @@ var _ conditions.Setter = &Cluster{}
 // ClusterSpec holds the desired state of the Cluster (from the client).
 type ClusterSpec struct {
 	KubeConfig string `json:"kubeconfig"`
+
+	// Unschedulable controls cluster schedulability of new workloads. By
+	// default, cluster is schedulable.
+	// +optional
+	Unschedulable bool `json:"unschedulable"`
```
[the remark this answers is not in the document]
sttts: something is wrong with kubebuilder. It creates a string "False". Maybe the syntax is lower-case false?

pkg/reconciler/namespace/namespace.go
```diff
@@ -326,14 +342,23 @@ func clusterLabelPatchBytes(val string) []byte {
...
+	if conditions.IsTrue(cl, clusterv1alpha1.ClusterReadyCondition) && !cl.Spec.Unschedulable && evicted {
 		// Revisit any unscheduled namespaces
 		var errs []error
 		errs = append(errs, c.enqueueNamespaces(ctx, unscheduledSelector))
 		errs = append(errs, c.enqueueNamespaces(ctx, scheduledToNothingSelector))
 		return errors.NewAggregate(errs)
 	}
 
+	if cl.Spec.EvictAfter != nil && cl.Spec.EvictAfter.After(now()) {
+		dur := now().Sub(cl.Spec.EvictAfter.Time)
```
sttts, line 358: if now is before eviction time, now-evictAfter is negative. Is that intentional?

pkg/reconciler/namespace/namespace.go
```diff
@@ -201,7 +212,7 @@ func (c *Controller) assignCluster(ctx context.Context, ns *corev1.Namespace) er
 }
 
 // isValidCluster checks whether the given cluster name exists and is valid for the
-// purposes of scheduling a namespace to it.
+// purposes of a namespace that is already scheduled to it.
```
ncdc, line 215: This could use some rewording (grammar) 😄

pkg/reconciler/namespace/namespace.go
```diff
@@ -211,17 +222,22 @@ func (c *Controller) isValidCluster(ctx context.Context, lclusterName, clusterNa
...
 }
 
 // ensureScheduled attempts to ensure the namespace is assigned to a viable cluster. This
-// will succeed without error if a cluster is assigned or if there are no viable clusters
-// to assign to. The condition of not being scheduled to a cluster will be reflected in
-// the namespace's status rather than by returning an error.
+// will succeed without error if a cluster is assigned to a valid clusteror if
+// there are no viable clusters to assign to. The condition of not being
+// scheduled to a cluster will be reflected in the namespace's status rather
+// than by returning an error.
```
ncdc, line 240: Undo this change?

pkg/reconciler/namespace/namespace.go
```diff
@@ -326,14 +342,23 @@ func clusterLabelPatchBytes(val string) []byte {
 func (c *Controller) observeCluster(ctx context.Context, cl *clusterv1alpha1.Cluster) error {
 	klog.Infof("Observing Cluster %s", cl.Name)
 
-	if conditions.IsTrue(cl, clusterv1alpha1.ClusterReadyCondition) {
+	evicted := cl.Spec.EvictAfter != nil && cl.Spec.EvictAfter.Time.Before(now())
+
+	// If cluster is Ready and not unschedulable or evicting, revisit any
+	// unscheduled namespaces.
+	if conditions.IsTrue(cl, clusterv1alpha1.ClusterReadyCondition) && !cl.Spec.Unschedulable && evicted {
```
ncdc, line 349: I think you mean `!evicted`?

ncdc (changes requested):

pkg/reconciler/namespace/namespace.go
```diff
@@ -172,6 +175,14 @@ func (c *Controller) assignCluster(ctx context.Context, ns *corev1.Namespace) er
 			klog.V(2).InfoS("assignCluster: excluding cluster with different metadata.clusterName", "ns.clusterName", ns.ClusterName, "check", allClusters[i].ClusterName)
 			continue
 		}
+		if allClusters[i].Spec.Unschedulable {
+			klog.V(2).InfoS("assignCluster: excluding unschedulable cluster", "metadata.name", allClusters[i].Name)
+			continue
+		}
+		if evictAfter := allClusters[i].Spec.EvictAfter; evictAfter != nil && evictAfter.Time.Before(now()) {
+			klog.V(2).InfoS("assignCluster: excluding evicted cluster", "metadata.name", allClusters[i].Name)
```
marun, line 183: nit: Is there a better term for 'cluster whose contents are marked for eviction' than 'evicted cluster'?

pkg/reconciler/namespace/namespace.go
```diff
@@ -211,17 +222,22 @@ func (c *Controller) isValidCluster(ctx context.Context, lclusterName, clusterNa
 	if err != nil {
 		return false, err
 	}
-	isReady := conditions.IsTrue(cluster, clusterv1alpha1.ClusterReadyCondition)
-	if !isReady {
+	if !conditions.IsTrue(cluster, clusterv1alpha1.ClusterReadyCondition) {
```
marun, line 225: (No action required) Consider retaining the use of `isReady` even if it is not needed for return. 'Named conditionals' tend to improve code readability.

test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -157,20 +157,51 @@ func TestNamespaceScheduler(t *testing.T) {
...
 
 			err = clusterClient.Delete(ctx, cluster1.Name, metav1.DeleteOptions{})
 			require.NoError(t, err, "failed to delete cluster1 %q", cluster1.Name)
 
+			err = expect(namespace, unschedulableMatcher())
+			require.NoError(t, err, "did not see namespace marked unschedulable")
+
+			t.Log("Uncordon the new cluster and expect the namespace to end up scheduled to it.")
+
+			cluster2.Spec.Unschedulable = false // uncordoned.
```
marun, line 176: (No action required) Would `Cordoned` be a better name for the field than `Unschedulable`?

test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -157,20 +157,51 @@ func TestNamespaceScheduler(t *testing.T) {
...
+			cluster2, err = clusterClient.Update(ctx, cluster2, metav1.UpdateOptions{})
+			require.NoError(t, err, "failed to uncordon cluster2")
+
+			err = expect(namespace, scheduledMatcher(cluster2.Name))
+			require.NoError(t, err, "did not see namespace marked scheduled for cluster2 %q", cluster2.Name)
+
+			t.Log("Mark the new cluster for immediate eviction, and see the namespace get unschedulable.")
+
+			now := metav1.NewTime(time.Now().Add(-time.Hour))
```
marun, line 185: nit: s/now/previousNow/ ?

test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -157,20 +157,51 @@ func TestNamespaceScheduler(t *testing.T) {
...
+			cluster2, err = clusterClient.Get(ctx, cluster2.Name, metav1.GetOptions{})
+			require.NoError(t, err, "failed to get cluster2")
+			cluster2.Spec.EvictAfter = &now
+			cluster2, err = clusterClient.Update(ctx, cluster2, metav1.UpdateOptions{})
+			require.NoError(t, err, "failed to evict cluster2")
+
+			err = expect(namespace, scheduledMatcher(cluster2.Name))
```
marun, line 192: Should the namespace remain scheduled to cluster2?

pkg/reconciler/namespace/namespace.go
```diff
@@ -157,6 +158,8 @@ func (c *Controller) updateSchedulableCondition(ctx context.Context, ns *corev1.
 	})
 }
 
+var now = time.Now
```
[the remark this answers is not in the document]
imjasonh: I think that was just force of habit from using `Now` at all. Since we seem to be fine with e2e test coverage I'll just remove this.

pkg/reconciler/namespace/namespace.go
imjasonh: Maybe. In the meantime, using more words to add clarity.

test/e2e/reconciler/namespace/controller_test.go
imjasonh: Maybe. I wanted to follow the precedent for Nodes, but we've already broken with that tradition by adding `EvictAfter` (or API-level eviction at all; node eviction/"drain" seems to be entirely client-side)

And in any case, even Nodes aren't consistent -- the API is `unschedulable bool`, but the CLI is `kubectl cordon <node>`

test/e2e/reconciler/namespace/controller_test.go
imjasonh: Good catch. The matcher was passing presumably because of delay in unscheduling it. Fixed.

test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -158,20 +158,51 @@ func TestNamespaceScheduler(t *testing.T) {
...
+
+			anHourAgo := metav1.NewTime(time.Now().Add(-time.Hour))
+			cluster2, err = clusterClient.Get(ctx, cluster2.Name, metav1.GetOptions{})
+			require.NoError(t, err, "failed to get cluster2")
+			cluster2.Spec.EvictAfter = &anHourAgo
+			cluster2, err = clusterClient.Update(ctx, cluster2, metav1.UpdateOptions{})
+			require.NoError(t, err, "failed to evict cluster2")
+
+			err = expect(namespace, unschedulableMatcher())
+			require.NoError(t, err, "did not see namespace marked scheduled")
```
marun, line 194: nit: s/scheduled/unscheduled/

marun: er, unschedulable?

test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -158,20 +158,51 @@ func TestNamespaceScheduler(t *testing.T) {
...
+
+			t.Log("Unevict the new cluster and see the namespace scheduled to it.")
+
+			cluster2.Spec.EvictAfter = nil // Not evicting.
+			cluster2, err = clusterClient.Update(ctx, cluster2, metav1.UpdateOptions{})
+			require.NoError(t, err, "failed to un-evict cluster2")
```
marun, line 200: nit: s/un-evict/unevict/ (the should the log line on 195 change?)

pkg/reconciler/namespace/namespace.go
```diff
@@ -211,11 +225,15 @@ func (c *Controller) isValidCluster(ctx context.Context, lclusterName, clusterNa
...
 	}
-	isReady := conditions.IsTrue(cluster, clusterv1alpha1.ClusterReadyCondition)
-	if !isReady {
+	if ready := !conditions.IsTrue(cluster, clusterv1alpha1.ClusterReadyCondition); !ready {
 		klog.Infof("Cluster %q|%q is not reporting ready", lclusterName, clusterName)
+		return false, nil
 	}
-	return isReady, nil
+	if evictAfter := cluster.Spec.EvictAfter; evictAfter != nil && evictAfter.Time.Before(time.Now()) {
+		klog.Infof("Cluster %q|%q is evicted", lclusterName, clusterName)
```
marun, line 233: nit: s/evicted/cordoned/ ? i.e. namespaces scheduled to a namespace are evicted. clusters are not evicted. 

Maybe this suggests using `cordonAfter`? There is ambiguity to me in what `cluster.evictAfter` means, since the natural implication for me is that the cluster should be evicted. `cluster.cordonAfter` has no such ambiguity.

imjasonh: Good point, about clusters not being evicted, only namespaces/workloads.

However, `evictAfter` != `cordonAfter` -- cordoning affects only new workloads making a scheduling decision, evicting affects already-scheduled workloads, kicking them out. Maybe `drainAfter`?

(Now I'm thinking maybe it's worth expressing `unschedulableAfter`/`cordonAfter` to be consistent with `evictAfter` -- or maybe we should just do whatever we can now, and aim to replace all of this with an actual scheduling constraint language at some point in the future?)

marun: The hardest task of software engineering: naming ;)

Maybe `unscheduleAfter` to be specific about what is being controlled? But I digress. I don't want to block merge of this on naming.

pkg/reconciler/namespace/namespace.go
```diff
@@ -211,11 +225,15 @@ func (c *Controller) isValidCluster(ctx context.Context, lclusterName, clusterNa
 	if err != nil {
 		return false, err
 	}
-	isReady := conditions.IsTrue(cluster, clusterv1alpha1.ClusterReadyCondition)
-	if !isReady {
+	if ready := !conditions.IsTrue(cluster, clusterv1alpha1.ClusterReadyCondition); !ready {
```
marun, line 228: I think you mean `ready := conditions.IsTrue`?

marun: I caught this in the process of preparing #559.

marun: You may want to include https://github.com/kcp-dev/kcp/pull/559/commits/631c173a8f195759fe85201939ac95c93e62d6d9 in your PR:

 - ensures test updates to clusters are retried. Otherwise the test is prone to racing with cluster-updating controllers and updates will randomly fail. Other options including be patching or using SSA instead of `Update()` 
 - adds checks of cluster readiness before checking that a namespace is scheduled to a cluster. This ensures against timing issues with the controller responsible for setting cluster readiness.

imjasonh: This is done now. If we're okay with this change as-is, we can just include your change in a followup PR. wdyt?

marun: My suggestion is to add that commit to this PR. Otherwise I think you'll have trouble passing e2e.

imjasonh: I've incorporated your change (thanks for that!) and also changed Updates into Patches, to avoid races. PTAL.

pkg/apis/cluster/v1alpha1/types.go
```diff
@@ -56,6 +56,18 @@ var _ conditions.Setter = &Cluster{}
...
+	// Unschedulable controls cluster schedulability of new workloads. By
+	// default, cluster is schedulable.
+	// +optional
+	// +kubebuilder:default=false
+	Unschedulable bool `json:"unschedulable"`
+
+	// EvictAfter controls cluster schedulability of new and existing workloads.
+	// After the EvictAfter time, any workload scheduled to the cluster
+	// will be unassigned from the cluster.
+	// By default, cluster is not evicted.
```
marun, line 69: (No action required) s/cluster is/workloads scheduled to the cluster are/

test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -158,20 +166,51 @@ func TestNamespaceScheduler(t *testing.T) {
...
 			require.NoError(t, err, "failed to delete cluster1 %q", cluster1.Name)
 
+			err = expect(namespace, unschedulableMatcher())
+			require.NoError(t, err, "did not see namespace marked unschedulable")
+
+			t.Log("Uncordon the new cluster and expect the namespace to end up scheduled to it.")
+
+			schedulablePatchBytes := []byte(`[{"op":"replace","path":"/spec/unschedulable","value":false}]`) // uncordoned.
+			cluster2, err = clusterClient.Patch(ctx, cluster2.Name, types.JSONPatchType, schedulablePatchBytes, metav1.PatchOptions{})
+			require.NoError(t, err, "failed to uncordon cluster2")
```
marun, line 187: (No action required) Maybe wait for cluster2 to be marked ready?

marun (approved):
Some suggestions inline, but I would be happy to see this merge.

test/e2e/reconciler/namespace/controller_test.go
imjasonh: Yeah, I believe your commit had it wait to become Ready here, but I think that was actually unnecessary -- at least while it only considers the `Ready` condition, and not the value of `.spec.unschedulable`.

Setting `unschedulable=true` (or `evictAfter` in the past) doesn't remove the `Ready` condition (as of right now), though perhaps it should?

marun: Not necessary, just defensive. If the test doesn't check that a cluster is ready, there is a chance the agent responsible for setting the ready condition might not have done so and the scheduled check would time out rather than failing for a detectable reason.

marun (approved):

PR #526, opened 2022-02-16
Default `make test-e2e` to a single run

marun:
Previously, `make test-e2e` would execute each test 5 times which could be a surprising result for some. Developers would commonly have to set `COUNT=1` if they wanted quick feedback on changes they were making locally. This commit defaults `make test-e2e` to executing a single run, with multiple runs configurable by setting `COUNT=`.

In keeping with this change, the `e2e-single-run` ci job is renamed to `e2e` and no longer sets an explicit `COUNT=` to ensure it uses the default behavior of `make test-e2e`. The previous `e2e` suite - executing 5 runs - is accordingly renamed `e2e-multiple-runs`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ $(OPENSHIFT_GOIMPORTS):
 imports: $(OPENSHIFT_GOIMPORTS)
 	$(OPENSHIFT_GOIMPORTS) -m github.com/kcp-dev/kcp
 
-COUNT ?= 5
+COUNT ?= 1
 E2E_PARALLELISM ?= 1
 
 .PHONY: test-e2e
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -80,8 +80,8 @@ jobs:
       - run: make build
       - run: PATH="${PATH}:$(pwd)/bin/" make test
 
-  e2e-single-run:
-    name: e2e-single-run
+  e2e:
+    name: e2e
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -97,7 +97,7 @@ $(OPENSHIFT_GOIMPORTS):
 imports: $(OPENSHIFT_GOIMPORTS)
 	$(OPENSHIFT_GOIMPORTS) -m github.com/kcp-dev/kcp
 
-COUNT ?= 5
+COUNT ?= 1
```
ncdc, line 100: This disables support for cached test results, which some folks might want. Should I file a separate issue to discuss if we want to support that or not?

ncdc: Let's see if anyone else has opinions - @kcp-dev/kcp-prototypers?

marun: Updated to set continue always setting `-count` for e2e, since the dependency on external binaries doesn't really allow for caching. For unit tests, only set -count if an explicit value for COUNT is provided.

Makefile
```diff
@@ -97,7 +97,18 @@ $(OPENSHIFT_GOIMPORTS):
...
 	$(OPENSHIFT_GOIMPORTS) -m github.com/kcp-dev/kcp
 
-COUNT ?= 5
+# Default to not supply `-count` to unit tests to enable
+# caching. Setting COUNT explicitly will supply the argument and
+# disable the caching.
+ifdef COUNT
+UT_COUNT_ARG = -count $(COUNT)
+else
+# If value is set for COUNT, default to 1 to ensure e2e tests don't
```
ncdc, line 106:
```suggestion
# If value is not set for COUNT, default to 1 to ensure e2e tests don't
```

?

Makefile
ncdc: It turns out that enabling `-coverprofile` (which we do) negates caching when `-count` is unset, e.g.:

```
go test -race ./pkg/crdpuller
ok      github.com/kcp-dev/kcp/pkg/crdpuller    (cached)

go test -race -coverprofile=coverage.txt ./pkg/crdpuller
ok      github.com/kcp-dev/kcp/pkg/crdpuller    0.653s  coverage: 24.7% of statements
```

So we either need to abandon this (for now?), or make coverage off by default.

marun: Arg. Open to either, which would you prefer?

ncdc: Let's keep coverage on by default for the time being, so I guess let's undo this bit if that's ok?

marun: Done

ncdc (approved):

PR #529, opened 2022-02-17
server: avoid panics through start-kube-namespace-controller context closing

sttts:
Hooks returning non-nil (even during termination because of context cancellation) lead to klog.Fatal and hence giant stack traces.


pkg/server/server.go
```diff
@@ -265,14 +265,15 @@ func (s *Server) Run(ctx context.Context) error {
...
-		wait.PollInfiniteWithContext(goContext(ctx), time.Second, func(ctx context.Context) (bool, error) {
+		if err := wait.PollInfiniteWithContext(goContext(ctx), time.Second, func(ctx context.Context) (bool, error) {
 			if err := s.bootstrapCRDs(ctx, apiextensionsClusterClient.Cluster(helper.OrganizationCluster).ApiextensionsV1().CustomResourceDefinitions()); err != nil {
 				klog.Errorf("failed to bootstrap CRDs: %v", err)
 				return false, nil // keep going
 			}
 			return true, nil
-		})
+		}); err != nil {
+			return nil // don't klog.Fatal. This only happens when context is cancelled.
```
davidfestal, line 275: Maybe disable the linter here ?

sttts: done

davidfestal (approved):

PR #530, opened 2022-02-17
e2e/framework: allow direct log output of kcp server

sttts:
Helps with debugging to see immediate what's going on.


stevekuznetsov (changes requested):
Could we please just do [this](https://github.com/openshift/ci-tools/blob/0d5af434229c81dd039749fe4ec8589e958d54f1/pkg/testhelper/accessory.go#L154-L156)?

```go
if stream {
	writers = append(writers, os.Stdout)
}
```

test/e2e/framework/test.go
```diff
@@ -36,4 +36,7 @@ type RunningServer interface {
 type KcpConfig struct {
 	Name string
 	Args []string
+
+	LogToConsole bool
```
marun, line 40: https://github.com/kcp-dev/kcp/pull/502 proposes to add test args to enable configuring running servers and controllers in process. Maybe rebase on that and add another arg for enabling logging to the console? Unless I'm missing something, I think it would be preferable to allow configuring the test run via cli args vs having to edit code.

sttts:
Updated to use a library.

PR #533, opened 2022-02-17
Generators outside gopath

### LINKED ISSUE #531: `make codegen` doesn't work outside of GOPATH
Repro:

1. clone the repo to somewhere outside of GOPATH
2. update some `types.go` to add a field, etc.
3. `make codegen` (succeeds)
4. deepcopy code isn't updated -- if you push a PR in this state, CI will fail ([example](https://github.com/kcp-dev/kcp/runs/5222837676?check_suite_focus=true))

---

This is using `controller-gen` v0.7.0, and ultimately I think it's a bug on their end.

We could either have `make codegen` detect you're not in GOPATH and fail more loudly, or do something like have it use a temp directory, make it GOPATH, do codegen and copy the result back to the original location.

### LINKED ISSUE #532: third_party: conditions deepcopy functions are not being generated
1. We are reusing these from Cluster API
2. Cluster API uses kubebuilder/controller-gen markers at the package level to enable deep copy generation
3. We didn't reuse Cluster API's `groupversion_info.go` file, which contains the markers
4. We didn't add our own markers to our `doc.go`
5. Therefore deepcopy generation isn't happening

ncdc:
~Requires https://github.com/kcp-dev/kubernetes/pull/44 to be merged first~

Fixes #531 
Fixes #532 

**NOTE**: now that we're correctly generating deepcopy functions for the code in third_party, those header files do have the copyright header change (to use our boilerplate).


go.mod
```diff
@@ -29,36 +29,36 @@ require (
 	k8s.io/klog/v2 v2.30.0
 	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65
 	k8s.io/kubectl v0.0.0
-	k8s.io/kubernetes v0.0.0
+	k8s.io/kubernetes v1.23.4
```
ncdc, line 32: Not sure why this is now showing a real version? This is after `go mod tidy` and even if I put in `v0.0.0` manually, it keeps coming back as `v1.23.4`. @sttts any ideas?

ncdc: Still not sure what's up here...

imjasonh: That is weird, but I also don't know that it matter since it gets `replace`d below anyway.

sttts (approved):

PR #534, opened 2022-02-17
Strip finalizers and owner references before syncing downstream

astefanutti:
This PR strips finalizers before syncing downstream, to prevent deletion of downstream resources from being blocked. It also strips owner references, to avoid broken references and orphaning, and to make sure cascading deletion happens only once upstream.

### DIFF
--- a/pkg/syncer/specsyncer.go
+++ b/pkg/syncer/specsyncer.go
@@ -155,6 +155,9 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
 	downstreamObj.SetNamespace(downstreamNamespace)
 	downstreamObj.SetManagedFields(nil)
 	downstreamObj.SetClusterName("")
+	// Deletion fields are immutable and set by the downstream API server
+	downstreamObj.SetDeletionTimestamp(nil)
+	downstreamObj.SetDeletionGracePeriodSeconds(nil)
 	// Strip owner references, to avoid orphaning by broken references,
 	// and make sure cascading deletion is only performed once upstream.
 	downstreamObj.SetOwnerReferences(nil)
--- a/pkg/syncer/specsyncer.go
+++ b/pkg/syncer/specsyncer.go
@@ -155,23 +155,16 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
 	downstreamObj.SetNamespace(downstreamNamespace)
 	downstreamObj.SetManagedFields(nil)
 	downstreamObj.SetClusterName("")
-
-	ownedByLabel := downstreamObj.GetLabels()["kcp.dev/owned-by"]
-	var ownerReferences []metav1.OwnerReference
-	for _, reference := range downstreamObj.GetOwnerReferences() {
-		if reference.Name == ownedByLabel {
-			continue
-		}
-		ownerReferences = append(ownerReferences, reference)
-	}
-	downstreamObj.SetOwnerReferences(ownerReferences)
+	// Strip owner references, to avoid orphaning by broken references,
+	// and make sure cascading deletion is only performed once upstream.
+	downstreamObj.SetOwnerReferences(nil)
+	// Strip finalizers to avoid the deletion of the downstream resource from being blocked.
+	downstreamObj.SetFinalizers(nil)
 
 	// Run name transformations on the downstreamObj.
 	transformName(downstreamObj, KcpToPhysicalCluster)
 
-	// TODO: wipe things like finalizers, owner-refs and any other life-cycle fields. The life-cycle
-	//       should exclusively owned by the syncer. Let's not some Kubernetes magic interfere with it.
-
+	// Marshalling the unstructured object is good enough as SSA patch
 	data, err := json.Marshal(downstreamObj)
 	if err != nil {
 		return err


ncdc:
@astefanutti thanks for your PR! This looks like something that would benefit from an issue describing the bug/feature request, as well as careful planning and design. Would you have time to create an issue?

Because of this, I am moving this to the `TBD` milestone. We will be frequently reviewing things in there as part of broader roadmap planning efforts, and then moving them to specific milestones.

pkg/syncer/specsyncer.go
```diff
@@ -149,20 +149,13 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
 	obj.SetResourceVersion("")
 	obj.SetNamespace(namespace)
 	obj.SetManagedFields(nil)
+	// Strip owner references, to avoid orphaning by broken references,
+	// and make sure cascading deletion is only performed once upstream.
+	obj.SetOwnerReferences([]metav1.OwnerReference{})
+	// Strip finalizers to avoid the deletion of the downstream resource from being blocked.
+	obj.SetFinalizers([]string{})
```
sttts, line 156: there will be finalizers added by downstream controllers. What happens with them?

sttts:
@marun can you review and shepherd this?

pkg/syncer/specsyncer.go
astefanutti: As Server-Side Apply is used to update the object downstream, and the `finalizers` key is removed from the SSA patch, the downstream finalizers are left untouched, so that controllers responsible for these finalizers can perform their logic.

I've added a commit that sets the deletion propagation policy to `Foreground`, so that the deletion cascades the entire ownership graph, and the `deletionTimestamp` field is set, so that the object remains in the key-value store until all its finalizers are removed.

astefanutti:
@marun this is ready for your review. Let me know if you prefer to have it based on #670. On the other hand, as it stands, #670 is opinionated per single GRV, while that one is transversal.

pkg/syncer/specsyncer.go
```diff
@@ -151,27 +151,24 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
 
 	downstreamObj := upstreamObj.DeepCopy()
 	downstreamObj.SetUID("")
+	downstreamObj.SetCreationTimestamp(metav1.Time{})
```
marun, line 154: No need to set this, it gets set by the api.

pkg/syncer/specsyncer.go
```diff
@@ -151,27 +151,24 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
...
-	var ownerReferences []metav1.OwnerReference
-	for _, reference := range downstreamObj.GetOwnerReferences() {
-		if reference.Name == ownedByLabel {
-			continue
-		}
-		ownerReferences = append(ownerReferences, reference)
-	}
-	downstreamObj.SetOwnerReferences(ownerReferences)
+	// Deletion fields are immutable and set by the downstream API server
+	downstreamObj.SetDeletionTimestamp(nil)
```
marun, line 160: If they are immutable, do they need to be cleared?

marun (commented):
I think the approach is fine for now but definitely all object mutation should be externalized to simplify testing.

pkg/syncer/specsyncer.go
astefanutti: Removed 👍🏼.

pkg/syncer/specsyncer.go
astefanutti: I think so. The update requests get rejected by the downstream API servers otherwise.

astefanutti: By removing them, they get removed from the SSA patch, do that there are left untouched downstream.

pkg/syncer/specsyncer.go
```diff
@@ -155,23 +155,19 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
 	downstreamObj.SetNamespace(downstreamNamespace)
 	downstreamObj.SetManagedFields(nil)
 	downstreamObj.SetClusterName("")
-
-	ownedByLabel := downstreamObj.GetLabels()["kcp.dev/owned-by"]
```
marun, line 159: Note to other reviewers: OwnerReferences being keyed to UID means they aren't portable across clusters.

pkg/syncer/specsyncer.go
```diff
@@ -155,23 +155,19 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
...
-	}
-	downstreamObj.SetOwnerReferences(ownerReferences)
+	// Deletion fields are immutable and set by the downstream API server
+	downstreamObj.SetDeletionTimestamp(nil)
+	downstreamObj.SetDeletionGracePeriodSeconds(nil)
+	// Strip owner references, to avoid orphaning by broken references,
+	// and make sure cascading deletion is only performed once upstream.
+	downstreamObj.SetOwnerReferences(nil)
+	// Strip finalizers to avoid the deletion of the downstream resource from being blocked.
+	downstreamObj.SetFinalizers(nil)
```
marun, line 165: Note to other reviews: Finalizers are not portable across clusters. They need to be set and removed by the same actor actively reconciling a given API.

marun (approved):
Thank you!

PR #508, opened 2022-02-15
Implement root workspace and organization workspaces

sttts:
This PR lifts the kcp platform to have real (Cluster)Workspace support by adding a hierarchy of the

1. root workspace (replacing admin)
2. organization workspaces (= workspaces of type Organization defined in root workspace)
3. non-org workspaces (= workspaces of other type, `Universal` by default, defined in org workspaces.

For this, the following bigger changes are included:
1. all controllers and e2e tests are extended to not depend on an admin workspace anymore. I.e. they create real org workspaces and in there real universal workspaces to do the actual work (cc @marun this might open up the possibility to use shared fixtures in a number of tests).
2. logical clusters are separated by `:` instead of `_`, mostly for cosmetical reason, and to align with the use of `:` in Kubernetes, e.g. in RBAC roles and bindings. Underscore feels foreign. With that a cluster name is either a) root b) <org>:<ws> or c) system:*. Org workspaces have logical cluster name of the shape `root:<org>`.
3. `Workspaces` and renamed to `ClusterWorkspaces` to allow us to have different sources of workspace definition in the future. The virtual personal workspace apiserver will return `Workspace`s which are simplified, supportable projections of `ClusterWorkspace`s. The latter can include low-level fields like location information, movement statemachine, quota, accounting data and more. We don't want to leak those as a service provider.
4. org workspaces are `ClusterWorkspace`s of type `Organization`. A `clusterworkspacetype_organization` controller is added and an admission plugin. The controllers bootstrap CRDs, and makes use of the initializer field of `ClusterWorkspace`s. The admission plugin makes the type field immutable and adds initializers defined by `ClusterWorkspaceType` objects. Providers of custom-workspaces can use that to have their own initialization flow.
5. the admin logical cluster is renamed to `system:admin`. In general, `system:*` logical clusters are supposed to be shard-local.
6. by default (through config/root) a `root:default` organisation is created on kcp startup. The default context in the `.kcp/admin.kubeconfig` points to `root:default`. In addition, there are `cross-cluster`(as before), `root` for the root workspace and `system:admin` for the former admin workspace.

Follow-ups:
- [ ] add authorization to ClusterWorkspaceType use @s-urbaniak 
- [ ] move test/e2e/virtual/framework/compositecmd into pkg/virtual @davidfestal 
- [ ] rewrite workspaceindex test, and adapt controller @sttts @stevekuznetsov 
- [ ] repo-wide require.NoErrorf => require.NoError @marun 
[...]

PR #537, opened 2022-02-21
Add a label to deactivate automatic namespace scheduling

### LINKED ISSUE #494: Namespace scheduler should support ignoring namespaces
As per internal discussion, it should be possible to mark a namespace in kcp as not being intended to be scheduled to a pcluster. So long as the namespace scheduler doesn't assign a namespace to a cluster, the namespace and its contents would never be created in a pcluster.

Suggested implementation would have the namespace scheduler ignore namespaces with the label `kcp.dev/ignore: true`. A label is prefered to an annotation so it can be selected on. 

Use case: Namespaces containing resources used to configure controllers operating against KCP but not intended to be 
propagated to pclusters.

astefanutti:
Currently the namespace controller owns the responsibility of assigning namespaces to physical clusters. The only option for other controllers, like the `deployment-splitter` one, or users manually, to distribute resources across physical clusters is to deactivate the namespace controller by starting KCP with `--run-controllers=false --unsupported-run-individual-controllers="workspace-scheduler,cluster"`.

Meanwhile more advanced scheduling / placement scenarios are supported, this PR proposes to introduce the `experimental.scheduling.kcp.dev/disabled` label, that can be added to namespaces, in order to deactivate the automatic placement to physical clusters, performed by the namespace controller.

Setting this label to `false` prevents the namespace controller from overriding the scheduling logic performed by the `deployment-splitter` controller, but also manual scheduling, when working on global load-balancing use cases for example.

Fixes #494

### DIFF
--- a/pkg/reconciler/namespace/namespace.go
+++ b/pkg/reconciler/namespace/namespace.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -28,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/clusters"
@@ -38,28 +40,35 @@ import (
 	"github.com/kcp-dev/kcp/third_party/conditions/util/conditions"
 )
 
-const clusterLabel = "kcp.dev/cluster"
+const (
+	clusterLabel          = "kcp.dev/cluster"
+	ScheduleDisabledLabel = "experimental.scheduling.kcp.dev/disabled"
+)
 
 var (
-	unscheduledSelector        labels.Selector
-	scheduledToNothingSelector labels.Selector
+	scheduleRequirement           labels.Requirement
+	scheduleEmptyLabelRequirement labels.Requirement
+	unscheduledRequirement        labels.Requirement
 )
 
 func init() {
-	var err error
-
-	// This matches namespaces without the label
-	unscheduled := "!" + clusterLabel
-	unscheduledSelector, err = labels.Parse(unscheduled)
-	if err != nil {
-		klog.Fatalf("error parsing label selector %q: %v", unscheduled, err)
+	// This matches namespaces that haven't been scheduled yet
+	if req, err := labels.NewRequirement(clusterLabel, selection.DoesNotExist, []string{}); err != nil {
+		klog.Fatalf("error creating the cluster label requirement: %v", err)
+	} else {
+		unscheduledRequirement = *req
 	}
-
-	// This matches namespaces with the label set to ""
-	scheduledToNothing := clusterLabel + "="
-	scheduledToNothingSelector, err = labels.Parse(scheduledToNothing)
-	if err != nil {
-		klog.Fatalf("error parsing label selector %q: %v", scheduledToNothing, err)
+	// This matches namespaces with the cluster label set to ""
+	if req, err := labels.NewRequirement(clusterLabel, selection.Equals, []string{""}); err != nil {
+		klog.Fatalf("error creating the cluster label requirement: %v", err)
+	} else {
+		scheduleEmptyLabelRequirement = *req
+	}
+	// This matches namespaces that should be scheduled automatically by the namespace controller
+	if req, err := labels.NewRequirement(ScheduleDisabledLabel, selection.DoesNotExist, []string{}); err != nil {
+		klog.Fatalf("error creating the schedule label requirement: %v", err)
+	} else {
+		scheduleRequirement = *req
 	}
 }
 
@@ -88,6 +97,13 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName string,
 		return err
 	}
 
+	if !scheduleRequirement.Matches(labels.Set(ns.Labels)) {
+		// Do not schedule the resource transitively, and let external controllers
+		// or users be responsible for it, consistently with the scheduling of the
+		// namespace.
+		return nil
+	}
+
 	lbls := unstr.GetLabels()
 	if lbls == nil {
 		lbls = map[string]string{}
@@ -100,10 +116,13 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName string,
 	}
 
 	// Update the resource's assignment.
-	klog.Infof("Patching to update cluster assignment for %s %s/%s: %q -> %q", gvr, ns.Name, unstr.GetName(), old, new)
-	if _, err := c.dynClient.Cluster(lclusterName).Resource(*gvr).Namespace(ns.Name).Patch(ctx, unstr.GetName(), types.MergePatchType, clusterLabelPatchBytes(new), metav1.PatchOptions{}); err != nil {
+	patchType, patchBytes := ClusterLabelPatchBytes(new)
+	if _, err = c.dynClient.Cluster(lclusterName).Resource(*gvr).Namespace(ns.Name).
+		Patch(ctx, unstr.GetName(), patchType, patchBytes, metav1.PatchOptions{}); err != nil {
 		return err
 	}
+	klog.Infof("Patched cluster assignment for %s %s/%s: %q -> %q", gvr, ns.Name, unstr.GetName(), old, new)
+
 	return nil
 }
 
@@ -112,7 +131,7 @@ func (c *Controller) reconcileGVR(ctx context.Context, gvr schema.GroupVersionRe
 	listers, _ := c.ddsif.Listers()
 	lister, found := listers[gvr]
 	if !found {
-		return fmt.Errorf("Informer for %s is not synced; re-enqueueing", gvr)
+		return fmt.Errorf("informer for %s is not synced; re-enqueueing", gvr)
 	}
 
 	// Enqueue workqueue items to reconcile every resource of this type, in
@@ -132,25 +151,35 @@ func (c *Controller) patchStatus(ctx context.Context, ns *corev1.Namespace, upda
 	if err != nil {
 		return err
 	}
-	_, err = c.kubeClient.Cluster(ns.ClusterName).CoreV1().Namespaces().Patch(ctx, ns.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{}, "status")
+	_, err = c.kubeClient.Cluster(ns.ClusterName).CoreV1().Namespaces().
+		Patch(ctx, ns.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{}, "status")
 	if err != nil {
 		return fmt.Errorf("failed to patch status on namespace %q|%q: %w", ns.ClusterName, ns.Name, err)
 	}
 	return nil
 }
 
-// Set the schedulable condition on namespaces and not the resources it
+// Set the scheduled condition on namespaces and not the resources it
 // contains. Not all native types support status conditions (e.g. configmaps
 // and secrets) and there is no guarantee that arbitrary CRDs will support
 // conditions.
-func (c *Controller) updateSchedulableCondition(ctx context.Context, ns *corev1.Namespace, schedulable bool) error {
-	klog.Infof("Patching namespace %q|%q status to indicate schedulable=%t", ns.ClusterName, ns.Name, schedulable)
+func (c *Controller) updateScheduledCondition(ctx context.Context, ns *corev1.Namespace, scheduled, disabled bool) error {
+	klog.Infof("Patching namespace %q|%q status to indicate scheduled=%t", ns.ClusterName, ns.Name, scheduled)
 
-	if schedulable {
+	if scheduled {
 		return c.patchStatus(ctx, ns, func(conditionsSetter conditions.Setter) {
 			conditions.MarkTrue(conditionsSetter, NamespaceScheduled)
 		})
 	}
+
+	if disabled {
+		return c.patchStatus(ctx, ns, func(conditionsSetter conditions.Setter) {
+			conditions.MarkFalse(conditionsSetter, NamespaceScheduled, NamespaceReasonSchedulingDisabled,
+				conditionsv1alpha1.ConditionSeverityNone, // NamespaceCondition doesn't support severity
+				"Automatic scheduling is deactivated and can be performed by setting the cluster label manually.")
+		})
+	}
+
 	return c.patchStatus(ctx, ns, func(conditionsSetter conditions.Setter) {
 		conditions.MarkFalse(conditionsSetter, NamespaceScheduled, NamespaceReasonUnschedulable,
 			conditionsv1alpha1.ConditionSeverityNone, // NamespaceCondition doesn't support severity
@@ -199,10 +228,12 @@ func (c *Controller) assignCluster(ctx context.Context, ns *corev1.Namespace) er
 
 	oldClusterName := ns.Labels[clusterLabel]
 	if oldClusterName != newClusterName {
-		if _, err := c.kubeClient.Cluster(ns.ClusterName).CoreV1().Namespaces().Patch(ctx, ns.Name, types.MergePatchType, clusterLabelPatchBytes(newClusterName), metav1.PatchOptions{}); err != nil {
+		patchType, patchBytes := ClusterLabelPatchBytes(newClusterName)
+		if _, err = c.kubeClient.Cluster(ns.ClusterName).CoreV1().Namespaces().
+			Patch(ctx, ns.Name, patchType, patchBytes, metav1.PatchOptions{}); err != nil {
 			return err
 		}
-		klog.Infof("Patched to update cluster assignment for namespace %q|%q: %q -> %q", ns.ClusterName, ns.Name, oldClusterName, newClusterName)
+		klog.Infof("Patched cluster assignment for namespace %q|%q: %q -> %q", ns.ClusterName, ns.Name, oldClusterName, newClusterName)
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


ncdc:
Thanks for the PR. Is this for #494?

ncdc:
cc @marun @imjasonh

astefanutti:
> Thanks for the PR. Is this for #494?

Ah I missed it, It looks like exactly for it, only the name of the label differs, naming is the hardest... :)

pkg/reconciler/namespace/namespace.go
```diff
@@ -268,6 +275,15 @@ func (c *Controller) reconcileNamespace(ctx context.Context, lclusterName string
 		ns.Labels = map[string]string{}
 	}
 
+	if ns.Labels[clusterLabel] == "" && !scheduleRequirement.Matches(labels.Set(ns.Labels)) {
+		klog.Infof("Skipping placement for namespace %q|%q", lclusterName, ns.Name)
+		return c.patchStatus(ctx, ns, func(conditionsSetter conditions.Setter) {
+			conditions.MarkFalse(conditionsSetter, NamespaceScheduled, NamespaceReasonUnschedulable,
```
marun, line 281: Maybe a new status reason is suggested to differentiate 'unable to schedule' from 'scheduling disabled'?

pkg/reconciler/namespace/namespace.go
```diff
@@ -190,11 +196,11 @@ func (c *Controller) assignCluster(ctx context.Context, ns *corev1.Namespace) er
 
 	oldClusterName := ns.Labels[clusterLabel]
 	if oldClusterName != newClusterName {
-		klog.Infof("Patching to update cluster assignment for namespace %q|%q: %q -> %q", ns.ClusterName, ns.Name, oldClusterName, newClusterName)
-		if _, err := c.kubeClient.Cluster(ns.ClusterName).CoreV1().Namespaces().Patch(ctx, ns.Name, types.MergePatchType, clusterLabelPatchBytes(newClusterName), metav1.PatchOptions{}); err != nil {
+		klog.Infof("Patching to update scheduled cluster for namespace %q|%q: %q -> %q", ns.ClusterName, ns.Name, oldClusterName, newClusterName)
```
marun, line 199: Why do you propose to update 'cluster assignment' to 'scheduled cluster' everywhere? A resource is scheduled to a cluster. A resource has a cluster assignment/a cluster assigned to it. A resource doesn't have a scheduled cluster.

pkg/reconciler/namespace/namespace.go
```diff
@@ -268,6 +275,15 @@ func (c *Controller) reconcileNamespace(ctx context.Context, lclusterName string
 		ns.Labels = map[string]string{}
 	}
 
+	if ns.Labels[clusterLabel] == "" && !scheduleRequirement.Matches(labels.Set(ns.Labels)) {
```
marun, line 278: I think it may be potentially confusing that the logic for disabling scheduling for a namespace is both here and in `ensureScheduled`. Maybe localize the change to simplify maintenance? e.g.

```golang
schedulingDisabled := !scheduleRequirement.Matches(labels.Set(ns.Labels))
if schedulingDisabled {
  return c.ensureSchedulingDisabledStatus(...)
}
```

I'm also not sure why it makes sense to check for an empty cluster assignment. Once a namespace has been labeled with `kcp.dev/schedule=false`, why should the controller consider the value of `kcp.dev/cluster` for that namespace?

pkg/reconciler/namespace/namespace.go
```diff
@@ -37,28 +38,28 @@ import (
...
 
 func init() {
-	var err error
-
-	// This matches namespaces without the label
-	unscheduled := "!" + clusterLabel
-	unscheduledSelector, err = labels.Parse(unscheduled)
-	if err != nil {
-		klog.Fatalf("error parsing label selector %q: %v", unscheduled, err)
+	// This matches namespaces that haven't been scheduled yet
```
marun, line 52: Why does it make sense to make this check more restrictive (label does not exist) than previous (label does not exist or does exist and has a non-empty value)? 

While I can see that your intent is that the controller remove the label, if the label were set to an empty value (e.g. `kcp.dev/cluster=`), the code as proposed would appear neither to ensure removal of the label or ensure queueing of the namespace when cluster readiness status changed.

sttts:
Am uncomfortable with adding features like this without a bigger picture of direction. Adding a label to control who schedules a namespace is clearly possible, but doesn't add much deeper insights in what we are building.

Can we turn this into some exploration doc about how deployment-splitter like behaviour fits into the bigger TMC picture?

astefanutti:
> Am uncomfortable with adding features like this without a bigger picture of direction. Adding a label to control who schedules a namespace is clearly possible, but doesn't add much deeper insights in what we are building.
> 
> Can we turn this into some exploration doc about how deployment-splitter like behaviour fits into the bigger TMC picture?

I agree, this is also my intuition as I've been working through this PR. It seems some exploration is required to address the scheduling / placement concern holistically, possibly looking at https://github.com/open-cluster-management-io/placement, and OCM in general.

On the other hand, the namespace controller, as it stands, pre-empts the scheduling, so the only solution to demo use cases that distribute namespace resources across clusters, is to deactivate it. So maybe fixing #494 in the short term with this PR and do the exploration work in the medium term is an acceptable tradeoff?

pkg/reconciler/namespace/namespace.go
astefanutti: Agreed, I've reverted the change. I think that was consistency gone wrong :)

pkg/reconciler/namespace/namespace.go
astefanutti: I agree, I've pushed 395fc13886175bcda5e5a1be685ba7729fbd5c28 (I'll squash once the review is complete), that adds the the `SchedulingDisabled` condition reason.

It has actually highlighted an issue in the condition update logic, where the condition was not created until the namespace was actually scheduled. It's fixed with 395fc13886175bcda5e5a1be685ba7729fbd5c28.

test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -28,21 +28,22 @@ import (
...
 	clientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	"github.com/kcp-dev/kcp/pkg/client/clientset/versioned/typed/cluster/v1alpha1"
 	nscontroller "github.com/kcp-dev/kcp/pkg/reconciler/namespace"
 	"github.com/kcp-dev/kcp/test/e2e/framework"
 )
 
 const clusterLabel = "kcp.dev/cluster"
 
 func TestNamespaceScheduler(t *testing.T) {
-	t.Parallel()
```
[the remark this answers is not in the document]
astefanutti: The namespace controller relies on the physical clusters created in the test fixture, and now that there are more than one test case, that manipulate that "shared state", that leads to race conditions. That explains why I've deactivated parallel execution of the test cases, as I wasn't sure what's the best approach. To isolate that clusters "state", each test case could use a dedicated workspace, but I went for the path of least resistance, so as to get your feedback before proceeding.

test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -73,13 +74,148 @@ func TestNamespaceScheduler(t *testing.T) {
...
+					},
+					Spec: clusterv1alpha1.ClusterSpec{
+						KubeConfig: string(server1Kubeconfig),
+					},
+				}, metav1.CreateOptions{})
+				require.NoError(t, err, "failed to create cluster1")
+
+				t.Log("Create a namespace with manual scheduling and expect it to be marked unscheduled")
+
+				namespace, err := server.client.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
```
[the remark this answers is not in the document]
astefanutti: I hesitated on that also :) That new "test case" can definitely be added to the existing one. I went for the "single responsibility" principle (if such thing applies to test cases :)), to avoid going down the road of writing a giant test case, but that may have been a premature decision.

That being said, my understanding is that the biggest contributor to total execution time is the creation and tear down of the servers, and adding an extra test case is negligible.

pkg/reconciler/namespace/namespace.go
```diff
@@ -97,10 +98,15 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName string,
 		// Already assigned to the right cluster.
 		return nil
 	}
+	if new == "" {
+		// The namespace hasn't been assigned a cluster.
```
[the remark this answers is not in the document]
astefanutti: Thinking about it more, that was only an approximation of what is probably a more general solution, that is, do not schedule resources transitively for namespaces with `kcp.dev/schedule=false`. That label means, the namespace controller should not be responsible for the scheduling of that namespace, nor its resources, so the responsibility is fully delegated to an external controller or the user.

I've amended the logic based on that rationale in b386ec6b88652bac5d1dbe720824bd624285ef99.

pkg/reconciler/namespace/namespace.go
astefanutti: Totally agree. I've updated the logic in 00b0aa88bc13e64e2ea8914ef6ad927ce220f5e2.

pkg/reconciler/namespace/namespace.go
astefanutti: My understanding is that the existing code has to deal with `kcp.dev/cluster=`, because the logic that updates the label does not handle the removal of the label, and simply sets it to `""`. So my rationale was that with properly handling the label deletion, there is no need to deal with `kcp.dev/cluster=` anymore.

astefanutti:
@marun thanks a lot for the review. I should have covered all your comments, if you want to give a second iteration on this.

pkg/reconciler/namespace/namespace.go
marun: I'm afraid it's not an option to assume that the label in question will only be managed by this controller. There is always the potential for other actors - including humans - to make changes that don't match this controller's expectations.  Not assuming anything about the input is also a good way to ensure forwards and backwards compatibility.

ncdc:
> Can we turn this into some exploration doc about how deployment-splitter like behaviour fits into the bigger TMC picture?

#557 

> On the other hand, the namespace controller, as it stands, pre-empts the scheduling, so the only solution to demo use cases that distribute namespace resources across clusters, is to deactivate it. So maybe fixing https://github.com/kcp-dev/kcp/issues/494 in the short term with this PR and do the exploration work in the medium term is an acceptable tradeoff?

I chatted with @sttts and we are 👍 on this plan. The label in this PR might not survive as-is for the long-term, but we will ensure that things like the deployment-splitter can coexist with namespace scheduling.

ncdc:
Would you want to also check for the label per resource, so that a namespace could be scheduled to a cluster, but you could opt out individual e.g. deployments?

ncdc:
@marun assigned to you 😄

sttts:
Please some clear indication that the label is experimental and going away. experimental.something.kcp.dev.

test/e2e/reconciler/namespace/controller_test.go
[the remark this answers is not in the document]
marun: I think it would be preferable for now to move the fixture instantiation inside `t.Run()` to ensure that each test gets clean fixture.

test/e2e/reconciler/namespace/controller_test.go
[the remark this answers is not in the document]
marun: I think that's fine for now. Note that corner-cases will ideally be validated in unit tests as per #559.

marun:
> Would you want to also check for the label per resource, so that a namespace could be scheduled to a cluster, but you could opt out individual e.g. deployments?

We had a use case for skipping entire namespaces as per #494. Do we also have one or more documented use cases for skipping individual resources? afaik the goal for this PR is allowing scheduling controllers (e.g. deployment-splitter) to manage specific types vs individual resources.

pkg/reconciler/namespace/namespace.go
marun: bump

pkg/reconciler/namespace/namespace.go
```diff
@@ -98,9 +106,10 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName string,
 		return nil
 	}
 
-	// Update the resource's assignment.
```
marun, line 101: nit: revert to 'assignment' from 'scheduled cluster'

pkg/reconciler/namespace/namespace.go
```diff
@@ -191,10 +209,10 @@ func (c *Controller) assignCluster(ctx context.Context, ns *corev1.Namespace) er
 	oldClusterName := ns.Labels[clusterLabel]
 	if oldClusterName != newClusterName {
 		klog.Infof("Patching to update cluster assignment for namespace %q|%q: %q -> %q", ns.ClusterName, ns.Name, oldClusterName, newClusterName)
-		if _, err := c.kubeClient.Cluster(ns.ClusterName).CoreV1().Namespaces().Patch(ctx, ns.Name, types.MergePatchType, clusterLabelPatchBytes(newClusterName), metav1.PatchOptions{}); err != nil {
+		t, patch := clusterLabelPatchBytes(newClusterName)
```
marun, line 212: nit: Please avoid using single-char variables except for receivers and loop indexes.

pkg/reconciler/namespace/namespace.go
```diff
@@ -236,24 +254,34 @@ func (c *Controller) ensureScheduled(ctx context.Context, lclusterName string, n
...
 func (c *Controller) ensureScheduledStatus(ctx context.Context, ns *corev1.Namespace) error {
-	scheduled := (ns.Labels[clusterLabel] != "")
-	statusChangeRequired := scheduled != IsScheduled(ns)
-	if statusChangeRequired {
-		return c.updateSchedulableCondition(ctx, ns, scheduled)
+	scheduled := ns.Labels[clusterLabel] != ""
+	disabled := !scheduleRequirement.Matches(labels.Set(ns.Labels))
+
+	if condition := conditions.Get(&NamespaceConditionsAdapter{ns}, NamespaceScheduled); condition != nil {
+		conditionStatus := scheduled != (condition.Status == corev1.ConditionTrue)
```
marun, line 274: nit: `s/conditionStatus/conditionStatusChanged/` ? i.e. indicate clearly in the name what the conditional indicates. Same comment for `conditionReason` below

test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -107,94 +244,43 @@ func TestNamespaceScheduler(t *testing.T) {
...
+			return fmt.Errorf("expected an unscheduled namespace, missing scheduled condition: %#v", object.Status.Conditions)
 		}
-		return nil
 	}
 }
 
 func scheduledMatcher(target string) namespaceExpectation {
 	return func(object *corev1.Namespace) error {
-		if !nscontroller.IsScheduled(object) {
+		if scheduled := nscontroller.IsScheduled(object); !scheduled {
```
marun, line 283: nit: This can be reverted

test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -230,3 +316,14 @@ func expectNamespaces(ctx context.Context, t *testing.T, client kubernetes.Inter
 		}, 30*time.Second)
 	}, nil
 }
+
+func patchNamespaceLabel(label, value string) []byte {
```
marun, line 320: Maybe just make this public in the target package so it can be reused?

pkg/reconciler/namespace/namespace.go
astefanutti: True. I've restore the existing logic.

pkg/reconciler/namespace/namespace.go
astefanutti: Reverted.

test/e2e/reconciler/namespace/controller_test.go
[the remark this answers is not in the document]
astefanutti: Agreed. I've moved the fixture instantiation where you suggested.

pkg/reconciler/namespace/namespace.go
astefanutti: I've renamed the variables.

pkg/reconciler/namespace/namespace.go
astefanutti: Agreed. I've updated both variables as per your suggestion.

test/e2e/reconciler/namespace/controller_test.go
astefanutti: Thanks for the catch! I had reverted half-way, now fully reverted.

test/e2e/reconciler/namespace/controller_test.go
astefanutti: I've made the function in the target package and removed that one.

astefanutti:
@marun thanks a lot for that second iteration. I should have addressed all your current comments, so it's available for another review cycle, and squashed into a single commit.

@sttts I've changed the label to `experimental.scheduling.kcp.dev/disabled`, and turned it into a _flag_ label, to remove ambiguity on the value.

marun:
Please rebase. And I'll be rebasing #559 on this PR to avoid causing you further pain.

astefanutti:
@marun thanks for the heads-up. Rebased 👍🏼.

marun:
I'm afraid #508 will need to land ahead of this, and I'll do the final review after the inevitable rebase.

astefanutti:
@marun thanks, again :) Rebased 👍🏼.

pkg/reconciler/namespace/namespace.go
```diff
@@ -323,10 +370,16 @@ func (c *Controller) reconcileNamespace(ctx context.Context, lclusterName string
 	return nil
 }
 
-// clusterLabelPatchBytes returns JSON patch bytes expressing an operation to
-// add or replace the cluster assignment label to the given value.
-func clusterLabelPatchBytes(val string) []byte {
-	return []byte(fmt.Sprintf(`
+// ClusterLabelPatchBytes returns JSON patch bytes expressing an operation to
+// add, replace to the given value, or delete the cluster assignment label.
+func ClusterLabelPatchBytes(val string) (types.PatchType, []byte) {
```
imjasonh, line 375: This seems to only be exported so that it can be called from e2e tests. If that's the case, I might suggest either reimplementing necessary bits in the test itself (i.e., test that the patch is `"[{"op":...}]` instead of reusing the helper), or perhaps moving the method to an `/internal/` package so that it can't be depended on outside of this Go module.

Exported methods have a nasty habit of getting depended on outside of our codebase, and sometimes beyond intended usage, which can make changes harder to make safely.

astefanutti: I agree. @marun what do you think of moving that function in the `/internal/` package? Could that be an acceptable compromise, as you raised concerns about duplicating it?

marun: Agreed that this detail shouldn't be exposed. I don't think jumping through the hoops of moving to internal is suggested though. How about we avoid using low-level patch semantics in testing and rely on `retry.RetryOnConflict` instead?

e.g. https://github.com/kcp-dev/kcp/blob/main/test/e2e/reconciler/ingress/controller_test.go#L102

pkg/reconciler/namespace/namespace.go
```diff
@@ -132,25 +151,35 @@ func (c *Controller) patchStatus(ctx context.Context, ns *corev1.Namespace, upda
 	if err != nil {
 		return err
 	}
-	_, err = c.kubeClient.Cluster(ns.ClusterName).CoreV1().Namespaces().Patch(ctx, ns.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{}, "status")
+	_, err = c.kubeClient.Cluster(ns.ClusterName).CoreV1().Namespaces().
+		Patch(ctx, ns.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{}, "status")
```
sttts, line 155: please split features from refactorings

pkg/reconciler/namespace/namespace.go
```diff
@@ -132,25 +151,35 @@ func (c *Controller) patchStatus(ctx context.Context, ns *corev1.Namespace, upda
...
 }
 
-// Set the schedulable condition on namespaces and not the resources it
+// Set the scheduled condition on namespaces and not the resources it
 // contains. Not all native types support status conditions (e.g. configmaps
 // and secrets) and there is no guarantee that arbitrary CRDs will support
 // conditions.
-func (c *Controller) updateSchedulableCondition(ctx context.Context, ns *corev1.Namespace, schedulable bool) error {
-	klog.Infof("Patching namespace %q|%q status to indicate schedulable=%t", ns.ClusterName, ns.Name, schedulable)
+func (c *Controller) updateScheduledCondition(ctx context.Context, ns *corev1.Namespace, scheduled, disabled bool) error {
```
sttts, line 166: please split features from refactorings

sttts:
As a general rule of thumb: we should try to split refactorings/cleanups from actual feature changes better. E.g. if we rename a func, a condition, or anything with lots of ripples: do that in its own commit.

marun:
There are some nits remaining (including not exporting the patch bytes method just for testing) that I will address in a refactor/cleanup I am working on (of which #559 is a part).

marun (approved):

PR #540, opened 2022-02-21
config: generalize to arbitrary resource bootstrapping

sttts:
To be used soon for the root workspace, and org workspaces.

Poor men's CVO.


config/helpers/bootstrap.go
```diff
@@ -0,0 +1,193 @@
...
+		return fmt.Errorf("failed to bootstrap CRDs into root workspace: %w", err)
+	}
+
+	// bootstrap non-crd resources
+	var transformers []TransformFileFunc
+	for _, opt := range opts {
+		transformers = append(transformers, opt.TransformFile)
+	}
+	return wait.PollImmediateInfiniteWithContext(ctx, time.Second, func(ctx context.Context) (bool, error) {
+		// +lint:noerrcheck
```
marun, line 92: (No action required) Why is this required here but not above line 77?

ncdc:
>Poor men's CVO.

What is CVO?

ncdc:
Can you rename `dummy-to-make-embed-fs-happy.yaml` to a more inclusive name (e.g. sentinel- or empty- or something like that)? Thanks!

config/helpers/bootstrap.go
```diff
@@ -0,0 +1,193 @@
...
+	mapper := restmapper.NewDeferredDiscoveryRESTMapper(memory.NewMemCacheClient(crdClient.Discovery()))
+
+	if err := wait.PollImmediateInfiniteWithContext(ctx, time.Second, func(ctx context.Context) (bool, error) {
+		if err := configcrds.Create(ctx, crdClient.ApiextensionsV1().CustomResourceDefinitions(), crds...); err != nil {
+			klog.Errorf("failed to bootstrap CRDs: %v", err)
+			return false, nil // keep retrying
+		}
+		return true, nil
+	}); err != nil {
+		return fmt.Errorf("failed to bootstrap CRDs into root workspace: %w", err)
```
ncdc, line 83: Would we ever use this outside of root?

config/helpers/bootstrap.go
```diff
@@ -0,0 +1,193 @@
...
+
+// CreateResourceFromFS creates given resource file.
+func CreateResourceFromFS(ctx context.Context, client dynamic.Interface, mapper meta.RESTMapper, filename string, fs embed.FS, transformers ...TransformFileFunc) error {
+	raw, err := fs.ReadFile(filename)
+	if err != nil {
+		return fmt.Errorf("could not read %s: %w", filename, err)
+	}
+
+	if len(raw) == 0 {
+		return nil // ignore dummy files
```
ncdc, line 128:
```suggestion
		return nil // ignore empty files
```

config/helpers/bootstrap.go
```diff
@@ -0,0 +1,193 @@
...
+		if err := configcrds.Create(ctx, crdClient.ApiextensionsV1().CustomResourceDefinitions(), crds...); err != nil {
+			klog.Errorf("failed to bootstrap CRDs: %v", err)
+			return false, nil // keep retrying
+		}
+		return true, nil
+	}); err != nil {
+		return fmt.Errorf("failed to bootstrap CRDs into root workspace: %w", err)
+	}
+
+	// bootstrap non-crd resources
```
ncdc, line 86: What's this for?

ncdc (changes requested):

config/helpers/bootstrap.go
sttts: yes, we would

config/helpers/bootstrap.go
sttts: I guess this is for line 95, ignoring the error

config/helpers/bootstrap.go
sttts: E.g. workspace types, or any other kind of resources we need like roles/rolebindings.

ncdc: Ok, so mostly forward-looking at this point?

sttts:
All adressed.

@ncdc CVO = OpenShift's cluster version operator

ncdc:
LGTM. Will wait for @marun as well, since you've been looking at this also.

marun (approved):

config/helpers/bootstrap.go
sttts: yes

PR #546, opened 2022-02-22
WIP e2e: Enable fake pclusters

marun:
**Note** This PR is not reviewable until #502 merges.


marun:
I assumed that it would be necessary to disable namespace syncing, but so long as a workspace lacks workload cluster resources namespace scheduling can't proceed. The namespace scheduler will set an error condition on the namespace, but that can likely be ignored for the purposes of testing. Given that there extensive syncer testing will need to be performed against clusters with potentially interfering controllers, this is probably fine.

PR #555, opened 2022-02-23
server: align with upstream logic to derive external address

sttts:
We had our own logic to derive an external name from binding host+port somehow. As the cert is created with the external address data, diverging asks for trouble. This PR uses the same `DefaultExternalAdddress()" us in upstream for the cert.

### DIFF
--- a/pkg/server/options/authentication.go
+++ b/pkg/server/options/authentication.go
@@ -107,16 +107,9 @@ func (s *AdminAuthentication) ApplyTo(config *genericapiserver.Config) (newToken
 	return newTokenOrEmpty, tokenHash, nil
 }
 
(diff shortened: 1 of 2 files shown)


ncdc:
@sttts with this change, if you do `--bind-address 127.0.0.1`, we see
```
I0223 15:47:47.578318    5919 secure_serving.go:266] Serving securely on 127.0.0.1:6443
```

but `admin.kubeconfig` has

```
apiVersion: v1
clusters:
- cluster:
    server: https://192.168.4.118:6443
```

ncdc:
Apparently this code path makes it so loopback is explicitly disallowed as the external address. Do we want/need our kubeconfig to contain localhost for the server url?

PR #559, opened 2022-02-23
Refactor namespace scheduler cluster assignment for testability

marun:
This PR extracts cluster selection from the namespace scheduling  controller and adds preliminary unit test coverage of the result.

API interaction is limited to functions that can be replaced in tests. Care was taken to ensure functional behavior which is easier to reason about and validate. Testing of API mutation is out-of-scope since the fixture cost would likely exceed the benefit.

This strategy of factoring out logic that can be unit tested enables trivial validation of corner-cases and is intended to reduce the cost of ongoing maintenance.

### DIFF
--- a/pkg/reconciler/namespace/namespace.go
+++ b/pkg/reconciler/namespace/namespace.go
@@ -19,12 +19,10 @@ package namespace
 import (
 	"context"
 	"fmt"
-	"math/rand"
 	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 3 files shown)


pkg/reconciler/namespace/namespace.go
```diff
@@ -187,111 +185,31 @@ func (c *Controller) updateScheduledCondition(ctx context.Context, ns *corev1.Na
...
-func (c *Controller) isValidCluster(ctx context.Context, lclusterName, clusterName string) (bool, error) {
-	cluster, err := c.clusterLister.Get(clusters.ToClusterAwareKey(lclusterName, clusterName))
-	if apierrors.IsNotFound(err) {
-		klog.Infof("Cluster %q|%q does not exist", lclusterName, clusterName)
-		return false, nil
+	scheduler := namespaceScheduler{
+		getCluster:   c.clusterLister.Get,
+		listClusters: c.clusterLister.List,
 	}
+	newClusterName, err := scheduler.AssignCluster(ns)
```
sttts, line 199: newPClusterName

marun: Maybe that kind of rename can wait for a separate PR?

pkg/reconciler/namespace/scheduler.go
```diff
@@ -0,0 +1,139 @@
...
+// the purposes of any namespace already scheduled to it (i.e., if it reports
+// as Ready, and any evictAfter value, if specified, has not yet passed).
+//
+// It doesn't take into account Unschedulable, and should only be used when
+// determining if a cluster that a namespace has already been assigned to
+// should keep having that namespace.
+func (s *namespaceScheduler) isValidCluster(lclusterName, clusterName string) (bool, error) {
+	cluster, err := s.getCluster(clusters.ToClusterAwareKey(lclusterName, clusterName))
+	if apierrors.IsNotFound(err) {
+		klog.Infof("Cluster %q|%q does not exist", lclusterName, clusterName)
```
sttts, line 82: nit: logs in predicates or getters have smell

sttts: make the func return `bool, message string, error`

marun: done

marun:
Rebased

pkg/reconciler/namespace/namespace.go
```diff
@@ -187,111 +185,32 @@ func (c *Controller) updateScheduledCondition(ctx context.Context, ns *corev1.Na
...
+	if err != nil {
+		return err
 	}
 
-	return nil
+	if oldPClusterName == newPClusterName {
+		return nil
+	}
+
+	klog.Infof("Patching to update cluster assignment for namespace %q|%q: %q -> %q",
```
imjasonh, line 208: Discussion in Slack sounded like it was heading toward `%s` over `%q`

pkg/reconciler/namespace/scheduler_test.go
```diff
@@ -0,0 +1,256 @@
...
+	testCases := map[string]struct {
+		cluster *clusterFixture
+		isValid bool
+	}{
+		"missing -> false": {},
+		"unready -> false": {
+			cluster: defaultClusterFixture(),
+		},
+		"ready -> true": {
+			cluster: defaultClusterFixture().withReady(),
```
imjasonh, line 164: (not blocking, style preference): It seems like quite a lot of boilerplate to manually write/review/maintain above just so you can use this builder pattern to make the test case slightly shorter.

I tend to prefer just being verbose and putting the whole `v1alpha1.Cluster` struct here. Even though it might be somewhat noisy, it means I only have to look in one place to see what `withReady` means, and I don't have to maintain a separate testing builder mini-framework. I've seen these things grow to include dozens of fields, and it's really toilsome to strip back out.

pkg/reconciler/namespace/namespace.go
marun: This isn't net-new code - it's a refactor - and I'd much prefer to defer rewriting log messages to a separate non-functional PR.

pkg/reconciler/namespace/scheduler_test.go
marun: Shorter isn't the goal here. It's making my intent clear. For simple self-contained unit tests like this one, I'd much rather have to maintain some simple helpers to make it crystal clear what the cluster represents vs requiring a maintainer to have to confirm that the cluster matches expectation for each test case. 

As far as the potential for the helpers to metastasize out of proportion to their value, I think we can cross to that bridge when we come to it. If we keep our tests clean and focused I don't think the scope creep you mention is likely. If we start testing too many things at once, then sure, but I think that's a factoring problem in and of itself.

marun: That said, that's just me. If consensus is a preference not to use this pattern in testing, happy to revert.

ncdc: I am a fan of helpers, as long as there's not really anything "hidden" in them (in other words, they tend to do zero to minimal logic and just set fields based on function parameters).

imjasonh: Sounds good to me.

imjasonh (approved):

pkg/reconciler/namespace/scheduler_test.go
marun: @imjasonh Please continue to keep me honest! I think your instinct to keep it simple is a good one, and I am generally in that camp too. If/when I stray I'll appreciate hearing about it from you.

marun:
Rebased (required by label rename PR). Also added a commit to revert an accidental name change merged with the rename PR.

pkg/reconciler/namespace/namespace.go
marun: https://github.com/kcp-dev/kcp/pull/600 does the rename.

imjasonh (approved):

ncdc:
@marun I'm good if @imjasonh's good. Merge when ready

Issue #550, opened 2022-02-22
Improve ingress dev/demo experience on Mac


sttts:
The prototype 2 demo does not work completely on Macs today because networking is more complicated through the podman or docker virtual machine in between kind and the host.

This issue is about improving the situation, either in documentation or even better in some automation, in demo scripting or in the ingress controller.

PR #561, opened 2022-02-24
Fix the label used to select Ingresses for Envoy

astefanutti:
That PR fixes the label used to select Ingresses ready to be exposed via Envoy, from `ingress-controller/envoy` to `ingress.kcp.dev/envoy`.

I faced this issue while working on #550, after a rebase on #478.

/cc @jmprusi

### DIFF
--- a/pkg/localenvoy/controllers/ingress/envoyingress.go
+++ b/pkg/localenvoy/controllers/ingress/envoyingress.go
@@ -25,11 +25,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
 	"k8s.io/klog/v2"
+
+	envoycontrolplane "github.com/kcp-dev/kcp/pkg/localenvoy/controlplane"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/localenvoy/controllers/ingress/envoyingress.go
```diff
@@ -90,7 +91,7 @@ func generateStatusHost(domain string, ingress *networkingv1.Ingress) string {
 		}
 	}
 
-	//TODO(jmprusi): Hardcoded to the first one...
+	// TODO(jmprusi): Hardcoded to the first one...
```
ncdc, line 94: nit: not really needed in this PR 😄 If you want to keep the diff scoped as small as possible, I'd revert this & the similar changes below... are they coming from code formatting?

ncdc (approved):

pkg/localenvoy/controllers/ingress/envoyingress.go
astefanutti: Ah yes, that my IDE that auto-formatted the file on save. I'll be careful next time.

PR #554, opened 2022-02-22
Update README for prototype2

ncdc:
Signed-off-by: Andy Goldstein <andy.goldstein@redhat.com>

Co-authored-by: Dr. Stefan Schimanski <stefan.schimanski@gmail.com>

PR #569, opened 2022-02-24
Add CONTRIBUTING.md

ncdc:
TODO:
- [x] Link to this from README.md after #554 merges

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,82 @@
+# Contributing to kcp
+
+Please read the following guide if you're interested in contributing to kcp.
+
+## Getting started
+
+### Prerequisites
+
+1. Clone this repository.
+2. [Install Go](https://golang.org/doc/install) (1.17+).
+3. Install [kubectl](https://kubernetes.io/docs/tasks/tools/#kubectl).
+
+### Build & verify
+1. In one terminal, build and start `kcp`:
+```
+go run ./cmd/kcp start
+```
+
+2. In another terminal, tell `kubectl` where to find the kubeconfig:
+
+```
+export KUBECONFIG=.kcp/admin.kubeconfig
+```
+
+3. Confirm you can connect to `kcp`:
+
+```
+kubectl api-resources
+```
+
+## Finding areas to contribute
+
+Starting to participate in a new project can sometimes be overwhelming, and you may not know where to begin. Fortunately, we are here to help! We track all of our tasks here in GitHub, and we label our issues to categorize them. Here are a couple of handy links to check out:
+
+* [Good first issue](https://github.com/kcp-dev/kcp/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) issues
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,82 @@
...
+    - `rootKcpClient`
+    - `orgKubeClient`
+- Configs intended for `NewClusterForConfig` (i.e. today often called "admin workspace config") should uniformely be called `clusterConfig`
+    - Note: with org workspaces, the second use for the admin workspaces will go away
+    - Note 2: sometimes we use clients for same purpose, but this can be harder to read
+- Cluster-aware clients should follow similar naming conventions:
+    - `crdClusterClient`
+    - `kcpClusterClient`
+    - `kubeClusterClient`
+- `clusterName` is a kcp term. It is **NOT** a name of a physical cluster. If we mean the latter, use `pclusterName` or similar.
```
marun, line 77: (No action required) Why is it important for kcp's `clusterName` to be unqualified? vs using `lclusterName` for a logical cluster and `pclusterName` for physical clusters?

CONTRIBUTING.md
```diff
@@ -0,0 +1,82 @@
...
+- In the syncer: upstream = kcp, downstream = pcluster. Depending on direction, "from" and "to" can have different meanings. `source` and `sink` are synonyms for upstream and downstream.
```
marun, line 78: (No action required) source/sink tends to be used to refer to event producers and consumers. Maybe source/destination would be more appropriate for our kcp?

marun (commented):
LGTM, thank you for getting the ball rolling!

CONTRIBUTING.md
ncdc: @sttts wrote these, but I'll hazard a guess: the guidance is to never use `clusterName` to mean a physical cluster. If/when unqualified, `clusterName` means a logic cluster.

I maybe want to also suggest that we never use `lcluster...` or `pcluster...` - it's too easy to wonder if that was intended, or if the leading `l`/`p` was a typo.

CONTRIBUTING.md
ncdc: My preference would be target instead of sink. What do y'all think?

CONTRIBUTING.md
sttts: `clusterName` is defined by kube's ObjectMeta. Also kcp is technically about clusters in the lcluster sense. Pcluster is something on-top. We are just not that far yet to split it appart and have the syncer in the same code base.

CONTRIBUTING.md
sttts: target of status syncer is kcp. Don't like that.

ncdc: Fair

ncdc: Then I'd vote for src/dest

sttts: or just kcp and pcluster.

sttts (approved):

PR #576, opened 2022-02-25
README clarifications

ncdc:
Incorporating comments from @smarterclayton in #554

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
-# `kcp` provides a Kubernetes-like control plane with true multitenancy, with flexible compute powered by real Kubernetes clusters.
+# `kcp` provides a true multi-tenant Kubernetes control plane for workloads on many clusters
 
 ![build status badge](https://github.com/kcp-dev/kcp/actions/workflows/ci.yaml/badge.svg)
 
-`kcp` is a generic [CustomResourceDefinition](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) (CRD) apiserver that is divided into multiple "[logical clusters](docs/investigations/logical-clusters.md)" that enable multitenancy of cluster-scoped resources such as CRDs and Namespaces. Each logical cluster is independent: the available APIs and data are separate from one logical cluster to another.
+`kcp` is a generic [CustomResourceDefinition](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) (CRD) apiserver that is divided into multiple "[logical clusters](docs/investigations/logical-clusters.md)" that enable multitenancy of cluster-scoped resources such as CRDs and Namespaces. Each of these logical clusters is fully isolated from the others, allowing different teams, workloads, and use cases to live side by side.
 
 By default, `kcp` only knows about:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,7 +14,7 @@ By default, `kcp` only knows about:
 
 ![kubectl api-resources showing kcp's API resources](./docs/images/kubectl-api-resources.png)
 
-Any other resources, including standard Kubernetes resources like [`Deployment`](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/)s and the rest, can be added as CRDs and optionally reconciled using the standard controllers launched against the `kcp` API, or via two-way replication into a Kubernetes cluster.
+Any other resources, including standard Kubernetes resources like [`Deployment`](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/)s and the rest, can be added as CRDs and and be replicated onto one or more Kubernetes clusters.
```
marun, line 17: nit: s/and and/and/

README.md
```diff
@@ -27,6 +27,10 @@ At the same time, a diverse and creative community of tools and services has spr
 
 Imagine a declarative Kubernetes-style API for _anything_, supported by an ecosystem of Kubernetes-aware tooling, separate from Kubernetes-the-container-orchestrator.
 
+But a control plane without any workloads isn't very compelling - so we want to uplevel the majority of Kubernetes applications today by allowing them to move and spread across clusters without changing a line of YAML. We believe moving a workload between clusters should be as easy as moving a pod between nodes.
```
marun, line 30: nit: s/ - /, / Or maybe just start a new sentance?

README.md
```diff
@@ -27,6 +27,10 @@ At the same time, a diverse and creative community of tools and services has spr
...
+
+Take a multi-tenant control plane for everything, and add to it most of the world's cloud-native workloads happily cruising across clouds, datacenters, and out to the edge?
```
marun, line 32: nit: Maybe use exclamation instead of question mark for this sentence? Or maybe add a following sentence emphasizing how awesome the preceding vision is?

README.md
ncdc: Updated

README.md
smarterclayton: Actually, an ellipsis would probably be the best continuation.

ncdc: Done

marun:
LGTM

README.md
```diff
@@ -27,6 +27,10 @@ At the same time, a diverse and creative community of tools and services has spr
 
 Imagine a declarative Kubernetes-style API for _anything_, supported by an ecosystem of Kubernetes-aware tooling, separate from Kubernetes-the-container-orchestrator.
 
+But a control plane without any workloads isn't very compelling... We want to uplevel the majority of Kubernetes applications today by allowing them to move and spread across clusters without changing a line of YAML. We believe moving a workload between clusters should be as easy as moving a pod between nodes.
```
smarterclayton, line 30: I would remove the ellipsis here and tweak a bit (after re-reading this about ten times:

> But even a generic control plane is only as good as the use case it enables.  We want to give existing Kubernetes users a new superpower - take your existing applications and allow them to move or spread across clusters without having to change a single line of YAML. We believe moving a workload between clusters should be as easy as moving a pod between nodes.

README.md
```diff
@@ -1,8 +1,8 @@
-# `kcp` provides a Kubernetes-like control plane with true multitenancy, with flexible compute powered by real Kubernetes clusters.
+# `kcp` provides a true multi-tenant Kubernetes control plane for workloads that span clusters
```
sttts, line 1: "that span cluster" sounds wrong for me. That wording suggests federation as the focus which it isn't.

smarterclayton: I definitely wanted it to be vague here - because we are not implying M:1, nor 1:N, but M:N.  And we're starting with M:N, not M -> 2.  If you have a word that is vague like spans, please suggest.

smarterclayton: > ... for workloads that fill clusters

> ... for workloads across clusters

> ... for workloads on many clusters

ncdc: I went with option 3

ncdc:
Updated, PTAL

sttts (approved):

PR #582, opened 2022-02-28
Add workspace invariant of valid baseURL

sttts:
This PR does the following:
1. To simplify the state machine as seen by users and controllers, this PR adds the invariant that

   - After scheduling `status.baseURL` of a ClusterWorkspace stays valid.

   In other words: there is no unscheduling. A workspace has state (obviously) and a workspace without a shard does not make sense.

   Corrolary: We have to make sure (through movement and high availability measures) that a workspace is available. Clients should not be expected to hande the case that a workspace is not accessible. Reason:
   - client-go clients don't watch the workspace object.
   - processes talk to some URL via clients, and clients cannot update the URL they talk to (that's not part of the Kubernetes client contract).

   tl/dr: we have to make sure the workspace baseURL keeps working at all time.
2. A `ClusterWorkspace` admission plugin is added verifying the state transitions of cluster workspace objects.

### DIFF
--- a/pkg/admission/clusterworkspace/admission.go
+++ b/pkg/admission/clusterworkspace/admission.go
@@ -1,0 +1,129 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package clusterworkspace
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+
+	"k8s.io/apimachinery/pkg/api/validation"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apiserver/pkg/admission"
+
+	kcpadmissionhelpers "github.com/kcp-dev/kcp/pkg/admission/helpers"
+	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
+)
+
+// Validate ClusterWorkspace creation and updates for
+// - immutability of fields like type
+// - valid phase transitions fulfilling pre-conditions
+// - status.location.current and status.baseURL cannot be unset.
+
+const (
+	PluginName = "tenancy.kcp.dev/ClusterWorkspace"
+)
+
+func Register(plugins *admission.Plugins) {
+	plugins.Register(PluginName,
+		func(_ io.Reader) (admission.Interface, error) {
+			return &clusterWorkspace{
+				Handler: admission.NewHandler(admission.Create, admission.Update),
+			}, nil
+		})
+}
+
+type clusterWorkspace struct {
+	*admission.Handler
+}
+
+// Ensure that the required admission interfaces are implemented.
+var _ = admission.ValidationInterface(&clusterWorkspace{})
+
+var phaseOrdinal = map[tenancyv1alpha1.ClusterWorkspacePhaseType]int{
+	tenancyv1alpha1.ClusterWorkspacePhaseType(""):     1,
+	tenancyv1alpha1.ClusterWorkspacePhaseScheduling:   2,
+	tenancyv1alpha1.ClusterWorkspacePhaseInitializing: 3,
+	tenancyv1alpha1.ClusterWorkspacePhaseReady:        4,
+}
+
+// Validate ensures that
+// - the workspace only does a valid phase transition
+// - has a valid type
+// - has valid initializers when transitioning to initializing
+func (o *clusterWorkspace) Validate(ctx context.Context, a admission.Attributes, _ admission.ObjectInterfaces) (err error) {
+	if a.GetResource().GroupResource() != tenancyv1alpha1.Resource("clusterworkspaces") {
+		return nil
+	}
+
+	obj, err := kcpadmissionhelpers.NativeObject(a.GetObject())
+	if err != nil {
+		// nolint: nilerr
+		return nil // only work on unstructured ClusterWorkspaces
+	}
+	cw, ok := obj.(*tenancyv1alpha1.ClusterWorkspace)
+	if !ok {
+		// nolint: nilerr
+		return nil // only work on unstructured ClusterWorkspaces
+	}
+
+	if a.GetOperation() == admission.Update {
+		obj, err = kcpadmissionhelpers.NativeObject(a.GetOldObject())
+		if err != nil {
+			return fmt.Errorf("unexpected unknown old object, got %v, expected ClusterWorkspace", a.GetOldObject().GetObjectKind().GroupVersionKind().Kind)
+		}
+		old, ok := obj.(*tenancyv1alpha1.ClusterWorkspace)
+		if !ok {
+			return fmt.Errorf("unexpected unknown old object, got %v, expected ClusterWorkspace", obj.GetObjectKind().GroupVersionKind().Kind)
+		}
+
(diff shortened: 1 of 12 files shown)


pkg/reconciler/workspace/controller.go
```diff
@@ -286,39 +286,67 @@ func (c *Controller) process(ctx context.Context, key string) error {
...
+		if current := workspace.Status.Location.Current; current != "" {
+			// make sure current shard still exists
+			if _, err := c.rootWorkspaceShardLister.Get(clusters.ToClusterAwareKey(tenancyhelper.RootCluster, current)); errors.IsNotFound(err) {
+				klog.Infof("De-scheduling workspace %q|%q from nonexistent shard %q", tenancyhelper.RootCluster, workspace.Name, current)
+				workspace.Status.Location.Current = ""
+			} else if err != nil {
+				return err
+			}
 		}
-		shard = currentShard
```
stevekuznetsov, line 299: Why did we lose this?

test/e2e/reconciler/workspace/controller_test.go
```diff
@@ -331,9 +340,11 @@ func unschedulable(object *tenancyv1alpha1.ClusterWorkspace) error {
 func scheduled(target string) func(workspace *tenancyv1alpha1.ClusterWorkspace) error {
 	return func(object *tenancyv1alpha1.ClusterWorkspace) error {
 		if isUnschedulable(object) {
+			klog.Infof("workspace %s is unschedulable, skipping scheduled check", object.Name)
```
stevekuznetsov, line 343: Why do we need the klogs, should the errors not be logged by `t.Error`?

pkg/reconciler/workspace/controller.go
stevekuznetsov: Flow seems correct but is a bit more implicit this way.

stevekuznetsov (approved):
Seems fine to me - how common is the admission pattern?

pkg/reconciler/workspace/controller.go
sttts: are you referring to line 375 where we reget the shard?

test/e2e/reconciler/workspace/controller_test.go
sttts: Just to have some indication when things go wrong. Silent wait loops are usually silent despite the logs when things are good, but should be noisy when things are bad.

pkg/reconciler/workspace/controller.go
sttts: updated. Actually, we only want to touch the baseURL during scheduling. After that it is set in stone forever.

pkg/admission/clusterworkspace/admission.go
```diff
@@ -0,0 +1,124 @@
...
+		if err != nil {
+			return fmt.Errorf("unexpected unknown old object, got %v, expected ClusterWorkspace", a.GetOldObject().GetObjectKind().GroupVersionKind().Kind)
+		}
+		old, ok := obj.(*tenancyv1alpha1.ClusterWorkspace)
+		if !ok {
+			return fmt.Errorf("unexpected unknown old object, got %v, expected ClusterWorkspace", obj.GetObjectKind().GroupVersionKind().Kind)
+		}
+
+		if old.Spec.Type != cw.Spec.Type {
+			return admission.NewForbidden(a, errors.New("spec.type is immutable"))
```
ncdc, line 94: Do you want to use any validation helpers, such as ValidateImmutableField?

pkg/admission/clusterworkspace/admission_test.go
```diff
@@ -0,0 +1,340 @@
...
+						Phase:        tenancyv1alpha1.ClusterWorkspacePhaseReady,
+						Initializers: []tenancyv1alpha1.ClusterWorkspaceInitializer{},
+						Location:     tenancyv1alpha1.ClusterWorkspaceLocation{Current: "somewhere"},
+						BaseURL:      "https://kcp.bigcorp.com/clusters/org:test",
+					},
+				}),
+			wantErr: true,
+		},
+		{
+			name: "ignors different resources",
```
ncdc, line 309:
```suggestion
			name: "ignores different resources",
```

pkg/admission/clusterworkspacetypeexists/admission_test.go
```diff
@@ -454,86 +427,8 @@ func TestValidate(t *testing.T) {
...
-						Phase:        tenancyv1alpha1.ClusterWorkspacePhaseInitializing,
-						Initializers: []tenancyv1alpha1.ClusterWorkspaceInitializer{"a"},
-						Location:     tenancyv1alpha1.ClusterWorkspaceLocation{Current: "somewhere"},
-						BaseURL:      "https://kcp.bigcorp.com/clusters/org:test",
-					},
-				}),
-		},
-		{
-			name: "ignors different resources",
+			name:  "ignors different resources",
```
ncdc, line 430:
```suggestion
			name:  "ignores different resources",
```

pkg/reconciler/workspace/controller.go
```diff
@@ -286,39 +288,105 @@ func (c *Controller) process(ctx context.Context, key string) error {
...
+				}
+				u.Path = path.Join(u.Path, targetShard.Status.ConnectionInfo.APIPath, "clusters", logicalCluster)
+
+				workspace.Status.BaseURL = u.String()
+				workspace.Status.Location.Current = targetShard.Name
+
+				conditions.MarkTrue(workspace, tenancyv1alpha1.WorkspaceScheduled)
+				klog.Infof("Scheduled workspace %q|%q to %q|%q", workspace.ClusterName, workspace.Name, targetShard.ClusterName, targetShard.Name)
+			} else {
+				conditions.MarkFalse(workspace, tenancyv1alpha1.WorkspaceScheduled, tenancyv1alpha1.WorkspaceReasonUnschedulable, conditionsv1alpha1.ConditionSeverityError, "No available shards to schedule workspace %q|%q.", workspace.ClusterName, workspace.Name)
```
ncdc, line 356: Since this is a condition on the workspace itself, can we omit the ClusterName and Name from the message?

pkg/reconciler/workspace/controller.go
```diff
@@ -286,39 +288,105 @@ func (c *Controller) process(ctx context.Context, key string) error {
...
+				failures := make([]string, 0, len(invalidShards))
+				for name, x := range invalidShards {
+					failures = append(failures, fmt.Sprintf("  %s: reason %q, message %q", name, x.reason, x.message))
+				}
+				klog.Infof("No valid shards found for workspace %q|%q, skipped:\n%s", workspace.ClusterName, workspace.Name, strings.Join(failures, "\n"))
```
ncdc, line 361: Ditto - can we omit workspace clusterName and name from the message?

pkg/admission/clusterworkspace/admission.go
sttts: oh, didn't know those exists

pkg/reconciler/workspace/controller.go
sttts: this is for logging, not exposed

sttts:
@ncdc Addressed comments.

test/e2e/virtual/workspaces/virtual_workspace_test.go
```diff
@@ -182,8 +182,8 @@ func TestWorkspacesVirtualWorkspaces(t *testing.T) {
 							return false, nil
 						}
 						return false, err
-					} else if !conditions.IsTrue(cw, tenancyv1alpha1.WorkspaceURLValid) {
-						return false, fmt.Errorf("ClusterWorkspace %s is not valid: %s", cw.Name, conditions.GetMessage(cw, tenancyv1alpha1.WorkspaceURLValid))
+					} else if !conditions.IsTrue(cw, tenancyv1alpha1.WorkspaceShardValid) {
+						return false, fmt.Errorf("ClusterWorkspace %s is not valid: %s", cw.Name, conditions.GetMessage(cw, tenancyv1alpha1.WorkspaceShardValid))
```
aojea, line 186: @sttts why is this a hard error, should not retry?
seens a failure here because of this https://github.com/kcp-dev/kcp/runs/5467154069?check_suite_focus=true

PR #584, opened 2022-02-28
prototype2-script: the Ingress demo can run on MacOS

### LINKED ISSUE #550: Improve ingress dev/demo experience on Mac

astefanutti:
This PR moves the Envoy proxy, used in the Ingress part of the P2 demo, to be running in container, so that it can connect to the network created by KinD, to be able to proxy downstream clusters, even while it runs in the extra virtualisation layer introduced by Docker or Podman on MacOS.

Fixes #550.

### DIFF
--- a/contrib/demo/clusters/kind/createKindClusters.sh
+++ b/contrib/demo/clusters/kind/createKindClusters.sh
@@ -56,13 +56,20 @@ detect_podman() {
         return
     fi
 
+    if [[ "$OSTYPE" == "darwin"* && -z "$(podman ps)" ]]; then
+        # Podman machine is not started
+        return
+    fi
+
     if [[ -z "$(podman system connection ls --format=json)" ]]; then
         return
     fi
 
+    KIND_EXPERIMENTAL_PROVIDER=podman
+
     PODMAN_VERSION=$(podman version -f '{{.Server.Version}}')
     PODMAN_MAJOR=$(echo "${PODMAN_VERSION}" | cut -d. -f1)
-    if [[ "${PODMAN_MAJOR}" == "3" ]]; then
+    if [[ "${PODMAN_MAJOR}" == "3" && "$OSTYPE" == "darwin"* ]]; then
         PODMAN_MAC_SSH_WORKAROUND=1
 
         # Setup the control socket if it's not already running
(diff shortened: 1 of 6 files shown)


astefanutti:
@ncdc @jmprusi this implements the approach we discussed. I've successfully tested it with Docker Desktop for Mac, as well as with Podman 3.

sttts:
@jmprusi can you review?

contrib/envoy/bootstrap.yaml
```diff
@@ -27,7 +27,7 @@ static_resources:
             endpoint:
               address:
                 socket_address:
-                  address: "localhost"
+                  address: BOOTSTRAP_ADDRESS
```
jmprusi, line 30: This file could be used by others, so instead of modifying it, can you copy it and add an extension that makes it look like a "templated file" (so, it's explicit that is not intended to be used directly by envoy) something like `bootstrap.yaml.tmpl` or so.

contrib/demo/clusters/kind/us-east1.config
```diff
@@ -12,10 +12,3 @@ nodes:
     nodeRegistration:
       kubeletExtraArgs:
         node-labels: "ingress-ready=true"
-  extraPortMappings:
```
jmprusi, line 15: no need to remove those, as it could be useful to go directly against the cluster ingress in some cases.

contrib/demo/clusters/kind/createKindClusters.sh
```diff
@@ -56,13 +56,20 @@ detect_podman() {
         return
     fi
 
+    if [[ "$OSTYPE" == "darwin"* && -z "$(podman ps)" ]]; then
+        # Podman machine is not started
```
jmprusi, line 60: can you add some error message here?

jmprusi (changes requested):
Thanks @astefanutti ! Can you take a look at the comments? I think we can merge it after that!

contrib/demo/clusters/kind/createKindClusters.sh
astefanutti: Good idea, added in 763f62bd90e8643cdce4e0d618a5f5ca584fc6d9.

contrib/envoy/bootstrap.yaml
astefanutti: Sure, added in 0e7af08b3b7783824fb7f6c7bb25f553d22b36dc.

astefanutti:
@jmprusi thanks a lot for the review. I think I've covered your comments. I've also added small fixes for:
* `|&` fails with GNU bash version 3.2.57 on MacOS
* Unbounded `IMAGE_FLAG` variable
* Duplicated calls to the Envoy container clean-up function

sttts:
Tested locally.

jmprusi (approved):

astefanutti:
@sttts @jmprusi I have re-pushed to make the tests pass (otherwise retries kept failing). Could you re-enable auto-merge once again so that it gets merged?

PR #589, opened 2022-03-01
ci: don't run e2e & other tests for documentation-only PRs

### LINKED ISSUE #573: CI: don't run e2e & other tests for documentation-only PRs
If a PR only touches documentation, we should skip running most of the CI jobs because the documentation changes do not affect e.g. e2e tests.

We can use https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths to accomplish this.

Possible ignore list:
- `docs/**`
- `*.md` (not sure if we want `**/*.md` for every md file, or only those in the repo root)
- `LICENSE`

I'm not sure if we currently have any jobs that we'd want to run for a docs-only change. WDYT?

avinal:
# What is changed?
    
 - Exclude docs and markdowns
 - Reformat workflow files
 - Fixes #573

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,6 +5,9 @@ on:
     branches: ["main"]
   pull_request:
     branches: ["main"]
+    paths-ignore:
+      - "docs/**"
+      - "**/*.md"
 
 jobs:
   boilerplate:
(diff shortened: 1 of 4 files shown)


avinal:
Hey @ncdc,

I think `LICENSE` need not be excluded since it is not meant to be changed (in normal circumstances), also I have formatted the workflow files, doesn't make any difference but just nice to have a standard. Please review.

ncdc:
Thanks @avinal!

imjasonh (approved):

PR #592, opened 2022-03-01
Refactor namespace scheduler status maintenance for testability

marun:


pkg/reconciler/namespace/conditions_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	conditionsapi "github.com/kcp-dev/kcp/third_party/conditions/apis/conditions/v1alpha1"
+	"github.com/kcp-dev/kcp/third_party/conditions/util/conditions"
+)
+
+func TestSetScheduledCondition(t *testing.T) {
+	testCases := map[string]struct {
```
imjasonh, line 32: (another fun non-blocking style nit comment!) I never really see this phrased in terms of a `map[string]struct`, usually just a `[]struct` with some description field. I suppose it's roughly equivalent, except that execution order will be randomized using a map. I don't think that would be a problem in this case, but I think that's probably why the idiomatic way is a `[]struct` (and maybe saving some indentation?).

test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -220,21 +219,6 @@ func TestNamespaceScheduler(t *testing.T) {
 
 				err = server.expect(namespace, unscheduledMatcher(nscontroller.NamespaceReasonSchedulingDisabled))
 				require.NoError(t, err, "did not see namespace marked with scheduling disabled")
-
-				t.Log("Assign a cluster to the namespace manually")
```
imjasonh, line 224: Was this purposefully removed?

pkg/reconciler/namespace/conditions_test.go
marun: I've gotten into the habit of using maps by default to avoid any dependency on test ordering. Tests relying on order of execution are by definition not hermetic, which is generally an anti-pattern.

I think you'll find there's effectively no difference in indentation from struct to map:


```golang
testCases := []struct{
   name string
   bar     string
} {
    {
        name: "foo",
        bar:     "baz",
    },
}

testCases := map[string]struct{
    foo string
} {
    "foo": {
        bar: "baz",
    },
}
```

test/e2e/reconciler/namespace/controller_test.go
marun: Yes, is intentional. The removed part of this test was checking that a namespace for which scheduling is disabled and a cluster is assigned has a condition of scheduled=true (implying automatic scheduling). As per the unit tests, the desired condition should be scheduled=false with reason 'SchedulingDisabled' (implying assignment is handled manually). Since we already have unit test coverage for this I chose to remove the broken test rather than rewrite.

marun: Or should any assigned cluster - regardless of whether automatically assigned - imply a condition of scheduled=true?  Definitely worth some thought as to how we want this to behave from a UX perspective.

pkg/reconciler/namespace/conditions_test.go
imjasonh: > I think you'll find there's effectively no difference in indentation from struct to map:

True enough, unless you collapse the indentation (like I tend to do), which is a bit harder to justify with a map:

```go
testCases := []struct{
    name string
    bar  string
} {{
    name: "foo",
    bar:  "baz",
}}
 
testCases := map[string]struct{
    foo string
} {
    "foo": {
        bar: "baz",
    },
}
```

Anyway, I've spent my per-PR budget on style nits. Resolving.

imjasonh (approved):

pkg/reconciler/namespace/conditions_test.go
marun: This is the first time I'm seeing the indentation being collapsed, now I see what you mean!

imjasonh: For bonus compactness points, I usually do this:

```golang
for _, c := range []struct {
  desc string
  ...
  want string
} {{
  desc: "first",
  want: "whatever",
}, {
  desc: "second",
  want: "another thing",
}} {
  t.Run(c.desc, func(t *testing.T) {
    ...
  }
}

```

marun: I think I'm convinced to use this compact form for unit tests that have no potential interaction. Maybe for e2e where shared fixture is involved, maps would be worth the loss of compactness.

PR #599, opened 2022-03-01
kube: back to upstream (dd0f1685f84)

ncdc:
Looks like #508 snuck in replace directives pointing at @sttts's kube fork 😄


go.mod
```diff
@@ -32,36 +32,36 @@ require (
 	k8s.io/klog/v2 v2.30.0
 	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65
 	k8s.io/kubectl v0.0.0
-	k8s.io/kubernetes v0.0.0
+	k8s.io/kubernetes v1.23.4
```
ncdc, line 35: I'm seeing this also in #533 and don't know why it's not staying at v0.0.0 any more, but it shouldn't matter because of the replace directives (? right ?)

sttts (approved):

Issue #552, opened 2022-02-22
Make signal handling consistent


ncdc:
We have a mixture of:

- `signal.NotifyContext`
- `genericapiserver.SetupSignalContext`
- `genericapiserver.SetupSignalHandler`

I'd like to unify to a single convention, if everyone agrees. My preference would be `genericapiserver.SetupSignalContext`.

PR #601, opened 2022-03-02
Convert cmd/syncer/main.go to be a cobra.Command

### LINKED ISSUE #553: Convert cmd/syncer/main.go to be a cobra.Command

RAMESSESII2:
### Migrating to cobra.Command
* added cobra command equivalent of existing go flags.
* moved the main logic from ./cmd/syncer/main.go to ./cmd/syncer/cmd/syncer.go
* created a new exported ```Options struct``` for flags binding.
* syncedResourceTypes is a flag now
image

## Testing
* Build the source code
```bash
go build ./cmd/syncer
```
* Run it
```bash
./syncer -h
```
- [x] Fixes #553

### DIFF
--- a/cmd/syncer/cmd/syncer.go
+++ b/cmd/syncer/cmd/syncer.go
@@ -18,27 +18,22 @@ package cmd
 
 import (
 	"context"
-	"errors"
-	"fmt"
 
 	"github.com/spf13/cobra"
-	"github.com/spf13/pflag"
+
 	"k8s.io/apimachinery/pkg/util/sets"
 	genericapiserver "k8s.io/apiserver/pkg/server"
-	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/component-base/logs"
 	"k8s.io/klog/v2"
 
-	nscontroller "github.com/kcp-dev/kcp/pkg/reconciler/namespace"
+	synceroptions "github.com/kcp-dev/kcp/cmd/syncer/options"
 	"github.com/kcp-dev/kcp/pkg/syncer"
 )
 
 const numThreads = 2
 
 func NewSyncerCommand() *cobra.Command {
-	options := NewDefaultOptions()
-
+	options := synceroptions.NewOptions()
 	syncerCommand := &cobra.Command{
 		Use:   "syncer",
 		Short: "Synchronizes resources in `kcp` assigned to the clusters",
@@ -52,7 +47,7 @@ func NewSyncerCommand() *cobra.Command {
 			}
 
 			ctx := genericapiserver.SetupSignalContext()
-			if err := options.Run(ctx); err != nil {
+			if err := Run(options, ctx); err != nil {
 				return err
 			}
 
@@ -62,89 +57,32 @@ func NewSyncerCommand() *cobra.Command {
 		},
 	}
 
-	options.BindFlags(syncerCommand.Flags())
+	options.AddFlags(syncerCommand.Flags())
 
 	return syncerCommand
 }
 
-func (options *Options) Validate() error {
-	if options.FromClusterName == "" {
-		return errors.New("--from-cluster is required")
-	}
-
-	return nil
-}
-
-func (options *Options) Complete() error {
+func Run(options *synceroptions.Options, ctx context.Context) error {
 	klog.Infof("Syncing the following resource types: %s", options.SyncedResourceTypes)
 
-	// Create a client to dynamically watch "from".
-	if cfg, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
-		&clientcmd.ClientConfigLoadingRules{ExplicitPath: options.FromKubeconfig}, nil).ClientConfig(); err != nil {
+	kcpConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: options.FromKubeconfig}, nil).ClientConfig()
+	if err != nil {
 		return err
-	} else {
-		options.fromConfig = cfg
 	}
 
-	if options.ToKubeconfig != "" {
-		toOverrides := clientcmd.ConfigOverrides{
+	toConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: options.ToKubeconfig},
+		&clientcmd.ConfigOverrides{
 			CurrentContext: options.ToContext,
-		}
-
-		if cfg, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
-			&clientcmd.ClientConfigLoadingRules{ExplicitPath: options.ToKubeconfig},
-			&toOverrides).ClientConfig(); err != nil {
-			return err
-		} else {
-			options.toConfig = cfg
-		}
+		}).ClientConfig()
+	if err != nil {
+		return err
 	}
 
-	return nil
-}
-
-func (options *Options) Run(ctx context.Context) error {
-	if err := syncer.StartSyncer(ctx, options.fromConfig, options.toConfig, sets.NewString(options.SyncedResourceTypes...), options.FromClusterName, options.PclusterID, numThreads); err != nil {
+	if err := syncer.StartSyncer(ctx, kcpConfig, toConfig, sets.NewString(options.SyncedResourceTypes...), options.FromClusterName, options.PclusterID, numThreads); err != nil {
 		return err
 	}
 
 	return nil
-}
-
-type Options struct {
-	FromKubeconfig      string
-	FromClusterName     string
-	ToKubeconfig        string
-	ToContext           string
-	PclusterID          string
-	Logs                *logs.Options
-	SyncedResourceTypes []string
-
-	//non-exported members
-	toConfig   *rest.Config
-	fromConfig *rest.Config
-}
-
-func NewDefaultOptions() *Options {
-	return &Options{
-		FromKubeconfig:      "",
-		FromClusterName:     "",
-		ToKubeconfig:        "",
-		ToContext:           "",
-		PclusterID:          "",
-		SyncedResourceTypes: []string{"deployments.apps"},
-		Logs:                logs.NewOptions(),
-	}
-}
-
-func (options *Options) BindFlags(fs *pflag.FlagSet) {
-	fs.StringVar(&options.FromKubeconfig, "from-kubeconfig", options.FromKubeconfig, "Kubeconfig file for -from cluster.")
-	fs.StringVar(&options.FromClusterName, "from-cluster", options.FromClusterName, "Name of the -from logical cluster.")
-	fs.StringVar(&options.ToKubeconfig, "to-kubeconfig", options.ToKubeconfig, "Kubeconfig file for -to cluster. If not set, the InCluster configuration will be used.")
-	fs.StringVar(&options.ToContext, "to-context", options.ToContext, "Context to use in the Kubeconfig file for -to cluster, instead of the current context.")
-	fs.StringVar(&options.PclusterID, "cluster", options.PclusterID,
-		fmt.Sprintf("ID of the -to cluster. Resources with this ID set in the '%s' label will be synced.", nscontroller.ClusterLabel))
-	fs.StringArrayVarP(&options.SyncedResourceTypes, "sync-resources", "r", options.SyncedResourceTypes, "Resources to be synchronized in kcp.")
-
-	options.Logs.AddFlags(fs)
 }
--- a/cmd/syncer/cmd/syncer.go
+++ b/cmd/syncer/cmd/syncer.go
@@ -1,0 +1,150 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/util/sets"
+	genericapiserver "k8s.io/apiserver/pkg/server"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/component-base/logs"
+	"k8s.io/klog/v2"
+
+	nscontroller "github.com/kcp-dev/kcp/pkg/reconciler/namespace"
+	"github.com/kcp-dev/kcp/pkg/syncer"
+)
+
+const numThreads = 2
+
+func NewSyncerCommand() *cobra.Command {
+	options := NewDefaultOptions()
+
+	syncerCommand := &cobra.Command{
+		Use:   "syncer",
+		Short: "Synchronizes resources in `kcp` assigned to the clusters",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := options.Complete(); err != nil {
+				return err
+			}
+
+			if err := options.Validate(); err != nil {
+				return err
+			}
+
+			ctx := genericapiserver.SetupSignalContext()
+			if err := options.Run(ctx); err != nil {
+				return err
+			}
+
+			<-ctx.Done()
+
+			return nil
+		},
+	}
+
+	options.BindFlags(syncerCommand.Flags())
+
+	return syncerCommand
+}
+
+func (options *Options) Validate() error {
+	if options.FromClusterName == "" {
+		return errors.New("--from-cluster is required")
+	}
+
+	return nil
+}
+
+func (options *Options) Complete() error {
+	klog.Infof("Syncing the following resource types: %s", options.SyncedResourceTypes)
+
+	// Create a client to dynamically watch "from".
+	if cfg, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: options.FromKubeconfig}, nil).ClientConfig(); err != nil {
+		return err
+	} else {
+		options.fromConfig = cfg
+	}
+
+	if options.ToKubeconfig != "" {
+		toOverrides := clientcmd.ConfigOverrides{
+			CurrentContext: options.ToContext,
+		}
+
+		if cfg, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+			&clientcmd.ClientConfigLoadingRules{ExplicitPath: options.ToKubeconfig},
+			&toOverrides).ClientConfig(); err != nil {
+			return err
+		} else {
+			options.toConfig = cfg
+		}
+	}
+
+	return nil
+}
+
+func (options *Options) Run(ctx context.Context) error {
+	if err := syncer.StartSyncer(ctx, options.fromConfig, options.toConfig, sets.NewString(options.SyncedResourceTypes...), options.FromClusterName, options.PclusterID, numThreads); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+type Options struct {
+	FromKubeconfig      string
+	FromClusterName     string
+	ToKubeconfig        string
+	ToContext           string
+	PclusterID          string
+	Logs                *logs.Options
+	SyncedResourceTypes []string
+
+	//non-exported members
+	toConfig   *rest.Config
+	fromConfig *rest.Config
+}
+
+func NewDefaultOptions() *Options {
+	return &Options{
+		FromKubeconfig:      "",
+		FromClusterName:     "",
+		ToKubeconfig:        "",
+		ToContext:           "",
+		PclusterID:          "",
+		SyncedResourceTypes: []string{"deployments.apps"},
+		Logs:                logs.NewOptions(),
+	}
+}
+
+func (options *Options) BindFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&options.FromKubeconfig, "from-kubeconfig", options.FromKubeconfig, "Kubeconfig file for -from cluster.")
+	fs.StringVar(&options.FromClusterName, "from-cluster", options.FromClusterName, "Name of the -from logical cluster.")
+	fs.StringVar(&options.ToKubeconfig, "to-kubeconfig", options.ToKubeconfig, "Kubeconfig file for -to cluster. If not set, the InCluster configuration will be used.")
+	fs.StringVar(&options.ToContext, "to-context", options.ToContext, "Context to use in the Kubeconfig file for -to cluster, instead of the current context.")
+	fs.StringVar(&options.PclusterID, "cluster", options.PclusterID,
+		fmt.Sprintf("ID of the -to cluster. Resources with this ID set in the '%s' label will be synced.", nscontroller.ClusterLabel))
+	fs.StringArrayVarP(&options.SyncedResourceTypes, "sync-resources", "r", options.SyncedResourceTypes, "Resources to be synchronized in kcp.")
+
+	options.Logs.AddFlags(fs)
+}
--- a/cmd/syncer/main.go
+++ b/cmd/syncer/main.go
@@ -17,87 +17,15 @@ limitations under the License.
 package main
 
 import (
-	"context"
-	"flag"
-	"fmt"
 	"os"
-	"os/signal"
-	"path"
-	"syscall"
 
-	"k8s.io/apimachinery/pkg/util/sets"
-	"k8s.io/client-go/rest"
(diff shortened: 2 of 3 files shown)


cmd/syncer/main.go
```diff
@@ -16,88 +16,6 @@ limitations under the License.
...
-
-	klog.Infoln("Starting workers")
-	if err := syncer.StartSyncer(ctx, fromConfig, toConfig, sets.NewString(syncedResourceTypes...), *fromClusterName, *pclusterID, numThreads); err != nil {
-		klog.Fatal(err)
-	}
-
-	<-ctx.Done()
-
-	klog.Infoln("Stopping workers")
+	ExecuteCmd()
```
ncdc, line 20: I'd look into following a pattern like you see at https://github.com/kubernetes/kubernetes/blob/88f97283393715c7fa870dd90be0fd0d0faad6bd/cmd/kube-controller-manager/controller-manager.go:

```go
func main() {
	command := app.NewControllerManagerCommand()
	code := cli.Run(command)
	os.Exit(code)
}
```

The reason to put the cobra command in a separate package is so others can import it & use it, and we can test it. The example I shared above puts it in `app`, which would be ok, or you could put it in `cmd` (`cmd/syncer/cmd`).

cmd/syncer/root.go
```diff
@@ -0,0 +1,125 @@
+/*
+Copyright 2021 The KCP Authors.
```
ncdc, line 2:
```suggestion
Copyright 2022 The KCP Authors.
```

cmd/syncer/root.go
```diff
@@ -0,0 +1,125 @@
+/*
```
ncdc, line 1: Please move this file into a subpackage (see comment above)

cmd/syncer/root.go
```diff
@@ -0,0 +1,125 @@
...
+const numThreads = 2
+
+var (
+	fromKubeconfig      string
+	fromClusterName     string
+	toKubeconfig        string
+	toContext           string
+	pclusterID          string
+	syncedResourceTypes []string
+)
```
ncdc, line 45: Please move this into an `Options` struct - see https://github.com/kcp-dev/kcp/blob/ba104a795fa57e7f452a91b5fb30f96463aa1599/cmd/ingress-controller/main.go#L122-L150 for a good reference example.

cmd/syncer/root.go
```diff
@@ -0,0 +1,125 @@
...
+
+var rootCmd = &cobra.Command{
```
ncdc, line 47: Instead of making this a private package variable, let's create a constructor (`NewSyncerCommand`) that returns the `*cobra.Command`. This function can also set up the options & bind flags.

cmd/syncer/root.go
```diff
@@ -0,0 +1,125 @@
...
+
+		synchronise()
+	},
+}
+
+func synchronise() {
+
+	// Create a client to dynamically watch "from".
+	if fromClusterName == "" {
+		klog.Fatal("--from_cluster is required")
```
ncdc, line 66: Let's move validation to `Options.Validate()`

cmd/syncer/root.go
```diff
@@ -0,0 +1,125 @@
...
+			klog.Fatal(err)
+		}
+	} else {
+		toConfig, err = rest.InClusterConfig()
+	}
+	if err != nil {
+		klog.Fatal(err)
+	}
+
+	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGILL, syscall.SIGINT)
```
ncdc, line 95: Since you're in here, please do #552 😄

cmd/syncer/root.go
```diff
@@ -0,0 +1,125 @@
...
+	defer cancel()
+
+	klog.Infoln("Starting workers")
```
ncdc, line 98: Can remove this - code inside `syncer.StartSyncer` does similar logging (with more details)

cmd/syncer/root.go
```diff
@@ -0,0 +1,125 @@
...
+	if err := syncer.StartSyncer(ctx, fromConfig, toConfig, sets.NewString(syncedResourceTypes...), fromClusterName, pclusterID, numThreads); err != nil {
+		klog.Fatal(err)
+	}
+
+	<-ctx.Done()
+
+	klog.Infoln("Stopping workers")
```
ncdc, line 105: Can remove this - code inside `syncer.StartSyncer` does similar logging (with more details)

cmd/syncer/root.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// Create a client to dynamically watch "from".
+	if fromClusterName == "" {
+		klog.Fatal("--from_cluster is required")
+	}
+
+	fromConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: fromKubeconfig}, nil).ClientConfig()
+	if err != nil {
+		klog.Fatal(err)
```
ncdc, line 72: Let's return any `err` instead of doing `klog.Fatal`.

cmd/syncer/root.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	<-ctx.Done()
+
+	klog.Infoln("Stopping workers")
+}
+
+func ExecuteCmd() {
+	rootCmd.SetHelpFunc(Help)
```
ncdc, line 109: Do we need a custom help func?

cmd/syncer/root.go
```diff
@@ -0,0 +1,125 @@
...
+	if err := syncer.StartSyncer(ctx, fromConfig, toConfig, sets.NewString(syncedResourceTypes...), fromClusterName, pclusterID, numThreads); err != nil {
+		klog.Fatal(err)
+	}
+
+	<-ctx.Done()
+
+	klog.Infoln("Stopping workers")
+}
+
+func ExecuteCmd() {
```
ncdc, line 108: Probably don't need this function any more, if you're doing
```go
code := cli.Run(command)
os.Exit(code)
```
in main.go

ncdc (changes requested):
@RAMESSESII2 thanks so much for the PR! If you have any questions about my review comments, please let me know.

cmd/syncer/main.go
RAMESSESII2: great idea, in fact, I already wanted to. I'm on it

cmd/syncer/root.go
RAMESSESII2: done

cmd/syncer/root.go
RAMESSESII2: done

cmd/syncer/main.go
RAMESSESII2: done

cmd/syncer/root.go
RAMESSESII2: @ncdc, should I keep the Logs field as well in the Syncer Options struct? Will there be any future prospects to it? WDYT

cmd/syncer/root.go
RAMESSESII2: done

cmd/syncer/root.go
ncdc: Yes, please. It configures `--v` and related flags.

cmd/syncer/root.go
RAMESSESII2: Not really, it's just originally, we had ``` 	fmt.Fprintf(os.Stderr, "Usage of %s [resourceTypes...]:\n", path.Base(os.Args[0])) ``` - inside flag.Usage
although yeah we don't really required it, cobra help will be enough ig.

cmd/cluster-controller/main.go
```diff
@@ -75,8 +74,7 @@ func (o *options) Validate() error {
 
 func main() {
 	// Setup signal handler for a cleaner shutdown
-	ctx, cancel := signal.NotifyContext(context.Background(), os.Kill, os.Interrupt)
-	defer cancel()
+	ctx := genericapiserver.SetupSignalContext()
```
ncdc, line 77: Re signal handling - sorry - I meant to only do the change specific to syncer. But that's awesome that you've done the others! Would you mind splitting them into a separate PR (and leave the syncer changes in this one)?

cmd/syncer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/component-base/logs"
+	"k8s.io/klog/v2"
+)
+
+const numThreads = 2
+
+func NewSyncerCommand() *cobra.Command {
+
```
ncdc, line 38:
```suggestion
```

cmd/syncer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+	options := NewDefaultOptions()
```
ncdc, line 39:
```suggestion
	options := NewDefaultOptions()
	
```

cmd/syncer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+
+	return nil
+}
+
+type Options struct {
+	fromKubeconfig  string
+	fromClusterName string
+	toKubeconfig    string
+	toContext       string
+	pclusterID      string
```
ncdc, line 108: Please make all of these fields exported - anyone programmatically creating & using an `Options` struct from outside this package needs to be able to set these values.

cmd/syncer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+		pclusterID:      "",
+		Logs:            logs.NewOptions(),
+	}
+}
+
+func (options *Options) BindFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&options.fromKubeconfig, "from_kubeconfig", options.fromKubeconfig, "Kubeconfig file for -from cluster.")
+	fs.StringVar(&options.fromClusterName, "from_cluster", options.fromClusterName, "Name of the -from logical cluster.")
+	fs.StringVar(&options.toKubeconfig, "to_kubeconfig", options.toKubeconfig, "Kubeconfig file for -to cluster. If not set, the InCluster configuration will be used.")
+	fs.StringVar(&options.toContext, "to_context", options.toContext, "Context to use in the Kubeconfig file for -to cluster, instead of the current context.")
```
ncdc, line 127: Please convert the flag names to use dashes (`-`) to separate words instead of underscores (`_`)

cmd/syncer/main.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The KCP Authors.
+Copyright 2022 The KCP Authors.
```
ncdc, line 2: Please don't change copyright dates of existing files (or if you're going to, I'd say only make additive changes; e.g. 2021-2022)

cmd/virtual-workspaces/main.go
```diff
@@ -33,13 +33,13 @@ import (
 )
 
 func main() {
-	stopCh := genericapiserver.SetupSignalHandler()
+	ctx := genericapiserver.SetupSignalContext()
```
ncdc, line 36: Split into separate PR (just focused on signal handling consistency)

cmd/syncer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
+/*
```
ncdc, line 1: nit: I would call this file syncer.go

cmd/syncer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+
+	options := NewDefaultOptions()
+	syncerCommand := &cobra.Command{
+		Use:   "syncer",
+		Short: "Synchronizes resources in `kcp` assigned to the clusters",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			err := options.validate(args)
+			return err
+		},
+	}
```
ncdc, line 47:
```suggestion
	}
	
```

cmd/syncer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+	options.BindFlags(syncerCommand.Flags())
+	return syncerCommand
```
ncdc, line 49:
```suggestion

	return syncerCommand
```

cmd/syncer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (options *Options) validate(args []string) error {
```
ncdc, line 52:
```suggestion
func (options *Options) Validate(args []string) error {
```

This needs to be exported so third party packages can invoke it

cmd/syncer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+	//non-flag arguements
+	syncedResourceTypes := args
+	if len(syncedResourceTypes) == 0 {
+		syncedResourceTypes = []string{"deployments.apps"}
```
ncdc, line 56: I would move the defaulting logic into a new Complete() function.

cmd/syncer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+			&toOverrides).ClientConfig()
+		if err != nil {
+			return err
+		}
+	} else {
+		toConfig, err = rest.InClusterConfig()
+	}
+	if err != nil {
+		return err
+	}
```
ncdc, line 90: I think I would probably move this into Complete() too. Anything you're creating as variables in here could be stored as unexported fields in Options, then used in Run()

cmd/syncer/cmd/root.go
```diff
@@ -0,0 +1,131 @@
...
+
+	ctx := genericapiserver.SetupSignalContext()
+
+	if err := syncer.StartSyncer(ctx, fromConfig, toConfig, sets.NewString(syncedResourceTypes...), options.fromClusterName, options.pclusterID, numThreads); err != nil {
+		return err
+	}
+
+	<-ctx.Done()
```
ncdc, line 98: Please move this to a new function called Run()

ncdc (changes requested):

cmd/syncer/cmd/root.go
ncdc: cc @sttts would you agree with this comment + the ones below?

cmd/cluster-controller/main.go
RAMESSESII2: yeah sure.

cmd/syncer/cmd/root.go
RAMESSESII2: sorry, I missed that, nevertheless, DONE

cmd/syncer/cmd/root.go
RAMESSESII2: yeah, right 👍🏼

cmd/syncer/cmd/root.go
RAMESSESII2: yeah sure, that makes more sense.

RAMESSESII2:
Hi @ncdc , I think I've incorporated the required changes. Please do check if I missed anything, thanks 😺

cmd/syncer/cmd/syncer.go
```diff
@@ -0,0 +1,164 @@
...
+	return syncerCommand
+}
+
+func (options *Options) Validate() error {
+	err := options.Complete()
+	if err != nil {
+		return err
+	}
+
+	err = options.Run()
```
sttts, line 68: this is odd. Why run in validation?

cmd/syncer/cmd/syncer.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	//bind flags
+	options.BindFlags(syncerCommand.Flags())
+
+	return syncerCommand
+}
+
+func (options *Options) Validate() error {
+	err := options.Complete()
```
sttts, line 63: would not expect completion inside of validate. Validate should be side-effect free, otherwise it would be called ValidateAndDoX

cmd/syncer/cmd/syncer.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	return nil
+}
+
+func (options *Options) Complete() error {
+	if len(options.SyncedResourceTypes) == 0 {
+		options.SyncedResourceTypes = []string{"deployments.apps"}
+		klog.Infoln("No resource types provided; using defaults.")
+	}
```
sttts, line 80: could we have that as default list in the Options struct on construction? Or is this a problem that a string set flag then cannot override it?

cmd/syncer/cmd/syncer.go
```diff
@@ -0,0 +1,164 @@
...
+
+	return nil
+}
+
+type Options struct {
+	FromKubeconfig  string
+	FromClusterName string
+	ToKubeconfig    string
+	ToContext       string
+	PclusterID      string
```
sttts, line 134: WorkloadClusterName

cmd/syncer/main.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The KCP Authors.
+Copyright 2021-2022 The KCP Authors.
```
sttts, line 2: just 2021, no range

cmd/syncer/main.go
```diff
@@ -17,87 +17,14 @@ limitations under the License.
...
-
-var (
-	fromKubeconfig  = flag.String("from_kubeconfig", "", "Kubeconfig file for -from cluster.")
-	fromClusterName = flag.String("from_cluster", "", "Name of the -from logical cluster.")
-	toKubeconfig    = flag.String("to_kubeconfig", "", "Kubeconfig file for -to cluster. If not set, the InCluster configuration will be used.")
-	toContext       = flag.String("to_context", "", "Context to use in the Kubeconfig file for -to cluster, instead of the current context.")
-	pclusterID      = flag.String("cluster", "",
-		fmt.Sprintf("ID of the -to cluster. Resources with this ID set in the '%s' label will be synced.", nscontroller.ClusterLabel))
+	"github.com/kcp-dev/kcp/cmd/syncer/cmd"
+	"k8s.io/component-base/cli"
```
sttts, line 23: we separate the kcp imports, i.e.
```
import (
   <go imports>

   <external imports>

   <kcp imports>
)
```
```

**Review by ncdc — CHANGES_REQUESTED** [review]
The changes are looking better! Thanks for your hard work!

File: cmd/syncer/cmd/syncer.go (hunk 27 above)
Comment (RAMESSESII2) [reply cmd/syncer/cmd/syncer.go:68]: yeah I realise that doesn't make any sense.

File: cmd/syncer/cmd/syncer.go [hunk 33]
```diff
@@ -0,0 +1,164 @@
...
+const numThreads = 2
+
+func NewSyncerCommand() *cobra.Command {
+	options := NewDefaultOptions()
+
+	syncerCommand := &cobra.Command{
+		Use:   "syncer",
+		Short: "Synchronizes resources in `kcp` assigned to the clusters",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			options.SyncedResourceTypes = args
```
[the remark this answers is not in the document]
RAMESSESII2: Sure, agreed

cmd/syncer/cmd/syncer.go
```diff
@@ -0,0 +1,164 @@
...
+func (options *Options) Complete() error {
+	if len(options.SyncedResourceTypes) == 0 {
+		options.SyncedResourceTypes = []string{"deployments.apps"}
+		klog.Infoln("No resource types provided; using defaults.")
+	}
+	klog.Infof("Syncing the following resource types: %s", options.SyncedResourceTypes)
+
+	// Create a client to dynamically watch "from".
+	if options.FromClusterName == "" {
+		return errors.New("--from_cluster is required")
```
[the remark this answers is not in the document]
RAMESSESII2: sorry, missed it

cmd/syncer/cmd/syncer.go
RAMESSESII2: sure, make sense

cmd/syncer/cmd/syncer.go
RAMESSESII2: Hey @ncdc , by "change", do you mean making ```SyncedResourceTypes``` a flag (StringArrayVar type)? I think that could be done.

cmd/syncer/cmd/syncer.go
RAMESSESII2: Hey @sttts, I'm sorry, I'm new to KCP but if I'm correct do you mean ClusterName is the same as WorkloadClusterName? And should I change ```FromClusterName``` to ```FromWorkloadClusterName```?

cmd/syncer/cmd/syncer.go
ncdc: @RAMESSESII2 that's right, we could make SyncedResourceTypes a flag.

RAMESSESII2:
Hey @ncdc, i think I've made the requested changes but still this can always be improved so let me know. 😄

ncdc:
@marun would you be willing to take over reviewing this?

marun (approved):
I've manually tested this to run the syncer with the cluster e2e.

cmd/syncer/cmd/syncer.go
```diff
@@ -0,0 +1,150 @@
...
+	FromClusterName     string
+	ToKubeconfig        string
+	ToContext           string
+	PclusterID          string
+	Logs                *logs.Options
+	SyncedResourceTypes []string
+
+	//non-exported members
+	toConfig   *rest.Config
+	fromConfig *rest.Config
```
sttts, line 125: this does not belong into options.

cmd/syncer/cmd/syncer.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (options *Options) Run(ctx context.Context) error {
+	if err := syncer.StartSyncer(ctx, options.fromConfig, options.toConfig, sets.NewString(options.SyncedResourceTypes...), options.FromClusterName, options.PclusterID, numThreads); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+type Options struct {
```
sttts, line 114: please move this into a sub-folder "options". Otherwise we cannot compose it.

cmd/syncer/cmd/syncer.go
sttts: These are derived from the options, aren't they?

Usually we have only flags values (with very few exceptions) in option structs. Would put those into cmd/syncer/options, with AddFlags, Complete and Validate methods.

Then at the cmd/syncer/cmd level you will use the option structs from the options package and turn e.g. the kubeconfig path name into rest.Configs.

cmd/syncer/cmd/syncer.go
marun: What are we intending to compose it with?

sttts: for example run in-process. We have it everywhere else. You won't know how to compose before you do it. Options struct are made exactly for that.

RAMESSESII2: Hi @sttts, I agree, it makes sense to put Options struct in its own package. And then let's say I create new Struct which will eventually be composed of options and the derived variables(toConfig and fromConfig). Something like this -
```
type SyncerOptions struct{
         options *options.Options
         toConfig *rest.Config
         fromConfig *rest.Config
}
```
1. Is  this somewhat you mean here - `Then at the cmd/syncer/cmd level you will use the option structs from the options package and turn e.g. the kubeconfig path name into rest.Configs`? And please if you can elaborate if I'm not on the same package?
2. If so, Is ```SyncerOptions``` a correct naming convention?

RAMESSESII2: Hi @sttts, I agree, it makes sense to put Options struct in its own package. And then let's say I create new Struct which will eventually be composed of options and the derived variables(toConfig and fromConfig). Something like this -
```
type SyncerOptions struct{
         options *options.Options
         toConfig *rest.Config
         fromConfig *rest.Config
}
```
1. Is this somewhat you mean here - Then at the cmd/syncer/cmd level you will use the option structs from the options package and turn e.g. the kubeconfig path name into rest.Configs? And please if you can elaborate if I'm not on the same package?
2. If so, Is SyncerOptions a correct naming convention?

sttts:
@RAMESSESII2 please see https://github.com/RAMESSESII2/kcp/pull/2.

sttts:
As background:

Syncer is simple enough that we can create the config just in the `Run` func. In bigger setups (like kube-apiserver) we would have an intermediate Config or Server struct holding the config. But here that would be overkill.

The structure in the options package is always the same, like in https://github.com/RAMESSESII2/kcp/pull/2. It's a pattern that proved to work well for composition of components in one binary.

RAMESSESII2:
> Syncer is simple enough that we can create the config just in the `Run` func. In bigger setups (like kube-apiserver) we would have an intermediate Config or Server struct holding the config. But here that would be overkill.
Ah, sure, makes sense and primarily that eliminates the problem of keeping track of those those two vaibales that are essentially derived from Options itself. Thanks 😄

sttts:
I made the linter unhappy:
```
--- a/cmd/syncer/options/options.go
+++ b/cmd/syncer/options/options.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	"github.com/spf13/pflag"
+
 	"k8s.io/component-base/logs"
 
 	nscontroller "github.com/kcp-dev/kcp/pkg/reconciler/namespace"
```
Can you fix that? Then it's ready to go.

RAMESSESII2:
> I made the linter unhappy:
> Can you fix that? Then it's ready to go.

yupp, IG we're good now 😄

sttts (approved):

ncdc (approved):

PR #609, opened 2022-03-03
admission: simplify informer sync in ClusterWorkspaceTypeExists

sttts:
Following @ncdc's example.


s-urbaniak (approved):
no concerns, it looks like this was double-locked, as `HasSynced()` has an internal lock as well and is goroutine-safe.

Also verified with other admission precedences, nice simplification 👍

PR #611, opened 2022-03-03
admission: add SubjectAccessReview to ClusterWorkspaceType use

sttts:
Authorize ClusterWorkspace creation against a clusterworkspacetypes verb=use permission via SubjectAccessReview.

By default everybody (with org access) can create universal ClusterWorkspaces. This is done through a bootstrapped ClusterRole+Binding in every org workspace. If this is not desired, that binding could be removed on a per-org basis.

### DIFF
--- a/config/organization/clusterworkspacetype-use-clusterrole.yaml
+++ b/config/organization/clusterworkspacetype-use-clusterrole.yaml
@@ -1,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:kcp:universal-clusterworkspacetype-use
+rules:
+- apiGroups: ["tenancy.kcp.dev"]
+  resources: ["clusterworkspacetypes"]
+  resourceNames: ["universal"]
+  verbs: ["use"]
--- a/config/organization/clusterworkspacetype-use-clusterrolebinding.yaml
+++ b/config/organization/clusterworkspacetype-use-clusterrolebinding.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: system:kcp:authenticated:universal-clusterworkspacetype-use
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:universal-clusterworkspacetype-use
+subjects:
+- apiGroup: rbac.authorization.k8s.io
+  kind: Group
+  name: system:authenticated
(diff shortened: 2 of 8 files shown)


config/organization/clusterworkspacetype-use-clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: system:authenticated:universal-clusterworkspacetype-use
```
sttts, line 4: @s-urbaniak is this the naming pattern we want to use? Does it match that of kube?

Alternatively, we could use a `system:kcp` prefix.

sttts: Renamed.

s-urbaniak: we used the `system:kcp` prefix 👍

config/organization/clusterworkspacetype-use-clusterrole.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:kcp:universal-clusterworkspacetype-use
+rules:
+- apiGroups: ["tenancy.kcp.dev"]
+  resources: ["clusterworkspacetypes"]
+  resourceNames: ["universal"]
+  verbs: ["use"]
```
s-urbaniak, line 9: since this is used only in the `create` case, it seems a bit counter-intuitive to use the verb `use` here. do we want to use that verb for other cases as well?

s-urbaniak (commented):
lgtm from a logical perspective. Suggestions:

1. let's add the restriction introduced here, i.e. in `workspaces.md`:

```
A cluster workspace of type `Universal` is a workspace without further initialization 
or special properties by default, and it can be used without a corresponding 
ClusterWorkspaceType object (though one can be added and its initializers will be 
applied).

Note: in order to create cluster workspaces of type `Universal` you must have `use` permissions against `clusterworkspacetypes` resources with the name `universal`.
```

2. let's add some more godoc to this admission handler and what its responsibilities are:
- adds initializers from type to workspace (i am not 100% clear on this use case)
- validates if the user has permissions to use universal cluster workspaces

config/organization/clusterworkspacetype-use-clusterrole.yaml
ncdc: I agree but I don't have any alternative suggestions at the moment. Let's maybe file an issue to re-review/follow-up in the future?

ncdc (approved):

config/organization/clusterworkspacetype-use-clusterrole.yaml
sttts: alternative: instantiate

s-urbaniak: created https://github.com/kcp-dev/kcp/issues/617

PR #614, opened 2022-03-04
e2e: Enable use of persistent server fixture

marun:
Since e2e test execution today is dominated by the ~9s it takes to start a kcp instance, enabling the use of a persistent server speeds up iteration on test development. Tests must be compatible with this, though, so more work will be needed before any given e2e test will work with persistent fixture.

Usage: `go test -v --kubeconfig=/path/to/.kcp/admin.kubeconfig`

Only tests that have been explicitly configured to use shared fixture will use a persistent server configured via `--kubeconfig`.


test/e2e/framework/fixture.go
```diff
@@ -47,12 +47,34 @@ func NewKcpFixture(t *testing.T, cfgs ...KcpConfig) *KcpFixture {
 	artifactDir, dataDir, err := ScratchDirs(t)
 	require.NoError(t, err, "failed to create scratch dirs: %v", err)
 
+	f.Servers = map[string]RunningServer{}
+
+	if len(TestConfig.KubeConfig) > 0 {
```
sttts, line 52: does it make sense to add `Disruptive: true` to the fixture config to opt-out of the possibly global fixture?

marun: I don't think we'll need it. For now, this is a developer-only feature intended to make it easier to get the tests to a hermetic state.  Long term, the goal is for all non-disruptive tests to move to shared fixture that they don't configure in any way, and for the disruptive tests to be the only tests defining test-local fixture.

marun: There is also a follow-up (#636) that switches to running test-managed syncer to enable a few more tests to run with shared fixture.

marun:
Rebased

test/e2e/framework/kcp.go
```diff
@@ -411,3 +406,74 @@ func (c *kcpServer) monitorEndpoint(client *rest.RESTClient, endpoint string) {
...
+	if fs.Size() == 0 {
+		return nil, fmt.Errorf("%s points to an empty file", kubeconfigPath)
+	}
+
+	rawConfig, err := clientcmd.LoadFromFile(kubeconfigPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load admin kubeconfig: %w", err)
+	}
+
+	return clientcmd.NewNonInteractiveClientConfig(*rawConfig, "root", nil, nil), nil
```
sttts, line 428: unrelated to the this PR: "root" here is strange. That's the root workspace, i.e. not system:admin. Wondering what this client config is used for. Is it for health checks?

test/e2e/framework/kcp.go
```diff
@@ -343,24 +344,18 @@ func (c *kcpServer) loadCfg() error {
...
-		}
-
-		rawConfig, err := clientcmd.LoadFromFile(c.kubeconfigPath)
+		config, err := loadKubeConfig(c.kubeconfigPath)
 		if err != nil {
-			lastError = fmt.Errorf("failed to load admin kubeconfig: %w", err)
+			// A missing file is likely caused by the server not
+			// having started up yet. Ignore these errors for the
+			// purposes of logging.
+			if !os.IsNotExist(err) {
```
sttts, line 352: there is a race that can lead to other errors: zero file size or corrupted because partially written file.

test/e2e/api_inheritance/api_inheritance_test.go
```diff
@@ -88,9 +80,6 @@ func TestAPIInheritance(t *testing.T) {
 				ctx = withDeadline
 			}
 
-			require.Equal(t, 1, len(f.Servers), "incorrect number of servers")
-			server := f.Servers[serverName]
```
sttts, line 92: :+1:

All these should go. Not sure why they were added in the first place. Lack of trust into the fixtures?

sttts: unrelated to this PR: can we have a helper for the context logic above? It's painful to see that everywhere.

sttts (approved):

test/e2e/api_inheritance/api_inheritance_test.go
marun: > unrelated to this PR: can we have a helper for the context logic above? It's painful to see that everywhere.

I'm happy to strip the context logic out entirely. With `t.Cleanup` it should be possible for anyone who wants to use a context to instantiate local to use vs relying on anything passed around.

test/e2e/framework/kcp.go
marun: Isn't that still handled by loadKubeConfig? I was trying to maintain the existing semantics, just factor out the common bits for reuse.

PR #616, opened 2022-03-04
e2e: Fix broken workspacetype initializer test

marun:
I noticed that the workspacetype initializer test was taking ~50s - quadruple the runtime of any other test. Investigation revealed that the test was broken in 3 ways:

 - incorrect deletion of an element from a slice
 - using Update instead of UpdateStatus to effect a status change
 - not checking the result of a subsequent expect meant the first 2 issues were not failing the test

The extended duration of test execution was the result of the expect taking 30s before timing out, 2x5s sleeps, and the usual 10s overhead of starting a kcp server fixture.

### DIFF
--- a/test/e2e/workspacetype/controller_test.go
+++ b/test/e2e/workspacetype/controller_test.go
@@ -101,13 +101,13 @@ func TestClusterWorkspaceTypes(t *testing.T) {
 			},
 		},
 		{
-			name: "create a workspace with a type that has initializers",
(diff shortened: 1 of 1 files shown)


marun:
Added a commit that simplifies the test to the use of only a single initializer. Happy to drop it if there's a good reason a single initializer provides insufficient coverage.

sttts (approved):

PR #619, opened 2022-03-04
Remove workloadclusters, apiresourceimports, negotiatedapiresources from root and org workspaces

sttts:
The root workspace and org workspaces are not meant for workloads. Hence, we remove these types fro now to not encourage an anti-pattern.


ncdc (approved):

test/e2e/api_inheritance/wildwest.dev_cowboys.yaml
```diff
@@ -0,0 +1,58 @@
+
```
marun, line 1: nit: Does this need to be duplicated?

test/e2e/api_inheritance/api_inheritance_test.go
```diff
@@ -18,49 +18,55 @@ package api_inheritance
 
 import (
 	"context"
+	"embed"
```
marun, line 21: (No action required) I'm also rewriting this test (for shared fixture #636) so I guess there's a messy rebase in my future.

cmd/cluster-controller/main.go
```diff
@@ -135,15 +135,7 @@ func main() {
...
-
-	if syncer != nil {
-		prepared, err := syncer.Prepare()
-		if err != nil {
-			fmt.Fprintln(os.Stderr, err)
-			os.Exit(1)
-		}
-		go prepared.Start(ctx)
-	}
+	go syncer.Start(ctx)
```
marun, line 138: I think you'll find that you still need the nil check here. If a syncer mode isn't specified, no syncer is not intended to run.

marun (commented):
Aha, I'm hopeful the config changes you've made to the cluster controllers will fix the problem I was seeing trying to run in-process for testing.

test/e2e/api_inheritance/wildwest.dev_cowboys.yaml
sttts: fixing

test/e2e/api_inheritance/api_inheritance_test.go
sttts: This one was necessary. The test was just wrong.

cmd/cluster-controller/main.go
sttts: sigh. Constructors returning nil.

sttts: fixed

pkg/reconciler/cluster/syncer/syncer_controller.go
```diff
@@ -77,51 +77,7 @@ func NewController(
...
-				}
-				if genericcontrolplanescheme.Scheme.IsVersionRegistered(gv) {
-					apiResourceList, err := discoveryClient.ServerResourcesForGroupVersion(gv.String())
-					if err != nil {
-						return PreparedController{}, err
-					}
-					for _, apiResource := range apiResourceList.APIResources {
-						gvk := gv.WithKind(apiResource.Kind)
-						if !strings.Contains(apiResource.Name, "/") && genericcontrolplanescheme.Scheme.Recognizes(gvk) {
-							c.syncerManager.genericControlPlaneResources = append(c.syncerManager.genericControlPlaneResources, gv.WithResource(apiResource.Name))
```
davidfestal, line 111: This code was to also include generic control plane resources, like config maps or secrets, in what can be synced by the syncer, even though there was no related API imports, since those resource types are by default present in every KCP workspace.

sttts: We should not use discovery for this. If this was about configmaps and secrets, we have to hardcode those very explicitly in the syncer. We cannot use some implicit mechanism like asking discovery.

davidfestal: Completely agree that the change made in commit https://github.com/kcp-dev/kcp/pull/619/commits/4bb843b2dc004344f7d0993223901cdf7d9fb11c is sufficient and better when hard-coded.

davidfestal (approved):

jmprusi (approved):

PR #622, opened 2022-03-04
Update README to clearly state that kcp has two layers

MikeSpreitzer:


README.md
```diff
@@ -1,10 +1,10 @@
-# `kcp` provides a true multi-tenant Kubernetes control plane for workloads on many clusters
+# `kcp` provides a multi-tenant Kubernetes API service and a transparent multi-cluster workload service
```
smarterclayton, line 1: I think the difference between these two sentences is:

current: Get people excited about potential
proposed: Be very precise about the components

Generally at this stage in a project and prototype I value both - but prioritize the former rather than the latter.  Do you understand why I would argue for that prioritization, and there are counter arguments you'd make (such that the first sentence doesn't actually accomplish getting people excited)?

MikeSpreitzer: Yes, I understand that prioritization.  What do you think of keeping the top heading as it is currently merged and doing the other changes in this PR?

README.md
```diff
@@ -1,10 +1,10 @@
...
 
 ![build status badge](https://github.com/kcp-dev/kcp/actions/workflows/ci.yaml/badge.svg)
 
-`kcp` is a generic [CustomResourceDefinition](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) (CRD) apiserver that is divided into multiple "[logical clusters](docs/investigations/logical-clusters.md)" that enable multitenancy of cluster-scoped resources such as CRDs and Namespaces. Each of these logical clusters is fully isolated from the others, allowing different teams, workloads, and use cases to live side by side.
+`kcp` has two layers.  The lower layer is a generic [CustomResourceDefinition](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) (CRD) apiserver that is divided into multiple "[logical clusters](docs/investigations/logical-clusters.md)" that enable multitenancy of cluster-scoped resources such as CRDs and Namespaces. Each of these logical clusters is fully isolated from the others, allowing different teams, workloads, and use cases to live side by side.
```
smarterclayton, line 5: I think instead of layers we should use a different word - maybe "components" or "components".  Transparent multi-cluster is a component that integrates with the control plane, and needs some semi-novel API access patterns and concepts from Kube.  It's the "killer app" to drive usage of the control plane.  Control plane will not be fundamentally aware of transparent-multicluster (control plane is useful without it).  If tmc struggles, we might separate it out.  Ideally everything service can be completely layered on top of control plane (either as a code plugin or a layer on top).

smarterclayton: I would probably use `kcp` to describe control plane.  I would describe tmc as a subsystem or sub service.

MikeSpreitzer: I do not understand https://github.com/kcp-dev/kcp/pull/622#discussion_r824055662 ; it starts by disliking the word "layers" but the goes on to describe an arrangement that is exactly layering.

This "kcp" repo includes tmc, so it seems odd to me to define "kcp" to mean something narrower.

sttts: tmc is in the repo to allow fast iterations. There is no technical reason we can't move it to kcp-dev/workloads eventually. Maybe we will.

tmc is also a tool to ensure what we build in kcp core (I like that term) is suitable for complex services. E.g. note that in v0.5 tmc uses APIExport/APIBindings, i.e. the core primitives every other service will use.

sftim, line 5: If I'm right, you don't _have_ to have multiple logical clusters; if that feature wasn't useful for you, you could use KCP without paying any attention to the logical cluster aspect at all.

If so:
```suggestion
`kcp` has two layers.  The lower layer is a generic [CustomResourceDefinition](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/) (CRD) apiserver that you can further divide into multiple "[logical clusters](docs/investigations/logical-clusters.md)". KCP logical clusters enable multitenancy for cluster-scoped resources such as CRDs and Namespaces. Each of these logical clusters is fully isolated from the others, allowing different teams, workloads, and use cases to live side by side.
```

README.md
```diff
@@ -16,12 +16,13 @@ By default, `kcp` only knows about:
 
 Any other resources, including standard Kubernetes resources like [`Deployment`](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/)s and the rest, can be added as CRDs and be replicated onto one or more Kubernetes clusters.
 
+The upper layer of `kcp` is a transparent multi-cluster workload service.  This service keeps track of backend "physical clusters" and syncs API resources defined in `kcp` to those physical clusters.
```
sftim, line 19:
```suggestion
The upper layer of `kcp` is a transparent multi-cluster workload service.  This service keeps track of one or more _backend clusters_ and syncs API resources (defined in the `kcp` front end cluster) to those backend clusters.
```

README.md
```diff
@@ -16,12 +16,13 @@ By default, `kcp` only knows about:
...
 
 ## Why would I want that?
 
 Kubernetes is mainly known as a container orchestration platform today, but we believe it can be even more.
 
-With the power of CRDs, Kubernetes provides a flexible platform for declarative APIs of _all types_, and the reconciliation pattern common to Kubernetes controllers is a powerful tool in building robust, expressive systems.
+Kubernetes itself can be seen as having two layers: (1) an extensible API service, and (2) a control plane built using that API service to keep track of worker nodes and manage containerized workloads on those worker nodes.  The API service of Kubernetes is extensible through two mechanisms: (i) aggregation of custom external apiservers (which _may_ be built in the same style as the `kube-apiserver` by building on [k/apiserver](github.com/kubernetes/apiserver)) and (ii) addition of CRDs.  `kcp` focuses on CRDs.  With its extensible API service, Kubernetes provides a flexible platform for declarative APIs of _all types_ --- and the reconciliation pattern common to Kubernetes controllers is a powerful tool in building robust, expressive systems.
```
sftim, line 25: This isn't quite right, in that the foundational APIs (think Namespace and kin) still also rely on control loops external to the API server. Having _just_ an API server isn't that useful. That holds whether you want the objects in the API in order to manage container compute, a factory, home automation, or any other weird and wonderful thing you could do with an extensible CRUD API like the Kubernetes core gives you.

I recommend having a think about how to make the point here but without sacrificing accuracy over the details.

ncdc:
Have a major update to the readme landing in #1911. Going to close this as outdated. Happy to continue to iterate on the docs in future PRs.

/close

Issue #604, opened 2022-03-02
Reconciliation of resources back to pclusters by KCP if deleted by mistake


samyak-jn:
**Describe the bug**
In the perfect world scenario, Ideally, KCP should reconcile back the resources which are by mistake/intentionally deleted from the physical cluster end. The only source of truth in such operation should be done via the KCP end only, but for now if resources are being deleted (need to check for create/update as well) the resources are not reconciled back to their original state of KCP. This should be fixed, as confirmed with team folks as well the tests for this aren't available (might be broken) 

**To Reproduce**
1. Create a ns in kcp and deploy a simple pod to reflect resources in pcluster
2. Using kubeconfig of pcluster, check for ns created 
3. Delete the above namespace from the pcluster end
4. Wait to see the reconciliation to happen but it is broken atm so you wont be getting that ns back but it should ideally come back as it is existing in KCP 

**Expected behavior**
_KCP should reconcile back the resources which are by mistake/intentionally deleted from the physical cluster end_

ncdc:
I think this is because the "spec syncer", which upserts from kcp to pclusters, only watches resources in kcp. If something is deleted in the pcluster, the spec syncer won't know about it. If you make any updates to those resources in kcp (e.g. label or annotate them), they should be rereconciled and recreated. Otherwise, for the time being, you have to wait for the resync period, which is many hours.

marun:
I'm going to take this. Not sure yet if P3 will be doable, but P4 for sure.

kasturinarra:
I have hit below issue as well during my test:
======================================
1) when updated the deployment spec in kcp to have replica count of 12, i see that number of pods run are around 12 in physical cluster which is fine, when i try to update deployment spec in physical cluster to have 14 replicas i see that pods does get increase in number and running oc get deployment also shows the available number as 14 which is not expected as we believe kcp is source of truth

kcp context:
=================
[knarra@knarra kcp]$ oc get deployment 
NAME    READY   UP-TO-DATE   AVAILABLE   AGE
kuard   14/12   14           14          18m

pyhiscal cluster context:
======================
[knarra@knarra ~]$ oc get pods --all-namespaces | grep kuard | wc -l
14

2) When i delete deployment from physical cluster i see that deployments does not show up in pyhsical cluster and ideally they should because kcp is source of truth.

ncdc:
@kasturinarra if you wait a bit, do the deployments go back to 12?

kasturinarra:
[...]

PR #628, opened 2022-03-04
syncer: Use UID pre-condition for delete to downstream operations

astefanutti:
This fixes possible races when upstream objects are deleted and re-created.

An example of such races I've encountered, is:
1. The downstream object is deleted for some reasons,
2. The upstream object is deleted, but the operation fails and is re-queued for retry with backoff
3. The upstream object is re-created, that is with the same name
4. The re-created upstream object is synced downstream
5. The delete operation that failed in step 2. is processed
6. The re-created downstream object is deleted

In that particular example, it could be possible to prevent the race by simply ignoring 404 errors, but it seems more robust to use a concurrency control mechanism, to avoid unanticipated races for delete operations, as for update ones.

The UID of the downstream objects, used in the delete pre-conditions for concurrency control, are stored in an annotation. That seemed the easiest way to transport the downstream UID, but there may be better options I'm not aware of.

### DIFF
--- a/pkg/syncer/specsyncer.go
+++ b/pkg/syncer/specsyncer.go
@@ -86,11 +86,13 @@ func NewSpecSyncer(from, to *rest.Config, syncedResourceTypes []string, kcpClust
 	return New(kcpClusterName, pclusterID, fromDiscovery, fromClient, toClient, KcpToPhysicalCluster, syncedResourceTypes, pclusterID)
 }
 
-func (c *Controller) deleteFromDownstream(ctx context.Context, gvr schema.GroupVersionResource, namespace, name string) error {
-	// TODO: get UID of just-deleted object and pass it as a precondition on this delete.
-	// This would avoid races where an object is deleted and another object with the same name is created immediately after.
-
-	return c.toClient.Resource(gvr).Namespace(namespace).Delete(ctx, name, metav1.DeleteOptions{})
+func (c *Controller) deleteFromDownstream(ctx context.Context, gvr schema.GroupVersionResource, upstreamNamespace, name string, downstreamUID string) error {
+	if err := c.toClient.Resource(gvr).Namespace(upstreamNamespace).Delete(ctx, name, metav1.DeleteOptions{
+		Preconditions: metav1.NewUIDPreconditions(downstreamUID),
+	}); err != nil && !k8serrors.IsNotFound(err) && !k8serrors.IsConflict(err) {
+		return err
+	}
+	return nil
 }
 
 const namespaceLocatorAnnotation = "kcp.dev/namespace-locator"
--- a/pkg/syncer/statussyncer.go
+++ b/pkg/syncer/statussyncer.go
@@ -18,15 +18,18 @@ package syncer
 
 import (
 	"context"
+	"encoding/json"
 
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/pointer"
 )
 
 func deepEqualStatus(oldObj, newObj interface{}) bool {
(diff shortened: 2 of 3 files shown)


pkg/syncer/specsyncer.go
```diff
@@ -86,11 +86,13 @@ func NewSpecSyncer(from, to *rest.Config, syncedResourceTypes []string, kcpClust
...
 
-func (c *Controller) deleteFromDownstream(ctx context.Context, gvr schema.GroupVersionResource, namespace, name string) error {
-	// TODO: get UID of just-deleted object and pass it as a precondition on this delete.
-	// This would avoid races where an object is deleted and another object with the same name is created immediately after.
-
-	return c.toClient.Resource(gvr).Namespace(namespace).Delete(ctx, name, metav1.DeleteOptions{})
+func (c *Controller) deleteFromDownstream(ctx context.Context, gvr schema.GroupVersionResource, upstreamNamespace, name string, downstreamUID string) error {
+	if err := c.toClient.Resource(gvr).Namespace(upstreamNamespace).Delete(ctx, name, metav1.DeleteOptions{
+		Preconditions: metav1.NewUIDPreconditions(downstreamUID),
+	}); err != nil && !k8serrors.IsNotFound(err) && !k8serrors.IsConflict(err) {
```
marun, line 92: What would `k8serrors.IsConflict` being true indicate, and why is it desirable to silently ignore that?

pkg/syncer/statussyncer.go
```diff
@@ -66,26 +70,43 @@ func NewStatusSyncer(from, to *rest.Config, syncedResourceTypes []string, kcpClu
...
 		return err
 	}
 
-	// TODO: verify that we really only update status, and not some non-status fields in ObjectMeta.
-	//       I believe to remember that we had resources where that happened.
+	if _, err := c.toClient.Resource(gvr).Namespace(upstreamNamespace).
+		Patch(ctx, downstreamObj.GetName(), types.ApplyPatchType, data, metav1.PatchOptions{
+			FieldManager: syncerApplyManager,
+			Force:        pointer.Bool(true),
+		}, "status"); err != nil {
```
marun, line 88: nit: Maybe add a blank line to make up for the lack of visual indication between the conditional and its block? Either that or maybe this form of error handling (`if err...; err != nil`) should only apply to single-line conditionals...?

pkg/syncer/syncer.go
```diff
@@ -45,9 +45,10 @@ import (
 )
 
 const (
-	resyncPeriod       = 10 * time.Hour
-	SyncerNamespaceKey = "SYNCER_NAMESPACE"
-	syncerApplyManager = "syncer"
+	resyncPeriod        = 10 * time.Hour
+	SyncerNamespaceKey  = "SYNCER_NAMESPACE"
+	syncerApplyManager  = "syncer"
+	targetUIDAnnotation = "workload.kcp.dev/target.uid"
```
marun, line 51: (No action required) Maybe dashes are used instead of periods in the non-domain part of the key (e.g. target-uid)?

pkg/syncer/specsyncer.go
```diff
@@ -86,11 +86,13 @@ func NewSpecSyncer(from, to *rest.Config, syncedResourceTypes []string, kcpClust
...
 }
 
-func (c *Controller) deleteFromDownstream(ctx context.Context, gvr schema.GroupVersionResource, namespace, name string) error {
-	// TODO: get UID of just-deleted object and pass it as a precondition on this delete.
-	// This would avoid races where an object is deleted and another object with the same name is created immediately after.
-
-	return c.toClient.Resource(gvr).Namespace(namespace).Delete(ctx, name, metav1.DeleteOptions{})
+func (c *Controller) deleteFromDownstream(ctx context.Context, gvr schema.GroupVersionResource, upstreamNamespace, name string, downstreamUID string) error {
+	if err := c.toClient.Resource(gvr).Namespace(upstreamNamespace).Delete(ctx, name, metav1.DeleteOptions{
+		Preconditions: metav1.NewUIDPreconditions(downstreamUID),
```
marun, line 91: What if downstreamUID is empty?

marun (commented):
My inline comments assume what you're proposing makes sense, but after reviewing I'm not clear that is the case. I would appreciate it if you would update the PR description with a crisp explanation of the problem you are trying to solve and how exactly the proposed solution addresses that problem. Please emphasise the 'why' rather than the 'how'.

Assuming this PR evolves into something we want to merge, I would also expect test changes to cover the new behavior.

astefanutti:
> My inline comments assume what you're proposing makes sense, but after reviewing I'm not clear that is the case. I would appreciate it if you would update the PR description with a crisp explanation of the problem you are trying to solve and how exactly the proposed solution addresses that problem. Please emphasise the 'why' rather than the 'how'.

Thanks for the review. I've updated the description with the use case that I faced, and that explains why I'm proposing this change. 
 
> Assuming this PR evolves into something we want to merge, I would also expect test changes to cover the new behavior.

Agreed.

pkg/syncer/specsyncer.go
astefanutti: In that case, I've noticed the pre-condition is simply ignored.

pkg/syncer/specsyncer.go
astefanutti: In that particular case, that is when the UID set in the pre-condition differs from the UID of the downstream object that actually exists, a conflict implies the object that was meant to be deleted no longer exists, and has been replaced with a new object, hence the operation that is being performed has already been fulfilled, and the error can be ignored, similarly to the 404 case.

pkg/syncer/statussyncer.go
astefanutti: I agree, I'd be inclined to go with the "apply to single-line conditionals only". I'll do it if we decide to pursue the work on that PR.

pkg/syncer/syncer.go
astefanutti: I agree, I'll do it if we decide to pursue the work on that PR.

marun:
That's helpful detail, thank you!

Unless I'm missing something, what you're seeing is a symptom of a syncer that is not responding to pcluster events (also reported by #604). I think it might be preferable to first ensure reconciliation is bidirectional and that there exists sufficient test coverage to understand whether additional coordination is required.

astefanutti:
> Unless I'm missing something, what you're seeing is a symptom of a syncer that is not responding to pcluster events (also reported by https://github.com/kcp-dev/kcp/issues/604). I think it might be preferable to first ensure reconciliation is bidirectional and that there exists sufficient test coverage to understand whether additional coordination is required.

Yes, I think that would greatly reduce the likelihood of the race in that case, but I don't think that would eliminate it nonetheless.

ncdc:
@astefanutti would you mind filing an issue for this?

astefanutti:
@ncdc sure, I've created #643. I'd be inclined to close that PR, and continue the discussion there, if that's alright with you.

Issue #634, opened 2022-03-04
CI: linter job does not print lines


sttts:
```
run golangci-lint
  Running [/home/runner/golangci-lint-1.44.2-linux-amd64/golangci-lint run --out-format=github-actions --new-from-patch=/tmp/tmp-1622-2iRAsmvu7xoU/pull.patch --new=false --new-from-rev= --timeout=5m] in [] ...
  Error: ineffectual assignment to err (ineffassign)
  
  Error: issues found
  Ran golangci-lint in 24648ms
```
(source https://github.com/kcp-dev/kcp/runs/5429150824?check_suite_focus=true#step:4:109)

This is not helpful.

PR #635, opened 2022-03-05
ci: use lint make target

jbpratt:
golanci-lint action job uses a custom format to annotate code within
GitHub, we can override it (forcefully) to achieve the desired output
(file and line numbers for lint issue in the output). Or we can just use
the make target directly :)

https://github.com/golangci/golangci-lint-action/issues/119#issuecomment-981090648

fixes https://github.com/kcp-dev/kcp/issues/634

Signed-off-by: jbpratt <jbpratt78@gmail.com>

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -61,10 +61,8 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: v1.17
-      - uses: golangci/golangci-lint-action@v2
-        with:
-          only-new-issues: true
-          args: --timeout=5m
+      - name: Run golangci-lint
+        run: make lint
 
   codegen:
     name: codegen
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,10 @@ OPENSHIFT_GOIMPORTS_BIN := openshift-goimports
 OPENSHIFT_GOIMPORTS := $(TOOLS_DIR)/$(OPENSHIFT_GOIMPORTS_BIN)-$(OPENSHIFT_GOIMPORTS_VER)
 export OPENSHIFT_GOIMPORTS # so hack scripts can use it
 
+GOLANGCI_LINT_VER := v1.44.2
+GOLANGCI_LINT_BIN := golangci-lint
+GOLANGCI_LINT := $(GOBIN_DIR)/$(GOLANGCI_LINT_BIN)-$(GOLANGCI_LINT_VER)
+
 all: build
 .PHONY: all
 
@@ -47,8 +51,11 @@ install:
 	go install ./cmd/...
 .PHONY: install
 
-lint:
-	golangci-lint run ./...
+$(GOLANGCI_LINT):
+	GOBIN=$(GOBIN_DIR) $(GO_INSTALL) github.com/golangci/golangci-lint/cmd/golangci-lint $(GOLANGCI_LINT_BIN) $(GOLANGCI_LINT_VER)
+
+lint: $(GOLANGCI_LINT)
+	$(GOLANGCI_LINT) run --timeout=10m ./...
 .PHONY: lint
 
 vendor: ## Vendor the dependencies


sttts:
Can you rebase? Not sure why e2es fail.

jbpratt:
rebased changed!

ncdc:
@stevekuznetsov & I were chatting a while ago about just running `make lint` instead of trying to use a GH action. WDYT?

.github/workflows/ci.yaml
```diff
@@ -64,7 +64,7 @@ jobs:
       - uses: golangci/golangci-lint-action@v2
         with:
           only-new-issues: true
-          args: --timeout=5m
+          args: --timeout=5m --out-${NO_FUTURE}format colored-line-number
```
ncdc, line 67: I would prefer to avoid using a flag where the tooling's guidance is to avoid using it, because it's not guaranteed to be stable going forward. How about running `make lint` instead?

jbpratt:
> @stevekuznetsov & I were chatting a while ago about just running `make lint` instead of trying to use a GH action. WDYT?

sounds good to me! This felt like a gross hack anyway :) I've pushed changes and had to add/bump the timeout time

failure with 5m timeout: https://github.com/jbpratt/kcp/runs/5450736838?check_suite_focus=true 
pass with 10m timeout: https://github.com/jbpratt/kcp/runs/5450903835?check_suite_focus=true

.github/workflows/ci.yaml
```diff
@@ -61,10 +61,11 @@ jobs:
       - uses: actions/setup-go@v2
         with:
           go-version: v1.17
-      - uses: golangci/golangci-lint-action@v2
-        with:
-          only-new-issues: true
-          args: --timeout=5m
+      - name: Install golangci-lint
+        run: |
+          curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin
```
ncdc, line 66: If we're going to do this, we should pin to a specific version. However, I'm wondering if it would be possible to make sure everyone is running the same lint version, both locally and in CI. Would you be willing to see if you could do something similar to https://github.com/kcp-dev/kcp/blob/467fc8a2fc2f8b67139eba965301eae8990cb463/Makefile#L24-L27 and https://github.com/kcp-dev/kcp/blob/467fc8a2fc2f8b67139eba965301eae8990cb463/Makefile#L59-L60?

jbpratt: I can give it a try, this is grabbing (technically) the same version that CI was using but I think pinning to a version would be the safest. Would it be fine to stuff away the `golangci-lint` install in the make target for linting (install if it isn't already)? I could easily drop this in the `bin/` directory so that we install the linter locally within the repo but this may not be ideal. :cat: open to ideas

ncdc: What I linked above does the install into `hack/tools`, if you follow that model.

jbpratt: I see, sounds good! I will give that a try.

jbpratt: wowee, you were way ahead of me :) made the requested changes
```
❯ make lint
GOBIN=/home/jbpratt/git/kcp/hack/tools ./hack/go-install.sh github.com/golangci/golangci-lint/cmd/golangci-lint golangci-lint v1.44.2
go: creating new go.mod: module fake/mod
go get: installing executables with 'go get' in module mode is deprecated.
	To adjust and download dependencies of the current module, use 'go get -d'.
	To install using requirements of the current module, use 'go install'.
	To install ignoring the current module, use 'go install' with a version,
	like 'go install example.com/cmd@latest'.
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.
go get: added 4d63.com/gochecknoglobals v0.1.0
go get: added github.com/Antonboom/errname v0.1.5
go get: added github.com/Antonboom/nilnil v0.1.0
go get: added github.com/BurntSushi/toml v1.0.0
go get: added github.com/Djarvur/go-err113 v0.0.0-20210108212216-aea10b59be24
go get: added github.com/Masterminds/semver v1.5.0
go get: added github.com/OpenPeeDeeP/depguard v1.1.0
go get: added github.com/alexkohler/prealloc v1.0.0
go get: added github.com/ashanbrown/forbidigo v1.3.0
go get: added github.com/ashanbrown/makezero v1.1.0
go get: added github.com/beorn7/perks v1.0.1
go get: added github.com/bkielbasa/cyclop v1.2.0
go get: added github.com/blizzy78/varnamelen v0.6.0
go get: added github.com/bombsimon/wsl/v3 v3.3.0
go get: added github.com/breml/bidichk v0.2.2
go get: added github.com/breml/errchkjson v0.2.3
go get: added github.com/butuzov/ireturn v0.1.1
go get: added github.com/cespare/xxhash/v2 v2.1.2
....
/home/jbpratt/git/kcp/hack/tools/golangci-lint-v1.44.2 run --timeout=10m ./...

❯ make lint
/home/jbpratt/git/kcp/hack/tools/golangci-lint-v1.44.2 run --timeout=10m ./...
```

ncdc: Cool!

ncdc:
LGTM. PTAL @sttts @marun @stevekuznetsov @davidfestal

ncdc (approved):

marun:
LGTM

PR #649, opened 2022-03-08
Move `TEST_ARGS` for `make test-e2e` to after the package list

marun:
This enables the use of `-args [custom args]` and `-- [custom args]`, which have to appear after the package list.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ E2E_PARALLELISM ?= 1
 .PHONY: test-e2e
 test-e2e: WHAT ?= ./test/e2e...
 test-e2e: install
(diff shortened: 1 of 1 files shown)


ncdc:
Is it possible to move TEST_ARGS to after the package list?

marun:
> Is it possible to move TEST_ARGS to after the package list?

I think we want both. TEST_ARGS for e.g. -v, EXTRA_TEST_ARGS for flags defined in the test (e.g. `--kubeconfig`)

ncdc:
I was going to suggest `TEST_ARGS='-v -- --kubeconfig` or something like that

marun:
> I was going to suggest `TEST_ARGS='-v -- --kubeconfig` or something like that

Done

ncdc (approved):

PR #651, opened 2022-03-08
e2e: Convert the virtual workspace test to shared fixture

marun:


test/e2e/virtual/workspaces/virtual_workspace_test.go
```diff
@@ -313,15 +322,15 @@ func testWorkspacesVirtualWorkspaces(t *testing.T, standalone bool) {
 				require.NoError(t, err, "failed to create root workspace roles")
 
 				err = server.virtualWorkspaceExpectations[0](func(w *tenancyv1beta1.WorkspaceList) error {
-					expectedOrgs := sets.NewString(orgName)
+					expectedOrgs := sets.NewString(orgName, "default")
```
marun, line 325: @davidfestal I needed this change to get the tests passing, but I'm assuming this is not desirable. Any idea what I might have done and how to fix it?

sttts: welcome to shared test data (users with groups) 🎉

Workaround: one group membership per test that depends on it.

sttts: actually even per-test users.

marun: One test did require a unique group to ensure auth isolation. I updated a different test that was using "root:default" to use a test org instead. PTAL

sttts (approved):

PR #659, opened 2022-03-08
Proxy to terminate TLS and handle client cert auth

csams:
kcp-front-proxy is a reverse proxy that accepts client certificates and forwards Common Name and Organizations to backend API servers in HTTP headers. The proxy terminates TLS and communicates with KCP and virtual workspace servers via mTLS. Traffic is routed based on paths.

The intent is to deploy the proxy in the same pod as KCP and virtual workspace server pairs.

### DIFF
--- a/hack/gen-certs.sh
+++ b/hack/gen-certs.sh
@@ -1,0 +1,69 @@
+#!/usr/bin/env bash
+
+# Copyright 2022 The KCP Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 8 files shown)


hack/gen-certs.sh
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
```
stevekuznetsov, line 1: We have all of this logic packaged into nice Go libraries already in k8s - might we use that instead of Bash?

csams: Sure, we can use that if you think it would be better. I wrote it in bash because `gen-certs.sh` is just a convenience script for local testing. We'd use cert manager in a deployment.

sttts:
Let's ship it. PR to switch to SecureServing is pending.

sttts (approved):

PR #636, opened 2022-03-06
e2e: Switch to test-managed syncer and update syncer-dependent tests to use shared fixture

marun:
Switch to test-managed syncer to enable syncer-requiring tests to use shared fixture.

PR #662, opened 2022-03-09
cli profiling

aojea:
Checking the e2e tests I've realized that it takes an average of 10 seconds to get the kcp server running, that sounds excessive for a local test.

I was trying to understand why it takes so long, but despite we can expose a profile endpoint, that is not enough to trace and profile the startup process, so I've added some flags to be able to run the profiles and traces directly from the CI (based on the `go test` flags)

```
bin/kcp start --memprofile=./mem.out --cpuprofile cpu.out --trace trace.out
```

As a result, we can observe that there is a lot of cpu used for crypto things 

```
 go tool pprof cpu.out
File: kcp
Type: cpu
Time: Mar 9, 2022 at 11:01am (CET)
Duration: 13.91s, Total samples = 4.76s (34.21%)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 3160ms, 66.39% of 4760ms total
Dropped 603 nodes (cum <= 23.80ms)
Showing top 10 nodes out of 346
      flat  flat%   sum%        cum   cum%
    1200ms 25.21% 25.21%     1240ms 26.05%  crypto/elliptic/internal/fiat.p521Mul
     650ms 13.66% 38.87%      650ms 13.66%  math/big.addMulVVW
     420ms  8.82% 47.69%      450ms  9.45%  crypto/elliptic/internal/fiat.p521Square
     220ms  4.62% 52.31%      870ms 18.28%  math/big.nat.montgomery
     120ms  2.52% 54.83%      130ms  2.73%  crypto/elliptic/internal/fiat.p521Add
     120ms  2.52% 57.35%      340ms  7.14%  runtime.pcvalue
     120ms  2.52% 59.87%      160ms  3.36%  runtime.step
     110ms  2.31% 62.18%      300ms  6.30%  runtime.scanobject
     100ms  2.10% 64.29%      480ms 10.08%  runtime.gentraceback
     100ms  2.10% 66.39%      180ms  3.78%  syscall.Syscall
```

The trace shows that at least 1 seconds is needed for generating the certificates ( can we save some time on the e2e with custom certificates as we are doing in k/k?)

path36485

I'm not very familiar with the cli flags, the code as is is ugly, I just wanted to check that it works,  but if there is interest I will make it prettier

### DIFF
--- a/cmd/kcp/kcp.go
+++ b/cmd/kcp/kcp.go
@@ -19,6 +19,10 @@ package main
 import (
 	"fmt"
 	"os"
+	"path/filepath"
+	"runtime"
+	"runtime/pprof"
+	"runtime/trace"
 
 	"github.com/spf13/cobra"
 
@@ -36,6 +40,19 @@ import (
 
 func main() {
 	help.FitTerminal()
+
+	// cli flags for profiling
+	var (
+		outputDir            string
+		memProfile           string
+		memProfileRate       int
+		cpuProfile           string
+		blockProfile         string
+		blockProfileRate     int
+		mutexProfile         string
+		mutexProfileFraction int
+		traceFile            string
+	)
 
 	cmd := &cobra.Command{
 		Use:   "kcp",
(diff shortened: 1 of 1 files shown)


marun:
@aojea While I appreciate your initiative, I'm not sure we want to worry too much about the cost of server startup in the near term. Operationally, a kcp server is likely to be long-running and features and reliability would seem more important in the immediate future than an optimized startup time.

That said, I share your instinct that it should be possible to minimize the the impact of server startup on e2e execution. Rather than worry about the server, I'm working to ensure that:

 - e2e tests can use shared fixture if possible
 - shared fixture can be persistent

The PR enabling persistent shared fixture for some tests has already merged: https://github.com/kcp-dev/kcp/pull/614 
A follow-on PR (still WIP) enables more e2e tests to use shared fixture: https://github.com/kcp-dev/kcp/pull/636

Running tests with `--kubeconfig=/path/to/.kcp/admin.kubeconfig` removes the 10s startup cost for compatible tests and makes it easier to interacting with the kcp server under test. The catch is that I'm seeing performance degradation when too many tests are hitting the same server. I'm not yet sure if this is something on the client or server side. Would you be interested in exploring this problem with me?

aojea:
> @aojea While I appreciate your initiative, I'm not sure we want to worry too much about the cost of server startup in the near term. Operationally, a kcp server is likely to be long-running and features and reliability would seem more important in the immediate future than an optimized startup time.

yeah, you are right, no worries, I'm starting to get familiar with the project so forgive me if some things are duplicates

> 
> That said, I share your instinct that it should be possible to minimize the the impact of server startup on e2e execution. Rather than worry about the server, I'm working to ensure that:
> 
> * e2e tests can use shared fixture if possible
> * shared fixture can be persistent
> 
> The PR enabling persistent shared fixture for some tests has already merged: #614 A follow-on PR (still WIP) enables more e2e tests to use shared fixture: #636

awesome

> Running tests with `--kubeconfig=/path/to/.kcp/admin.kubeconfig` removes the 10s startup cost for compatible tests and makes it easier to interacting with the kcp server under test. The catch is that I'm seeing performance degradation when too many tests are hitting the same server. I'm not yet sure if this is something on the client or server side. Would you be interested in exploring this problem with me?

glad to help,

marun:
@sttts Would what is proposed here be useful beyond e2e optimization?

PR #663, opened 2022-03-09
kubectl-kcp: fix logic to derive the orgClusterName

### LINKED ISSUE #630: In root workspace, kcp cli plugin does not understand current workspace
While in root workspace:

```
$ kubectl kcp workspace current 
error: The current context is not a KCP workspace !
```

And then (demo is an org):
```
$ kubectl kcp workspace use demo
$ kubectl kcp workspace current 
Current all workspace is "demo".
error: Scope of the workspace-directory ('all') doesn't match the scope of the workspace-directory server ('https://192.168.178.56:6444/services/workspaces/root:demo/personal').
Cannot check the current workspace existence
```

Also note that the output should rather be `Current all workspace is organization "demo".`

sttts:
`kubectl kcp workspaces` now operates on the following org:

- when in the root, the org is the root
- when in an org, the org is the same org
- when in a non-org, the org is the parent.

Also adding a `workspaces` alias to the subcommand, and removing the confusion `'all'` scope output from current workspace. The latter needs another design session probably, but removing at least removes the confusion for now.

Closes #630.

### DIFF
--- a/pkg/cliplugins/workspace/plugin/kubeconfig.go
+++ b/pkg/cliplugins/workspace/plugin/kubeconfig.go
@@ -32,7 +32,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/clientcmd/api"
 
-	"github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1/helper"
+	tenancyhelpers "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1/helper"
 	tenancyv1beta1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1beta1"
 	tenancyclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
 	workspacecmd "github.com/kcp-dev/kcp/pkg/virtual/framework/cmd"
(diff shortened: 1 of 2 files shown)


davidfestal (approved):

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -108,14 +108,27 @@ func (kc *KubeConfig) ensureWorkspaceDirectoryContextExists(options *Options) (*
...
+				} else if org, _, err := tenancyhelpers.ParseLogicalClusterName(clusterName); err != nil {
+					return nil, fmt.Errorf("unable to parse cluster name %s", clusterName)
+				} else if org == "system:" {
+					return nil, fmt.Errorf("no workspaces are accessible from %s", clusterName)
+				} else if org == tenancyhelpers.RootCluster {
+					// already in an org workspace
+					orgClusterName = clusterName
+				} else {
+					// some other workspace, return org cluster name
+					orgClusterName, err = tenancyhelpers.ParentClusterName(clusterName)
```
davidfestal, line 126:
```suggestion
					orgClusterName, err = tenancyhelpers.ParentClusterName(clusterName)
					if err != nil {
						return nil, err
					}					
```

davidfestal (changes requested):

davidfestal (approved):

PR #666, opened 2022-03-09
server: add --experimental-bind-free-port

sttts:
For race-less port allocation in tests.

### DIFF
--- a/pkg/server/options/options.go
+++ b/pkg/server/options/options.go
@@ -18,6 +18,7 @@ package options
 
 import (
 	"fmt"
+	"net"
 	"os"
 	"path/filepath"
 	"time"
@@ -25,6 +26,7 @@ import (
 	"github.com/spf13/pflag"
 
 	"k8s.io/apimachinery/pkg/util/sets"
+	genericapiserveroptions "k8s.io/apiserver/pkg/server/options"
 	cliflag "k8s.io/component-base/cli/flag"
 	_ "k8s.io/kubernetes/pkg/features"
 	"k8s.io/kubernetes/pkg/genericcontrolplane/options"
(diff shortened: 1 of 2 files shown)


marun:
Awesome! Please fix the lint issue and I'm happy to merge.

pkg/server/options/options.go
```diff
@@ -178,6 +193,14 @@ func (o *Options) Complete() (*CompletedOptions, error) {
 		o.AdminAuthentication.KubeConfigPath = filepath.Join(o.Extra.RootDirectory, o.AdminAuthentication.KubeConfigPath)
 	}
 
+	if o.Extra.ExperimentalBindFreePort {
+		listener, _, err := genericapiserveroptions.CreateListener("tcp", fmt.Sprintf("%s:0", o.GenericControlPlane.SecureServing.BindAddress), net.ListenConfig{})
```
aojea, line 197:
```suggestion
		listener, _, err := genericapiserveroptions.CreateListener("tcp",net.JoinHostPort(o.GenericControlPlane.SecureServing.BindAddress,"0"), net.ListenConfig{})
```

pkg/server/options/options.go
```diff
@@ -77,16 +80,18 @@ func NewOptions() *Options {
...
+			ShardKubeconfigFile:      "",
+			EnableSharding:           false,
+			DiscoveryPollInterval:    60 * time.Second,
+			ExperimentalBindFreePort: false,
 		},
 	}
 
 	// override all the stuff
 	o.GenericControlPlane.SecureServing.ServerCert.CertDirectory = "" // will be completed to be relative to root dir
+	o.GenericControlPlane.SecureServing.Required = false
```
aojea, line 94: does this have any implication, defaulting to false?

sttts: no, it doesn't. It would change the flag description which we don't want. Hence, I don't set it earlier. Maybe worth a go doc.

sttts: added comment

stevekuznetsov:
What was wrong with the race-less implementation we have in test code?

marun (approved):

PR #668, opened 2022-03-09
Syncer object mutators

jmprusi:
Adds a simple interface for registering mutators, that based on the object's GVR,  applies changes at different phases of the sync.

### DIFF
--- a/pkg/syncer/mutators/interface.go
+++ b/pkg/syncer/mutators/interface.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package mutators
 
 import (
--- a/pkg/syncer/mutators/interface.go
+++ b/pkg/syncer/mutators/interface.go
@@ -4,10 +4,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-const originalNameLabel = "syncer.kcp.dev/originalname"
-
 type Mutator interface {
 	ApplySpec(downstreamObj *unstructured.Unstructured) error
 	ApplyStatus(upstreamObj *unstructured.Unstructured) error
-	ApplyDownstreamName(downstreamObj *unstructured.Unstructured) error
 }
--- a/pkg/syncer/mutators/interface.go
+++ b/pkg/syncer/mutators/interface.go
@@ -1,0 +1,13 @@
+package mutators
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+const originalNameLabel = "syncer.kcp.dev/originalname"
+
+type Mutator interface {
+	ApplySpec(downstreamObj *unstructured.Unstructured) error
+	ApplyStatus(upstreamObj *unstructured.Unstructured) error
+	ApplyDownstreamName(downstreamObj *unstructured.Unstructured) error
+}
--- a/pkg/syncer/mutators/interface_test.go
+++ b/pkg/syncer/mutators/interface_test.go
@@ -1,0 +1,107 @@
+/*
+Copyright 2022 The KCP Authors.
+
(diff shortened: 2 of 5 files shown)


jmprusi:
@marun as we talked yesterday, I've moved the name transformations to another PR, and outside the transformers.

https://github.com/kcp-dev/kcp/pull/679

pkg/syncer/mutators/interface_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"testing"
+
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func TestMutators(t *testing.T) {
+	// Register the mutator.
```
marun, line 31: (No action required) Is this comment relevant?

pkg/syncer/mutators/interface_test.go
```diff
@@ -0,0 +1,114 @@
...
+		},
+		Status: appsv1.DeploymentStatus{
+			ObservedGeneration: int64(0),
+		},
+	}
+
+	// Convert to unstructured.
+	syncedObject := &unstructured.Unstructured{}
+	err := convert(&deployment, syncedObject)
+	if err != nil {
```
marun, line 53: (No action required) Maybe prefer the succinctness afforded by testify (e.g. require.NoError)?

marun, line 53: (No action required) Maybe prefer the succinctness afforded by testify/require (e.g. require.NoError)?

pkg/syncer/mutators/interface.go
```diff
@@ -0,0 +1,26 @@
...
+*/
+
+package mutators
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type Mutator interface {
+	ApplySpec(downstreamObj *unstructured.Unstructured) error
```
marun, line 24: If spec and status mutation are orthogonal, might it make sense to just use a function (e.g. `mutate(*unstructured.Unstructured) error`) and define mutators separately for spec and status? 

That would simplify testing - there wouldn't have to be special-case handling of spec vs status. The effective type of a mutator would be determined by which syncer (spec or status) it was registered to. They could still be defined together, but having a mutator for one type of syncer wouldn't require an implementation (even if empty) for the other.

marun, line 24: (No action required) If spec and status mutation are orthogonal, would it make sense to just use a function (e.g. `mutate(*unstructured.Unstructured) error`) and register spec and status mutators separately?

pkg/syncer/syncer.go
```diff
@@ -74,6 +75,7 @@ func StartSyncer(ctx context.Context, upstream, downstream *rest.Config, resourc
 	return nil
 }
 
+type mutatorGvrMap map[schema.GroupVersionResource]mutators.Mutator
```
marun, line 78: (No action required) If the intent is to only have a single mutator per gvr, maybe that suggests an `add` method that can throw an error if one is already defined?

marun, line 78: (No action required) If the intent is to only have a single mutator per gvr, maybe that suggests an `add` method that can throw an error if one is already defined?

pkg/syncer/syncer.go
```diff
@@ -105,6 +108,11 @@ func New(kcpClusterName, pcluster string, fromDiscovery discovery.DiscoveryInter
 		direction:           direction,
 		upstreamClusterName: kcpClusterName,
 		syncerNamespace:     os.Getenv(SyncerNamespaceKey),
+		mutators:            make(mutatorGvrMap),
+	}
+
+	if len(mutators) > 0 {
```
marun, line 114: Why is it necessary to check the input if we are writing the caller and can assure that what is provided is valid?

marun, line 114: Why is it necessary to check the input if we are writing the caller and can assure that what is provided is valid?

pkg/syncer/mutators/interface_test.go
jmprusi: it's a leftover after some changes, removed.

pkg/syncer/syncer.go
jmprusi: As it can be used from outside the package, just making sure to use a set of mutators if you pass those. But I can remove it if not needed.

pkg/syncer/mutators/interface.go
jmprusi: Yes, that's a good point. I can do that.. the initial plan, was to make the Name transformations as part of the ApplySpec (yes.. perhaps not the best name) so we can change that.

jmprusi:
After talking with @marun I will simplify the interface and move this PR into https://github.com/kcp-dev/kcp/pull/670

PR #669, opened 2022-03-09
Add DCO information

ncdc:


pweil- (commented):
Changes LGTM.  Are we going to configure the enforcement of `Signed-off-by:` in commit messages and should we also add a pull request template to reference this?

ncdc:
We don't need sign off per commit - the blanket language in CONTRIBUTING.md covers it. 

I do think a PR template with a comment directing people to the contributing doc would also be good.

ncdc:
Added a very simple PR template for starters.

stevekuznetsov (approved):

pweil- (approved):

Issue #280, opened 2021-12-03
Deploy an Application: SA Kubeconfigs point to KCP


pweil-:
# Use Case: Deploy an unmodified application and make it transparently talk to the kcp apiserver

### Objectives:

- [x] An unmodified controller that needs to manipulate its own CRs (e.g. Tekton) deployed to kcp "just works"
- [x] The user doesn't need to do anything to make the controller running on a pcluster know how to talk to the kcp apiserver

### Details

Deploying an application to KCP should detect workloads that configure SAs and inject a kubeconfig that points them back up to kcp

### Action Items:
- [x] **Research:** make sure we can rewrite a Deployment with custom service account logic and not fight with Kube service account controllers / kubelet.

> Using "automountServiceAccountToken: false" in deployment.spec.template.spec allows us to handle what volumes should be mounted on the PODs, this way we can override the default kube-root-ca.crt or any other names

~~- [ ] #1192
    - [ ] (note from Clayton) The service account token volume projection on a local cluster is by default stripped and replaced with a secret from the logical cluster service account, possibly via a CSI driver (to accomplish rotation)~~
- [x] sync down service accounts referenced by synced resources, possibly deconflicting names (https://github.com/kcp-dev/kcp/pull/679 && https://github.com/kcp-dev/kcp/pull/680)
- [x] sync down service account secrets and rewrite namespace, possibly deconflicting names (https://github.com/kcp-dev/kcp/pull/679 && https://github.com/kcp-dev/kcp/pull/680)
- [x] Make the rootCACertPublisher workspace aware (merge: https://github.com/kcp-dev/kubernetes/pull/49)
- [x] Wire the rootCACertPublisher into KCP (merge: https://github.com/kcp-dev/kcp/pull/667)
- [x] sync down `kube-root-ca.crt` ConfigMap, and deconflict name (https://github.com/kcp-dev/kcp/pull/679 && https://github.com/kcp-dev/kcp/pull/680 )
- [x] Ensure to sync only the configmap/secret/servicetoken when referenced by a deployment.  (https://github.com/kcp-dev/kcp/pull/680) ---> pushed to p4 (https://github.com/kcp-dev/kcp/issues/727)
- [x] implement transformation of Deployments pointing to the right service account in the pcluster (related: https://github.com/kcp-dev/kcp/pull/668 https://github.com/kcp-dev/kcp/pull/670)
[...]

PR #670, opened 2022-03-09
Make Pods from Synced Deployments in a pCluster connect to KCP when using inCluster config.

jmprusi:
This PR adds makes the inClusterConfig of any PODs (that pertain to a synced Deployment) point to the KCP api-server instead of the pcluster api-server where it is scheduled.

This is done via a deployment mutator that handles the mounting of the kcp root cert and default tokens into the proper path, and also overrides the KUBERNETES_SERVICE_XXXX  env vars that are required by inClusterConfig. 

related to https://github.com/kcp-dev/kcp/issues/280


pkg/syncer/mutators/deployment.go
```diff
@@ -0,0 +1,234 @@
...
+}
+func int64ptr(i int64) *int64 {
+	return &i
+}
+
+func int32ptr(i int32) *int32 {
+	return &i
+}
+
+func boolPtr(b bool) *bool {
```
sttts, line 232: these exist in k8s.io/utils

pkg/server/handler.go
```diff
@@ -91,6 +94,24 @@ func WithClusterScope(apiHandler http.Handler) http.HandlerFunc {
...
+				return
+			}
+			if resp != nil {
+				extraInfo := resp.User.GetExtra()
+				if val, ok := extraInfo[serviceaccount.ClusterNameKey]; ok {
+					klog.Infof("Set cluster name from request to: %q", val[0])
+					clusterName = val[0]
+				}
+			}
+		}
```
sttts, line 114: this is the wrong spot. We have to make the service account on the pcluster side to talk to the right server URL.

pkg/syncer/mutators/deployment.go
```diff
@@ -0,0 +1,247 @@
...
+		Version:  "v1",
+		Resource: "deployments",
+	}
+}
+
+func (dt *DeploymentMutator) Register(mutators map[schema.GroupVersionResource]Mutator) {
+	if _, ok := mutators[dt.getGVR()]; !ok {
+		mutators[dt.getGVR()] = dt
+	}
+}
```
sttts, line 33: this is an odd pattern. Why should a mutator know how to register into some map of mutators?

pkg/syncer/mutators/interface_test.go
```diff
@@ -0,0 +1,96 @@
+/*
```
sttts, line 1: wrong filename?

pkg/syncer/syncer.go
```diff
@@ -453,6 +461,12 @@ func transformName(syncedObject *unstructured.Unstructured, direction SyncDirect
 		if syncedObject.GetKind() == "ServiceAccount" && syncedObject.GetName() == "default" {
 			syncedObject.SetName("kcp-default")
 		}
+		// TODO(jmprusi): We are rewriting the name of the object into a non random one so we can reference it from the deployment transformer
+		//                but this means that means than more than one default-token-XXXX object will overwrite the same "kcp-default-token"
+		//				  object. This must be fixed.
+		if syncedObject.GetKind() == "Secret" && strings.Contains(syncedObject.GetName(), "default-token-") {
```
sttts, line 467: make all of these fully-qualified group-version-kind comparisons. There could be foo/v1 Secrets too.

cmd/syncer/options/options.go
```diff
@@ -52,7 +53,7 @@ func (options *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&options.PclusterID, "workload-cluster-name", options.PclusterID,
 		fmt.Sprintf("ID of the -to cluster. Resources with this ID set in the '%s' label will be synced.", nscontroller.ClusterLabel))
 	fs.StringArrayVarP(&options.SyncedResourceTypes, "sync-resources", "r", options.SyncedResourceTypes, "Resources to be synchronized in kcp.")
-
+	fs.StringVar(&options.KCPHost, "kcp-host", options.KCPHost, "Hostname of the kcp server (url:port) if not set, defaults to the FromKubeConfig host.")
```
sttts, line 56: is this format validated? Does url imply `https://`?

full-stop after the parenthesis.

pkg/reconciler/cluster/syncer/pushmanager.go
```diff
@@ -38,11 +38,13 @@ const numSyncerThreads = 2
 
 type pushSyncerManager struct {
 	syncerCancelFuncs map[string]func()
+	kcpHost           string
 }
 
-func newPushSyncerManager() SyncerManager {
+func newPushSyncerManager(KCPHost string) SyncerManager {
```
sttts, line 44: kcpHost

pkg/reconciler/cluster/syncer/syncer_controller_start.go
```diff
@@ -37,6 +38,7 @@ func BindOptions(o *Options, fs *pflag.FlagSet) *Options {
 	fs.StringVar(&o.SyncerImage, "syncer-image", o.SyncerImage, "Syncer image to install on clusters")
 	fs.BoolVar(&o.PullMode, "pull-mode", o.PullMode, "Deploy the syncer in registered physical clusters in POD, and have it sync resources from KCP")
 	fs.BoolVar(&o.PushMode, "push-mode", o.PushMode, "If true, run syncer for each cluster from inside cluster controller")
+	fs.StringVar(&o.KCPHost, "kcp-host", o.KCPHost, "The KCP Api-server host in host:port format.")
```
sttts, line 41: API server

Wondering, "host" suggests that this is a hostname. But it isn't. It's an https URL, no? Maybe `--kcp-address` or `--kcp-url`? @marun opinions?

pkg/syncer/mutators/deployment.go
```diff
@@ -0,0 +1,247 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/rest"
+	utilspointer "k8s.io/utils/pointer"
+)
+
+type DeploymentMutator struct {
```
marun, line 16: Hmm, what happened to the idea of mutators being a function and not a type? Fine to special-case instantiation but that doesn't require special-casing registration or use e.g.

```golang
deploymentMutator := NewDeploymentMutator(...)
c.registerMutator(deploymentMutator.GVR, deploymentMutator.Mutate)
c.registerMutator(otherGVR, func(u *unstructured.Unstructured) error {
   // Do my thing
})
```

pkg/syncer/mutators/deployment.go
```diff
@@ -0,0 +1,247 @@
...
+func (dt *DeploymentMutator) Mutate(downstreamObj *unstructured.Unstructured) error {
+	var deployment appsv1.Deployment
+	err := runtime.DefaultUnstructuredConverter.FromUnstructured(
+		downstreamObj.UnstructuredContent(),
+		&deployment)
+	if err != nil {
+		return err
+	}
+
+	// If the deployment has not serviceaccount defined or it is "default", that means that it is using the default one
```
marun, line 52: nit: `s/not serviceaccount/no serviceaccount/`

pkg/syncer/mutators/deployment.go
```diff
@@ -0,0 +1,247 @@
...
+	// and we need will need to override it with our own, "kcp-default"
+	//
+	// If the deployment has a serviceaccount defined, that means we need to synchronize the created service account
+	// down to the workloadcluster, so we don't modify it, and expect the scheduler to do the job.
+	//
+	if deployment.Spec.Template.Spec.ServiceAccountName == "" || deployment.Spec.Template.Spec.ServiceAccountName == "default" {
```
marun, line 58: (No action required): So many dots! Maybe alias the template spec to simplify things?

```golang
templateSpec := deployment.Spec.Template.Spec
```

pkg/syncer/mutators/deployment.go
```diff
@@ -0,0 +1,247 @@
...
+
+func NewDeploymentMutator(fromConfig *rest.Config, kcpHost string) *DeploymentMutator {
+	return &DeploymentMutator{
+		fromConfig: fromConfig,
+		kcpHost:    kcpHost,
+	}
+}
+
+// ApplySpec makes modifications to the Spec of the deployment object.
+func (dt *DeploymentMutator) Mutate(downstreamObj *unstructured.Unstructured) error {
```
marun, line 43: (No action required) What does `dt` mean? Maybe use `dm` or even `m` instead for clarity?

pkg/syncer/mutators/deployment.go
```diff
@@ -0,0 +1,247 @@
...
+	deployment.Spec.Template.Spec.AutomountServiceAccountToken = utilspointer.BoolPtr(false)
+
+	var hostname, port string
+	if dt.kcpHost != "" {
+		hostname, port, err = net.SplitHostPort(dt.kcpHost)
+		if err != nil {
+			return err
+		}
+	} else {
+		u, err := url.Parse(dt.fromConfig.Host)
```
marun, line 73: (No action required) Maybe limit single-char var names to loop index variables and receivers?

pkg/syncer/mutators/deployment.go
```diff
@@ -0,0 +1,247 @@
...
+		if err != nil {
+			return err
+		}
+		hostname = u.Hostname()
+		port = u.Port()
+	}
+
+	kcpExternalPort := port
```
marun, line 81: nit: Maybe just use these variable names in the first place?

pkg/syncer/mutators/deployment.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// This is the VolumeMount that we will append to all the containers of the deployment
+	serviceAccountMount := corev1.VolumeMount{
+		Name:      "kcp-api-access",
+		MountPath: "/var/run/secrets/kubernetes.io/serviceaccount",
+		ReadOnly:  true,
+	}
+
+	// This is the Volume that we will add to the Deployment in order to control
+	// the name of the ca.crt references (kcp-root-ca.crt vs kube-root-ca.crt)
```
marun, line 98: Maybe add a bit more detail as to why this renaming is necessary (e.g. kube already writes its own ca secret so we need to differentiate the one for kcp)?

pkg/syncer/mutators/deployment.go
```diff
@@ -0,0 +1,247 @@
...
+	// and the serviceaccount reference.
+	serviceAccountVolume := corev1.Volume{
+		Name: "kcp-api-access",
+		VolumeSource: corev1.VolumeSource{
+			Projected: &corev1.ProjectedVolumeSource{
+				DefaultMode: utilspointer.Int32Ptr(420),
+				Sources: []corev1.VolumeProjection{
+					{
+						// TODO(jmprusi): Instead of using a secret directly we should use the serviceaccount
```
marun, line 107: Do we actually want to do this? 

Bound tokens need to be refreshed periodically, which is something that the kubelet is guaranteed sufficient permission to do on behalf of a pod. Using that same model for a kcp-dependent pod would require a syncer component to have the same reliability of the kubelet. Except, if a kubelet fails to do its job, the pod goes unhealthy and is restarted somewhere else. If a syncer stops running for whatever reason, would it be preferable to allow the workload to keep running or have it die if its token isn't refreshed in a timely manner?

pkg/syncer/mutators/deployment.go
```diff
@@ -0,0 +1,247 @@
...
+		return err
+	}
+
+	// Set the changes back into the obj.
+	downstreamObj.SetUnstructuredContent(unstructured)
+
+	return nil
+}
+
+//findEnv finds an env in a list of envs
```
marun, line 205: nit: Maybe include a leading space for godoc comments for consistency?

pkg/syncer/mutators/interface_test.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	// Apply mutators to the object.
+	err = tdm.Mutate(syncedObject)
+	if err != nil {
+		t.Errorf("Mutate() = %v", err)
+	}
+
+	// Convert the unstructured object back to a deployment.
+	err = runtime.DefaultUnstructuredConverter.FromUnstructured(syncedObject.UnstructuredContent(), &deployment)
```
marun, line 64: (No action required) Maybe separate out the conversion to/from unstructured from the mutation function to simplify testing? 

```golang
def mutate(obj *unstructured.Unstructured) error {
  deployment := unmarshal(...)
  err := updateDeployment(deployment)
  if err != nil {
   return err
  } 
  runtime.DefaultUnstructuredConverter.FromUnstructured(obj.UnstructuredContent(), deployment)
}
```

And then target `updateDeployment` for testing of mutation changes.

pkg/syncer/mutators/secrets.go
```diff
@@ -0,0 +1,68 @@
...
+		&secret)
+	if err != nil {
+		return err
+	}
+
+	// We need to transform the default kcp token, in order to avoid the pcluster to rewrite it.
+	if secret.Name == "kcp-default-token" {
+		secret.Type = corev1.SecretTypeOpaque
+		if strings.Contains(secret.Name, "kcp-default-token") {
+			// Override with the default kcp serviceaccount
```
marun, line 51: Why are we setting these annotations if they are managed by the pcluster controller? What purpose do they serve?

pkg/syncer/mutators/interface_test.go
jmprusi: please review the new deployment_test file.

cmd/syncer/options/options.go
aojea: why do we want to have an override of the kubeconfig host field and not set it correctly the first time?

jmprusi: When the Syncer is in push mode, the From address is "localhost", that's not going to work when trying to connect back to KCP from a downstream POD.

aojea: then this is my gap, I was thinking that in push mode the user pass the kubeconfig file to the syncer ... how is this kubeconfig passed to the syncer then? 
The key is to know how this  kubeconfig is generated and why we can not pass the right url there :)

cmd/syncer/options/options.go
```diff
@@ -52,7 +54,7 @@ func (options *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&options.PclusterID, "workload-cluster-name", options.PclusterID,
 		fmt.Sprintf("ID of the -to cluster. Resources with this ID set in the '%s' label will be synced.", nscontroller.ClusterLabel))
 	fs.StringArrayVarP(&options.SyncedResourceTypes, "sync-resources", "r", options.SyncedResourceTypes, "Resources to be synchronized in kcp.")
-
+	fs.StringVar(&options.KCPHost, "kcp-host", options.KCPHost, "Hostname of the kcp server (host:port). If not set, defaults to the --from-kubeconfig cluster URL.")
```
sttts, line 57: Credits to @aojea for the question: do we know why we need this flag?

My (maybe naive) assumption: every erver URL good for the syncer in the provided `--from-kubeconfig` is good enough for every other pods.

cmd/syncer/options/options.go
marun: FYI we're not intending to support 'push mode' going forward.

aojea: I use to mix both modes, I'm trying to say the one that the syncer connects to kcp, is that pull?

marun: - 'push': the syncer for a given cluster runs as part of the kcp server process
- 'pull': the syncer for a given cluster runs external to kcp

We're not intending for kcp to keep running syncers because we don't want to depend on workload clusters being network accessible to kcp.

aojea: > When the Syncer is in push mode, the From address is "localhost", that's not going to work when trying to connect back to KCP from a downstream POD.

if the pod has hostNetwork:true it will work

marun: So syncer will need to run with `hostNetwork` to function? 😄

In seriousness, I don't think we want this arg. The kcp kubeconfig will need to be valid for a syncer running out-of-process.

jmprusi: I agree, we can get rid of this Arg, what do you think, if instead of setting 127.0.0.1 by default on the kubeconfig used by the syncer, we use the Binding address?

marun: I think that's fine for this PR. And whatever mechanism creates syncer yaml will need to know how to set that externally-accessible address.

pkg/server/server.go
```diff
@@ -382,8 +382,11 @@ func (s *Server) Run(ctx context.Context) error {
 			return err
 		}
 
-		// TODO(sttts): this is a hack, using the loopback config as a blueprint. Syncer should never use a loopback connection.
-		if err := s.installSyncerController(ctx, controllerConfig, CreateLoopbackUpstreamKubeConfig(server)); err != nil {
+		syncerConfig, err := s.options.AdminAuthentication.GetPushModeSyncerKubeconfig(genericConfig, newTokenOrEmpty, tokenHash, s.options.GenericControlPlane.GenericServerRunOptions.ExternalHost, servingOpts.BindPort)
```
aojea, line 385: I think that ExternalAddress has logic to get always the right IP, 
```
	// ExternalAddress is the host name to use for external (public internet) facing URLs (e.g. Swagger)
	// Will default to a value based on secure serving info and available ipv4 IPs.
	ExternalAddress string
```

externalHost depends on be populated IIRC
```
staging/src/k8s.io/apiserver/pkg/server/options/server_run_options.go:  ExternalHost                string
staging/src/k8s.io/apiserver/pkg/server/options/server_run_options.go:  c.ExternalAddress = s.ExternalHost
staging/src/k8s.io/apiserver/pkg/server/options/server_run_options.go:  fs.StringVar(&s.ExternalHost, "external-hostname", s.ExternalHost,
```

jmprusi: It iss already being used to generate the kubeconfig written to disk, so It has to be set.

pkg/server/options/authentication.go
```diff
@@ -107,6 +107,18 @@ func (s *AdminAuthentication) ApplyTo(config *genericapiserver.Config) (newToken
 	return newTokenOrEmpty, tokenHash, nil
 }
 
+func (s *AdminAuthentication) GetPushModeSyncerKubeconfig(config *genericapiserver.Config, newToken string, tokenHash []byte, externalAddress string, servingPort int) (*clientcmdapi.Config, error) {
```
sttts, line 112: `// TODO: remove when push mode is gone`

pkg/server/options/authentication.go
```diff
@@ -107,6 +107,18 @@ func (s *AdminAuthentication) ApplyTo(config *genericapiserver.Config) (newToken
...
+	externalCACert, _ := config.SecureServing.Cert.CurrentCertKeyContent()
+	externalKubeConfigHost := fmt.Sprintf("https://%s:%d", externalAddress, servingPort)
+
+	externalAdminUserName := "admin"
+	if newToken == "" {
+		return nil, fmt.Errorf("cannot create the PushModeSyncer kubeConfig with an empty token for the %s user", externalAdminUserName)
```
sttts, line 118: which means this break on server restart.

jmprusi: So, that's the reason it reads from the kubeconfig file...

pkg/server/options/authentication.go
```diff
@@ -107,6 +107,18 @@ func (s *AdminAuthentication) ApplyTo(config *genericapiserver.Config) (newToken
 	return newTokenOrEmpty, tokenHash, nil
 }
 
+func (s *AdminAuthentication) GetPushModeSyncerKubeconfig(config *genericapiserver.Config, newToken string, tokenHash []byte, externalAddress string, servingPort int) (*clientcmdapi.Config, error) {
+	externalCACert, _ := config.SecureServing.Cert.CurrentCertKeyContent()
+	externalKubeConfigHost := fmt.Sprintf("https://%s:%d", externalAddress, servingPort)
```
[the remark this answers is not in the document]
jmprusi: fixed! thanks.

pkg/syncer/mutators/deployment.go
jmprusi: done, thanks.

pkg/syncer/mutators/deployment.go
jmprusi: leftover from when it was a "DeploymentTransformer".  fixed!

pkg/syncer/mutators/deployment.go
jmprusi: Changed it, please take a look..

pkg/syncer/syncer.go
```diff
@@ -453,6 +461,12 @@ func transformName(syncedObject *unstructured.Unstructured, direction SyncDirect
 		if syncedObject.GetKind() == "ServiceAccount" && syncedObject.GetName() == "default" {
 			syncedObject.SetName("kcp-default")
 		}
+		// TODO(jmprusi): We are rewriting the name of the object into a non random one so we can reference it from the deployment transformer
```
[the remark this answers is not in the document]
sttts: @jmprusi can you move that to a follow-up task to investigage? For p4.

pkg/syncer/syncer.go
```diff
@@ -445,20 +453,41 @@ func (c *Controller) process(ctx context.Context, h holder) error {
...
 }
+
+func getDefaultMutators(from *rest.Config) mutatorGvrMap {
+	mutatorsMap := make(mutatorGvrMap)
+
+	deploymentMutator := mutators.NewDeploymentMutator(from)
+	secretMutator := mutators.NewSecretMutator()
+
+	mutatorsMap[deploymentMutator.GVR()] = deploymentMutator.Mutate
+	mutatorsMap[secretMutator.GVR()] = secretMutator.Mutate
```
sttts, line 491: nit: why the multiple steps, just `mutatorsMap[deploymentMutator.GVR()] = mutators.NewDeploymentMutator(from)`

sttts (approved):

pkg/syncer/syncer.go
jmprusi: deploymentMutator.GVR() is a method of the deployment mutator, so I need to instantiate it first, I can change it to a single line, but readability suffers a little bit:

```
mutatorsMap[mutators.NewDeploymentMutator(from).GVR()] = mutators.NewDeploymentMutator(from).Mutate
```

marun: (No action required) Maybe GVR doesn't have to be provided by a mutator type method? It's static data, it could just be a var in the same namespace (e.g. DeploymentGVR).

pkg/syncer/mutators/deployment.go
```diff
@@ -0,0 +1,242 @@
...
+		return err
+	}
+
+	// If the deployment has no serviceaccount defined or it is "default", that means that it is using the default one
+	// and we need will need to override it with our own, "kcp-default"
+	//
+	// If the deployment has a serviceaccount defined, that means we need to synchronize the created service account
+	// down to the workloadcluster, so we don't modify it, and expect the scheduler to do the job.
+	//
+	// Alias for the template.spec to improve readability.
```
marun, line 65: (No action required) Maybe this comment is redundant?

marun (approved):

pkg/server/handler.go
```diff
@@ -156,6 +159,44 @@ func WithWildcardListWatchGuard(apiHandler http.Handler) http.HandlerFunc {
...
+
+		if strings.HasPrefix(req.RequestURI, "/clusters/") {
+			handler.ServeHTTP(w, req)
+			return
+		}
+
+		// attempt to authenticate service account JWT
+		clone := utilnet.CloneRequest(req)
+		resp, ok, err := unsafeServiceAccountPreAuth.AuthenticateRequest(clone)
```
ncdc, line 182: I'm curious why we're cloning the request here? Does AuthenticateRequest modify it? Or were we intending to modify the clone below, when we adjust the path and requestURI?

jmprusi: Yes, authenticateRequest removes the token header.

ncdc: Thanks - maybe add a comment in a follow-up?

PR #640, opened 2022-03-07
virtual-workspaces: wire into kcp server and remove topology leaking to cli plugin

sttts:
Adds
```
--run-virtual-workspaces=true/false # Run the virtual workspace apiservers in-process (defaults to true!)
--virtual-workspace-address=<url> # Address of a stand-alone virtual workspace apiservers
--virtual-workspaces-workspaces-base-path # The prefix of the workspaces API server root path.
```
to `kcp start`.

Note: `virtual-workspaces-workspaces-base-path` is not a typo 🤷‍♂️ 

In case of `--run-virtual-workspaces=false` it requires `--virtual-workspace-address` to set a redirect (301) URL. Obviously in a production environment with uniform URLs, this can be taken over by a front-proxy (cc @kylape @csams).

Among the in-process run mode, this also closes https://github.com/kcp-dev/kcp/issues/638.

TODOs:
- [ ] adapt demo to use in-process mode
- [ ] adapt services deployment @csams @kylape

PR #673, opened 2022-03-09
Better scope management and CLI plugin cleanup

davidfestal:
- Cleanup the CLI `workspaces` plugin to:
  - remove completely the `workspace-directory` overrides
  - use the same hostname and port as the current context by default (an alternate port can still be specified through a CLI if necessary, especially if this PR is merged before #640 )  
  - have the scope be specified as a CLI arg
- Fixed an error when getting the `current` workspace on a non-org workspace
- Added (quick and dirty fix) the support of the `exec` user authentication method on the client-side.

### DIFF
--- a/pkg/cliplugins/workspace/cmd/cmd.go
+++ b/pkg/cliplugins/workspace/cmd/cmd.go
@@ -45,7 +45,7 @@ func NewCmdWorkspace(streams genericclioptions.IOStreams) (*cobra.Command, error
 
 	cmd := &cobra.Command{
 		Aliases:          []string{"ws", "workspaces"},
-		Use:              "workspace [--workspace-directory-server=] <current|use|list>",
+		Use:              "workspace [--scope=<personal|all>] <create|list|use|current|delete>",
 		Short:            "Manages KCP workspaces",
 		Example:          fmt.Sprintf(workspaceExample, "kubectl kcp"),
 		SilenceUsage:     true,
@@ -53,10 +53,6 @@ func NewCmdWorkspace(streams genericclioptions.IOStreams) (*cobra.Command, error
 	}
 
 	opts.BindFlags(cmd)
-	kubeconfig, err := plugin.NewKubeConfig(opts)
-	if err != nil {
-		return nil, err
-	}
 
 	useCmd := &cobra.Command{
 		Use:          "use < workspace name | - >",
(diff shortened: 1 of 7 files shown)


pkg/cliplugins/workspace/cmd/cmd.go
```diff
@@ -45,25 +45,25 @@ func NewCmdWorkspace(streams genericclioptions.IOStreams) (*cobra.Command, error
...
-		return nil, err
-	}
 
 	useCmd := &cobra.Command{
 		Use:          "use < workspace name | - >",
 		Short:        "Uses the given workspace as the current workspace. Using - means previous workspace",
 		Example:      "kcp workspace use my-worspace",
 		SilenceUsage: true,
 		RunE: func(c *cobra.Command, args []string) error {
+			kubeconfig, err := plugin.NewKubeConfig(opts)
```
marun, line 63: (No action required) What's the point of in-lining this if the argument it depends on is provided outside the function scope?

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -76,104 +82,94 @@ func NewKubeConfig(overridingOptions *Options) (*KubeConfig, error) {
...
+	if kc.port != 0 {
+		port = fmt.Sprintf(":%d", kc.port)
+	} else {
+		if currentServerURL.Port() != "" {
+			port = fmt.Sprintf(":%s", currentServerURL.Port())
 		}
-		workspaceDirectoryAwareConfig.Clusters[workspaceDirectoryOverrides.CurrentContext] = workspaceDirectoryCluster
-		workspaceDirectoryAwareConfig.Contexts[workspaceDirectoryOverrides.CurrentContext] = workspaceDirectoryContext
 	}
+	workspaceDirectoryCluster.Server = fmt.Sprintf("%s://%s%s%s/%s/%s", currentServerURL.Scheme, currentServerURL.Hostname(), port, workspacebuilder.DefaultRootPathPrefix, orgClusterName, kc.scope)
```
marun, line 136: (No action required) Maybe use `net.JoinHostPort` as per https://github.com/kcp-dev/kcp/pull/666#discussion_r822951636

marun:
Are there (or could there be) unit tests for this code?

pkg/cliplugins/workspace/cmd/cmd.go
ncdc: `opts` fields have their values changed when the command line flags are parsed

PR #679, opened 2022-03-10
Handle name transformation to avoid default object conflicts

jmprusi:
Transform the object names to avoid upstream/downstream name conflicts while keeping the kubernetes conventions on naming.

* For `serviceaccounts`: when syncing a default sa into a pcluster, prefix it with "kcp-", when syncing back to kcp remove the "kcp-" prefix.

* For `configmaps`: when syncing the default "kube-root-ca.crt" configmap to a pcluster, rename it to "kcp-root-ca.crt" to avoid any conflicts, when syncing back to kcp, rename it to "kube-root-ca.crt"

related to https://github.com/kcp-dev/kcp/issues/280


pkg/syncer/syncer.go
```diff
@@ -431,3 +431,27 @@ func (c *Controller) process(ctx context.Context, h holder) error {
 
 	return err
 }
+
+// transformName changes the object name into the desired one based on the Direction:
+// - if the object is a configmap it handles the "kube-root-ca.crt" name mapping
+// - if the object is a serviceaccount it handles the "default" name mapping
+func transformName(syncedObject *unstructured.Unstructured, direction Direction) {
+
```
sttts, line 439: nit: drop empty line

PR #681, opened 2022-03-10
Start service account controller and enable legacy+bound service account token authn+authz

sttts:
Depends on https://github.com/kcp-dev/kubernetes/pull/51.

- start service account and service account tokenn controllers (credits to @marun)
- auto-generate service account key+cert to `.kcp/sa.{key,crt}` (credits to @ncdc; 💪  pair programming with @s-urbaniak)
- enable service account authentication (💪  pair programming with @s-urbaniak)
- give `access` level (= authenticated) permission for service accounts to their own workspace (💪  pair programming with @s-urbaniak)
- add service account e2e tests for legacy+bound tokens (credits to @marun for the former; 💪  pair programming with @s-urbaniak))

@ncdc this does not yet change the namespace indexer keys. So we depend on the objects per bucket to be unique. But this is the case luckily. But clearly a list will return something wrong.


pkg/server/options/controllers.go
```diff
@@ -55,6 +78,33 @@ func (c *Controllers) AddFlags(fs *pflag.FlagSet) {
...
+}
+
+func (c *Controllers) Complete(rootDir string) error {
+	if c.SAController.ServiceAccountKeyFile == "" {
+		// use sa.key and auto-generate if not existing
+		c.SAController.ServiceAccountKeyFile = filepath.Join(rootDir, "sa.key")
+		if _, err := os.Stat(c.SAController.ServiceAccountKeyFile); os.IsNotExist(err) {
+			klog.Infof("Generating service account key file %s", c.SAController.ServiceAccountKeyFile)
+			// TODO(ncdc): should alrogithm & key size be configurable?
+			key, err := rsa.GenerateKey(cryptorand.Reader, 2048)
```
marun, line 92: (No action required) Maybe update this to 4096 and remove the comment above? I think rsa+4096 should be good enough until we learn otherwise.

pkg/server/options/options.go
```diff
@@ -100,9 +100,10 @@ func NewOptions() *Options {
 		WithClientCert().
 		WithOIDC().
 		WithRequestHeader().
-		// WithServiceAccounts().
+		WithServiceAccounts().
 		WithTokenFile()
 	//WithWebHook()
+	o.GenericControlPlane.Authentication.ServiceAccounts.Issuers = []string{"https://kcp.default.svc"}
```
marun, line 106: The kube default points to the apiserver service that is always reachable within a cluster, but with kcp lacking services it's not clear we have a good default. Maybe that suggests we should not be enabling bound tokens without an explicit issuer being configured? Or maybe we should provide a default that is not a url (if possible - maybe validation precludes this) to make it clear that the bound tokens issued by default are not intended to be verified via an oidc endpoint?

test/e2e/framework/kcp.go
```diff
@@ -98,6 +99,16 @@ func newKcpServer(t *testing.T, cfg KcpConfig, artifactDir, dataDir string) (*kc
 	if err := os.MkdirAll(dataDir, 0755); err != nil {
 		return nil, fmt.Errorf("could not create data dir: %w", err)
 	}
+
+	saPrivateKey, err := GenerateRSAPrivateKey()
```
marun, line 103: Given that a default key is being set (and this test code isn't even supplying the result of generation as an argument to the server), this change and the rsa.go addition should likely be dropped.

test/e2e/authorizer/serviceaccounts_test.go
```diff
@@ -0,0 +1,199 @@
...
+		for _, secret := range secrets.Items {
+			if secret.Annotations[corev1.ServiceAccountNameKey] == "default" {
+				tokenSecret = secret
+				return true
+			}
+		}
+		return false
+	}, wait.ForeverTestTimeout, time.Millisecond*100, "token secret for default service account not created")
+
+	type TokenTestCase struct {
```
marun, line 129: (No action required) Maybe prefer idiomatic golang test case definition for consistency with existing code? e.g.

```golang
testCases := []struct{
  ...
}{
  ...
}
for _, tc := range testCases {
  ...
}
```

For extra credit, use map[string]struct{} instead of []struct{} to define e2e test cases. The use of a map will randomize execution order which avoids the potential for dependency between tests creeping in.

test/e2e/authorizer/serviceaccounts_test.go
```diff
@@ -0,0 +1,199 @@
...
+						UID:        tokenSecret.UID,
+					},
+				},
+			}, metav1.CreateOptions{})
+			require.NoError(t, err, "failed to create token")
+			return boundToken.Status.Token
+		}},
+	} {
+		t.Run(ttc.name, func(t *testing.T) {
+			saRestConfig, err := server.DefaultConfig()
```
marun, line 156: (No action required) I'm going to update DefaultConfig to accept `t` so it can use require on any errors. Test helpers that require checking of trivial errors is an anti-pattern.

marun (commented):
The new tests are passing, nice work!  I'm not qualified to validate the auth changes, but modulo a couple of inline comments this LGTM.

pkg/server/options/controllers.go
sttts: done

pkg/server/options/options.go
sttts: I would leave it on. We can have multiple in the future if we have to change to some other URL for some reason.

test/e2e/framework/kcp.go
sttts: done

test/e2e/authorizer/serviceaccounts_test.go
sttts: done.

sttts:
@marun addressed you comments. Ptal.

marun (approved):
Nice work, thank you!

PR #682, opened 2022-03-10
Update deployment manifest

kylape:
- Add new flags to kcp and virtual-workspace containers
- Pull external hostname into github actions variable
- Use variable to set external hostname via kustomize
- Update transformation command for virtual workspace kubeconfig

Co-authored-by: Chris Sams <csams@redhat.com>

### DIFF
--- a/manifest/kcp.yaml
+++ b/manifest/kcp.yaml
@@ -105,12 +105,18 @@ spec:
         - --tls-private-key-file=/etc/kcp/tls/server/tls.key
         - --tls-cert-file=/etc/kcp/tls/server/tls.crt
         - --root-directory=/tmp/kcp
+        - --run-virtual-workspaces=false
+        - --virtual-workspace-address=https://$(EXTERNAL_HOSTNAME)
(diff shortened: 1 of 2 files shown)


manifest/kcp.yaml
```diff
@@ -105,12 +105,18 @@ spec:
         - --tls-private-key-file=/etc/kcp/tls/server/tls.key
         - --tls-cert-file=/etc/kcp/tls/server/tls.crt
         - --root-directory=/tmp/kcp
+        - --run-virtual-workspaces=false
+        - --virtual-workspace-address=https://$(EXTERNAL_HOSTNAME)
```
sttts, line 109: no port?

sttts: or is the front proxy doing the magic?

kylape: Since this is exposed using an Openshift route for now, the port (443) is implied.  If we want to make this template more generic for use outside of our shared service, we would need to do more work to generalize.

ncdc (approved):

PR #684, opened 2022-03-10
kcp terminology doc

pkprzekwas:
I took text from the original google doc created by @sttts and moved it here as I found it very helpful while learning about kcp terminology.

### DIFF
--- a/docs/terminology.md
+++ b/docs/terminology.md
@@ -1,47 +1,94 @@
 # Terminology for control plane
-Contains the definitions shared across design documents around prototyping a kube-like control plane (in KCP).  This is a derivative work of other design documents intended to frame terminology.  All future statements that may be changed by designs is covered by those designs, and not duplicated here.
+Contains the definitions shared across design documents around prototyping a kube-like control plane (in KCP).  This is
+a derivative work of other design documents intended to frame terminology.  All future statements that may be changed by
+designs is covered by those designs, and not duplicated here.
 
 ## Logical cluster
-A logical cluster is a way to subdivide a single kube-apiserver + etcd storage into multiple clusters (different APIs, separate semantics for access, policy, and control) without requiring multiple instances.  A logical cluster is a mechanism for achieving separation, but may be modelled differently in different use cases.  A logical cluster is similar to a virtual cluster as defined by sig-multicluster, but is able to amortize the cost of a new cluster to be near-zero memory and storage so that we can create tens of millions of empty clusters cheaply.
+A logical cluster is a way to subdivide a single kube-apiserver + etcd storage into multiple clusters (different APIs,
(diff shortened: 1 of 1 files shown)


docs/terminology.md
```diff
@@ -0,0 +1,47 @@
+# Terminology for control plane
+Contains the definitions shared across design documents around prototyping a kube-like control plane (in KCP).  This is a derivative work of other design documents intended to frame terminology.  All future statements that may be changed by designs is covered by those designs, and not duplicated here.
+
+## Logical cluster
+A logical cluster is a way to subdivide a single kube-apiserver + etcd storage into multiple clusters (different APIs, separate semantics for access, policy, and control) without requiring multiple instances.  A logical cluster is a mechanism for achieving separation, but may be modelled differently in different use cases.  A logical cluster is similar to a virtual cluster as defined by sig-multicluster, but is able to amortize the cost of a new cluster to be near-zero memory and storage so that we can create tens of millions of empty clusters cheaply.
```
stevekuznetsov, line 5: Technically as implemented today there is actually zero cost to a "new empty" logical cluster, since there is no book-keeping...

docs/terminology.md
```diff
@@ -0,0 +1,47 @@
...
+
+A logical cluster is a storage level concept that adds an additional attribute to an object’s identifier on a kube-apiserver.  Regular servers identify objects by (group, version, resource, optional namespace, name).  A logical cluster prepends an identifier: (logical cluster name, group, version, resource, optional namespace, name).
```
stevekuznetsov, line 7: Depending on what is meant by "prepend" this is incorrect, currently at the storage layer (when implemented in etcd) the logical cluster is after the gvr but before ns/name

docs/terminology.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Physical cluster
```
stevekuznetsov, line 9: Don't we have a new term for this?

marun: 'Workload Cluster'

marun:
@pkprzekwas Please reformat to ensure each line is no more than 120 chars. That ensures that review comments can be unambiguously localized vs potentially referring to anywhere in the 3-4 wrapped lines that github is displaying to me.

docs/terminology.md
pkprzekwas: changed to `zero or near zero`. Let me know what do you think.

docs/terminology.md
pkprzekwas: `prepends` -> `enriches` + changed order

docs/terminology.md
pkprzekwas: Replaced as suggested.

pkprzekwas:
> unambiguously

Wrapped to 120 lines.

sttts (approved):

PR #686, opened 2022-03-10
kubectl-kcp: point to authInfo config instead of clone

kylape:
Currently when running e.g. `kubectl kcp workspaces use my-workspace`, the context created in the user's kubeconfig will point to a new user that was cloned from the previously-active context.  This can fill a user's kubeconfig with lots of duplicate users if there are lots of workspace switches.  It also allows the authInfo to drift from the original context and the workspace-specific context, whereas they very likely need to be in sync.

This change will make the new context point to the already-existing authInfo instead of making a copy of it.

Co-authored-by: Chris Sams <csams@redhat.com>

### DIFF
--- a/pkg/cliplugins/workspace/plugin/kubeconfig.go
+++ b/pkg/cliplugins/workspace/plugin/kubeconfig.go
@@ -181,8 +181,9 @@ func (kc *KubeConfig) workspaceDirectoryRestConfig(options *Options, parent bool
 	}
 
 	kubectlOverrides := options.KubectlOverrides
+	_, authInfo := prioritizedAuthInfo(&kubectlOverrides.AuthInfo, currentContextAuthInfo)
 	overrides := &clientcmd.ConfigOverrides{
-		AuthInfo: *prioritizedAuthInfo(&kubectlOverrides.AuthInfo, currentContextAuthInfo),
+		AuthInfo: *authInfo,
 	}
 	return clientcmd.NewDefaultClientConfig(*workpaceDirectoryAwareConfig, overrides).ClientConfig()
 }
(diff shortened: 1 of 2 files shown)


ncdc:
Thanks for doing this - it was on my list of things to talk to @davidfestal about 😄

sttts:
Please describe why in the description.

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -232,26 +232,11 @@ func (kc *KubeConfig) UseWorkspace(ctx context.Context, opts *Options, workspace
...
 
-	kubectlOverrides := opts.KubectlOverrides
-
-	workspaceAuthInfo := prioritizedAuthInfo(&kubectlOverrides.AuthInfo, currentContextAuthInfo)
-	kc.startingConfig.AuthInfos[workspaceContextName] = workspaceAuthInfo
-
 	kc.startingConfig.Contexts[workspaceContextName] = &api.Context{
 		Cluster:  workspaceContextName,
-		AuthInfo: workspaceContextName,
+		AuthInfo: kc.startingConfig.Contexts[kc.startingConfig.CurrentContext].AuthInfo,
```
davidfestal, line 239: Seems like you completely disable to use of `kubectl` overrides. Which means that you cannot do something like:
`kubectl kcp workspace --token=user-1-token use workspace1` to override AuthInfo to be the `user1` when accessing `workspace1`.

Is it really what we want ?

kylape: Probably not.  I'll revise.

davidfestal: I'm refactoring the whole piece of code, so that I have to touch this part as well.
Is it OK if I include the requirement you started answering in this PR ?

stevekuznetsov: Asking people to wait on small PRs because enormous ones are coming is IMHO an anti-pattern - not only because enormous PRs are not great to write and review, but because no contributor likes to hear that their work will be done by someone else / that they need to wait.

kylape: I updated it, but no worries if you refactor it.

davidfestal (approved):
Thanks @kylape !

PR #691, opened 2022-03-11
config/universal: create default namespace by default

sttts:


ncdc (approved):

ncdc:
Rebased on top of main which now has the fix for the bootstrapping flake.

PR #695, opened 2022-03-12
Make namespace scheduling opt-in via label on containing workspace

marun:
Not all use of kcp will involve namespace scheduling or synchronization with a workload cluster.

One such use case involves using a logical cluster (workspace) as a workload cluster for the purposes of syncer testing. Were namespace scheduling to remain enabled for such a workspace, the cluster label of namespaces created by the syncer in that workspace would be removed (due to the cluster not existing in the workspace). Since the syncer's downstream informer filters on the presence of the label in question, the syncer would not be able to see any resources for which the namespace scheduler had removed the label. Disabling the namespace scheduler for the workspace ensures the contents of the workspace more precisely reflect how a kubernetes cluster would behave.

Unlike disabling scheduling for a namespace - where status is still set by the scheduler on the namespace - the goal here is disabling the scheduler entirely. To that end, only log messages will indicate that scheduling is disabled for the workspace.

### DIFF
--- a/pkg/apis/tenancy/v1alpha1/helper/helper.go
+++ b/pkg/apis/tenancy/v1alpha1/helper/helper.go
@@ -60,7 +60,7 @@ func EncodeOrganizationAndClusterWorkspace(organization, workspace string) strin
 // <OrganizationCluster>#$#<ws>. Otherwise, the key will be of the format
 // <OrganizationClsuter>_<org>#$#<ws>.
 func WorkspaceKey(org, ws string) string {
-	if org == RootCluster || strings.HasPrefix(org, LocalSystemClusterPrefix) {
+	if org == RootCluster || (org == "" && ws == RootCluster) || strings.HasPrefix(org, LocalSystemClusterPrefix) {
 		return clusters.ToClusterAwareKey(org, ws)
 	}
 
--- a/pkg/apis/tenancy/v1alpha1/helper/helper_test.go
+++ b/pkg/apis/tenancy/v1alpha1/helper/helper_test.go
@@ -144,6 +144,18 @@ func TestWorkspaceKey(t *testing.T) {
 		want string
 	}{
 		{
+			name: "root ws",
+			org:  "",
+			ws:   RootCluster,
+			want: RootCluster,
+		},
+		{
+			name: "fake root ws",
+			org:  "myorg",
+			ws:   RootCluster,
+			want: "root:myorg#$#root",
+		},
+		{
 			name: "org ws",
 			org:  RootCluster,
 			ws:   "myws",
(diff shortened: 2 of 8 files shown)


test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -146,6 +149,39 @@ func TestNamespaceScheduler(t *testing.T) {
...
+				t.Log("Create a namespace when workspace scheduling is disabled and expect it to remain unscheduled with no condition")
+				namespace, err := server.client.CoreV1().Namespaces().Create(ctx, &corev1.Namespace{
+					ObjectMeta: metav1.ObjectMeta{
+						GenerateName: "e2e-nss-",
+					},
+				}, metav1.CreateOptions{})
+				require.NoError(t, err, "failed to create namespace")
+
+				// Give the namespace scheduler time to reconcile the namespace
+				time.Sleep(5 * time.Second)
```
sttts, line 167: can't we poll for it?

marun: Updated to poll for the lack of a change, PTAL.

marun:
Updated PTAL

pkg/apis/tenancy/v1alpha1/helper/helper.go
```diff
@@ -60,7 +60,7 @@ func EncodeOrganizationAndClusterWorkspace(organization, workspace string) strin
 // <OrganizationCluster>#$#<ws>. Otherwise, the key will be of the format
 // <OrganizationClsuter>_<org>#$#<ws>.
 func WorkspaceKey(org, ws string) string {
-	if org == RootCluster || strings.HasPrefix(org, LocalSystemClusterPrefix) {
+	if org == RootCluster || (org == "" && ws == RootCluster) || strings.HasPrefix(org, LocalSystemClusterPrefix) {
```
sttts, line 63: is it only me who think this function is highly confusion?

I see calls with org being an org name, and calls where it is a org ClusterName.

pkg/apis/tenancy/v1alpha1/helper/helper_test.go
```diff
@@ -143,6 +143,18 @@ func TestWorkspaceKey(t *testing.T) {
...
+			org:  "",
+			ws:   RootCluster,
+			want: RootCluster,
+		},
+		{
+			name: "fake root ws",
+			org:  "myorg",
+			ws:   RootCluster,
+			want: "root:myorg#$#root",
+		},
```
sttts, line 157: `ws:RootCluster` is not really RootCluster, but just a random workspace called "root".

sttts (approved):

pkg/apis/tenancy/v1alpha1/helper/helper.go
marun: I definitely it our key handling confusing, and I think it is likely to be a persistent source of confusion and misunderstanding that all workspaces use the same type with very different behavior.

PR #710, opened 2022-03-14
Support adding labels to workspace on creation

marun:
A field `AdditionalWorkspaceLabels` is added to `ClusterWorkspaceType`. Labels defined by this field for a given
`ClusterWorkspaceType` will be added to `ClusterWorkspace` resources of that type. Labels will only be added on creation of a `ClusterWorkspace` and only if the labels are not already defined (i.e. existing values will not be overwritten).

This PR supports the use case of controlling workspace scheduling by label, as per #695.


pkg/admission/clusterworkspacetypeexists/admission.go
```diff
@@ -94,6 +91,35 @@ func (o *clusterWorkspaceTypeExists) Admit(ctx context.Context, a admission.Attr
...
+
+	clusterName, err := genericapirequest.ClusterNameFrom(ctx)
+	if err != nil {
+		return apierrors.NewInternalError(err)
+	}
+
+	cwt, err := o.typeLister.Get(clusters.ToClusterAwareKey(clusterName, strings.ToLower(cw.Spec.Type)))
+	if err != nil && apierrors.IsNotFound(err) {
+		if cw.Spec.Type == "Universal" {
+			return nil // Universal is always valid
```
sttts, line 106: you better want to fall through. Otherwise the control flow gets tricky.

pkg/admission/clusterworkspacetypeexists/admission.go
```diff
@@ -283,3 +284,33 @@ func (o *clusterWorkspaceTypeExists) SetKcpInformers(informers kcpinformers.Shar
...
+	return nil
+}
+
+// addAdditionlWorkspaceLabels adds labels defined by the workspace
+// type to the workspace if they are not already present.
+func addAdditionalWorkspaceLabels(
+	cwt *tenancyv1alpha1.ClusterWorkspaceType,
+	cw *tenancyv1alpha1.ClusterWorkspace,
+) {
+	if cwt.Spec.AdditionalWorkspaceLabels != nil {
```
sttts, line 304: len() > 0

sttts (approved):

pkg/admission/clusterworkspacetypeexists/admission.go
marun: Done

pkg/admission/clusterworkspacetypeexists/admission.go
marun: If a cw is valid without the cwt for type Universal, that means there is no cwt defining additional labels to apply, which seems ok to me?

marun:
Rebased

pkg/admission/clusterworkspacetypeexists/admission.go
sttts: Universal is the only type that does not have to exist to pass admission. Your code is fine. Was just complaining about the execution flow with this deeply nested return (which I introduced, didn't I?).

marun: Yes, all you 😄

PR #711, opened 2022-03-15
reconcilers: get rid of clientcmdapi.Config and pass rest.Config

sttts:
Non-functional change, just getting rid of the in-process loopback kubeconfig with multiple contexts.

### DIFF
--- a/pkg/server/controllers.go
+++ b/pkg/server/controllers.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	_ "net/http/pprof"
-	"net/url"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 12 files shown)


pkg/server/controllers.go
```diff
@@ -227,34 +227,34 @@ func readCA(file string) ([]byte, error) {
 	return rootCA, err
 }
 
-func (s *Server) installNamespaceScheduler(ctx context.Context, workspaceLister tenancylisters.ClusterWorkspaceLister, clientConfig clientcmdapi.Config, server *genericapiserver.GenericAPIServer) error {
-	kubeClient, err := kubernetes.NewClusterForConfig(server.LoopbackClientConfig)
+func (s *Server) installNamespaceScheduler(ctx context.Context, config *rest.Config) error {
+	kubeClient, err := kubernetes.NewClusterForConfig(config)
```
aojea, line 231: why are we creating multiple clients instead of sharing the same client?

aojea: this will end having multiple TCP connections, can we make the client part of the struct? 
not asking for doing it in this PR,

aojea (approved):
I think we should plumb down the client so it is shared, instead of the config ... otherwise we'll have multiple controller creating multiple TCP connections that will have different delays causing races ... is not that having a single TCP connection completely solves that but since everything is serialized in the same stream it is easier to debug and to troubleshoot

sttts:
For reference: https://github.com/kcp-dev/kcp/issues/714 as follow-up.

PR #715, opened 2022-03-15
kcp-front-proxy: implement options pattern

sttts:
Reuse `SecureServing` from kube, and align default options with `kcp start`.


cmd/kcp-front-proxy/main.go
```diff
@@ -40,26 +66,39 @@ forwards Common Name and Organizations to backend API servers in HTTP headers.
...
 	}
 
-	// setup klog
-	fs := goflags.NewFlagSet("klog", goflags.PanicOnError)
-	klog.InitFlags(fs)
-	cmd.PersistentFlags().AddGoFlagSet(fs)
+	options.AddFlags(cmd.Flags())
 
-	cmd.PersistentFlags().StringVar(&server.ListenAddress, "listen-address", ":8083", "Address and port for the proxy to listen on")
-	cmd.PersistentFlags().StringVar(&server.ClientCACert, "client-ca-cert", "certs/kcp-client-ca-cert.pem", "CA cert used to validate client certs")
```
csams, line 53: @sttts The `--client-ca-file` option isn't available to verify client certs.

csams (changes requested):
Thanks for the changes and opportunity to see a small part of how secure serving is set up!

How do I configure the CA to validate client certs? The `--client-ca-file` option isn't available.

cmd/kcp-front-proxy/main.go
sttts: that's via authentication. Have not wired that, but left your code in place. Might be worth to explore using the kube authenticators too.

sttts: Would be good to know whether I broke the old code or whether it still works. We can attempt to use kube's client cert auth in follow-up.

csams: The old code was validating client certs, so it's broken. :) I chatted with @ncdc and got some pointers for wiring up kube's client cert auth, though. I'll merge and start on a follow up.

csams (approved):
Will follow up with a PR to use kube's client cert auth.

Issue #716, opened 2022-03-15
Authorize ClusterWorkspaceType use in virtual-workspace


sttts:
As in https://github.com/kcp-dev/kcp/blob/main/pkg/admission/clusterworkspacetypeexists/admission.go#L261 for `ClusterWorkspace` create admission, but in the virtual workspace apiserver.

PR #718, opened 2022-03-15
virtual: authorize clusterworkspacetype use

sttts:
Fixes https://github.com/kcp-dev/kcp/issues/716

We had that in ClusterWorkspaceTypeExists admission. But because the virtual-workspace apiserver creates as admin, that did not apply. Also impersonation would not work because users cannot create ClusterWorkspaces in the org workspace.

### DIFF
--- a/pkg/admission/apibinding/apibinding_admission.go
+++ b/pkg/admission/apibinding/apibinding_admission.go
@@ -32,9 +32,9 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 
-	kcpadmissionhelpers "github.com/kcp-dev/kcp/pkg/admission/helpers"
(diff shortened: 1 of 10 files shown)


slaskawi (commented):
LGTM. There are some linter complains though.

davidfestal (commented):
Just a dummy question: is it expected that we don't do a similar check on ckusterworkspaces dor the delete action ?

sttts:
Had somehow lost the commit moving the admission authz helper. Updated.

davidfestal:
> > is it expected that we don't do a similar check on ckusterworkspaces dor the delete action ?
> 
> why delete?

Soryr, I had misread. Forget my question.

davidfestal (approved):

Issue #678, opened 2022-03-10
"kubectl kcp workspace current" in root is broken


sttts:
As discussed, the "current" command inside the root workspace is broken right now, but will be fixed with a planned cleanup/simplification of the plugin. This is just a reminder to track that.

PR #720, opened 2022-03-15
Simplify `kubectl kcp workspace` plugin usage and implementation

davidfestal:
- **Always** derive the current workspace name and org from the server URL of the current `kubeconfig` context (instead of reading it from the current workspace kubeconfig context key). This fixes issue #678
- Store **only one** `kubeconfig` context to store the current workspace context, and one for the previous workspace (instead of the whole history of current contexts)
- Correctly manage pretty names, so that they are shown **as an alias** in the `use` or `current` commands as shown below:
```
$ kubectl kcp workspace --token=another-token create my-workspace
Workspace "my-workspace" created.
$ kubectl kcp workspace --token=user-1-token create my-workspace 
Workspace "my-workspace" created.
# It is a distinct workspace, whose org-wide name will be disambiguated. But the name requested by the user was given as an alias in the personal scope.

$ kubectl kcp workspace --token=user-1-token current
Current workspace is "my-workspace" (an alias for "my-workspace--1") in organization "default".

# Now with the org-wide ("all") scope instead of the personal scope, we directly see the workspace real, org-wide name
$ kubectl kcp workspace --token=user-1-token --scope=all current
Current workspace is "my-workspace--1" in organization "default".``` 
```

- In addition, to use a given context instead of the current context to specify the KCP server URL, no need to reset the current kubeconfig context first. You can now use the standard `--current` `kubectl` override:
```
$ kubectl kcp workspace use my-workspace
Current workspace is "my-workspace" in organization "default".
$ kubectl kcp workspace --context=root use default
Current workspace is "default" in organization "root".
``` 

More details:

For historical reasons until now the plugin used to add, inside the current `kubeconfig`, **one new context for each new current workspace** (after a "use" call), with the context name containing the name of the workspace and the related scope.
This way of storing current workspace contexts has become inconsistent with the current plugin behavior which builds the `workspaces` virtual workspace URL automatically based on the current context server URL, by extracting the current workspace logical cluster name from it.

### DIFF
--- a/pkg/cliplugins/workspace/plugin/helpers.go
+++ b/pkg/cliplugins/workspace/plugin/helpers.go
@@ -17,41 +17,113 @@ limitations under the License.
 package plugin
 
 import (
+	"context"
+	"fmt"
+	"net/url"
+	"path"
 	"strings"
 
-	"k8s.io/client-go/tools/clientcmd/api"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	"github.com/kcp-dev/kcp/pkg/virtual/workspaces/registry"
+	virtualcommandoptions "github.com/kcp-dev/kcp/cmd/virtual-workspaces/options"
+	tenancyhelpers "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1/helper"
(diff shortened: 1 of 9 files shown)


davidfestal:
@varshaprasad96 You might be interested by this PR, above which #632 should be much easier to implement.

sttts (approved):

PR #724, opened 2022-03-16
WIP: tenancy/helpers: move WorkspaceKey to pkg/server/apiexensions.go, this is not a generic helper

sttts:
This helper is very workspace specific. So we should not promote reuse. Depends on the context.

### DIFF
--- a/pkg/apis/tenancy/v1alpha1/helper/helper.go
+++ b/pkg/apis/tenancy/v1alpha1/helper/helper.go
@@ -21,7 +21,6 @@ import (
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/tools/clusters"
 
 	tenancyapi "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
 )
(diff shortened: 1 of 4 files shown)


ncdc:
Following up from Slack & Meet conversations. I would recommend keeping this helper, somewhere. We should simplify it so it looks like this:

```go
func WorkspaceKey(org, ws string) string {
  strings.HasPrefix(org, LocalSystemClusterPrefix) {
    panic("programmer error") // we could do this, or not, but it should never be called this way
  }
   
  if org == RootCluster {
     return clusters.ToClusterAwareKey(org, ws)
   }

   return clusters.ToClusterAwareKey(EncodeOrganizationAndClusterWorkspace(RootCluster, org), ws)
 }
```

If we remove this helper, than anyone who ever needs to look up a workspace using an indexer key will need to recreate this logic, because the key changes if you're looking up an org workspace or non-org workspace.

marun:
Updated WorkspaceKey to panic when the workspace key would be invalid and added a halper (HasClusterWorkspace) to enable detecting when an org:ws pair are not valid for retrieving a ClusterWorkspace: https://github.com/kcp-dev/kcp/pull/735

ncdc:
Closing, replaced by #735.

PR #735, opened 2022-03-16
Update WorkspaceKey to panic for org+ws without ClusterWorkspace

marun:
This ensures a clear indication to callers attempting to retrieve a ClusterWorkspace for logical clusters like `root` and `system:admin` that what they are trying to do is not supported.

### DIFF
--- a/pkg/server/apiextensions.go
+++ b/pkg/server/apiextensions.go
@@ -210,6 +210,11 @@ func (c *inheritanceCRDLister) GetWithContext(ctx context.Context, name string) 
 		return nil, apierrors.NewNotFound(apiextensionsv1.Resource("customresourcedefinitions"), name)
 	}
 
+	if !helper.HasClusterWorkspace(org, ws) {
+		// Without a cluster workspace no inheritance is possible
+		return nil, apierrors.NewNotFound(apiextensionsv1.Resource("customresourcedefinitions"), name)
+	}
+
 	// Check for API inheritance
 	targetWorkspaceKey := helper.WorkspaceKey(org, ws)
 	workspace, err := c.workspaceLister.Get(targetWorkspaceKey)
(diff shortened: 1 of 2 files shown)


pkg/apis/tenancy/v1alpha1/helper/helper_test.go
```diff
@@ -143,18 +143,6 @@ func TestWorkspaceKey(t *testing.T) {
 		ws   string
 		want string
 	}{
-		{
```
ncdc, line 146: Testify has some assertions for panics, doesn't it?

marun: Done.

pkg/apis/tenancy/v1alpha1/helper/helper_test.go
```diff
@@ -136,25 +138,44 @@ func TestParseLogicalClusterName(t *testing.T) {
...
-		want string
+		want bool
+	}{{
+		name: "root ws",
+		org:  "",
+		ws:   RootCluster,
+		want: false,
+	}, {
+		name: "system:admin ws",
+		org:  "system:admin",
```
marun, line 154: Why do we supply `ws: "root"` but `org: "system:admin` ?

marun: I mean to say, the root cluster is represented by `org: "", ws: "root"`, whereas the system:admin cluster is represented by `org: "system:admin", ws: "<arbitrary>"`. I inherited this, I'd like to understand why it makes sense.

marun:
I'm not entirely convinced it wouldn't be preferable to just return an error for callers to handle...

pkg/apis/tenancy/v1alpha1/helper/helper_test.go
sttts: that's probably from a blind rewrite from our former admin:foo. We basically have split former admin into system:admin and root. So +1 for removing the case. system:admin should never be in org.

sttts:
> I'm not entirely convinced it wouldn't be preferable to just return an error for callers to handle...

No error please. This is a programming error. We don't want super soft type system. It does not help code quality, but the opposite.

pkg/apis/tenancy/v1alpha1/helper/helper.go
```diff
@@ -55,13 +55,30 @@ func EncodeOrganizationAndClusterWorkspace(organization, workspace string) strin
...
 // <OrganizationCluster>#$#<ws>. Otherwise, the key will be of the format
 // <OrganizationClsuter>_<org>#$#<ws>.
-func WorkspaceKey(org, ws string) string {
-	if org == RootCluster || (org == "" && ws == RootCluster) || strings.HasPrefix(org, LocalSystemClusterPrefix) {
-		return clusters.ToClusterAwareKey(org, ws)
+func WorkspaceKey(org string, ws string) string {
+	if org == "" && ws == RootCluster {
+		panic("The root cluster is not backed by a ClusterWorkspace")
+	}
+	if strings.HasPrefix(org, LocalSystemClusterPrefix) {
```
sttts, line 76: would go one step further: check for colon. Colon is a sign that we did something wrong.

ncdc: +1

pkg/apis/tenancy/v1alpha1/helper/helper_test.go
ncdc: +1

marun: I've update the code to check for `:` and updated this test case accordingly.

marun: Done.

ncdc (approved):

PR #736, opened 2022-03-16
e2e: Update workspace shard tests to use shared fixture

marun:
Also fixed a bug in the workspace shard controller that was uncovered in the process of updating the tests.

### DIFF
--- a/pkg/reconciler/workspace/controller.go
+++ b/pkg/reconciler/workspace/controller.go
@@ -320,7 +320,7 @@ func (c *Controller) reconcile(ctx context.Context, workspace *tenancyv1alpha1.C
 			}
 
 			if len(validShards) > 0 {
-				targetShard := shards[rand.Intn(len(shards))]
+				targetShard := validShards[rand.Intn(len(validShards))]
 
 				u, err := url.Parse(targetShard.Status.ConnectionInfo.Host)
 				if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/reconciler/workspace/controller.go
```diff
@@ -320,7 +320,7 @@ func (c *Controller) reconcile(ctx context.Context, workspace *tenancyv1alpha1.C
 			}
 
 			if len(validShards) > 0 {
-				targetShard := shards[rand.Intn(len(shards))]
+				targetShard := validShards[rand.Intn(len(validShards))]
```
sttts, line 323: 🎉

test/e2e/reconciler/workspace/controller_test.go
```diff
@@ -171,6 +171,8 @@ func TestWorkspaceController(t *testing.T) {
 		},
 	}
 
+	server := framework.SharedKcpServer(t)
```
sttts, line 174: line 77 is potentially disruptive. Have you checked that it isn't, e.g. by not calling t.parallel? I don't think this shared fixture will work otherwise.

marun: Oof. I'll submit it separately then.

marun: Removed for now.

sttts (approved):

ncdc (approved):

Issue #418, opened 2022-02-03
Prototype 3-4: PCluster Registration/Admin Demo


pweil-:
### Demo Objective
- [x] Document a repeatable process for registering a physical cluster with kcp
- [x] (stretch goal) Add support in the `kubectl kcp` plugin to register a physical cluster with kcp

### Demo Steps
1. User (or kcp admin, or org admin) needs to add new cluster (capacity, location, whatever)
    1. Choose/Create workspace (this will be the one the syncer will sync).
    2. Create SA in that workspace. The controller automatically generates a token in a `"token"` secret next to the SA.
    3. pcluster admin create kubeconfig with the upper token, and creates that as a secret in the pcluster syncer namespace
1. Reviews the yaml that creates the agent on the cluster
    1.  Sees reasonable roles+rolebindings, a deployment, and has to adapt the kubeconfig secret mount to match secret created above.
    2. YAML contains address of KCP to connect to
    3. YAML contains the unique name of this cluster (according to admin user)
    4. YAML references SA+token to authenticate to KCP, copied from KCP previously
1. Deploys the yaml 
    1. `kubectl apply -f syncer.yaml` 
    2. **Out-of-scope for p3** Possibly more automation around this
    3. Syncer starts by creating a `ClusterWorkspace` CR in KCP
1. Sees the WorkloadCluster show up in the workspace
1. **Out-of-scope for p3**: Does administrative action that makes this available to a workspace
1. Options:
    1. Decommission old cluster
        1. Triggers moves and drains
    1. Add new capacity
        1. The moment the WorkloadCluster is available, the new workload which wasn’t scheduled suddenly gets scheduled and placed (might be a stretch)
    1. Failure
        1. Sees an alert because the cluster is not responding? Don’t have any of our monitoring or conditions done (might be a stretch)

### Action Items

- [x] Scope the current demo as necessary to fit in prototype boundaries
- Technical tasks
   - [x] https://github.com/kcp-dev/kcp/pull/681 wire TokensController into pkg/server/server.go  @sttts  @marun 
   - ~~[ ] create WorkloadCluster object from syncer~~ find and document a procedure to make the syncer work through the service-account @aojea / @marun  
   - [x] build syncer controller that heartbeats @marun
   - [x] #738 build cluster-health controller in the server @imjasonh
   - [x] #723 @aojea 
   - [x] minimal, experimental/temporary RBAC policy on the pcluster
[...]

PR #738, opened 2022-03-17
Cluster heartbeat controller

imjasonh:
## Summary

This adds a controller to periodically check clusters' `.Status.Conditions[].LastHeartbeatTime` values, which the syncer should update periodically (cc @jmprusi), and marks them as not `Ready` if a heartbeat hasn't been seen in a specified, configurable amount of time (default currently 1 minute, open to discussion).

## Related issue(s)

- #418

### DIFF
--- a/pkg/apis/workload/v1alpha1/types.go
+++ b/pkg/apis/workload/v1alpha1/types.go
@@ -87,10 +87,6 @@ type WorkloadClusterStatus struct {
 
 	// +optional
 	SyncedResources []string `json:"syncedResources,omitempty"`
-
-	// LastHeartbeat represents the last time the cluster's syncer successfully made a request to update this value.
-	// +optional
-	LastHeartbeat *metav1.Time `json:"lastHeartbeat,omitempty"`
 }
 
 // WorkloadClusterList is a list of WorkloadCluster resources
--- a/pkg/apis/workload/v1alpha1/types.go
+++ b/pkg/apis/workload/v1alpha1/types.go
@@ -87,6 +87,10 @@ type WorkloadClusterStatus struct {
 
 	// +optional
 	SyncedResources []string `json:"syncedResources,omitempty"`
+
+	// LastHeartbeat represents the last time the cluster's syncer successfully made a request to update this value.
+	// +optional
+	LastHeartbeat *metav1.Time `json:"lastHeartbeat,omitempty"`
 }
 
 // WorkloadClusterList is a list of WorkloadCluster resources
@@ -130,6 +134,9 @@ const (
 
 	// ErrorStartingAPIImporterReason indicates an error starting the API Importer.
 	ErrorStartingAPIImporterReason = "ErrorStartingAPIImporter"
+
+	// ErrorHeartbeat indicates that a heartbeat update was not received within the configured threshold.
+	ErrorHeartbeat = "ErrorHeartbeat"
 )
 
 func (in *WorkloadCluster) SetConditions(c conditionsv1alpha1.Conditions) {
--- a/pkg/apis/workload/v1alpha1/types.go
+++ b/pkg/apis/workload/v1alpha1/types.go
@@ -135,8 +135,8 @@ const (
 	// ErrorStartingAPIImporterReason indicates an error starting the API Importer.
 	ErrorStartingAPIImporterReason = "ErrorStartingAPIImporter"
 
-	// ErrorHeartbeat indicates that a heartbeat update was not received within the configured threshold.
-	ErrorHeartbeat = "ErrorHeartbeat"
+	// ErrorHeartbeatMissedReason indicates that a heartbeat update was not received within the configured threshold.
+	ErrorHeartbeatMissedReason = "ErrorHeartbeat"
 )
 
 func (in *WorkloadCluster) SetConditions(c conditionsv1alpha1.Conditions) {
--- a/pkg/reconciler/cluster/cluster_controller.go
+++ b/pkg/reconciler/cluster/cluster_controller.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cluster
+
+import (
+	"time"
+
+	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	apiresourceinformer "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/apiresource/v1alpha1"
+	workloadinformer "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/workload/v1alpha1"
+)
+
+func NewController(
+	kcpClusterClient *kcpclient.Cluster,
+	clusterInformer workloadinformer.WorkloadClusterInformer,
+	apiResourceImportInformer apiresourceinformer.APIResourceImportInformer,
+	heartbeatThreshold time.Duration,
+) (*ClusterReconciler, error) {
+
+	cm := &clusterManager{heartbeatThreshold: heartbeatThreshold}
+
+	return NewClusterReconciler(
+		"kcp-cluster-manager",
+		cm,
+		kcpClusterClient,
+		clusterInformer,
+		apiResourceImportInformer,
+	)
+}
--- a/pkg/reconciler/cluster/cluster_controller_start.go
+++ b/pkg/reconciler/cluster/cluster_controller_start.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cluster
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/spf13/pflag"
+)
+
+func DefaultOptions() *Options {
+	return &Options{
+		HeartbeatThreshold: time.Minute,
+	}
+}
+
+func BindOptions(o *Options, fs *pflag.FlagSet) *Options {
+	fs.DurationVar(&o.HeartbeatThreshold, "cluster-heartbeat-threshold", o.HeartbeatThreshold, "Amount of time to wait for a successful heartbeat before marking the cluster as not ready")
+	return o
+}
+
+type Options struct {
+	HeartbeatThreshold time.Duration
+}
+
+func (o *Options) Validate() error {
+	if o.HeartbeatThreshold < 0 {
+		return fmt.Errorf("heartbeat threshold must be >0 (%s)", o.HeartbeatThreshold)
+	}
+	return nil
+}
--- a/pkg/reconciler/cluster/cluster_manager.go
+++ b/pkg/reconciler/cluster/cluster_manager.go
@@ -35,7 +35,7 @@ func (c *clusterManager) Reconcile(ctx context.Context, cluster *workloadv1alpha
 	if cluster.Status.LastHeartbeat != nil && time.Since(cluster.Status.LastHeartbeat.Time) > c.heartbeatThreshold {
 		conditions.MarkFalse(cluster,
 			workloadv1alpha1.WorkloadClusterReadyCondition,
-			workloadv1alpha1.ErrorHeartbeat,
+			workloadv1alpha1.ErrorHeartbeatMissedReason,
 			conditionsv1alpha1.ConditionSeverityError,
 			"No heartbeat since %s", cluster.Status.LastHeartbeat)
 	} else {
--- a/pkg/reconciler/cluster/cluster_manager.go
+++ b/pkg/reconciler/cluster/cluster_manager.go
@@ -1,0 +1,52 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cluster
+
+import (
+	"context"
+	"time"
+
+	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
+	conditionsv1alpha1 "github.com/kcp-dev/kcp/third_party/conditions/apis/conditions/v1alpha1"
+	"github.com/kcp-dev/kcp/third_party/conditions/util/conditions"
+)
+
+var _ ClusterReconcileImpl = (*clusterManager)(nil)
+
+type clusterManager struct {
+	heartbeatThreshold time.Duration
+}
+
+func (c *clusterManager) Reconcile(ctx context.Context, cluster *workloadv1alpha1.WorkloadCluster) error {
+	if cluster.Status.LastHeartbeat != nil && time.Since(cluster.Status.LastHeartbeat.Time) > c.heartbeatThreshold {
+		conditions.MarkFalse(cluster,
+			workloadv1alpha1.WorkloadClusterReadyCondition,
+			workloadv1alpha1.ErrorHeartbeat,
+			conditionsv1alpha1.ConditionSeverityError,
+			"No heartbeat since %s", cluster.Status.LastHeartbeat)
+	} else {
+		conditions.MarkTrue(cluster,
+			workloadv1alpha1.WorkloadClusterReadyCondition)
+
+		// TODO: enqueue another check in $heartbeatThreshold.
+	}
+
+	return nil
+}
+
+func (c *clusterManager) Cleanup(ctx context.Context, deletedCluster *workloadv1alpha1.WorkloadCluster) {
+}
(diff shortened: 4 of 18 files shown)


pkg/apis/workload/v1alpha1/types.go
```diff
@@ -130,6 +134,9 @@ const (
 
 	// ErrorStartingAPIImporterReason indicates an error starting the API Importer.
 	ErrorStartingAPIImporterReason = "ErrorStartingAPIImporter"
+
+	// ErrorHeartbeat indicates that a heartbeat update was not received within the configured threshold.
+	ErrorHeartbeat = "ErrorHeartbeat"
```
ncdc, line 139: ErrorHeartbeatMissedReason?

imjasonh: Good idea, done.

pkg/reconciler/cluster/cluster_manager.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/kcp-dev/kcp/third_party/conditions/util/conditions"
+)
+
+var _ ClusterReconcileImpl = (*clusterManager)(nil)
+
+type clusterManager struct {
+	heartbeatThreshold time.Duration
+}
+
+func (c *clusterManager) Reconcile(ctx context.Context, cluster *workloadv1alpha1.WorkloadCluster) error {
```
marun, line 34: (No action required) Maybe factor this method so that it's logic could be tested without instantiating the manager or providing a context? Also, maybe add a unit test?

pkg/reconciler/cluster/cluster_controller.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"time"
+
+	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	apiresourceinformer "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/apiresource/v1alpha1"
+	workloadinformer "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/workload/v1alpha1"
+)
+
+func NewController(
```
sttts, line 27: be specific what this does: NewHeartbeatWatcherController or so

pkg/reconciler/cluster/cluster_controller_start.go
```diff
@@ -0,0 +1,46 @@
...
+package cluster
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/spf13/pflag"
+)
+
+func DefaultOptions() *Options {
```
sttts, line 26: NewOptions is our pattern elsewhere

pkg/reconciler/cluster/cluster_controller_start.go
```diff
@@ -0,0 +1,46 @@
...
+	return &Options{
+		HeartbeatThreshold: time.Minute,
+	}
+}
+
+func BindOptions(o *Options, fs *pflag.FlagSet) *Options {
```
sttts, line 32: Options.AddFlags

pkg/reconciler/cluster/cluster_controller_start.go
```diff
@@ -0,0 +1,46 @@
...
+	fs.DurationVar(&o.HeartbeatThreshold, "cluster-heartbeat-threshold", o.HeartbeatThreshold, "Amount of time to wait for a successful heartbeat before marking the cluster as not ready")
+	return o
+}
+
+type Options struct {
+	HeartbeatThreshold time.Duration
+}
+
+func (o *Options) Validate() error {
```
sttts, line 41: []error as return type

pkg/reconciler/cluster/cluster_controller_start.go
```diff
@@ -0,0 +1,46 @@
...
+	if o.HeartbeatThreshold < 0 {
+		return fmt.Errorf("heartbeat threshold must be >0 (%s)", o.HeartbeatThreshold)
```
sttts, line 43: I believe we write `--cluster-heartbeat-threshold must be ...` elsewhere

pkg/reconciler/cluster/cluster_controller_start.go
```diff
@@ -0,0 +1,46 @@
...
+	fs.DurationVar(&o.HeartbeatThreshold, "cluster-heartbeat-threshold", o.HeartbeatThreshold, "Amount of time to wait for a successful heartbeat before marking the cluster as not ready")
+	return o
+}
+
+type Options struct {
+	HeartbeatThreshold time.Duration
+}
+
+func (o *Options) Validate() error {
+	if o.HeartbeatThreshold < 0 {
```
sttts, line 42: <=

pkg/reconciler/cluster/cluster_manager.go
```diff
@@ -0,0 +1,52 @@
...
+		conditions.MarkFalse(cluster,
+			workloadv1alpha1.WorkloadClusterReadyCondition,
+			workloadv1alpha1.ErrorHeartbeatMissedReason,
+			conditionsv1alpha1.ConditionSeverityError,
+			"No heartbeat since %s", cluster.Status.LastHeartbeat)
+	} else {
+		conditions.MarkTrue(cluster,
+			workloadv1alpha1.WorkloadClusterReadyCondition)
+
+		// TODO: enqueue another check in $heartbeatThreshold.
```
sttts, line 45: would do that right away. It's kind of essential for the controller, isn't it?

pkg/reconciler/cluster/cluster_manager.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	return nil
+}
+
+func (c *clusterManager) Cleanup(ctx context.Context, deletedCluster *workloadv1alpha1.WorkloadCluster) {
```
sttts, line 51: what is this going to do?

pkg/reconciler/cluster/cluster_controller.go
```diff
@@ -0,0 +1,43 @@
+/*
```
sttts, line 1: would very much prefer to see this not in pkg/reconciler/cluster. This dir is far too crowded already. One dir, one controller, e.g. `pkg/reconciler/workloadclusterheartbeat`.

pkg/server/controllers.go
```diff
@@ -510,6 +511,38 @@ func (s *Server) installSyncerController(ctx context.Context, config *rest.Confi
 	return nil
 }
 
+func (s *Server) installClusterController(ctx context.Context, config *rest.Config) error {
+	config = rest.AddUserAgent(rest.CopyConfig(config), "kcp-cluster-controller")
```
sttts, line 515: kcp-cluster-heartbeat-controller

pkg/reconciler/cluster/cluster_controller_start.go
sttts: ic, our controller are still non-standard. Feel free to leave the options func names as they are.

pkg/reconciler/cluster/cluster_controller_start.go
sttts: as above

pkg/server/options/controllers.go
```diff
@@ -40,12 +41,14 @@ type Controllers struct {
 	ApiImporter         ApiImporterController
 	ApiResource         ApiResourceController
 	Syncer              SyncerController
+	Cluster             ClusterController
```
sttts, line 44: WorkloadCluster

pkg/reconciler/cluster/cluster_controller_start.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+func DefaultOptions() *Options {
+	return &Options{
+		HeartbeatThreshold: time.Minute,
+	}
+}
+
+func BindOptions(o *Options, fs *pflag.FlagSet) *Options {
+	fs.DurationVar(&o.HeartbeatThreshold, "cluster-heartbeat-threshold", o.HeartbeatThreshold, "Amount of time to wait for a successful heartbeat before marking the cluster as not ready")
```
sttts, line 33: workloadcluster-heartbeat-threshold

pkg/server/handler.go
```diff
@@ -42,7 +42,6 @@ import (
 	apiserverdiscovery "k8s.io/apiserver/pkg/endpoints/discovery"
 	"k8s.io/apiserver/pkg/endpoints/handlers/responsewriters"
 	"k8s.io/apiserver/pkg/endpoints/request"
-	genericapirequest "k8s.io/apiserver/pkg/endpoints/request"
```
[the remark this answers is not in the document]
imjasonh: It is, my editor was bugging me about the duplicate import. Happy to revert it and just ignore the editor if you'd rather.

pkg/reconciler/cluster/cluster_manager.go
imjasonh: AFAIK it's got nothing to do. In theory it could cancel any queued-in-the-future checks, but I'm not really sure it's worth it. It just needs this to satisfy the interface.

pkg/reconciler/cluster/cluster_controller_start.go
imjasonh: All the others return `error` -- I think if we decide to surface multiple errors we should do so using `multierr`.

pkg/reconciler/cluster/cluster_controller.go
imjasonh: With the new package name is `workloadclusterheartbeat.NewController`, which I think is descriptive enough without being repetitive. Let me know if you disagree.

pkg/server/handler.go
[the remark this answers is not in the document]
marun: It's fine, maybe break out into a separate commit?

pkg/reconciler/cluster/workloadclusterheartbeat/cluster_controller.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/kcp-dev/kcp/pkg/reconciler/cluster"
+)
+
+func NewController(
+	kcpClusterClient *kcpclient.Cluster,
+	clusterInformer workloadinformer.WorkloadClusterInformer,
+	apiResourceImportInformer apiresourceinformer.APIResourceImportInformer,
+	heartbeatThreshold time.Duration,
+) (*cluster.ClusterReconciler, error) {
+
```
sttts, line 34: nit: no empty line

pkg/server/options/flags.go
```diff
@@ -144,6 +144,7 @@ var (
 		"run-virtual-workspaces",                 // Run the virtual workspaces apiservers in-process
 		"syncer-image",                           // Syncer image to install on clusters
 		"unsupported-run-individual-controllers", // Run individual controllers in-process. The controller names can change at any time.
+		"cluster-heartbeat-threshold",            // Amount of time to wait for a successful heartbeat before marking the cluster as not ready.
```
sttts, line 147: needs update

sttts:
flags_test needs an update. Other than that LGTM.

sttts:
Will approve, please fix the test and then it's ready to merge.

pkg/apis/workload/v1alpha1/types.go
```diff
@@ -87,6 +87,10 @@ type WorkloadClusterStatus struct {
 
 	// +optional
 	SyncedResources []string `json:"syncedResources,omitempty"`
+
+	// LastHeartbeat represents the last time the cluster's syncer successfully made a request to update this value.
+	// +optional
+	LastHeartbeat *metav1.Time `json:"lastHeartbeat,omitempty"`
```
marun, line 93: I'm not sure syncer health is accurately represented by this field. That the syncer can update this field indicates that it has connectivity (both network and configuration) to the kcp api and its token or cert gives it the permissions necessary to update the cluster.

However, there are other dimensions that the syncer will likely want to communicate in the not-too-distance future:
 - that the resource syncer has the necessary connectivity and permission to the pcluster that it needs
 - that an eventual separately permissioned namespace syncer has connectivity to kcp
 - that the namespace syncer has the necessary connectivity and permission to the pcluster that it needs

Note the implication of 'connectivity to pcluster' - the syncer may not actually be running in the pcluster in all configurations.

I'm not suggesting that all these details need to be represented in this PR, but I do wonder if it might make sense to anticipate the need for more than just heartbeat as an indication of syncer health. Maybe use conditions (or something similar) to allow the provision of more detail and provide an easier path for future extensibility? And assume the need for separate reporting for an eventual namespace syncer?

imjasonh: Thanks for the feedback. I agree. I'll refactor to make this a condition instead of a single field. In the future, other syncer-set conditions can be added, and the heartbeat controller (or whatever it's called then) can aggregate those multiple signals into `Ready` or `Available` or whatever.

pkg/reconciler/cluster/workloadclusterheartbeat/cluster_manager.go
```diff
@@ -0,0 +1,53 @@
...
+		conditions.MarkFalse(cluster,
+			workloadv1alpha1.WorkloadClusterReadyCondition,
+			workloadv1alpha1.ErrorHeartbeatMissedReason,
+			conditionsv1alpha1.ConditionSeverityError,
+			"No heartbeat since %s", cluster.Status.LastHeartbeat)
+	} else {
+		conditions.MarkTrue(cluster,
+			workloadv1alpha1.WorkloadClusterReadyCondition)
+
+		// TODO: enqueue another check in $heartbeatThreshold.
```
sttts, line 46: this is a blocker to merge this PR, isn't it? Without the controller just does not do its job, does it?

imjasonh: This is a blocker, and why the PR is still WIP :) 

I've added this, though it's a bit hacky how it gets in there given our current reconciler/controller/manager frameworks.

pkg/apis/workload/v1alpha1/types.go
```diff
@@ -130,6 +130,9 @@ const (
 
 	// ErrorStartingAPIImporterReason indicates an error starting the API Importer.
 	ErrorStartingAPIImporterReason = "ErrorStartingAPIImporter"
+
+	// ErrorHeartbeatMissedReason indicates that a heartbeat update was not received within the configured threshold.
+	ErrorHeartbeatMissedReason = "ErrorHeartbeat"
```
marun, line 135: nit: -> `ErrorHeartbeatMissed` ?

pkg/reconciler/cluster/workloadclusterheartbeat/cluster_manager.go
```diff
@@ -0,0 +1,67 @@
...
+var _ cluster.ClusterReconcileImpl = (*clusterManager)(nil)
+
+type clusterManager struct {
+	heartbeatThreshold  time.Duration
+	enqueueClusterAfter func(*workloadv1alpha1.WorkloadCluster, time.Duration)
+}
+
+func (c *clusterManager) Reconcile(ctx context.Context, cluster *workloadv1alpha1.WorkloadCluster) error {
+	latestHeartbeat := time.Time{}
+	for _, c := range cluster.Status.Conditions {
```
marun, line 38: (No action required) Would it make sense to look for a condition/conditions specific to the syncer? Or are we expecting only conditions relevant to cluster readiness here?

pkg/reconciler/cluster/workloadclusterheartbeat/cluster_manager.go
```diff
@@ -0,0 +1,67 @@
...
+		if c.LastHeartbeatTime.Time.After(latestHeartbeat) {
+			latestHeartbeat = c.LastHeartbeatTime.Time
+		}
+	}
+	if latestHeartbeat.IsZero() {
+		// Never seen a heartbeat; for now, this means the syncer hasn't heartbeated -- in the future, this should be un-Ready.
+		return nil
+	}
+
+	if time.Since(latestHeartbeat) > c.heartbeatThreshold {
```
marun, line 48: (No action required) Maybe add a TODO to use something like a sliding window to avoid setting unready for a transient issue?

marun (commented):
What's proposed LGTM, but I'm not sure it should merge by itself. Without the corresponding  change proposed in #356, this new controller ensures that clusters go not ready after the default duration of 1m (since nothing will be heartbeating). 

Would you prefer to merge this PR with the controller disabled and re-enable when heartbeating is available, or include the changes from #356 in this PR so that healthy clusters will stay ready?  In either case, I think additional e2e coverage is suggested when the controller is enabled to validate the behavior and ensure against regression.

pkg/apis/workload/v1alpha1/types.go
sttts: Am confused. Where is the heartbeat stored? Do we update some condition timestamp regularly?

sttts: Also all these are reasons, not conditions. Every condition should have a clear owner. Who owns WorkloadClusterReadyCondition ?

imjasonh:
> What's proposed LGTM, but I'm not sure it should merge by itself. Without the corresponding change proposed in #356, this new controller ensures that clusters go not ready after the default duration of 1m (since nothing will be heartbeating).
> 
> Would you prefer to merge this PR with the controller disabled and re-enable when heartbeating is available, or include the changes from #356 in this PR so that healthy clusters will stay ready? In either case, I think additional e2e coverage is suggested when the controller is enabled to validate the behavior and ensure against regression.

If no heartbeat timestamps are seen at all, the controller currently does nothing: https://github.com/kcp-dev/kcp/pull/738/files#diff-0676de99724f6d912c663cadb712e10a44e45cea872dcb779dd75371c4a52988R44 -- there's a TODO to make this update to not-`Ready`, when the syncer sends those.

pkg/reconciler/cluster/workloadclusterheartbeat/cluster_manager.go
imjasonh: My thinking was that the heartbeat interval would be some fraction of this controller's heartbeat threshold, say 10s. So the controller would only mark the cluster as not-`Ready` if N (=6) heartbeats were missed. Both of these would be configurable from either end, so you could tweak that to your liking based on the relative amount of transient issues you expect from clusters in the system.

pkg/reconciler/cluster/workloadclusterheartbeat/cluster_manager.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func (c *clusterManager) Reconcile(ctx context.Context, cluster *workloadv1alpha1.WorkloadCluster) error {
+	latestHeartbeat := time.Time{}
+	for _, c := range cluster.Status.Conditions {
+		if c.LastHeartbeatTime.Time.After(latestHeartbeat) {
+			latestHeartbeat = c.LastHeartbeatTime.Time
+		}
+	}
+	if latestHeartbeat.IsZero() {
```
marun, line 43: Aha, I didn't grasp this until your prompting.

marun (approved):

marun:
https://github.com/kcp-dev/kcp/issues/739 strikes again. I'm going to get to the bottom of it.

marun:
@imjasonh Merge when ready. I wasn't sure if you wanted it merged since the title still says WIP

marun:
For next time: maybe we get into the habit of squashing commits to allow for a coherent git history?

PR #741, opened 2022-03-17
Use genericConfig.ExternalAddress instead of ExternalHost

kylape:
The apiserver config allows the port to be added to the
`--external-hostname` option and then applies it to the `ExternalAddress`
field in the `genericConfig` (with other defaulting code in case port is
not provided).

This change uses the computed `genericConfig.ExternalAddress` instead of
trying to compute it in kcp code.


kylape:
I ended up significantly changing this PR, so I'm going to close this and reopen another one.

PR #743, opened 2022-03-18
Create new external.kubeconfig

### LINKED ISSUE #707: Split apart our single kubeconfig into multiple
We have a single kubeconfig that is shared by kcp in-process, humans (admin.kubeconfig), and potentially components of the architecture such as the virtual workspaces apiserver. We need to split these apart:

- [x] The loopback client must be for internal, in-process use only, by kcp & its built-in controllers. It must not take into account `--external-hostname`. It must not have any kubeconfig contexts.
- [ ] `admin.kubeconfig` is for human consumption. It must honor `--external-hostname`. We may need to add `--external-hostname-port`. This is the only kubeconfig in which we generate contexts.
- [ ] service.kubeconfig, for components such as the virtual-workspaces standalone apiserver

kylape:
`external.kubeconfig` uses the `--external-hostname` option to populate the
hostname *and port*.  The generic apiserver config allows the port to be added to the
`--external-hostname`.

`admin.kubeconfig` will now always use the IP address of the locally bound
interface and the port from `--secure-port`.

`--kubeconfig-path` now points to the directory containing the kubeconfig
files.  This now means that the `admin.kubeconfig` filename cannot be
changed.

Fixes #707

### DIFF
--- a/pkg/server/options/authentication.go
+++ b/pkg/server/options/authentication.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 
 	"github.com/google/uuid"
 	"github.com/spf13/pflag"
@@ -45,7 +46,7 @@ type AdminAuthentication struct {
 
 func NewAdminAuthentication() *AdminAuthentication {
 	return &AdminAuthentication{
-		KubeConfigPath:    "admin.kubeconfig",
+		KubeConfigPath:    ".",
 		TokenHashFilePath: ".admin-token-store",
 	}
 }
@@ -58,7 +59,7 @@ func (s *AdminAuthentication) Validate() []error {
 	errs := []error{}
 
 	if s.TokenHashFilePath == "" && s.KubeConfigPath != "" {
-		errs = append(errs, fmt.Errorf("--admin-kubeconfig requires --admin-token-hash-file-path"))
+		errs = append(errs, fmt.Errorf("--kubeconfig-path requires --admin-token-hash-file-path"))
 	}
 
 	return errs
@@ -70,7 +71,7 @@ func (s *AdminAuthentication) AddFlags(fs *pflag.FlagSet) {
 	}
 
 	fs.StringVar(&s.KubeConfigPath, "kubeconfig-path", s.KubeConfigPath,
-		"Path to which the administrative kubeconfig should be written at startup. If this is relative, it is relative to --root-directory.")
+		"Dir to which the kubeconfigs should be written at startup. If this is relative, it is relative to --root-directory.")
 	fs.StringVar(&s.TokenHashFilePath, "authentication-admin-token-path", s.TokenHashFilePath,
 		"Path to which the administrative token hash should be written at startup. If this is relative, it is relative to --root-directory.")
 }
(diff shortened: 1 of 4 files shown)


pkg/server/options/authentication.go
```diff
@@ -58,7 +59,7 @@ func (s *AdminAuthentication) Validate() []error {
 	errs := []error{}
 
 	if s.TokenHashFilePath == "" && s.KubeConfigPath != "" {
-		errs = append(errs, fmt.Errorf("--admin-kubeconfig requires --admin-token-hash-file-path"))
+		errs = append(errs, fmt.Errorf("--kubeconfig-path requires --admin-token-hash-file-path"))
```
sttts, line 62: am not a fan of this change. Just add another flag for the other kubeconfig.

pkg/server/virtual.go
```diff
@@ -40,7 +39,7 @@ type mux interface {
 	Handle(pattern string, handler http.Handler)
 }
 
-func (s *Server) installVirtualWorkspaces(ctx context.Context, kubeClusterClient kubernetesclient.ClusterInterface, kcpClusterClient kcpclient.ClusterInterface, auth genericapiserver.AuthenticationInfo, preHandlerChainMux mux) error {
+func (s *Server) installVirtualWorkspaces(ctx context.Context, kubeClusterClient kubernetesclient.ClusterInterface, kcpClusterClient kcpclient.ClusterInterface, genericConfig *genericapiserver.Config, preHandlerChainMux mux) error {
```
sttts, line 42: don't like this change. Pass the info you need explicitly. The change here encourages deep coupling that we don't want.

sttts:
`"external-hostname",                    // The hostname to use when generating externalized URLs for this master (e.g. Swagger API Docs or OpenID Discovery).` 

Don't like that we tweak its meaning. Just add `--external-port` as a flag.

pkg/server/options/authentication.go
kylape: This line is actually correcting a mismatch in option names.  `--admin-kubeconfig` is not an actual flag.  Given that, I know I am still changing the meaning of `--kubeconfig-path`.  Would you still like me to add another flag for the new kubeconfig?

kylape:
Here's how I followed the code to make me believe `--external-hostname` allows a port in kube-apiserver:

`ExternalHost` gets copied over to `ExternalAddress` in the config:

https://github.com/kcp-dev/kubernetes/blob/10415751c5a9f503df2dca997114a2d4de855340/staging/src/k8s.io/apiserver/pkg/server/options/server_run_options.go#L88

Then when `Complete` is called on the config, the `ExternalHost` is parsed to see if a port is in it:

https://github.com/kcp-dev/kubernetes/blob/10415751c5a9f503df2dca997114a2d4de855340/staging/src/k8s.io/apiserver/pkg/server/config.go#L482-L483

This implies that a port _can_ be included in the `--external-hostname` flag.

pkg/server/options/authentication.go
sttts: ic.

But still wondering what we actually want here. Why again do we need the same kubeconfig with external and internal address? Why not always the external one?

We had a discussion about a second kubeconfig for other components, with only one context. This PR is not that, is it?

kylape: Wrote up [a comment on the linked issue](https://github.com/kcp-dev/kcp/issues/707#issuecomment-1073411354) to try and explain my thoughts as I worked on this PR.

PR #746, opened 2022-03-18
Add logging options to virtual workspaces server

csams:
## Summary
Add logging options to the virtual workspaces server.

### DIFF
--- a/cmd/virtual-workspaces/command/cmd.go
+++ b/cmd/virtual-workspaces/command/cmd.go
@@ -22,7 +22,6 @@ import (
 	"io"
 	"io/ioutil"
 	"net/url"
-	"os"
 	"time"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


cmd/virtual-workspaces/command/cmd.go
```diff
@@ -60,6 +60,10 @@ func NewCommand(out, errout io.Writer, stopCh <-chan struct{}) *cobra.Command {
 		Long:  "Start a virtual workspace apiserver to managing personal, shared or organization workspaces",
 
 		Run: func(c *cobra.Command, args []string) {
+			if err := opts.Logs.ValidateAndApply(); err != nil {
+				klog.Fatal(err)
```
sttts, line 64: this print stack trace. Better print the error normally, and then os.Exit

sttts: better: use the RunE variant for the command and just return the error, also below

csams: good call. I converted to RunE but kept the block that prints details if the error was invalid. Is that still useful, or should I remove it, too?

sttts: honestly I don't understand why that code is there. Would remove it.

csams: Done

sttts (approved):

PR #747, opened 2022-03-18
crd bootstrap: better goroutine support

ncdc:
## Summary
Add support for multiple goroutines (e.g. post start hooks) that try to
install the same CRD at the same time. There was a race where one
goroutine checked for the CRD's existence, didn't find it, and proceeded
to create it. Meanwhile, another goroutine created it. Before this
change, the first goroutine would return an error ("already exists").
Now, it recognizes the existence and tries to do an update instead of
returning an error.

## Related issue(s)

Fixes #

### DIFF
--- a/config/crds/bootstrap.go
+++ b/config/crds/bootstrap.go
@@ -104,19 +104,36 @@ func createSingleFromFS(ctx context.Context, client apiextensionsv1client.Custom
 		return fmt.Errorf("decoded CRD %s into incorrect type, got %T, wanted %T", gr.String(), rawCrd, &apiextensionsv1.CustomResourceDefinition{})
 	}
 
-	crdResource, err := client.Get(ctx, rawCrd.Name, metav1.GetOptions{})
+	updateNeeded := false
+	crd, err := client.Get(ctx, rawCrd.Name, metav1.GetOptions{})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


config/crds/bootstrap.go
```diff
@@ -104,19 +104,36 @@ func createSingleFromFS(ctx context.Context, client apiextensionsv1client.Custom
...
 			return fmt.Errorf("error fetching CRD %s: %w", gr.String(), err)
 		}
 	} else {
-		rawCrd.ResourceVersion = crdResource.ResourceVersion
+		updateNeeded = true
+	}
+
+	if updateNeeded {
+		rawCrd.ResourceVersion = crd.ResourceVersion
 		crd, err := client.Update(ctx, rawCrd, metav1.UpdateOptions{})
```
stevekuznetsov, line 137: Is all of this dance necessary if we just use SSA?

sttts: yes, SSA is a good idea

sttts: PR welcome if you want to do something non-crdb ;-)

sttts: SSA is follow-up.

sttts (approved):

PR #749, opened 2022-03-18
Update namespace scheduler to enqueue resources on scheduling change

marun:
Previously the scheduler enqueued resources contained by a namespace when the namespace was reconciled for any reason. This change updates the scheduler to keep track of namespace scheduling decisions and only enqueue contained resources if the scheduling is changed by the controller.

### DIFF
--- a/pkg/reconciler/workload/namespace/namespace_controller.go
+++ b/pkg/reconciler/workload/namespace/namespace_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/kcp-dev/apimachinery/pkg/logicalcluster"
@@ -87,6 +88,8 @@ func NewController(
 		clusterLister:   clusterLister,
 		namespaceLister: namespaceLister,
 		kubeClient:      kubeClusterClient,
+
+		namespaceContentsEnqueuedForMap: map[string]string{},
 	}
 	clusterInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc:    func(obj interface{}) { c.enqueueCluster(obj) },
@@ -98,7 +101,17 @@ func NewController(
 		Handler: cache.ResourceEventHandlerFuncs{
 			AddFunc:    func(obj interface{}) { c.enqueueNamespace(obj) },
 			UpdateFunc: func(_, obj interface{}) { c.enqueueNamespace(obj) },
-			DeleteFunc: nil, // Nothing to do.
+			DeleteFunc: func(obj interface{}) {
+				// Ensure the scheduling cache entry for the namespace is removed
+				key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+				if err != nil {
+					runtime.HandleError(err)
+					return
+				}
+				c.namespaceContentsEnqueuedForLock.Lock()
+				defer c.namespaceContentsEnqueuedForLock.Unlock()
+				delete(c.namespaceContentsEnqueuedForMap, key)
+			},
 		},
 	})
 	workspaceInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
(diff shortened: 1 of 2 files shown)


pkg/reconciler/namespace/controller.go
```diff
@@ -132,6 +145,11 @@ type Controller struct {
 	kubeClient      kubernetes.ClusterInterface
 	ddsif           informer.DynamicDiscoverySharedInformerFactory
 	gvkTrans        *gvk.GVKTranslator
+
+	// Mapping of namespace key to the last scheduling decision for
```
sttts, line 149: "last scheduling decision" = a pcluster name?

pkg/reconciler/namespace/namespace.go
```diff
@@ -187,6 +187,11 @@ func (c *Controller) ensureScheduled(ctx context.Context, ns *corev1.Namespace)
 	patchType, patchBytes := clusterLabelPatchBytes(newPClusterName)
 	_, err = c.kubeClient.Cluster(ns.ClusterName).CoreV1().Namespaces().
 		Patch(ctx, ns.Name, patchType, patchBytes, metav1.PatchOptions{})
+	if err != nil {
+		// Update the label to enable the caller to detect a scheduling change.
+		ns.Labels[ClusterLabel] = newPClusterName
```
sttts, line 192: don't we get the namespace back in the Patch call? Would rather return that as it is guaranteed to be the real truth.

pkg/reconciler/namespace/namespace.go
```diff
@@ -269,9 +280,35 @@ func (c *Controller) reconcileNamespace(ctx context.Context, lclusterName string
...
+// namepaceContentsEnqueuedFor retrieves the last scheduling decision
+// for which the contents of the given namespace were successfully
+// enqueued.
+func (c *Controller) namepaceContentsEnqueuedFor(ns *corev1.Namespace) (string, bool) {
+	key := clusters.ToClusterAwareKey(ns.ClusterName, ns.Name)
+	c.namespaceContentsEnqueuedForLock.RLock()
+	defer c.namespaceContentsEnqueuedForLock.RUnlock()
+	scheduling, ok := c.namespaceContentsEnqueuedForMap[key]
+	return scheduling, ok
+}
```
sttts, line 300: these funcs with local locking make me nervous that we might get races due to multiple go routines influencing each other? E.g. the `DeleteFunc` definitely runs outside of the workers.

sttts: One could move the `DeleteFunc` logic into the reconciler too. Per ns key there is only one go routine operating on it at a given time.

pkg/reconciler/namespace/controller.go
```diff
@@ -98,7 +101,17 @@ func NewController(
 		Handler: cache.ResourceEventHandlerFuncs{
 			AddFunc:    func(obj interface{}) { c.enqueueNamespace(obj) },
 			UpdateFunc: func(_, obj interface{}) { c.enqueueNamespace(obj) },
-			DeleteFunc: nil, // Nothing to do.
+			DeleteFunc: func(obj interface{}) {
+				// Ensure the scheduling cache entry for the namespace is removed
+				key, err := cache.MetaNamespaceKeyFunc(obj)
```
ncdc, line 106: Safer to use DeletionHandlingMetaNamespaceKeyFunc here

pkg/reconciler/namespace/namespace.go
ncdc: I'm also curious about this scenario - if the patch fails, we're forcing the ClusterLabel to be what was calculated in this func, but failed to patch?

pkg/reconciler/namespace/namespace.go
ncdc: DeleteFunc has locking - is this comment outdated?

pkg/reconciler/namespace/controller.go
marun: For now, yes.

pkg/reconciler/namespace/namespace.go
marun: Done.

pkg/reconciler/namespace/controller.go
marun: Done.

marun:
Rebased and updated.

pkg/reconciler/namespace/namespace.go
marun: Please restate your request, I don't understand what action is implied.

pkg/reconciler/namespace/namespace.go
ncdc: Do you mean `err == nil`?

marun: Yes! `<facepalm>`
Done.

ncdc (approved):

ncdc:
@imjasonh do you have time to 👀 ?

imjasonh (approved):

PR #753, opened 2022-03-22
WIP: e2e: run via go-run by default, GORUN_FIXTURE=0 for opt-out from Makefile

sttts:


sttts:
Blocked on https://github.com/golang/go/issues/40467.

marun:
Is it necessary or desirable long-term that our e2e suite be responsible for starting servers? vs requiring test servers be started before the test run, which presumably could use go run and maybe not run into this issue?

test/e2e/framework/fixture.go
```diff
@@ -137,6 +137,11 @@ func InProcessEnvSet() bool {
 	return inProcess
 }
 
+func LogToConsoleEnvSet() bool {
+	inProcess, _ := strconv.ParseBool(os.Getenv("LOG_TO_CONSOLE"))
```
marun, line 141: nit: maybe rename the var to something either more or less specific?

marun, line 141: nit: maybe rename the var to something either more or less specific?

Issue #417, opened 2022-02-03
Prototype 3: API Author Demo


pweil-:
# Demo Objective
Close over consumer details, evolve APIs

1. For ingress (shouldn’t ingress have a wildcard?)
1. For serving (reencrypt)
1. For talking to another service

# Demo Steps
1. Create new workspace for our new cert-manager API backend
1. Run cert-manager (as Deployment if possible, locally is also fine because not focus of this epic) with a kubeconfig of a kcp controller workspace.
1. Export the cert-manager APIs (those needed to issue a serving cert)
1. In another workspace, import the cert-manager APIs from above
1. Show webapp getting a certificate within the workspace
1. Stetch goal: **A compatible API schema change is deployed**
1. Stetch goal: **See the workspaces to see the new schema automatically**

# Action Items

- [x] Scope the current demo as necessary to fit in prototype boundaries
- [x] Create and link required tasks to realize demo
- [x] https://github.com/kcp-dev/kcp/pull/583 @sttts 
- [x] automatically install CRDs into Universal workspaces
- [x] Create APIBindingController: @ncdc #755
  - create a CRD from APIResourceSchema in shadow workspace
  - phase transitions of the APIBinding
  - follow-ups:
    - ~~[ ] set the storage versions in the APIBinding's status @ncdc~~ moved to https://github.com/kcp-dev/kcp/issues/842
    - [x] stop unbinding, only unbinding we want is deletion, see https://github.com/kcp-dev/kcp/pull/702/files#r830041070
    - [x] ~~replace CRDReady condition with Established, see https://github.com/kcp-dev/kcp/pull/702#discussion_r830022232~~
    - ~~[ ] have more speaking names of CRDs in shadow namespace, see https://github.com/kcp-dev/kcp/pull/702#discussion_r830045834~~ moved https://github.com/kcp-dev/kcp/issues/843
     - [x] consolodate system CRD installation in couple of post-start-hooks and by system CRD provider, see kcp-install-api-resource-controller
    - ~~[ ] add admission to only allow `apis.kcp.dev/bound-crd` annotation in `system:bound-crds` workspace https://github.com/kcp-dev/kubernetes/pull/48/files#r827736066~~ moved to https://github.com/kcp-dev/kcp/issues/844
    - ~~[ ]  stop CRDs and APIBindings to add the same GVRs~~ moved to https://github.com/kcp-dev/kcp/issues/748
[...]

PR #763, opened 2022-03-23
apibinding demo

sttts:
Part of #417.

Co-authored by @varshaprasad96.

### DIFF
--- a/contrib/demo/prototype3-script/script-apibindings.sh
+++ b/contrib/demo/prototype3-script/script-apibindings.sh
@@ -53,7 +53,7 @@ clear
 pe "kubectl config use-context root"
 pe "kubectl kcp workspace use demo"
 
-c "Now we swtich roles. We are the customer now, in need of certificate generation for our web application. We are lucky: kcp has a cert-manager service 🎉"
+c "Now we switch roles. We are the customer now, in need of certificate generation for our web application. We are lucky: kcp has a cert-manager service 🎉"
 pe "kubectl kcp workspace create webapp"
 pe "kubectl kcp workspace use webapp"
 
--- a/contrib/demo/prototype3-script/script-apibindings.sh
+++ b/contrib/demo/prototype3-script/script-apibindings.sh
@@ -1,0 +1,145 @@
+#!/usr/bin/env bash
+
+# Copyright 2022 The KCP Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+DEMO_DIR="$(dirname "${BASH_SOURCE[0]}")"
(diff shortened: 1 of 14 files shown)


contrib/demo/prototype3-script/script-apibindings.sh
```diff
@@ -0,0 +1,140 @@
...
+c "Let's wait for the deployment to be running in kind"
+pe "while ! kubectl --kubeconfig ${CLUSTERS_DIR}/us-east1.kubeconfig wait --for=jsonpath='{.status.availableReplicas}'=1 -A -l 'workloads.kcp.dev/cluster=kind-us-east1' --field-selector 'metadata.name=kuard' deployments; do
+  sleep 1
+done"
+
+c "Let's describe it to see its status - take note of Replicas"
+pe "kubectl describe deployment/kuard"
+
+c "Does it actually serve with cert-manager-produced certs? We will find out."
+pe ""
```
varshaprasad96, line 140: The last few steps are:
1. Applying this service:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: kuard
spec:
  selector:
    app: kuard
  ports:
    - protocol: TCP
      port: 9376
      targetPort: 8080
```

2. Make sure the service is created:
```
pe "kubectl get svc"
```

3. Port forward to expose the web app
```
kubectl port-forward service/kuard 9376:9376
``` 

4. `curl --cert ${KCP_DATA_DIR}/ca.crt --key ${KCP_DATA_DIR}/ca.key http://127.0.0.1:9376 -v`
We are able to reach the server. You should be able to see a `200 OK` in the verbose output to confirm :)

sttts: Was a little harder. The port-forward didn't work because we don't know the namespace name on the workload cluster (finding out about it would be cheating).

The kcp service does not support port-forward yet.

So with @jmprusi's help (thank you!) I switched to nginx ingress, which is deployed anyway. After enabling ssl-passthrough in the nginx deployment, and after fixing the curl flags to use `--cacert` and `--resolve` to workaround my broken internet provider (not allowing 127.0.0.1 from nip.io) it started working 🎉

fabianvf (approved):
Awesome that it's working with ingress, haven't pulled it down to run these updates yet but if it's working for you lgtm

njhale (commented):
Looking good @sttts!

Do we also want a make target for prototype 3 (like we have for the other demos)? If so, I think we'll need to rename `script-apibindings.sh` to `script` to match the convention used by `runDemoScripts.sh`.

FYI: when I added the make target and tried to run this locally (macos/zsh), I ran into issues with the generated KCP kubeconfig file paths; I'm currently looking into it further.

contrib/demo/prototype3-script/script-apibindings.sh
```diff
@@ -0,0 +1,145 @@
...
+pe "kubectl apply -f ${DEMO_DIR}/apibinding/cert-manager-apiexport.yaml"
+c "Note that it references our APIResourceSchema, more concretely today's version. In the future, we might want to have another version and roll that out to customers."
+
+wait
+clear
+
+pe "kubectl config use-context root"
+pe "kubectl kcp workspace use demo"
+
+c "Now we swtich roles. We are the customer now, in need of certificate generation for our web application. We are lucky: kcp has a cert-manager service 🎉"
```
ncdc, line 56:
```suggestion
c "Now we switch roles. We are the customer now, in need of certificate generation for our web application. We are lucky: kcp has a cert-manager service 🎉"
```

ncdc (approved):

PR #765, opened 2022-03-23
reconciler/namespace: wire metadata client and allow differing schemas for PartialObjectMetadata wildcard requests

sttts:
Like #764, but without APIBindings, whichever merges first.


csams (approved):
All related errors are gone. Thanks!

PR #766, opened 2022-03-24
e2e: unify make-build bin/ dir and e2e search path for binaries

sttts:
How many times did I miss that I have to `go install` instead of `make build` to run an e2e test via `go test`. This is probably the reason for half of the times where we I added `RunInProcess: true` to the fixture because without the test behaves differently in Goland than via `make test-e2e`. This PR unifies that by basing both on `<repo-dir>/bin`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ E2E_PARALLELISM ?= 1
 
 .PHONY: test-e2e
 test-e2e: WHAT ?= ./test/e2e...
-test-e2e: install
+test-e2e: build
 	go test -race -count $(COUNT) -p $(E2E_PARALLELISM) -parallel $(E2E_PARALLELISM) $(WHAT) $(TEST_ARGS)
 
 .PHONY: test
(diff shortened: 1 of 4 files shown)


stevekuznetsov:
Why not do this in the other direction? Having to add `$REPO/bin` to `$PATH` for every repo is pretty annoying, and makes it hard to e.g. use `kcp` binary in `kubernetes` checkout, etc. `go install` is the normal way to create binaries on the system for reuse ...

sttts:
I still have problems with installing "for every repo". I find that annoying. Tests should work without tainting my system.

sttts:
Is there a way to reference the `go build` output somehow?

sttts:
Btw, there is no need to put `bin` into your path. And you can of course call `make install` instead of `make` if you prefer binaries installed. My point here in the PR is that a test should not have side-effects, e.g. by installing binaries in the system. Does this make sense @stevekuznetsov ?

marun (approved):

PR #767, opened 2022-03-24
kubectl-kcp: add create --type flag

sttts:
Like in `kubectl kcp ws create foo --type Organization`.


davidfestal (approved):
Thanks !

Just a question: Is `Universal` the default workspace type is not type is specified ?

sttts:
Yes, `Universal` is the default, done through CRD defaulting.

sttts:
https://github.com/kcp-dev/kcp/blob/main/pkg/apis/tenancy/v1alpha1/types.go#L83

Issue #415, opened 2022-02-03
Prototype 3: Transparent Multi-Cluster End-User Demo


pweil-:
### Prototype 3 Themes
* Use Workspaces and API inheritance to run real apps
* Real apps can run in multiple Locations
* Apps have resilience to compute failure

### Demo Objective
User has a multi-cluster placeable application that can move transparently

### Demo Steps
1. User begins journey with a KCP URL of a hosted KCP instance.
1. User `kubectl krew`-installs kcp plugin.
1. User logs into the cluster with `kubectl kcp login <url>` and some OIDC authn (e.g. Github)
1. User lands in a default workspace.
1. User creates a stateless web application
1. User decides to migrate the app between Locations for Reasons 
    1. Option 1: Admin is upgrading the cluster with downtime
    1. Option 2: Cluster is out of capacity
    1. Option 3: Cluster hard fails
1. Client-facing traffic sees no interruption (?)
1. Client has a multi-location security problem (needs a cert to access some external resource), THUS needs the API End-User Demo

### Action Items

- [ ] Scope the current demo as necessary to fit in prototype boundaries
- [ ] Create and link required tasks to realize demo
- [ ] Contribute to final demo script and recording

pweil-:
Notes from discussion:

> User logs into the cluster with kubectl kcp login <url> and some OIDC authn (e.g. Github)

if there is an existing kubectl login plugin we should just use it

marun:
> User decides to migrate the app between Locations for Reasons
Option 1: Admin is upgrading the cluster with downtime
Option 2: Cluster is out of capacity
Option 3: Cluster hard fails

Do these specific options matter for the purposes of the demo? vs just forcing a move from one cluster to another? afaik today we aren't even targeting allowing a 'User' (vs an admin) to choose a cluster to deploy to let alone migrate between - workload is just deployed to available compute (ala the 'Transparent' adjective of TCM).

> Client has a multi-location security problem (needs a cert to access some external resource), THUS needs the https://github.com/kcp-dev/kcp/issues/416

Any context on why this is part of this flow?

marun:
> Client-facing traffic sees no interruption (?)

@jmprusi Thoughts as to what will be required to demo this? I know this can work with a cloud lb, but will that also be required for ci or is there a lighter-weight way to ensure this flow is tested once supported?

marun:
[...]

PR #774, opened 2022-03-25
Add demo script for cordon and evict behavior

imjasonh:
## Summary

Adds demo script to show workload cluster cordoning (`unschedulable: true`) and immediate eviction (`evictAfter: $date-in-the-past`).

## Related issue(s)

#415

### DIFF
--- a/contrib/demo/prototype3-script/deployment-kuard.yaml
+++ b/contrib/demo/prototype3-script/deployment-kuard.yaml
@@ -1,0 +1,22 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: kuard
+  name: kuard
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: kuard
+  template:
+    metadata:
+      labels:
+        app: kuard
+    spec:
+      containers:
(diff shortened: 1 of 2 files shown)


marun:
This would appear to be duplicating the exercise of functionality we're already validating in unit and e2e tests. If we're past the prototype phase and want to best spend our effort, maybe demoing effort is better spent writing documentation for features we implement rather than effectively duplicating testing effort?

marun:
cc @pweil-

pweil-:
>This would appear to be duplicating the exercise of functionality we're already validating in unit and e2e tests. If we're past the prototype phase and want to best spend our effort, maybe demoing effort is better spent writing documentation for features we implement?

I think we should discuss if we've passed the prototype phase with P3 in a community call and what that would even mean.  I support the idea that we should focus on documentation vs codified demos if the goal of the effort is changing (regular, community call demos still encouraged).  I don't think it has, though.  Worth discussing.

imjasonh:
Added to this week's agenda: https://github.com/kcp-dev/kcp/issues/786

sttts:
Seeing ingress e2e flakes. Maybe rebase. Those flakes should be gone in main.

sttts (approved):

PR #779, opened 2022-03-25
6. kubectl-kcp: add create-context command

sttts:
TODO:
- [x] unit tests

### DIFF
--- a/pkg/cliplugins/workspace/cmd/cmd.go
+++ b/pkg/cliplugins/workspace/cmd/cmd.go
@@ -49,6 +49,12 @@ var (
 
 	# create a workspace and immediately enter it
 	%[1]s workspace create my-workspace --use
+
+	# create a context with the current workspace, e.g. root:default:my-workspace
+	%[1]s workspace create-context
+
+	# create a context with the current workspace, named context-name
+	%[1]s workspace create-context context-name
 `
 )
 
@@ -78,7 +84,7 @@ func NewCmdWorkspace(streams genericclioptions.IOStreams) (*cobra.Command, error
 
 	cmd := &cobra.Command{
 		Aliases:          []string{"ws", "workspaces"},
-		Use:              "workspace [list|create|<workspace>|..|-|<root:absolute:workspace>]",
+		Use:              "workspace [list|create|create-context|<workspace>|..|-|<root:absolute:workspace>]",
 		Short:            "Manages KCP workspaces",
 		Example:          fmt.Sprintf(workspaceExample, "kubectl kcp"),
 		SilenceUsage:     true,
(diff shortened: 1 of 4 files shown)


pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -406,3 +406,52 @@ func (kc *KubeConfig) ListWorkspaces(ctx context.Context, opts *Options) error {
...
+
+	if err := kc.modifyConfig(newKubeConfig); err != nil {
+		return err
+	}
+
+	if existedBefore {
+		if kc.startingConfig.CurrentContext == name {
+			fmt.Fprintf(kc.Out, "Updated context %q.\n", name) // nolint: errcheck
+		} else {
+			fmt.Fprintf(kc.Out, "Updated context %q and switched to it.\n", name) // nolint:	errcheck
```
ncdc, line 450: Should we ask the user if they want to overwrite a preexisting non-current context, or make them specify `--overwrite`? e.g. I have a context called `prod`, I switch my current workspace to `test`, and then do `k kcp create-context prod`, I probably don't actually want to do that?

pkg/cliplugins/workspace/plugin/kubeconfig_test.go
```diff
@@ -701,6 +702,204 @@ func TestUse(t *testing.T) {
...
+				currentContext: tt.config.CurrentContext,
+				overrides:      tt.overrides,
+
+				modifyConfig: func(config *clientcmdapi.Config) error {
+					got = config
+					return nil
+				},
+				IOStreams: streams,
+			}
+			err := kc.CreateContext(context.TODO(), tt.param)
```
ncdc, line 873: `context.Background()` is probably better here, as `TODO()` is more for when you need a context, but don't have a way to pass one in (i.e. you intend to come back and plumb it).

pkg/cliplugins/workspace/plugin/kubeconfig.go
sttts: Flag seems to be fine. Has there been anything interactive in kubectl and oc before?

pkg/cliplugins/workspace/plugin/kubeconfig_test.go
sttts: done

pkg/cliplugins/workspace/plugin/kubeconfig.go
ncdc: I would do what `kubectl label` does:

```
 If --overwrite is true, then existing labels can be overwritten, otherwise attempting to overwrite a label will
result in an error.
```

pkg/cliplugins/workspace/cmd/cmd.go
```diff
@@ -162,6 +164,35 @@ func NewCmdWorkspace(streams genericclioptions.IOStreams) (*cobra.Command, error
...
+
+			arg := ""
+			if len(args) == 1 {
+				arg = args[0]
+			}
+
+			return kubeconfig.CreateContext(cmd.Context(), arg, overwriteContext)
+		},
+	}
+	createCmd.Flags().BoolVar(&overwriteContext, "overwrite", overwriteContext, "Overwrite the context if it already exists")
```
ncdc, line 194:
```suggestion
	createContextCmd.Flags().BoolVar(&overwriteContext, "overwrite", overwriteContext, "Overwrite the context if it already exists")
```

sttts: fixed already.

ncdc (approved):

Issue #739, opened 2022-03-17
Test flake (e2e-multiple-runs): TestIngressController/ingress_lifecycle


ncdc:
From #702 (it's possibly my changes created a flake scenario, although I don't think so...?)

https://github.com/kcp-dev/kcp/runs/5586190871?check_suite_focus=true

```
--- FAIL: TestIngressController (0.00s)
    --- FAIL: TestIngressController/ingress_lifecycle (51.10s)
        util.go:119: Saving test artifacts for test "TestIngressController/ingress_lifecycle" under "/tmp/e2e/TestIngressController/ingress_lifecycle/556494561".
        fixture.go:101: Starting kcp servers...
        kcp.go:169: running: /home/runner/go/bin/kcp start --root-directory /tmp/e2e/TestIngressController/ingress_lifecycle/556494561/data/kcp/source --secure-port=38121 --embedded-etcd-client-port=43035 --embedded-etcd-peer-port=44417 --embedded-etcd-wal-size-bytes=5000 --kubeconfig-path=admin.kubeconfig --push-mode --auto-publish-apis=true --resources-to-sync=ingresses.networking.k8s.io,deployments.apps,services --discovery-poll-interval=5s
        kcp.go:169: running: /home/runner/go/bin/kcp start --root-directory /tmp/e2e/TestIngressController/ingress_lifecycle/556494561/data/kcp/sink --secure-port=35907 --embedded-etcd-client-port=39595 --embedded-etcd-peer-port=33969 --embedded-etcd-wal-size-bytes=5000 --kubeconfig-path=admin.kubeconfig --run-controllers=false
        kcp.go:396: success contacting https://10.1.0.18:38121/readyz
        kcp.go:396: success contacting https://10.1.0.18:38121/livez
        kcp.go:396: success contacting https://10.1.0.18:35907/livez
        kcp.go:396: success contacting https://10.1.0.18:35907/readyz
        fixture.go:128: Started kcp servers after 12.487015693s
        controller_test.go:183: Installing test CRDs into source cluster...
        controller_test.go:197: Installing test CRDs into sink cluster...
        controller_test.go:205: Installing sink cluster...
        controller_test.go:209: Installed sink cluster after 64.088108ms
        controller_test.go:211: Creating namespace in source cluster...
        controller_test.go:217: Creating service in source cluster...
        controller_test.go:250: Starting ingress-controller...
[...]

PR #784, opened 2022-03-26
0. ingress-controller: fix race leading to too many leaves when informers are stale

sttts:
ingress-controller used generated-name, i.e. without any guarantee that there is only one leaf per workload cluster. Moreover, the controller didn't delete the leaves that are too many. Finally, the e2e test waits for exactly one 🤷‍♂️ 

This PR switches to `<root-name>-<cluster-name>`.

Also: revert 034f5c4f0 which was wrong. The list is a string list.


slaskawi (approved):
/lgtm

s-urbaniak:
lgtm

marun:
@sttts Does this fix #739 ?

PR #785, opened 2022-03-28
authorization: give everybody authenticated root org access

sttts:
This is preprartion for the virtual workspace apiserver to do proper authorization, and also for users to access the root and see their org.


s-urbaniak (approved):
lgtm

PR #787, opened 2022-03-29
server: send warning header to client on /clusters/<org>:<workspace> access

sttts:
```
$ kubectl kcp workspace create foo <- old kubectl plugin
W0329 10:22:54.753682   33759 warnings.go:70] the /clusters/<org>:<cluster> URL pattern is deprecated. Update your kubeconfig and use /clusters/root:<org>:<cluster> instead.
Workspace "foo" (type "Universal") created.
```

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -30,6 +30,7 @@ import (
 	apiextensionsexternalversions "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apiserver/pkg/admission"
+	"k8s.io/apiserver/pkg/endpoints/filters"
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	"k8s.io/apiserver/pkg/util/webhook"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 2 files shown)


pkg/server/server.go
```diff
@@ -225,6 +226,10 @@ func (s *Server) Run(ctx context.Context) error {
 		apiHandler = WithClusterScope(genericapiserver.DefaultBuildHandlerChain(apiHandler, c))
 		apiHandler = WithInClusterServiceAccountRequestRewrite(apiHandler, unsafeServiceAccountPreAuth)
 
+		// this will be replace in DefaultBuildHandlerChain. So at worst we get twice as many warning.
```
davidfestal, line 229:
```suggestion
		// this will be replaced in DefaultBuildHandlerChain. So at worst we get twice as many warning.
```

pkg/server/server.go
```diff
@@ -225,6 +226,10 @@ func (s *Server) Run(ctx context.Context) error {
...
+		// But this is not harmful as the kcp warning are not many.
```
davidfestal, line 230:
```suggestion
		// But this is not harmful as the kcp warnings are not many.
```

davidfestal (approved):

PR #791, opened 2022-03-29
e2e: Use ./bin/kcp instead of finding it from the path

marun:
This ensures that a test run will use the most recent built artifact
rather than a potentially stale binary in the path somewhere.


sttts (approved):

marun:
Nope, that doesn't work. The path varies by the test being invoked.

PR #792, opened 2022-03-29
Update manifest to persist KCP configs between restarts

kylape:
Currently the admin token and the sa.key are getting regenerated every time the kcp pod restarts. We need to start persisting this, especially because folks are starting to use service account tokens.

This also includes a commit to change the update strategy to recreate (this helps with the PVC getting stuck upon rollout).

### DIFF
--- a/manifest/kcp.yaml
+++ b/manifest/kcp.yaml
@@ -82,6 +82,8 @@ spec:
   selector:
     matchLabels:
       app: kcp
+  strategy:
+    type: Recreate
   template:
     metadata:
       labels:
--- a/manifest/kcp.yaml
+++ b/manifest/kcp.yaml
@@ -1,4 +1,17 @@
 ---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: kcp
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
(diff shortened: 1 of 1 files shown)


csams (approved):
lgtm

manifest/kcp.yaml
```diff
@@ -1,4 +1,17 @@
 ---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: kcp
+spec:
+  accessModes:
+  - ReadWriteOnce
```
sttts, line 8: This is fine for now.

We might want to start thinking about a multi-shard deployment. Then we likely have to put the service account keys into a secret and share that secret between shards.

sttts: We might also want to have service account key per shard, but distribute trust (some CA? Does that work with service accounts? cc @marun).

sttts (approved):

manifest/kcp.yaml
kylape: https://github.com/kcp-dev/kcp/issues/799

marun: @sttts Token signing keys are rsa, so each server would either share the same key or have an independent key. There is no option for a certificate-based approach with CA afaik.

PR #794, opened 2022-03-30
e2e: cleanup misleading log output from apibinding test

marun:
cc: @sttts @ncdc


sttts (approved):

ncdc:
Good catch, thanks!!!

PR #804, opened 2022-03-30
e2e: Add --use-default-server to simplify use of persistent fixture

marun:
Rather than requiring developers to supply the qualified path to
.kcp/admin.kubeconfig to run against shared fixture, rely on the fact
that the repository root is discoverable to allow --use-default-server
to supply the same configuration.

### DIFF
--- a/test/e2e/framework/fixture.go
+++ b/test/e2e/framework/fixture.go
@@ -64,16 +64,17 @@ type KcpFixture struct {
 
 // SharedKcpServer returns a kcp server fixture intended to be shared
 // between tests. A persistent server will be configured if
-// `--kubeconfig` is supplied to the test runner. Otherwise a
-// test-managed server will be started. Only tests that are known to
-// be hermetic are compatible with shared fixture.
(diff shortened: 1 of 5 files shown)


test/e2e/framework/fixture.go
```diff
@@ -64,16 +64,17 @@ type KcpFixture struct {
 
 // SharedKcpServer returns a kcp server fixture intended to be shared
 // between tests. A persistent server will be configured if
-// `--kubeconfig` is supplied to the test runner. Otherwise a
-// test-managed server will be started. Only tests that are known to
-// be hermetic are compatible with shared fixture.
+// `--kubeconfig` or `--use-default-config` is supplied to the test
```
ncdc, line 67:
```suggestion
// `--kubeconfig` or `--use-default-server` is supplied to the test
```

marun: Good catch, done!

ncdc (approved):

Issue #796, opened 2022-03-30
apibinding e2e test fails if run more than once against a persistent server


marun:
The [apibinding e2e test](https://github.com/kcp-dev/kcp/blob/main/test/e2e/apibinding/apibinding_test.go) fails if run more than once against a persistent server. This is curious, since the testing appears to be isolated within test-local organizations and workspaces.

**To Reproduce**
1. Start a persistent test server in one terminal:
```
$ rm -rf .kcp && make build && ./bin/test-server 2>&1 | tee kcp.log
```
2. Run the test in another terminal (it should pass):
```
 go test -v  test/e2e/apibinding/apibinding_test.go -args --kubeconfig=$(pwd)/.kcp/admin.kubeconfig
```
3. Run the same test again. It should stall after logging something like the following:
```
  apibinding_test.go:109: Make sure "<group>" API group shows up in workspace "<workspace name>" group discovery
```

ncdc:
Thanks for filing! Here is what is happening:

1. Test run 1 creates org 1 with all the required resources (2 workspaces, apiresourceschema, apiexport, apibinding)
2. A CRD is created in system:bound-crds corresponding to the apiresourceschema for this org 1
3. Test run 2 creates org 2 with the same resources
4. A 2nd CRD is created in system:bound-crds (different name) corresponding to the apiresourceschema from org 2
5. The 2nd CRD has these conditions:

```
  - lastTransitionTime: "2022-03-30T15:24:16Z"
    message: '"CowboyList" is already in use'
    reason: ListKindConflict
    status: "False"
    type: NamesAccepted
  - lastTransitionTime: "2022-03-30T15:24:16Z"
    message: not all names are accepted
    reason: NotAccepted
    status: "False"
    type: Established
```

ncdc:
@sttts I guess we need to patch more apiextensions controllers to relax restrictions when evaluating our generated CRDs, i.e. those with the `apis.kcp.dev/bound-crd` annotation?

marun:
It appears the `TestCrossLogicalClusterList` has a similar problem when run more than once against persistent fixture?

```
=== RUN   TestCrossLogicalClusterList
=== PAUSE TestCrossLogicalClusterList
=== RUN   TestCrossLogicalClusterListPartialObjectMetadata
=== PAUSE TestCrossLogicalClusterListPartialObjectMetadata
=== CONT  TestCrossLogicalClusterList
[...]

PR #809, opened 2022-03-30
e2e: Run e2e against shared server only a single time.

marun:
Multiple times is broken by #796 and even a broken non-gating CI job
has a negative impact. A fix for #796 should update the count back to
2.

-------


marun:
Rebased

sttts (approved):

Issue #730, opened 2022-03-16
API import does not work when adding the same cluster into multiple workspaces


astefanutti:
**Describe the bug**

Create a WorkloadCluster resource in a workspace. The API import works as expected.

Then create the same WorkloadCluster resource, in any other workspaces, the APIs are not imported from the cluster into these other workspaces.

**To Reproduce**

```sh
$ kubectl kcp workspace --token user-1-token create workspace1 --use
$ cat <<EOF | kubectl apply -f -
apiVersion: workload.kcp.dev/v1alpha1
kind: WorkloadCluster
metadata:
  name: kind-cluster
spec:
  kubeconfig: |
$(sed 's,^,    ,' cluster.kubeconfig)
EOF
workloadcluster.workload.kcp.dev/kind-cluster created
$ kubectl apply -f contrib/demo/prototype2-script/deployment-kuard.yaml
deployment.apps/deployment created

$ kubectl kcp workspace --token user-1-token create workspace2 --use
$ cat <<EOF | kubectl apply -f -
apiVersion: workload.kcp.dev/v1alpha1
kind: WorkloadCluster
metadata:
  name: kind-cluster
spec:
  kubeconfig: |
$(sed 's,^,    ,' cluster.kubeconfig)
EOF
workloadcluster.workload.kcp.dev/kind-cluster created
$ kubectl apply -f contrib/demo/prototype2-script/deployment-kuard.yaml
error: unable to recognize "contrib/demo/prototype2-script/deployment-kuard.yaml": no matches for kind "Deployment" in version "apps/v1"
```

**Expected behavior**

API import should work in the other workspaces.

**Additional context**

It is caused because the API importer are cached by cluster name key: https://github.com/kcp-dev/kcp/blob/62354fc9bf4f1e4d72f3a72d7f4613b182f0691c/pkg/reconciler/cluster/apiimporter/apiimporter_manager.go#L57

ncdc:
@astefanutti I see you're running against older code (Cluster instead of WorkloadCluster, etc.). I don't expect we've necessarily fixed things, but could you please try with the latest from main, making sure to install the latest kubectl-kcp plugin too? (Also with main, you don't need `--workspace-directory-insecure-skip-tls-verify` any more).

astefanutti:
@ncdc you're right, this is based on the stable `release-prototype-2` branch. However, I've checked in main and it seems the logic of keying API importers by cluster name is still there:

[...]

PR #810, opened 2022-03-31
Move api importer to the syncer

marun:
Another step towards deprecating syncer management. Next: provide an
easy path to manual deployment (i.e. generate a bunch of yaml).

Fixes: #730


ncdc:
Is the question "why" or ?

marun:
> I probably miss some context to really understand the reason for this change.

We are moving towards not requiring that kcp have network connectivity to workload clusters since that is unlikely to be a consistent possibility. All communication will be initiated from the syncer, including api importing.
 
> But currently, to test the syncer virtual workspace along with 2 external syncers (no pull and no push mode), I still benefit from the APIImport mechanism as soon as a WorkloadCluster is added manually, which is good IMO.

I encourage you to explore performing such testing with e2e fixture. All the tests requiring syncers have been updated to run the syncer (and with this PR, api import) in the same process as the test. This simplifies debugging and speeds up iteration (no need to redeploy kcp to get new behavior).

The method responsible for starting the syncer is `framework.StartWorkspaceSyncer`, e.g.

https://github.com/kcp-dev/kcp/blob/main/test/e2e/reconciler/ingress/controller_test.go#L194
 
> Would this change make it impossible to have the APIImport work even with a completely external syncer (ie neither push nor pull ) ?

As per above, the syncer running in e2e is already `external` in the sense that it is not managed by kcp at all.  Syncer management and push/pull are already deprecated, no longer tested in e2e, and will be removed as soon as we have a straightforward path to syncer deployment (e.g. kcp plugin is capable of generating yaml or directly updating the api with the necessary configuration).

marun:
I realize this work was a bit hasty. The crd pulling needs to be part of the syncer, but the rest needs to stay in kcp. Converting to draft and will fix.

davidfestal:
> What is a non-push, non-pull syncer?

I was referring to the case when you start KCP with neither the `--pull-mode` nor the `push-mode`.

marun:
Granted syncer permission for crds needed for api import to the example manifest.

marun:
As per offline conversation with @ncdc, we'd like to get this merged in more-or-less it's current form to allow syncer management to be retried. The expectation is that what is proposed here is a stop-gap, and that the implementation of api import can be revisited during the iteration on the syncer are implied by:
 
 - refactoring for increased test coverage
 - updating to target virtual workspaces
 - ???

@davidfestal Let's talk next week about how best to ensure you can easily start the api importer independent of the syncer for dev/test.

manifest/syncer.yaml
```diff
@@ -34,6 +34,14 @@ rules:
...
   verbs:
   - "*"
+- apiGroups:
+  - "apiextensions.k8s.io"
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - "get"
+  - "watch"
+  - "list"
```
ncdc, line 44: We also need apiresourceimports create/update/delete, right?

ncdc: Oh, this is for the pcluster?

ncdc:
Do we need to edit docs/syncer.md to create a clusterrole/binding for the syncer SA so it has the right permissions (e.g. get/list/watch/update/delete deployments, apiresourceimports, etc)?

manifest/syncer.yaml
marun: Yes, this is for the pcluster.

marun:
> Do we need to edit docs/syncer.md to create a clusterrole/binding for the syncer SA so it has the right permissions (e.g. get/list/watch/update/delete deployments, apiresourceimports, etc)?

I was assuming the syncer SA gets cluster admin for the workspace? Not sure we want to worry about fine-grained permission for the kcp side in advance of virtual workspace.   

For future on the pcluster side, I'm thinking we need a namespace syncer with cluster admin to manage namespaces and grant the resource syncer full access to all resources in the managed namespaces. I don't think we want to worry about granting permissions per resource type.

marun:
@ncdc Logs from a syncer built from this PR and deployed as per [documented deployment procedure](https://github.com/kcp-dev/kcp/blob/main/docs/syncer.md) (modulo actually specifying a command for the container): 

```
            I0403 07:22:13.135458       1 syncer.go:67] Syncing the following resource types: [deployments.apps ingresses.networking.k8s services]
            I0403 07:22:13.135949       1 specsyncer.go:75] Creating spec syncer for clusterName root:default:my-workspace to pcluster local, resources [deployments.apps ingresses.networking.k8s services
]
            E0403 07:22:13.146670       1 run.go:74] "command failed" err="The following resource types were requested to be synced, but were not found in the KCP logical cluster: [deployments.apps ingresses.networking.k8s services]"
```

The syncer crashloops with above error. Am I responsible for creating the types configured for syncing, or does the api importer need to be given a chance to ensure the resources marked for syncing are available before attempting to start the spec and status syncers?

marun:
Updating docs to reflect some of what I've been doing for local testing. I include these changes not because I think they should merge, but to ensure interested parties can replicate the behavior I'm seeing if so inclined.

docs/syncer.md
```diff
@@ -11,64 +11,104 @@ users have to install a special component named [syncer](https://github.com/kcp-
...
+$ export KCP_SERVER_URL=$(grep server .kcp/admin.kubeconfig | grep -v clusters | awk '{print $2}')
 ```
 
-Obtain the KCP server URL and the logical cluster name:
+1. Obtain the logical cluster name from `kubectl kcp workspace`:
 
-- KCP_SERVER_URL=https://192.168.1.150:6443
-- KCP_LOGICAL_CLUSTER_NAME=default:my-workspace--1
+```sh
+$ export KCP_LOGICAL_CLUSTER_NAME=$(kubectl kcp workspace | awk '{print $4}' | sed 's/".$/"/' | sed 's/"//g')
```
fgiloux, line 31: Don't you need "current" here?
kubectl kcp workspace current | awk '{print $4}' | sed 's/".$/"/' | sed 's/"//g'

ncdc: Not any more

fgiloux:
@marun I am surely missing something but I did not grok how the PR allows to match a WorkloadCluster/syncer to multiple workspaces, apart from creating an additional syncer for each workspace.
It seems [here](https://github.com/kcp-dev/kcp/pull/810/files#diff-0705d04ebc88b239145fe94b232d72bae8460b6cda37c12cd437ef611107be5cL108-L109) that there is a 1-1 match at the syncer level between a WorkloadCluster and a logical cluster. This is all good when the logical cluster is a virtual workspace but my understanding is that this will come later and that it points to a personal workspace in the meantime. What am I missing?

marun:
> @marun I am surely missing something but I did not grok how the PR allows to match a WorkloadCluster/syncer to multiple workspaces, apart from creating an additional syncer for each workspace.

I wasn't suggesting that this PR allows apis for a single cluster to be imported to multiple workspaces without multiple syncers. The problem reported in #730, though, seemed specific to the problem of multiple workspaces with the same cluster not resulting in multiple api importers. Since this PR localizes api import to each syncer it will ensure the behavior you proposed in #856 is retained.

fgiloux:
@marun thanks for the clarification.

marun:
Updated the e2e test to validate that the api importer is working as intended. I think this PR is ready for review and can be merged in advance of kind testing.

pkg/syncer/syncer.go
```diff
@@ -67,15 +70,65 @@ const SyncDown SyncDirection = "down"
...
+	if err != nil {
+		return err
+	}
+	fromDiscovery := discoveryClient.WithCluster(kcpClusterName)
+
+	// Block syncer start on gvr discovery completing successfully and
+	// including the resources configured for syncing. The spec and status
+	// syncers depend on the types being present to start their informers.
+	var gvrs []string
+	err = wait.PollImmediateInfinite(gvrQueryInterval, func() (bool, error) {
```
ncdc, line 101: Possible follow-up: set a condition on the WorkloadCluster that it's waiting for API import/discovery, for user visibility

test/e2e/syncer/syncer_test.go
```diff
@@ -0,0 +1,126 @@
...
+	require.NoError(t, err, "failed to read embedded deployment")
+
+	var deployment *appsv1.Deployment
+	err = yaml.Unmarshal(deploymentYAML, &deployment)
+	require.NoError(t, err, "failed to unmarshal deployment")
+
+	// Since this test created a new workspace that initially lacked support for
+	// deployments, successful creation of a deployment in the workspace proves that
+	// the api importer imported the downstream type to the upstream server.
+	upstreamDeployment, err := upstreamKubeClient.AppsV1().Deployments(upstreamNamespace.Name).Create(ctx, deployment, metav1.CreateOptions{})
```
ncdc, line 100: Any chance this could fail due to timing - does this need to be in a poll/eventually block?

docs/syncer.md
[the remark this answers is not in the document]
marun: I don't think this change should be included, actually. I will drop it.

test/e2e/syncer/syncer_test.go
marun: The syncer waits for the gvrs to become available in the logical cluster before heartbeating, and the fixture Start method is checking for a ready status that depends on the heartbeat.

marun:
Updated to drop syncer doc updates, those will either be included as part of #881 or dropped entirely (since the process will be automated).

test/e2e/syncer/syncer_test.go
marun: I'll update the comment to be clear about this.

marun: Updated, PTAL

pkg/syncer/syncer.go
marun: I'm not sure how critical this is yet - our current model of only syncing a few resource types takes on the order of seconds. Maybe more details status can wait until enough resources are synced that import takes on the order of 10s of seconds?

test/e2e/syncer/syncer_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	deploymentYAML, err := embeddedResources.ReadFile("deployment.yaml")
+	require.NoError(t, err, "failed to read embedded deployment")
+
+	var deployment *appsv1.Deployment
+	err = yaml.Unmarshal(deploymentYAML, &deployment)
+	require.NoError(t, err, "failed to unmarshal deployment")
+
+	// This test created a new workspace that initially lacked support for deployments, but once the
+	// workload cluster cluster went ready (checked by the syncer fixture's Start method) the api
```
ncdc, line 98:
```suggestion
	// workload cluster went ready (checked by the syncer fixture's Start method) the api
```

?

ncdc (approved):

pkg/syncer/syncer.go
ncdc: Totally fine!

test/e2e/syncer/syncer_test.go
marun: Thanks! Done.

PR #811, opened 2022-03-31
e2e: Update cross logical cluster testing to use private fixture

### LINKED ISSUE #802: e2e: TestCrossLogicalClusterList fails when run more than once against persistent fixture
The [TestCrossLogicalClusterList e2e test](https://github.com/kcp-dev/kcp/blob/main/test/e2e/conformance/cross_logical_cluster_list_test.go) fails if run more than once against a persistent server. This is curious, since the testing appears to be isolated within test-local organizations and workspaces.

**To Reproduce**
1. Start a persistent test server in one terminal:
```
$ rm -rf .kcp && make build && ./bin/test-server 2>&1 | tee kcp.log
```
2. Run the test in another terminal (it should pass):
```
 go test -v  test/e2e/conformance/cross_logical_cluster_list_test.go -args --kubeconfig=$(pwd)/.kcp/admin.kubeconfig
```
3. Run the same test again and it should quickly fail.

marun:
These tests operate against global state which precludes successive
invocations against shared fixture without error.

Fixes #802

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -143,7 +143,7 @@ jobs:
           ARTIFACT_DIR=/tmp/e2e \
           PATH="${PATH}:$(pwd)/bin/" \
           TEST_ARGS='-args --use-default-server' \
-          COUNT=2 \
+          COUNT=1 \
           E2E_PARALLELISM=2 \
           make test-e2e
       - uses: actions/upload-artifact@v2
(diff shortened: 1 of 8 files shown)


sttts:
I don't think this is necessary. Just make the global state unique enough to not fall over other parallel tests.

marun:
> I don't think this is necessary. Just make the global state unique enough to not fall over other parallel tests.

If you - or whoever wrote this test - want to fix it, by all means. I've already spent considerable effort getting e2e running with shared fixture and at this point my priority is getting the e2e-shared-server job to the point where it can prevent further regression. This PR accomplishes that, and in no way precludes future changes that would re-enable use of shared fixture.

FWIW, moving to private fixture also ensures these tests execute in a reasonable amount of time. When running against shared fixture even once, this test takes 45s on my machine. It takes <10s running independently.

Issue #759, opened 2022-03-23
Code-generate scoping wrappers for arbitrary k8s.io/code-generator generated clients and informers


sttts:
This epic is about generating wrappers for client-gen generated clients and informer-gen generated informers (i.e. those e.g. in k8s.io/client-go) which support scoping to a logical cluster, but **do not require changes to the clients and informers (or their generators) themselves**.

1. For clients this means that we want an interface generated like
```golang
package clusterclientset

import clientset "k8s.io/client-go/kubernetes"

type Interface interface {
    Cluster(lcluster) clientset.Interface
}
```
automatically generated which support to scope the given client down on-demand, the resulting client should talk to `/clusters/<lcluster>` instead of  ´/`.

There is a sketch at https://github.com/ncdc/kcp-client-wrappers/blob/main/kubernetes/cluster_client.go for a (manually created; we want it code generated) wrapper. Moreover, in https://github.com/kcp-dev/kubernetes/blob/feature-logical-clusters-1.23/staging/src/k8s.io/client-go/kubernetes/clientset.go#L79 we have modified upstream clientsets, but this approach is not viable as we have to fork k8s.io/code-generator and every client repo like client-go for that.

2. For informers we want cross-workspace-capable informer factories and informer types that can be scoped down (via some index to be performant) to an lcluster. The scoped down informer/list should only return objects for that lcluster, and behave like a normal non-kcp-aware informer/lister. See https://github.com/kcp-dev/kcp/blob/main/pkg/virtual/framework/wrappers/rbac/cluster_filtering.go for an example of a manually created wrapper. We need this code-generated.

3. We'll also need a ClusterDiscoveryClient that looks like this:

```go
type ClusterDiscoveryClient struct {
// ...
}

func (c *ClusterDiscoveryClient) Cluster(clusterName logicalcluster.LogicalCluster) discovery.DiscoveryInterface {
}
```

ncdc:
@varshaprasad96 @fabianvf @njhale I updated the description to include the need for a ClusterDiscoveryClient

fabianvf:
Sounds good, I'll take a look a that tomorrow. Presumably that client would live in the same place that https://github.com/kcp-dev/apimachinery/pull/21 ends up?

fabianvf:
also linking https://github.com/kcp-dev/code-generator/pull/1 (clientset generator) and https://github.com/kcp-dev/code-generator/pull/2 (listers/informer example)

PR #818, opened 2022-04-01
Enabling Validating and Mutating Webhook's for KCP

shawn-hurley:
## Summary
Adding a conformance test to prove https://github.com/kcp-dev/kubernetes/pull/52 is working as expected

Another follow on PR is going to need to test this using the APIBindings type to pass webhooks along.
## Related issue(s)

Fixes #


test/e2e/conformance/webhook_test.go
```diff
@@ -0,0 +1,224 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/kubernetes/pkg/genericcontrolplane/clientutils"
+)
+
+func TestWebhookInLogicalWorkspace(t *testing.T) {
```
sttts, line 30: LocicalCluster or Workspace. Logical Workspace does not exist as a term :)

test/e2e/conformance/webhook_test.go
```diff
@@ -0,0 +1,224 @@
...
+		t.Logf("Bootstrapping ClusterWorkspace CRDs in logical cluster %s", logicalCluster)
+		apiExtensionsClients, err := apiextensionsclient.NewClusterForConfig(cfg)
+		require.NoError(t, err, "failed to construct apiextensions client for server")
+		crdClient := apiExtensionsClients.Cluster(logicalCluster).ApiextensionsV1().CustomResourceDefinitions()
+		wildwest.Create(t, crdClient, metav1.GroupResource{Group: "wildwest.dev", Resource: "cowboys"})
+
+		// install the webhook in the first logical cluster base level.
+		url := "https://localhost:8090/hello"
+		sideEffect := admissionregistrationv1.SideEffectClassNone
+		if i == 0 {
```
sttts, line 80: can't you move it behind the loop?

test/e2e/conformance/webhook_test.go
```diff
@@ -0,0 +1,224 @@
...
+	_, err = cowbyClients.Cluster(logicalClusters[1]).WildwestV1alpha1().Cowboys("default").Create(ctx, &cowboy, metav1.CreateOptions{})
+	require.NoError(t, err, "failed to create cowboy resource in second logical cluster")
+	if testWebhook.Calls != 0 {
+		t.Fail()
+	}
+	_, err = cowbyClients.Cluster(logicalClusters[0]).WildwestV1alpha1().Cowboys("default").Create(ctx, &cowboy, metav1.CreateOptions{})
+	require.NoError(t, err, "failed to create cowboy resource in second logical cluster")
+	if testWebhook.Calls != 1 {
+		t.Fatalf("Calls is %v but expected %v", testWebhook.Calls, 1)
+	}
```
sttts, line 124: test in general looks fine. You will notice that you run into some race:
1. the CRD creation and the server being ready to serve are asynchronous, you your Create here will possibly fail for O(tens or hundreds of milliseconds). `require.Eventually` will help.
2. the admission webhook is informer driven inside of kube. That's also asynchronously update. I.e. it could be that these Create work, but your webhook is not yet called. Again, some `require.Eventually` will help.

test/e2e/conformance/webhook_test.go
```diff
@@ -0,0 +1,224 @@
...
+
+}
+
+type testWebhookServer struct {
+	Response  v1.AdmissionResponse
+	ObjectGVK schema.GroupVersionKind
+	t         *testing.T
+	Calls     int
```
sttts, line 132: add locking

test/e2e/conformance/webhook_test.go
```diff
@@ -0,0 +1,224 @@
...
+	// verify the content type is accurate
+	contentType := req.Header.Get("Content-Type")
+	if contentType != "application/json" {
+		msg := fmt.Sprintf("contentType=%s, expect application/json", contentType)
+		t.t.Logf("%v", msg)
+		http.Error(resp, msg, http.StatusBadRequest)
+		return
+	}
+
+	scheme := runtime.NewScheme()
```
sttts, line 172: usually a scheme is constructed outside in some init func.

test/e2e/conformance/webhook_test.go
```diff
@@ -0,0 +1,224 @@
...
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	t.Cleanup(cancelFunc)
+
+	cfg := server.DefaultConfig(t)
+	cfg.CertFile = filepath.Join(dirPath, "apiserver.crt")
+	cfg.KeyFile = filepath.Join(dirPath, "apiserver.key")
+
+	// Until we get rid of the multiClusterClientConfigRoundTripper and replace it with scoping,
+	// make sure we don't break cross-logical cluster client listing.
+	clientutils.EnableMultiCluster(cfg, nil, true)
```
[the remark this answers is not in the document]
shawn-hurley: Copied over don't have a use case

test/e2e/conformance/webhook_test.go
shawn-hurley: Thank you for this knowledge! 

If I understand correctly I need to eventually require that the create call will work. 

I would like to understand point 2 more fully for my knowledge. Does this mean that I could expect that a controller gets an object that did not get processed by the validating webhook?

Thanks.

test/e2e/conformance/webhook_test.go
```diff
@@ -0,0 +1,230 @@
...
+	"k8s.io/client-go/rest"
+)
+
+var scheme = runtime.NewScheme()
+
+func init() {
+	admissionregistrationv1.AddToScheme(scheme)
+	v1.AddToScheme(scheme)
+	v1alpha1.AddToScheme(scheme)
+}
```
ncdc, line 38: Mind moving the creation of the scheme into TestWebhookInWorkspace, require.NoError each AddToScheme call, and pass the scheme as a field into testWebhookServer?

test/e2e/conformance/webhook_test.go
```diff
@@ -0,0 +1,230 @@
...
+		Spec: v1alpha1.CowboySpec{},
+	}
+	require.Eventually(t, func() bool {
+		_, err = cowbyClients.Cluster(logicalClusters[1]).WildwestV1alpha1().Cowboys("default").Create(ctx, &cowboy, metav1.CreateOptions{})
+		if err != nil && !errors.IsAlreadyExists(err) {
+			return false
+		}
+		return true
+
+	}, 10*time.Second, 2*time.Second)
```
ncdc, line 129: We typically use `wait.ForeverTestTimeout` for the timeout, and a shorter interval (e.g. 100ms)

test/e2e/conformance/webhook_test.go
```diff
@@ -0,0 +1,230 @@
...
+	require.NoError(t, err, "failed to construct cowboy client for server")
+
+	//Insall the Cowboy resources into logical clusters
+	for _, logicalCluster := range logicalClusters {
+		t.Logf("Bootstrapping ClusterWorkspace CRDs in logical cluster %s", logicalCluster)
+		apiExtensionsClients, err := apiextensionsclient.NewClusterForConfig(cfg)
+		require.NoError(t, err, "failed to construct apiextensions client for server")
+		crdClient := apiExtensionsClients.Cluster(logicalCluster).ApiextensionsV1().CustomResourceDefinitions()
+		wildwest.Create(t, crdClient, metav1.GroupResource{Group: "wildwest.dev", Resource: "cowboys"})
+	}
```
ncdc, line 86: Please add whitespace after to separate chunks of code for easier readability

test/e2e/conformance/webhook_test.go
ncdc: Also, whitespace

test/e2e/conformance/webhook_test.go
```diff
@@ -0,0 +1,230 @@
...
+	contentType := req.Header.Get("Content-Type")
+	if contentType != "application/json" {
+		msg := fmt.Sprintf("contentType=%s, expect application/json", contentType)
+		t.T.Logf("%v", msg)
+		http.Error(resp, msg, http.StatusBadRequest)
+		return
+	}
+
+	var codecs = serializer.NewCodecFactory(scheme)
+	deserializer := codecs.UniversalDeserializer()
```
ncdc, line 181: You should be able to do this once and store as fields on the struct

test/e2e/conformance/webhook_test.go
```diff
@@ -0,0 +1,230 @@
...
+	if !ok {
+		//return an error
+		msg := fmt.Sprintf("Expected AdmissionReview but got: %T", obj)
+		t.T.Logf("%v", msg)
+		http.Error(resp, msg, http.StatusBadRequest)
+		return
+	}
+	obj, objGVK, err := deserializer.Decode(requestedAdmissionReview.Request.Object.Raw, nil, nil)
+	if err != nil {
+		t.T.Errorf("%v", err)
```
ncdc, line 203: `http.Error` and `return`?

test/e2e/conformance/webhook_test.go
```diff
@@ -0,0 +1,230 @@
...
+		http.Error(resp, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	t.Lock.Lock()
+	t.Calls = t.Calls + 1
+	t.Lock.Unlock()
+	resp.Header().Set("Content-Type", "application/json")
+	if _, err := resp.Write(respBytes); err != nil {
+		t.T.Logf("%v", err)
+	}
```
ncdc, line 229: Whitespace

test/e2e/conformance/webhook_test.go
```diff
@@ -0,0 +1,230 @@
...
+	_, err = kubeClusterClient.Cluster(logicalClusters[0]).AdmissionregistrationV1().ValidatingWebhookConfigurations().Create(ctx, webhook, metav1.CreateOptions{})
+	require.NoError(t, err, "failed to add validating webhook configurations")
+	cowboy := v1alpha1.Cowboy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "testing",
+			Namespace: "default",
+		},
+		Spec: v1alpha1.CowboySpec{},
+	}
+	require.Eventually(t, func() bool {
```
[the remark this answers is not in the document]
shawn-hurley: @sttts stated that there is a race condition here where the CRD may not be ready in the logical cluster and served yet so we want to wait for it to be available.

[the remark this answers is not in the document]
ncdc: Thanks. Please add a comment. Or you could have an eventually loop that checks discovery for the availability of the API.

pkg/admission/validatingwebhook/plugin.go
```diff
@@ -0,0 +1,182 @@
...
+	if workspace, isAPIBinding := p.getAPIBindingWorkspace(attr, lcluster); isAPIBinding {
+		wh := p.restrictToLogicalCluster(hooks, workspace)
+		whAccessor = append(whAccessor, wh...)
+	}
+
+	return p.dipatcher.Dispatch(ctx, attr, o, whAccessor)
+}
+
+func (p *Plugin) getAPIBindingWorkspace(attr admission.Attributes, lc logicalcluster.LogicalCluster) (logicalcluster.LogicalCluster, bool) {
+	l, err := p.apiBinding.List(labels.Everything())
```
sttts, line 144: We need something more clever here, something to lookup the information without iterating over 1000 apibindings. An informer index to filter by workspace would help a lot.

ncdc: Can we TODO pending #759 ?

pkg/admission/validatingwebhook/plugin.go
```diff
@@ -0,0 +1,182 @@
...
+		for _, br := range apiBinding.Status.BoundResources {
+			if br.Group == attr.GetResource().Group && br.Resource == attr.GetResource().Resource {
+				return logicalcluster.New(apiBinding.Status.BoundAPIExport.Workspace.WorkspaceName), true
+			}
+		}
+	}
+	return logicalcluster.New(""), false
+}
+
+// In the future use a restricted list call
```
sttts, line 161: same here: add a workspace index to the webhook informer and you can get the list of webhooks in constant time.

pkg/admission/validatingwebhook/plugin.go
```diff
@@ -0,0 +1,182 @@
...
+		return admission.NewForbidden(attr, fmt.Errorf("not yet ready to handle request"))
+	}
+
+	hooks := p.hookSource.Webhooks()
+	whAccessor := p.restrictToLogicalCluster(hooks, lcluster)
+
+	// Determine the type of request, is it api binding or not.
+	if workspace, isAPIBinding := p.getAPIBindingWorkspace(attr, lcluster); isAPIBinding {
+		wh := p.restrictToLogicalCluster(hooks, workspace)
+		whAccessor = append(whAccessor, wh...)
```
sttts, line 137: I think we want to replace the hooks with `wh`, not just add them, at least that's true for status updates (= status subresource). cc @ncdc

shawn-hurley: a cluster-admin couldn't have a webhook for a bound resource in their cluster then? at least that was the use case that I was thinking of

test/e2e/apibinding/apibinding_webhook_test.go
```diff
@@ -0,0 +1,334 @@
...
+	cowboy := v1alpha1.Cowboy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "testing",
+		},
+		Spec: v1alpha1.CowboySpec{},
+	}
+
+	// Avoid race condition here by making sure that CRD is served after installing the types into logical clusters
+	t.Logf("Creating cowboy resource in target logical cluster")
+	require.Eventually(t, func() bool {
```
sttts, line 176: I think this Eventually must span the testWebhook.Calls check below. For that you can use generated name to get unique objects. And replace `== 1` with `>= 1`.

sttts: same for the other tests of course.

test/e2e/apibinding/apibinding_webhook_test.go
```diff
@@ -0,0 +1,334 @@
...
+			return false
+		}
+		return true
+
+	}, wait.ForeverTestTimeout, 100*time.Millisecond)
+
+	//Avoid race condition where webhook informer is not updated before the call to create was made.
+	t.Log("Verify webhook is eventually called")
+	require.Eventually(t, func() bool {
+		return testWebhook.Calls == 1
```
sttts, line 188: would also like to see a check that the in-workspace webhook (to be installed) is not called.

pkg/admission/webhook/generic_webhook.go
```diff
@@ -22,29 +22,35 @@ import (
...
+
 type WebhookDispatcher struct {
-	dispatcher       generic.Dispatcher
-	hookSource       generic.Source
-	apiBinding       apisv1alpha1.APIBindingLister
-	apiBindHasSynced func() bool
+	dispatcher           generic.Dispatcher
+	hookSource           generic.Source
+	apiBindingsIndexer   cache.Indexer
+	apiBindingsHasSynced func() bool
```
shawn-hurley, line 48: for my knowledge why the change here in the ordering of the fields?

pkg/admission/webhook/generic_webhook.go
```diff
@@ -80,47 +88,56 @@ func (p *WebhookDispatcher) Dispatch(ctx context.Context, attr admission.Attribu
...
 			}
 		}
 	}
-	return logicalcluster.New(""), false
+	return logicalcluster.New(""), false, nil
 }
 
 // In the future use a restricted list call
 func (p *WebhookDispatcher) restrictToLogicalCluster(hooks []webhook.WebhookAccessor, lc logicalcluster.LogicalCluster) []webhook.WebhookAccessor {
+	// TODO(sttts): this might not scale if there are many webhooks. This is called per request, and traverses all
```
shawn-hurley, line 120: In the future, could we add an indexer for webhooks based on logical cluster like we have for api bindings?

pkg/admission/plugins.go
```diff
@@ -78,13 +82,15 @@ func RegisterAllKcpAdmissionPlugins(plugins *admission.Plugins) {
...
 
 var defaultOnPluginsInKcp = sets.NewString(
-	lifecycle.PluginName,              // NamespaceLifecycle
-	limitranger.PluginName,            // LimitRanger
-	mutatingwebhook.PluginName,        // MutatingAdmissionWebhook
-	validatingwebhook.PluginName,      // ValidatingAdmissionWebhook
+	lifecycle.PluginName,   // NamespaceLifecycle
+	limitranger.PluginName, // LimitRanger
+	// mutatingwebhook.PluginName,        // MutatingAdmissionWebhook
+	// validatingwebhook.PluginName,      // ValidatingAdmissionWebhook
```
shawn-hurley, line 93: Should I remove these completly?

pkg/admission/webhook/generic_webhook.go
sttts: I only switched to the indexer instead of the lister, to avoid listing all (many!) APIBindings on a shard.

pkg/admission/webhook/generic_webhook.go
```diff
@@ -80,47 +88,56 @@ func (p *WebhookDispatcher) Dispatch(ctx context.Context, attr admission.Attribu
 	return p.dispatcher.Dispatch(ctx, attr, o, whAccessor)
 }
 
-func (p *WebhookDispatcher) getAPIBindingWorkspace(attr admission.Attributes, lc logicalcluster.LogicalCluster) (logicalcluster.LogicalCluster, bool) {
-	l, err := p.apiBinding.List(labels.Everything())
+func (p *WebhookDispatcher) getAPIBindingWorkspace(attr admission.Attributes, clusterName logicalcluster.LogicalCluster) (logicalcluster.LogicalCluster, bool, error) {
+	parentClusterName, hasParent := clusterName.Parent()
+	if !hasParent {
+		// APIBindings in root are not possible (they can only point to sibling workspaces).
+		return logicalcluster.New(""), false, nil
```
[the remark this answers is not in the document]
sttts: CRDs in the root have to work.

pkg/admission/webhook/generic_webhook.go
sttts: the `Source` interface of kube does not allow that without patching. I think it is not really worth it because there are never many webhooks. We can of course add our own source implementation and use another interface. That might give us the option to improve this. But would wait until we really hit this bottleneck. I don't expect it.

pkg/admission/plugins.go
sttts: yes

test/e2e/apibinding/apibinding_webhook_test.go
```diff
@@ -0,0 +1,458 @@
...
+	}
+
+	// Avoid race condition here by making sure that CRD is served after installing the types into logical clusters
+	t.Logf("Creating cowboy resource in target logical cluster")
+	require.Eventually(t, func() bool {
+		_, err = cowbyClients.Cluster(targetWorkspace).WildwestV1alpha1().Cowboys("default").Create(ctx, &cowboy, metav1.CreateOptions{})
+		if err != nil && !errors.IsAlreadyExists(err) {
+			return false
+		}
+		return testWebhook.Calls == 0
```
sttts, line 455: this won't work. You cannot really know that you have waited long enough. I would add the server to the old tests, set up the webhook before the one in the source workspace. Then you can just check at the very end that the in-workspace webhook was not called. Because it had been set up before the source-workspace one, you can be sure you waited enough.

go.mod
```diff
@@ -37,30 +37,30 @@ require (
...
-	k8s.io/kube-scheduler => github.com/kcp-dev/kubernetes/staging/src/k8s.io/kube-scheduler v0.0.0-20220412170948-9b23827cdd1b
-	k8s.io/kubectl => github.com/kcp-dev/kubernetes/staging/src/k8s.io/kubectl v0.0.0-20220412170948-9b23827cdd1b
-	k8s.io/kubelet => github.com/kcp-dev/kubernetes/staging/src/k8s.io/kubelet v0.0.0-20220412170948-9b23827cdd1b
-	k8s.io/kubernetes => github.com/kcp-dev/kubernetes v0.0.0-20220412170948-9b23827cdd1b
-	k8s.io/legacy-cloud-providers => github.com/kcp-dev/kubernetes/staging/src/k8s.io/legacy-cloud-providers v0.0.0-20220412170948-9b23827cdd1b
-	k8s.io/metrics => github.com/kcp-dev/kubernetes/staging/src/k8s.io/metrics v0.0.0-20220412170948-9b23827cdd1b
-	k8s.io/mount-utils => github.com/kcp-dev/kubernetes/staging/src/k8s.io/mount-utils v0.0.0-20220412170948-9b23827cdd1b
-	k8s.io/pod-security-admission => github.com/kcp-dev/kubernetes/staging/src/k8s.io/pod-security-admission v0.0.0-20220412170948-9b23827cdd1b
-	k8s.io/sample-apiserver => github.com/kcp-dev/kubernetes/staging/src/k8s.io/sample-apiserver v0.0.0-20220412170948-9b23827cdd1b
+	k8s.io/api => github.com/sttts/kubernetes/staging/src/k8s.io/api v0.0.0-20220427164015-329d68c4e92d
```
sttts, line 40: you can bump to the official branch now that the PR has merged.

test/e2e/conformance/webhook_test.go
```diff
@@ -0,0 +1,292 @@
...
+		_, err = cowbyClients.Cluster(logicalClusters[0]).WildwestV1alpha1().Cowboys("default").Create(ctx, &cowboy, metav1.CreateOptions{})
+		if err != nil && !errors.IsAlreadyExists(err) {
+			return false
+		}
+		return true
+
+	}, wait.ForeverTestTimeout, 100*time.Millisecond)
+
+	//Avoid race condition where webhook informer is not updated before the call to create was made.
+	t.Log("Verify webhook is eventually called")
```
sttts, line 164: here we also have the race about the two eventually calls.

sttts:
For this to land, also for @ncdc to pick this up in review later today:
1. rebase to non-private kube branch
2. merge https://github.com/shawn-hurley/kcp/pull/1
3. sort out https://github.com/kcp-dev/kcp/pull/818/files#r861929387
4. bonus: either get rid of the two workspaces in the conformance tests, or add something testing that the webhook is only called for one of them (where the webhook is registered in).

test/e2e/apibinding/apibinding_webhook_test.go
sttts: Resolved in https://github.com/shawn-hurley/kcp/pull/1

shawn-hurley: Done

go.mod
shawn-hurley: Done

pkg/admission/plugins.go
shawn-hurley: Done

shawn-hurley:
@sttts 

I believe that the first three are all taken care of!

for the bonus, I believe that I am doing this. [Here]( https://github.com/kcp-dev/kcp/pull/818/files#diff-b06c2b9be1be99fae953b78f8da36ff0ae75445fb20dd50f27c94f1219a049e1R144) I am using logical cluster one to create a resource and [here](https://github.com/kcp-dev/kcp/pull/818/files#diff-b06c2b9be1be99fae953b78f8da36ff0ae75445fb20dd50f27c94f1219a049e1R151) We check that the webhook is not called.

test/e2e/conformance/webhook_test.go
```diff
@@ -0,0 +1,292 @@
...
+	t.Logf("Creating cowboy resource in second logical cluster")
+	require.Eventually(t, func() bool {
+		_, err = cowbyClients.Cluster(logicalClusters[1]).WildwestV1alpha1().Cowboys("default").Create(ctx, &cowboy, metav1.CreateOptions{})
+		if err != nil && !errors.IsAlreadyExists(err) {
+			return false
+		}
+		return true
+
+	}, wait.ForeverTestTimeout, 100*time.Millisecond)
+	require.Equal(t, 0, testWebhook.Calls(), "expected that the webhook is not called for logical cluster where webhook is not installed")
```
sttts, line 151: this is not helpful. You don't know that the webhook informer in the apiserver is actually up-to-date and would call the webhook if there were a bug.

sttts: What you can do:
1. create webhook
2. ```
    eventually {
       create in first
       return calls == 1
    }
    create in second
    require calls == 1
   ```

shawn-hurley: Done

sttts (approved):

PR #819, opened 2022-04-01
manifest: Add explicit port number to external hostname

kylape:
If port is not present in external hostname, the --secure-port value is
used instead.


ncdc:
Would it make more sense to change it so if external-hostname is set, we **don't** append secure-port?

sttts (approved):

ncdc:
Ok, thanks, since that's coming from upstream, no worries.

PR #822, opened 2022-04-01
Enable CI/CD for release branches

kylape:
Start using Github Action Environments so the `CICD_KUBE_*` environment
variables can be set depending on which deployment environment (i.e.
stable/unstable) is targeted.

### DIFF
--- a/.github/workflows/kcp-image.yaml
+++ b/.github/workflows/kcp-image.yaml
@@ -7,13 +7,14 @@ on:
   push:
     branches:
       - main
+      - 'release-*'
 
 jobs:
   kcp-image:
     if: github.repository_owner == 'kcp-dev'
     name: Build and Publish KCP Image
+    environment: unstable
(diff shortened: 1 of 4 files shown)


ncdc (approved):

.github/workflows/kcp-image.yaml
```diff
@@ -7,13 +7,14 @@ on:
   push:
     branches:
       - main
+      - 'release-*'
 
 jobs:
   kcp-image:
     if: github.repository_owner == 'kcp-dev'
     name: Build and Publish KCP Image
+    environment: unstable
```
ncdc, line 16: Is this going to have different values in different branches?

kylape: yes

ncdc: Thanks

ncdc (approved):

PR #823, opened 2022-04-01
apibindings: check for conflicts, refactor controller

### LINKED ISSUE #796: apibinding e2e test fails if run more than once against a persistent server

### LINKED ISSUE #842: Follow-up APIBindings:  set the storage versions in the APIBinding's status
Follow-up of #417:

Set the storage versions in the APIBinding's status correctly.

ncdc:
## Summary
When processing an APIBinding, check for naming conflicts with any other
bound CRDs in the same workspace.

This also pulls in an update to our Kubernetes fork to allow multiple CRDs in the system:bound-crds workspace to have naming conflicts.

## Related issue(s)

Fixes #796 
Fixes #842


pkg/reconciler/apibinding/name_conflict_checker.go
```diff
@@ -0,0 +1,131 @@
...
+	apislisters "github.com/kcp-dev/kcp/pkg/client/listers/apis/v1alpha1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	crdlister "k8s.io/apiextensions-apiserver/pkg/client/listers/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/clusters"
+)
+
+type nameConflictChecker struct {
+	apiBindingLister        apislisters.APIBindingLister
```
marun, line 31: (No action required) Suggest providing function signatures (e.g. listAPIBinding, getAPIExport, etc) so that they can be trivially substituted in unit tests.

pkg/reconciler/apibinding/name_conflict_checker.go
```diff
@@ -0,0 +1,131 @@
...
+	existingNames := sets.NewString()
+	existingKinds := sets.NewString()
+
+	crds, err := ncc.getBoundCRDs(clusterName)
+	if err != nil {
+		return fmt.Errorf("error checking for naming conflicts for cluster %s: error getting CRDs: %w", clusterName, err)
+	}
+
+	for _, crd := range crds {
+		existingNames.Insert(crd.Status.AcceptedNames.Plural)
```
marun, line 100: (No action required) Hmm, so each of these 3 field values not only has to be unique across the same field in different crds but also across the other 2 fields in different crds?

pkg/reconciler/apibinding/name_conflict_checker.go
```diff
@@ -0,0 +1,131 @@
...
+		existingNames.Insert(crd.Status.AcceptedNames.Singular)
+		existingNames.Insert(crd.Status.AcceptedNames.ShortNames...)
+
+		existingKinds.Insert(crd.Status.AcceptedNames.Kind)
+		existingKinds.Insert(crd.Status.AcceptedNames.ListKind)
+	}
+
+	if existingNames.Has(crd.Spec.Names.Plural) {
+		return fmt.Errorf("plural %q is already in use", crd.Spec.Names.Plural)
```
marun, line 109: (No action required) Might it make sense to indicate which crd is already using the value to aid in troubleshooting? Question applies to any of the conflict messages.

test/e2e/apibinding/apibinding_test.go
```diff
@@ -63,92 +67,123 @@ func TestAPIBinding(t *testing.T) {
...
+		require.False(t, groupExists(groups, wildwest.GroupName),
+			"should not have seen %s API group in service provider workspace %q group discovery", wildwest.GroupName, serviceProvider1Workspace)
+
+		t.Logf("Make sure %q API group shows up in consumer workspace %q group discovery", wildwest.GroupName, consumerWorkspace)
+		err = wait.PollImmediateWithContext(ctx, 100*time.Millisecond, wait.ForeverTestTimeout, func(c context.Context) (done bool, err error) {
+			groups, err := kcpClients.Cluster(consumerWorkspace).Discovery().ServerGroups()
+			if err != nil {
+				return false, fmt.Errorf("error retrieving consumer workspace %q group discovery: %w", consumerWorkspace, err)
+			}
+			if groupExists(groups, wildwest.GroupName) {
```
marun, line 123: (No action required) `return groupExists(groups, wildwest.GroupName), nil`

pkg/reconciler/apibinding/apibinding_reconcile.go
```diff
@@ -290,16 +298,27 @@ func (r *workspaceAPIExportReferenceReconciler) reconcile(ctx context.Context, a
...
+			// Check for naming conflicts
+			if err := r.nameConflictChecker.checkForConflicts(crd, logicalcluster.From(apiBinding)); err != nil {
+				conditions.MarkFalse(
+					apiBinding,
+					apisv1alpha1.CRDReady,
+					apisv1alpha1.NamingConflictsReason,
+					conditionsv1alpha1.ConditionSeverityError,
+					err.Error(),
+				)
+				return reconcileStatusStop, nil // don't retry
```
marun, line 317: (No action required) I'm not sure what to make of the persistent `// don't retry` comments. I think it may be as important to not return an error to ensure the status change (`conditions.MarkFalse`) is persisted?

pkg/reconciler/apibinding/name_conflict_checker.go
ncdc: Yes - this is a reimplementation of the logic in https://github.com/kubernetes/kubernetes/blob/00bd91e558612de9b89b4c76c7a42121ac2835c6/staging/src/k8s.io/apiextensions-apiserver/pkg/controller/status/naming_controller.go#L123.

pkg/reconciler/apibinding/name_conflict_checker.go
sttts: In reality these conflicts never happen in some tricky to diagnose way. Have never seen this in 8 years of kube.

pkg/reconciler/apibinding/apibinding_reconcile.go
```diff
@@ -312,11 +331,15 @@ func (r *workspaceAPIExportReferenceReconciler) reconcile(ctx context.Context, a
 						err,
 					)
 					return reconcileStatusStop, nil // don't retry
+				case !apierrors.IsAlreadyExists(err):
+					// TODO(ncdc): Do we want to set a user-facing condition here?
+					// This case and the next have the same effect (retry). But this one could loop, whereas the default
+					// case should proceed on the next reconcile iteration.
```
sttts: this can only loop for some limited time, no? You do a get above. So only while stale the cache will return NotFound before and we land here.

sttts: In other words: retry is fine.

pkg/reconciler/apibinding/apibinding_reconcile.go
```diff
@@ -330,48 +353,25 @@ func (r *workspaceAPIExportReferenceReconciler) reconcile(ctx context.Context, a
...
+
 		boundResources = append(boundResources, apisv1alpha1.BoundAPIResource{
 			Group:    schema.Spec.Group,
 			Resource: schema.Spec.Names.Plural,
 			Schema: apisv1alpha1.BoundAPIResourceSchema{
 				Name: schema.Name,
 				UID:  string(schema.UID),
 			},
-			StorageVersions: crd.Status.StoredVersions,
+			StorageVersions: storageVersions,
```
sttts, line 287: When we switch schema a new bound CRD is created, right? Do we carry along storage versions?

ncdc: Yes, a new bound CRD would be created. I need to merge old+new storage versions.

pkg/reconciler/apibinding/apibinding_reconcile.go
ncdc: Old code/comment - deleted locally 😄

test/e2e/apibinding/apibinding_test.go
```diff
@@ -20,6 +20,8 @@ import (
 	"context"
 	"embed"
 	"fmt"
+	"github.com/kcp-dev/apimachinery/pkg/logicalcluster"
+	"github.com/kcp-dev/kcp/third_party/conditions/util/conditions"
```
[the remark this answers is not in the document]
ncdc: Helpful hint - save yourself time: ignore reviewing imports - CI will catch it & I'll always fix before it's ready for merge 😄

pkg/reconciler/apibinding/name_conflict_checker.go
ncdc: Agreed, but this is different because users never have access to the bound CRDs. I will have a fix for this so it shows which APIBinding it conflicts with.

pkg/reconciler/apibinding/name_conflict_checker.go
ncdc: Am working on this 😄

pkg/reconciler/apibinding/apibinding_reconcile.go
ncdc: Good point about an error precluding the status patching. I have adjusted the reconciler to always patch (open for discussion)

sttts:
Let me know when this is ready to review (regarding https://github.com/kcp-dev/kcp/pull/823#issuecomment-1096846758).

ncdc:
It's ready

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -69,19 +69,50 @@ func NewController(
...
+		apiBindingsLister: apiBindingInformer.Lister(),
+		listAPIBindings: func(clusterName logicalcluster.LogicalCluster) ([]*apisv1alpha1.APIBinding, error) {
+			list, err := apiBindingInformer.Lister().List(labels.Everything())
+			if err != nil {
+				return nil, err
+			}
+
+			var ret []*apisv1alpha1.APIBinding
+
+			for i := range list {
```
sttts, line 86: (no action necessary) I have started adding an index by workspace elsewhere. Is that correct / necessary / sensible? These loops here definitely won't scale long-term.

sttts, line 86: (no action necessary) I have started adding an index by workspace elsewhere. Is that correct / necessary / sensible? These loops here definitely won't scale long-term.

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -69,19 +69,50 @@ func NewController(
...
+		getAPIResourceSchema: func(clusterName logicalcluster.LogicalCluster, name string) (*apisv1alpha1.APIResourceSchema, error) {
+			return apiResourceSchemaInformer.Lister().Get(clusters.ToClusterAwareKey(clusterName, name))
+		},
 		apiResourceSchemaIndexer: apiResourceSchemaInformer.Informer().GetIndexer(),
-		crdLister:                crdInformer.Lister(),
-		crdIndexer:               crdInformer.Informer().GetIndexer(),
-		deletedCRDTracker:        newLockedStringSet(),
+
+		createCRD: func(ctx context.Context, crd *apiextensionsv1.CustomResourceDefinition) (*apiextensionsv1.CustomResourceDefinition, error) {
+			return crdClusterClient.Cluster(logicalcluster.From(crd)).ApiextensionsV1().CustomResourceDefinitions().Create(ctx, crd, metav1.CreateOptions{})
```
sttts, line 109: (no action necessary) am not a fan of implicitly passing the cluster name through `metadata.clusterName` here in code. IMO it is clear to have it explicitly everywhere.

sttts, line 109: (no action necessary) am not a fan of implicitly passing the cluster name through `metadata.clusterName` here in code. IMO it is clear to have it explicitly everywhere.

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -176,7 +212,7 @@ func (c *controller) enqueueAPIBinding(obj interface{}) {
 		return
 	}
 
-	klog.Infof("Queueing APIBinding %q", key)
+	klog.V(4).Infof("Queueing APIBinding %q", key)
```
sttts, line 215: this might be worth a Slack thread and an update to the contributor guide to declare some verbosity level for different kinds of controller logging. We run in `V(2)` now by default. So `V(4)` seems to be fine, but maybe also `V(2)` might make sense for the main resource queuing in a controller.

sttts, line 215: this might be worth a Slack thread and an update to the contributor guide to declare some verbosity level for different kinds of controller logging. We run in `V(2)` now by default. So `V(4)` seems to be fine, but maybe also `V(2)` might make sense for the main resource queuing in a controller.

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -301,9 +336,10 @@ func (c *controller) process(ctx context.Context, key string) error {
 	old := obj
 	obj = obj.DeepCopy()
 
-	if err := c.reconcile(ctx, obj); err != nil {
-		return err
-	}
+	reconcileErr := c.reconcile(ctx, obj)
+
+	// Regardless of whether reconcile returned an error or not, always try to patch status if needed. Return the
+	// reconciliation error at the end.
```
sttts, line 342: (no action necessary) is this something we have to do elsewhere in other controllers too?

sttts, line 342: (no action necessary) is this something we have to do elsewhere in other controllers too?

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -47,159 +42,43 @@ const (
...
-	}
-
-	boundSchemaUIDs := sets.NewString()
-	for _, boundResource := range apiBinding.Status.BoundResources {
-		boundSchemaUIDs.Insert(boundResource.Schema.UID)
-	}
-
-	return !exportedSchemaUIDs.Equal(boundSchemaUIDs)
-}
+func (c *controller) reconcileNew(ctx context.Context, apiBinding *apisv1alpha1.APIBinding) error {
```
sttts, line 67: (no action necessary) am curious why you moved to this new style of method on `controller`.

sttts, line 67: (no action necessary) am curious why you moved to this new style of method on `controller`.

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -237,209 +116,300 @@ func (r *workspaceAPIExportReferenceReconciler) reconcile(ctx context.Context, a
...
+			listAPIBindings:      c.listAPIBindings,
+			getAPIExport:         c.getAPIExport,
+			getAPIResourceSchema: c.getAPIResourceSchema,
+			getCRD:               c.getCRD,
+		}
+
+		if err := nameConflictChecker.checkForConflicts(crd, apiBinding); err != nil {
+			conditions.MarkFalse(
+				apiBinding,
+				apisv1alpha1.APIsBound,
```
sttts, line 196: on rebind this is a little odd that we set `APIsBound` to false, no? The old schemas are still bound (I assume and hope), just the new binding failed. So maybe a slightly different condition name like `APIBindingSuccessful` or so would make it clear?

sttts, line 196: on rebind this is a little odd that we set `APIsBound` to false, no? The old schemas are still bound (I assume and hope), just the new binding failed. So maybe a slightly different condition name like `APIBindingSuccessful` or so would make it clear?

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -237,209 +116,300 @@ func (r *workspaceAPIExportReferenceReconciler) reconcile(ctx context.Context, a
...
-			getAPIResourceSchema: c.getAPIResourceSchema,
-			getCRD:               c.getCRD,
-			createCRD:            c.createCRD,
-			updateCRD:            c.updateCRD,
-			deletedCRDTracker:    c.deletedCRDTracker,
-		},
+func (c *controller) reconcileBound(ctx context.Context, apiBinding *apisv1alpha1.APIBinding) error {
+	apiExportClusterName, err := getAPIExportClusterName(apiBinding)
+	if err != nil {
+		// Should never happen
```
sttts, line 321: can happen if the export gets deleted by the owner, can't it? Also in the future with sharding it might fail here when the APIExport holding shard goes down.

sttts, line 321: can happen if the export gets deleted by the owner, can't it? Also in the future with sharding it might fail here when the APIExport holding shard goes down.

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -237,209 +116,300 @@ func (r *workspaceAPIExportReferenceReconciler) reconcile(ctx context.Context, a
...
+			conditionsv1alpha1.ConditionSeverityError,
+			err.Error(),
+		)
+
+		return nil
 	}
 
-	var errs []error
+	if referencedAPIExportChanged(apiBinding) {
+		klog.Infof("APIBinding %s|%s needs rebinding because it now points to a different APIExport", apiBinding.ClusterName, apiBinding.Name)
```
sttts, line 334: V(4) ?

sttts, line 334: V(4) ?

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
sttts: nevermind. ClusterName.

sttts: nevermind. ClusterName.

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -237,209 +116,300 @@ func (r *workspaceAPIExportReferenceReconciler) reconcile(ctx context.Context, a
...
-}
+		// Return nil here so we don't retry. If/when there is an informer event for the correct APIExport, this
+		// APIBinding will automatically be requeued.
+		return nil
+	}
+	if err != nil {
+		conditions.MarkFalse(
+			apiBinding,
+			apisv1alpha1.APIExportValid,
+			"TODO",
```
sttts, line 361: internal error?

sttts, line 361: internal error?

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -237,209 +116,300 @@ func (r *workspaceAPIExportReferenceReconciler) reconcile(ctx context.Context, a
...
-	return c.crdLister.Get(clusters.ToClusterAwareKey(clusterName, name))
-}
+	var exportedSchemas []*apisv1alpha1.APIResourceSchema
+	for _, schemaName := range apiExport.Spec.LatestResourceSchemas {
+		apiResourceSchema, err := c.getAPIResourceSchema(apiExportClusterName, schemaName)
+		if apierrors.IsNotFound(err) {
+			conditions.MarkFalse(
+				apiBinding,
+				apisv1alpha1.APIExportValid,
+				"TODO",
```
sttts, line 379: APIExportInvalidSchemas

sttts, line 379: APIExportInvalidSchemas

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -237,209 +116,300 @@ func (r *workspaceAPIExportReferenceReconciler) reconcile(ctx context.Context, a
...
-}
+			// Return nil here so we don't retry. If/when there is an informer event for the correct APIExport, this
+			// APIBinding will automatically be requeued.
+			return nil
+		}
+		if err != nil {
+			conditions.MarkFalse(
+				apiBinding,
+				apisv1alpha1.APIExportValid,
+				"TODO",
```
sttts, line 392: internal error

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -483,32 +453,43 @@ func crdFromAPIResourceSchema(schema *apisv1alpha1.APIResourceSchema) (*apiexten
...
-type lockedStringSet struct {
-	lock sync.RWMutex
-	s    sets.String
+func getAPIExportClusterName(apiBinding *apisv1alpha1.APIBinding) (logicalcluster.LogicalCluster, error) {
+	parent, hasParent := logicalcluster.From(apiBinding).Parent()
+	if !hasParent {
+		return logicalcluster.LogicalCluster{}, fmt.Errorf("a APIBinding in %s cannot reference a workspace", logicalcluster.From(apiBinding))
+	}
+	if apiBinding.Spec.Reference.Workspace == nil {
+		return logicalcluster.LogicalCluster{}, fmt.Errorf("APIBinding does not specify an APIExport")
```
sttts, line 462: `// cannot happen due to APIBinding validation`

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -483,32 +453,43 @@ func crdFromAPIResourceSchema(schema *apisv1alpha1.APIResourceSchema) (*apiexten
...
 	}
-}
 
-func (l *lockedStringSet) Add(s string) {
-	l.lock.Lock()
-	defer l.lock.Unlock()
-	l.s.Insert(s)
-}
+	if apiBinding.Status.BoundAPIExport == nil {
+		return false
```
sttts, line 474: had expected true here and below. So this means that "changed" is actually "was set + set to something else now". Might be clear to return true here and have the distinction between first-bind and rebind at the call site.

test/e2e/authorizer/authorizer_test.go
```diff
@@ -48,6 +48,17 @@ import (
...
+		/* TODO: this test creates a clusterrole/clusterrolebinding in genericcontrolplane.LocalAdminCluster. Running
+		this test multiple times will not work in the current form because the names will collide. However, unique
+		names is not a sufficient solution, because any clusterrole/binding from a previous test run will pollute
+		the intent of the test. A possible fix would be to attempt to clean up any possible resources this test
+		intends to create in genericcontrolplane.LocalAdminCluster at the beginning of the test, or to split it
+		into parts that can run against a shared server, vs. those that truly cannot.
+		*/
+		t.Skip("Skipping until this test is updated to support multiple runs against a shared server")
+	}
+
```
sttts, line 61: addressed in https://github.com/kcp-dev/kcp/pull/921

pkg/reconciler/apis/apibinding/apibinding_controller.go
ncdc: I'd like to propose this as a pattern to implement (ideally in a base controller)

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
ncdc: Mainly because the logic seemed to vary primarily based on phase (new vs binding vs bound)

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
ncdc: Yeah, wanted to discuss this with you. Probably worth a live chat

pkg/reconciler/apis/apibinding/apibinding_controller.go
ncdc: Updated to take in clusterName in incoming commit

ncdc: Updated to take in clusterName in incoming commit

pkg/reconciler/apis/apibinding/apibinding_controller.go
ncdc: 👍. If V(2) is the default, I'm wondering if we should instead move those 2s to just klog.Info and go with the unset (0/1) default?

Do you think the kcp admin would always want to see "Queueing xyz" for a standard deployment? Or only when running in a more verbose mode for debugging?

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
ncdc: Going with 2 for now in incoming commit

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
ncdc: 👍 in incoming commit

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
ncdc: Do we want to expose this information to the user?

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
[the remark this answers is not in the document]
ncdc: Fixed

test/e2e/authorizer/authorizer_test.go
ncdc: Undone

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
ncdc: Fixed

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
ncdc: Went with internal error, simplified a bit

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
ncdc: How about just removing the 2 if blocks that check Status.BoundAPIExport?

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
sttts: not necessary I think

sttts (approved):

PR #827, opened 2022-04-04
tenancy: clean up ClusterWorkspaceShard type

sttts:
- rename `WorkspaceShard` -> `ClusterWorkspaceShard`
- remove credential logic (reference to some secret)
- remove unfinished/unused location history
- replace location info with simple URLs: baseURL for direct connection, externalURL to be copied into workspace URLs
- add admission to default those URLs
- remove virtual workspaces kubeconfig subresource whose only motivation was to publish CA certs to the plugin.
- remove unfinished workspaceindex code (we can resurrect it from git history when we are there for sharding).


davidfestal (changes requested):
What about changes in the kubectl plugin to replace the use of the kubeconfig sub-resource by the direct use of the URL on the Workspace status ?
Is it something that you already implemented on a previous PR and I missed it ?

sttts:
> What about changes in the kubectl plugin to replace the use of the kubeconfig sub-resource by the direct use of the URL on the Workspace status ?

Am not aware of anything remaining related to that in the plugin.

PR #828, opened 2022-04-04
apis: add API identity as a concept to APIExport and APIBinding

sttts:
The APIExport references an "identity secret" with a file called key. That file will have a cryptographic-strong random string, defaulted for the API if not created and set explicitly by the APIExport creator.

This identity will determine some unique etcd prefix in the shards, for safe, conflict-free wildcard requests.

The virtual workspace APIExport apiserver will use some derived (sha256) value to ask for values from the shards.


ncdc:
Incorporated into #877

PR #833, opened 2022-04-05
e2e/virtual: add test that a workspace list never returns invalid resource,

sttts:


sttts:
Part of https://github.com/kcp-dev/kcp/pull/858.

PR #849, opened 2022-04-06
Kind-e2e fixture

jmprusi:
## Summary

Allows the creation of Kind clusters for e2e tests purposes


marun:
Closing - we don't want to manage kind instances on a per-test basis and it's not possible to share fixture within a test run across packages.

PR #850, opened 2022-04-06
dockerfile: Use make to build go binaries

### LINKED ISSUE #645: GitVersion is not set in kubectl version
**Describe the bug**

Knative framework checks the version of kubernetes endpoint. With the current kcp image, the GitVersion is a placeholder and the regexp is failing.

Tekton cannot run because of that.

The check is here: https://github.com/knative/pkg/blob/80c511aa340f678c7449e1cc360d43b168a140aa/version/version.go#L52

**To Reproduce**
1. Run a knative controller

**Expected behavior**

A version :)

kylape:
This is required to properly get the version strings baked into the
binaries.

* .git dir is copied in to build image, but it doesn't end up in the
  shipped image.  We can follow up later to remove the need to copy over
  the whole .git dir.
* CGO_ENABLED=0 is required for alpine/distroless, so I've kept that
  here.
* Had to list out the binaries to put into the shipped image.  Using `*`
  resulted in an 800MB image!

Fixes #645

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,13 +31,15 @@ COPY config/ config/
 COPY pkg/ pkg/
 COPY cmd/ cmd/
 COPY third_party/ third_party/
+COPY .git/ .git/
 
-RUN mkdir bin; CGO_ENABLED=0 go build -o bin ./cmd/...
+RUN apt-get update && apt-get install -y jq && mkdir bin
+RUN CGO_ENABLED=0 make
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 # FROM gcr.io/distroless/static:nonroot
 FROM alpine:3.15
 WORKDIR /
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -31,13 +31,14 @@ COPY config/ config/
 COPY pkg/ pkg/
 COPY cmd/ cmd/
 COPY third_party/ third_party/
+COPY .git/ .git/
 
-RUN mkdir bin; CGO_ENABLED=0 go build -o bin ./cmd/...
+RUN apt-get update && apt-get install -y jq && mkdir bin && CGO_ENABLED=0 make
```
ncdc, line 36: I'd put the `make` invocation as its own `RUN` line so the `apt` & `mkdir` calls can be cached separately. WDYT?

kylape: Yes, makes sense. Updated.

ncdc (approved):

ncdc:
@kylape merge when you're happy w/it & ready

PR #855, opened 2022-04-07
server: Set external address based on completed serving config

### LINKED ISSUE #853: `kubectl kcp workspace use` with relative logical cluster argument breaks current kubeconfig
**Describe the bug**

Navigating to a workspace with the workspace CLI plugin using relative logical cluster path breaks the client config.

**To Reproduce**

```
$ kcp start ...
$ kubectl kcp workspace create test
$ kubectl kcp workspace use test
$ kubectl get ns
The connection to the server 192.168.0.24 was refused - did you specify the right host or port?
```

**Expected behavior**

Navigating to a workspace using its relative logical cluster path should work.

### LINKED ISSUE #854: Root ClusterWorkspaceShard URLs are not complete
**Describe the bug**

The default `root` ClusterWorkspaceShard resource, that's created when starting KCP, does not have the default host port reported in its URLs.

This also impacts the URLs of the Workspaces that are attached to this shard.

The root cause seems the URL is initialised before the generic API server config is completed.

**To Reproduce**

```
$ kcp start ...
$ kubectl --context root get clusterworkspaceshards.tenancy.kcp.dev -o yaml
apiVersion: v1
items:
- apiVersion: tenancy.kcp.dev/v1alpha1
  kind: ClusterWorkspaceShard
  metadata:
    clusterName: root
    name: root
  spec:
    baseURL: https://192.168.0.24
    externalURL: https://192.168.0.24
```

**Expected behavior**

The default port should be reported.

**Additional context**

This is likely the root cause of #853.

astefanutti:
## Summary

This PR makes sure the external address admission initialiser relies on a completed configuration, with the default secure serving port.

The logic is taken from the generic options completion method. While it fixes the issues below, there may be a better, more idiomatic, way to fully complete the options.

## Related issue(s)

Fixes #853 
Fixes #854

### DIFF
--- a/pkg/admission/clusterworkspaceshard/admission_test.go
+++ b/pkg/admission/clusterworkspaceshard/admission_test.go
@@ -70,10 +70,12 @@ func updateAttr(ws, old *tenancyv1alpha1.ClusterWorkspaceShard) admission.Attrib
 
 func TestAdmit(t *testing.T) {
 	tests := []struct {
-		name        string
-		a           admission.Attributes
-		expectedObj runtime.Object
-		wantErr     bool
+		name                      string
+		a                         admission.Attributes
+		emptyExternalAddress      bool
+		noExternalAddressProvider bool
+		expectedObj               runtime.Object
+		wantErr                   bool
 	}{
 		{
 			name: "does nothing on update when baseURL and externalURL are set",
@@ -305,12 +307,96 @@ func TestAdmit(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "default externalURL on create when baseURL is set",
+			a: createAttr(&tenancyv1alpha1.ClusterWorkspaceShard{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test",
+				},
+				Spec: tenancyv1alpha1.ClusterWorkspaceShardSpec{
+					BaseURL: "https://boston2.kcp.dev",
+				},
+			}),
+			noExternalAddressProvider: true,
+			expectedObj: &tenancyv1alpha1.ClusterWorkspaceShard{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test",
+				},
+				Spec: tenancyv1alpha1.ClusterWorkspaceShardSpec{
+					BaseURL:     "https://boston2.kcp.dev",
+					ExternalURL: "https://boston2.kcp.dev",
+				},
+			},
+		},
+		{
+			name: "fails on create when baseURL is not set and external address provider is nil",
+			a: createAttr(&tenancyv1alpha1.ClusterWorkspaceShard{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test",
+				},
+				Spec: tenancyv1alpha1.ClusterWorkspaceShardSpec{},
+			}),
+			noExternalAddressProvider: true,
+			wantErr:                   true,
+		},
+		{
+			name: "fails on update when baseURL is not set and external address provider is nil",
+			a: updateAttr(&tenancyv1alpha1.ClusterWorkspaceShard{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test",
+				},
+				Spec: tenancyv1alpha1.ClusterWorkspaceShardSpec{},
+			},
+				&tenancyv1alpha1.ClusterWorkspaceShard{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "test",
+					},
+					Spec: tenancyv1alpha1.ClusterWorkspaceShardSpec{},
+				}),
+			noExternalAddressProvider: true,
+			wantErr:                   true,
+		},
+		{
+			name: "fails on create when baseURL is not set and external address provider returns empty string",
+			a: createAttr(&tenancyv1alpha1.ClusterWorkspaceShard{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test",
+				},
+				Spec: tenancyv1alpha1.ClusterWorkspaceShardSpec{},
+			}),
+			emptyExternalAddress: true,
+			wantErr:              true,
+		},
+		{
+			name: "fails on update when baseURL is not set and external address provider returns empty string",
+			a: updateAttr(&tenancyv1alpha1.ClusterWorkspaceShard{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test",
+				},
+				Spec: tenancyv1alpha1.ClusterWorkspaceShardSpec{},
+			},
+				&tenancyv1alpha1.ClusterWorkspaceShard{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "test",
+					},
+					Spec: tenancyv1alpha1.ClusterWorkspaceShardSpec{},
+				}),
+			emptyExternalAddress: true,
+			wantErr:              true,
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			o := &clusterWorkspaceShard{
-				Handler:         admission.NewHandler(admission.Create, admission.Update),
-				externalAddress: "external.kcp.dev",
+				Handler:                 admission.NewHandler(admission.Create, admission.Update),
+				externalAddressProvider: func() string { return "external.kcp.dev" },
+			}
+			if tt.noExternalAddressProvider {
+				o.externalAddressProvider = nil
+			} else if tt.emptyExternalAddress {
+				o.externalAddressProvider = func() string {
+					return ""
+				}
 			}
 			ctx := request.WithCluster(context.Background(), request.Cluster{Name: logicalcluster.New("root:org")})
 			if err := o.Admit(ctx, tt.a, nil); (err != nil) != tt.wantErr {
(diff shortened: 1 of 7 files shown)


ncdc:
@astefanutti thanks for finding this! What do you think about this alternative approach (note, was a quick hack, so names may need some tidying up):

```diff
diff --git a/pkg/admission/clusterworkspaceshard/admission.go b/pkg/admission/clusterworkspaceshard/admission.go
index bcc35b38..52a5cb0a 100644
--- a/pkg/admission/clusterworkspaceshard/admission.go
+++ b/pkg/admission/clusterworkspaceshard/admission.go
@@ -51,7 +51,7 @@ func Register(plugins *admission.Plugins) {
 type clusterWorkspaceShard struct {
 	*admission.Handler
 
-	externalAddress string
+	externalAddress func() string
 }
 
 // Ensure that the required admission interfaces are implemented.
@@ -102,7 +102,7 @@ func (o *clusterWorkspaceShard) Admit(ctx context.Context, a admission.Attribute
 	}
 
 	if cws.Spec.BaseURL == "" {
-		cws.Spec.BaseURL = "https://" + o.externalAddress
+		cws.Spec.BaseURL = "https://" + o.externalAddress()
 	}
 	if cws.Spec.ExternalURL == "" {
 		cws.Spec.ExternalURL = cws.Spec.BaseURL
@@ -117,6 +117,6 @@ func (o *clusterWorkspaceShard) Admit(ctx context.Context, a admission.Attribute
 	return nil
 }
 
-func (o *clusterWorkspaceShard) SetExternalAddress(externalAddress string) {
+func (o *clusterWorkspaceShard) SetExternalAddress(externalAddress func() string) {
 	o.externalAddress = externalAddress
 }
diff --git a/pkg/admission/initializers/initializer.go b/pkg/admission/initializers/initializer.go
index 3b80a180..8dc3b26f 100644
--- a/pkg/admission/initializers/initializer.go
+++ b/pkg/admission/initializers/initializer.go
@@ -87,19 +87,19 @@ func (i *kcpClusterClientInitializer) Initialize(plugin admission.Interface) {
 // NewExternalAddressInitializer returns an admission plugin initializer that injects
 // the external address of the shard into the admission plugin.
 func NewExternalAddressInitializer(
-	externalAddress string,
+	externalAddressProvider func() string,
 ) *externalAddressInitializer {
 	return &externalAddressInitializer{
-		externalAddress: externalAddress,
+		externalAddressProvider: externalAddressProvider,
 	}
 }
 
 type externalAddressInitializer struct {
-	externalAddress string
+	externalAddressProvider func() string
 }
 
 func (i *externalAddressInitializer) Initialize(plugin admission.Interface) {
 	if wants, ok := plugin.(WantExternalAddress); ok {
-		wants.SetExternalAddress(i.externalAddress)
+		wants.SetExternalAddress(i.externalAddressProvider)
 	}
 }
diff --git a/pkg/admission/initializers/interfaces.go b/pkg/admission/initializers/interfaces.go
index ecc94128..c5f04bb8 100644
--- a/pkg/admission/initializers/interfaces.go
+++ b/pkg/admission/initializers/interfaces.go
@@ -44,5 +44,5 @@ type WantsKcpClusterClient interface {
 // WantExternalAddress interface should be implemented by admission plugins
 // that want to have a external shard address injected.
 type WantExternalAddress interface {
-	SetExternalAddress(externalAddress string)
+	SetExternalAddress(externalAddress func() string)
 }
diff --git a/pkg/server/server.go b/pkg/server/server.go
index 0f8dff21..62ab206c 100644
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -242,7 +242,7 @@ func (s *Server) Run(ctx context.Context) error {
 		kcpadmissioninitializers.NewKcpInformersInitializer(s.kcpSharedInformerFactory),
 		kcpadmissioninitializers.NewKubeClusterClientInitializer(kubeClusterClient),
 		kcpadmissioninitializers.NewKcpClusterClientInitializer(kcpClusterClient),
-		kcpadmissioninitializers.NewExternalAddressInitializer(genericConfig.ExternalAddress),
+		kcpadmissioninitializers.NewExternalAddressInitializer(func() string { return genericConfig.ExternalAddress }),
 	}
 
 	apisConfig, err := genericcontrolplane.CreateKubeAPIServerConfig(genericConfig, s.options.GenericControlPlane, s.kubeSharedInformerFactory, admissionPluginInitializers, storageFactory)
```

ncdc:
This lets the normal `Complete()` flow set the external address like it normally does (to get the port in there if needed), and makes it lazily evaluated for the clusterworkspaceshard

astefanutti:
@ncdc ah that lazy initializer is exactly what I tried in the first place, but I didn't get the `clusterWorkspaceShard` part right for some reasons :) Let me rework it now that you've showed me the complete solution 😄 Thanks a lot!

pkg/server/server.go
```diff
@@ -238,6 +240,19 @@ func (s *Server) Run(ctx context.Context) error {
...
 
+	// Update the generic config external address based on the secure serving default port,
+	// so its complete representation is provided to the external address initializer.
+	if _, _, err := net.SplitHostPort(genericConfig.ExternalAddress); err != nil {
+		if genericConfig.SecureServing == nil {
+			klog.Fatalf("cannot derive external address port without listening on a secure port.")
+		}
+		_, port, err := genericConfig.SecureServing.HostPort()
+		if err != nil {
+			klog.Fatalf("cannot derive external address from the secure port: %v", err)
```
sttts, line 251: return error, don't klog.Fatal.

pkg/server/server.go
```diff
@@ -238,6 +240,19 @@ func (s *Server) Run(ctx context.Context) error {
...
+		}
+		genericConfig.ExternalAddress = net.JoinHostPort(genericConfig.ExternalAddress, strconv.Itoa(port))
```
sttts, line 253: is this a bug in kube?

astefanutti:
@ncdc @sttts I've updated the PR. I went for using `*string`, instead of a _provider_/_supplier_ function approach, that simplifies things even more, and added a couple of unit and e2e tests.

pkg/server/server.go
astefanutti: I don't think so. It's the admission plugin that's initialised in KCP before the `Complete` method is actually called on the generic options.

pkg/server/server.go
```diff
@@ -242,7 +242,9 @@ func (s *Server) Run(ctx context.Context) error {
 		kcpadmissioninitializers.NewKcpInformersInitializer(s.kcpSharedInformerFactory),
 		kcpadmissioninitializers.NewKubeClusterClientInitializer(kubeClusterClient),
 		kcpadmissioninitializers.NewKcpClusterClientInitializer(kcpClusterClient),
-		kcpadmissioninitializers.NewExternalAddressInitializer(genericConfig.ExternalAddress),
+		// The external address pointer is passed as its value may be updated
+		// with the default secure port, when the config is completed.
+		kcpadmissioninitializers.NewExternalAddressInitializer(&genericConfig.ExternalAddress),
```
ncdc, line 247: Passing in the pointer means that any admission plugin could potentially mutate genericConfig.ExternalAddress. We presumably wouldn't code that, but it is possible. Do you think we should guard against that?

pkg/admission/clusterworkspaceshard/admission_test.go
```diff
@@ -309,8 +309,82 @@ func TestAdmit(t *testing.T) {
...
+				expected := helpers.ToUnstructuredOrDie(tt.expectedObj)
+				if !apiequality.Semantic.DeepEqual(expected, got) {
+					t.Fatalf("unexpected result (A expected, B got): %s", diff.ObjectDiff(expected, got))
+				}
+			}
+		})
+	}
+}
+
+func TestAdmitWithoutExternalAddress(t *testing.T) {
```
ncdc, line 330: Do you think you could combine this into the previous test? Maybe add a bool to the test struct `provideExternalAddress` and have that set/not set `externalAddressProvider` in `o`?

astefanutti: Good idea. I've added the plugin under test to the struct, so the test cases can still be run in parallel if we ever want it to.

pkg/admission/clusterworkspaceshard/admission_test.go
```diff
@@ -69,14 +69,28 @@ func updateAttr(ws, old *tenancyv1alpha1.ClusterWorkspaceShard) admission.Attrib
...
+		Handler: admission.NewHandler(admission.Create, admission.Update),
+	}
+	pluginExternalAddressProviderReturnsEmptyString := &clusterWorkspaceShard{
+		Handler:                 admission.NewHandler(admission.Create, admission.Update),
+		externalAddressProvider: func() string { return "" },
+	}
+
 	tests := []struct {
 		name        string
+		plugin      *clusterWorkspaceShard
```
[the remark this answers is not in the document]
astefanutti: If I understanding it correctly, wouldn't this approach prevent from running the test cases in parallel in the future? Is that OK?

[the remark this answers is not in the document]
ncdc: I don't think so. Each subtest is still constructing its own `clusterWorkspaceShard` instance. Right?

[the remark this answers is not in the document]
astefanutti: Ah you're right, I've overlooked it, sorry :( This should now be closer to your expectation.

ncdc (approved):

PR #863, opened 2022-04-07
Apply resource constraints to kcp instance

kylape:
This includes:

* kcp container
* virtual-workspaces container
* 3 etcd pods

The initial goal here is to protect the infrastructure running KCP from a memory leak in KCP.  This should give us some decent head room for now.

We've actually seen node instability (i.e. memory thresholds hit on nodes) on the OSD cluster running the KCP shared service, so this is a step to try and prevent that from happening.

### DIFF
--- a/manifest/kcp.yaml
+++ b/manifest/kcp.yaml
@@ -148,6 +148,13 @@ spec:
             path: readyz
             port: 6443
             scheme: HTTPS
+        resources:
+          limits:
+            cpu: '2'
+            memory: 2Gi
+          requests:
+            cpu: '1'
+            memory: 1Gi
         volumeMounts:
         - name: etcd-certs
           mountPath: /etc/etcd/tls/server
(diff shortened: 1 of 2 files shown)


manifest/kcp.yaml
```diff
@@ -148,6 +148,13 @@ spec:
             path: readyz
             port: 6443
             scheme: HTTPS
+        resources:
+          limits:
+            cpu: '1'
+            memory: 1Gi
+          requests:
+            cpu: 500m
+            memory: 512Mi
```
ncdc, line 157: Do you want to start this high, given that we've observed the pod at ~120mb steady state (unless I misread the data)?

kylape:
@ncdc I've updated the resource limits based on a closer look at the metrics dashboard in our cluster.

ncdc (approved):

ncdc:
Thanks, let's keep an eye on the metrics after this change is deployed to make sure we aren't hitting any issues.

PR #864, opened 2022-04-07
Stop apiimporter, syncer, heartbeat from fighting over the Ready condition

### LINKED ISSUE #865: apiimport_manager, push/pull managers, and heartbeat manager fight over workload cluster readiness
**Describe the bug**
If a workload cluster becomes not ready, various controllers update the Ready condition with different reasons and messages. This causes them to hotloop fighting over the condition.

**To Reproduce**
1. Create a kind cluster
2. Attach it to a workspace
3. Delete the kind cluster
4. Watch the logs fly

**Alternatively**
1. Create a kind cluster
2. Attach it to a workspace
3. Update the SA in the cluster so the credentials in the workloadcluster resource are invalid
4. Watch the logs fly

**Expected behavior**
Ready should be set once, and other conditions should explain the details of various checks each controller has performed.

csams:
Give each controller its own condition and make heartbeat manager responsible for setting `Ready` by accounting for all conditions via `conditions.SetSummary`

Update the namespace scheduler to depend on total readiness.

Fixes #865

### DIFF
--- a/pkg/apis/workload/v1alpha1/types.go
+++ b/pkg/apis/workload/v1alpha1/types.go
@@ -105,8 +105,14 @@ type WorkloadClusterList struct {
 
 // Conditions and ConditionReasons for the kcp WorkloadCluster object.
 const (
-	// WorkloadClusterReadyCondition means the WorkloadCluster is available.
-	WorkloadClusterReadyCondition conditionsv1alpha1.ConditionType = "Ready"
+	// SyncerReady means the syncer is ready to transfer resources between KCP and the WorkloadCluster.
+	SyncerReady conditionsv1alpha1.ConditionType = "SyncerReady"
+
+	// APIImporterReady means the APIImport component is ready to import APIs from the WorkloadCluster.
+	APIImporterReady conditionsv1alpha1.ConditionType = "APIImporterReady"
+
+	// HeartbeatHealthy means the HeartbeatManager is able to contact the WorkloadCluster at a specified internval.
+	HeartbeatHealthy conditionsv1alpha1.ConditionType = "HeartbeatHealthy"
 
 	// WorkloadClusterUnknownReason documents a WorkloadCluster which readiness is unknown.
 	WorkloadClusterUnknownReason = "WorkloadClusterStatusUnknown"
--- a/pkg/apis/workload/v1alpha1/types.go
+++ b/pkg/apis/workload/v1alpha1/types.go
@@ -111,7 +111,7 @@ const (
 	// APIImporterReady means the APIImport component is ready to import APIs from the WorkloadCluster.
 	APIImporterReady conditionsv1alpha1.ConditionType = "APIImporterReady"
 
-	// HeartbeatHealthy means the HeartbeatManager is able to contact the WorkloadCluster at a specified internval.
+	// HeartbeatHealthy means the HeartbeatManager has seen a heartbeat for the WorkloadCluster within the expected interval.
 	HeartbeatHealthy conditionsv1alpha1.ConditionType = "HeartbeatHealthy"
 
 	// WorkloadClusterUnknownReason documents a WorkloadCluster which readiness is unknown.
--- a/pkg/reconciler/workload/heartbeat/heartbeat_manager.go
+++ b/pkg/reconciler/workload/heartbeat/heartbeat_manager.go
@@ -24,7 +24,7 @@ import (
 
 	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
 	"github.com/kcp-dev/kcp/pkg/reconciler/workload/basecontroller"
-	conditionsv1alpha1 "github.com/kcp-dev/kcp/third_party/conditions/apis/conditions/v1alpha1"
+	conditionsapi "github.com/kcp-dev/kcp/third_party/conditions/apis/conditions/v1alpha1"
 	"github.com/kcp-dev/kcp/third_party/conditions/util/conditions"
 )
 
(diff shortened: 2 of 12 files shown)


marun:
I'm still unsure that this is a desirable change given that only the heartbeat controller will be setting the ready condition in the immediate future. And unless we have a concrete plan to implement one or more controllers that would be competing with the heartbeat controller for managing the ready status of workload clusters, I'd suggest YAGNI.

pkg/apis/workload/v1alpha1/types.go
```diff
@@ -105,8 +105,14 @@ type WorkloadClusterList struct {
...
 const (
-	// WorkloadClusterReadyCondition means the WorkloadCluster is available.
-	WorkloadClusterReadyCondition conditionsv1alpha1.ConditionType = "Ready"
+	// SyncerReady means the syncer is ready to transfer resources between KCP and the WorkloadCluster.
+	SyncerReady conditionsv1alpha1.ConditionType = "SyncerReady"
+
+	// APIImporterReady means the APIImport component is ready to import APIs from the WorkloadCluster.
+	APIImporterReady conditionsv1alpha1.ConditionType = "APIImporterReady"
+
+	// HeartbeatHealthy means the HeartbeatManager is able to contact the WorkloadCluster at a specified internval.
```
ncdc, line 114:
```suggestion
	// HeartbeatHealthy means the HeartbeatManager has seen a heartbeat for the WorkloadCluster within the expected interval.
```

Trying to future-proof for when push mode is gone

pkg/reconciler/workload/heartbeat/heartbeat_manager.go
```diff
@@ -36,27 +36,37 @@ type clusterManager struct {
 }
 
 func (c *clusterManager) Reconcile(ctx context.Context, cluster *workloadv1alpha1.WorkloadCluster) error {
+
```
ncdc, line 39:
```suggestion
```

pkg/reconciler/workload/heartbeat/heartbeat_manager.go
```diff
@@ -24,7 +24,7 @@ import (
 
 	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
 	"github.com/kcp-dev/kcp/pkg/reconciler/workload/basecontroller"
-	conditionsv1alpha1 "github.com/kcp-dev/kcp/third_party/conditions/apis/conditions/v1alpha1"
+	conditionsapi "github.com/kcp-dev/kcp/third_party/conditions/apis/conditions/v1alpha1"
```
ncdc, line 27: Not super important, but curious why the change here?

pkg/reconciler/workload/heartbeat/heartbeat_manager.go
```diff
@@ -36,27 +36,37 @@ type clusterManager struct {
...
+		),
+	)
+
 	latestHeartbeat := time.Time{}
 	if cluster.Status.LastSyncerHeartbeatTime != nil {
 		latestHeartbeat = cluster.Status.LastSyncerHeartbeatTime.Time
 	}
 	if latestHeartbeat.IsZero() {
-		klog.V(5).Infof("Marking WorkloadCluster %s|%s not ready due to no heartbeat", cluster.ClusterName, cluster.Name)
+		klog.V(5).Infof("Marking Syncer %s|%s not ready due to no heartbeat", cluster.ClusterName, cluster.Name)
```
ncdc, line 54: Maybe change the log message here so instead of talking about marking not ready, we say we're setting HeartbeathHealthy to false b/c no heartbeat?

pkg/reconciler/workload/heartbeat/heartbeat_manager.go
```diff
@@ -36,27 +36,37 @@ type clusterManager struct {
...
 		conditions.MarkFalse(cluster,
-			workloadv1alpha1.WorkloadClusterReadyCondition,
+			workloadv1alpha1.HeartbeatHealthy,
 			workloadv1alpha1.ErrorHeartbeatMissedReason,
-			conditionsv1alpha1.ConditionSeverityWarning,
+			conditionsapi.ConditionSeverityWarning,
 			"No heartbeat yet seen")
 	} else if time.Since(latestHeartbeat) > c.heartbeatThreshold {
-		klog.V(5).Infof("Marking WorkloadCluster %s|%s not ready due to a stale heartbeat", cluster.ClusterName, cluster.Name)
+		klog.V(5).Infof("Marking Syncer %s|%s not ready due to a stale heartbeat", cluster.ClusterName, cluster.Name)
```
ncdc, line 61: Ditto

pkg/reconciler/workload/heartbeat/heartbeat_manager.go
```diff
@@ -36,27 +36,37 @@ type clusterManager struct {
...
-			workloadv1alpha1.WorkloadClusterReadyCondition,
+			workloadv1alpha1.HeartbeatHealthy,
 			workloadv1alpha1.ErrorHeartbeatMissedReason,
-			conditionsv1alpha1.ConditionSeverityWarning,
+			conditionsapi.ConditionSeverityWarning,
 			"No heartbeat since %s", latestHeartbeat)
 	} else {
-		klog.V(5).Infof("Marking WorkloadCluster %s|%s ready", cluster.ClusterName, cluster.Name)
-		conditions.MarkTrue(cluster, workloadv1alpha1.WorkloadClusterReadyCondition)
+		klog.V(5).Infof("Marking Heartbeat healthy for %s|%s", cluster.ClusterName, cluster.Name)
```
ncdc, line 68:
```suggestion
		klog.V(5).Infof("Marking Heartbeat healthy for WorkloadCluster %s|%s", cluster.ClusterName, cluster.Name)
```

pkg/reconciler/workload/syncer/pushmanager.go
```diff
@@ -91,7 +91,6 @@ func (m *pushSyncerManager) update(ctx context.Context, cluster *workloadv1alpha
 	}
 
 	klog.Infof("Started push mode syncer from clusterName %s for pcluster %s", kcpClusterName, cluster.Name)
-	conditions.MarkTrue(cluster, workloadv1alpha1.WorkloadClusterReadyCondition)
```
ncdc, line 94: We don't need SyncerReady=True here?

csams: Yep, not sure why I removed it. `checkHealth` doesn't get called in the func that calls update if an update was needed.

pkg/reconciler/workload/heartbeat/heartbeat_manager.go
csams: The `conditions` library uses `conditionsapi` as the conventional import name. I'd been spelunking in the lib and was going for consistency with it. I now see `conditionsv1alpha1` is used in KCP, though, so I'll switch to that.

ncdc: I think we have a mix here & there. Either is fine for now.

ncdc (approved):

ncdc:
@marun @imjasonh how does this look to you?

For background/additional context: the conditions library we're using (from Cluster API) went through extensive R&D, settling on the use of SetSummary to summarize a resource's Ready condition. It's a good practice to follow, and even if we only currently have a single condition that represents Ready, for future-proofing, we should choose a unique name for that condition, and have it flow into Ready via SetSummary.

imjasonh (approved):
This looks like a good change, thanks! 👍

PR #873, opened 2022-04-08
Syncer Tunneler: Pod logs, ....

### LINKED ISSUE #333: Prototype kcp apiserver -> pcluster egress
The kcp apiserver has to talk to the pcluster for
- `kubectl exec|attach|logs`
- webhooks in CRD conversion
- webhooks in admission.

For that kubernetes uses egress, compare https://kubernetes.io/docs/tasks/extend-kubernetes/setup-konnectivity/.

The task here is to enable kcp to make use of this mechanism, configured for each shard (not customizable for the users).

### Acceptance criteria:

1. `kubectl exec/attach/logs` works
2. webhooks work

aojea:
## Summary

Add functionality to the Syncer to establish reverse connections and act as "reverse proxy for the downstream clusters.

The KCP server exposes 2 URLs that allows to proxy over reverse connections with the following format
- `https://host/services/syncer-tunnels/clusters/<ws>/apis/workload.kcp.dev/v1alpha1/synctargets/<name>/connect` : connections received in this URL are hold used to forward connections over them (the first connection in this URL establish the control channel between the syncer and kcp)
- `https://host/services/syncer-tunnels/clusters/<ws>/apis/workload.kcp.dev/v1alpha1/synctargets/<name>/proxy/{path}` : proxies the {path} over the reverse connection identified by the {cluster} and {syncer-name} to the syncer apiserver. The proxied path uses the `syncer` service account permissions


You can use all `kubectl` commands against the downstream cluster by using the "syncer-proxy" url as base:

`kubectl --server=https://{host}/services/syncer-tunnels/clusters/<ws>/apis/workload.kcp.dev/v1alpha1/synctargets/<name>/proxy get pods -A`

This kubectl inherits the `syncer` ServiceAccount permissions

**WARNING**

This first implementation exposes the whole downstream cluster without any security
The endpoint https://host/services/syncer-tunnels/clusters/<ws>/apis/workload.kcp.dev/v1alpha1/synctargets/<name>/proxy/{path}  is highly experimental and temporary, it is under the feature gate `KCPSyncerTunnel`

### Example

1. Start your kcp cluster:

`kcp start`

2. Create the syncer configuration:

`kubectl kcp workload sync SYNCER_NAME --syncer-image=SYNCER_IMAGE -o syncer.yaml`

3. Append permissions to read pods and pods/log and exec on pods on the generated manifest, per example

```yaml
kind: ClusterRole                                                                                                                                                                                                                                                        
metadata:                                                                                                                                                                                                                                                                
  name: kcp-syncer-cluster1-195k2b7i                                                                                                                                                                                                                                     
rules:                                                                                                                                                                                                                                                                   
- apiGroups:                                                                                                                                                                                                                                                             
  - ""                                                                                                                                                                                                                                                                   
  resources:                                                                                                                                                                                                                                                             
  - namespaces                                                                                                                                                                                                                                                           
  verbs:                                                                                                                                                                                                                                                                 
  - "create"                                                                                                                                                                                                                                                             
  - "list"                                                                                                                                                                                                                                                               
  - "watch"                                                                                                                                                                                                                                                              
- apiGroups:                                                                                                                                                                                                                                                             
  - ""                                                                                                                                                                                                                                                                   
  resources:                                                                                                                                                                                                                                                             
  - pods                                                                                                                                                                                                                                                                 
  - pods/log
  - pods/exec                                                                                                                                                                                                                                                             
  verbs:                                                                                                                                                                                                                                                                 
  - "get"                                                                                                                                                                                                                                                             
  - "list"     
(...snipped...)
```

4. Apply the syncer manifest to the downstream cluster

`kubect apply -f syncer.yaml`

5. After some seconds, the syncer will connect to the kcp cluster and you'll be able to use `kubectl` on the exposed URL with the reverse connections, in this example:

Get the downstream deployments:

```
kubectl --server=https://{host}/services/syncer-tunnels/clusters/<ws>/apis/workload.kcp.dev/v1alpha1/synctargets/<name>/proxy get deployments -A
NAMESPACE                      NAME                           READY   UP-TO-DATE   AVAILABLE   AGE
kcp-syncer-cluster1-195k2b7i   kcp-syncer-cluster1-195k2b7i   1/1     1            1           34m
kube-system                    coredns                        2/2     2            2           4h53m
local-path-storage             local-path-provisioner         1/1     1            1           4h53m
```

Get the downstream pods:
```
kubectl --server=https://{host}/services/syncer-tunnels/clusters/<ws>/apis/workload.kcp.dev/v1alpha1/synctargets/<name>/proxy get pods -A
NAMESPACE                      NAME                                            READY   STATUS    RESTARTS   AGE
kcp-syncer-cluster1-195k2b7i   kcp-syncer-cluster1-195k2b7i-7bbc5698dc-pjm2t   1/1     Running   0          32m
kube-system                    coredns-6d4b75cb6d-6g5jz                        1/1     Running   0          4h54m
kube-system                    coredns-6d4b75cb6d-qb22j                        1/1     Running   0          4h54m
kube-system                    etcd-kind-control-plane                         1/1     Running   0          4h54m
kube-system                    kindnet-8mtcr                                   1/1     Running   0          4h54m
kube-system                    kube-apiserver-kind-control-plane               1/1     Running   0          4h54m
kube-system                    kube-controller-manager-kind-control-plane      1/1     Running   0          4h54m
kube-system                    kube-proxy-7jxp7                                1/1     Running   0          4h54m
kube-system                    kube-scheduler-kind-control-plane               1/1     Running   0          4h54m
local-path-storage             local-path-provisioner-9cd9bd544-84rm6          1/1     Running   0          4h54m
```

Get one pod logs:

```
kubectl --server=https://{host}/services/syncer-tunnels/clusters/<ws>/apis/workload.kcp.dev/v1alpha1/synctargets/<name>/proxy -n kube-system logs coredns-6d4b75cb6d-6g5jz
.:53
[INFO] plugin/reload: Running configuration MD5 = db32ca3650231d74073ff4cf814959a7
CoreDNS-1.8.6
linux/amd64, go1.17.1, 13a9191
```

or exec into a pod:
```
kubectl --server=https://{host}/services/syncer-tunnels/clusters/<ws>/apis/workload.kcp.dev/v1alpha1/synctargets/<name>/proxy-n kube-system exec -it kube-proxy-7jxp7 sh                                                                                    
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
# 
# ls
bin   dev  go-runner  lib    media  opt   product_name  root  sbin  sys  usr
boot  etc  home       lib64  mnt    proc  product_uuid  run   srv   tmp  var

```




## Related issue(s)

Fixes #333


pkg/server/tunneler.go
```diff
@@ -0,0 +1,43 @@
+/*
+Copyright 2021 The KCP Authors.
```
sttts, line 2: 2022

pkg/revdial/e2e_test.go
```diff
@@ -0,0 +1,207 @@
...
+		l.Close()
+		server.Close()
+		publicServer.Close()
+		backend.Close()
+	}
+	return publicServer.Client(), publicServer.URL + "/proxy/d001/", stop
+
+}
+
+func Test_e2e(t *testing.T) {
```
sttts, line 74: move these into test/e2e? Or is this actually an integration test?

pkg/syncer/tunneler.go
```diff
@@ -0,0 +1,96 @@
...
+		return err
+	}
+	proxy.Transport = trDownstream
+
+	// create the reverse connection
+	dst := "https://" + upstream.Host + "/" + revdial.DefaultRootPathPrefix
+	id := fmt.Sprintf("%s_%s", kcpClusterName, pcluster)
+	l, err := revdial.NewListener(clientUpstream, dst, id)
+	if err != nil {
+		panic(err)
```
sttts, line 55: return err

pkg/syncer/syncer.go
```diff
@@ -217,6 +217,14 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 
 	go specSyncer.Start(ctx, numSyncerThreads)
 	go statusSyncer.Start(ctx, numSyncerThreads)
+	// connect to create the reverse tunnels
+	go wait.UntilWithContext(ctx, func(ctx context.Context) {
+		klog.V(5).Infof("Starting tunnel for clusterName %s from pcluster", cfg.KCPClusterName, cfg.WorkloadClusterName)
+		err := startTunneler(ctx, upstreamConfig, downstreamConfig, cfg.KCPClusterName.String(), cfg.WorkloadClusterName)
```
sttts, line 223: I think we have to actually connect to the virtual workspace apiserver rather than the one shard that the WorkloadCluster object is one.

pkg/revdial/e2e_test.go
aojea: integration,

pkg/syncer/syncer.go
aojea: can you link me to some examples or snippet on the code? how can I connect " to the virtual workspace apiserver" ?

pkg/syncer/syncer.go
```diff
@@ -210,6 +210,14 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 
 	go specSyncer.Start(ctx, numSyncerThreads)
 	go statusSyncer.Start(ctx, numSyncerThreads)
+	// connect to create the reverse tunnels
+	go wait.UntilWithContext(ctx, func(ctx context.Context) {
+		klog.V(5).Infof("Starting tunnel for clusterName %s from pcluster", cfg.KCPClusterName, cfg.SyncTargetName)
+		err := startTunneler(ctx, upstreamConfig, downstreamConfig, cfg.KCPClusterName.String(), cfg.SyncTargetName)
```
sttts, line 216: is this blocking? Why 1 min delay below?

pkg/revdial/const.go
```diff
@@ -0,0 +1,29 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package revdial
+
+// define the constants used to build the URL
+// The listener connects to an user with path [host:port/base]/revdial?id=[id]
```
sttts, line 20: to make this more concrete for our use-case:
- the syncer is 1:1 with a SyncTarget in a certain workspace
- the id is probably `root:org:ws:<sync-target-name>`.
- we have to authorize this request via some RBAC SAR against the SyncTarget, maybe with a custom subresource `synctarget/revdial`  or a custom verb.

pkg/revdial/const.go
```diff
@@ -0,0 +1,29 @@
...
+// The dialer listens on the urls:
+// [host:port/base]/revdial for the reverse connections
+// [host:port/base]/proxy/[id]/[path] for the reverse proxied to [path]
```
sttts, line 23: this is too static I think. We need a more flexible way to map these path to IDs, i.e. dynamic in code, not with fixed paths in some mux.

E.g. a request comes in for `/clusters/root:org:ws/api/v1/namespaces/default/pods/foo/logs`. We have to lookup the pod and check where it is scheduled to via `state.workload.kcp.dev/<synctarget-name>` label (cc @jmprusi). `<synctarget-name>` (this will be either a UID or a hash, see https://docs.google.com/document/d/1RyyIuOq20PX6kAGVuzCvXdqpipsnci5E1c887Y5GKbg/edit#), but should equal the revdial id. Hence, when we have that id from the pod's label, we pick the right reverse proxy.

At this point, the request is already authorized by kcp. But it would be good to have some more authorization (= filtering of paths) on the syncer side. Note that the syncer today is pretty privileged.

sttts: Note: we don't have pods yet in the system. We are looking into inverse syncing which will make pods replicated from downstream back to kcp. Hence, for the time being we could pick `state.workload.kcp.dev/<synctarget-name>` and with that the id from the namespace instead of the pod.

pkg/syncer/syncer.go
aojea: it's blocking, the 1 mins is to wait for reconnecting

pkg/server/config.go
```diff
@@ -315,6 +316,7 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
 		*c.preHandlerChainMux = append(*c.preHandlerChainMux, mux)
 		apiHandler = mux
 
+		apiHandler = revdial.WithSyncerTunnel(apiHandler)
```
aojea, line 319: @s-urbaniak how can I add authentication to this handler?

ezrasilvera:
Is there a way to disable this feature? (through config or something) ?  I can think of cases that don't require this functionality and want to avoid the additional security exposure.

pkg/server/config.go
s-urbaniak: @aojea is authentication or is it authorization you want to add here? how and why should it be different from the default authz/authn we already configure in kcp?

Generally. your handler here already gets authz as described in https://github.com/kcp-dev/kcp/blob/main/docs/authorization.md.

When it comes to authentication it also gets the stock authentication scheme from within kcp.

s-urbaniak: oh i see, this handler is before authorization is initialized, and as discussed OOB you cannot put it later in the chain. i suggest we discuss together with @sttts what we envision authz wise.

aojea: thanks sur :+1:

pkg/syncer/syncer.go
```diff
@@ -218,6 +218,14 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 
 	go specSyncer.Start(ctx, numSyncerThreads)
 	go statusSyncer.Start(ctx, numSyncerThreads)
+	// connect to create the reverse tunnels
+	go wait.UntilWithContext(ctx, func(ctx context.Context) {
+		klog.V(5).Infof("Starting tunnel for clusterName %s from pcluster", cfg.SyncTargetWorkspace, cfg.SyncTargetName)
```
sttts, line 223:
```suggestion
		klog.V(5).Infof("Starting tunnel for clusterName %s:%s from pcluster", cfg.SyncTargetWorkspace, cfg.SyncTargetName)
```

sttts:
The endpoint `https://{host}/clusters/{cluster}/services/tunnels/syncer-proxy/{syncer-name}/{proxied-path}` is highly experimental and temporary. Can you clarify that in the description and code?

Am fine with moving the wiring of a real `/clusters/ws/api/namespaces/ns/pods/name/logs` to the proxy into a follow-up.

pkg/syncer/tunneler.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	proxy.Transport = clientDownstream.Transport
+
+	u, err := url.Parse(upstream.Host)
+	if err != nil {
+		return err
+	}
+	// create the reverse connection
+	// virtual workspaces
+	// host/clusters/root:andy/syncer-connect/(syncer-name)/syncer-connect
```
sttts, line 67: outdated comment

pkg/syncer/tunneler.go
```diff
@@ -0,0 +1,93 @@
...
+	dst := fmt.Sprintf("%s://%s/clusters/%s", u.Scheme, u.Host, syncTargetWorkspace)
+	klog.Infof("syncer: connecting to %s with key %s", dst, syncTargetName)
```
sttts, line 69: please always print the workspace too, i.e. `%s|%s`

sttts: nit: the "syncer: " prefix is not needed. This goes into syncer logs anyway.

pkg/syncer/syncer.go
```diff
@@ -218,6 +218,14 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
...
+		err := startTunneler(ctx, upstreamConfig, downstreamConfig, cfg.SyncTargetWorkspace, cfg.SyncTargetName)
+		if err != nil {
+			klog.Errorf("Failed to create tunnel for clusterName %s from pcluster %s: %v", cfg.SyncTargetWorkspace, cfg.SyncTargetName, err)
```
sttts, line 226:
```suggestion
			klog.Errorf("Failed to create tunnel for SyncTarget %s|%s: %v", cfg.SyncTargetWorkspace, cfg.SyncTargetName, err)
```

pkg/syncer/syncer.go
```diff
@@ -218,6 +218,14 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
...
+		}
+	}, 60*time.Second)
```
sttts, line 228: a sensible variant with backoff would be nice.

pkg/tunneler/listener.go
```diff
@@ -0,0 +1,290 @@
...
+func syncerTunnelURL(host string, id string) (string, error) {
+	if id == "" {
+		return "", fmt.Errorf("id can not be empty")
+	}
+	hostURL, err := url.Parse(host)
+	if err != nil || hostURL.Scheme != "https" || hostURL.Host == "" {
+		return "", fmt.Errorf("wrong url format, expected https://host<:port>/<path>: %w", err)
+	}
+	host = strings.Trim(host, "/")
+	return host + defaultTunnelPathPrefix + "/" + cmdTunnelConnect + "/" + id, nil
```
sttts, line 284: the URL should be `https://host/services/tunnels/clusters/ws/...`, not services behind clusters.

sttts: while at it, would even rename it to `the URL should be `https://host/services/syncer-tunnels/clusters/ws/...`

sttts: And to make it complete: `https://host/services/syncer-tunnels/clusters/<ws>/apis/workload.kcp.dev/v1alpha1/synctargets/<name>/tunnel`

aojea: if I do that , don't I miss the cluster name from the context added in `WithClusterScope`?

```
func WithClusterScope(apiHandler http.Handler) http.HandlerFunc {
	return func(w http.ResponseWriter, req *http.Request) {
		var clusterName logicalcluster.Name
		if path := req.URL.Path; strings.HasPrefix(path, "/clusters/") {
			path = strings.TrimPrefix(path, "/clusters/")

			i := strings.Index(path, "/")
```

sttts: yes, under `/services` you have to extract the cluster manually yourself. That's expected.

sttts: Note: with a follow-up this will only be a problem for the syncer tunnel side, not the kubectl/client side. The later will move to  `/clusters/<ws>/api/v1/namespace/<ns>/pods/<name>/log`.

pkg/syncer/syncer.go
```diff
@@ -218,6 +219,26 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
...
+		initBackoff   = 5 * time.Second
+		maxBackoff    = 5 * time.Minute
+		resetDuration = 1 * time.Minute
+		backoffFactor = 2.0
+		jitter        = 1.0
+		clock         = &clock.RealClock{}
+		sliding       = true
+	)
+
+	backoffMgr := wait.NewExponentialBackoffManager(initBackoff, maxBackoff, resetDuration, backoffFactor, jitter, clock)
```
sttts, line 233: there is a default backoff. Isn't good enough?

sttts: and if not we should rather copy that down into some func. Now the StartSyncer func gets crowded too much by these details.

aojea: the retry.DefaultBackoff? that is too quick for these things

test/e2e/syncer/syncer_test.go
```diff
@@ -251,6 +255,76 @@ func TestSyncerLifecycle(t *testing.T) {
...
+		podsAllRole := &rbacv1.Role{
+			ObjectMeta: metav1.ObjectMeta{Name: "pods-all"},
+			Rules: []rbacv1.PolicyRule{
+				rbachelper.NewRule("*").Groups("").Resources("pods").RuleOrDie(),
+				rbachelper.NewRule("*").Groups("").Resources("pods/log").RuleOrDie(),
+				rbachelper.NewRule("*").Groups("").Resources("pods/exec").RuleOrDie(),
+			},
+		}
+
+		_, err = upstreamKubeClusterClient.RbacV1().Roles(upstreamNamespace.Name).Create(logicalcluster.WithCluster(ctx, wsClusterName), podsAllRole, metav1.CreateOptions{})
```
sttts, line 269: this confuses me. Why do we need upstream rbac?

sttts: Would have expected them downstream for the syncer service account only.

test/e2e/syncer/syncer_test.go
```diff
@@ -251,6 +255,76 @@ func TestSyncerLifecycle(t *testing.T) {
...
+		proxiedConfig := rest.CopyConfig(syncerFixture.SyncerConfig.UpstreamConfig)
+		proxiedConfig.Host = fmt.Sprintf("%s://%s/clusters/%s/services/tunnels/syncer-proxy/%s", u.Scheme, u.Host, wsClusterName, syncerFixture.SyncerConfig.SyncTargetName)
+
+		proxiedKcpClient, err := kubernetesclientset.NewForConfig(proxiedConfig)
+		require.NoError(t, err)
+
+		framework.Eventually(t, func() (bool, string) {
+			deployment, err = proxiedKcpClient.AppsV1().Deployments(downstreamNamespaceName).Get(ctx, iccUpstreamDeployment.Name, metav1.GetOptions{})
+			if apierrors.IsNotFound(err) {
+				return false, ""
```
sttts, line 296: return the error as string: `return false, fmt.Sprintf("some context: %v", err)`

test/e2e/syncer/syncer_test.go
```diff
@@ -251,6 +255,76 @@ func TestSyncerLifecycle(t *testing.T) {
...
+			}
+			require.NoError(t, err)
+			pods, err := proxiedKcpClient.CoreV1().Pods(downstreamNamespaceName).List(ctx, metav1.ListOptions{})
+			if apierrors.IsUnauthorized(err) {
+				return false, ""
```
sttts, line 301: same here

test/e2e/syncer/syncer_test.go
```diff
@@ -251,6 +255,76 @@ func TestSyncerLifecycle(t *testing.T) {
...
+						Namespace(pod.Namespace).
+						Name(pod.Name).SubResource("log").
+						Param("container", c.Name)
+					logs, err := request.Do(context.TODO()).Raw()
+					if err != nil {
+						klog.Errorf("Failed to get logs for pod %s/%s container %s: %v", pod.Namespace, pod.Name, c.Name, err)
+						continue
+					}
+					for _, line := range strings.Split(string(logs), "\n") {
+						t.Logf("Pod %s/%s container %s: %s", pod.Namespace, pod.Name, c.Name, line)
```
sttts, line 319: will this spam the test logs? How many lines will we see? How many containers?

test/e2e/syncer/syncer_test.go
aojea: I was thinking they were synced, and is how an user will use them ... are they not synced?

aojea: oh, the service accounts are synced, not the roles, damn

pkg/syncer/syncer.go
aojea: or do we have some Default in other place?

test/e2e/syncer/syncer_test.go
aojea: I've copied/pasted from the existing `podLogs` function, there is only one pod with no many logs

aojea: let me change it

test/e2e/syncer/syncer_test.go
aojea: yes

pkg/syncer/syncer.go
sttts: this different backoff is fine. Just move it out of this function.

pkg/syncer/tunneler.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	proxy.Transport = clientDownstream.Transport
+
+	u, err := url.Parse(upstream.Host)
+	if err != nil {
+		return err
+	}
+	// create the reverse connection
+	// virtual workspaces
+	dst := fmt.Sprintf("%s://%s/clusters/%s", u.Scheme, u.Host, syncTargetWorkspace)
```
sttts, line 67: nit: `syncTargetWorkspace.Path()` gives you `/clusters/%s`

pkg/tunneler/listener.go
sttts: this is still open

pkg/syncer/tunneler.go
```diff
@@ -0,0 +1,92 @@
...
+	klog.Infof("connecting to %s|%s", dst, syncTargetWorkspace, syncTargetName)
```
sttts, line 68:
```suggestion
	klog.Infof("Connecting to %s|%s", dst, syncTargetWorkspace, syncTargetName)
```

pkg/tunneler/integration_test.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	proxy := httputil.NewSingleHostReverseProxy(url)
+	proxy.Transport = backend.Client().Transport
+	server := &http.Server{Handler: proxy}
+	//nolint:errcheck
+	go server.Serve(l)
+
+	// client
+	// wait for the reverse connection to be established
+	time.Sleep(1 * time.Second)
```
sttts, line 80: why a second? Can this flake?

pkg/tunneler/integration_test.go
```diff
@@ -0,0 +1,228 @@
...
+	proxy := httputil.NewSingleHostReverseProxy(url)
+	proxy.Transport = backend.Client().Transport
+	server := &http.Server{Handler: proxy}
+	//nolint:errcheck
+	go server.Serve(l)
+	defer server.Close()
+
+	// client
+	// wait for the reverse connection to be established
+	time.Sleep(1 * time.Second)
```
sttts, line 176: why 1s? Can this flake?

pkg/tunneler/listener.go
```diff
@@ -0,0 +1,290 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tunneler
+
+// Based on https://github.com/golang/build/blob/master/tunneler/v2/tunneler.go
```
sttts, line 19: worth to comment why we need a fork?

pkg/tunneler/listener.go
```diff
@@ -0,0 +1,290 @@
...
+		return err
+	}
+
+	t2.ReadIdleTimeout = time.Duration(30) * time.Second
+	t2.PingTimeout = time.Duration(15) * time.Second
+	return nil
+}
+
+// syncerTunnelURL builds the destination url with the Dialer expected format of the URL
+func syncerTunnelURL(host string, id string) (string, error) {
```
sttts, line 275: where does the forked tunneler.go code end? Can we move everything into a different file that is not coming from the upstream file?

pkg/tunneler/integration_test.go
aojea: I run all my tests with `go race -c` and then `stress` to check they don't flake , I can't remember now if I did it with last versions, but I'll do it befor emerge

pkg/tunneler/listener.go
aojea: is not really a fork, I took only the idea and the algorithm, but the code is  different, I just wanted to give credit

pkg/tunneler/listener.go
aojea: the code is not really forked, there are some tweaks I've done

pkg/tunneler/integration_test.go
aojea: it doesn't flake

```
./tunneler.test -test.run Test_integration_listener_reconnect
5s: 24 runs so far, 0 failures
10s: 48 runs so far, 0 failures
15s: 72 runs so far, 0 failures
20s: 96 runs so far, 0 failures
25s: 144 runs so far, 0 failures
30s: 168 runs so far, 0 failures
35s: 192 runs so far, 0 failures
40s: 216 runs so far, 0 failures
45s: 240 runs so far, 0 failures
50s: 288 runs so far, 0 failures
55s: 312 runs so far, 0 failures
1m0s: 336 runs so far, 0 failures
1m5s: 360 runs so far, 0 failures
```

pkg/tunneler/integration_test.go
aojea: https://github.com/kcp-dev/kcp/pull/873/commits/ca18e5b3e6dd09216c294fd93cdc2e4e4350e049#r960457232

pkg/tunneler/listener_test.go
```diff
@@ -0,0 +1,81 @@
...
+		target  string
+		want    string
+		wantErr bool
+	}{
+		{
+			name:   "valid",
+			host:   "https://host:9443/base",
+			ws:     "myws",
+			target: "syncer001",
+			want:   "https://host:9443/base/services/syncer-tunnels/clusters/myws/apis/workload.kcp.dev/v1alpha1/synctargets/syncer001",
```
aojea, line 37: @sttts the new urls

pkg/tunneler/tunnel.go
```diff
@@ -0,0 +1,261 @@
...
+		return "", fmt.Errorf("wrong url format, expected https://host<:port>/<path>: %w", err)
+	}
+	host = strings.Trim(host, "/")
+	return host + defaultTunnelPathPrefix + "/" + ws + "/apis/" + workloadv1alpha1.SchemeGroupVersion.String() + "/synctargets/" + target, nil
+}
+
+// HTTP Handler that handles reverse connections and reverse proxy requests using 2 different paths:
+//
+// https://host/services/syncer-tunnels/clusters/<ws>/apis/workload.kcp.dev/v1alpha1/synctargets/<name>/connect establish reverse connections and queue them so it can be consumed by the dialer
+// https://host/services/syncer-tunnels/clusters/<ws>/apis/workload.kcp.dev/v1alpha1/synctargets/<name>/proxy/{path} proxies the {path} through the reverse connection identified by the cluster and syncer name
```
aojea, line 115: new urls

pkg/tunneler/listener.go
aojea: fixed, now

pkg/syncer/syncer.go
aojea: done

pkg/syncer/tunneler.go
aojea: fixed

pkg/tunneler/integration_test.go
stevekuznetsov: (now run it in a CPU-constrained environment)

aojea: that is how we check the tests for Kubernetes/kubernetes , we didn't have flakes since we are doing that

aojea: all credit is for Jordan Liggitt of course https://gist.github.com/liggitt/6a3a2217fa5f846b52519acfc0ffece0

pkg/tunneler/integration_test.go
sttts: what I mean is: if you didn't wait, would it flake?

pkg/tunneler/listener.go
sttts: ack

pkg/tunneler/tunnel.go
```diff
@@ -0,0 +1,261 @@
...
+			path[4] != "synctargets" {
+			http.Error(w, "invalid path", http.StatusInternalServerError)
+			return
+		}
+
+		clusterName := path[0]
+		syncerName := path[5]
+		command := path[6]
+
+		klog.V(5).InfoS("tunneler connection received", "command", command, "clusterName", clusterName, "syncerName", syncerName)
```
sttts, line 146: not sure CI complains, but this should use the logger from the request context.

pkg/tunneler/tunnel.go
```diff
@@ -0,0 +1,261 @@
...
+			}
+			// only proxy the proxied path and don't forward the authentication header
+			proxy.Director = func(req *http.Request) {
+				// strip the non-proxied path
+				proxypath := "/"
+				if len(path) > 7 {
+					proxypath += strings.Join(path[7:], "/")
+				}
+				req.URL.Path = proxypath
+				// TODO: strip authorization header?????
```
sttts, line 225: definitely, this is a must.

sttts:
/lgtm
/approve

Now we need green tests 🙏

go.sum
```diff
@@ -90,6 +90,10 @@ github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d/go.mod h1:rBZYJk5
 github.com/antihax/optional v1.0.0/go.mod h1:uupD/76wgC+ih3iEmQUL+0Ugr19nfwCT1kdvxnR2qWY=
 github.com/antlr/antlr4/runtime/Go/antlr v0.0.0-20210826220005-b48c857c3a0e h1:GCzyKMDDjSGnlpl3clrdAK7I1AaVoaiKDOYkUzChZzg=
 github.com/antlr/antlr4/runtime/Go/antlr v0.0.0-20210826220005-b48c857c3a0e/go.mod h1:F7bn7fEU90QkQ3tnmaTx3LTKLEDqnwWODIYppRQ5hnY=
+github.com/aojea/rwconn v0.0.0-20220622092013-13e3141a2a4e h1:rldE7HndxqaNyukzKCUFnWEgwt5WtwESLt6evEvoeHw=
+github.com/aojea/rwconn v0.0.0-20220622092013-13e3141a2a4e/go.mod h1:LUO0QX1YNsA51BR48slR87GsvEMiTTOWNdC6aoG+BTA=
```
hardys, line 94: Old version here, I think you need to `go mod tidy` to clean this up?

aojea: oh, good catch

pkg/tunneler/integration_test.go
aojea: yes, the listener needs to connect to the handler so the reverse connection is established before start testing

We can do an active check instead of the time.sleep , but this is reliable and scoped to this setup() function only ... I'll add it to my TODO backlog

PR #877, opened 2022-04-08
Add APIExport controller

ncdc:
## Summary
- Add APIExport controller that generates and/or verifies an APIExport's
identity secret.
- Update APIBindings to record the hash of the APIExport's identity when
binding.
- CRDs are now stored in etcd in /registry/$group/$version/$apiExportIdentity/...
    - If a CRD isn't part of an APIExport, the $apiExportIdentity value is `customresources`

## Related issue(s)

Requires https://github.com/kcp-dev/kubernetes/pull/62


pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -0,0 +1,383 @@
...
+				apisv1alpha1.IdentityGenerationFailedReason,
+				conditionsv1alpha1.ConditionSeverityError,
+				"Error generating identity private key: %v",
+				err,
+			)
+
+			return fmt.Errorf("error generating APIExport %s|%s identity private key: %w", clusterName, apiExport.Name, err)
+		}
+
+		encoded, err := keyutil.MarshalPrivateKeyToPEM(privateKey)
```
sttts, line 302: technically we only need a random secret. An RSA keht would allow us to sign, if we want to. Still wondering how (other than cross-shard SubjectAccessReview) we make a shard to accept wildcard reads. Maybe signing could help (also note that a service account has the workspace today, and with that at least can be checked offline by a shard).

sttts: Make me think whether we need some generic authorization proof concept in SA JWTs. The bound service tokens actually could give us that. The APIExport hash could be an audience in the JWT.

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -421,6 +422,7 @@ func crdFromAPIResourceSchema(schema *apisv1alpha1.APIResourceSchema) (*apiexten
 				annotationBoundCRDKey:      "",
 				annotationSchemaClusterKey: schema.ClusterName,
 				annotationSchemaNameKey:    schema.Name,
+				annotationAPIIdentityKey:   hash,
```
sttts, line 426: this will mean two schemas with different identities (can happen with multiple APIExports) will fight.

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -0,0 +1,465 @@
...
+		kubeClusterClient: kubeClusterClient,
+		getNamespace: func(clusterName logicalcluster.LogicalCluster, name string) (*corev1.Namespace, error) {
+			return namespaceInformer.Lister().Get(clusters.ToClusterAwareKey(clusterName, name))
+		},
+		createNamespace: func(ctx context.Context, clusterName logicalcluster.LogicalCluster, ns *corev1.Namespace) error {
+			_, err := kubeClusterClient.Cluster(clusterName).CoreV1().Namespaces().Create(ctx, ns, metav1.CreateOptions{})
+			return err
+		},
+		secretLister:    secretInformer.Lister(),
+		secretNamespace: "default",
```
sttts, line 88: this might be a security problem. Would rather prefer a namespace like kube-system where users have no secret-read access.

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -0,0 +1,465 @@
...
+
+	hashBytes := sha256.Sum256(key)
+	hash := fmt.Sprintf("%x", hashBytes)
+
+	if apiExport.Status.IdentityHash == "" {
+		apiExport.Status.IdentityHash = hash
+	}
+
+	if apiExport.Status.IdentityHash != hash {
+		return fmt.Errorf("hash mismatch: identity secret hash %q should match status.identityHash %q", hash, apiExport.Status.IdentityHash)
```
sttts, line 459: s/should/must/

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
ncdc: Thanks, will remove

pkg/reconciler/apis/apiexport/apiexport_controller.go
ncdc: Use case: user has create-apiexport permission but not read-secrets?

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
ncdc: I knew I wanted to come back to this! In https://github.com/kcp-dev/kubernetes/pull/57/files that we just merged, it uses this annotation to alter the resource prefix of the CRs for the CRD in etcd.

I'm guessing you're thinking of a scenario like this?

- APIExport E1 exports APIResourceSchemas S1, S2
- APIExport E2 exports S1, S2' (a variant of S2)
- User creates APIBinding B1 in ws W1 pointing to E1
    - This creates CRD C1 in system:bound-crds with name=S1.metadata.UID, identityAnnotation=E1.identityHash
    - This creates CRD C2 in system:bound-crds with name=S2.metadata.UID, identityAnnotation=E1.identityHash
- User creates APIBinding B2 in ws W2 pointing to E2
    - Because C1 already exists (for S1), we don't try to create a new CRD (even though E2 has a different identityHash)
    - This creates CRD C3 in system:bound-crds with name=S2'.metadata.UID, identityAnnotation=E2.identityHash

What this tells me is that bound CRDs can't be associated back to identity hashes.

So the question becomes: how do we specify a schema's identity for storage in etcd, for instances of said schema? It's more like each APIResourceSchema needs a separate associated APIIdentity object (?) to represent compatibility and to be able to distinguish when there are breaking changes. For example, S2 and S2' could both share an APIIdentity if their actual schemas are compatible (e.g. S2' adds a new optional field). That isn't tied to the APIExport per se.

Thoughts?

ncdc: Chatted on Slack a few days ago. When getting a CRD for a specific known identity, we'll add the annotation on the fly (not stored in etcd). And the identity is via the APIExport, not a separate resource.

pkg/reconciler/apis/apiexport/apiexport_controller.go
ncdc: Fixed

ncdc:
@sttts ready for review (don't merge until we merge the kube PR)

ncdc:
FYI @kylape @csams this changes where CRs are stored in etcd.

pkg/apis/apis/v1alpha1/types.go
```diff
@@ -251,6 +267,26 @@ type APIBindingList struct {
...
+
+	IdentityVerificationFailedReason = "IdentityVerificationFailed"
+	IdentityGenerationFailedReason   = "IdentityGenerationFailed"
+)
+
+// These are for APIExport identity.
+const (
+	// AnnotationKeyAPIExportCluster is the key for an annotation on an identity secret for an APIExport for the
+	// cluster name for the associated APIExport.
+	AnnotationKeyAPIExportCluster = "apis.kcp.dev/apiexport-cluster"
```
sttts, line 282: why is this needed? Isn't the secret in the same workspace?

pkg/apis/apis/v1alpha1/types.go
```diff
@@ -251,6 +267,26 @@ type APIBindingList struct {
...
+	// AnnotationKeyAPIExportName is the key for an annotation on an identity secret for an APIExport for the
+	// name for the associated APIExport.
+	AnnotationKeyAPIExportName = "apis.kcp.dev/apiexport-name"
```
sttts, line 285: does this mean we cannot share the identity secret between exports?

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -0,0 +1,489 @@
...
+
+// enqueueAPIBinding enqueues an APIExport .
+func (c *controller) enqueueAPIExport(obj interface{}) {
+	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+
+	klog.V(4).Infof("Queueing APIExport %q", key)
```
sttts, line 175: `V(2)` for pure progress

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -0,0 +1,489 @@
...
+
+	apiExportCluster := secret.Annotations[apisv1alpha1.AnnotationKeyAPIExportCluster]
+	apiExportName := secret.Annotations[apisv1alpha1.AnnotationKeyAPIExportName]
+	if apiExportCluster == "" || apiExportName == "" {
+		return
+	}
+
+	key := clusters.ToClusterAwareKey(logicalcluster.New(apiExportCluster), apiExportName)
+
+	klog.V(4).Infof("Queueing APIExport %q via identity secret %s|%s", key, logicalcluster.From(secret), secret.Name)
```
sttts, line 207: `V(2)` for pure progress.

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -0,0 +1,489 @@
...
+
+	if secret.Annotations == nil {
+		return
+	}
+
+	apiExportCluster := secret.Annotations[apisv1alpha1.AnnotationKeyAPIExportCluster]
+	apiExportName := secret.Annotations[apisv1alpha1.AnnotationKeyAPIExportName]
+	if apiExportCluster == "" || apiExportName == "" {
+		return
+	}
```
sttts, line 203: I think this coupling of the secret and the export is not good. We should make the secret reusable, and also very important: never delete it, i.e. not owner reference (haven't checked whether we have set one).

In other words, here we have to go through all exports (there won't be many, usually one) and check the secret ref.

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -0,0 +1,489 @@
...
+	if err := c.patchIfNeeded(ctx, old, obj); err != nil {
+		errs = append(errs, err)
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+func (c *controller) patchIfNeeded(ctx context.Context, old, obj *apisv1alpha1.APIExport) error {
+	specChanged := !equality.Semantic.DeepEqual(old.Spec, obj.Spec)
+	statusChanged := !equality.Semantic.DeepEqual(old.Status, obj.Status)
```
sttts, line 282: this pattern is interesting. In other controller we update spec and metadata from the reconciler. Maybe this pattern is better.

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -0,0 +1,489 @@
...
+
+	if specChanged && statusChanged {
+		klog.Fatalf("Programmer error: spec and status changed in same reconcile iteration")
```
sttts, line 285: not agreeing this is wrong.

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -0,0 +1,489 @@
...
+	// In case the lister is slow to catch up, try a live read
+	secret, err = c.kubeClusterClient.Cluster(clusterName).CoreV1().Secrets(ns).Get(ctx, name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	return secret, nil
+}
+
+func (c *controller) reconcile(ctx context.Context, apiExport *apisv1alpha1.APIExport) error {
```
sttts, line 349: I liked the separation into controller and reconciler we used before. It made it clear how to unit test a reconciler and separated business logic nicely from plumbing. This pattern here in the file will be hard to maintain and will quickly diverge. With the separation you know what is standard plumbing and what is custom business logic.

pkg/server/apiextensions.go
```diff
@@ -144,27 +142,23 @@ func (p *systemCRDProvider) Keys(ctx context.Context, clusterName logicalcluster
...
 
-// apiBindingAwareCRDLister is a CRD lister combines APIs coming from APIBindings with CRDs in a workspace.
-type apiBindingAwareCRDLister struct {
-	kcpClusterClient  kcpclientset.ClusterInterface
-	crdLister         apiextensionslisters.CustomResourceDefinitionLister
-	workspaceLister   tenancylisters.ClusterWorkspaceLister
-	apiBindingLister  apislisters.APIBindingLister
-	systemCRDProvider *systemCRDProvider
+// APIBindingAwareCRDLister is a CRD lister combines APIs coming from APIBindings with CRDs in a workspace.
+type APIBindingAwareCRDLister struct {
```
sttts, line 146: used outside of the package?

pkg/server/apiextensions.go
```diff
@@ -173,10 +167,19 @@ func (c *apiBindingAwareCRDLister) ListWithContext(ctx context.Context, selector
 		return crd.Spec.Names.Plural + "." + crd.Spec.Group
 	}
 
+	selectAll := selector.Empty()
+	matchesSelector := func(crd *apiextensionsv1.CustomResourceDefinition) bool {
+		if selectAll {
```
sttts, line 172: that's the default behaviour, no?

pkg/server/apiextensions.go
```diff
@@ -245,117 +259,226 @@ func (c *apiBindingAwareCRDLister) ListWithContext(ctx context.Context, selector
...
+
+	updatedCRD, err := c.crdLister.Get(crdKey)
 	if err != nil {
 		return nil, err
 	}
 
-	if strings.HasSuffix(name, ".") {
-		name = name + "core"
+	if crd.Annotations != nil {
+		if identity := crd.Annotations[apisv1alpha1.AnnotationAPIIdentityKey]; identity != "" {
```
sttts, line 281: does this mean the identity is stored in the bound CRD? The bound CRD is only named according to the schema UID. So we might get into conflict if multiple export with different identities reference the same schema? Seems to be wrong.

pkg/server/apiextensions.go
```diff
@@ -245,117 +259,226 @@ func (c *apiBindingAwareCRDLister) ListWithContext(ctx context.Context, selector
...
+	crdKey := clusters.ToClusterAwareKey(logicalcluster.From(crd), crd.Name)
+
+	updatedCRD, err := c.crdLister.Get(crdKey)
 	if err != nil {
 		return nil, err
 	}
 
-	if strings.HasSuffix(name, ".") {
-		name = name + "core"
+	if crd.Annotations != nil {
```
sttts, line 280: no need for the check.

pkg/server/apiextensions.go
```diff
@@ -245,117 +259,226 @@ func (c *apiBindingAwareCRDLister) ListWithContext(ctx context.Context, selector
...
+
+	if crd == nil {
+		return nil, errors.NewNotFound(apiextensionsv1.Resource("customresourcedefinitions"), name)
+	}
+
+	return crd, nil
+}
+
+func (c *APIBindingAwareCRDLister) getForIdentity(name, identity string) (*apiextensionsv1.CustomResourceDefinition, error) {
+	apiExports, err := c.apiExportIndexer.ByIndex("byIdentity", identity)
```
sttts, line 383: use string constant for index name.

pkg/server/handler.go
```diff
@@ -130,7 +129,8 @@ func WithClusterScope(apiHandler http.Handler) http.HandlerFunc {
 		case clusterName == logicalcluster.Wildcard:
 			// HACK: just a workaround for testing
 			cluster.Wildcard = true
-			fallthrough
+			//fallthrough
```
sttts, line 132: no fallthrough anymore?

pkg/server/handler.go
```diff
@@ -234,7 +234,100 @@ func WithInClusterServiceAccountRequestRewrite(handler http.Handler, unsafeServi
 	})
 }
 
-func mergeCRDsIntoCoreGroup(crdLister v1.CustomResourceDefinitionLister, crdHandler, coreHandler func(res http.ResponseWriter, req *http.Request)) restful.FilterFunction {
+func WithWildcardIdentityGuard(handler http.Handler) http.Handler {
```
sttts, line 237: doc what this does

pkg/server/handler.go
```diff
@@ -234,7 +234,100 @@ func WithInClusterServiceAccountRequestRewrite(handler http.Handler, unsafeServi
...
+				errorCodecs, schema.GroupVersion{}, w, req,
+			)
+			return
+		}
+		if !requestInfo.IsResourceRequest {
+			handler.ServeHTTP(w, req)
+			return
+		}
+
+		// SUPER HACK
```
sttts, line 258: don't we have a list of system CRDs elsewhere? I think I also saw here in the PR that `kcp.dev` is taken as hint that it is a system CRD.

pkg/server/handler.go
```diff
@@ -234,7 +234,100 @@ func WithInClusterServiceAccountRequestRewrite(handler http.Handler, unsafeServi
...
+			// Wildcard but not a possible API resource request - let it through
+			handler.ServeHTTP(w, req)
+			return
+		}
+
+		if resource == "" {
+			responsewriters.ErrorNegotiated(
+				apierrors.NewInternalError(fmt.Errorf("missing resource")),
+				errorCodecs, schema.GroupVersion{}, w, req,
+			)
```
sttts, line 298: not 404 or 403?

pkg/server/handler.go
```diff
@@ -234,7 +234,100 @@ func WithInClusterServiceAccountRequestRewrite(handler http.Handler, unsafeServi
...
+				return
+			}
+
+			ctx := WithIdentity(req.Context(), identity)
+			req = req.WithContext(ctx)
+
+			req.URL.Path = strings.Replace(req.URL.Path, resourceAndIdentity, resource, 1)
+			req.URL.RawPath = strings.Replace(req.URL.RawPath, resourceAndIdentity, resource, 1)
+			requestInfo.Resource = resource
+		}
```
sttts, line 324: this func deserves a unit-test

test/e2e/conformance/cross_logical_cluster_list_test.go
```diff
@@ -63,23 +59,23 @@ func TestCrossLogicalClusterList(t *testing.T) {
 
 	// Until we get rid of the multiClusterClientConfigRoundTripper and replace it with scoping,
 	// make sure we don't break cross-logical cluster client listing.
-	clientutils.EnableMultiCluster(cfg, nil, true)
+	//clientutils.EnableMultiCluster(cfg, nil, true)
```
sttts, line 62: remove?

test/e2e/conformance/cross_logical_cluster_list_test.go
```diff
@@ -63,23 +59,23 @@ func TestCrossLogicalClusterList(t *testing.T) {
...
-		require.NoError(t, err, "failed to bootstrap CRDs")
+		//t.Logf("Bootstrapping ClusterWorkspace CRDs in logical cluster %s", logicalCluster)
+		//apiExtensionsClients, err := apiextensionsclient.NewClusterForConfig(cfg)
+		//require.NoError(t, err, "failed to construct apiextensions client for server")
+		//crdClient := apiExtensionsClients.Cluster(logicalCluster).ApiextensionsV1().CustomResourceDefinitions()
+		//workspaceCRDs := []metav1.GroupResource{
+		//	{Group: tenancy.GroupName, Resource: "clusterworkspaces"},
+		//}
+		//err = configcrds.Create(ctx, crdClient, workspaceCRDs...)
+		//require.NoError(t, err, "failed to bootstrap CRDs")
```
sttts, line 78: remove?

test/e2e/conformance/cross_logical_cluster_list_test.go
```diff
@@ -134,26 +129,82 @@ func TestCrossLogicalClusterListPartialObjectMetadata(t *testing.T) {
 
 	cfg := server.DefaultConfig(t)
 
+	// Make sure the informers aren't throttled
+	cfg.QPS = 500
+	cfg.Burst = 1000
+
```
sttts, line 135: why not?

pkg/server/apiextensions.go
```diff
@@ -144,27 +142,23 @@ func (p *systemCRDProvider) Keys(ctx context.Context, clusterName logicalcluster
...
+var _ kcp.ClusterAwareCRDLister = &APIBindingAwareCRDLister{}
 
-// List lists those CustomResourceDefinitions in the underlying store matching selector. This method does not
-// support scoping to logical clusters or APIBindings.
-func (c *apiBindingAwareCRDLister) List(selector labels.Selector) ([]*apiextensionsv1.CustomResourceDefinition, error) {
-	return c.crdLister.ListWithContext(context.Background(), selector)
-}
-
-// ListWithContext lists all CustomResourceDefinitions that come in via APIBindings as well as all in the current
+// List lists all CustomResourceDefinitions that come in via APIBindings as well as all in the current
```
sttts, line 158: ... current logical cluster retrieved from the context.

pkg/server/apiextensions.go
```diff
@@ -144,27 +142,23 @@ func (p *systemCRDProvider) Keys(ctx context.Context, clusterName logicalcluster
...
 // logical cluster.
-func (c *apiBindingAwareCRDLister) ListWithContext(ctx context.Context, selector labels.Selector) ([]*apiextensionsv1.CustomResourceDefinition, error) {
-	cluster, err := genericapirequest.ValidClusterFrom(ctx)
+func (c *APIBindingAwareCRDLister) List(ctx context.Context, selector labels.Selector) ([]*apiextensionsv1.CustomResourceDefinition, error) {
```
sttts, line 160: what's the selector used for?

ncdc: Standard lister signature

test/e2e/conformance/cross_logical_cluster_list_test.go
ncdc: This is just for testing because it is starting the dynamic shared informer factory and it makes lots of discovery calls. Shortens test execution time.

pkg/server/apiextensions.go
```diff
@@ -217,18 +224,25 @@ func (c *apiBindingAwareCRDLister) ListWithContext(ctx context.Context, selector
 			}
 
 			// Priority 2: Add APIBinding CRDs. These take priority over those from the local workspace.
+
+			// Add the APIExport identity hash as an annotation to the CRD so the RESTOptionsGetter can assign
+			// the correct etcd resource prefix.
+			crd = crd.DeepCopy()
```
sttts, line 230: if you don't mind, would rather just do 
```
shallowCopy := *crd
shallowCopy.Annotations = <some-map-deepcopy>
....
```
CRDs are giant. This is a pretty heavy operation.

pkg/server/apiextensions.go
```diff
@@ -217,18 +224,25 @@ func (c *apiBindingAwareCRDLister) ListWithContext(ctx context.Context, selector
...
+			crd.Annotations[apisv1alpha1.AnnotationAPIIdentityKey] = boundResource.Schema.IdentityHash
+
 			ret = append(ret, crd)
 			seen.Insert(crdName(crd))
 		}
 	}
 
-	crds, err := c.crdLister.List(labels.Everything())
+	// TODO use scoping lister when available
+	crds, err := c.crdLister.List(selector)
```
sttts, line 239: index not worth it for now?

pkg/server/apiextensions.go
```diff
@@ -245,117 +259,226 @@ func (c *apiBindingAwareCRDLister) ListWithContext(ctx context.Context, selector
...
 	}
 
-	if strings.HasSuffix(name, ".") {
-		name = name + "core"
+	if crd.Annotations != nil {
+		if identity := crd.Annotations[apisv1alpha1.AnnotationAPIIdentityKey]; identity != "" {
+			if updatedCRD.Annotations == nil {
+				updatedCRD.Annotations = make(map[string]string)
+			}
+			updatedCRD.Annotations[apisv1alpha1.AnnotationAPIIdentityKey] = identity
```
sttts, line 285: is this mutating the lister object?

pkg/server/apiextensions.go
```diff
@@ -245,117 +259,226 @@ func (c *apiBindingAwareCRDLister) ListWithContext(ctx context.Context, selector
...
-			}
-		} else {
-			var equal bool // true if all the found CRDs have the same spec
-			crd, equal = findCRD(name, crds)
-			if !equal {
-				err = apierrors.NewInternalError(fmt.Errorf("error resolving resource: cannot watch across logical clusters for a resource type with several distinct schemas"))
-				return nil, err
-			}
-		}
+	if strings.Contains(name, "kcp.dev") {
```
sttts, line 308: this means we have different concepts of system CRDs. Can't we call `getSystemCRD` and fall through on IsNotFound?

pkg/server/apiextensions.go
```diff
@@ -245,117 +259,226 @@ func (c *apiBindingAwareCRDLister) ListWithContext(ctx context.Context, selector
...
-	if systemCRDKeys.Has(systemCRDKeyName) {
-		crd, err = c.crdLister.Get(systemCRDKeyName)
-		if err != nil && !apierrors.IsNotFound(err) {
-			// something went wrong w/the lister - could only happen if meta.Accessor() fails on an item in the store.
-			return nil, err
+	return crd, nil
+}
+
+func partialMetadataCRD(crd *apiextensionsv1.CustomResourceDefinition) *apiextensionsv1.CustomResourceDefinition {
+	crd = crd.DeepCopy()
```
sttts, line 334: as above, shallow copy

pkg/server/apiextensions.go
sttts, line 383: this confuses me. Why do you need the export in the first place? What is missing in the binding to be sufficient here?

pkg/server/apiextensions.go
```diff
@@ -245,117 +259,226 @@ func (c *apiBindingAwareCRDLister) ListWithContext(ctx context.Context, selector
...
+	group, resource := crdNameToGroupResource(name)
+
+	// If for some reason multiple APIExports share the same identity, that shouldn't matter. We just need any
+	// one of them.
+	apiExport := apiExports[0].(*apisv1alpha1.APIExport)
+	apiExportClusterName := logicalcluster.From(apiExport)
+
+	var boundCRDName string
+
+	for _, schemaName := range apiExport.Spec.LatestResourceSchemas {
```
sttts, line 401: the export schemas can be totally different by now. The binding should count, until transactionally we switch over to new schemas and store them in the binding status.

pkg/server/apiextensions.go
```diff
@@ -371,13 +494,19 @@ func (c *apiBindingAwareCRDLister) GetWithContext(ctx context.Context, name stri
 					// something went wrong w/the lister - could only happen if meta.Accessor() fails on an item in the store.
 					return nil, err
 				}
+
+				// Add the APIExport identity hash as an annotation to the CRD so the RESTOptionsGetter can assign
+				// the correct etcd resource prefix.
+				crd = crd.DeepCopy()
```
sttts, line 500: shallow copy

pkg/apis/apis/v1alpha1/types.go
ncdc: That wasn't the intent, after chatting with you a couple of weeks ago. I need to undo this.

pkg/apis/apis/v1alpha1/types.go
ncdc: Probably just me adding unnecessary things. Will see about undoing.

ncdc: Oh yeah, now I remember why I added these. I wanted a way to requeue an APIExport whenever a secret event came in for a secret that was used as an identity. I can probably rework this with an index.

pkg/apis/apis/v1alpha1/types.go
ncdc: Fixed

pkg/server/handler.go
ncdc: I took this out because "wildcard, but cluster.Name is actually system:admin" was breaking things in the CRD lister. All the e2es continue to pass... 🤷 😄

pkg/server/handler.go
ncdc: We do have a list, but it's grouped by workspace type. That's why I took this approach.

And I later wrote the code you reference, to just check for kcp.dev. I think I prefer that approach plus admission to reject kcp.dev anywhere other than system:system-crds.

ncdc: Actually, I can remove this check. It's not needed. I originally had it because this handler was going to reject wildcard requests that didn't have an identity, but now that identity is optional, we can remove it.

pkg/server/handler.go
ncdc: Refactored & added unit test

pkg/server/handler.go
ncdc: Added

pkg/server/handler.go
ncdc:
```
k get --raw '/clusters/*/apis/0b7ba5a9-e9ae-4e5d-b876-5ee5076261ac.io/v1/sheriffs:'
Error from server (InternalError): Internal error occurred: invalid resource "sheriffs": missing identity

k get --raw '/clusters/*/apis/0b7ba5a9-e9ae-4e5d-b876-5ee5076261ac.io/v1/sheriffs:a'
Error from server (NotFound): the server could not find the requested resource
```

What do you think makes the most sense?

pkg/reconciler/apis/apiexport/apiexport_controller.go
ncdc: Changed to kcp-system

pkg/reconciler/apis/apiexport/apiexport_controller.go
ncdc: Fixed

pkg/reconciler/apis/apiexport/apiexport_controller.go
ncdc: Fixed

pkg/reconciler/apis/apiexport/apiexport_controller.go
ncdc: Fixed

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -0,0 +1,489 @@
...
+	// If the object being reconciled changed as a result, update it.
+	if err := c.patchIfNeeded(ctx, old, obj); err != nil {
+		errs = append(errs, err)
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+func (c *controller) patchIfNeeded(ctx context.Context, old, obj *apisv1alpha1.APIExport) error {
+	specChanged := !equality.Semantic.DeepEqual(old.Spec, obj.Spec)
```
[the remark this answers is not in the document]
ncdc: 👍 will fix

pkg/reconciler/apis/apiexport/apiexport_controller.go
ncdc: If you want to safely, reliably change spec and status in 1 iteration, you really need something like https://github.com/kubernetes-sigs/cluster-api/blob/2982a3219bc6589a0a50b712b2d300a6f0b96026/util/patch/patch.go#L83.

Are you ok merging as-is & we can discuss later? Or do you want to talk first?

pkg/reconciler/apis/apiexport/apiexport_controller.go
ncdc: Fixing

pkg/server/apiextensions.go
ncdc: I don't remember why I exported it. Fixing

pkg/server/apiextensions.go
ncdc: Matching this logic: https://github.com/kubernetes/kubernetes/blob/e7a2ce75e5df96ba6ea51d904bf2735397b3e203/staging/src/k8s.io/client-go/tools/cache/listers.go#L35-L40

pkg/server/apiextensions.go
ncdc: Discussed elsewhere but recording here: no, the identity annotation is not stored in etcd. This GetLatest method is more like a Refresh (happy to rename if needed). It's only currently called by apiextensions-apiserver's getOrCreateServingInfo. And it's only here to refresh from the cache to avoid a race condition. The flow is:

1. CustomResourceHandler gets the CRD (using our lister)
2. CustomResourceHandler wants to get/create serving info for the given *CustomResourceDefinition
3. CustomResourceHandler wants to refresh from the lister (cache) to make sure it's not operating on stale data (because another goroutine maybe updated it in between step 1 and now)

Because the CRD came from our lister, it went through all the logic for identity, partial metadata, APIExports, etc. When refreshing from the cache, we need to make sure the refreshed CRD has the identity annotation from the original, and we trim the schema if it was originally for a partial metadata request.

pkg/server/apiextensions.go
ncdc: Fixing

pkg/server/apiextensions.go
ncdc: Fixing

test/e2e/conformance/cross_logical_cluster_list_test.go
ncdc: Done

test/e2e/conformance/cross_logical_cluster_list_test.go
ncdc: Done

pkg/server/apiextensions.go
ncdc: Great catch, thanks. Fixing

pkg/server/apiextensions.go
ncdc: Will do.

ncdc: Actually, will do later w/the wrappers 😄

pkg/server/apiextensions.go
ncdc: Good catch, thanks

pkg/server/apiextensions.go
ncdc: k

pkg/server/apiextensions.go
ncdc: This is only for `/clusters/*/apis/$group/$version/$resource:identity`. We don't know the binding.

pkg/apis/apis/v1alpha1/types.go
```diff
@@ -161,6 +162,9 @@ const (
 	APIExportInvalidReferenceReason = "APIExportInvalidReference"
 	// APIExportNotFoundReason is a reason for the APIExportValid condition that the referenced APIExport is not found.
 	APIExportNotFoundReason = "APIExportNotFound"
+	// APIExportIdentityMismatch is a reason for APIExportValid condition that the referenced APIExport has a different identity than
+	// the existin bound APIResourceSchemas.
```
sttts, line 166: typo, existing

pkg/reconciler/apis/apibinding/apibinding_reconcile_test.go
```diff
@@ -232,6 +232,11 @@ func TestReconcileBinding(t *testing.T) {
 			wantAPIExportInternalError: true,
 			wantError:                  true,
 		},
+		"APIExport doesn't have identity hash yet": {
+			apiBinding: binding.DeepCopy().
+				WithWorkspaceReference("some-workspace", "no-identity-hash").Build(),
+			wantAPIExportValid: false,
```
sttts, line 238: just to double check: in this state the CR endpoint is not up?

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -0,0 +1,359 @@
...
+)
+
+const (
+	controllerName = "kcp-apiexport"
+
+	defaultIdentitySecretNamespace = "kcp-system"
+
+	indexAPIExportBySecret   = "bySecret"
+	IndexAPIExportByIdentity = "byIdentity"
+	//indexAPIExportByIdentityGroupResource = "byIdentityGroupResource"
```
sttts, line 57: remove?

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -0,0 +1,359 @@
...
+			//		schema, err := apiResourceSchemaInformer.Lister().Get(schemaKey)
+			//		if err != nil {
+			//			// TODO log
+			//			continue
+			//		}
+			//
+			//		ret = append(ret, fmt.Sprintf("%s/%s/%s", apiExport.Status.IdentityHash, schema.Spec.Group, schema.Spec.Names.Plural))
+			//	}
+			//	return ret, nil
+			//},
```
sttts, line 114: remove?

pkg/reconciler/apis/apiexport/apiexport_controller.go
sttts: yes. If the controller makes progress. I assume it does.

sttts: Philosophy comment: it might be a programming error because actually these should be two controller if spec and status are changed.

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -0,0 +1,359 @@
...
+}
+
+func (c *controller) readThroughGetSecret(ctx context.Context, clusterName logicalcluster.LogicalCluster, ns, name string) (*corev1.Secret, error) {
+	secret, err := c.secretLister.Secrets(ns).Get(clusters.ToClusterAwareKey(clusterName, name))
+	if err == nil {
+		return secret, nil
+	}
+
+	// In case the lister is slow to catch up, try a live read
+	secret, err = c.kubeClusterClient.Cluster(clusterName).CoreV1().Secrets(ns).Get(ctx, name, metav1.GetOptions{})
```
sttts, line 353: am curious: what happens if you reference a non-existing secret. Will the controller exponentially reach max backoff (10min?) for that export?

pkg/reconciler/apis/apiexport/apiexport_controller_test.go
```diff
@@ -0,0 +1,245 @@
...
+		apiExportHasExpectedHash             bool
+		apiExportHasSomeOtherHash            bool
+		hasPreexistingVerifyFailure          bool
+		wantError                            bool
+		wantCreateSecretCalled               bool
+		wantUnsetIdentity                    bool
+		wantDefaultSecretRef                 bool
+		wantStatusHashSet                    bool
+		wantVerifyFailure                    bool
+		wantIdentityValid                    bool
```
sttts, line 55: taste comment: would add some empty lines between block of similar fields. Makes it easier to read.

sttts: same below in the test cases, at least <input><empty line><expected>

pkg/reconciler/apis/apiexport/apiexport_reconcile.go
```diff
@@ -0,0 +1,175 @@
...
+			conditionsv1alpha1.ConditionSeverityError,
+			err.Error(),
+		)
+	}
+
+	return nil
+}
+
+func (c *controller) ensureSecretNamespaceExists(ctx context.Context, clusterName logicalcluster.LogicalCluster) {
+	if _, err := c.getNamespace(clusterName, c.secretNamespace); errors.IsNotFound(err) {
```
sttts, line 100: no action needed: a direct create would do too.

pkg/server/apiextensions.go
sttts: not wrong. I just thought am empty selector matches everything anyway.

pkg/server/apiextensions.go
sttts: still open

pkg/server/apiextensions.go
```diff
@@ -262,117 +277,238 @@ func (c *apiBindingAwareCRDLister) ListWithContext(ctx context.Context, selector
...
+		return nil, err
+	}
+
+	if len(apiBindings) == 0 {
+		return nil, errors.NewNotFound(apiextensionsv1.Resource("customresourcedefinitions"), name)
+	}
 
-		if crd != nil {
-			return crd, nil
+	apiBinding := apiBindings[0].(*apisv1alpha1.APIBinding)
```
sttts, line 431: this is the point where we could add gcd logic of some kind? Worth a TODO maybe.

test/e2e/conformance/cross_logical_cluster_list_test.go
sttts: ack

sttts: add a sentence in the comment "... because dynamic informers does lots of discovery which slows down tests"

test/e2e/framework/fixture.go
```diff
@@ -202,11 +202,11 @@ func NewOrganizationFixture(t *testing.T, server RunningServer) (orgClusterName
...
-		err := clusterClient.Cluster(tenancyv1alpha1.RootCluster).TenancyV1alpha1().ClusterWorkspaces().Delete(ctx, org.Name, metav1.DeleteOptions{})
-		if apierrors.IsNotFound(err) {
-			return // ignore not found error
-		}
-		require.NoErrorf(t, err, "failed to delete organization workspace %s", org.Name)
+		//err := clusterClient.Cluster(tenancyv1alpha1.RootCluster).TenancyV1alpha1().ClusterWorkspaces().Delete(ctx, org.Name, metav1.DeleteOptions{})
+		//if apierrors.IsNotFound(err) {
+		//	return // ignore not found error
+		//}
+		//require.NoErrorf(t, err, "failed to delete organization workspace %s", org.Name)
```
sttts, line 209: must be readded

test/e2e/framework/fixture.go
```diff
@@ -252,11 +252,11 @@ func NewWorkspaceWithWorkloads(t *testing.T, server RunningServer, orgClusterNam
...
-		err := clusterClient.Cluster(orgClusterName).TenancyV1alpha1().ClusterWorkspaces().Delete(ctx, ws.Name, metav1.DeleteOptions{})
-		if apierrors.IsNotFound(err) {
-			return // ignore not found error
-		}
-		require.NoErrorf(t, err, "failed to delete workspace %s", ws.Name)
+		//err := clusterClient.Cluster(orgClusterName).TenancyV1alpha1().ClusterWorkspaces().Delete(ctx, ws.Name, metav1.DeleteOptions{})
+		//if apierrors.IsNotFound(err) {
+		//	return // ignore not found error
+		//}
+		//require.NoErrorf(t, err, "failed to delete workspace %s", ws.Name)
```
sttts, line 259: must be readded

pkg/reconciler/apis/apiexport/apiexport_controller.go
ncdc: >If the controller makes progress. I assume it does.
Yes, it's coded to change spec, then wait for next reconcile to update status

pkg/reconciler/apis/apibinding/apibinding_reconcile_test.go
ncdc: Correct. It won't create the bound CRD until there is an identity

pkg/reconciler/apis/apiexport/apiexport_controller.go
ncdc: No, the return out of reconcile is `nil`. There's an event handler on secrets that will enqueue all APIExports that have the secret listed as the identity ref.

pkg/reconciler/apis/apiexport/apiexport_reconcile.go
ncdc: When I was working on Velero, we had a user complain because GCP/GKE alerts were triggering every time Velero tried to create its namespace (the namespace had been precreated and the Velero SA didn't have permission to create it). So I tend to do get-then-create now 😄

test/e2e/framework/fixture.go
ncdc: Dropped the commit

test/e2e/framework/fixture.go
ncdc: Dropped the commit

pkg/server/apiextensions.go
ncdc: Added

pkg/reconciler/apis/apiexport/apiexport_controller_test.go
ncdc: Fixed

pkg/reconciler/apis/apiexport/apiexport_controller.go
ncdc: Done

pkg/reconciler/apis/apiexport/apiexport_controller.go
ncdc: Done

pkg/apis/apis/v1alpha1/types.go
ncdc: Removed this reason

sttts (approved):

PR #880, opened 2022-04-11
virtual: Disable workspace deletion for now

kylape:
Since workspace deletion currently leaves all resources inside that
workspace effectively orphaned, we should not allow users to delete
workspaces for now.  Admins can still delete clusterworkspaces, and we
can always add the "delete" verb to the "workspace" subresource by hand.

### DIFF
--- a/pkg/virtual/workspaces/registry/rest.go
+++ b/pkg/virtual/workspaces/registry/rest.go
@@ -415,7 +415,11 @@ const (
 var roleRules = map[RoleType][]rbacv1.PolicyRule{
 	OwnerRoleType: {
 		{
-			Verbs:     []string{"get", "delete"},
+			Verbs: []string{
+				"get",
+				// TODO: Allow deleting workspaces after kcp supports it
+				//"delete",
+			},
 			Resources: []string{"clusterworkspaces/workspace"},
 		},
 		{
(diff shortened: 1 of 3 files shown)


pkg/virtual/workspaces/registry/rest.go
```diff
@@ -678,28 +682,14 @@ func (s *REST) Delete(ctx context.Context, name string, deleteValidation rest.Va
 		Name:            internalName,
 		ResourceRequest: true,
 	}
-	if decision, _, err := authz.Authorize(ctx, deleteWorkspaceAttr); err != nil {
+	decision, reason, err := authz.Authorize(ctx, deleteWorkspaceAttr)
+	if err != nil {
```
ncdc, line 686: Curious why you split these into 2 lines?

pkg/virtual/workspaces/registry/rest.go
```diff
@@ -678,28 +682,14 @@ func (s *REST) Delete(ctx context.Context, name string, deleteValidation rest.Va
...
 		klog.Errorf("failed to authorize user %q to %q clusterworkspaces/workspace name %q in %s", userInfo.GetName(), "delete", internalName, orgClusterName)
 		return nil, false, kerrors.NewForbidden(tenancyv1beta1.Resource("workspaces"), "", fmt.Errorf("deletion in workspace %q is not allowed", orgClusterName))
 	} else if decision != authorizer.DecisionAllow {
-		// check for admin verb on the content
```
ncdc, line 685: Do you want to comment this out instead of deleting it? I imagine we'll want to reenable it once we have proper delete?

ncdc:
@davidfestal PTAL

davidfestal (commented):
Do we also want to temporarily remove ( == comment probably) the `delete` subcommand of the kcp plugin ?

pkg/virtual/workspaces/registry/rest.go
kylape: Yeah I can just comment out.

kylape:
Yes, I think that makes sense, @davidfestal.  I can comment that out, too.

kylape:
Actually the delete subcommand just prints a message to run a `kubectl delete`: 

https://github.com/kcp-dev/kcp/blob/976a437043ee5134b689d7fa601397cdcc14f9ca/pkg/cliplugins/workspace/cmd/cmd.go#L207-L216

Given that, would it make sense to just remove the subcommand altogether for now?

pkg/virtual/workspaces/registry/rest.go
```diff
@@ -678,28 +682,33 @@ func (s *REST) Delete(ctx context.Context, name string, deleteValidation rest.Va
 		Name:            internalName,
 		ResourceRequest: true,
 	}
-	if decision, _, err := authz.Authorize(ctx, deleteWorkspaceAttr); err != nil {
+	if decision, reason, err := authz.Authorize(ctx, deleteWorkspaceAttr); err != nil {
 		klog.Errorf("failed to authorize user %q to %q clusterworkspaces/workspace name %q in %s", userInfo.GetName(), "delete", internalName, orgClusterName)
 		return nil, false, kerrors.NewForbidden(tenancyv1beta1.Resource("workspaces"), "", fmt.Errorf("deletion in workspace %q is not allowed", orgClusterName))
 	} else if decision != authorizer.DecisionAllow {
+		klog.Errorf("user %q lacks (%s) clusterworkspaces/workspace %q permission for %q in %s: %s", userInfo.GetName(), decisions[decision], "delete", internalName, orgClusterName, reason)
+		return nil, false, kerrors.NewForbidden(tenancyv1beta1.Resource("workspaces"), internalName, fmt.Errorf("deletion in workspace %q is not allowed", orgClusterName))
```
ncdc, line 690: I wonder if this should be `name` instead of `internalName`?

ncdc (approved):

ncdc:
Just 1 minor question, but otherwise 👍. Thanks for doing this!

ncdc:
> Given that, would it make sense to just remove the subcommand altogether for now?

I think it's ok to leave it for the time being, in case anyone was using it and isn't aware of the change.

pkg/virtual/workspaces/registry/rest.go
kylape: If you look up few lines, you'll see that there is still code in here around the internal/pretty name:

https://github.com/kcp-dev/kcp/blob/78dbe8e2d3b3ab897f67330918e161dba4dc798c/pkg/virtual/workspaces/registry/rest.go#L656-L667

So I think we are getting rid of that, but that would likely be a separate PR to clean it up.

kylape: Looked at this some more.  The `NotFound` error was returned to the user, so it should use `name`.  the `klog.Errorf` is logging (not to the user obviously) about the `clusterworkspace`, which should reference the `internalName`.

PR #887, opened 2022-04-13
Add branch name to the tag list of kcp and syncer images

fgiloux:

## Summary

By adding branch names to the tags of the kcp and syncer images it is easier to consume them from other projects

### DIFF
--- a/.github/workflows/kcp-image.yaml
+++ b/.github/workflows/kcp-image.yaml
@@ -30,7 +30,7 @@ jobs:
         uses: redhat-actions/buildah-build@v2
         with:
           image: kcp
-          tags: latest ${{ steps.vars.outputs.sha_short }}
+          tags: latest ${{ steps.vars.outputs.sha_short }} ${{ github.ref_name }}
           containerfiles: |
             ./Dockerfile
 
@@ -39,7 +39,7 @@ jobs:
         uses: redhat-actions/push-to-registry@v2
         with:
           image: ${{ steps.build-image.outputs.image }}
-          tags: ${{ steps.build-image.outputs.tags }}
+          tags: ${{ steps.build-image.outputs.tags }} ${{ github.ref_name }}
           registry: ghcr.io/${{ github.repository_owner }}
           username: ${{ github.actor }}
           password: ${{ github.token }}
--- a/.github/workflows/kcp-image.yaml
+++ b/.github/workflows/kcp-image.yaml
@@ -8,6 +8,8 @@ on:
     branches:
       - main
       - 'release-*'
+    tags:
+    - 'v*'
 
 jobs:
   build:
--- a/.github/workflows/syncer-image.yaml
+++ b/.github/workflows/syncer-image.yaml
@@ -19,6 +19,6 @@ jobs:
         with:
           go-version: v1.17
 
-      # Build and push a syncer image, tagged with the commit SHA.
+      # Build and push a syncer image, tagged with the commit SHA and the branch name.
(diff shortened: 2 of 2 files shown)


.github/workflows/kcp-image.yaml
```diff
@@ -25,13 +25,18 @@ jobs:
         id: vars
         run: echo "::set-output name=sha_short::$(echo ${{ github.sha }} | cut -b -7)"
 
+      - name: Extract branch name
+        shell: bash
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
```
ncdc, line 30: We should probably be consistent in our set-output commands? i.e. `::` instead of `##`?

fgiloux: I just saw that ${{ github.ref_name }} should now provide the branch name.
Regarding consistency I was wondering why kcp-image,yaml uses buildah  and syncer-image.yaml uses ko.

ncdc: > Regarding consistency I was wondering why kcp-image,yaml uses buildah and syncer-image.yaml uses ko.

Let's use a separate issue for that, please. Mind filing one for a discussion?

fgiloux: > Let's use a separate issue for that, please. Mind filing one for a discussion?
For the records, here is [the link](https://github.com/kcp-dev/kcp/issues/902) to the issue

fgiloux:
Any concerns with the idea or the approach?

.github/workflows/syncer-image.yaml
```diff
@@ -19,6 +19,6 @@ jobs:
         with:
           go-version: v1.17
 
-      # Build and push a syncer image, tagged with the commit SHA.
+      # Build and push a syncer image, tagged with the commit SHA and the branch name.
       - uses: imjasonh/setup-ko@v0.4
-      - run: ko publish ./cmd/syncer -t $(git rev-parse --short "$GITHUB_SHA")
+      - run: ko publish ./cmd/syncer -t $(git rev-parse --short "$GITHUB_SHA"),${{ github.ref_name }}
```
sttts, line 24: how will the syncer image be tagged? Does this mean here that there won't be a git tag on the syncer image?

fgiloux: it should have two tags:
- git rev-parse --short "$GITHUB_SHA", which is the short version of the commit
- github.ref_name, the branch name
Example: https://github.com/fgiloux/kcp/pkgs/container/kcp%2Fsyncer-c2e3073d5026a8f7f2c47a50c16bdbec
It has b952d3d and release-0.42

sttts: how can we get a tagged one too?

fgiloux: Tags would need to be added to the event list:
~~~
on:
  push:
    branches:
    - main
    - 'release-*'
    tags:
    - 'v*'
~~~

I can add it to this PR or open a separate one if you prefer. Would you want the action to be triggered on every tag or on tags having a specific pattern v* above?

sttts: `v*`

sttts: Just add it here I would say.

fgiloux: I have pushed the change here.

sttts (approved):

sttts:
:ship:

PR #894, opened 2022-04-14
Remove the global config embedding.

jpeach:
## Summary

Each level of config embeds it's YAML files independently now, so the
global embed of the directory is no longer used.

Signed-off-by: James Peach <jpeach@cloudflare.com>


ncdc (approved):

ncdc:
Welcome to the project, and thanks!

PR #895, opened 2022-04-14
Remove stty dependency.

jpeach:
## Summary

There is already internal code to detect the terminal width, so we don't
need to carry a dependency that does it by exec'ing stty.

Signed-off-by: James Peach <jpeach@cloudflare.com>


ncdc (approved):

ncdc:
@jpeach please run `make imports`, amend your commit, and force push. Thanks!

PR #882, opened 2022-04-12
Disable deploy pipeline on release branch

kylape:
No longer want to deploy to the "stable" environment from the release branch.

PR #900, opened 2022-04-14
[0.3] Disable deploy pipeline on release branch

kylape:
Backports #882.  The intent is for the image to still be built on this branch, but the deploy step is disabled.

### DIFF
--- a/.github/workflows/kcp-image.yaml
+++ b/.github/workflows/kcp-image.yaml
@@ -44,33 +44,3 @@ jobs:
           registry: ghcr.io/${{ github.repository_owner }}
           username: ${{ github.actor }}
           password: ${{ github.token }}
-
-  deploy:
-    if: "github.repository_owner == 'kcp-dev' && github.ref_name == 'main'"
-    name: Deploy KCP
-    needs: build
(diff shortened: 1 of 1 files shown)


.github/workflows/kcp-image.yaml
```diff
@@ -45,6 +45,17 @@ jobs:
...
+  deploy:
+    if: "github.repository_owner == 'kcp-dev' && github.ref_name == 'main'"
+    name: Deploy KCP
+    needs: build
+    environment: unstable
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get the short sha
+        id: vars
+        run: echo "::set-output name=sha_short::$(echo ${{ github.sha }} | cut -b -7)"
```
ncdc, line 57: Given this is on the `release-0.3` branch, we probably don't need it here?

kylape: Line 49 disables this for all branches except main.  but we could just delete this whole job from this branch.  I can go either way.

ncdc: Let's delete it from the branch

PR #908, opened 2022-04-18
Ensure the kcp config root has reasonable permissions.

### LINKED ISSUE #907: kcp creates a config directory with permissions that it doesn't like
**Describe the bug**

Running kcp emits this error:
```
{"level":"warn","ts":1650247225.6866188,"caller":"fileutil/fileutil.go:57","msg":"check file permission","error":"directory \"/Users/jpeach/upstream/kcp/.kcp\" exist, but the permission is \"drwxr-xr-x\". The recommended permission is \"-rwx------\" to prevent possible unprivileged access to the data"}
```

But kcp created this directory using `os.MkdirAll(dir, 0755)`, so it should know what it wants and just do that.

**To Reproduce**

Run kcp.

**Expected behavior**

No error.

jpeach:
## Summary

Some of the components that kcp starts write private configuration into
the configuration root before the kcp server comes up. This means that
to create the configuration directory with the permissions we want,
we need to do it really early so we aren't affected by these side-effects.

For the embedded etcd specifically, we should use the etcd "TouchDirAll"
API that is specifically provided for creating the configuration root.
It's also useful to make the etcd state persist in a subdirectory to
avoid any possibility of file name collisions.

Adding the new "MkdirRoot" utility to the "cmd" package isn't perhaps the
greatest choice, but circular import problems mean that it can't go in
"pkg/server", and there is some reasonable probability that multiple
commands could use it.

## Related issue(s)

Fixes #907


pkg/cmd/dir.go
```diff
@@ -0,0 +1,57 @@
...
+// server. This has to be done early before we start bringing up server
+// components to ensure that we set the initial permissions correctly,
+// since otherwise components will create it as a side-effect.
+func MkdirRoot(dir string) error {
+	if dir == "" {
+		return errors.New("missing root directory configuration")
+	}
+
+	fi, err := os.Stat(dir)
+	if err != nil {
```
stevekuznetsov, line 37: Check for the appropriate error here and return if it's unexpected, eg

```go
if errors.Is(err, os.NotExist)
```

pkg/server/server.go
```diff
@@ -42,7 +41,7 @@ import (
 	"k8s.io/kubernetes/pkg/genericcontrolplane"
 
 	configroot "github.com/kcp-dev/kcp/config/root"
-	"github.com/kcp-dev/kcp/config/system-crds"
+	systemcrds "github.com/kcp-dev/kcp/config/system-crds"
```
stevekuznetsov, line 44: Is this a no-op?

pkg/cmd/dir.go
jpeach: That happens below on line 39. I will update to move the log message below there.

pkg/cmd/dir.go
```diff
@@ -0,0 +1,57 @@
...
+	"os"
+
+	"k8s.io/klog/v2"
+)
+
+// MkdirRoot creates the root configuration directory for the KCP
+// server. This has to be done early before we start bringing up server
+// components to ensure that we set the initial permissions correctly,
+// since otherwise components will create it as a side-effect.
+func MkdirRoot(dir string) error {
```
sttts, line 31: does this have to be a public func? I only ee it used once.

pkg/cmd/dir.go
```diff
@@ -0,0 +1,57 @@
...
+	if dir == "" {
+		return errors.New("missing root directory configuration")
+	}
+
+	fi, err := os.Stat(dir)
+	if err != nil {
+		klog.Infof("Creating root directory %s", dir)
```
sttts, line 38: logging in a generic helper like this is odd.

pkg/server/options/options.go
```diff
@@ -181,6 +182,12 @@ func (o *Options) Complete() (*CompletedOptions, error) {
 		}
 		o.Extra.RootDirectory = filepath.Join(pwd, o.Extra.RootDirectory)
 	}
+
+	// Create the configuration root correctly before other components get a chance.
+	if err := cmd.MkdirRoot(o.Extra.RootDirectory); err != nil {
+		return nil, err
+	}
+
```
sttts, line 190: Is there a technical reason to have this side-effect in a completion func? I would expect this to happen in the Run code path.

jpeach: Yeh it's surprising, but the `.kcp` directory gets created as a side-effect in `o.Controllers.Complete()`

pkg/cmd/dir.go
jpeach: I found that the config path was a bit non-obvious, so clearly logging the persistent state directory was helpful.

pkg/server/server.go
jpeach: Yeh, Goland seemed to think it was a good idea 🤷

pkg/cmd/dir.go
jpeach: I guess not. It's public to make it a more official policy rather than just something this server does.

pkg/server/options/options.go
sttts: Why don't we fix that code then? I see `if err := keyutil.WriteKey(c.SAController.ServiceAccountKeyFile, encoded); err != nil` there. `WriteKey` is only mkdir+write. Just replace that with a call to a `0700` mkdir and `ioutil.WriteFile`.

pkg/cmd/dir.go
sttts: With https://github.com/kcp-dev/kcp/pull/908#discussion_r853852591, can we get rid of `MkdirRoot`?

pkg/server/options/options.go
jpeach: We could do that, but that still results in the root config dir being created as a side effect. It's better to create it intentionally up front rather than depending on this spooky action at a distance.

sttts: I can see that if we move MkdirRoot into pkg/server and make it private. Then I would also be fine with the logging output. Am not a fan of helpers that are not used and probably will never be used by other code.

jpeach: Sounds good!

jpeach:
@sttts @stevekuznetsov Updated with review changes.

sttts (approved):

sttts:
Lgtm. Ship it 🚢

ncdc:
@jpeach are you able to rebase? We'll get it merged after!

ncdc (approved):

PR #909, opened 2022-04-18
e2e: Enable testing against a deployed syncer with kind

marun:
- [x] Ensure that syncer deployment and deployed syncer operation can be tested against kind 

As part of this PR, the `e2e-shared-server` job needs to changed to required to ensure against regression of syncer deployment.


marun:
I won't be able to implement testing of kcp api access from synced deployments in this PR. I've filed #917 and assigned to P4 with `priority/urgent` to ensure this gets tackled asap.

docs/syncer.md
```diff
@@ -11,77 +11,79 @@ users have to install a special component named [syncer](https://github.com/kcp-
...
 NAME     READY   UP-TO-DATE   AVAILABLE   AGE
-syncer   1/1     1            1           13m
-```
\ No newline at end of file
+kcp-syncer   1/1     1            1           13m
+```
+
+1. Wait for the kcp workload cluster to go ready.
+
+TODO(marun)
```
ncdc, line 89: Is this for this PR or a follow up?

docs/syncer.md
```diff
@@ -11,77 +11,79 @@ users have to install a special component named [syncer](https://github.com/kcp-
...
-```sh
-$ kubectl get secret
-NAME                  TYPE                                  DATA   AGE
-default-token-nkpxj   kubernetes.io/service-account-token   3      29m
-syncer-token-2mtgf    kubernetes.io/service-account-token   3      29m
+Alternately, create a `kind` cluster with a local registry to simplify syncer development by executing the
+following script:
 
-```
+https://raw.githubusercontent.com/kubernetes-sigs/kind/main/site/static/examples/kind-with-registry.sh
```
sttts, line 45: what is special about that script? Why is the first way not enough?

marun: The script deploys kind with a registry, which is ideal for building the image locally and testing it without having to push it to a regsitry.

test/e2e/framework/fixture.go
```diff
@@ -271,89 +271,187 @@ func NewWorkspaceWithWorkloads(t *testing.T, server RunningServer, orgClusterNam
...
+	KubectlApply(t, downstreamKubeconfigPath, syncerYAML)
 
 	// Extract the configuration for an in-process syncer from the resources that were
 	// applied to the downstream server. This maximizes the parity between the
 	// configuration of a deployed and in-process syncer.
-	syncerNamespace := workloadcliplugin.GetSyncerID(cfg.WorkspaceClusterName.String(), cfg.WorkloadClusterName)
+	syncerNamespace := workloadcliplugin.GetSyncerID(sf.WorkspaceClusterName.String(), sf.WorkloadClusterName)
 	syncerConfig := syncerConfigFromCluster(t, downstreamConfig, syncerNamespace)
 
+	ctx, cancelFunc := context.WithCancel(context.Background())
```
ncdc, line 371: Should we pass ctx in as an arg to Start()?

test/e2e/framework/fixture.go
```diff
@@ -271,89 +271,187 @@ func NewWorkspaceWithWorkloads(t *testing.T, server RunningServer, orgClusterNam
...
 	}
+
+	// The workload cluster becoming ready indicates the syncer is healthy and has
+	// successfully sent a heartbeat to kcp.
+	startedSyncer.WaitForClusterReadyReason(t, ctx, "")
+
+	return startedSyncer
+}
+
+// StartedSyncerFixture contains the configuration used to start a syncer and interact with its
```
ncdc, line 442: Are we still using this to start a syncer?

pkg/syncer/syncer.go
```diff
@@ -424,6 +428,18 @@ type NamespaceLocator struct {
...
+	annotation := annotations[namespaceLocatorAnnotation]
+	if len(annotation) == 0 {
+		return nil, nil
+	}
+	var locator NamespaceLocator
+	if err := json.Unmarshal([]byte(annotation), &locator); err != nil {
+		return nil, err
+	}
+	return &locator, nil
+}
```
sttts, line 441: this annotation and it format are preexisting?

sttts: ic, was inlined

test/e2e/reconciler/cluster/syncerheartbeat_test.go
```diff
@@ -1,49 +0,0 @@
...
-
-import (
-	"context"
-	"testing"
-
-	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
-	"github.com/kcp-dev/kcp/test/e2e/framework"
-)
-
-func TestSyncerHeartbeat(t *testing.T) {
```
sttts, line 27: why is this gone? Was useless?

ncdc: See https://github.com/kcp-dev/kcp/pull/909/files#diff-91621de609392e23db5fc8ea76095bd9e1fd7cfefc8d496188205888f316210eR55

marun: Yeah, it's effectively subsumed by the new syncer test. Also, the syncer fixture is checking for a ready cluster which is implicitly validating the heartbeat behavior.

test/e2e/framework/fixture.go
marun: 'Started' is the key word here - it results from starting the fixture. If you don't like it, easy enough to change, but I thought it was useful to allow configuring the fixture from a struct (to allow not specifying optional things) and then start it subsequently.

test/e2e/framework/fixture.go
marun: I don't think there's a need to manage context for most syncer, it can be an internal detail of the fixture. It's not like the use of context in a request chain, the only requirement is that the fixture gets cleaned up at the end of the test so it's not necessary to pass a single context around everywhere.

docs/syncer.md
marun: Follow-up. I typically think of FIXME as something to finish in the current PR and TODO as 'eventually'.

ncdc (approved):

Issue #527, opened 2022-02-16
Workspace deletion


ncdc:
**Is your feature request related to a problem? Please describe.**
When deleting a workspace, we need to clean up all its associated content automatically.

**Describe the solution you'd like**
This would include all cluster and namespace scoped resources in the workspace. We may decide to use the upstream kube namespace controller for part of this, or design our own solution.

ncdc:
Related: #400 (may need to adjust its description)

PR #913, opened 2022-04-19
delete resources in workspace when workspace is deleted

qiujian16:

## Summary
basically reuse the namespace deletion code to delete all resources in a workspace when workspace is deleting. Still need to add tests

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/527

### DIFF
--- a/pkg/admission/namespacelifecycle/admission.go
+++ b/pkg/admission/namespacelifecycle/admission.go
@@ -1,0 +1,164 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package namespacelifecycle
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apiserver/pkg/admission"
+	"k8s.io/apiserver/pkg/admission/initializer"
+	"k8s.io/apiserver/pkg/admission/plugin/namespace/lifecycle"
+	genericapirequest "k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clusters"
+
+	kcpinitializers "github.com/kcp-dev/kcp/pkg/admission/initializers"
+	kcpinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
+	tenancyv1alpha1lister "github.com/kcp-dev/kcp/pkg/client/listers/tenancy/v1alpha1"
+)
+
+const (
+	// PluginName indicates the name of admission plug-in
+	PluginName = "WorkspaceNamespaceLifecycle"
+)
+
+// Register registers a plugin
+func Register(plugins *admission.Plugins) {
+	plugins.Register(PluginName, func(config io.Reader) (admission.Interface, error) {
+		return newLifcycle()
+	})
+}
+
+// workspaceNamespaceLifecycle is a wrapper of kubernetes namespace lifecycle admission control
+// It uses legacy namespace lifecycle admission control by default, and ignores
+// immortal namespaces when the workspaces is deleting. This can ensure we can remove all immortal
+// namespaces when workspaces is deleting.
+type workspaceNamespaceLifecycle struct {
+	// legacyNamespaceLifecycle is the kube legacy namespace lifecycle
+	legacyNamespaceLifecycle *lifecycle.Lifecycle
+
+	// namespaceLifecycle is used only when workspace is deleting
+	namespaceLifecycle *lifecycle.Lifecycle
+
+	*admission.Handler
+	workspaceLister tenancyv1alpha1lister.ClusterWorkspaceLister
+}
+
+func newLifcycle() (*workspaceNamespaceLifecycle, error) {
+	legacyLifecycle, err := lifecycle.NewLifecycle(sets.NewString(metav1.NamespaceDefault, metav1.NamespaceSystem, metav1.NamespacePublic))
+	if err != nil {
+		return nil, err
+	}
+
+	lifecycle, err := lifecycle.NewLifecycle(sets.NewString())
+	if err != nil {
+		return nil, err
+	}
+
+	return &workspaceNamespaceLifecycle{
+		Handler:                  admission.NewHandler(admission.Create, admission.Update, admission.Delete),
+		legacyNamespaceLifecycle: legacyLifecycle,
+		namespaceLifecycle:       lifecycle,
+	}, nil
+}
+
+var _ = kcpinitializers.WantsKcpInformers(&workspaceNamespaceLifecycle{})
+var _ = initializer.WantsExternalKubeInformerFactory(&workspaceNamespaceLifecycle{})
+var _ = initializer.WantsExternalKubeClientSet(&workspaceNamespaceLifecycle{})
+
+// Admit makes an admission decision based on the request attributes
+func (l *workspaceNamespaceLifecycle) Admit(ctx context.Context, a admission.Attributes, o admission.ObjectInterfaces) error {
+
+	// call legacy namespace lifecycle at first
+	admissionErr := l.legacyNamespaceLifecycle.Admit(ctx, a, o)
+
+	// when the workspace is being deleted, we want to allow deletion of all namespaces. Hence, after
+	// legacyNamespaceLifecycle has forbidden the deletion, we give it a second chance be delegating to
+	// namespaceLifecycle which does not protect any namespace beyond normal life-cycle.
+	if !apierrors.IsForbidden(admissionErr) {
+		return admissionErr
+	}
+
+	if a.GetOperation() != admission.Delete || a.GetKind().GroupKind() != corev1.SchemeGroupVersion.WithKind("Namespace").GroupKind() {
+		return admissionErr
+	}
+
+	clusterName, err := genericapirequest.ClusterNameFrom(ctx)
+	if err != nil {
+		return apierrors.NewInternalError(err)
+	}
+
+	org, hasParent := clusterName.Parent()
+	if !hasParent {
+		return admissionErr
+	}
+
+	workspaceKey := clusters.ToClusterAwareKey(org, clusterName.Base())
+
+	workspace, err := l.workspaceLister.Get(workspaceKey)
+	// The shard hosting the workspace could be down,
+	// just return error from legacy namespace lifecycle admission in this case
+	if err != nil && !apierrors.IsNotFound(err) {
+		return admissionErr
+	}
+
+	if workspace.DeletionTimestamp.IsZero() {
+		return admissionErr
+	}
+
+	return l.namespaceLifecycle.Admit(ctx, a, o)
+}
+
+// SetExternalKubeInformerFactory implements the WantsExternalKubeInformerFactory interface.
+func (l *workspaceNamespaceLifecycle) SetExternalKubeInformerFactory(f informers.SharedInformerFactory) {
+	l.legacyNamespaceLifecycle.SetExternalKubeInformerFactory(f)
+	l.namespaceLifecycle.SetExternalKubeInformerFactory(f)
+}
+
+// SetExternalKubeClientSet implements the WantsExternalKubeClientSet interface.
+func (l *workspaceNamespaceLifecycle) SetExternalKubeClientSet(client kubernetes.Interface) {
+	l.legacyNamespaceLifecycle.SetExternalKubeClientSet(client)
+	l.namespaceLifecycle.SetExternalKubeClientSet(client)
+}
+
+func (l *workspaceNamespaceLifecycle) SetKcpInformers(informers kcpinformers.SharedInformerFactory) {
+	l.SetReadyFunc(informers.Tenancy().V1alpha1().ClusterWorkspaces().Informer().HasSynced)
+	l.workspaceLister = informers.Tenancy().V1alpha1().ClusterWorkspaces().Lister()
+}
+
+// ValidateInitialization implements the InitializationValidator interface.
+func (l *workspaceNamespaceLifecycle) ValidateInitialization() error {
+	if err := l.legacyNamespaceLifecycle.ValidateInitialization(); err != nil {
+		return err
+	}
+
+	if err := l.namespaceLifecycle.ValidateInitialization(); err != nil {
+		return err
+	}
+
+	if l.workspaceLister == nil {
+		return fmt.Errorf("missing workspaceLister")
+	}
+	return nil
+}
--- a/pkg/admission/plugins.go
+++ b/pkg/admission/plugins.go
@@ -45,10 +45,12 @@ import (
 	"github.com/kcp-dev/kcp/pkg/admission/clusterworkspaceshard"
 	"github.com/kcp-dev/kcp/pkg/admission/clusterworkspacetype"
 	"github.com/kcp-dev/kcp/pkg/admission/clusterworkspacetypeexists"
+	workspacenamespacelifecycle "github.com/kcp-dev/kcp/pkg/admission/namespacelifecycle"
 )
 
 // AllOrderedPlugins is the list of all the plugins in order.
 var AllOrderedPlugins = beforeWebhooks(kubeapiserveroptions.AllOrderedPlugins,
+	workspacenamespacelifecycle.PluginName,
 	apiresourceschema.PluginName,
 	clusterworkspace.PluginName,
 	clusterworkspaceshard.PluginName,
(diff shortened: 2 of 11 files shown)


pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_gc_controller.go
```diff
@@ -0,0 +1,175 @@
...
+		UpdateFunc: func(_, obj interface{}) { c.enqueue(obj) },
+	})
+
+	c.deleter = deletion.NewWorkspacedResourcesDeleter(kcpClient, metadataClient, discoverResourcesFn, "ws-finalize")
+	c.workspaceSynced = workspaceInformer.Informer().HasSynced
+
+	return c
+}
+
+type GCController struct {
```
sttts, line 64: this is not really GC, but just namespace and object creation, right? Would just call it `clusterworkspacedeletion` as a package. 

Also, don't mix it with the existing clusterworkspace controller. Deletion is separate IMO.

pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_gc_controller.go
```diff
@@ -0,0 +1,175 @@
...
+				hasFinalizer = true
+				break
+			}
+		}
+		if !hasFinalizer {
+			workspace.Finalizers = append(workspace.Finalizers, deletion.WorkspaceFinalizer)
+			_, err := c.kcpClient.Cluster(logicalcluster.From(workspace)).TenancyV1alpha1().ClusterWorkspaces().Update(ctx, workspace, metav1.UpdateOptions{})
+			return err
+		}
+	}
```
sttts, line 172: return? You don't want to run into the deletion below, do you?

pkg/reconciler/tenancy/clusterworkspace/deletion/status_condition_util.go
```diff
@@ -0,0 +1,233 @@
...
+)
+
+// NamespaceConditionUpdater interface that translates namespace deleter errors
+// into namespace status conditions.
+type WorkspaceConditionUpdater interface {
+	ProcessDiscoverResourcesErr(e error)
+	ProcessGroupVersionErr(e error)
+	ProcessDeleteContentErr(e error)
+	Update(*tenancyv1alpha1.ClusterWorkspace) bool
+}
```
sttts, line 38: can you follow the patterns for updating conditions in pkg/reconciler/pis/apibindings? This code here feels different in style.

pkg/reconciler/tenancy/clusterworkspace/deletion/workspace_resource_deletor.go
```diff
@@ -0,0 +1,536 @@
+/*
+Copyright 2021 The KCP Authors.
```
sttts, line 2: 2022

pkg/reconciler/tenancy/clusterworkspace/deletion/workspace_resource_deletor.go
```diff
@@ -0,0 +1,536 @@
+/*
```
sttts, line 1: how much of this file is from namespace deletion?

pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_gc_controller.go
```diff
@@ -0,0 +1,175 @@
+/*
+Copyright 2021 The KCP Authors.
```
sttts, line 2: 2022

pkg/reconciler/tenancy/clusterworkspace/deletion/workspace_resource_deletor.go
```diff
@@ -0,0 +1,536 @@
...
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	genericapirequest "k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/metadata"
+	"k8s.io/client-go/util/retry"
+)
+
+const WorkspaceFinalizer = "workspace-finalizer"
```
sttts, line 41: Would rather use something with `tenancy.kcp.dev`.

pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_gc_controller.go
```diff
@@ -0,0 +1,175 @@
...
+		klog.Infof("Workspace has been deleted %v", key)
+		return nil
+	}
+	if err != nil {
+		runtime.HandleError(fmt.Errorf("unable to retrieve workspace %v from store: %v", key, err))
+		return err
+	}
+
+	workspace = workspace.DeepCopy()
+	hasFinalizer := false
```
sttts, line 159: move into the if clause

pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_gc_controller.go
qiujian16: renamed

pkg/reconciler/tenancy/clusterworkspace/deletion/status_condition_util.go
qiujian16: updated

pkg/reconciler/tenancy/clusterworkspace/deletion/workspace_resource_deletor.go
qiujian16: I think 70%. I remove the opCache part, and change how conditions are updated. Add a comment, and also add kubernetes copyright in this file.

qiujian16:
@sttts PTAL on this PR again. Thanks!

pkg/admission/namespacelifecycle/admission.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	org, hasParent := clusterName.Parent()
+	if !hasParent {
+		return l.legacyNamespaceLifecycle.Admit(ctx, a, o)
+	}
+
+	workspaceKey := clusters.ToClusterAwareKey(org, clusterName.Base())
+
+	workspace, err := l.workspaceLister.Get(workspaceKey)
```
sttts, line 108: this won't scale. I would rather prefer some way to decide locally, e.g. by looking at the request user, to decide whether we want to allow deletion.

Background: as soon as we get sharding you won't have a `workspaceLister` anymore that holds all workspaces, only those on the same shard.

pkg/admission/namespacelifecycle/admission.go
```diff
@@ -0,0 +1,148 @@
...
+// ValidateInitialization implements the InitializationValidator interface.
+func (l *workspaceNamespaceLifecycle) ValidateInitialization() error {
+	if err := l.legacyNamespaceLifecycle.ValidateInitialization(); err != nil {
+		return err
+	}
+
+	if err := l.namespaceLifecycle.ValidateInitialization(); err != nil {
+		return err
+	}
+	return nil
```
sttts, line 147: no check for the kcp informer?

pkg/admission/plugins.go
```diff
@@ -45,10 +45,12 @@ import (
 	"github.com/kcp-dev/kcp/pkg/admission/clusterworkspaceshard"
 	"github.com/kcp-dev/kcp/pkg/admission/clusterworkspacetype"
 	"github.com/kcp-dev/kcp/pkg/admission/clusterworkspacetypeexists"
+	"github.com/kcp-dev/kcp/pkg/admission/namespacelifecycle"
```
sttts, line 48: maybe alias as `kcpnamespacelifecycle`, to make clear it is our version

pkg/admission/namespacelifecycle/admission.go
sttts: Or let's say this admission plugin must not fail normal namespace deletion when the shard with the workspace object is down. During workspace deletion though, failing in that case is ok.

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -190,6 +190,23 @@ const (
...
+	WorkspaceDeletionGVParsingSuccess conditionsv1alpha1.ConditionType = "WorkspaceDeletionGVParsingSuccess"
+
+	// WorkspaceDeletionContentSuccess represents the status that all resources in the workspace is deleting
+	WorkspaceDeletionContentSuccess conditionsv1alpha1.ConditionType = "WorkspaceDeletionContentSuccess"
+
+	// WorkspaceContentClean represents the status that all resources in the workspace is cleaned.
+	WorkspaceContentClean conditionsv1alpha1.ConditionType = "WorkspaceContentClean"
+
+	// WorkspaceFinalizersClean represents the status that finalizers of all resources in the workspace is cleaned.
+	WorkspaceFinalizersClean conditionsv1alpha1.ConditionType = "WorkspaceFinalizersClean"
```
sttts, line 209: these are many conditions. Can we combine them into one, but with many reasons?

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
```diff
@@ -0,0 +1,245 @@
...
+	klog.Infof("Queueing workspace %q", key)
+	c.queue.Add(key)
+}
+
+func (c *Controller) Start(ctx context.Context, numThreads int) {
+	defer runtime.HandleCrash()
+	defer c.queue.ShutDown()
+
+	klog.Info("Starting ClusterWorkspace GC controller")
+	defer klog.Info("Shutting down ClusterWorkspace GC controller")
```
sttts, line 93: update controller name

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
```diff
@@ -0,0 +1,245 @@
...
+	copiedFinalizers := []string{}
+	for i := range workspace.Finalizers {
+		if workspace.Finalizers[i] == deletion.WorkspaceFinalizer {
+			continue
+		}
+		copiedFinalizers = append(copiedFinalizers, workspace.Finalizers[i])
+	}
+	if len(workspace.Finalizers) != len(copiedFinalizers) {
+		workspace.Finalizers = copiedFinalizers
+		_, err := c.kcpClient.Cluster(logicalcluster.From(workspace)).TenancyV1alpha1().ClusterWorkspaces().Update(ctx, workspace, metav1.UpdateOptions{})
```
sttts, line 241: patch?

pkg/admission/namespacelifecycle/admission.go
qiujian16: good point! updated.

pkg/apis/tenancy/v1alpha1/types.go
qiujian16: reduced to two conditions WorkspaceContentClean and WorkspaceContentDeleteSuccess

test/e2e/framework/expect.go
```diff
@@ -51,7 +51,7 @@ type Expecter interface {
 
 // NewExpecter creates a informer-driven registry of expectations, which will
 // be triggered on every event that the informer ingests.
-func NewExpecter(informer cache.SharedIndexInformer) *expectationController {
+func NewExpecter(informer cache.SharedIndexInformer) Expecter {
```
ncdc, line 54: I'm curious why you made this change?

qiujian16: I think this should return an interface but not a private class? I need to call ExpectBefore method of this interface https://github.com/kcp-dev/kcp/pull/913/files#diff-d37089ee294b6de484fc1d0d69e3ce8357b20367d45f1336df40e23a2461da7bR67

ncdc: See https://github.com/golang/go/wiki/CodeReviewComments#interfaces - it's better to return the concrete type

qiujian16: gotcha, updated :)

pkg/admission/namespacelifecycle/admission.go
```diff
@@ -0,0 +1,161 @@
...
+func (l *workspaceNamespaceLifecycle) Admit(ctx context.Context, a admission.Attributes, o admission.ObjectInterfaces) error {
+
+	// call legacy namespace lifecycle at first
+	admissionErr := l.legacyNamespaceLifecycle.Admit(ctx, a, o)
+
+	// Only handle the case when namespace deletion is forbidden
+	if !apierrors.IsForbidden(admissionErr) {
+		return admissionErr
+	}
+
```
sttts, line 99:
```suggestion
// when the workspace is being deleted, we want to allow deletion of all workspaces. Hence, after 
// legacyNamespaceLifecycle has forbidden the deletion, we give it a second chance be delegating to
// namespaceLifecycle which does not protect any namespace beyond normal life-cycle.
```

pkg/admission/namespacelifecycle/admission.go
```diff
@@ -0,0 +1,161 @@
...
+	if !hasParent {
+		return admissionErr
+	}
+
+	workspaceKey := clusters.ToClusterAwareKey(org, clusterName.Base())
+
+	workspace, err := l.workspaceLister.Get(workspaceKey)
+	// The shard hosting the workspace could done,
+	// just return error from legacy namespace lifecycle addmission in this case
+	if err != nil {
```
sttts, line 119: do we want to handle 404 differently? A 404 is a definitive response, i.e. the workspace is really gone. Maybe we want to keep deleting the namespace then?

pkg/admission/plugins.go
```diff
@@ -78,10 +80,11 @@ func RegisterAllKcpAdmissionPlugins(plugins *admission.Plugins) {
 	clusterworkspacetypeexists.Register(plugins)
 	apiresourceschema.Register(plugins)
 	apibinding.Register(plugins)
+	kcpnamespacelifecycle.Register(plugins)
 }
 
 var defaultOnPluginsInKcp = sets.NewString(
-	lifecycle.PluginName,              // NamespaceLifecycle
+	kcpnamespacelifecycle.PluginName,  // NamespaceLifecycle
```
sttts, line 87: WorkspaceNamespaceLifecycle

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -190,6 +190,12 @@ const (
 	// WorkspaceShardValidReasonShardNotFound reason in WorkspaceShardValid condition means that the
 	// referenced ClusterWorkspaceShard object got deleted.
 	WorkspaceShardValidReasonShardNotFound = "ShardNotFound"
+
+	// WorkspaceDeletionContentSuccess represents the status that all resources in the workspace is deleting
+	WorkspaceDeletionContentSuccess conditionsv1alpha1.ConditionType = "WorkspaceDeletionContentSuccess"
+
+	// WorkspaceContentClean represents the status that all resources in the workspace is cleaned.
+	WorkspaceContentClean conditionsv1alpha1.ConditionType = "WorkspaceContentClean"
```
sttts, line 198: Clean = Deleted?

test/e2e/reconciler/clusterworkspacedeletion/controller_test.go
```diff
@@ -0,0 +1,197 @@
...
+					if !conditions.IsTrue(workspace, tenancyv1alpha1.WorkspaceDeletionContentSuccess) {
+						return false, fmt.Errorf("expect true status for condition %s", tenancyv1alpha1.WorkspaceDeletionContentSuccess)
+					}
+
+					if !conditions.IsFalse(workspace, tenancyv1alpha1.WorkspaceContentClean) {
+						return false, fmt.Errorf("expect true status for condition %s", tenancyv1alpha1.WorkspaceContentClean)
+					}
+
+					return true, nil
+				}, 30*time.Second)
```
sttts, line 128: `wait.ForeverTestTimeout`

test/e2e/reconciler/clusterworkspacedeletion/controller_test.go
```diff
@@ -0,0 +1,197 @@
...
+
+				// finally check if all resources has been removed.
+				nslist, err := kubeClient.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+				require.NoError(t, err, "failed to list namespaces in workspace %s", workspace.Name)
+				require.Equal(t, 0, len(nslist.Items))
+
+				cmlist, err := kubeClient.CoreV1().ConfigMaps(metav1.NamespaceAll).List(ctx, metav1.ListOptions{})
+				require.NoError(t, err, "failed to list configmaps in workspace %s", workspace.Name)
+				require.Equal(t, 0, len(cmlist.Items))
+			},
```
sttts, line 155: would like to see another cluster-scoped resource in the test.

sttts: Also add a second test case where an org workspace is deleted. That has the obvious complexity of nested deletions, but it also has virtual resources (`Workspace` which is mapped into the org workspace, but is actually served by the virtual workspace apiserver).

sttts: cc @davidfestal – interesting topic when we map in virtual resources that are marked writeable in the fake CRD we put into the workspace, but readonly in the virtual workspace.

qiujian16: also create a ns in the workspace, ensuring it is deleted and another test to delete org workspace.

pkg/admission/namespacelifecycle/admission.go
```diff
@@ -92,7 +93,9 @@ func (l *workspaceNamespaceLifecycle) Admit(ctx context.Context, a admission.Att
 	// call legacy namespace lifecycle at first
 	admissionErr := l.legacyNamespaceLifecycle.Admit(ctx, a, o)
 
-	// Only handle the case when namespace deletion is forbidden
+	// when the workspace is being deleted, we want to allow deletion of all workspaces. Hence, after
```
sttts, line 96: "all namespaces". Maybe my fault in the suggestion :)

pkg/admission/plugins.go
```diff
@@ -80,17 +80,17 @@ func RegisterAllKcpAdmissionPlugins(plugins *admission.Plugins) {
...
 
 var defaultOnPluginsInKcp = sets.NewString(
-	kcpnamespacelifecycle.PluginName,  // NamespaceLifecycle
-	limitranger.PluginName,            // LimitRanger
-	mutatingwebhook.PluginName,        // MutatingAdmissionWebhook
-	validatingwebhook.PluginName,      // ValidatingAdmissionWebhook
-	certapproval.PluginName,           // CertificateApproval
-	certsigning.PluginName,            // CertificateSigning
-	certsubjectrestriction.PluginName, // CertificateSubjectRestriction
+	workspacenamespacelifecycle.PluginName, // NamespaceLifecycle
```
sttts, line 87: it's still called `NamespaceLifecycle`?

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
```diff
@@ -205,7 +209,7 @@ func (c *Controller) process(ctx context.Context, key string) error {
 }
 
 func (c *Controller) patchStatus(ctx context.Context, old, new *tenancyv1alpha1.ClusterWorkspace) error {
-	if !equality.Semantic.DeepEqual(old.Status.Conditions, new.Status.Conditions) {
+	if equality.Semantic.DeepEqual(old.Status.Conditions, new.Status.Conditions) {
```
sttts, line 212: with this it actually became `patchConditions`, no?

sttts (approved):

PR #930, opened 2022-04-25
cmd/kcp: add --feature-gates flag

sttts:
Used as in kube:
- on the command line: `--feature-gates=Foo=true`
- in code: `if utilfeature.DefaultFeatureGate.Enabled(kcpfeatures.Foo)`.
- In `kcp start options`: Bildschirmfoto 2022-04-25 um 14 07 13


davidfestal (approved):
Nice !

PR #829, opened 2022-04-04
apis: add scheduling.kcp.dev/v1alpha1 API

sttts:
Implementing https://github.com/kcp-dev/kcp/issues/912 

First attempt to add `Location`, ~~`LocationDomain`~~ (dropped) ~~, `Placement`~~ (postponed).

TODO:
- [x] split into bare minimum core API for p4, and extended "future" PR for p5+
- [x] rename LocationPool -> LocationDomain maybe
- [x] add locationdomain admission for validation
   - [x] unit tests
- [x] add locationdomain controller to create location objects in the org
   - [x] e2e test
   - [x] unit tests
- [x] add domain scheduler
   - [x] e2e test
   - [x] unit tests
- [x] add placement scheduler
   - [x] e2e test
   - [x] unit tests
- [x] add workload apiexport controller that create APIResourceSchemas and updates the export
   - [x] e2e test
   - [x] unit tests
- [x] add workload apibinding controller that create an APIBinding after domain scheduling. 
   - [x] e2e test
   - [x] unit tests

=== stretch goals from here ===

- [ ] add virtual workspace for location filtering and projecting into user workspaces.
- [x] come up with plan what this API means for sharding and availability.
- [ ] sketch WorkspacePlacement (vs. Placement per namespace)
- [ ] move assignment annotation to placement object status
- [ ] update placement dynamically if no workload object has been created yet

follow-ups:
1. ~~[ ] rolling updates of changes in the transformations~~ outside of this PR
2. ~~[ ] rethink transformation which should be in the virtual workspace and which in the syncer (like service account)~~ outside of this PR
3. [ ] lazy binding of the pool – am not sure how this can go if the pool means an APIImport.
4. [ ] make location controller generic to support non-WorkloadCluster GVRs (which means dynamic client informer).
5. [ ] add admission to forbid normal users to set placement annotations to specific WorkloadCluster. Maybe for that split location scheduling, and WorkloadCluster assignment (via uid).

PR #937, opened 2022-04-26
apis: allow APIResourceSchemas with single word groups used by kube

sttts:


shawn-hurley (commented):
Looks good to me, I wonder if a test should be added?

sttts:
Test will come through https://github.com/kcp-dev/kcp/pull/829 when deployments.apps are exported.

davidfestal (approved):

PR #942, opened 2022-04-28
Advanced scheduling

jmprusi:
This PRs adds the required changes for the advanced scheduling demo in milestone 0.4, and includes a set of changes that are feature gated and others that aren't.

The feature gate is added to the workloadcluster resources: `featuregates.experimental.workloads.kcp.dev/advancedscheduling: "true"`, the syncers should be started AFTER the annotation is set, or restarted manually to be taken into account.

Non-feature-gated changes:
* New label for defining the assigned workload clusters to a resource: `state.internal.workloads.kcp.dev/<workloadClusterName>` that replaces the previous, and now deprecated `workloads.kcp.dev/cluster: "<workloadClusterName>"`

Feature-gated changes:
* Syncer support for soft finalizers per workloadCluster: `finalizers.workloads.kcp.dev/"<workloadClusterName>": "finalizer1,finalizer2,..."`
* Syncer SpecDiff JsonPatch support per workloadCluster, the syncer will apply jsonPatches to the spec of an object: `experimental.spec-diff.workloads.kcp.dev/<workloadClusterName>: "'[{"op":"replace","path":"/replicas","value":3}]'"`
* Syncer support for per-workloadCluster deletion-timestamp annotations: `deletion.internal.workloads.kcp.dev/<workloadClusterName>: "2020-01-01T00:00:00Z"`
* Syncer will add a finalizer to the object: `workloads.kcp.dev/syncer-<workloadClusterName>` that will block deletion from upstream until the per-workspacecluster deletion timestamp.
* The syncer will push the resource status into an annotation per-workloadCluster: `experimental.status.workloads.kcp.dev/<workloadClusterName>`

How to test the features introduced in this PR: 

1. Start KCP
2. Create workspace
```
kubectl kcp ws create demo --enter
```

3. Create two kind clusters
4. Enable Syncer for both kind clusters (don't start those yet!)
```
kubectl kcp workload sync cluster1 --syncer-image a:a --replicas 0 --resources=ingresses.networking.k8s.io,deployments.apps,services | KUBECONFIG=cluster1.kubeconfig kubectl apply -f -

kubectl kcp workload sync cluster2 --syncer-image a:a --replicas 0 --resources=ingresses.networking.k8s.io,deployments.apps,services | KUBECONFIG=cluster2.kubeconfig kubectl apply -f -
```

5. Annotate the WorkloadClusters to enable the feature gate
```
kubectl annotate workloadcluster cluster2 featuregates.experimental.workloads.kcp.dev/advancedscheduling="true" --overwrite
kubectl annotate workloadcluster cluster1 featuregates.experimental.workloads.kcp.dev/advancedscheduling="true" --overwrite
```

6. Start both Syncer:
```
./bin/syncer --workload-cluster-name cluster1 --from-context system:admin --from-kubeconfig .kcp/admin.kubeconfig --from-cluster root:default:demo --to-kubeconfig cluster1.kubeconfig -r=ingresses.networking.k8s.io -r=deployments.apps -r=services

./bin/syncer --workload-cluster-name cluster2 --from-context system:admin --from-kubeconfig .kcp/admin.kubeconfig --from-cluster root:default:demo --to-kubeconfig cluster2.kubeconfig -r=ingresses.networking.k8s.io -r=deployments.apps -r=services
```

7. Disable namespace scheduler:
```
kubectl label namespace default experimental.workloads.kcp.dev/scheduling-disabled="true"
``` 
8. Now deploy the ingress-demo example yaml file:
```
kubectl apply -f contrib/examples/ingress-demo/full-example.yaml
```

9. Annotate the namespace and other resources:
```
kubectl label secret $(kubectl get secrets -o=jsonpath='{.items[?(@..annotations.kubernetes\.io/service-account\.name=="default")].metadata.name}') "state.internal.workloads.kcp.devcluster2"="Sync" "state.internal.workloads.kcp.dev/cluster1"="Sync" --overwrite

kubectl label configmap kube-root-ca.crt  "state.internal.workloads.kcp.dev/cluster2"="Sync" "state.internal.workloads.kcp.dev/cluster1"="Sync" --overwrite

kubectl label namespace default "state.internal.workloads.kcp.dev/cluster2"="Sync" "state.internal.workloads.kcp.dev/cluster1"="Sync" --overwrite
```

10. Now annotate the Deployment with an empty WorkloadClusterLabel:
```
kubectl label deployment httpecho  "state.internal.workloads.kcp.dev/cluster2"="" "state.internal.workloads.kcp.dev/cluster1"="" --overwrite
```

11. The deployment will not be scheduled yet. Let's add an SpecDiff to it so the Deployment will have different replicas in the different locations, `3` replicas in the `cluster1`, and `1` replica in the `cluster2`, this is done by setting the proper annotation with the desired JsonPatch value. NOTE: The "root" path of the json object is already the "spec"
```
kubectl annotate deployment httpecho experimental.specdiff.workloads.kcp.dev/cluster1='[{"op":"replace","path":"/replicas","value":3}]' experimental.specdiff.workloads.kcp.dev/cluster2='[{"op":"replace","path":"/replicas","value":1}]' --overwrite
```

12. Now let's sync the deployment to both clusters:
```
kubectl label deployment httpecho "state.internal.workloads.kcp.dev/cluster2"="Sync" "state.internal.workloads.kcp.dev/cluster1"="Sync" --overwrite
```

13. Observe the Deployment object, see how the Status field is empty, you will find the per-location status in the annotations: `experimental.status.workloads.kcp.dev/<workloadClusterName>` notice the reported statuses with different replicas.

14. Let's add a "soft finalizer" for "cluster1" used by external controllers as a way to add a back-pressure mechanism:
```
kubectl annotate deployment httpecho finalizers.workloads.kcp.dev/cluster1="mycontroller"
```

15. Now issue a deletion of the object:
```
kubectl delete deployment httpecho
```

16. Notice how the deletion is blocked, now in another terminal, let's add the per-location deletion timestamps (yes, for now, it is manual...):
```
kubectl annotate deployment httpecho deletion.internal.workloads.kcp.dev/cluster1="$(date -u +\"%Y-%m-%dT%H:%M:%SZ\")" --overwrite

kubectl annotate deployment httpecho deletion.internal.workloads.kcp.dev/cluster2="$(date -u +\"%Y-%m-%dT%H:%M:%SZ\")" --overwrite
``` 

17. Check the Deployment object again. All the labels and syncer finalizers referencing cluster2 should be gone, not cluster1 as we have a soft finalizer set. Let's remove it.
```
kubectl annotate deployment finalizers.workloads.kcp.dev/cluster1-
```

18. Check the previous terminal, where the kubectl command was blocked waiting for the deployment deletion. It should have finished, and the deployment httpEcho should be gone from your KCP Workspace.


cmd/syncer/options/options.go
```diff
@@ -50,11 +52,13 @@ func NewOptions() *Options {
 
 func (options *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&options.FromKubeconfig, "from-kubeconfig", options.FromKubeconfig, "Kubeconfig file for -from cluster.")
+	fs.StringVar(&options.FromContext, "from-context", options.FromContext, "Context to use in the Kubeconfig file for -from cluster, instead of the current context.")
+	fs.StringVar(&options.FromServer, "from-server", options.FromServer, "The overridden address and port of the Kubernetes API server to use in the Kubeconfig file for -from cluster.")
```
sttts, line 56: already commented earlier. Why this change? Seems unrelated and unneeded.

pkg/localenvoy/controllers/ingress/envoyingress.go
```diff
@@ -29,14 +29,14 @@ import (
...
-	nscontroller "github.com/kcp-dev/kcp/pkg/reconciler/workload/namespace"
+	"github.com/kcp-dev/kcp/pkg/syncer"
 )
 
 // reconcile is triggered on every change to an ingress resource.
 func (c *Controller) reconcile(ctx context.Context, ingress *networkingv1.Ingress) error {
 	klog.InfoS("reconciling Ingress", "ClusterName", logicalcluster.From(ingress), "Namespace", ingress.Namespace, "Name", ingress.Name)
 
-	if ingress.Labels[nscontroller.ClusterLabel] == "" {
+	if syncer.GetAssignedWorkloadCluster(ingress.Labels) == "" {
```
sttts, line 39: wrong package. This is something from the workload API and should live in pkg/apis/workload/helpers or so.

pkg/syncer/specsyncer.go
```diff
@@ -88,7 +116,11 @@ func (c *Controller) deleteFromDownstream(ctx context.Context, gvr schema.GroupV
 	// TODO: get UID of just-deleted object and pass it as a precondition on this delete.
 	// This would avoid races where an object is deleted and another object with the same name is created immediately after.
 
-	return c.toClient.Resource(gvr).Namespace(namespace).Delete(ctx, name, metav1.DeleteOptions{})
+	if err := c.toClient.Resource(gvr).Namespace(namespace).Delete(ctx, name, metav1.DeleteOptions{}); errors.IsNotFound(err) {
```
sttts, line 119: Better style:

```
... err != nil && !IsNotFound(err) {
   return err
}

return nil
```

pkg/syncer/helpers.go
```diff
@@ -0,0 +1,62 @@
...
+
+func SyncerFinalizerName(workloadClusterName string) string {
+	return "workloads.kcp.dev/syncer-" + workloadClusterName
+}
+
+func WorkloadClusterLabelName(workloadClusterName string) string {
+	return "cluster.internal.workloads.kcp.dev/" + workloadClusterName
+}
+
+func GetAssignedWorkloadCluster(labels map[string]string) string {
```
sttts, line 45: so this is "GetSomeAssignedWorkloadCluster" ? Is it healthy that we have this? It's undeterministic. Where is it used?

pkg/reconciler/workload/namespace/namespace_reconcile.go
```diff
@@ -40,12 +40,13 @@ import (
...
 	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
 	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/syncer"
 	conditionsapi "github.com/kcp-dev/kcp/third_party/conditions/apis/conditions/v1alpha1"
 	"github.com/kcp-dev/kcp/third_party/conditions/util/conditions"
 )
 
 const (
-	ClusterLabel            = "workloads.kcp.dev/cluster"
+	ScheduledClusterLabel   = "workloads.kcp.dev/scheduled-cluster"
```
sttts, line 49: am confused. What is this now after we added the new labels?

pkg/reconciler/workload/namespace/namespace_reconcile.go
```diff
@@ -64,13 +65,13 @@ var (
 
 func init() {
 	// This matches namespaces that haven't been scheduled yet
-	if req, err := labels.NewRequirement(ClusterLabel, selection.DoesNotExist, []string{}); err != nil {
+	if req, err := labels.NewRequirement(ScheduledClusterLabel, selection.DoesNotExist, []string{}); err != nil {
```
sttts, line 68: does this mean we need a label to tell that we are scheduled, and then one per cluster? Would it make sense to have a list of cluster names in this one, and then call the other "state"?

pkg/syncer/specsyncer.go
```diff
@@ -143,11 +175,44 @@ func (c *Controller) ensureDownstreamNamespaceExists(ctx context.Context, downst
...
+		upstreamFinalizers := upstreamObj.GetFinalizers()
+		hasFinalizer := false
+		for _, finalizer := range upstreamFinalizers {
+			if finalizer == SyncerFinalizerName(c.pcluster) {
+				hasFinalizer = true
+			}
+		}
+		if !hasFinalizer {
+			upstreamFinalizers = append(upstreamFinalizers, SyncerFinalizerName(c.pcluster))
+			upstreamObj.SetFinalizers(upstreamFinalizers)
```
sttts, line 193: this is bad style. Never mutate inputs if it is not super clear that the func is intended to do that. `notifySyncerOwnership` does not suggest side-effects, actually not even that is does some client call. `ensureSyncerFinalizer` would be much better.

sttts: this is even much worse if I am not mistaken: the upstream object comes directly from an informer. So this func mutates cached data, leading to inconsistencies, and also to races.

pkg/syncer/specsyncer.go
```diff
@@ -143,11 +175,44 @@ func (c *Controller) ensureDownstreamNamespaceExists(ctx context.Context, downst
...
+	klog.Infof("Updated resource %s|%s/%s to notify syncer ownership", c.upstreamClusterName, namespace, name)
+	return nil
+}
+
+func (c *Controller) applyToDownstream(ctx context.Context, eventType watch.EventType, gvr schema.GroupVersionResource, downstreamNamespace string, upstreamObj *unstructured.Unstructured) error {
 	if err := c.ensureDownstreamNamespaceExists(ctx, downstreamNamespace, upstreamObj); err != nil {
 		return err
 	}
 
+	if eventType == watch.Added {
```
sttts, line 210: don't act by edges, but by level (= state). Just check whether the finalizer is there and update if it is not.

pkg/syncer/specsyncer.go
```diff
@@ -173,9 +238,66 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
 		}
 	}
 
+	if advancedSchedulingFeatureEnabled {
+		specDiffPatch := upstreamObj.GetAnnotations()[LocationSpecDiffAnnotationName(c.pcluster)]
+		if specDiffPatch != "" {
+			upstreamSpec, specExists, err := unstructured.NestedFieldCopy(upstreamObj.UnstructuredContent(), "spec")
```
sttts, line 244: why copy?

pkg/syncer/helpers.go
```diff
@@ -0,0 +1,62 @@
...
+
+package syncer
+
+import "strings"
+
+func LocationDeletionAnnotationName(workloadClusterName string) string {
+	return "deletion.internal.workloads.kcp.dev/" + workloadClusterName
+}
+
+func LocationFinalizersAnnotationName(workloadClusterName string) string {
```
sttts, line 25: go doc what this is

pkg/syncer/helpers.go
```diff
@@ -0,0 +1,62 @@
...
+	return "finalizers.workloads.kcp.dev/" + workloadClusterName
+}
+
+func LocationStatusAnnotationName(workloadClusterName string) string {
```
sttts, line 29: go doc what this is. Better: move into api

pkg/syncer/helpers.go
```diff
@@ -0,0 +1,62 @@
...
+	return "experimental.status.workloads.kcp.dev/" + workloadClusterName
+}
+
+func LocationSpecDiffAnnotationName(workloadClusterName string) string {
+	return "experimental.specdiff.workloads.kcp.dev/" + workloadClusterName
```
sttts, line 34: go doc what this is.

pkg/syncer/helpers.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func SyncerFinalizerName(workloadClusterName string) string {
+	return "workloads.kcp.dev/syncer-" + workloadClusterName
+}
+
+func WorkloadClusterLabelName(workloadClusterName string) string {
+	return "cluster.internal.workloads.kcp.dev/" + workloadClusterName
```
sttts, line 42: go doc what this is. Better: move into api

pkg/syncer/specsyncer.go
```diff
@@ -173,9 +238,66 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
...
 
+	// TODO(jmprusi): When using syncer virtual workspace we would check the DeletionTimestamp on the upstream object, instead of the DeletionTimestamp annotation,
+	//                as the virtual workspace will set the the deletionTimestamp() on the location view by a transformation.
+	intendedToBeRemovedFromLocation := upstreamObj.GetAnnotations()[LocationDeletionAnnotationName(c.pcluster)] != ""
+
+	// TODO(jmprusi): When using syncer virtual workspace this condition would not be necessary anymore, since directly tested on the virtual workspace side.
+	stillOwnedByExternalActorForLocation := upstreamObj.GetAnnotations()[LocationFinalizersAnnotationName(c.pcluster)] != ""
+
+	if intendedToBeRemovedFromLocation && !stillOwnedByExternalActorForLocation {
+		if err := c.toClient.Resource(gvr).Namespace(downstreamNamespace).Delete(ctx, downstreamObj.GetName(), metav1.DeleteOptions{}); err != nil {
```
sttts, line 285: why do we delete the namespace?!?!

sttts: nvmd. Misread.

pkg/syncer/statussyncer.go
```diff
@@ -64,7 +70,60 @@ func NewStatusSyncer(from, to *rest.Config, gvrs []string, kcpClusterName logica
 	return New(kcpClusterName, pclusterID, fromClient, toClient, SyncUp, gvrs, pclusterID, mutatorsMap)
 }
 
-func (c *Controller) updateStatusInUpstream(ctx context.Context, gvr schema.GroupVersionResource, upstreamNamespace string, downstreamObj *unstructured.Unstructured) error {
+func (c *Controller) deleteFromUpstream(ctx context.Context, gvr schema.GroupVersionResource, upstreamNamespace, name string) error {
+	return c.removeFinalizersAndUpdate(ctx, c.toClient, gvr, upstreamNamespace, name)
+}
+
+func (c *Controller) removeFinalizersAndUpdate(ctx context.Context, upstreamClient dynamic.Interface, gvr schema.GroupVersionResource, upstreamNamespace, name string) error {
```
sttts, line 77: give it a better name. Is this about upstream? Downstream?

pkg/syncer/statussyncer.go
```diff
@@ -64,7 +70,60 @@ func NewStatusSyncer(from, to *rest.Config, gvrs []string, kcpClusterName logica
...
+	delete(annotations, LocationSpecDiffAnnotationName(c.pcluster))
+	upstreamObj.SetAnnotations(annotations)
+
+	// remove the cluster label.
+	upstreamLabels := upstreamObj.GetLabels()
+	delete(upstreamLabels, WorkloadClusterLabelName(c.pcluster))
+	upstreamObj.SetLabels(upstreamLabels)
+	// - End of block to be removed once the virtual workspace syncer is integrated -
+
+	if _, err := upstreamClient.Resource(gvr).Namespace(upstreamObj.GetNamespace()).Update(ctx, upstreamObj, metav1.UpdateOptions{}); err != nil {
```
sttts, line 118: what if this conflicts? If I am not mistaken this is edge based now.

pkg/syncer/statussyncer.go
```diff
@@ -64,7 +70,60 @@ func NewStatusSyncer(from, to *rest.Config, gvrs []string, kcpClusterName logica
 	return New(kcpClusterName, pclusterID, fromClient, toClient, SyncUp, gvrs, pclusterID, mutatorsMap)
 }
 
-func (c *Controller) updateStatusInUpstream(ctx context.Context, gvr schema.GroupVersionResource, upstreamNamespace string, downstreamObj *unstructured.Unstructured) error {
+func (c *Controller) deleteFromUpstream(ctx context.Context, gvr schema.GroupVersionResource, upstreamNamespace, name string) error {
+	return c.removeFinalizersAndUpdate(ctx, c.toClient, gvr, upstreamNamespace, name)
```
sttts, line 74: with this the function name is lying

pkg/syncer/statussyncer.go
```diff
@@ -90,11 +158,28 @@ func (c *Controller) updateStatusInUpstream(ctx context.Context, gvr schema.Grou
...
+		newUpstream := existing.DeepCopy()
+		statusAnnotationValue, err := json.Marshal(downstreamStatus)
+		if err != nil {
+			return err
+		}
+		newUpstreamAnnotations := newUpstream.GetAnnotations()
+		newUpstreamAnnotations[LocationStatusAnnotationName(c.pcluster)] = string(statusAnnotationValue)
+		newUpstream.SetAnnotations(newUpstreamAnnotations)
+
+		if _, err := c.toClient.Resource(gvr).Namespace(upstreamNamespace).Update(ctx, newUpstream, metav1.UpdateOptions{}); err != nil {
```
sttts, line 172: missing deepequal comparisons before updating. Also elsewhere.

pkg/syncer/syncer.go
```diff
@@ -130,6 +135,21 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 		return err
 	}
 
+	// TODO(marun) Report pcluster connectivity to kcp
```
sttts, line 138: ?

pkg/syncer/syncer.go
```diff
@@ -180,15 +192,17 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 }
 
 type mutatorGvrMap map[schema.GroupVersionResource]func(obj *unstructured.Unstructured) error
-type UpsertFunc func(ctx context.Context, gvr schema.GroupVersionResource, namespace string, unstrob *unstructured.Unstructured) error
+type UpsertFunc func(ctx context.Context, eventType watch.EventType, gvr schema.GroupVersionResource, namespace string, unstrob *unstructured.Unstructured) error
```
sttts, line 195: as written above: event type is a warning sign of edge based logic.

pkg/syncer/syncer.go
```diff
@@ -222,29 +238,34 @@ func New(kcpClusterName logicalcluster.LogicalCluster, pcluster string, fromClie
 		c.deleteFn = c.deleteFromDownstream
 	} else {
 		c.upsertFn = c.updateStatusInUpstream
+		c.deleteFn = c.deleteFromUpstream
 	}
 
 	fromInformers := dynamicinformer.NewFilteredDynamicSharedInformerFactory(fromClient, resyncPeriod, metav1.NamespaceAll, func(o *metav1.ListOptions) {
-		o.LabelSelector = fmt.Sprintf("%s=%s", nscontroller.ClusterLabel, pclusterID)
+		o.LabelSelector = WorkloadClusterLabelName(pclusterID) + " = Sync"
```
sttts, line 245: nit: no spaces around =

pkg/syncer/syncer.go
```diff
@@ -340,9 +361,10 @@ type holder struct {
 	clusterName logicalcluster.LogicalCluster
 	namespace   string
 	name        string
+	eventType   watch.EventType
```
sttts, line 364: as written before: no edge driven logic in controllers.

pkg/syncer/specsyncer.go
```diff
@@ -43,12 +45,38 @@ func deepEqualApartFromStatus(oldObj, newObj interface{}) bool {
...
 		return false
 	}
-	if !equality.Semantic.DeepEqual(oldUnstrob.GetAnnotations(), newUnstrob.GetAnnotations()) {
+
+	// TODO(jmprusi): Remove this after switching to virtual workspaces.
+	// remove status annotation from oldObj and newObj before comparing
+	oldAnnotations := oldUnstrob.GetAnnotations()
+	for k := range oldAnnotations {
+		if !strings.HasPrefix(k, LocationStatusAnnotationName("")) {
+			delete(oldAnnotations, k)
```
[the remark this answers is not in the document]
jmprusi: In this case, we need to remove from the comparison the status annotations to avoid reconcile loops, I'm not "setting" those annotations back into the Object (SetAnnotations()) and NestedStringMap() (used by GetAnnotations() seems to return a copy the annotations)

pkg/reconciler/workload/namespace/namespace_reconcile.go
```diff
@@ -307,26 +308,85 @@ func (c *Controller) setNamepaceContentsEnqueuedFor(ns *corev1.Namespace) {
...
+	}
+
+	patchLines = []string{
+		fmt.Sprintf("%q: %s,", syncer.WorkloadClusterLabelName(old), "null"),
+		fmt.Sprintf("%q: %q", syncer.WorkloadClusterLabelName(new), "Sync"),
 	}
 
-	return types.MergePatchType,
-		[]byte(fmt.Sprintf(`
+	return types.MergePatchType, []byte(fmt.Sprintf("%s%s%s", beginPatch, strings.Join(patchLines, "\n"), endPatch))
```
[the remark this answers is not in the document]
jmprusi: Done!

pkg/syncer/specsyncer.go
jmprusi: deepcopied the object, and changed the name.

pkg/syncer/statussyncer.go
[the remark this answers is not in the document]
jmprusi: Done

jmprusi: Done

pkg/syncer/syncer.go
jmprusi: done

pkg/syncer/helpers.go
sttts: have marked all of these deprecated.

pkg/syncer/specsyncer.go
```diff
@@ -181,35 +180,33 @@ func (c *Controller) ensureSyncerFinalizer(ctx context.Context, gvr schema.Group
...
 
-	// If the advanced scheduling feature is enabled, add the Syncer Finalizer to the upstream object
-	if advancedSchedulingFeatureEnabled {
-		upstreamFinalizers := upstreamObjCopy.GetFinalizers()
-		hasFinalizer := false
-		for _, finalizer := range upstreamFinalizers {
-			if finalizer == syncerFinalizerNamePrefix+c.pcluster {
-				hasFinalizer = true
-			}
+	upstreamFinalizers := upstreamObjCopy.GetFinalizers()
```
sttts, line 183: only deepcopy on demand, i.e. move the deepcopy call down behind the check

pkg/syncer/specsyncer.go
jmprusi: done

pkg/syncer/specsyncer.go
```diff
@@ -181,35 +180,33 @@ func (c *Controller) ensureSyncerFinalizer(ctx context.Context, gvr schema.Group
...
 		}
-		if !hasFinalizer {
-			upstreamFinalizers = append(upstreamFinalizers, syncerFinalizerNamePrefix+c.pcluster)
-			upstreamObjCopy.SetFinalizers(upstreamFinalizers)
+	}
+	if !hasFinalizer {
+		upstreamFinalizers = append(upstreamFinalizers, syncerFinalizerNamePrefix+c.pcluster)
+		upstreamObjCopy.SetFinalizers(upstreamFinalizers)
+		if _, err := c.fromClient.Resource(gvr).Namespace(namespace).Update(ctx, upstreamObjCopy, metav1.UpdateOptions{}); err != nil {
+			klog.Errorf("Failed updating to notify syncer ownership on resource %s|%s/%s: %v", c.upstreamClusterName, namespace, name, err)
```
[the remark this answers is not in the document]
jmprusi: done

cmd/syncer/options/options.go
sttts: still open

pkg/syncer/specsyncer.go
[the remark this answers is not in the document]
sttts: `GetAnnotations` does not guarantee that. Look at `metav1.Object`:
```
func (meta *ObjectMeta) GetAnnotations() map[string]string            { return meta.Annotations }
```
We cannot depend on that behaviour of unstructured. Just call `NestedStringMap` manually. Then it doesn't work here by coincidence.

pkg/syncer/specsyncer.go
```diff
@@ -43,12 +46,38 @@ func deepEqualApartFromStatus(oldObj, newObj interface{}) bool {
...
+	oldAnnotations := oldUnstrob.GetAnnotations()
+	for k := range oldAnnotations {
+		if !strings.HasPrefix(k, workloadv1alpha1.InternalClusterStatusAnnotationPrefix) {
+			delete(oldAnnotations, k)
+		}
+	}
+
+	newAnnotations := newUnstrob.GetAnnotations()
+	for k := range newAnnotations {
+		if strings.HasPrefix(k, workloadv1alpha1.InternalClusterStatusAnnotationPrefix) {
```
sttts, line 61: compare above: here the check is positive, above negative. Intentional?

pkg/syncer/specsyncer.go
```diff
@@ -173,9 +237,66 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
...
+		if err := c.toClient.Resource(gvr).Namespace(downstreamNamespace).Delete(ctx, downstreamObj.GetName(), metav1.DeleteOptions{}); err != nil {
+			if errors.IsNotFound(err) {
+				// That's not an error.
+				// Just think about removing the finalizer from the KCP location-specific resource:
+				if err := c.removeUpstreamSyncerOwnership(ctx, gvr, upstreamObj.GetNamespace(), upstreamObj.GetName()); err != nil {
+					return err
+				}
+				return nil
+			}
+			klog.Infof("Error deleting %s %s/%s from downstream %s|%s/%s: %v", gvr.Resource, upstreamObj.GetNamespace(), upstreamObj.GetName(), upstreamObj.GetClusterName(), upstreamObj.GetNamespace(), upstreamObj.GetName(), err)
```
sttts, line 293: klog.Error

pkg/syncer/specsyncer.go
```diff
@@ -173,9 +237,66 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
...
+			return err
+		}
+		klog.Infof("Deleted %s %s/%s from downstream %s|%s/%s", gvr.Resource, upstreamObj.GetNamespace(), downstreamObj.GetName(), upstreamObj.GetClusterName(), upstreamObj.GetNamespace(), upstreamObj.GetName())
```
sttts, line 296: V(2) at least

pkg/syncer/specsyncer.go
```diff
@@ -173,9 +238,66 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
...
+				if err := c.removeFinalizersAndUpdate(ctx, c.fromClient, gvr, upstreamObj.GetNamespace(), upstreamObj.GetName()); err != nil {
+					return err
+				}
+				return nil
+			}
+			klog.Infof("Error deleting %s %s/%s from downstream %s|%s/%s: %v", gvr.Resource, upstreamObj.GetNamespace(), upstreamObj.GetName(), upstreamObj.GetClusterName(), upstreamObj.GetNamespace(), upstreamObj.GetName(), err)
+			return err
+		}
+		klog.Infof("Deleted %s %s/%s from downstream %s|%s/%s", gvr.Resource, upstreamObj.GetNamespace(), downstreamObj.GetName(), upstreamObj.GetClusterName(), upstreamObj.GetNamespace(), upstreamObj.GetName())
+		return nil
```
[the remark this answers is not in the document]
sttts: still open

pkg/syncer/statussyncer.go
```diff
@@ -64,6 +71,59 @@ func NewStatusSyncer(from, to *rest.Config, gvrs []string, kcpClusterName logica
...
+	upstreamLabels := upstreamObj.GetLabels()
+	delete(upstreamLabels, workloadv1alpha1.InternalClusterResourceStateLabelPrefix+c.pcluster)
+	upstreamObj.SetLabels(upstreamLabels)
+	// - End of block to be removed once the virtual workspace syncer is integrated -
+
+	if _, err := c.toClient.Resource(gvr).Namespace(upstreamObj.GetNamespace()).Update(ctx, upstreamObj, metav1.UpdateOptions{}); err != nil {
+		klog.Errorf("Failed updating after removing the finalizers of resource %s|%s/%s: %v", c.upstreamClusterName, upstreamNamespace, upstreamObj.GetName(), err)
+		return err
+	}
+	klog.Infof("Updated resource %s|%s/%s after removing the finalizers", c.upstreamClusterName, upstreamNamespace, upstreamObj.GetName())
```
sttts, line 123: at least V(2)

pkg/apis/workload/v1alpha1/workloadcluster_types.go
```diff
@@ -0,0 +1,154 @@
...
+// +crd
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster,categories=kcp
+// +kubebuilder:printcolumn:name="Location",type="string",JSONPath=`.metadata.name`,priority=1
+// +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=`.status.conditions[?(@.type=="Ready")].status`,priority=2
+// +kubebuilder:printcolumn:name="Synced API resources",type="string",JSONPath=`.status.syncedResources`,priority=3
+
```
sttts, line 38: remove empty line (was pre-existing probably?)

pkg/syncer/specsyncer.go
```diff
@@ -173,9 +237,66 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
...
+	intendedToBeRemovedFromLocation := upstreamObj.GetAnnotations()[workloadv1alpha1.InternalClusterDeletionTimestampAnnotationPrefix+c.pcluster] != ""
+
+	// TODO(jmprusi): When using syncer virtual workspace this condition would not be necessary anymore, since directly tested on the virtual workspace side.
+	stillOwnedByExternalActorForLocation := upstreamObj.GetAnnotations()[workloadv1alpha1.ClusterFinalizerAnnotationPrefix+c.pcluster] != ""
+
+	if intendedToBeRemovedFromLocation && !stillOwnedByExternalActorForLocation {
+		if err := c.toClient.Resource(gvr).Namespace(downstreamNamespace).Delete(ctx, downstreamObj.GetName(), metav1.DeleteOptions{}); err != nil {
+			if errors.IsNotFound(err) {
+				// That's not an error.
+				// Just think about removing the finalizer from the KCP location-specific resource:
```
sttts, line 287: assume the syncer crashes here. How do we recover?

pkg/syncer/specsyncer.go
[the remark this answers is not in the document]
sttts: from discussion with @davidfestal: here the deletion process is not finished downstream. We have to depend on another reconcile.

pkg/syncer/statussyncer.go
```diff
@@ -64,6 +71,59 @@ func NewStatusSyncer(from, to *rest.Config, gvrs []string, kcpClusterName logica
 	return New(kcpClusterName, pclusterID, fromClient, toClient, SyncUp, gvrs, pclusterID, mutatorsMap)
 }
 
+func (c *Controller) deleteFromUpstream(ctx context.Context, gvr schema.GroupVersionResource, upstreamNamespace, name string) error {
+	return c.removeUpstreamSyncerOwnership(ctx, gvr, upstreamNamespace, name)
+}
+
+func (c *Controller) removeUpstreamSyncerOwnership(ctx context.Context, gvr schema.GroupVersionResource, upstreamNamespace, name string) error {
```
sttts, line 78: can we rename to ensureUpstreamFinalizerRemoved ?

pkg/syncer/specsyncer.go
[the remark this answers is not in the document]
davidfestal: Thanks for this explanation @sttts . Useful.

pkg/syncer/statussyncer.go
davidfestal: The changes done to this method based on [this comment](https://github.com/kcp-dev/kcp/pull/942#discussion_r861610345) don't seem right:
- the `upstreamClient dynamic.Interface` argument has been removed, and in replaced in the method by calling `c.toClient`
- However this method is also called, in a corner-case, [in the specsyncer](https://github.com/kcp-dev/kcp/pull/942/files#diff-e03cd260985652e7bb1a27c1c23905e46d846815d4084dd87145f55860c619ddR288) when the downstream object should be deleted, and in fact it has already been deleted.
- So when this method is called from the specsyncer, using the `c.toClient` corresponds to using to client to the downstream kube cluster, while we want this method to act n the upstream KCP logical cluster *in all cases*

To conclude I'd rather make this method a helper function, not a method of the controller, by adding the following parameters to it:
- `upstreamClient dynamic.Interface` (which has been dropped)
- `pcluster string` (take the opportunity to rename it to `workloadCluster`)
- `upstreamClusterName string` (take the opportunity to rename it as `logicalClusterName`)

davidfestal (changes requested):

pkg/syncer/statussyncer.go
```diff
@@ -64,6 +70,13 @@ func NewStatusSyncer(from, to *rest.Config, gvrs []string, kcpClusterName logica
 	return New(kcpClusterName, pclusterID, fromClient, toClient, SyncUp, gvrs, pclusterID, mutatorsMap)
 }
 
+func (c *Controller) deleteFromUpstream(ctx context.Context, gvr schema.GroupVersionResource, upstreamNamespace, name string) error {
+	if advancedSchedulingFeatureEnabled {
+		return removeUpstreamSyncerOwnership(ctx, gvr, c.toClient, upstreamNamespace, c.pcluster, c.upstreamClusterName, name)
```
sttts, line 75: removeSyncerFinalizerFromUpstream

pkg/syncer/statussyncer.go
```diff
@@ -64,6 +70,13 @@ func NewStatusSyncer(from, to *rest.Config, gvrs []string, kcpClusterName logica
 	return New(kcpClusterName, pclusterID, fromClient, toClient, SyncUp, gvrs, pclusterID, mutatorsMap)
 }
 
+func (c *Controller) deleteFromUpstream(ctx context.Context, gvr schema.GroupVersionResource, upstreamNamespace, name string) error {
```
sttts, line 73: move `if advancedSchedulingFeatureEnabled {` up to where `c.deleteFn` is called, with a closure.

pkg/syncer/statussyncer.go
jmprusi: done

pkg/syncer/syncer.go
sttts: would remove it. It's copied, but not sure about value

pkg/syncer/specsyncer.go
sttts: fixed above

pkg/syncer/specsyncer.go
sttts: Talking to @davidfestal: looks like this is still open. We have to also ensure finalizer removal when the downstream object is "NotFound".

test/e2e/reconciler/cluster/controller_test.go
```diff
@@ -69,7 +69,7 @@ func TestClusterController(t *testing.T) {
 					ObjectMeta: metav1.ObjectMeta{
 						Name: "timothy",
 						Labels: map[string]string{
-							nscontroller.ClusterLabel: sinkClusterName,
```
sttts, line 72: this should rather be a string literal. We want that tests break when we change a label or annotation in the main code.

pkg/syncer/syncer_process_test.go
```diff
@@ -0,0 +1,859 @@
...
+				deployment("theDeployment", "kcp0124d7647eb6a00b1fcb6f2252201601634989dd79deb7375c373973", "", nil, nil, nil),
+				addDeploymentStatus(appsv1.DeploymentStatus{
+					Replicas: 15,
+				})),
+			toResources: []runtime.Object{
+				deployment("theDeployment", "test", "root:org:ws", map[string]string{
+					"state.internal.workloads.kcp.dev/us-west1": "Sync",
+				}, nil, nil),
+			},
+			resourceToProcessLogicalClusterName: "",
```
sttts, line 201: what does it mean if this empty?

pkg/syncer/syncer_process_test.go
```diff
@@ -0,0 +1,859 @@
...
+			resourceToProcessName:               "theDeployment",
+			workloadClusterName:                 "us-west1",
+			advancedSchedulingEnabled:           false,
+
+			expectError:         false,
```
sttts, line 206: can we drop these if errors are not expected? That should be the default.

pkg/syncer/syncer_process_test.go
```diff
@@ -0,0 +1,859 @@
...
+									"containers": nil,
+								},
+							}, "spec", "template"),
+							setNestedField(map[string]interface{}{}, "status"),
+						),
+					),
+				),
+			},
+		},
+		"SpecSyncer with AdvancedSchedulingFeature deletion, object exist downstream": {
```
sttts, line 331: nit: exists

pkg/syncer/syncer_process_test.go
```diff
@@ -0,0 +1,859 @@
...
+			expectActionsOnTo: []clienttesting.Action{
+				createNamespaceAction(
+					"",
+					changeUnstructured(
+						toUnstructured(t, namespace("kcp0124d7647eb6a00b1fcb6f2252201601634989dd79deb7375c373973", "",
+							map[string]string{
+								"state.internal.workloads.kcp.dev/us-west1": "Sync",
+							},
+							map[string]string{
+								"kcp.dev/namespace-locator": "{\"logical-cluster\":\"root:org:ws\",\"namespace\":\"test\"}",
```
sttts, line 369: nit: use back-tick string syntax to avoid escaping.

pkg/syncer/syncer_process_test.go
```diff
@@ -0,0 +1,859 @@
...
+			expectActionsOnFrom: []clienttesting.Action{
+				getDeploymentAction("theDeployment", "test"),
+				updateDeploymentAction("test",
+					changeUnstructured(
+						toUnstructured(t, changeDeployment(
+							deployment("theDeployment", "test", "root:org:ws", nil, map[string]string{"another.valid.annotation/this": "value"}, nil),
+						),
+						),
+						// TODO(jmprusi): Those next changes do "nothing", it's just for the test to pass
+						//                as the test expects some null fields to be there...
```
sttts, line 415: what does it mean? We are unnecessarily set those zero values in the controller?

pkg/syncer/syncer_process_test.go
```diff
@@ -0,0 +1,859 @@
...
+									"containers": nil,
+								},
+							}, "spec", "template"),
+							setNestedField(map[string]interface{}{}, "status"),
+						),
+					),
+				),
+			},
+		},
+		"StatusSyncer with AdvancedSchedulingFeature": {
```
sttts, line 517: StatusSyncer what? Be a little more descriptive. Is this about updating a downstream object? Or creating, or deleting?

pkg/syncer/statussyncer.go
```diff
@@ -28,24 +29,29 @@ import (
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
+
+	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
 )
 
-func deepEqualStatus(oldObj, newObj interface{}) bool {
+func deepEqualFinalizersAndStatus(oldObj, newObj interface{}) bool {
```
sttts, line 36: simplify this by typing oldObj and newObj as unstructured. Where this is called we know everything is unstructured. That will also remove some test cases in the unit tests.

pkg/syncer/specsyncer.go
```diff
@@ -34,7 +37,7 @@ import (
 	"k8s.io/klog/v2"
 	"k8s.io/utils/pointer"
 
-	nscontroller "github.com/kcp-dev/kcp/pkg/reconciler/workload/namespace"
+	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
 )
 
 func deepEqualApartFromStatus(oldObj, newObj interface{}) bool {
```
sttts, line 43: make this strongly typed with unstructured. Where this is called you know everything is unstructured.

jmprusi: done

pkg/syncer/syncer_process_test.go
jmprusi: done

pkg/syncer/syncer_process_test.go
jmprusi: done

pkg/syncer/syncer_process_test.go
jmprusi: done

test/e2e/reconciler/cluster/controller_test.go
jmprusi: done

pkg/apis/workload/v1alpha1/workloadcluster_types.go
jmprusi: done

pkg/syncer/statussyncer.go
```diff
@@ -90,11 +105,37 @@ func (c *Controller) updateStatusInUpstream(ctx context.Context, gvr schema.Grou
...
+		}
+		newUpstreamAnnotations := newUpstream.GetAnnotations()
+		if newUpstreamAnnotations == nil {
+			newUpstreamAnnotations = make(map[string]string)
+		}
+		newUpstreamAnnotations[workloadv1alpha1.InternalClusterStatusAnnotationPrefix+c.pcluster] = string(statusAnnotationValue)
+		newUpstream.SetAnnotations(newUpstreamAnnotations)
+
+		if reflect.DeepEqual(existing, newUpstream) {
+			klog.V(2).Infof("No need to update the status of resource %s|%s/%s from pcluster namespace %s", c.upstreamClusterName, upstreamNamespace, name, downstreamObj.GetNamespace())
```
sttts, line 123: `V(4)`

Compare convention in https://github.com/kcp-dev/kcp/blob/main/CONTRIBUTING.md:

> - Default log-level is 2.
> - Controllers should generally log (a) one line (not more) non-error progress per item with klog.V(2) (b) actions like create/update/delete via klog.V(3) and (c) skipped actions, i.e. what was not done for reasons via klog.V(4).

pkg/syncer/syncer.go
```diff
@@ -130,29 +137,39 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
...
+	kcpClient := kcpClusterClient.Cluster(cfg.KCPClusterName)
+	workloadClustersClient := kcpClient.WorkloadV1alpha1().WorkloadClusters()
+	// Check whether we're in the Advanced Scheduling feature-gated mode.
+	workloadCluster, err := workloadClustersClient.Get(ctx, cfg.WorkloadClusterName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	advancedSchedulingEnabled := false
+	if workloadCluster.GetAnnotations()[advancedSchedulingFeatureAnnotation] == "true" {
+		klog.Infof("Advanced Scheduling feature is enabled for workloadCluster %s", cfg.WorkloadClusterName)
```
sttts, line 154: nit: WorkloadCluster

pkg/syncer/syncer.go
```diff
@@ -185,10 +202,12 @@ type DeleteFunc func(ctx context.Context, gvr schema.GroupVersionResource, names
 type HandlersProvider func(c *Controller, gvr schema.GroupVersionResource) cache.ResourceEventHandlerFuncs
 
 type Controller struct {
-	name  string
-	queue workqueue.RateLimitingInterface
+	name     string
+	pcluster string
```
sttts, line 206: workloadClusterName

pkg/syncer/syncer.go
```diff
@@ -197,20 +216,25 @@ type Controller struct {
 
 	upstreamClusterName logicalcluster.LogicalCluster
 	mutators            mutatorGvrMap
+
+	advancedSchedulingEnabled bool
 }
 
 // New returns a new syncer Controller syncing spec from "from" to "to".
-func New(kcpClusterName logicalcluster.LogicalCluster, pcluster string, fromClient, toClient dynamic.Interface, direction SyncDirection, gvrs []string, pclusterID string, mutators mutatorGvrMap) (*Controller, error) {
+func New(kcpClusterName logicalcluster.LogicalCluster, pcluster string, fromClient, toClient dynamic.Interface, direction SyncDirection, gvrs []string, pclusterID string, mutators mutatorGvrMap, advancedSchedulingEnabled bool) (*Controller, error) {
```
sttts, line 224: pclusterID => workloadClusterID

pkg/reconciler/workload/namespace/namespace_reconcile.go
jmprusi: done!

cmd/syncer/options/options.go
jmprusi: done

pkg/syncer/syncer_process_test.go
jmprusi: In the SyncUp cases, the logicalClusterName is empty.

pkg/syncer/statussyncer.go
jmprusi: done

pkg/syncer/statussyncer.go
jmprusi: done

sttts (approved):

PR #952, opened 2022-05-02
e2e: create syncer log artifacts when running in kind

sttts:
Untested.


ncdc:
I don't think the `t.Cleanup` functions are executed when an assertion like `require.Eventually` fails 😦

ncdc (approved):

PR #959, opened 2022-05-03
e2e/syncer: stop (some of) the bleeding

sttts:
Use a smaller busybox image 1.6GB => 2.66MB 🤡

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -145,22 +145,27 @@ jobs:
 
       # Publish a syncer image to the local registry
       - uses: imjasonh/setup-ko@v0.4
-      - run: KO_DOCKER_REPO=localhost:5000 ko publish ./cmd/syncer --insecure-registry -t $(git rev-parse --short "$GITHUB_SHA") > /tmp/syncer-image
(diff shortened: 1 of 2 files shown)


ncdc (approved):

.github/workflows/ci.yaml
```diff
@@ -145,22 +145,27 @@ jobs:
 
       # Publish a syncer image to the local registry
       - uses: imjasonh/setup-ko@v0.4
-      - run: KO_DOCKER_REPO=localhost:5000 ko publish ./cmd/syncer --insecure-registry -t $(git rev-parse --short "$GITHUB_SHA") > /tmp/syncer-image
+      - run: KO_DOCKER_REPO=${KIND_REGISTRY} ko publish ./cmd/syncer --insecure-registry -t $(git rev-parse --short "$GITHUB_SHA") > /tmp/syncer-image
```
aojea, line 148: where is `KIND_REGISTRY` defined?

sttts: from the kind action above I assume

aojea: ic , `container-tools/kind-action@v1`

ncdc:
According to https://hub.docker.com/r/bitnami/kubectl/tags, the compressed size is 46MB. But still, that's big 😄

PR #969, opened 2022-05-03
Towards more stable syncer 2

sttts:


pkg/informer/informer.go
```diff
@@ -208,45 +202,46 @@ func (d *DynamicDiscoverySharedInformerFactory) discoverTypes(ctx context.Contex
...
-		gvr, _ := schema.ParseResourceArg(gvrstr)
-		if gvr == nil {
-			// TODO(ncdc): consider tracking where each gvrstr came from (which workspace) so we can include that in the error.
-			multierr.AppendInto(&merr, fmt.Errorf("unable to parse %q to a GroupVersionResource", gvrstr))
-			continue
-		}
-
-		inf := d.dsif.ForResource(*gvr).Informer()
+	for _, gvr := range newGVRs {
+		klog.Infof("Adding informer for %q in workspace %s", gvr, logicalClusterNames)
```
ncdc, line 233: IIRC these are wildcard informers, so we can probably remove the part about which workspace the gvr came from, unless you want to log that it was this workspace that triggered the add?

ncdc:
Looks like this actually broke e2e?

pkg/informer/informer.go
sttts: Yes, that's the intention to understand what triggered the addition. Also note that we have no logic to remove them when the resource goes away.

sttts:
> Looks like this actually broke e2e?

Yes, looks like. Will run tests locally. Change looks sane, but maybe I missed something. Or accidently changed some odd behaviour from before. I still don't get how informers were readded.

PR #973, opened 2022-05-03
Add front proxy

kylape:
This adds the front proxy into the topology of the deployment manifest.

Instead of terminating TLS at the ingress router, it will instead terminate at the front proxy, and the proxy will route requests to either KCP or the virtual workspaces process.

This also leverages cert-manager to add the ability to authenticate using a certificate.  Groups will be assigned using the organizations listed on the cert.

### DIFF
--- a/manifest/kcp-front-proxy.yaml
+++ b/manifest/kcp-front-proxy.yaml
@@ -103,8 +103,8 @@ metadata:
   name: kcp-front-proxy-virtual-workspaces-client-cert
 spec:
   secretName: kcp-front-proxy-virtual-workspaces-client-cert
-  duration: 2160h # 90d
-  renewBefore: 360h # 15d
+  duration: 2160h0m0s # 90d
+  renewBefore: 360h0m0s # 15d
   subject:
     organizations:
       - redhat
--- a/manifest/kcp-front-proxy.yaml
+++ b/manifest/kcp-front-proxy.yaml
@@ -1,0 +1,258 @@
+---
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  name: kcp-front-proxy
+spec:
+  host: kcp
+  port:
+    targetPort: 8443
+  tls:
+    termination: passthrough
+  to:
+    kind: Service
+    name: kcp-front-proxy
+    weight: 100
+  wildcardPolicy: None
+---
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: kcp-front-proxy-issuer
+spec:
+  acme:
+    server: https://acme-v02.api.letsencrypt.org/directory
+    privateKeySecretRef:
+      name: kcp-front-proxy-issuer-account-key
+    solvers:
+    - http01:
+        ingress:
+          serviceType: ClusterIP
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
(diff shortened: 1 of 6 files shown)


aojea:
does this allow to use http2 to the backend?

csams:
@aojea The backends are setup here: https://github.com/kcp-dev/kcp/blob/main/pkg/proxy/proxy.go#L61-L68. We may need to update the `Transport` to enable http2. Would that just be setting `ForceAttemptHTTP2` to true?

manifest/kcp-front-proxy.yaml
```diff
@@ -0,0 +1,258 @@
...
+        - containerPort: 8443
+        command:
+        - /kcp-front-proxy
+        args:
+        - --secure-port=8443
+        - --tls-private-key-file=/etc/kcp-front-proxy/tls/tls.key
+        - --tls-cert-file=/etc/kcp-front-proxy/tls/tls.crt
+        - --client-ca-file=/etc/kcp-front-proxy/client/tls/ca.crt
+        - --mapping-file=/etc/kcp-front-proxy/config/path-mapping.yaml
+        - --v=4
```
csams, line 183: Let's change this to 6 so the proxy will log basic info about requests to the backends.

csams (changes requested):

kylape:
We can try out the changes on https://github.com/kcp-dev/kcp/pull/873 with this manifest to see if http2 works.

aojea:
> @aojea The backends are setup here: https://github.com/kcp-dev/kcp/blob/main/pkg/proxy/proxy.go#L61-L68. We may need to update the `Transport` to enable http2. Would that just be setting `ForceAttemptHTTP2` to true?

I would build the transport differently, the ReverseProxy uses http.DefaultTransport by default https://pkg.go.dev/net/http/httputil#ReverseProxy
```
// DefaultTransport is the default implementation of Transport and is
// used by DefaultClient. It establishes network connections as needed
// and caches them for reuse by subsequent calls. It uses HTTP proxies
// as directed by the $HTTP_PROXY and $NO_PROXY (or $http_proxy and
// $no_proxy) environment variables.
var DefaultTransport RoundTripper = &Transport{
	Proxy: ProxyFromEnvironment,
	DialContext: defaultTransportDialContext(&net.Dialer{
		Timeout:   30 * time.Second,
		KeepAlive: 30 * time.Second,
	}),
	ForceAttemptHTTP2:     true,
	MaxIdleConns:          100,
	IdleConnTimeout:       90 * time.Second,
	TLSHandshakeTimeout:   10 * time.Second,
	ExpectContinueTimeout: 1 * time.Second,
}
```

initializing it https://github.com/kcp-dev/kcp/blob/main/pkg/proxy/proxy.go#L61-L68 this way you are overriding some sane defaults. I would initialize it as:
```go
        transport = http.DefaultTransport
        transport.TLSClientConfig = &tls.Config{
			Certificates: []tls.Certificate{cert},
			RootCAs:      caCertPool,
		},
	proxy := httputil.NewSingleHostReverseProxy(target)
	proxy.Transport = transport
```

kylape:
Note that this is checking in an Openshift-specific resource (Route).  This is a short term solution; we intend to have a more generic manifest available for the community that doesn't assume any particular distribution.

Issue #920, opened 2022-04-19
`kubectl kcp workload sync` is not idempotent


Roming22:
**Describe the bug**
Running `kubectl kcp workload sync` twice with the same set of params triggers a failure on the 2nd run because the resources habe already been declared.

**To Reproduce**
1. Run `kubectl kcp workload sync my-test --syncer-image ghcr.io/kcp-dev/kcp/syncer-c2e3073d5026a8f7f2c47a50c16bdbec:41ca72b` -> OK
2. Run `kubectl kcp workload sync my-test --syncer-image ghcr.io/kcp-dev/kcp/syncer-c2e3073d5026a8f7f2c47a50c16bdbec:41ca72b` -> Failure

**Expected behavior**
The resources should be updated if necessary in the background, and the program should exit successfully.

**Additional context**
@marun Related to your latest commits.

itdove:
@ncdc 
1. What should be updated in the workloadcluster as the method provides only the name of the workload.
2. I guess once we will fix the workloadcluster create issue, we will have the same issue with the SA and clusterRoleBinding

Wonder if it exists we just have to skip because I don't see right away which information must be updated.

ncdc:
1. Nothing needs to be updated. I'd say if the workloadcluster exists, keep going instead of erroring.
2. You'll probably want to ensure the SA's ownerref is correct. And make sure the CRB has the right subjects + role ref.

Issue #947, opened 2022-04-29
Race condition in syncer: namespace vs. resource event ordering & cache freshness


qiujian16:
**Describe the bug**
see TestSyncerLifecycle fails sometimes in  e2e-test-sharedserver.

log as below

```
fixture.go:458: Waiting for cluster "pcluster-01" condition "Ready" to have reason ""
    fixture.go:480: Cluster "pcluster-01" is Ready
    syncer_test.go:71: Creating upstream namespace...
    syncer_test.go:92: Waiting for downstream service account kcp015dc2d48934724ea9e4592e34ab9b641cd5960514c116524629471b/kcp-default to be created...
    syncer_test.go:106: Waiting for downstream service account secret kcp015dc2d48934724ea9e4592e34ab9b641cd5960514c116524629471b/kcp-default-token to be created...
    syncer_test.go:119: Waiting for downstream configmap kcp015dc2d48934724ea9e4592e34ab9b641cd5960514c116524629471b/kcp-root-ca.crt to be created...
    syncer_test.go:132: Creating upstream deployment...
    syncer_test.go:148: Waiting for downstream deployment kcp015dc2d48934724ea9e4592e34ab9b641cd5960514c116524629471b/syncer-test to be created...
    syncer_test.go:149: 
        	Error Trace:	syncer_test.go:149
        	Error:      	Condition never satisfied
        	Test:       	TestSyncerLifecycle
        	Messages:   	downstream deployment kcp015dc2d48934724ea9e4592e34ab9b641cd5960514c116524629471b/syncer-test was not synced
    syncer_test.go:155: saw an error waiting for downstream deployment kcp015dc2d48934724ea9e4592e34ab9b641cd5960514c116524629471b/syncer-test to be created: context canceled
FAIL
FAIL	github.com/kcp-dev/kcp/test/e2e/syncer	49.501s
?   	github.com/kcp-dev/kcp/test/e2e/virtual/framework/compositecmd	[no test files]
ok  	github.com/kcp-dev/kcp/test/e2e/virtual/workspaces	3.376s
ok  	github.com/kcp-dev/kcp/test/e2e/workspacetype	11.625s
```

**Expected behavior**
the test should pass

**Details**
1. Namespace is scheduled to a workload cluster and has its labels patched

[...]

PR #975, opened 2022-05-03
Make 'kubectl kcp workload sync' idempotent

itdove:
Signed-off-by: Dominique Vernier <dvernier@redhat.com>

## Summary
Avoid creating workloadcluster, sa and clusterolebinding if they already exist
Overwrite the sa, crb with the calculated values
## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/920

### DIFF
--- a/pkg/cliplugins/workload/plugin/sync.go
+++ b/pkg/cliplugins/workload/plugin/sync.go
@@ -30,6 +30,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	kubernetesclientset "k8s.io/client-go/kubernetes"
@@ -131,17 +132,23 @@ func enableSyncerForWorkspace(ctx context.Context, config *rest.Config, workload
 		return "", fmt.Errorf("failed to create kcp client: %w", err)
 	}
 
-	// Create the workload cluster that will serve as a point of coordination between
-	// kcp and the syncer (e.g. heartbeating from the syncer and virtual cluster urls
-	// to the syncer).
-	workloadCluster, err := kcpClient.WorkloadV1alpha1().WorkloadClusters().Create(ctx,
-		&workloadv1alpha1.WorkloadCluster{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: workloadClusterName,
-			},
-		},
-		metav1.CreateOptions{},
+	workloadCluster, err := kcpClient.WorkloadV1alpha1().WorkloadClusters().Get(ctx,
+		workloadClusterName,
+		metav1.GetOptions{},
 	)
+	if errors.IsNotFound(err) {
+		// Create the workload cluster that will serve as a point of coordination between
+		// kcp and the syncer (e.g. heartbeating from the syncer and virtual cluster urls
+		// to the syncer).
+		workloadCluster, err = kcpClient.WorkloadV1alpha1().WorkloadClusters().Create(ctx,
+			&workloadv1alpha1.WorkloadCluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: workloadClusterName,
+				},
+			},
+			metav1.CreateOptions{},
+		)
+	}
 	if err != nil {
 		return "", fmt.Errorf("failed to create WorkloadCluster: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


ncdc:
Please remove the double # and make sure the issue number is immediately following the # - thanks!

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -131,17 +132,23 @@ func enableSyncerForWorkspace(ctx context.Context, config *rest.Config, workload
...
+			&workloadv1alpha1.WorkloadCluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: workloadClusterName,
+				},
+			},
+			metav1.CreateOptions{},
+		)
+	}
 	if err != nil {
 		return "", fmt.Errorf("failed to create WorkloadCluster: %w", err)
```
ncdc, line 153: If the get call fails, and it's something other than a not-found error, the error returned here won't be correct. What if we redid the entire flow like this?

1. Try to get
2. If err != nil && !errors.IsNotFound -> return error getting
3. If NotFound
    1. Try to create
    2. If err != nil && !errors.IsAlreadyExists -> return error creating

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -161,34 +168,65 @@ func enableSyncerForWorkspace(ctx context.Context, config *rest.Config, workload
...
-	sa, err := kubeClient.CoreV1().ServiceAccounts(namespace).Create(ctx, &corev1.ServiceAccount{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:            authResourceName,
-			OwnerReferences: workloadClusterOwnerReferences,
-		},
-	}, metav1.CreateOptions{})
-	if err != nil {
-		return "", fmt.Errorf("failed to create ServiceAccount: %w", err)
+	sa, err := kubeClient.CoreV1().ServiceAccounts(namespace).Get(ctx, authResourceName,
+		metav1.GetOptions{})
```
ncdc, line 172: Suggest making this just 1 line

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -161,34 +168,65 @@ func enableSyncerForWorkspace(ctx context.Context, config *rest.Config, workload
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:            authResourceName,
+				OwnerReferences: workloadClusterOwnerReferences,
+			},
+		}, metav1.CreateOptions{}); err != nil {
+			return "", fmt.Errorf("failed to create ServiceAccount: %w", err)
+		}
+	case err == nil:
+		//Overwrite the ownerref in case of a user changed it
+		sa.ObjectMeta.OwnerReferences = workloadClusterOwnerReferences
```
ncdc, line 186: It's fine to do sa.OwnerReferences (I saw you had a separate commit to adjust) - we usually omit the embedded ObjectMeta in the line

ncdc: It's potentially safer to add the ownerref if it doesn't exist, but don't overwrite any other existing ones (although I'm not sure how likely that case would be).

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -161,34 +168,65 @@ func enableSyncerForWorkspace(ctx context.Context, config *rest.Config, workload
...
+		if sa, err = kubeClient.CoreV1().ServiceAccounts(namespace).Update(ctx, sa, metav1.UpdateOptions{}); err != nil {
```
ncdc, line 187: We try to do `Patch()` instead of `Update()` to avoid potential data loss (old client imports old types that don't know about new fields -> Update will cause the new fields to be dropped, Patch will not). Please look in the repo for examples where we call Patch(). Thanks!

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -161,34 +168,65 @@ func enableSyncerForWorkspace(ctx context.Context, config *rest.Config, workload
...
+
+	switch {
+	case errors.IsNotFound(err):
+		if sa, err = kubeClient.CoreV1().ServiceAccounts(namespace).Create(ctx, &corev1.ServiceAccount{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:            authResourceName,
+				OwnerReferences: workloadClusterOwnerReferences,
+			},
+		}, metav1.CreateOptions{}); err != nil {
+			return "", fmt.Errorf("failed to create ServiceAccount: %w", err)
```
ncdc, line 182: Global comment: please include the full name of the resource in log and error messages.

If you know the logical cluster (aka clusterName): `"%s|%s/%s", clusterName, namespace, name`

If not: `"%s/%s", namespace, name`

ncdc: Also, just in case there's some (weird) race, don't error on AlreadyExists

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -161,34 +168,65 @@ func enableSyncerForWorkspace(ctx context.Context, config *rest.Config, workload
...
+	case errors.IsNotFound(err):
+		if _, err = kubeClient.RbacV1().ClusterRoleBindings().Create(ctx, &rbacv1.ClusterRoleBinding{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:            authResourceName,
+				OwnerReferences: workloadClusterOwnerReferences,
+			},
+			Subjects: subjects,
+			RoleRef:  roleRef,
+		}, metav1.CreateOptions{}); err != nil {
+			return "", err
```
ncdc, line 219: Just in case there's some (weird) race, don't error on AlreadyExists

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -161,34 +168,65 @@ func enableSyncerForWorkspace(ctx context.Context, config *rest.Config, workload
...
+		}
+	case err == nil:
+		//Overwrite ownerref,subject and roleRef in case a user changed it
+		crb.ObjectMeta.OwnerReferences = workloadClusterOwnerReferences
```
ncdc, line 223: As above, ensure it's contained within, instead of doing a wholesale replacement

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -161,34 +168,65 @@ func enableSyncerForWorkspace(ctx context.Context, config *rest.Config, workload
...
+		crb.Subjects = subjects
+		crb.RoleRef = roleRef
+		if _, err = kubeClient.RbacV1().ClusterRoleBindings().Update(ctx, crb, metav1.UpdateOptions{}); err != nil {
```
ncdc, line 226: Update -> Patch

ncdc (changes requested):
This is a great start, thanks for picking it up so quickly!

What do you think about adding an e2e test for this, to make sure it works?

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -161,34 +175,116 @@ func enableSyncerForWorkspace(ctx context.Context, config *rest.Config, workload
...
+				OwnerReferences: workloadClusterOwnerReferences,
+			},
+		}, metav1.CreateOptions{}); err != nil && !errors.IsAlreadyExists(err) {
+			return "", fmt.Errorf("failed to create ServiceAccount %s|%s/%s: %w", workloadClusterName, namespace, authResourceName, err)
+		}
+	case err == nil:
+		oldData, err := json.Marshal(corev1.ServiceAccount{
+			ObjectMeta: metav1.ObjectMeta{
+				UID:             sa.UID,
+				ResourceVersion: sa.ResourceVersion,
```
ncdc, line 194: If you want the UID & ResourceVersion in the json path (as preconditions), you'll need to remove them from oldData

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -161,34 +175,116 @@ func enableSyncerForWorkspace(ctx context.Context, config *rest.Config, workload
...
+				OwnerReferences: sa.OwnerReferences,
+			},
+		})
+		if err != nil {
+			return "", fmt.Errorf("failed to Marshal old data for ServiceAccount %s|%s/%s: %w", workloadClusterName, namespace, authResourceName, err)
```
ncdc, line 199:
```suggestion
			return "", fmt.Errorf("failed to marshal old data for ServiceAccount %s|%s/%s: %w", workloadClusterName, namespace, authResourceName, err)
```

ncdc: Global comment

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -161,34 +175,116 @@ func enableSyncerForWorkspace(ctx context.Context, config *rest.Config, workload
...
+			Subjects: subjects,
+			RoleRef:  roleRef,
+		}, metav1.CreateOptions{}); err != nil && !errors.IsAlreadyExists(err) {
+			return "", err
+		}
+	case err == nil:
+		oldData, err := json.Marshal(rbacv1.ClusterRoleBinding{
+			ObjectMeta: metav1.ObjectMeta{
+				UID:             crb.UID,
+				ResourceVersion: crb.ResourceVersion,
```
ncdc, line 256: Remove to use as preconditions

ncdc:
LGTM, please squash, thanks!

ncdc (approved):

ncdc:
Unrelated test failure #947

Issue #950, opened 2022-05-02
e2e: clusterworkspacedeletion flakes


sttts:
```
--- FAIL: TestWorkspaceDeletionController (16.34s)
    util.go:117: Saving test artifacts for test "TestWorkspaceDeletionController" under "/tmp/e2e/TestWorkspaceDeletionController/1866300474".
    fixture.go:144: Starting kcp servers...
    kcp.go:200: running: /home/runner/work/kcp/kcp/bin/kcp start --root-directory /tmp/e2e/TestWorkspaceDeletionController/1866300474/data/kcp/shared --secure-port=44697 --embedded-etcd-client-port=[34](https://github.com/kcp-dev/kcp/runs/6255297079?check_suite_focus=true#step:5:34)957 --embedded-etcd-peer-port=38459 --embedded-etcd-wal-size-bytes=5000 --kubeconfig-path=admin.kubeconfig --auto-publish-apis --discovery-poll-interval=5s --token-auth-file /tmp/e2e/TestWorkspaceDeletionController/1866300474/data/auth-tokens.csv --run-virtual-workspaces=true --feature-gates=KCPLocationAPI=true
    kcp.go:4[36](https://github.com/kcp-dev/kcp/runs/6255297079?check_suite_focus=true#step:5:36): success contacting https://10.1.0.199:44697/readyz
    kcp.go:436: success contacting https://10.1.0.199:44697/livez
    fixture.go:171: Started kcp servers after 16.340057932s
    --- FAIL: TestWorkspaceDeletionController/nested_worksapce_cleanup_when_an_org_workspace_is_deleted ([39](https://github.com/kcp-dev/kcp/runs/6255297079?check_suite_focus=true#step:5:39).28s)
        controller_test.go:155: Should have finalizer in org workspace
        controller_test.go:170: Create a workspace with in the org workspace
        controller_test.go:174: Should have finalizer added in workspace
        controller_test.go:188: Create namespace in both the workspace and org workspace
        controller_test.go:203: Delete org workspace
        controller_test.go:207: Ensure namespace in the workspace is deleted
        controller_test.go:217: Ensure namespace in the org workspace is deleted
        controller_test.go:227: Ensure workspace in the org workspace is deleted
        controller_test.go:228: 
            	Error Trace:	controller_test.go:228
            	            				controller_test.go:271
            	Error:      	Condition never satisfied
[...]

PR #978, opened 2022-05-04
reconciler/workload/scheduling: remove wrong resource enqueuing avoiding optimization

sttts:
The namespace scheduler used `namespaceContentsEnqueuedForMap` to avoid double requeuing of resources. @ncdc found a bug of this logic that led to resources missed during reschedule of the namespace resources:

> I have a theory on the syncer flakes:
> 1. Namespace get scheduled to `cluster-x`
>     a. It’s patched
>     b. `c.namespaceContentsEnqueuedForMap[nsKey]` is set to `cluster-x`
> 2. Dynamic informer sees new deployment, ns controller `reconcileResource` is called
>     a. When it gets the ns from the cache, it’s stale (doesn’t have the cluster label yet)
>     b. It compares the deployment’s cluster label (unset) with the ns’s cluster label (unset) and returns early -> no assignment
> 3. NS informer sees updated ns (after it was patched), reconcileNamespace is called, which calls `enqueueResourcesForNamespace`, but because it was previously reconciled and the scheduling decision hasn’t changed, it returns early

`enqueueResourcesForNamespace` is a pure in-memory operation. Optimizing it away is a premature optimization. Instead we should be level-driven as much as we can. It's easy to just postpone the resource enqueuing until after the namespace cache is not stale anymore, i.e. on the next reconciliation of the namespace. This is what this PR is doing, avoiding resource scheduling with a stale namespace lister.

Note these thoughts about races:
1. after the patch of the namespace we can be sure, the namespace is requeued. When that happens, we can compare the cluster label on the resources and only enqueue those that differ. We know that those from a stale resource cache will "come back" with another resource reconcilation. By that time the namespace cache is up-to-date.
2. on the other hand, if a resource is reconciled before the namespace cache is fresh, the resource will possibly scheduling to the wrong location. But the namespace will be reconciled and will requeue this resource to fix the wrong location.

### DIFF
--- a/pkg/reconciler/workload/namespace/namespace_reconcile.go
+++ b/pkg/reconciler/workload/namespace/namespace_reconcile.go
@@ -93,11 +93,11 @@ func init() {
 // any type, to match the cluster where its namespace is assigned.
 func (c *Controller) reconcileResource(ctx context.Context, lclusterName logicalcluster.LogicalCluster, unstr *unstructured.Unstructured, gvr *schema.GroupVersionResource) error {
 	if gvr.Group == "networking.k8s.io" && gvr.Resource == "ingresses" {
-		klog.V(2).Infof("Skipping reconciliation of ingress %s/%s", unstr.GetNamespace(), unstr.GetName())
+		klog.V(4).Infof("Skipping reconciliation of ingress %s/%s", unstr.GetNamespace(), unstr.GetName())
 		return nil
 	}
 
-	klog.Infof("Reconciling %s %s|%s/%s", gvr.String(), lclusterName, unstr.GetNamespace(), unstr.GetName())
+	klog.V(4).Infof("Reconciling %s %s|%s/%s", gvr.String(), lclusterName, unstr.GetNamespace(), unstr.GetName())
 
 	// If the resource is not namespaced (incl if the resource is itself a
 	// namespace), ignore it.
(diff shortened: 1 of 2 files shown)


pkg/reconciler/workload/namespace/namespace_reconcile.go
```diff
@@ -244,79 +239,67 @@ func (c *Controller) reconcileNamespace(ctx context.Context, lclusterName logica
...
+		first := true
 		for _, obj := range objs {
-			c.enqueueResource(gvr, obj)
+			u := obj.(*unstructured.Unstructured)
+			objLocation := u.GetLabels()[ClusterLabel]
+			if objLocation != nsLocation {
+				c.enqueueResource(gvr, obj)
+
+				if first {
+					klog.V(2).Infof("Enqueuing some %s in namespace %s|%s to schedule to %q", gvr, ns.ClusterName, ns.Name, nsLocation)
```
ncdc, line 282: I'm confused about this... we have obj - why log this the first time through at v(2)?

ncdc: Also it's only enqueuing this specific obj, not the entire gvr

ncdc:
test is still failing w/the same issue?

pkg/reconciler/workload/namespace/namespace_reconcile.go
sttts: to see progress, but without too much noise with many objects.

sttts: not perfect, but best I could come up with without logging per-object. Any better idea? We could log a limit number of resources, e.g. 10, in some slice in that line.

ncdc: I see - that's fine to do. Maybe add a comment indicating that?

pkg/reconciler/workload/namespace/namespace_reconcile.go
```diff
@@ -172,7 +172,7 @@ func (c *Controller) reconcileGVR(ctx context.Context, gvr schema.GroupVersionRe
 // will succeed without error if a cluster is assigned or if there are no viable clusters
 // to assign to. The condition of not being scheduled to a cluster will be reflected in
 // the namespace's status rather than by returning an error.
-func (c *Controller) ensureScheduled(ctx context.Context, ns *corev1.Namespace) (*corev1.Namespace, error) {
+func (c *Controller) ensureScheduled(ctx context.Context, ns *corev1.Namespace) (*corev1.Namespace, bool, error) {
```
ncdc, line 175: I would suggest using named return variables so it's clear what this bool is. Also would update the godoc to add it.

Alternatively (follow-up), refactor this so it's not assigning + patching + returning true/false all in one method?

pkg/reconciler/workload/namespace/namespace_reconcile.go
```diff
@@ -244,79 +244,61 @@ func (c *Controller) reconcileNamespace(ctx context.Context, lclusterName logica
...
 		}
+
 		for _, obj := range objs {
-			c.enqueueResource(gvr, obj)
+			u := obj.(*unstructured.Unstructured)
+			objLocation := u.GetLabels()[ClusterLabel]
+			if objLocation != nsLocation {
+				c.enqueueResource(gvr, obj)
+
+				klog.Infof("Enqueuing %s %s|%s/%s to schedule to %q", gvr.GroupVersion().WithKind(u.GetKind()), ns.ClusterName, ns.Name, u.GetName(), nsLocation)
```
ncdc, line 285: I'm guessing you'll lower this to 2 or 4 after we're confident in the fix?

pkg/reconciler/workload/namespace/namespace_reconcile.go
```diff
@@ -244,79 +244,61 @@ func (c *Controller) reconcileNamespace(ctx context.Context, lclusterName logica
...
+			} else {
+				klog.Infof("Skipping %s %s|%s/%s because it is already scheduled to %q", gvr.GroupVersion().WithKind(u.GetKind()), ns.ClusterName, ns.Name, u.GetName(), nsLocation)
```
ncdc, line 287: I'm guessing you'll lower this to 2 or 4+ after we're confident in the fix?

ncdc:
e2e-shared-server passed 😄

pkg/reconciler/workload/namespace/namespace_reconcile.go
sttts: yes, that's just in the DEBUG commit

pkg/reconciler/workload/namespace/namespace_reconcile.go
sttts: yes, that's just in the DEBUG commit

sttts:
> e2e-shared-server passed 😄

I hoped it would fail.

ncdc:
Let's make sure we have enough at V(2) so we can triage in the future if it fails again?

ncdc:
e2e-multiple-runs: #950

ncdc (approved):

ncdc:
:happydance:

PR #987, opened 2022-05-04
DO NOT MERGE DEBUG

ncdc:

## Related issue(s)

Fixes #


ncdc:
Failure was #950

pkg/informer/informer.go
```diff
@@ -51,18 +51,66 @@ type clusterDiscovery interface {
...
+
+	if inf != nil {
+		return inf
+	}
+
+	// Grab the write lock, then find-or-create
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	return d.informerForResourceLH(gvr)
```
stevekuznetsov, line 87: What does `LH` mean?

pkg/informer/informer.go
```diff
@@ -104,15 +150,16 @@ func NewDynamicDiscoverySharedInformerFactory(
...
-	dsif := dynamicinformer.NewDynamicSharedInformerFactory(dynClient, resyncPeriod)
 	return DynamicDiscoverySharedInformerFactory{
 		workspaceLister: workspaceLister,
 		disco:           disco,
-		dsif:            dsif,
+		dynamicClient:   dynClient,
 		handler:         handler,
 		filterFunc:      filterFunc,
-		gvrs:            sets.NewString(),
+		gvrs:            make(map[schema.GroupVersionResource]struct{}),
```
stevekuznetsov, line 159: super-nit: a couple lines makes this into a first-class Set (which we can do generically now, I guess) and will improve readability

pkg/informer/informer.go
```diff
@@ -212,42 +260,78 @@ func (d *DynamicDiscoverySharedInformerFactory) discoverTypes(ctx context.Contex
...
+			informersToRemove = append(informersToRemove, gvr)
+		}
+	}
+	d.mu.RUnlock()
+
+	if len(informersToAdd) == 0 && len(informersToRemove) == 0 {
+		return nil
+	}
+
+	// We have to add/remove, so we need the write lock
```
stevekuznetsov, line 289: Isn't there a race here in case someone else grabbed the write lock first and mucked with things since we last read?

pkg/reconciler/workload/namespace/namespace_reconcile.go
```diff
@@ -262,18 +269,29 @@ func (c *Controller) reconcileNamespace(ctx context.Context, lclusterName logica
 // enqueueResourcesForNamespace adds the resources contained by the given
 // namespace to the queue if there scheduling label differs from the namespace's.
 func (c *Controller) enqueueResourcesForNamespace(ns *corev1.Namespace) error {
+	clusterName := logicalcluster.From(ns)
+
 	nsLocation := ns.Labels[ClusterLabel]
 
+	klog.V(4).Infof("enqueueResourcesForNamespace(%s|%s): getting listers", clusterName, ns.Name)
```
stevekuznetsov, line 276: We can start proliferating structured logs at any point, this might be

```go
klog.V(4).InfoS("getting listers", "method", "enqueueResourcesForNamespace", "clusterName", clusterName, "namespace", ns.Name)
```

pkg/informer/informer.go
ncdc: LockHeld

pkg/informer/informer.go
ncdc: We're still on 1.17

ncdc: But I can wrap this in a set if desired

pkg/informer/informer.go
```diff
@@ -51,18 +51,66 @@ type clusterDiscovery interface {
...
-	return d.dsif.ForResource(gvr).Informer().GetIndexer()
+	return d.InformerForResource(gvr).Informer().GetIndexer()
+}
+
+// InformerForResource returns the GenericInformer for gvr, creating it if needed.
+func (d *DynamicDiscoverySharedInformerFactory) InformerForResource(gvr schema.GroupVersionResource) informers.GenericInformer {
+	// See if we already have it
+	inf := func() informers.GenericInformer {
+		d.mu.RLock()
+		defer d.mu.RUnlock()
```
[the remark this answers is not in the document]
ncdc: Ok

pkg/informer/informer.go
ncdc: Will ponder

pkg/informer/informer.go
sttts: am used to WithLock suffix

pkg/informer/informer.go
sttts: would leave it as it is IMO. Sets are Java'ism.

pkg/informer/informer.go
ncdc: Just pushed LockHeld. That ok or want another rename?

pkg/informer/informer.go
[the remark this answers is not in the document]
ncdc: Fixed

pkg/informer/informer.go
ncdc: Fixed

pkg/informer/informer.go
stevekuznetsov: Eh. Less Java-ism and more readability. `foo[thing] = interface{}{}` is strictly worse than `set.Insert(thing)` :)

Feel free to ignore.

stevekuznetsov (commented):
With the race fixed, this LGTM. Wonder if we might want to prefer the simple write-lock, check-once code unless we can benchmark the impact easily? Having to make sure we *always* follow the read, calculate, maybe-early-exit, write, re-check flow will be hard on review in perpetuity.

PR #989, opened 2022-05-05
Fix workspace deletion e2e flake

qiujian16:
1. increase wait time
2. pass the test when list workspace return 404

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/950

### DIFF
--- a/test/e2e/reconciler/clusterworkspacedeletion/controller_test.go
+++ b/test/e2e/reconciler/clusterworkspacedeletion/controller_test.go
@@ -227,6 +227,11 @@ func TestWorkspaceDeletionController(t *testing.T) {
 				t.Logf("Ensure workspace in the org workspace is deleted")
 				require.Eventually(t, func() bool {
 					wslist, err := server.orgKcpClient.TenancyV1alpha1().ClusterWorkspaces().List(ctx, metav1.ListOptions{})
+					// 404 could be returned if the org workspace is deleted.
+					if apierrors.IsNotFound(err) {
+						return true
+					}
+
 					if err != nil {
 						return false
 					}
@@ -238,7 +243,6 @@ func TestWorkspaceDeletionController(t *testing.T) {
 				require.Eventually(t, func() bool {
 					_, err := server.rootKcpClient.TenancyV1alpha1().ClusterWorkspaces().Get(ctx, orgWorkspace.Name, metav1.GetOptions{})
 					return apierrors.IsNotFound(err)
-
 				}, wait.ForeverTestTimeout, 1*time.Second)
 			},
 		},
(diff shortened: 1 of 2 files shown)


test/e2e/reconciler/clusterworkspacedeletion/controller_test.go
```diff
@@ -227,19 +227,23 @@ func TestWorkspaceDeletionController(t *testing.T) {
 				t.Logf("Ensure workspace in the org workspace is deleted")
 				require.Eventually(t, func() bool {
 					wslist, err := server.orgKcpClient.TenancyV1alpha1().ClusterWorkspaces().List(ctx, metav1.ListOptions{})
+					// 404 could be returned if the org workspace is deleted.
```
stevekuznetsov, line 230: Why isn't this also true for the namespace list calls above this?

test/e2e/reconciler/clusterworkspacedeletion/controller_test.go
```diff
@@ -227,19 +227,23 @@ func TestWorkspaceDeletionController(t *testing.T) {
...
+						return true
+					}
+
 					if err != nil {
 						return false
 					}
 
 					return len(wslist.Items) == 0
-				}, wait.ForeverTestTimeout, 1*time.Second)
+				}, 2*wait.ForeverTestTimeout, 1*time.Second)
```
stevekuznetsov, line 240: 2 * forever = smell. Do we know what takes more than 30s here?

ncdc: I actually am thinking that handling the 404 will fix the problem, so we could start with that, and not do this 2* change now

test/e2e/reconciler/clusterworkspacedeletion/controller_test.go
ncdc: I believe this is coming from https://github.com/kcp-dev/kcp/blob/decced493792aeda1e171b9c860cc9efbcff00fd/pkg/server/apiextensions.go#L147. I see this getting logged repeatedly when the e2e is about to fail, presumably from the test trying to list clusterworkspaces every 1s

stevekuznetsov: Hm, so you don't get a 404 when you ask for a built-in type even when the workspace does not exist? Interesting.

stevekuznetsov (approved):

test/e2e/reconciler/clusterworkspacedeletion/controller_test.go
ncdc:
```
k get --raw '/clusters/root:bob/api/v1/namespaces'
{"kind":"NamespaceList","apiVersion":"v1","metadata":{"resourceVersion":"1983"},"items":[]}
```

sttts: What I don't like here: with this change we do not test what we wanted to test anymore. I think we have to use this etcd client pattern of checking directly in etcd. A 404 of a list does not say anything about deletion of the objects.

ncdc: https://github.com/kcp-dev/kcp/issues/992

ncdc: Let's get this in, then follow up

ncdc:
Rebased on top of main to pick up fix for #947

PR #991, opened 2022-05-05
Don't log when systemCRDProvider can't find ws

ncdc:
## Summary
The systemCRDProvider was logging an error every time a request for a system CRD was received and the workspace couldn't be found. Stop doing that, as it's a legitimate situation.

## Related issue(s)

Fixes #


sttts (approved):

ncdc:
Rebased to pick up race fix

PR #1005, opened 2022-05-06
add kcp workload cordon/uncordon/drain

### LINKED ISSUE #525: kubectl kcp workload cordon|drain [cluster]

As a pcluster admin, I want to easily cordon, uncordon and drain clusters using the kcp CLI plugin

Add CLI surfaces to the `kubectl kcp` CLI plugin to cordon and drain clusters:

- `kubectl kcp workload cordon [cluster]`
- `kubectl kcp workload drain [cluster]`
- `kubectl kcp workload uncordon [cluster]`

People can just patch the Cluster resource today, but that's a bit clunkier.

**Additional context**

The behavior of the CLI should mirror the existing `kubectl` surface for (un-)cordoning and draining nodes in a cluster: https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/

chrisahl:
Signed-off-by: Chris Ahl <cahl@redhat.com>

## Summary
Add kcp cli cmds for workload cordon/drain/uncordon

## Related issue(s)
Fixes #525

### DIFF
--- a/pkg/cliplugins/workload/cmd/cmd.go
+++ b/pkg/cliplugins/workload/cmd/cmd.go
@@ -42,7 +42,7 @@ var (
 	%[1]s workload uncordon <workload-cluster-name>
 `
 	drainExample = `
-	# Drain a workload cluster in preparation for maintenance.
+	# Start draining a workload cluster in preparation for maintenance.
 	%[1]s workload drain <workload-cluster-name>
 `
 )
@@ -182,7 +182,7 @@ func New(streams genericclioptions.IOStreams) (*cobra.Command, error) {
 	// drain
 	drainCmd := &cobra.Command{
 		Use:          "drain <workload-cluster-name>",
-		Short:        "Drain workload cluster in preparation for maintenance",
+		Short:        "Start draining workload cluster in preparation for maintenance",
 		Example:      fmt.Sprintf(drainExample, "kubectl kcp"),
 		SilenceUsage: true,
 		RunE: func(c *cobra.Command, args []string) error {
--- a/pkg/cliplugins/workload/cmd/cmd.go
+++ b/pkg/cliplugins/workload/cmd/cmd.go
@@ -33,6 +33,18 @@ var (
 	# Ensure a syncer is running on the specified workload cluster.
 	%[1]s workload sync <workload-cluster-name> --syncer-image <kcp-syncer-image>
 `
+	cordonExample = `
+	# Mark a workload cluster as unschedulable.
+	%[1]s workload cordon <workload-cluster-name>
+`
+	uncordonExample = `
+	# Mark a workload cluster as schedulable.
+	%[1]s workload uncordon <workload-cluster-name>
+`
+	drainExample = `
+	# Drain a workload cluster in preparation for maintenance.
+	%[1]s workload drain <workload-cluster-name>
+`
 )
 
 // New provides a cobra command for workload operations.
@@ -112,6 +124,87 @@ func New(streams genericclioptions.IOStreams) (*cobra.Command, error) {
 	enableSyncerCmd.Flags().StringVar(&kcpNamespaceName, "kcp-namespace", kcpNamespaceName, "The name of the kcp namespace to create a service account in.")
 
 	cmd.AddCommand(enableSyncerCmd)
+
+	// cordon
+	cordonCmd := &cobra.Command{
+		Use:          "cordon <workload-cluster-name>",
+		Short:        "Mark workload cluster as unschedulable",
+		Example:      fmt.Sprintf(cordonExample, "kubectl kcp"),
+		SilenceUsage: true,
+		RunE: func(c *cobra.Command, args []string) error {
+			if err := opts.Validate(); err != nil {
+				return err
+			}
+			kubeconfig, err := plugin.NewConfig(opts)
+			if err != nil {
+				return err
+			}
+
+			if len(args) != 1 {
+				return cmd.Help()
+			}
+
+			workloadClusterName := args[0]
+
+			return kubeconfig.Cordon(c.Context(), workloadClusterName)
+		},
+	}
+
+	cmd.AddCommand(cordonCmd)
+
+	// uncordon
+	uncordonCmd := &cobra.Command{
+		Use:          "uncordon <workload-cluster-name>",
+		Short:        "Mark workload cluster as schedulable",
+		Example:      fmt.Sprintf(uncordonExample, "kubectl kcp"),
+		SilenceUsage: true,
+		RunE: func(c *cobra.Command, args []string) error {
+			if err := opts.Validate(); err != nil {
+				return err
+			}
+			kubeconfig, err := plugin.NewConfig(opts)
+			if err != nil {
+				return err
+			}
+
+			if len(args) != 1 {
+				return cmd.Help()
+			}
+
+			workloadClusterName := args[0]
+
+			return kubeconfig.Uncordon(c.Context(), workloadClusterName)
+		},
+	}
+
+	cmd.AddCommand(uncordonCmd)
+
+	// drain
+	drainCmd := &cobra.Command{
+		Use:          "drain <workload-cluster-name>",
+		Short:        "Drain workload cluster in preparation for maintenance",
+		Example:      fmt.Sprintf(drainExample, "kubectl kcp"),
+		SilenceUsage: true,
+		RunE: func(c *cobra.Command, args []string) error {
+			if err := opts.Validate(); err != nil {
+				return err
+			}
+			kubeconfig, err := plugin.NewConfig(opts)
+			if err != nil {
+				return err
+			}
+
+			if len(args) != 1 {
+				return cmd.Help()
+			}
+
+			workloadClusterName := args[0]
+
+			return kubeconfig.Drain(c.Context(), workloadClusterName)
+		},
+	}
+
+	cmd.AddCommand(drainCmd)
 
 	return cmd, nil
 }
--- a/pkg/cliplugins/workload/plugin/workload.go
+++ b/pkg/cliplugins/workload/plugin/workload.go
@@ -134,7 +134,7 @@ func modifyCordon(ctx context.Context, config *rest.Config, workloadClusterName 
 	return nil
 }
 
-// Uncordon the workload cluster and mark it as schedulable
+// Start draining the workload cluster and mark it as unschedulable
 func (c *Config) Drain(ctx context.Context, workloadClusterName string) error {
 	config, err := clientcmd.NewDefaultClientConfig(*c.startingConfig, c.overrides).ClientConfig()
 	if err != nil {
--- a/pkg/cliplugins/workload/plugin/workload.go
+++ b/pkg/cliplugins/workload/plugin/workload.go
@@ -1,0 +1,116 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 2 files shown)


ncdc:
@chrisahl thanks! If this closes out the related issue, please make sure in the description it says `Fixes #525`

pkg/cliplugins/workload/plugin/workload.go
```diff
@@ -0,0 +1,123 @@
...
+		return fmt.Errorf("failed to get workloadcluster %s: %w", workloadClusterName, err)
+	}
+
+	workloadCluster.Spec.Unschedulable = cordon
+	// if uncordon, make sure evictAfter is not set
+	if !cordon {
+		workloadCluster.Spec.EvictAfter = nil
+	}
+
+	_, err = kcpClient.WorkloadV1alpha1().WorkloadClusters().Update(ctx, workloadCluster, metav1.UpdateOptions{})
```
ncdc, line 82: Mind switching this to a Patch call instead? Example: https://github.com/kcp-dev/kcp/blob/3f393f5389079c194625f6cd357b847553f11d4a/pkg/cliplugins/workload/plugin/sync.go#L191-L216

pkg/cliplugins/workload/plugin/workload.go
```diff
@@ -0,0 +1,123 @@
...
+		return fmt.Errorf("failed to get workloadcluster %s: %w", workloadClusterName, err)
+	}
+
+	nowTime := metav1.NewTime(time.Now())
+	workloadCluster.Spec.EvictAfter = &nowTime
+
+	//ensure unschedulable is also set
+	workloadCluster.Spec.Unschedulable = true
+
+	_, err = kcpClient.WorkloadV1alpha1().WorkloadClusters().Update(ctx, workloadCluster, metav1.UpdateOptions{})
```
ncdc, line 116: Ditto re Patch

pkg/cliplugins/workload/plugin/workload.go
```diff
@@ -0,0 +1,123 @@
...
+	)
+	if err != nil {
+		return fmt.Errorf("failed to get workloadcluster %s: %w", workloadClusterName, err)
+	}
+
+	nowTime := metav1.NewTime(time.Now())
+	workloadCluster.Spec.EvictAfter = &nowTime
+
+	//ensure unschedulable is also set
+	workloadCluster.Spec.Unschedulable = true
```
ncdc, line 114: UX wise, would you want to reject drain if unschedulable is unset, or automatically set it to true?

pkg/cliplugins/workload/plugin/workload.go
```diff
@@ -0,0 +1,123 @@
...
+	workloadCluster, err := kcpClient.WorkloadV1alpha1().WorkloadClusters().Get(ctx,
+		workloadClusterName,
+		metav1.GetOptions{},
+	)
+	if err != nil {
+		return fmt.Errorf("failed to get workloadcluster %s: %w", workloadClusterName, err)
+	}
+
+	nowTime := metav1.NewTime(time.Now())
+	workloadCluster.Spec.EvictAfter = &nowTime
```
ncdc, line 111: Do you think it's useful to let the user optionally specify a time in the future, e.g "drain in 30 minutes"?

chrisahl:
So this is still WIP and I was not yet expecting you to review.  The `drain` is currently async and for nodes they have it sync.  Still a bunch to investigate on `drain`, (and maybe un/cordon too) but I did want to at least open the PR in draft to be sure I could get a PR pushed. 

I will incorporate all the comments already provided (Thanks!) and will have to get used to how your team does PR flow.  Seems all teams do something slightly different so a learning curve on that.

ncdc:
> So this is still WIP and I was not yet expecting you to review

Sorry about that, was trying to provide some fast feedback 😄. I'll hold off and wait for you to let us know when it's ready. Thanks for the PR!

pkg/cliplugins/workload/plugin/workload.go
chrisahl: Based on what I see in kubectl, `drain` causes unschedulable to be set and `evictAfter` to be set

pkg/cliplugins/workload/plugin/workload.go
chrisahl: Changed to patches

pkg/cliplugins/workload/plugin/workload.go
chrisahl: Changed to patches

pkg/cliplugins/workload/plugin/workload.go
chrisahl: At a future point, if a customer really needs it, a time delay option might be able to be added.  But right now, I don't see if being needed.   There are various aspects of drain for pod that cannot be done for workloadcluster.  Most importantly the ability to have it be sync so that when it returns, the drain is complete.  I think there would need to be additional status added to workloadcluster in order for drain to know when to return.

pkg/cliplugins/workload/plugin/workload.go
```diff
@@ -0,0 +1,165 @@
...
+	// make the changes needed to the resource
+	workloadCluster.Spec.Unschedulable = cordon
+
+	// if uncordon, make sure evictAfter is not set
+	if !cordon {
+		workloadCluster.Spec.EvictAfter = nil
+	}
+
+	// Get a JSON copy of the changed data
+	newData, err := json.Marshal(workloadCluster)
```
stevekuznetsov, line 97: If I understand this correctly, this will encode both the UID and resourceVersion in the patch, which will cause this command to fail when either change while it's running. For UID, that seems like what we want. What do we expect for UX on the resourceVersion? Do we want this command to fail if it races with some other status update or something? If not, we should put this into a retry loop or remove the resourceVersion from the patch.

stevekuznetsov: [Here](https://github.com/kcp-dev/kcp/blob/69aed93556a6729fcc8c5e11eed458b11603e0f8/pkg/reconciler/tenancy/clusterworkspaceshard/clusterworkspaceshard_controller.go#L165-L171) is an example of building an explicit new object to diff against. You'd just want to do it with the Spec.

chrisahl: @stevekuznetsov I have submitted a change based on your comments

pkg/cliplugins/workload/plugin/workload.go
```diff
@@ -0,0 +1,204 @@
...
+	if (!cordon) && (workloadCluster.Spec.EvictAfter != nil) {
+		newObj.Spec.EvictAfter = nil
+	}
+
+	newData, err := json.Marshal(newObj)
+	if err != nil {
+		return fmt.Errorf("failed to marshal new data for WorkloadCluster %s: %w", workloadClusterName, err)
+	}
+
+	patchBytes, err := jsonpatch.CreateMergePatch(oldData, newData)
```
sttts, line 118: isn't a jsonpatch basically a one-liner for the same 20 merge patch generation code here?

pkg/cliplugins/workload/plugin/workload.go
```diff
@@ -0,0 +1,204 @@
...
+	if cordon {
+		fmt.Println(workloadClusterName, "cordoned")
+	} else {
+		fmt.Println(workloadClusterName, "uncordoned")
+	}
+
+	return nil
+}
+
+// Uncordon the workload cluster and mark it as schedulable
```
sttts, line 137: wrong comment?

pkg/cliplugins/workload/cmd/cmd.go
```diff
@@ -113,5 +125,86 @@ func New(streams genericclioptions.IOStreams) (*cobra.Command, error) {
...
+
+			return kubeconfig.Uncordon(c.Context(), workloadClusterName)
+		},
+	}
+
+	cmd.AddCommand(uncordonCmd)
+
+	// drain
+	drainCmd := &cobra.Command{
+		Use:          "drain <workload-cluster-name>",
```
sttts, line 184: wouldn't a `--evict-after` flag make sense to override the "now" timestamp?

pkg/cliplugins/workload/cmd/cmd.go
```diff
@@ -32,6 +32,18 @@ var (
...
+	cordonExample = `
+	# Mark a workload cluster as unschedulable.
+	%[1]s workload cordon <workload-cluster-name>
+`
+	uncordonExample = `
+	# Mark a workload cluster as schedulable.
+	%[1]s workload uncordon <workload-cluster-name>
+`
+	drainExample = `
+	# Drain a workload cluster in preparation for maintenance.
```
sttts, line 45: more like "Start draining". This command does not wait. We might want to add waiting in the future vi a `--wait`

pkg/cliplugins/workload/plugin/workload.go
chrisahl: fixing

pkg/cliplugins/workload/plugin/workload.go
chrisahl: Not sure I understand your comment.  Previous comment by @stevekuznetsov indicated to change from Update to Patch and gave an example that I followed.

pkg/cliplugins/workload/plugin/workload.go
```diff
@@ -0,0 +1,204 @@
...
+	} else if !cordon && !workloadCluster.Spec.Unschedulable {
+		fmt.Println(workloadClusterName, "already uncordoned")
+		return nil
+	}
+
+	// Build a JSON copy of the changed data
+	newObj := workloadv1alpha1.WorkloadCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			UID:             workloadCluster.UID,
+			ResourceVersion: workloadCluster.ResourceVersion,
```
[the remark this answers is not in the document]
chrisahl: The patch did not work if that was not there.
```
k kcp workload uncordon cahl-aws-4-10-5         
error: failed to update WorkloadCluster cahl-aws-4-10-5: workloadclusters.workload.kcp.dev "cahl-aws-4-10-5" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update
```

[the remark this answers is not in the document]
chrisahl: And this follows the example you linked to.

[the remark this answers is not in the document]
ncdc: You'd need to clear the resource version in the old object before converting it to json. Then you can exclude it from the new object too

pkg/cliplugins/workload/plugin/workload.go
ncdc: Stefan is suggesting to manually create the patch bytes

pkg/cliplugins/workload/cmd/cmd.go
chrisahl: an evict-after option could be added later if really needed but I do not see it as being needed to start with.  kubectl node drain does not offer such an option

pkg/cliplugins/workload/plugin/workload.go
chrisahl: I have changed the way the patch is built. See if it is what you are looking for (there are several ways to build patches)

chrisahl:
@sttts @ncdc @stevekuznetsov Can I get another review?

pkg/cliplugins/workload/cmd/cmd.go
sttts: Agree.

pkg/cliplugins/workload/cmd/cmd.go
sttts: +1

ncdc (approved):

PR #1006, opened 2022-05-06
Forbid kcp.dev CRD groups outside system:system-crds

### LINKED ISSUE #772: Forbid CRD creation of kcp.dev groups outside of system:system-crds
*.kcp.dev CRDs should be reserved for the system, i.e. a user should not be able to install these outside of the shadow workspace `system:system-crds` (compare #755 where it was introduced to host the kcp CRDs). This issue is about adding an admission plugin to stop `*.kcp.dev` CRD creation in any other workspace.

robinbobbitt:
Signed-off-by: Robin Y Bobbitt <rbobbitt@redhat.com>

## Summary

Adds an admission plugin that ensures *.kcp.dev CRDs are only allowed in system:system-crds.

## Related issue(s)

Fixes #772

### DIFF
--- a/pkg/admission/reservedcrdgroups/admission.go
+++ b/pkg/admission/reservedcrdgroups/admission.go
@@ -1,0 +1,76 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reservedcrdgroups
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"strings"
+
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	"k8s.io/apiserver/pkg/admission"
+	"k8s.io/apiserver/pkg/endpoints/request"
+)
+
+const (
+	PluginName = "kcp.dev/ReservedCRDGroups"
+)
+
+func Register(plugins *admission.Plugins) {
+	plugins.Register(PluginName,
+		func(_ io.Reader) (admission.Interface, error) {
+			return &reservedCRDGroups{
+				Handler: admission.NewHandler(admission.Create, admission.Update),
+			}, nil
+		})
+}
+
+type reservedCRDGroups struct {
+	*admission.Handler
+}
+
+// Ensure that the required admission interfaces are implemented.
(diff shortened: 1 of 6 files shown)


pkg/admission/reservedcrdgroups/admission.go
```diff
@@ -0,0 +1,76 @@
...
+	crd, ok := a.GetObject().(*apiextensions.CustomResourceDefinition)
+	if !ok {
+		return fmt.Errorf("unexpected type %T", a.GetObject())
+	}
+
+	clusterName, err := request.ClusterNameFrom(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve cluster from context: %w", err)
+	}
+	if clusterName.String() == "system:system-crds" {
```
ncdc, line 68: Not sure if this would cause an import cycle, and/or if there's a better location for it, but we do have a const in pkg/server for this: `SystemCRDLogicalCluster`

ncdc (commented):
Thanks for this! What do you think about also adding an end to end test that validates this as well (to make sure we don't regress having the admission plugin enabled & running)?

pkg/admission/reservedcrdgroups/admission.go
sttts: literal is fine. If we want to be safe, we could turn this into a public constant here and make sure in unit test in the server that both match.

sttts:
Sgtm. Small e2e would be good.

pkg/admission/reservedcrdgroups/admission.go
robinbobbitt: Yes I used the const from `server` originally and hit the import cycle, and didn't see a more general spot for constants to move `SystemCRDLogicalCluster`. I can try @sttts suggestion or should we start a constants file somewhere?

robinbobbitt:
Thank you both for the review. I will look into adding an e2e test.

test/e2e/customresourcedefinition/wildwest.kcp.dev_cowboys.yaml
```diff
@@ -0,0 +1,59 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.7.0
+    apis.kcp.dev/bound-crd: t
```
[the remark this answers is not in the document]
robinbobbitt: Doh! Good catch. Removing.

pkg/admission/reservedcrdgroups/admission.go
robinbobbitt: Unit test added [here](https://github.com/kcp-dev/kcp/pull/1006/commits/a503ecba0f4358aefd8e71f9f6639c1b4bcbb848) per @sttts

ncdc:
@robinbobbitt please run `make imports` - thanks 😄

robinbobbitt:
> @robinbobbitt please run `make imports` - thanks 😄

Done! Thanks.

ncdc (approved):

PR #1011, opened 2022-05-09
syncer: refactor into two normal controllers

sttts:
Turn the syncer into two normal, separate controllers, removing all the interdependencies between status and spec sync.


pkg/syncer/controller.go
```diff
@@ -0,0 +1,443 @@
...
+		},
+	)
+}
+
+// Start starts N worker processes processing work items.
+func (c *Controller) Start(ctx context.Context, numThreads int) {
+	defer runtime.HandleCrash()
+	defer c.queue.ShutDown()
+
+	c.fromInformers.Start(ctx.Done())
```
davidfestal, line 184: Am I right that the informer of the other side passed to either `NewSpecSyncer` or `NewStatusSyncer` is never started and waited for cache sync ?

sttts: Good point. Updated.

pkg/syncer/controller.go
```diff
@@ -0,0 +1,443 @@
...
+					if !deepEqualFinalizersAndStatus(oldUnstrob, newUnstrob) {
+						c.AddToQueue(*gvr, newUnstrob)
+					}
+				}
+			},
+			DeleteFunc: func(obj interface{}) {
+				c.AddToQueue(*gvr, obj)
+			},
+		})
+		klog.InfoS("Set up informer", "direction", c.direction, "clusterName", kcpClusterName, "pcluster", pcluster, "gvr", gvr)
```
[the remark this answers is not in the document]
davidfestal: Might be worth doing it before @itdove work on managing the delete from downstream case though, isn't it ?

pkg/syncer/controller.go
davidfestal: Do you still need to `Start` and `WaitForCacheSync` here now that it is done here: https://github.com/kcp-dev/kcp/pull/1011/files#diff-b043590efbb8609f4eced1377afbe4f51358c5c628fb79c0fdd24e8f9d553551R186 ?

sttts: oh, lost that change during rebase it seems.

sttts: fixed

pkg/syncer/syncer.go
```diff
@@ -93,6 +94,35 @@ func (sc *SyncerConfig) ID() string {
...
+	if err != nil {
+		return err
+	}
+	upstreamDiscoveryClient, err := discovery.NewDiscoveryClientForConfig(cfg.UpstreamConfig)
+	if err != nil {
+		return err
+	}
+
+	upstreamInformers := dynamicinformer.NewFilteredDynamicSharedInformerFactory(upstreamDynamicClient.Cluster(cfg.KCPClusterName), resyncPeriod, metav1.NamespaceAll, func(o *metav1.ListOptions) {
+		o.LabelSelector = workloadv1alpha1.InternalClusterResourceStateLabelPrefix + cfg.WorkloadClusterName + "=" + string(workloadv1alpha1.ResourceStateSync)
```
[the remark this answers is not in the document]
sttts: this will change in followup from @davidfestal. This is now only moved code.

pkg/syncer/controller.go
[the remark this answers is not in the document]
sttts: The spec syncer will hook into the downstream informer from its constructor, not from the shared code here.

pkg/syncer/status/status_process.go
```diff
@@ -44,25 +46,72 @@ func deepEqualFinalizersAndStatus(oldUnstrob, newUnstrob *unstructured.Unstructu
...
-	// Register the default mutators
-	mutatorsMap := getDefaultMutators(from)
+	// get the downstream object
+	obj, exists, err := c.downstreamInformers.ForResource(gvr).Informer().GetIndexer().GetByKey(key)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		// deleted downstream => remove finalizer upstream
+		klog.InfoS("Downstream GVR %q object %s|%s/%s does not exist. Removing finalizer upstream", gvr.String(), downstreamClusterName, upstreamNamespace, name)
```
davidfestal, line 102: Don't you want to add this log message inside the conditional block ?

sttts: 👍 , fixed

davidfestal (approved):
Great !

PR #1017, opened 2022-05-10
DRAFT: apis/tenancy: rework ClusterWorkspaceType

sttts:
- make `ClusterWorkspaceTypes` hierarchical in the sense that the workspace creation walks up to the root the find the type.
- add bootstrap manifests
- specify initialization authorization
- allow extensibility of types in sub-workspaces
- add default workspace type for sub-workspaces created
- add allowed workspace types for sub-workspaces created
- enforce that everything an initializer controller can do in a workspace, a normal user can also do manually
- add status with virtual workspace URLs.

### DIFF
--- a/pkg/apis/tenancy/v1alpha1/types.go
+++ b/pkg/apis/tenancy/v1alpha1/types.go
@@ -170,7 +170,20 @@ type ClusterWorkspaceTypeSpec struct {
 	// created in a workspace of the given type.
 	//
 	// If a workspace type extends one at a higher level the sets of allowed sub-workspace
-	// types are merged.
+	// types are merged. Sub-workspace types that are existing in a super-workspace
+	// are ignored.
+	//
+	// For example: if Organization exists at the root, a ClusterWorkspaceType called
+	// "Organization" cannot add itself to the possible allowedSubWorkspaceTypes. I.e.
+	// the nesting must be allowed at the root and cannot be overridden. I.e. the
+	// organization admin CANNOT allow further nesting of system provided workspace
+	// types.
+	//
+	// For example: if Organization exists at the root, a ClusterWorkspaceType called
+	// "Team" which does not exist at the root at root:org, can be added to the
+	// allowedSubWorkspaceTypes of the "Organization" workspace type at root:org. I.e.
+	// the organization admin can further custom workspace types to be nested under a
+	// system provided workspace type.
 	//
 	// By default no type is allowed. This means no other workspace can be nested
 	// within a workspace of the given type.
--- a/pkg/apis/tenancy/v1alpha1/types.go
+++ b/pkg/apis/tenancy/v1alpha1/types.go
@@ -123,8 +123,8 @@ type ClusterWorkspaceTypeSpec struct {
 	// bootstrapManifests references a set of Kubernetes manifests that are applied to the
 	// workspace on initialisation.
 	//
-	// If a workspace type extends on at a higher level, the bootstrap manifests of
-	// the higher level are applied first, then those of the lower levels.
+	// If a workspace type extends a type definition of the same name at a higher level,
+	// the bootstrap manifests of the higher level are applied first, then those of the lower levels.
 	//
 	// Every bootstrap manifest created is authorized against the user creating the
 	// workspace.
@@ -140,8 +140,8 @@ type ClusterWorkspaceTypeSpec struct {
 	// be granted verb="initialize" permissions on the ClusterWorkspaceType resource.
 	// These controllers are not able to access a workspace at any other phase.
 	//
-	// If a workspace type extends one at a high level the lists of initializers
-	// are merged.
+	// If a workspace type extends a type definition of the same name at a higher level,
+	// the lists of initializers are merged.
 	//
 	// Every request done by the controller is authorized against the user creating the
 	// workspace.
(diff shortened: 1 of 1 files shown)


pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -102,21 +102,143 @@ type ClusterWorkspaceType struct {
...
+	ClusterWorkspaceCreatorGroupsLabel = "tenancy.kcp.dev/creator-groups"
+)
+
 type ClusterWorkspaceTypeSpec struct {
-	// initializers are set of a ClusterWorkspace on creation and must be
+	// bootstrapManifests references a set of Kubernetes manifests that are applied to the
+	// workspace on initialisation.
+	//
+	// If a workspace type extends on at a higher level, the bootstrap manifests of
+	// the higher level are applied first, then those of the lower levels.
```
stevekuznetsov, line 127: Do we have any learnings (perhaps from CVO?) about what sort of flexibility we need in ordering here? Do we expect this to grow to be generic manifests?

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -102,21 +102,143 @@ type ClusterWorkspaceType struct {
...
+
+	// initializers are set on a ClusterWorkspace on creation and must be
 	// cleared by a controller before the workspace can be used. The workspace
 	// will stay in the phase "Initializing" state until all initializers are cleared.
 	//
+	// A controller responsible for initializing workspaces of a given type must
+	// be granted verb="initialize" permissions on the ClusterWorkspaceType resource.
+	// These controllers are not able to access a workspace at any other phase.
+	//
+	// If a workspace type extends one at a high level the lists of initializers
```
stevekuznetsov, line 143: nit: "extends one at a high level" is a bit hard to read/grok

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -102,21 +102,143 @@ type ClusterWorkspaceType struct {
...
 	// +optional
 	AdditionalWorkspaceLabels map[string]string `json:"additionalWorkspaceLabels,omitempty"`
+
+	// defaultSubWorkspaceType is the cluster workspace type that will be used
+	// by default if another, nested ClusterWorkspace is created in a workspace
+	// of the given type. The default behaviour equals "Universal".
+	//
+	// If a workspace type extends one at a higher level and they disagree about the default
+	// the one in the sub-workspace trumps. Not specifying defaultSubWorkspaceType or
+	// an empty string means to inherit the value from the super-workspace.
```
stevekuznetsov, line 167: nit: do we care to use `*string` to disambiguate the unset/empty case?

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -102,21 +102,143 @@ type ClusterWorkspaceType struct {
...
+// workspace on initialisation.
+type BootstrapManifests struct {
+	// ConfigMaps is a list of config maps that are applied to the workspace on initialisation.
+	//
+	// +optional
+	ConfigMaps []BootstrapConfigMap `json:"configMaps,omitempty"`
+}
+
+// BootstrapConfigMap references a config map that is applied to a workspace on initialisation.
+type BootstrapConfigMap struct {
```
stevekuznetsov, line 212: Why does this not specify logical cluster?

pkg/apis/tenancy/v1alpha1/types.go
sttts: just pure creation without templating. For anything advanced we have initializers. Would try to keep out of templating business for sure.

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -102,21 +102,143 @@ type ClusterWorkspaceType struct {
...
+	// workspace.
+	//
 	// +optional
 	Initializers []ClusterWorkspaceInitializer `json:"initializers,omitempty"`
 
 	// additionalWorkspaceLabels are a set of labels that will be added to a
 	// ClusterWorkspace on creation.
 	//
+	// If a workspace type extends one at a higher level, both maps are merged. If they
+	// disagree about a value, the one in the sub-workspace trumps.
```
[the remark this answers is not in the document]
sttts: general rule of thumb of the whole API: you can add to behaviour of higher level behaviour. You cannot opt-out.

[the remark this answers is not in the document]
sttts: So "ours" is certainly not what we want. Error, maybe.

pkg/apis/tenancy/v1alpha1/types.go
sttts: same workspace.

pkg/apis/tenancy/v1alpha1/types.go
sttts: What should empty string mean?

pkg/apis/tenancy/v1alpha1/types.go
sttts: compare the new wording. Am open for something better.

pkg/apis/tenancy/v1alpha1/types.go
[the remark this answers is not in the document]
sttts: You are right. I think we should make either the higher level trump or even reject via admission. Or in the former case at least print warnings via warning headers.

PR #1024, opened 2022-05-10
Add helpers in `CommonAPIResourceSpec`

davidfestal:
## Summary

This PR add helpers to the `CommonAPIResourceSpec` to make the conversion to `CustomResource` objects easier.

This PR is based on PR #1023 to which the first commit belongs, so that only the last commit is really part of this PR.

## Related issue(s)

The changes in this PR were part of the work initially done in the context of PR #724 (issue  #487)

Issue #487, opened 2022-02-10
Build a MVP implementation of the Syncer Virtual APIServer


davidfestal:
This issue is for implementing an MVP of the Syncer Virtual APIServer, whose complete implementation is described by issue https://github.com/kcp-dev/kcp/issues/350

This MVP imlementation would mainly:
- **skip the `ExternalSync***` aspect of things, as well as generic schema negotiation
- limit the synced APIs to some identified, well-known Kube APIs ( => **no need for schema negotiation**)
- focus on:
    - providing a virtual workspace for each physical cluster syncer, that would:
        - return a **filtered view** of the resources to sync
        - in a **scalable way**
    - instruct the physical clusters to point to the corresponding virtual workspace endpoint URL.
    - possibly explore:
        - a mechanism to plug-in live, golang-based transformations (one-object-to-one-object) of objects in kcp
        - a mechanism to plug-in live, golang-based transformations that do 1:n mapping of an object in kcp to multiple objects to be synced.

This issue relates to the [TMC End-user Demo](https://github.com/kcp-dev/kcp/issues/415)

PR #1025, opened 2022-05-10
Prepare for other types of virtuall workspaces

davidfestal:
## Summary

Prepare for other types of virtuall workspaces by moving an `openapiv2` special case out of the `rootapiserver`.

This PR is based on PR #1024 to which the 2 first commits belongs, so that only the last commit is really part of this PR.

## Related issue(s)

The changes in this PR were part of the work initially done in the context of PR #724 (issue  #487)


sttts:
Lgtm when https://github.com/kcp-dev/kcp/pull/1024/files is addressed.

sttts (approved):

Issue #897, opened 2022-04-14
Make leader election work transparently


ncdc:
**Is your feature request related to a problem? Please describe.**
The client-go leader election code typically defaults to the in-cluster namespace (from `/var/run/secrets/kubernetes.io/serviceaccount/namespace` mounted into the pod). For pods in synced namespaces, the namespace name in the workload cluster (kcp<hash>) is different from the namespace name in the control plane in kcp. If the kcp<hash> namespace doesn't exist in the control plane, the leader election code will fail to create the lock resource.

**Describe the solution you'd like**
Transparently support leader election in workload clusters, somehow.

**Describe alternatives you've considered**

**Additional context**

stevekuznetsov:
As of today, all Pods running in the physical cluster get direct access to `kcp` in their in-cluster configurations, so if this does *not* work, it's a bug, right? As noted, right now the `namespace` key in the secret we mount uses DownwardAPI, but this is the namespace in the physical cluster, so in what case is that actually what we want?

/cc @jmprusi

ncdc:
It should be the namespace name from kcp

jmprusi:
@ncdc can we close this one? looks like the code is already merged into KCP https://github.com/kcp-dev/kcp/commit/4252b4fa48782930b189c54ecd1edd39a7a5ee53

ncdc:
Yes, thanks

PR #1033, opened 2022-05-11
use kcp namespace name for SA in-cluster configuration

geoberle:
## Summary
replaces the downstream namespace name provided via the downwards API with the namespace name from KCP
this is required for kube clients to pick up the correct in-cluster namespace name for API communication with KCP.

## Related issue(s)
#897

Signed-off-by: Gerd Oberlechner <goberlec@redhat.com>


stevekuznetsov:
Can we put this into #1044 since there is no test for this behavior, as it's in that PR?

geoberle:
change cherry-picked into #1044

Issue #985, opened 2022-05-04
Data race in component-base when running tests with an in-process server


sttts:
Likely related to the recent `V(2)` default.
```
WARNING: DATA RACE
I0504 17:52:23.357965   18849 controllers.go:85] Generating service account key file /var/folders/gg/zn9yfdld21j7xg5jnj5zz5d80000gn/T/tmp.tq9LeR2K/TestAPIBinding557958044/001/TestAPIBinding/2149933257/data/kcp/shared/sa.key
Write at 0x00c00027a788 by goroutine 86:
  k8s.io/component-base/logs/registry.(*LogFormatRegistry).Freeze()
      /Users/sts/Quellen/origin/pkg/mod/github.com/kcp-dev/kubernetes/staging/src/k8s.io/component-base@v0.0.0-20220502152525-1ac78681cb61/logs/registry/registry.go:109 +0x23c
  k8s.io/component-base/logs.BindLoggingFlags()
      /Users/sts/Quellen/origin/pkg/mod/github.com/kcp-dev/kubernetes/staging/src/k8s.io/component-base@v0.0.0-20220502152525-1ac78681cb61/logs/config.go:69 +0x210
  k8s.io/component-base/logs.(*Options).AddFlags()
      /Users/sts/Quellen/origin/pkg/mod/github.com/kcp-dev/kubernetes/staging/src/k8s.io/component-base@v0.0.0-20220502152525-1ac78681cb61/logs/options.go:75 +0x448
  k8s.io/kubernetes/pkg/genericcontrolplane/options.(*ServerRunOptions).Flags()
      /Users/sts/Quellen/origin/pkg/mod/github.com/kcp-dev/kubernetes@v0.0.0-20220502152525-1ac78681cb61/pkg/genericcontrolplane/options/flags.go:37 +0x3f8
  github.com/kcp-dev/kcp/pkg/server/options.(*Options).rawFlags()
      /Users/sts/Quellen/origin/src/github.com/kcp-dev/kcp/pkg/server/options/options.go:136 +0x48
  github.com/kcp-dev/kcp/pkg/server/options.(*Options).Flags()
      /Users/sts/Quellen/origin/src/github.com/kcp-dev/kcp/pkg/server/options/options.go:122 +0x44
  github.com/kcp-dev/kcp/test/e2e/framework.(*kcpServer).Run()
[...]

PR #1036, opened 2022-05-12
server: initialize logs early to avoid races

sttts:
In the upstream fork, https://github.com/kcp-dev/kubernetes/commit/a525fc348b7ad73f634904aca0a55f842f0ef5dc removes the late klog initialization, and this PR here moves it to our kcp server Cobra command. This is the same that upstream does, e.g. in https://github.com/kubernetes/kubernetes/blob/edac6fce2ad4a1366a73afc8ac9ca50b2fbf7fc7/cmd/kube-apiserver/app/server.go#L112.

Fixes https://github.com/kcp-dev/kcp/issues/985.


cmd/kcp/kcp.go
```diff
@@ -97,6 +97,11 @@ func main() {
 				return errors.NewAggregate(errs)
 			}
 
+			// rnu as early as possible to avoid races later when some components (e.g. grpc) start early using klog
```
qiujian16, line 100: typo: run?

cmd/kcp/kcp.go
```diff
@@ -97,6 +97,11 @@ func main() {
...
+			if err := completed.GenericControlPlane.Logs.ValidateAndApply(); err != nil {
+				return err
```
s-urbaniak, line 102: i see we don't use the pattern of `return fmt.Errorf("error validating and applying generic control plane logs: %w, err)`. I am a bit worried this will not expose the full stack trace picture and logs and will make debugging harder? (not related to this PR as I see the same a few lines below).

s-urbaniak (approved):
one comment regarding general error handling, unrelated to this PR, hence approving.

cmd/kcp/kcp.go
sttts: Validation errors are usually pretty good with context. Don't think we need this here.

s-urbaniak (approved):

cmd/kcp/kcp.go
sttts: fixed

s-urbaniak (approved):

Issue #862, opened 2022-04-07
Add logical cluster annotation to audit events


kylape:
**Is your feature request related to a problem? Please describe.**
Currently audit events don't have any context regarding logic cluster.  This should be added to help audit events be more useful.

**Describe the solution you'd like**
[Add an annotation to audit events that includes the logical cluster](https://kubernetes.slack.com/archives/C021U8WSAFK/p1649358931928699?thread_ts=1649358177.770489&cid=C021U8WSAFK).

> we have `func AddAuditAnnotation(ctx context.Context, key, value string)`, so we can add an annotation via `pkg/server/handler.go` in kcp

geoberle:
i would like to grab this one

ibihim:
Sorry, I reacted in the slack chat and forgot to mention that I am working on it too.

geoberle:
@ibihim no problem. go for it

ibihim:
thank you :pray:

sttts:
Merged

PR #1042, opened 2022-05-12
pkg/server: add workspace annotation

ibihim:
Signed-off-by: Krzysztof Ostrowski <kostrows@redhat.com>

## Summary

Annotate audit logs with cluster name set in request context.

## Related issue(s)

https://github.com/kcp-dev/kcp/issues/862

Fixes #

### DIFF
--- a/pkg/server/handler.go
+++ b/pkg/server/handler.go
@@ -40,6 +40,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	utilnet "k8s.io/apimachinery/pkg/util/net"
 	"k8s.io/apimachinery/pkg/util/sets"
+	kaudit "k8s.io/apiserver/pkg/audit"
 	"k8s.io/apiserver/pkg/authentication/authenticator"
 	authserviceaccount "k8s.io/apiserver/pkg/authentication/serviceaccount"
 	apiserverdiscovery "k8s.io/apiserver/pkg/endpoints/discovery"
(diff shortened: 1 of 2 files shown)


pkg/server/handler.go
```diff
@@ -20,6 +20,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/emicklei/go-restful"
+	"github.com/kcp-dev/logicalcluster"
```
sttts, line 24: order

pkg/server/server.go
```diff
@@ -224,6 +224,7 @@ func (s *Server) Run(ctx context.Context) error {
 		apiHandler = WithClusterScope(apiHandler)
 		apiHandler = WithInClusterServiceAccountRequestRewrite(apiHandler, unsafeServiceAccountPreAuth)
 		apiHandler = WithAcceptHeader(apiHandler)
+		apiHandler = WithClusterAnnotation(apiHandler)
```
sttts, line 227: now your code runs too early. It's like shells, called from the outside to inside.

ibihim:
Do we have already auditing in KCP?

I created a PR 1042 - add workspace annotation for issue 862 - add logical cluster annotation to audit events, under the assumption, that it exists already. Otherwise we would need to add filters.WithAuditAnnoations and audit options, right?

sttts:
Looks good now. A small e2e would be nice. Like enable the audit file in the generic e2e server flag config, and check that some request arrived with the workspace set.

sttts (approved):

sttts:
e2e as follow-up.

Issue #846, opened 2022-04-06
APIBindings: delete CRs when the APIBinding is deleted


sttts:
Follow-up of #417.

When an APIBinding is deleted, there is no controller deleting the CRs. The resource will just disappear and leave dead data in etcd.

thetechnick:
I have had a quick dive through the code to see how to implement this, my rough idea would be:
- Ensure a finalizer on the APIBinding object. `finalizers.kc.dev/gc` Is there prior art for the finalizer name?
- Ensure a backref from CRDs to it's "Owning" APIBinding, so we can get it from an indexer for deletion, even when the APIExport object is no longer present
- Add a new reconciler acting only on objects with `.DeletionTimestamp` set, deleting all CRDs belonging to the deleted APIBinding

Unrelated to this bug, there might also be another cleanup issue.
When `apiExport.Spec.LatestResourceSchemas` changes to a smaller list, previously created CRDs are not cleaned up - as far as I have seen - need to play a bit more with KCP to be sure :)

PR #1040, opened 2022-05-12
Delete custom resources when apibinding is deleted

qiujian16:
Signed-off-by: Jian Qiu <jqiu@redhat.com>

## Summary

add controller to delete all CRs with apibinding is deleting

followup: avoid CR creation when apibinding is deleting

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/846

PR #1043, opened 2022-05-13
forbid resource creation when APIBinding is deleting

qiujian16:
## Summary

follow up PR of https://github.com/kcp-dev/kcp/pull/1040
## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/846

### DIFF
--- a/pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
+++ b/pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
@@ -178,7 +178,11 @@ func (c *Controller) process(ctx context.Context, key string) error {
 			return nil
 		}
 
-		return c.patchFinalizer(ctx, workspace, append(workspaceCopy.Finalizers, deletion.WorkspaceFinalizer))
+		workspaceCopy.Finalizers = append(workspaceCopy.Finalizers, deletion.WorkspaceFinalizer)
+		_, err := c.kcpClient.Cluster(logicalcluster.From(workspaceCopy)).TenancyV1alpha1().ClusterWorkspaces().Update(
+			ctx, workspaceCopy, metav1.UpdateOptions{})
+
+		return err
 	}
 
 	err = c.deleter.Delete(ctx, workspaceCopy)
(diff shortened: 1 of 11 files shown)


pkg/server/apiextensions.go
```diff
@@ -382,6 +382,31 @@ func shallowCopyCRD(in *apiextensionsv1.CustomResourceDefinition) *apiextensions
...
+func decorateCRDWithBinding(in *apiextensionsv1.CustomResourceDefinition, identity string, deleteTime *metav1.Time) *apiextensionsv1.CustomResourceDefinition {
+	out := shallowCopyCRD(in)
+
+	out.Annotations[apisv1alpha1.AnnotationAPIIdentityKey] = identity
+
+	if deleteTime.IsZero() {
+		return out
+	}
+
+	out.Status.Conditions = make([]apiextensionsv1.CustomResourceDefinitionCondition, len(in.Status.Conditions), len(in.Status.Conditions)+1)
```
stevekuznetsov, line 397: The slice does not need to grow if there's already a terminating condition, right?

pkg/server/apiextensions.go
```diff
@@ -382,6 +382,31 @@ func shallowCopyCRD(in *apiextensionsv1.CustomResourceDefinition) *apiextensions
...
+
+	out.Status.Conditions = append(out.Status.Conditions, in.Status.Conditions...)
+
+	out.DeletionTimestamp = deleteTime.DeepCopy()
+	apiextensionshelpers.SetCRDCondition(out, apiextensionsv1.CustomResourceDefinitionCondition{
```
stevekuznetsov, line 402: This will take care of appending a new member to the slice if needed, right?

stevekuznetsov (commented):
Are we happy with the racy implication of requiring APIBinding deletion to propagate to another object in order to stop creation?

pkg/server/apiextensions.go
sttts, line 397: these conditions here are not visible to anybody, are they? They are only faked while feeding the apiextensions-apiserver CR registry.

sttts: nvmd. The Terminating condition is used by apiextensions. Odd. Maybe add a comment to clarify that. Using a condition always feel odd to me for any kind of critical logic.

pkg/server/apiextensions.go
qiujian16: removed the setting of capacity.

stevekuznetsov:
@sttts for CRDs? If it is, seems reasonable to me :)

sttts:
In the sense that it is informer based in kube. So it can race.

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
```diff
@@ -242,6 +229,28 @@ func (c *Controller) patchCondition(ctx context.Context, old, new *tenancyv1alph
...
+	}
+
+	patch, err := json.Marshal(finalizerData)
+	if err != nil {
+		return err
+	}
+
+	// remove finalizers field if there is no remaining finalizers,
+	if len(finalizers) == 0 {
+		patch = []byte("{\"metadata\": {\"finalizers\": []}}")
```
sttts, line 246: not sure this is safe. Removing the field is not really necessary. Just merge-patch your finalizer out. This patch here has the potential to override other finalizers just added. 

same for the same func above.

qiujian16: added uid/rv as precond in this case. The finalizer is opmitempty, so I think we have to build this json string explicitly when finalizer is empty.

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
```diff
@@ -242,6 +229,31 @@ func (c *Controller) patchCondition(ctx context.Context, old, new *tenancyv1alph
...
+
+	patch, err := json.Marshal(finalizerData)
+	if err != nil {
+		return err
+	}
+
+	// remove finalizers field if there is no remaining finalizers,
+	if len(finalizers) == 0 {
+		patch = []byte(
+			fmt.Sprintf("{\"metadata\": {\"finalizers\": [], \"uid\": %q, \"resourceVersion\": %q}}", ws.UID, ws.ResourceVersion))
```
sttts, line 249: At this point it's much easier to reason about a simple Update call with `ws` from above with the changes directly in the object.

qiujian16: updated

sttts (approved):

PR #1046, opened 2022-05-13
Recreate downstream resource

itdove:
Signed-off-by: Dominique Vernier <dvernier@redhat.com>

## Summary

Add a downstream watcher in the spec_controller in order to recreate a resource when it gets deleted

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/604

Currently blocked with e2e failing when we're using a workspace as a fake cluster. #1066 should hopefully allows us to unblock


pkg/cliplugins/workload/plugin/syncer.yaml
```diff
@@ -25,7 +25,9 @@ rules:
   verbs:
   - "create"
   - "list"
+  - "get"
   - "watch"
+  - "patch"
```
stevekuznetsov, line 30: Why do we need `patch` on `Namespace`?

pkg/syncer/spec/spec_controller.go
```diff
@@ -96,7 +99,56 @@ func NewSpecSyncer(gvrs []schema.GroupVersionResource, upstreamClusterName logic
...
+				if err != nil {
+					klog.Error(err)
+					return
+				}
+				if ns == nil {
+					klog.Error(fmt.Errorf("namespace %s is nil", unstrob.GetNamespace()))
+					return
+				}
+				//If no annotations we can not find the logical-cluster name
+				if ns.GetAnnotations() == nil {
```
stevekuznetsov, line 128: You don't need to make this check, indexing into a `nil` map is fine: https://go.dev/play/p/JYLjZKFE2qr

pkg/syncer/spec/spec_controller.go
```diff
@@ -96,7 +99,56 @@ func NewSpecSyncer(gvrs []schema.GroupVersionResource, upstreamClusterName logic
...
 		})
-		klog.InfoS("Set up informer", "clusterName", upstreamClusterName, "pcluster", workloadClusterName, "gvr", gvr.String())
+		klog.InfoS("Set up upstream informer", "clusterName", upstreamClusterName, "pcluster", workloadClusterName, "gvr", gvr.String())
+
+		downstreamInformers.ForResource(gvr).Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+			DeleteFunc: func(obj interface{}) {
+				unstrob := obj.(*unstructured.Unstructured)
+				klog.InfoS("downstream delete event", "gvr", gvr, "namespace", unstrob.GetNamespace(), "name", unstrob.GetName())
+				//if there is no label then it is not owned by kcp
+				if unstrob.GetLabels() == nil {
```
stevekuznetsov, line 109: This is enforced by the list options going into the informer, so why do we need this check? Also, check the key-value, not for the existence of labels in general.

pkg/syncer/spec/spec_controller.go
```diff
@@ -96,7 +99,56 @@ func NewSpecSyncer(gvrs []schema.GroupVersionResource, upstreamClusterName logic
...
+					Group:    "",
+					Version:  "v1",
+					Resource: "namespaces",
+				})
+				ns, err := namespaces.Get(context.TODO(), unstrob.GetNamespace(), metav1.GetOptions{})
+				if err != nil {
+					klog.Error(err)
+					return
+				}
+				if ns == nil {
```
ncdc, line 123: We typically don't check that the object is non-nil if err comes back nil. No harm in doing this, just not something you'll usually see in the code.

pkg/syncer/spec/spec_controller.go
```diff
@@ -96,7 +99,56 @@ func NewSpecSyncer(gvrs []schema.GroupVersionResource, upstreamClusterName logic
 				c.AddToQueue(gvr, obj)
 			},
 		})
-		klog.InfoS("Set up informer", "clusterName", upstreamClusterName, "pcluster", workloadClusterName, "gvr", gvr.String())
+		klog.InfoS("Set up upstream informer", "clusterName", upstreamClusterName, "pcluster", workloadClusterName, "gvr", gvr.String())
+
+		downstreamInformers.ForResource(gvr).Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+			DeleteFunc: func(obj interface{}) {
```
stevekuznetsov, line 105: Likely we want a queue for this, or we are edge-driven and transient failures leave in a broken state forever

pkg/cliplugins/workload/plugin/sync_test.go
```diff
@@ -50,7 +50,9 @@ rules:
   verbs:
   - "create"
   - "list"
+  - "get"
   - "watch"
+  - "patch"
```
ncdc, line 55: We can remove this?

pkg/syncer/spec/spec_controller.go
```diff
@@ -96,7 +99,56 @@ func NewSpecSyncer(gvrs []schema.GroupVersionResource, upstreamClusterName logic
...
+				if unstrob.GetLabels() == nil {
+					return
+				}
+				//Use namespace lister
+				namespaces := downstreamClient.Resource(schema.GroupVersionResource{
+					Group:    "",
+					Version:  "v1",
+					Resource: "namespaces",
+				})
+				ns, err := namespaces.Get(context.TODO(), unstrob.GetNamespace(), metav1.GetOptions{})
```
stevekuznetsov, line 118: Please use the appropriate context, not `context.TODO()`

pkg/syncer/spec/spec_controller.go
```diff
@@ -96,7 +99,56 @@ func NewSpecSyncer(gvrs []schema.GroupVersionResource, upstreamClusterName logic
...
+				}
+				m := &metav1.ObjectMeta{
+					ClusterName: nsLocator.LogicalCluster.String(),
+					Namespace:   nsLocator.Namespace,
+					Name:        unstrob.GetName(),
+				}
+				c.AddToQueue(gvr, m)
+			},
+		})
+		klog.InfoS("Set up downstream informer", "clusterName", upstreamClusterName, "pcluster", workloadClusterName, "gvr", gvr.String())
```
ncdc, line 151: V(2)

pkg/syncer/spec/spec_controller.go
stevekuznetsov, line 123: This case is only true if `err != nil`, so this code is unreachable

pkg/syncer/spec/spec_process.go
```diff
@@ -222,6 +222,9 @@ func (c *Controller) ensureSyncerFinalizer(ctx context.Context, gvr schema.Group
 }
 
 func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVersionResource, downstreamNamespace string, upstreamObj *unstructured.Unstructured) error {
+	if gvr.Group == "" && gvr.Version == "v1" && gvr.Resource == "namespaces" {
```
ncdc, line 225: Let's add a comment why we're doing this

stevekuznetsov, line 225: ?

test/e2e/syncer/syncer_test.go
```diff
@@ -170,6 +170,30 @@ func TestSyncerLifecycle(t *testing.T) {
...
+			t.Errorf("saw an error waiting for downstream deployment %s/%s to be created: %v", downstreamNamespaceName, upstreamDeployment.Name, err)
+		}
+		//Test UID
+		if deployment.UID == newDeployment.UID {
+			return false
+		}
+		//Set the deployment to the new created one
+		deployment = newDeployment
+		return true
+	}, wait.ForeverTestTimeout, time.Second*10, "downstream deployment %s/%s was not synced", downstreamNamespaceName, upstreamDeployment.Name)
```
ncdc, line 195: Please change the poll interval from 10 seconds to 100 milliseconds. We want the tests to finish as quickly as possible.

stevekuznetsov, line 195: 10s is a very long poll. 100ms please

test/e2e/syncer/syncer_test.go
```diff
@@ -170,6 +170,30 @@ func TestSyncerLifecycle(t *testing.T) {
...
+		}
+		if err != nil {
+			t.Errorf("saw an error waiting for downstream deployment %s/%s to be created: %v", downstreamNamespaceName, upstreamDeployment.Name, err)
+		}
+		//Test UID
+		if deployment.UID == newDeployment.UID {
+			return false
+		}
+		//Set the deployment to the new created one
+		deployment = newDeployment
```
ncdc, line 193: Do you need to do this?

pkg/cliplugins/workload/plugin/syncer.yaml
itdove: This was missing as line  https://github.com/kcp-dev/kcp/blob/8b834d3e9055202f0f8a246dbc63f36fb892a6fa/pkg/syncer/spec/spec_process.go#L327 is doing a patch and an error land in the log:
```
E0511 15:10:29.940950       1 spec_process.go:328] Error upserting namespaces kcpf202fbc0d7de063cb1d7e373943c34f7d24eae3fbb01edff25395c17/default from upstream root:default:my-ws|/default: namespaces "default" is forbidden: User "system:serviceaccount:kcpsyncbe03c3b5a346b42da8ca24903ad985e493d1cb4f0ed17667c76a6733:kcp-syncer" cannot patch resource "namespaces" in API group "" in the namespace "kcpf202fbc0d7de063cb1d7e373943c34f7d24eae3fbb01edff25395c17"
E0511 15:10:29.941149       1 spec_controller.go:159] kcp-workload-syncer-spec failed to sync {{"" "v1" "namespaces"} "root:default:my-ws#$#default"}, err: namespaces "default" is forbidden: User "system:serviceaccount:kcpsyncbe03c3b5a346b42da8ca24903ad985e493d1cb4f0ed17667c76a6733:kcp-syncer" cannot patch resource "namespaces" in API group "" in the namespace "kcpf202fbc0d7de063cb1d7e373943c34f7d24eae3fbb01edff25395c17"
```

pkg/syncer/spec/spec_controller.go
ncdc: A queue for what, beyond the once this code is adding to?

pkg/syncer/spec/spec_controller.go
itdove: ok

pkg/cliplugins/workload/plugin/syncer.yaml
ncdc: You should be safe to remove it now

pkg/syncer/spec/spec_controller.go
itdove: yes, I had to add due to another bug I had and yes now I can remove it.

pkg/syncer/spec/spec_process.go
itdove: @ncdc suggested me to add that line as there is another bug somewhere in the code that needs to be fixed.

test/e2e/syncer/syncer_test.go
itdove: 100ms is not enough, I will try to shorten it

pkg/cliplugins/workload/plugin/syncer.yaml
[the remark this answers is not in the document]
itdove: why?

pkg/cliplugins/workload/plugin/syncer.yaml
```diff
@@ -25,7 +25,9 @@ rules:
   verbs:
   - "create"
   - "list"
+  - "get"
```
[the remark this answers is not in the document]
itdove: @stevekuznetsov suggested me to use live client as using a lister will be too costly (keep all ns in cache) 
to resolve an edge case (downstream deletion).

pkg/syncer/spec/spec_controller.go
```diff
@@ -96,7 +99,56 @@ func NewSpecSyncer(gvrs []schema.GroupVersionResource, upstreamClusterName logic
 				c.AddToQueue(gvr, obj)
 			},
 		})
-		klog.InfoS("Set up informer", "clusterName", upstreamClusterName, "pcluster", workloadClusterName, "gvr", gvr.String())
+		klog.InfoS("Set up upstream informer", "clusterName", upstreamClusterName, "pcluster", workloadClusterName, "gvr", gvr.String())
+
+		downstreamInformers.ForResource(gvr).Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+			DeleteFunc: func(obj interface{}) {
+				unstrob := obj.(*unstructured.Unstructured)
+				klog.InfoS("downstream delete event", "gvr", gvr, "namespace", unstrob.GetNamespace(), "name", unstrob.GetName())
```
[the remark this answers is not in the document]
itdove: OK I can do but:
1) didn't see much V(2) nor V().
2) tried to activate using `-- -v=3` in the deployment yaml but not working.

pkg/syncer/spec/spec_process.go
itdove: Please give me the comment to add as I don't know that code and why we have to do that.

test/e2e/syncer/syncer_test.go
itdove: As mentioned above 100ms is not enough, I will try to reduce it.

ncdc: Please use 100ms. This is the polling interval, not the timeout.

test/e2e/syncer/syncer_test.go
itdove: To realign deployment var with the latest deployment in case if someone else wants to add new test in the same test so the deployment var is up to date.

pkg/cliplugins/workload/plugin/syncer.yaml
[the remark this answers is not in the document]
ncdc: The syncer is not patching namespaces any more

test/e2e/syncer/syncer_test.go
ncdc: We can do that in a future PR if needed. I'd remove it from this one.

test/e2e/syncer/syncer_test.go
itdove: 300ms seems ok

pkg/syncer/spec/spec_controller.go
stevekuznetsov: For deleted keys - so we can ensure they end up in the queue this is adding to? What do we want to allow in case some of the intermediate logic here fails and we can't queue? With the current setup we'd just sit around until a relist with no object on the downstream cluster, right?

test/e2e/syncer/syncer_test.go
ncdc: If 100ms is not working, something is wrong inside the eventually block

pkg/syncer/spec/spec_controller.go
```diff
@@ -96,7 +99,56 @@ func NewSpecSyncer(gvrs []schema.GroupVersionResource, upstreamClusterName logic
...
+				//if there is no label then it is not owned by kcp
+				if unstrob.GetLabels() == nil {
+					return
+				}
+				//Use namespace lister
```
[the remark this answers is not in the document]
stevekuznetsov: I was mostly saying - users deleting things seems rare, why not wait to index in memory until we know we need the optimization?

pkg/syncer/spec/spec_controller.go
stevekuznetsov: Refactor the method to have one? `context.TODO` should have no place in merged production code :)

pkg/syncer/spec/spec_controller.go
ncdc: So when a downstream deployment is deleted, you want to stick the key for it into a queue?

If the syncer was offline while the downstream deployment was deleted, when it restarts, the normal upstream->downstream logic will kick in and upsert the deployment into downstream.

pkg/cliplugins/workload/plugin/syncer.yaml
itdove: See comment below.... seems as we add the `if` statement on namespace, the patch is not needed anymore but there is maybe a bug somewhere else that need to be resolved and then the `if` statement would not be needed.

pkg/syncer/spec/spec_controller.go
itdove: Changed also the upstream message than,

pkg/syncer/spec/spec_controller.go
ncdc: Not possible. This is an informer event handler

test/e2e/syncer/syncer_test.go
```diff
@@ -170,6 +170,29 @@ func TestSyncerLifecycle(t *testing.T) {
...
+		if err != nil {
+			t.Errorf("saw an error waiting for downstream deployment %s/%s to be created: %v", downstreamNamespaceName, upstreamDeployment.Name, err)
+		}
+		//Test UID
+		if deployment.UID == newDeployment.UID {
+			return false
+		}
+		//Set the deployment to the new created one
+		return true
+	}, wait.ForeverTestTimeout, time.Millisecond*300, "downstream deployment %s/%s was not synced", downstreamNamespaceName, upstreamDeployment.Name)
```
ncdc, line 194: This works fine for me at 100ms. What problems are you running into?

pkg/syncer/spec/spec_controller.go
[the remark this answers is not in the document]
ncdc: The dynamic informer is already caching namespaces, so let's switch to using that here. We can revisit if we want to change that behavior separately.

If you all agree, I'd suggest getting a reference to the dynamic namespace lister outside of the gvrs for loop. 

```go
namespaceGVR := schema.GroupVersionResource{
  Group:    "",
  Version:  "v1",
  Resource: "namespaces",
}

namespaceLister := downstreamInformers.ForResource(namespaceGVR).Lister()
```

Then you can do `namespaceLister.Get()` here.

pkg/syncer/spec/spec_controller.go
ncdc: It'll go away if we switch to the lister 😛

pkg/syncer/spec/spec_controller.go
```diff
@@ -96,7 +99,43 @@ func NewSpecSyncer(gvrs []schema.GroupVersionResource, upstreamClusterName logic
 				c.AddToQueue(gvr, obj)
 			},
 		})
-		klog.InfoS("Set up informer", "clusterName", upstreamClusterName, "pcluster", workloadClusterName, "gvr", gvr.String())
+		klog.V(2).InfoS("Set up upstream informer", "clusterName", upstreamClusterName, "pcluster", workloadClusterName, "gvr", gvr.String())
+
+		downstreamInformers.ForResource(gvr).Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+			DeleteFunc: func(obj interface{}) {
+				unstrob := obj.(*unstructured.Unstructured)
```
ncdc, line 106: Sorry I missed adding a comment about checking for a tombstone here (this is for ingress, but just change that to Unstructured):

```go
ingress, ok := obj.(*networkingv1.Ingress)
			if !ok {
				tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
				if !ok {
					runtime.HandleError(fmt.Errorf("unexpected object type: %T", obj))
					return
				}

				ingress, ok = tombstone.Obj.(*networkingv1.Ingress)
				if !ok {
					runtime.HandleError(fmt.Errorf("unexpected object type: %T", obj))
					return
				}
			}
```

pkg/syncer/spec/spec_controller.go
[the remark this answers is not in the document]
ncdc: Talked to Steve on Meet. He's ok with using the lister

pkg/syncer/spec/spec_controller.go
```diff
@@ -96,7 +99,47 @@ func NewSpecSyncer(gvrs []schema.GroupVersionResource, upstreamClusterName logic
 				c.AddToQueue(gvr, obj)
 			},
 		})
-		klog.InfoS("Set up informer", "clusterName", upstreamClusterName, "pcluster", workloadClusterName, "gvr", gvr.String())
+		klog.V(2).InfoS("Set up upstream informer", "clusterName", upstreamClusterName, "pcluster", workloadClusterName, "gvr", gvr.String())
+
+		downstreamInformers.ForResource(gvr).Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+			DeleteFunc: func(obj interface{}) {
+				unstrob, ok := obj.(*unstructured.Unstructured)
```
sttts, line 106: this is not good. I did the same mistake here: https://github.com/kcp-dev/kcp/pull/1049. 
Background: if the informer restarts in just this moment, it can be that no old object is return, but some tombstone object. You have to use the key helper like in upper PR.

pkg/syncer/spec/spec_controller.go
```diff
@@ -96,7 +99,47 @@ func NewSpecSyncer(gvrs []schema.GroupVersionResource, upstreamClusterName logic
...
+				if !ok {
+					runtime.HandleError(fmt.Errorf("unexpected object type: %T", obj))
+					return
+				}
+				klog.V(2).InfoS("downstream delete event", "gvr", gvr, "namespace", unstrob.GetNamespace(), "name", unstrob.GetName())
```
sttts, line 111: Our convention is more like V(3), and add a V(2) with the key above the enqueue. Compare https://github.com/kcp-dev/kcp/blob/main/CONTRIBUTING.md.

pkg/syncer/spec/spec_controller.go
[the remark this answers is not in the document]
sttts: +1, definitely lister. We cannot do live-gets for every deleted object.

pkg/syncer/spec/spec_controller.go
sttts: Why exactly do we user any lister or client at all? Why not just work completely with keys. It is shorter, less error-phrone, and it matches the logic of the status syncer.

test/e2e/syncer/syncer_test.go
```diff
@@ -170,6 +170,28 @@ func TestSyncerLifecycle(t *testing.T) {
 	}
 
 	// TODO(marun) Check that the deployment was able to contact kcp
+
+	//Delete the deployment
```
sttts, line 174: spaces after // please (convention)

pkg/syncer/spec/spec_controller.go
ncdc: @sttts see https://github.com/kcp-dev/kcp/pull/1046#discussion_r872716552

test/e2e/syncer/syncer_test.go
```diff
@@ -170,6 +170,28 @@ func TestSyncerLifecycle(t *testing.T) {
...
+	require.Eventually(t, func() bool {
+		newDeployment, err := downstreamKubeClient.AppsV1().Deployments(downstreamNamespaceName).Get(ctx, upstreamDeployment.Name, metav1.GetOptions{})
+		if apierrors.IsNotFound(err) {
+			return false
+		}
+		if err != nil {
+			t.Errorf("saw an error waiting for downstream deployment %s/%s to be created: %v", downstreamNamespaceName, upstreamDeployment.Name, err)
+		}
+		//Test UID
+		if deployment.UID == newDeployment.UID {
```
sttts, line 189: why this check?

stevekuznetsov: In Slack I had suggested that semantic equality + unequal UID would be a rigorous way to check for recreation. IDK if we need it.

pkg/syncer/spec/spec_controller.go
ncdc: We need the ns

test/e2e/syncer/syncer_test.go
ncdc: Make sure it was deleted & recreated

pkg/syncer/spec/spec_controller.go
ncdc: Talked to Steve offline. Any error in this delete handler is unlikely (type assertions should succeed, namespaces should exist, etc.). If we were to add the key to a separate queue, we would almost certainly hit the exact same error(s) again, i.e. permanent fail.

ncdc:
Updated description with current status (blocked)

ncdc (approved):

PR #1059, opened 2022-05-16
virtualworkspace: fix delegated UPDATE semantics

stevekuznetsov:
When a user does an UPDATE while specifying a resourceVersion, we
absolutely need to be using that information in our delegated GET calls.

/assign @sttts @ncdc

### DIFF
--- a/pkg/virtual/framework/forwardingregistry/store.go
+++ b/pkg/virtual/framework/forwardingregistry/store.go
@@ -179,10 +179,36 @@ func (s *Store) Update(ctx context.Context, name string, objInfo rest.UpdatedObj
 		return nil, false, err
 	}
 
+	var resourceVersion *string
+	var resourceVersionSpecified bool
+	if preconditions := objInfo.Preconditions(); preconditions != nil {
+		if preconditions.ResourceVersion != nil {
+			resourceVersion = preconditions.ResourceVersion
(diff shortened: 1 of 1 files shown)


sttts:
❤️

sttts:
After a test as Andy says, lgtm.

stevekuznetsov:
Looks like we pass the option down correctly to the delegated call even without this patch, so the failure is there anyway. I don't really want to think through the exact flow where this gets weird. Just to confirm - we are translating all Patch requests to the VW into UPDATE for the delegate, right?

sttts:
> we are translating all Patch requests to the VW into UPDATE for the delegate, right?

yes

stevekuznetsov:
/shrug

Did we do some survey to make sure that is correct in all cases? Seems like there might be some weirdness with it.

ncdc:
Update can lead to data loss in certain situations:

1. Old client has old version of go struct
2. New field is added to the API
3. Instances in etcd have the new field
4. Old client gets an instance, makes some change, issues an update
5. New field is silently removed, because update is a wholesale replacement of the contents

sttts:
This is using a dynamic client, doesn't it? No risk to lose anything, even with update.

sttts:
New test failed again, have seen it at least twice in a row. Maybe it's real.

PR #1064, opened 2022-05-16
PROOF: Syncer virtual workspace transformations and strategies

davidfestal:


pkg/virtual/syncer/strategies/apis.go
```diff
@@ -0,0 +1,54 @@
+/*
+Copyright 2021 The KCP Authors.
```
sttts, line 2: 2022

PR #1018, opened 2022-05-10
Adds virtualworkspaceurls-controller

jmprusi:
## Summary

This PR adds a new controller (virtualworkspaceurls) that populates a new status field in the WorkloadCluster resource: `syncerVirtualWorkspaceURLs`.

The SyncerVirtualWorkspaceURLs are built by concatenating several fields:

"http://external-host/services/syncer/demo:root:yourworkspace/test-cluster"
  
  * '`http://external-host`': ClusterWorkspaceShard external-url
  * '`/services`': virtualworkspacesoptions.DefaultRootPathPrefix
  * '`/syncer`': synceroptions.SyncerVirtualWorkspaceName
  * '`demo:root:yourworkspace`': WorkloadCluster ClusterName.
  * '`/test-cluster`': WorkloadCluster Name.

The controller will look for changes in the ClusterWorkspaceShards and WorkloadClusters resource in all KCP workspaces.

## Related issue(s)

https://github.com/kcp-dev/kcp/issues/1012

PR #1067, opened 2022-05-17
Wire syncer to the virtual workspace

davidfestal:
## Summary

This PR builds on top of PR https://github.com/kcp-dev/kcp/pull/1018 and wires the syncer to the syncer virtual workspace.

This is a minimal implementation that reads only the first VirtualWorkspace URL at start.

When tackling sharding, we would have to revisit this to create clients / informers (or even SpecSyncer / StatusSyncer)  for each virtual workspace URL (and maintain them if the list of URLs changes)

### DIFF
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -89,9 +89,8 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 	// TODO(david): we need to provide user-facing details if this polling goes on forever. Blocking here is a bad UX.
 	// TODO(david): Also, any regressions in our code will make any e2e test that starts a syncer (at least in-process)
 	// TODO(david): block until it hits the 10 minute overall test timeout.
-	err = wait.PollImmediateInfinite(gvrQueryInterval, func() (bool, error) {
-		klog.Infof("Attempting to retrieve the Syncer virtual workspace URL from WorkloadCluster %s|%s", cfg.KCPClusterName, cfg.WorkloadClusterName)
-
+	klog.Infof("Attempting to retrieve the Syncer virtual workspace URL from WorkloadCluster %s|%s", cfg.KCPClusterName, cfg.WorkloadClusterName)
+	err = wait.PollImmediateInfinite(5*time.Second, func() (bool, error) {
 		workloadCluster, err := kcpClusterClient.Cluster(cfg.KCPClusterName).WorkloadV1alpha1().WorkloadClusters().Get(ctx, cfg.WorkloadClusterName, metav1.GetOptions{})
 		if err != nil {
 			return false, err
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/pkg/version"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 
@@ -73,15 +74,21 @@ func (sc *SyncerConfig) ID() string {
 func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, importPollInterval time.Duration) error {
 	klog.Infof("Starting syncer for logical-cluster: %s, workload-cluster: %s", cfg.KCPClusterName, cfg.WorkloadClusterName)
 
-	kcpClusterClient, err := kcpclient.NewClusterForConfig(rest.AddUserAgent(rest.CopyConfig(cfg.UpstreamConfig), "kcp#syncer/v0.0.0"))
+	kcpVersion := version.Get().GitVersion
+
+	kcpClusterClient, err := kcpclient.NewClusterForConfig(rest.AddUserAgent(rest.CopyConfig(cfg.UpstreamConfig), "kcp#syncer/"+kcpVersion))
 	if err != nil {
 		return err
 	}
 
-	// For now we retrieve the syncerVirtualWorkpaceURL at start, since we temporarily stick to a single URL (sharding not supported).
-	// But the complete implementation should setup a WorkloadCluster informer, create spec and status syncer for every URLs found in the
-	// Status.SyncerVirtualWorkspaceURLs slice, and update them each time this list changes.
+	// TODO(david): Implement real support for several virtual workspace URLs that can change over time.
+	// TODO(david): For now we retrieve the syncerVirtualWorkpaceURL at start, since we temporarily stick to a single URL (sharding not supported).
+	// TODO(david): But the complete implementation should setup a WorkloadCluster informer, create spec and status syncer for every URLs found in the
+	// TODO(david): Status.SyncerVirtualWorkspaceURLs slice, and update them each time this list changes.
 	var syncerVirtualWorkspaceURL string
+	// TODO(david): we need to provide user-facing details if this polling goes on forever. Blocking here is a bad UX.
+	// TODO(david): Also, any regressions in our code will make any e2e test that starts a syncer (at least in-process)
+	// TODO(david): block until it hits the 10 minute overall test timeout.
 	err = wait.PollImmediateInfinite(gvrQueryInterval, func() (bool, error) {
 		klog.Infof("Attempting to retrieve the Syncer virtual workspace URL from WorkloadCluster %s|%s", cfg.KCPClusterName, cfg.WorkloadClusterName)
 
@@ -101,7 +108,6 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 		return true, nil
 	})
 	if err != nil {
-		// Should never happen
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/syncer/syncer.go
```diff
@@ -73,15 +73,58 @@ func (sc *SyncerConfig) ID() string {
 func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, importPollInterval time.Duration) error {
 	klog.Infof("Starting syncer for logical-cluster: %s, workload-cluster: %s", cfg.KCPClusterName, cfg.WorkloadClusterName)
 
+	kcpClusterClient, err := kcpclient.NewClusterForConfig(rest.AddUserAgent(rest.CopyConfig(cfg.UpstreamConfig), "kcp#syncer/v0.0.0"))
```
ncdc, line 76: Either TODO to fill in the version, or pull it from either `k8s.io/component-base/version` or `k8s.io/client-go/pkg/version`?

pkg/syncer/syncer.go
```diff
@@ -73,15 +73,58 @@ func (sc *SyncerConfig) ID() string {
...
+	if err != nil {
+		return err
+	}
+
+	// For now we retrieve the syncerVirtualWorkpaceURL at start, since we temporarily stick to a single URL (sharding not supported).
```
ncdc, line 81: Please make sure this comment has `TODO` in there somewhere - thanks!

pkg/syncer/syncer.go
```diff
@@ -73,15 +73,58 @@ func (sc *SyncerConfig) ID() string {
...
+	// workspace.
+	apiImporter, err := NewAPIImporter(cfg.UpstreamConfig, cfg.DownstreamConfig, resources, cfg.KCPClusterName, cfg.WorkloadClusterName)
+	if err != nil {
+		return err
+	}
+	go apiImporter.Start(ctx, importPollInterval)
+
 	upstreamConfig := rest.CopyConfig(cfg.UpstreamConfig)
+	upstreamConfig.Host = syncerVirtualWorkspaceURL
 	upstreamConfig.UserAgent = "kcp#spec-syncer/v0.0.0"
```
ncdc, line 124: TODO get version or fill in now

pkg/syncer/syncer.go
```diff
@@ -73,15 +73,58 @@ func (sc *SyncerConfig) ID() string {
...
 	downstreamConfig := rest.CopyConfig(cfg.DownstreamConfig)
 	downstreamConfig.UserAgent = "kcp#status-syncer/v0.0.0"
```
ncdc, line 126: TODO get version or fill in now

ncdc (commented):
Does line 141

```
upstreamInformers := dynamicinformer.NewFilteredDynamicSharedInformerFactory(upstreamDynamicClient.Cluster(cfg.KCPClusterName), resyncPeriod, metav1.NamespaceAll, func(o *metav1.ListOptions) {
```

need to switch to be a wildcard cluster?

pkg/syncer/syncer.go
```diff
@@ -73,15 +73,58 @@ func (sc *SyncerConfig) ID() string {
...
+	// For now we retrieve the syncerVirtualWorkpaceURL at start, since we temporarily stick to a single URL (sharding not supported).
+	// But the complete implementation should setup a WorkloadCluster informer, create spec and status syncer for every URLs found in the
+	// Status.SyncerVirtualWorkspaceURLs slice, and update them each time this list changes.
+	var syncerVirtualWorkspaceURL string
+	err = wait.PollImmediateInfinite(gvrQueryInterval, func() (bool, error) {
+		klog.Infof("Attempting to retrieve the Syncer virtual workspace URL from WorkloadCluster %s|%s", cfg.KCPClusterName, cfg.WorkloadClusterName)
+
+		workloadCluster, err := kcpClusterClient.Cluster(cfg.KCPClusterName).WorkloadV1alpha1().WorkloadClusters().Get(ctx, cfg.WorkloadClusterName, metav1.GetOptions{})
+		if err != nil {
+			return false, err
```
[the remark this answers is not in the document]
davidfestal: > If the WorkloadCluster doesn't exist, err will be NotFound, and this will exit the polling loop with an error... and then on line 104, it says "Should never happen". Would recommend either changing this logic to be accepting of NotFound errors, or adjust the comment below.

The "Should never happen" comment was just a wrong copy-paste. Let me remove it and keep the same logic (error when WorkloadCluster is not found), since this is already applied later on, in line [177](https://github.com/kcp-dev/kcp/blob/fd2d0affb194438440282df4f340b922e0fd57b4/pkg/syncer/syncer.go#L177)

> Also, if this is polling forever, the user has no idea what is going on unless they look in the syncer pod logs. Would recommend setting a condition on the WorkloadCluster before entering this poll loop indicating current status. Could do as a TODO (follow-up) or in this PR

I'd rather add a TODO and leave that for a next PR, since it seems to be the same need as what is in line [148](https://github.com/kcp-dev/kcp/blob/fd2d0affb194438440282df4f340b922e0fd57b4/pkg/syncer/syncer.go#L148).

In addition to that, I assume that when we start managing the list of URLs, we'll have to refactor this deeply.

pkg/syncer/syncer.go
davidfestal: Fixed

pkg/syncer/syncer.go
davidfestal: Fixed

pkg/syncer/syncer.go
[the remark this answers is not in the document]
davidfestal: Dummy comment removed

pkg/syncer/syncer.go
davidfestal: Fixed

pkg/syncer/syncer.go
davidfestal: Fixed

pkg/syncer/syncer.go
```diff
@@ -73,15 +74,63 @@ func (sc *SyncerConfig) ID() string {
...
 	klog.Infof("Starting syncer for logical-cluster: %s, workload-cluster: %s", cfg.KCPClusterName, cfg.WorkloadClusterName)
 
-	upstreamConfig := rest.CopyConfig(cfg.UpstreamConfig)
-	upstreamConfig.UserAgent = "kcp#spec-syncer/v0.0.0"
-	downstreamConfig := rest.CopyConfig(cfg.DownstreamConfig)
-	downstreamConfig.UserAgent = "kcp#status-syncer/v0.0.0"
+	kcpVersion := version.Get().GitVersion
 
-	kcpClusterClient, err := kcpclient.NewClusterForConfig(upstreamConfig)
+	kcpClusterClient, err := kcpclient.NewClusterForConfig(rest.AddUserAgent(rest.CopyConfig(cfg.UpstreamConfig), "kcp#syncer/"+kcpVersion))
```
sttts, line 79: please call the client-go default func as in the cli plugin. No need to invent a new scheme.

pkg/syncer/syncer.go
```diff
@@ -73,15 +74,63 @@ func (sc *SyncerConfig) ID() string {
...
+
+	// TODO(david): Implement real support for several virtual workspace URLs that can change over time.
+	// TODO(david): For now we retrieve the syncerVirtualWorkpaceURL at start, since we temporarily stick to a single URL (sharding not supported).
+	// TODO(david): But the complete implementation should setup a WorkloadCluster informer, create spec and status syncer for every URLs found in the
+	// TODO(david): Status.SyncerVirtualWorkspaceURLs slice, and update them each time this list changes.
+	var syncerVirtualWorkspaceURL string
+	// TODO(david): we need to provide user-facing details if this polling goes on forever. Blocking here is a bad UX.
+	// TODO(david): Also, any regressions in our code will make any e2e test that starts a syncer (at least in-process)
+	// TODO(david): block until it hits the 10 minute overall test timeout.
+	err = wait.PollImmediateInfinite(gvrQueryInterval, func() (bool, error) {
```
sttts, line 92: why gvrQueryInterval ? Can we make that 5s or so?

pkg/syncer/syncer.go
```diff
@@ -73,15 +74,63 @@ func (sc *SyncerConfig) ID() string {
...
+		klog.Infof("Attempting to retrieve the Syncer virtual workspace URL from WorkloadCluster %s|%s", cfg.KCPClusterName, cfg.WorkloadClusterName)
```
sttts, line 93: move out of the loop

pkg/syncer/syncer.go
```diff
@@ -73,15 +74,63 @@ func (sc *SyncerConfig) ID() string {
...
+		syncerVirtualWorkspaceURL = workloadCluster.Status.VirtualWorkspaces[0].URL
+		return true, nil
+	})
+	if err != nil {
+		return err
+	}
+
+	// Resources are accepted as a set to ensure the provision of a
+	// unique set of resources, but all subsequent consumption is via
+	// slice whose entries are assumed to be unique.
```
sttts, line 116: is there value in this comment?

davidfestal: Was already there (moved code)

pkg/syncer/syncer.go
```diff
@@ -73,15 +74,63 @@ func (sc *SyncerConfig) ID() string {
...
+	// workspace.
+	apiImporter, err := NewAPIImporter(cfg.UpstreamConfig, cfg.DownstreamConfig, resources, cfg.KCPClusterName, cfg.WorkloadClusterName)
+	if err != nil {
+		return err
+	}
+	go apiImporter.Start(ctx, importPollInterval)
+
+	upstreamConfig := rest.CopyConfig(cfg.UpstreamConfig)
+	upstreamConfig.Host = syncerVirtualWorkspaceURL
+	upstreamConfig.UserAgent = "kcp#spec-syncer/" + kcpVersion
```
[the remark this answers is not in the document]
davidfestal: like this afaict: `v1.23.5+kcp-v0.4.0-alpha.0-178-gba18125a1635e2`

PR #1076, opened 2022-05-17
virtual: expand APIDefinitionSetGetter

stevekuznetsov:
I saw in https://github.com/kcp-dev/kcp/pull/1052 @ncdc had similar requirements, so I'm pulling this out for us to share.

/assign @davidfestal @sttts @ncdc

### DIFF
--- a/pkg/virtual/framework/dynamic/apiserver/handler_test.go
+++ b/pkg/virtual/framework/dynamic/apiserver/handler_test.go
@@ -48,7 +48,7 @@ type mockedAPISetRetriever apidefinition.APIDefinitionSet
 
 var _ apidefinition.APIDefinitionSetGetter = (*mockedAPISetRetriever)(nil)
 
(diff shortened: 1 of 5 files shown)


ncdc:
Is this to move the logical cluster from the APIDomainKey into the ctx?

stevekuznetsov:
@ncdc no - I needed the context for a client call (but I guess I should be using a lister)

stevekuznetsov:
@ncdc removed the ctx plumbing

pkg/virtual/framework/dynamic/apiserver/handler_test.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package apiserver
 
 import (
+	"context"
```
ncdc, line 20: Remove?

ncdc (approved):

PR #1077, opened 2022-05-17
Adding apibinding authorizer to allow service bindings ability to limit its access to exported resources

### LINKED ISSUE #1013: Add Authorizer that let's APIExport owner restrict what a user can do with objects of an APIBinding
As an exported set of resources owner, I would like to restrict what a user can do with those resources once bound to a workspace.

There should be a virtual workspace for a policy that an API export owner can create RBAC rules that will be applied by a new authorizer that will use this RBAC. 
 
**Open Questions**

~~1. I don't understand the type of RBAC rules as an example for this feature. Would we have deny rules?~~
1. Are the RBAC Rules going to all be cluster scoped for the same workspace that the APIExport is created in?

**Current Implementation Plan:**
1. Add an authorizer that will determine if a call to a resource is bound, if it is, and the workspace is defined, and make new RBAC requests against that workspace. if it passes, then delegate down, if it fails return.
2. if the request is not bound, delegate it down the flow chain.

**User Stories:** 
1. API Export where every user is able to create, but only a pre-determined SA can update the status. -> most users should be restricted from updating status, SA should be able to update status.
2. API Export with no additional roles, but the workspace admin has configured RBAC to only allow get/list for a particular user. -> user should be denied to create, but can list
3. API Export where every user can create/get/list/update but not update status. The workspace also has an RBAC that only allows get/list. -> user should be denied to create, should be able to list
4. API Export allows users to get/list but not create/update etc. SA can create and update. No RBAC policy should be allowed to be created in bound workspace that allows for create/update.

shawn-hurley:
## Summary
Adding Authorizer that will check if a APIBinding exists, and if it does exist check the RBAC in the Exports Cluster, to determine if the request is allowed.
## Related issue(s)

Fixes #1013

### DIFF
--- a/pkg/authorization/apibinding_authorizer.go
+++ b/pkg/authorization/apibinding_authorizer.go
@@ -1,0 +1,142 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package authorization
+
+import (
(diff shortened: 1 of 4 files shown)


shawn-hurley:
Accidentally pushed  branch to origin closing in favor of PR from my fork

PR #1078, opened 2022-05-17
Adding apibinding authorizer to allow service bindings ability to limit access to exported resources

### LINKED ISSUE #1013: Add Authorizer that let's APIExport owner restrict what a user can do with objects of an APIBinding

shawn-hurley:
## Summary
Adding Authorizer that will check if a APIBinding exists, and if it does exist check the RBAC in the Exports Cluster, to determine if the request is allowed.

## Related issue(s)

Fixes #1013


pkg/authorization/apibinding_authorizer.go
```diff
@@ -0,0 +1,137 @@
...
+	apiBindingIndexer  cache.Indexer
+	delegate           authorizer.Authorizer
+}
+
+func (a *apiBindingAccessAuthorizer) Authorize(ctx context.Context, attr authorizer.Attributes) (authorized authorizer.Decision, reason string, err error) {
+	// Determine if resource being requested is bound.
+	// If it is bound get the workspace.
+	// request for the user in the cluster.
+
+	apiBindingAccessDenied := fmt.Sprintf("api %v.%v access is not permitted", attr.GetResource(), attr.GetAPIGroup())
```
sttts, line 75: which reason does one usually get from the kube RBAC authorizer? Would at least use some similar formatting

Wondering how much we can leak. I tend to think that we can leak that it is a bound API and not a local one.

pkg/authorization/apibinding_authorizer.go
```diff
@@ -0,0 +1,137 @@
...
+	kcpinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
+	rbacwrapper "github.com/kcp-dev/kcp/pkg/virtual/framework/wrappers/rbac"
+)
+
+const (
+	byWorkspaceIndex = "apiBindingAuthorizer-byWorkspace"
+)
+
+// NewAPIBindingAccessAuthorizer returns an authorizer that checks for
+func NewAPIBindingAccessAuthorizer(versionedInformers clientgoinformers.SharedInformerFactory, f kcpinformers.SharedInformerFactory, delegate authorizer.Authorizer) authorizer.Authorizer {
```
sttts, line 43: (taste) `f` is good for local vars, not for func arguments. I believe our convention is `kcpInformers`.

sttts: and `kubeInformers` for the first arg

pkg/authorization/apibinding_authorizer.go
```diff
@@ -0,0 +1,137 @@
...
+				return []string{logicalcluster.From(obj.(metav1.Object)).String()}, nil
+			},
+			},
+		)
+		if err != nil {
+			// nothing we can do here. But this should also never happen. We check for existence before.
+			klog.Errorf("failed to add indexer for APIBindings: %v", err)
+		}
+	}
+
```
sttts, line 56: this is missing to call `Rbac().V1().{all the resources we need}`. Background: informers only start those watches for the resource where at least one `Rbac().V1().Resource().Informers()` was called.

pkg/authorization/apibinding_authorizer.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	for _, obj := range objs {
+		apiBinding := obj.(*apisv1alpha1.APIBinding)
+		for _, br := range apiBinding.Status.BoundResources {
+			if apiBinding.Status.BoundAPIExport.Workspace == nil {
+				// this will never happen today. But as soon as we add other reference types (like exports), this log output will remind out of necessary work here.
+				continue
+			}
+			if br.Group == attr.GetAPIGroup() && br.Resource == attr.GetResource() {
+				return parentClusterName.Join(apiBinding.Status.BoundAPIExport.Workspace.WorkspaceName), true, nil
```
sttts, line 132: @s-urbaniak @ncdc please build an opinion where we want to have the upper-bound policy. Here it is in the APIExport workspace. Is that good, or does it mangle the policy with the APIExport objects? Am not sure. Alternative: have a field in the APIExport for the policy workspace? (can all be follow-up).

sttts:
Looks all very good. @shawn-hurley well done!

pkg/authorization/apibinding_authorizer.go
shawn-hurley: So, I am following this: https://github.com/kcp-dev/kcp/blob/698be764b4e946cb2369449ae724ffb66b4c9791/pkg/authorization/toplevel_org_authorizer.go#L72

The error that I am seeing is this: 

```cowboys.wildwest.dev "cowboy-e2e-workspace-6shcm" is forbidden: User "user-1" cannot update resource "cowboys/status" in API group "wildwest.dev" in the namespace "default"```

I spent some time looking around and wondering if: "bound api access is not permitted" is what you are thinking?

File: test/e2e/apibinding/apibinding_authorizer_test.go [hunk 5]
```diff
@@ -0,0 +1,216 @@
...
+
+		t.Logf("Make sure there is 1 cowboy in consumer workspace %q", consumerWorkspace)
+		cowboys, err = cowboyClient.List(ctx, metav1.ListOptions{})
+		require.NoError(t, err, "error listing cowboys in consumer workspace %q", consumerWorkspace)
+		require.Equal(t, 1, len(cowboys.Items), "expected 1 cowboy in consumer workspace %q", consumerWorkspace)
+		require.Equal(t, cowboyName, cowboys.Items[0].Name, "unexpected name for cowboy in consumer workspace %q", consumerWorkspace)
+
+		t.Logf("Make sure that the status of cowboy can not be updated")
+		_, err = cowboyClient.UpdateStatus(ctx, &cowboys.Items[0], metav1.UpdateOptions{})
+		require.Error(t, err, "expected error updating status of cowboys")
```
ncdc, line 203: I feel like I'm blind - what is blocking this?

shawn-hurley:
@sttts Had a quick chat with @ncdc about the user experience with this change. if a service provider does not create any RBAC, this will currently fail closed. One thought Andy had was if we some add some basic RBAC for the resource on creation of APIExport?

pkg/authorization/apibinding_authorizer.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+// NewAPIBindingAccessAuthorizer returns an authorizer that checks for
+func NewAPIBindingAccessAuthorizer(versionedInformers clientgoinformers.SharedInformerFactory, f kcpinformers.SharedInformerFactory, delegate authorizer.Authorizer) authorizer.Authorizer {
+	if _, found := f.Apis().V1alpha1().APIBindings().Informer().GetIndexer().GetIndexers()[byWorkspaceIndex]; !found {
+		err := f.Apis().V1alpha1().APIBindings().Informer().AddIndexers(
+			cache.Indexers{byWorkspaceIndex: func(obj interface{}) ([]string, error) {
+				return []string{logicalcluster.From(obj.(metav1.Object)).String()}, nil
+			},
+			},
```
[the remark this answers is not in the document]
shawn-hurley: +1 I was trying to make this line more readable, but this didn't help.

test/e2e/apibinding/apibinding_authorizer_test.go
```diff
@@ -0,0 +1,216 @@
...
+		})
+		require.NoError(t, err)
+
+		t.Logf("Make sure cowboys API resource shows up in consumer workspace %q group version discovery", consumerWorkspace)
+		resources, err := kcpClients.Cluster(consumerWorkspace).Discovery().ServerResourcesForGroupVersion(wildwestv1alpha1.SchemeGroupVersion.String())
+		require.NoError(t, err, "error retrieving consumer workspace %q API discovery", consumerWorkspace)
+		require.True(t, resourceExists(resources, "cowboys"), "consumer workspace %q discovery is missing cowboys resource", consumerWorkspace)
+
+		// User a new user in Cconsumer workspace
+		// New User
```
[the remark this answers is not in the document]
shawn-hurley: updated test

test/e2e/apibinding/apibinding_authorizer_test.go
```diff
@@ -0,0 +1,216 @@
...
+					Resources: []string{"cowboys"},
+				},
+			},
+		}
+		_, err = kubeClusterClient.Cluster(serviceProviderWorkspace).RbacV1().ClusterRoles().Create(ctx, clusterRole, metav1.CreateOptions{})
+		require.NoError(t, err)
+
+		clusterRoleBinding := &rbacv1.ClusterRoleBinding{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "test-sytstemauth",
```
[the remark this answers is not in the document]
shawn-hurley: updated

test/e2e/apibinding/apibinding_authorizer_test.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+
+		_, err = kcpClients.Cluster(consumerWorkspace).ApisV1alpha1().APIBindings().Create(ctx, apiBinding, metav1.CreateOptions{})
+		require.NoError(t, err)
+
+		t.Logf("Make sure %s API group does NOT show up in workspace %q group discovery", wildwest.GroupName, providerWorkspace)
+		groups, err := kcpClients.Cluster(providerWorkspace).Discovery().ServerGroups()
+		require.NoError(t, err, "error retrieving service provider workspace %q group discovery", providerWorkspace)
+		require.False(t, groupExists(groups, wildwest.GroupName),
+			"should not have seen %s API group in service provider workspace %q group discovery", wildwest.GroupName, providerWorkspace)
```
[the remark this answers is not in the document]
shawn-hurley: removed!

ncdc:
If local is unset, don't check RBAC?

shawn-hurley:
@ncdc that is correct

ncdc:
SGTM. My only thought is a bikeshedding one: should we call it something like accessControl instead of policy?

shawn-hurley:
My only thought is that we can not have deny rules access control might give the wrong impression?

pkg/authorization/apibinding_authorizer.go
shawn-hurley: done

pkg/authorization/apibinding_authorizer.go
shawn-hurley: Added a field for this as discussed below

pkg/apis/apis/v1alpha1/types.go
```diff
@@ -358,6 +362,15 @@ type Identity struct {
 	SecretRef *corev1.SecretReference `json:"secretRef"`
 }
 
+// APIExportPolicy is a wrapper type around the multiple options that would be allowed.
+type APIExportPolicy struct {
+	Local *LocalAPIExportPolicy `json:"local"`
```
shawn-hurley, line 367: Must add this as optional.

shawn-hurley: resolved

pkg/apis/apis/v1alpha1/types.go
```diff
@@ -339,6 +339,10 @@ type APIExportSpec struct {
 	//
 	// +optional
 	Identity *Identity `json:"identity"`
+
+	// policy will allow for a service provider to set a upper bound on what is allowed
+	// +optional
+	Policy *APIExportPolicy `json:"policy"`
```
sttts, line 345: omitempty

pkg/apis/apis/v1alpha1/types.go
```diff
@@ -350,6 +354,18 @@ type Identity struct {
 	SecretRef *corev1.SecretReference `json:"secretRef"`
 }
 
+// APIExportPolicy is a wrapper type around the multiple options that would be allowed.
+type APIExportPolicy struct {
```
sttts, line 358: this needs a yaml-patch file to enforce at least one of the variants (here only one), compare https://github.com/kcp-dev/kcp/blob/main/config/crds/apis.kcp.dev_apibindings.yaml-patch

pkg/apis/apis/v1alpha1/types.go
```diff
@@ -350,6 +354,18 @@ type Identity struct {
...
+	//Local is policy that is defined in same namespace as API Export.
```
sttts, line 359:
```suggestion
	// local is policy that is defined in same namespace as API Export.
```

API go doc are always about the low-case field names as used in JSON.

pkg/apis/apis/v1alpha1/types.go
```diff
@@ -339,6 +339,10 @@ type APIExportSpec struct {
 	//
 	// +optional
 	Identity *Identity `json:"identity"`
+
+	// policy will allow for a service provider to set a upper bound on what is allowed
```
sttts, line 343:
```suggestion
	// policy will allow for a service provider to set a upper bound on what is allowed for a consumer of this API.
```

sttts: Wondering about a better field name. Something like "maximalPermissionPolicy" ?

pkg/apis/apis/v1alpha1/types.go
shawn-hurley: Is this generated or do I have to make it manually? 

"enforce the variants" can you explain this to me offline, because I don't know if I follow or understand.

pkg/apis/apis/v1alpha1/types.go
shawn-hurley: Done

pkg/apis/apis/v1alpha1/types.go
shawn-hurley: Done

pkg/apis/apis/v1alpha1/types.go
shawn-hurley: Got an explanation from Andy and now I understand and have added.

pkg/apis/apis/v1alpha1/types.go
shawn-hurley: Changed the name, with a TODO to come back before beta.

sttts (approved):

pkg/apis/apis/v1alpha1/types.go
```diff
@@ -346,6 +346,14 @@ type APIExportSpec struct {
...
 	// +optional
 	Identity *Identity `json:"identity"`
+
+	// TODO: before beta we should re-evaluate this field name
+
+	// maximalPermissionPolicy will allow for a service provider to set a upper bound on what is allowed
+	// for a consumer of this API. If the policy is not set, no upper bound is applied,
+	// i.e the consuming users can do whatever the user workspace allows the user to do.
+	// +optional
+	MaximalPermissionPolicy *APIExportPolicy `json:"maximalPermissionPolicy,omitempty"`
```
sttts, line 356: @ncdc any ideas for a better field name? Feels like we didn't get it quite right yet (the cosmetical part).

ncdc: I'm ok on the current name. We can bike shed for a future alpha & definitely before beta

pkg/authorization/apibinding_authorizer.go
```diff
@@ -0,0 +1,190 @@
...
+		err := kcpInformers.Apis().V1alpha1().APIBindings().Informer().AddIndexers(
+			cache.Indexers{
+				byWorkspaceIndex: func(obj interface{}) ([]string, error) {
+					return []string{logicalcluster.From(obj.(metav1.Object)).String()}, nil
+				},
+			},
+		)
+		if err != nil {
+			// nothing we can do here. But this should also never happen. We check for existence before.
+			klog.Errorf("failed to add indexer for APIBindings: %v", err)
```
s-urbaniak, line 57: @sttts this will not fail the constructor and might mask iffy states if the indexer 🤔 is this really intended just to silently ignore the error?

PR #1080, opened 2022-05-18
admission: make api binding mutating admission work

stevekuznetsov:

/assign @ncdc


ncdc:
Mind removing the initializers field entirely from APIBindings instead? It's not used and not needed at this time.

ncdc:
Converted to draft so this doesn't accidentally get merged as-is 😄

stevekuznetsov:
@ncdc can we merge this, then a commit to remove it all, so that any revert brings it back to an actually functional state and nobody has to re-learn what is going on ?

ncdc:
Sure thing!

ncdc (approved):

PR #1082, opened 2022-05-18
Fix wildcard partial metadata list/watch requests

### LINKED ISSUE #1072: Make partial metadata list/get/watch work correctly - wildcard + non wildcard
TODO fill in details

ncdc:
## Summary
Fix wildcard partial metadata list/watch requests. There was an issue where not all instances across all CRDs implementing a group-resource were retrieved.

### TODO
- [x] Better docs/comments
- [x] e2e tests

## Related issue(s)

Fixes #1072

Depends on https://github.com/kcp-dev/kubernetes/pull/70


pkg/server/handler.go
```diff
@@ -128,12 +129,18 @@ func WithClusterScope(apiHandler http.Handler) http.HandlerFunc {
 		} else {
 			clusterName = logicalcluster.New(req.Header.Get("X-Kubernetes-Cluster"))
 		}
+
 		var cluster request.Cluster
+
+		if _, params, err := mime.ParseMediaType(req.Header.Get("Accept")); err == nil {
```
sttts, line 135: see WithAcceptHeader nad use isPartialMetadataRequest func.

sttts:
module WithAcceptHeader comment, lgtm

pkg/reconciler/workload/resource/resource_reconcile.go
```diff
@@ -76,6 +76,15 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName logical
...
 
+	// TODO(ncdc): wildcard partial metadata requests are now including CRs that come via APIBindings. Multiple e2e tests
+	// manually assign a scheduling label to a resource but not its namespace. The resources were getting synced based
+	// on the manually assigned label, and staying with that assignment. But now because wildcard partial metadata
+	// lists include these bound CRs, not having the namespace assigned is a problem. For now, don't unassign a
+	// resource that is scheduled if the namespace's assignment is unset.
+	if previousCluster != "" && newCluster == "" {
+		return nil
+	}
+
```
sttts, line 87: this seems wrong. The namespace is driving resource scheduling. We should fix the tests.

sttts: Which tests do you have in mind? I don't see any setting the label, at least in #1084.

sttts (approved):

pkg/reconciler/workload/resource/resource_reconcile.go
ncdc: It's existing tests that are doing it, such as https://github.com/kcp-dev/kcp/blob/19a74feffe74926ac2e5935ef352260f59d1265f/test/e2e/virtual/syncer/virtualworkspace_test.go#L417. Before the fixes in this PR, the resource controller would never see bound CRs, so it would never try to schedule them. Now it does, and if the namespace has `""`, the resource controller will update the resource to match the namespace's label value. So you either have to manually set the label on the namespace first, or do this.

pkg/server/apiextensions.go
```diff
@@ -619,10 +628,6 @@ func findCRD(name string, crds []*apiextensionsv1.CustomResourceDefinition) (*ap
 
 func findFirstCRDMatchingGroupResource(group, resource string, crds []*apiextensionsv1.CustomResourceDefinition) *apiextensionsv1.CustomResourceDefinition {
 	for _, crd := range crds {
-		if _, bound := crd.Annotations[apisv1alpha1.AnnotationBoundCRDKey]; bound {
-			continue
-		}
-
```
sttts, line 625: why got this deleted?

sttts: ic, it's just used in `getForPartialMetadata`.

PR #1091, opened 2022-05-18
Avoid trying to sync namespaces as namespaced resources.

davidfestal:
## Summary

After recent changes, it seems that the syncer constantly tries to sync namespace resources, as other namespaced resources, which of course is not possible.
Surely namespaces should be received by the informer and present in the informer caches, but not effectively synced by the syncer `process` functions.

Fixes #

Not sure there was an issue for this.

### DIFF
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -235,7 +235,7 @@ func getAllGVRs(discoveryClient discovery.DiscoveryInterface, resourcesToSync ..
 	}
 	// TODO(jmprusi): Added ServiceAccounts, Configmaps and Secrets to the default syncing, but we should figure out
 	//                a way to avoid doing that: https://github.com/kcp-dev/kcp/issues/727
-	gvrstrs := sets.NewString("namespaces.v1.", "serviceaccounts.v1.", "configmaps.v1.", "secrets.v1.") // A syncer should always watch namespaces, serviceaccounts, secrets and configmaps.
+	gvrstrs := sets.NewString("serviceaccounts.v1.", "configmaps.v1.", "secrets.v1.") // A syncer should always watch serviceaccounts, secrets and configmaps.
 	for _, r := range rs {
 		// v1 -> v1.
 		// apps/v1 -> v1.apps
--- a/pkg/syncer/status/status_controller.go
+++ b/pkg/syncer/status/status_controller.go
@@ -43,6 +43,7 @@ type Controller struct {
 
 	upstreamClient, downstreamClient       dynamic.Interface
 	upstreamInformers, downstreamInformers dynamicinformer.DynamicSharedInformerFactory
+	downstreamNamespaceLister              cache.GenericLister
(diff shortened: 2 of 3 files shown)


pkg/syncer/spec/spec_controller.go
```diff
@@ -98,6 +98,7 @@ func NewSpecSyncer(gvrs []schema.GroupVersionResource, upstreamClusterName logic
 		})
 		klog.InfoS("Set up informer", "clusterName", upstreamClusterName, "pcluster", workloadClusterName, "gvr", gvr.String())
 	}
+	upstreamInformers.ForResource(schema.GroupVersionResource{Version: "v1", Resource: "namespaces"}).Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{})
```
sttts, line 101: the last call afaik is not necessary.

davidfestal: well, when not there the namespaces are not present in the informer cache.

sttts: Alternative, much more in line with other code: have a namespaceLister in the controller object, and here you would call Lister() instead of Informer().

ncdc: 1046 uses the lister but is blocked on other things

davidfestal: I can change for the lister option, which would allow merging this fix without waiting for the rest of the other PR ?

davidfestal: Done.

davidfestal:
@itdove This PR does something that your PR #1046  also fixed. WDYT of merging this one first, and thus decoupling this fix from the main purpose of your PR #1046 ?
Your PR should be quite easy to rebase on this change which is very similar to what your PR includes.

pkg/syncer/syncer.go
```diff
@@ -235,7 +235,7 @@ func getAllGVRs(discoveryClient discovery.DiscoveryInterface, resourcesToSync ..
 	}
 	// TODO(jmprusi): Added ServiceAccounts, Configmaps and Secrets to the default syncing, but we should figure out
 	//                a way to avoid doing that: https://github.com/kcp-dev/kcp/issues/727
-	gvrstrs := sets.NewString("namespaces.v1.", "serviceaccounts.v1.", "configmaps.v1.", "secrets.v1.") // A syncer should always watch namespaces, serviceaccounts, secrets and configmaps.
+	gvrstrs := sets.NewString("serviceaccounts.v1.", "configmaps.v1.", "secrets.v1.") // A syncer should always watch namespaces, serviceaccounts, secrets and configmaps.
```
itdove, line 238: Same comment as @ncdc did to me, please update the trailing comment (remote namespace)

davidfestal: Fixed

ncdc (approved):

PR #1095, opened 2022-05-18
apibinding: remove initializers

stevekuznetsov:
These were unused and are not needed.


ncdc:
Need to run `make codegen` to get the CRD yaml updated?

PR #1101, opened 2022-05-18
virtualworkspace: inspect available verbs from storage

stevekuznetsov:

/assign @ncdc @sttts @davidfestal

### DIFF
--- a/pkg/virtual/framework/dynamic/apiserver/discovery.go
+++ b/pkg/virtual/framework/dynamic/apiserver/discovery.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/apiserver/pkg/endpoints/discovery"
 	"k8s.io/apiserver/pkg/endpoints/handlers/negotiation"
 	"k8s.io/apiserver/pkg/endpoints/handlers/responsewriters"
+	"k8s.io/apiserver/pkg/registry/rest"
 	"k8s.io/kubernetes/pkg/genericcontrolplane/aggregator"
 
 	apiresourcev1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apiresource/v1alpha1"
@@ -107,9 +108,6 @@ func (r *versionDiscoveryHandler) ServeHTTP(w http.ResponseWriter, req *http.Req
 		apiResourceSpec := apiDef.GetAPIResourceSpec()
 		subresources = apiResourceSpec.SubResources
 
-		// TODO: get the list of verbs from the REST storage instance
-		verbs := metav1.Verbs([]string{"get", "list", "patch", "create", "update", "watch"})
-
 		storageVersionHash = discovery.StorageVersionHash(apiDef.GetClusterName().String(), gvr.Group, gvr.Version, apiDef.GetAPIResourceSpec().Kind)
 
 		apiResourcesForDiscovery = append(apiResourcesForDiscovery, metav1.APIResource{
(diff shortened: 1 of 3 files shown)


stevekuznetsov:
This whole flow that we have looks very different from the APIInstaller in k8s - to what end do we want these two to look the same?

stevekuznetsov:
Also - because of the way we're using `customresource.NewStorageWithCustomStore` we're locked in to implementing `store.StandardStorage` for everything, so we can't use the contents of this patch yet.

pkg/virtual/framework/dynamic/apiserver/handler_test.go
```diff
@@ -191,11 +244,27 @@ func TestRouting(t *testing.T) {
 			ExpectDelegateCalled: false,
 			ExpectStatus:         200,
 			ExpectResponse: func(t *testing.T, r *http.Response, b []byte) {
-				if r.Header.Get("Content-Type") == "application/json" && r.StatusCode == 200 {
-					require.Equal(t,
-						"{\"kind\":\"APIGroup\",\"apiVersion\":\"v1\",\"name\":\"custom\",\"versions\":[{\"groupVersion\":\"custom/v1\",\"version\":\"v1\"}],\"preferredVersion\":{\"groupVersion\":\"custom/v1\",\"version\":\"v1\"}}\n",
```
stevekuznetsov, line 196: I don't think that asserting on some string here is correct - JSON has no ordering w.r.t. fields at the same level of nesting. anyway. Plus, actually comparing objects gives us sensible diffs.

pkg/virtual/framework/dynamic/apiserver/handler_test.go
```diff
@@ -251,11 +336,29 @@ func TestRouting(t *testing.T) {
...
+						APIVersion: "v1",
+					},
+					GroupVersion: "custom/v1",
+					APIResources: []metav1.APIResource{
+						{
+							Name:               "customresources",
+							SingularName:       "customresource",
+							Namespaced:         true,
+							Kind:               "CustomResource",
+							Verbs:              []string{"get", "list"},
```
stevekuznetsov, line 357: This just parrots whatever data we put into the mocks, perhaps we should use resource names etc that don't have any meaning, so as to not confuse people - certainly I was surprised to learn this was not reading some real store of CRD information.

pkg/virtual/framework/dynamic/apiserver/handler_test.go
```diff
@@ -397,11 +546,34 @@ func TestRouting(t *testing.T) {
...
+					APIResources: []metav1.APIResource{
+						{
+							Name:               "services",
+							SingularName:       "service",
+							Namespaced:         true,
+							Kind:               "Service",
+							Verbs:              []string{"get"},
+							StorageVersionHash: "+iYBRzoiY8o=",
+						},
+						{
```
stevekuznetsov, line 569: We did not test status subresources in the past. Now we do.

pkg/virtual/framework/dynamic/apiserver/handler_test.go
```diff
@@ -484,11 +679,26 @@ func TestRouting(t *testing.T) {
 			ExpectDelegateCalled: false,
 			ExpectStatus:         405,
 			ExpectResponse: func(t *testing.T, r *http.Response, b []byte) {
-				if r.Header.Get("Content-Type") == "application/json" {
-					require.JSONEq(t,
```
stevekuznetsov, line 488:
```
- require.JSONEq
+ actual comparison of objects
```

pkg/virtual/framework/dynamic/apiserver/discovery.go
```diff
@@ -107,29 +108,25 @@ func (r *versionDiscoveryHandler) ServeHTTP(w http.ResponseWriter, req *http.Req
 		apiResourceSpec := apiDef.GetAPIResourceSpec()
 		subresources = apiResourceSpec.SubResources
 
-		// TODO: get the list of verbs from the REST storage instance
-		verbs := metav1.Verbs([]string{"get", "list", "patch", "create", "update", "watch"})
```
stevekuznetsov, line 111: ... why did this not include delete ?

davidfestal: Because it was initially written for the Syncer virtual workspace, where we never intent to delete anything on the kcp logical cluster.

sttts (approved):

PR #1104, opened 2022-05-19
api: Enables the binding of APIs in protected groups

### LINKED ISSUE #1083: Cannot export or bind to resources in protected groups
**Describe the bug**
Cannot bind to api export of apiresourceschemas in protected groups (e.g. k8s.io)

**To Reproduce**
1. create and enter a new workspace (e.g. apigateway)
2. create APIResourceSchemas for a resource in a protected group (e.g. httproutes.gateway.networking.k8s.io)
3. create an APIExport resource that exports this resource
4. create and enter a new workspace (e.g. apigatewaybinding)
5. create an apibinding which binds to the previous apiexport
6. look in KCP logs for the error: `Error creating CRD for APIBinding ... <snip> ... is invalid: metadata.annotations[api-approved.kubernetes.io]: Required value: protected groups must have approval annotation "api-approved.kubernetes.io", see https://github.com/kubernetes/enhancements/pull/1111`

**Expected behavior**
This problem is happening because I cannot set a metadata annotation on the CRD in the apiresourceschema, I need to be able to set this annotation to enable me to export / bind to API resources in protected groups.

**Additional context**
The error that is reported on the API Binding is obscure, it would help if the error in the KCP logs could be logged against the API Binding for help debugging and identifying this issue in the future.

astefanutti:
## Summary

This PR enables the binding of APIs whose groups are _protected_, as defined in kubernetes/enhancements#1111.

The proposed solution delegates to the user the responsibility of adding the `api-approved.kubernetes.io` annotation to the `APIResourceSchema` resource, as for standard CRDs, so it makes it explicit, and prevents from accidentally declaring APIs in a protected group. With this PR, the APIBinding controller will simply propagate the annotation to the projected CRDs.

In the API import case, where CRDs are generated following the creation of `APIResourceImport` resources, the `NegociatedAPIResource` controller already adds the _approval_ annotation, assuming the API have already been explicitly declared as approved in the downstream workload cluster.    

TODO:
- [x] Propagate the _approval_ annotation from the `APIResourceSchema` resources to the generated CRDs
- [x] Add admission to reject `APIResouceSchema` resources in protected groups without the _approval_ annotation
- [x] Improve `APIBinding` condition message when the generated CRD is invalid
- [x] Add e2e test

## Related issue(s)

Fixes #1083

### DIFF
--- a/pkg/reconciler/apis/apibinding/apibinding_reconcile.go
+++ b/pkg/reconciler/apis/apibinding/apibinding_reconcile.go
@@ -19,6 +19,7 @@ package apibinding
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"sort"
 	"strings"
@@ -255,6 +256,39 @@ func (c *controller) reconcileBinding(ctx context.Context, apiBinding *apisv1alp
 			// Create flow
 
 			if _, err := c.createCRD(ctx, ShadowWorkspaceName, crd); err != nil {
+				if apierrors.IsInvalid(err) {
+					status := apierrors.APIStatus(nil)
+					// The error is guaranteed to implement APIStatus here
+					errors.As(err, &status)
+					conditions.MarkFalse(
+						apiBinding,
+						apisv1alpha1.BindingUpToDate,
+						apisv1alpha1.APIResourceSchemaInvalidReason,
+						conditionsv1alpha1.ConditionSeverityError,
+						fmt.Sprintf("APIResourceSchema %s|%s is invalid: %v\"", schema.ClusterName, schemaName, status.Status().Details.Causes),
+					)
+					// Only change InitialBindingCompleted if it's false
+					if conditions.IsFalse(apiBinding, apisv1alpha1.InitialBindingCompleted) {
+						conditions.MarkFalse(
+							apiBinding,
+							apisv1alpha1.InitialBindingCompleted,
+							apisv1alpha1.APIResourceSchemaInvalidReason,
+							conditionsv1alpha1.ConditionSeverityError,
+							fmt.Sprintf("APIResourceSchema %s|%s is invalid: %v\"", schema.ClusterName, schemaName, status.Status().Details.Causes),
+						)
+					}
+
+					klog.Errorf(
+						"Error creating CRD for APIBinding %s|%s, APIExport %s|%s, APIResourceSchema %s|%s: %v",
+						apiBinding.ClusterName, apiBinding.Name,
+						apiExport.ClusterName, apiExport.Name,
+						apiExport.ClusterName, schemaName,
+						err,
+					)
+
+					return nil
+				}
+
 				conditions.MarkFalse(
 					apiBinding,
 					apisv1alpha1.BindingUpToDate,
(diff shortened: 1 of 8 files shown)


pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -519,3 +545,30 @@ func apiExportLatestResourceSchemasChanged(apiBinding *apisv1alpha1.APIBinding,
 
 	return !exportedSchemaUIDs.Equal(boundSchemaUIDs)
 }
+
+func propagateLabelAndAnnotations(source, target metav1.Object) {
```
sttts, line 549: would rather expect a positive list. Why should we copy other annotation?

sttts:
One comment. Otherwise, looks very good.

sttts:
Compare https://github.com/kcp-dev/kcp/pull/1084/commits/47d0d378f6e02624655efa5a443d7d60cd4564c4. Aren't those changes missing?

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
astefanutti: I wasn't sure, but I could see values in copying the set of annotations and labels, for metering for instance, or simply to convey the annotations that the API providers generally set. For example the Gateway API set a number of annotations with the `gateway.networking.k8s.io` prefix.

astefanutti:
> Compare [47d0d37](https://github.com/kcp-dev/kcp/commit/47d0d378f6e02624655efa5a443d7d60cd4564c4) in #1084. Aren't those changes missing?

Unless I'm mistaken, the changes from 1ed69206724e27a0bf21d059ef5611b5228b0d8c aren't strictly required for the protected API use case. My understanding is that the _protected_ annotation is set at the end of the import execution path, when the `NegociatedAPIResource` is projected into the published CRD:

https://github.com/kcp-dev/kcp/blob/5dd2f2a6cd35fdcafc8f2e9282403997fe137f15/pkg/reconciler/apis/apiresource/negotiation.go#L702-L707

That being said, it would be required if we were to propagated the set of annotations arbitrarily, then the set would have to be transported along the resources chain (`APIResourceImport`, `NegotiatedAPIResource`, ...).

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
sttts: But nobody will see these CRDs. And there is the risk of side-effect, even security related.

sttts:
> That being said, it would be required if we were to propagated the set of annotations arbitrarily, then the set would have to be transported along the resources chain

But it does not hurt either. Bonus: we don't get this fake pull request number 4, but the real one in case of CRDs on the pcluster side.

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
astefanutti: Make sense, let me rework it quickly.

astefanutti:
> > That being said, it would be required if we were to propagated the set of annotations arbitrarily, then the set would have to be transported along the resources chain
> 
> But it does not hurt either. Bonus: we don't get this fake pull request number 4, but the real one in case of CRDs on the pcluster side.

Agreed, that's a good point about that fake approval PR 👍🏼.

sttts (approved):

PR #1109, opened 2022-05-19
Apply audit config to server

kylape:
This will enable audit logging if configured via flags.

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -183,6 +183,10 @@ func (s *Server) Run(ctx context.Context) error {
 		return err
 	}
 
+	if err := s.options.GenericControlPlane.Audit.ApplyTo(genericConfig); err != nil {
+		return err
+	}
+
 	// create service-account-only authenticator without any lookup for objects, just to extract the logical cluster name from the JWT.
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -183,6 +183,10 @@ func (s *Server) Run(ctx context.Context) error {
 		return err
 	}
 
+	if err := s.options.GenericControlPlane.Audit.ApplyTo(genericConfig); err != nil {
+		return err
+	}
+
```
sttts, line 189: oh, is this really needed? Don't we already call it through the options package somehow?

sttts: If we don't and you check this lgtm.

kylape: I did a local test where I did: 

```
go run ./cmd/kcp start --audit-log-maxage=10 --audit-log-maxbackup=1 --audit-log-maxsize=100 --audit-log-path=audit.log --audit-policy-file=audit-policy.yml
```

and no `audit.log` was created.  I then added the code in this PR, ran the same command, and the `audit.log` was created.

sttts (approved):

PR #1110, opened 2022-05-19
virtual: snip dependency on CRD storage builder

stevekuznetsov:
We don't need every possible endpoint to implement rest.StandardStorage
in every case, so we can refactor to reflect that. It seems like we're
getting a couple things from the current custom resource REST:

- shallow-copies of metadata, needed in the past when the generic
  registry was mutating it to set self-links, no longer needed
- providing categories
- forcing no creation on update behavior

By not using the custom resource REST implementation, we need to
duplicate ~10 lines of code, which does not seem risky.

/assign @sttts


sttts (dismissed):

stevekuznetsov:
Collided with my own previous PR, oops - rebased.

pkg/virtual/framework/forwardingregistry/rest.go
```diff
@@ -18,70 +18,69 @@ package forwardingregistry
...
+		return ret
+	}
+	listFactory := func() runtime.Object {
+		// lists are never stored, only manufactured, so stomp in the right kind
+		ret := &unstructured.UnstructuredList{}
+		ret.SetGroupVersionKind(listKind)
+		return ret
+	}
+	destroyer := func() {
+		// TODO: what do we do on Destroy()?
```
sttts, line 59: destroy watch servers.

pkg/virtual/framework/forwardingregistry/functions.go
```diff
@@ -0,0 +1,141 @@
...
+type UpdaterFunc func(ctx context.Context, name string, objInfo rest.UpdatedObjectInfo, createValidation rest.ValidateObjectFunc, updateValidation rest.ValidateObjectUpdateFunc, forceAllowCreate bool, options *metav1.UpdateOptions) (runtime.Object, bool, error)
+
+func (f UpdaterFunc) Update(ctx context.Context, name string, objInfo rest.UpdatedObjectInfo, createValidation rest.ValidateObjectFunc, updateValidation rest.ValidateObjectUpdateFunc, forceAllowCreate bool, options *metav1.UpdateOptions) (runtime.Object, bool, error) {
+	return f(ctx, name, objInfo, createValidation, updateValidation, forceAllowCreate, options)
+}
+
+var _ rest.Updater = &struct {
+	FactoryFunc
+	UpdaterFunc
+}{}
```
sttts, line 99: this is clever.

pkg/virtual/framework/forwardingregistry/rest.go
stevekuznetsov: Not today we don't ;)

sttts: ack

sttts (approved):

test/e2e/virtual/syncer/virtualworkspace_test.go
```diff
@@ -66,7 +66,7 @@ func deploymentsAPIResourceList(workspaceName string) *metav1.APIResourceList {
 				Name:               "deployments",
 				SingularName:       "deployment",
 				Namespaced:         true,
-				Verbs:              metav1.Verbs{"create", "delete", "deletecollection", "get", "list", "patch", "update", "watch"},
```
stevekuznetsov, line 69: @sttts we now lose these verbs as we expected to

PR #1112, opened 2022-05-19
Bump kcp-dev/kubernetes dependency

fabianvf:
This requires adding Destroy methods to anything implementing rest.Storage


pkg/virtual/syncer/builder/forwarding.go
```diff
@@ -78,6 +78,9 @@ func provideForwardingRestStorage(ctx context.Context, clusterClient dynamic.Clu
 
 		subresourceStorages = make(map[string]rest.Storage)
 		if statusEnabled {
+			// TODO storage.Status does not implement rest.Storage
```
fabianvf, line 81: This is the only thing still failing compilation, trying to find an interface that status and scale both implement that doesn't require the Destroy method (because the subresources do not implement rest.Storage)

fabianvf: fixed with https://github.com/kcp-dev/kubernetes/pull/71

ncdc (approved):

PR #1117, opened 2022-05-23
Makefile: fix WHAT for test-e2e

sttts:
Align with `make test`:
```
$ make test-e2e WHAT=./test/e2e/reconcilers/...
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,6 +68,10 @@ build: ## Build the project
 	go build -ldflags="$(LDFLAGS)" -o bin $(WHAT)
 .PHONY: build
 
+.PHONY: build-all
+build-all: WHAT := ./cmd/...
+build-all: build
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -68,6 +68,10 @@ build: ## Build the project
 	go build -ldflags="$(LDFLAGS)" -o bin $(WHAT)
 .PHONY: build
 
+.PHONY: build-all
+build-all: WHAT := ./cmd/...
```
qiujian16, line 72: just curious, is it necessary to set WHAT here?

sttts: yes, it overrides the value coming from the command line, to always enforce all commands being built for e2e tests (which makes sense as we call those binaries).

qiujian16: ah, I see

qiujian16 (approved):

Issue #1058, opened 2022-05-16
syncer: synced service account secrets are immediately deleted by the downstream cluster


ncdc:
**Describe the bug**
Service account secrets that are synced from kcp to a downstream cluster are immediately deleted.

**To Reproduce**
1. In the downstream cluster: `kubectl get secret --all-namespaces -w`
2. In kcp: `kubectl create serviceaccount foo`
3. See updates in the downstream cluster watch from (1)
4. Try to `kubectl get` the secret in the downstream cluster
5. See a non found error

**Expected behavior**
Either the secret shouldn't be immediately deleted, or (more likely) we should not be syncing this downstream in the first place

**Additional context**
This happens because the service account secret has an annotation `kubernetes.io/service-account.uid` that is the UID of the associated service account. But when kcp syncs the secret to the downstream cluster, it contains the UID for the service account in kcp. This doesn't match the UID of the service account in the downstream cluster, which is uniquely generated when it's created.

stevekuznetsov:
Unless I am mistaken, it does not seem necessary to put _anything_ in the downstream Pod's specification for a service account name. We are mounting in the in-cluster configuration data ourselves, anyway. So, we can just stop creating `ServiceAccount` objects in physical clusters, right?

It's not immediately obvious what it means to even create a `ServiceAccount` downstream - we cannot do so without racing with the downstream token minting controller, if it's on, and the UX here is extremely ambiguous, so perhaps let's turn off this feature unless we have a clear use-case?

ncdc:
Sounds good to me, at least until we find a use case where downstream needs a physical SA

ncdc:
What if we started by deleting these 2 annotations when syncing a secret down:

```
 kubernetes.io/service-account.name: default
 kubernetes.io/service-account.uid: 2734aec7-cab5-4f70-8659-72b97476f0d6
```

jmprusi:
This should be "fixed" by https://github.com/kcp-dev/kcp/pull/1123

jmprusi:
Fixed as https://github.com/kcp-dev/kcp/pull/1127 got merged, and now the SA Secrets are stripped of the ServiceAccount annotations and the Secret type is set to "Opaque" instead of ServiceAccountType.

PR #1123, opened 2022-05-23
Don't sync ServiceAccounts by default

jmprusi:
We are syncing the default serviceaccount from KCP to downstream, but it is not actually used downstream.

* Syncer: Stop syncing the serviceaccount by default.
* Syncer: Don't rename serviceaccount resources.
* Mutators: Deployment, don't add the serviceaccount to the downstream deployment podTemplate spec.


sttts:
This intelligence about which service account is needed seems to be at the wrong spot now. Rather than the syncher, the workload resource controller should have this knowledge and just not label unneeded service accounts.

sttts (approved):

jmprusi:
related https://github.com/kcp-dev/kcp/issues/1058

PR #1113, opened 2022-05-19
pkg/server: init audit annotations

ibihim:
Signed-off-by: Krzysztof Ostrowski <kostrows@redhat.com>

## Summary

Initializes Audit Annotations with `kaudit.WithAuditAnnotations`.

It is necessary to be done by some early middleware. I verified that with the change we annotate the workspace.

## Related issue(s)
https://github.com/kcp-dev/kcp/issues/862

PR #1124, opened 2022-05-23
pkg/server: add audit annotation init

ibihim:
Signed-off-by: Krzysztof Ostrowski <kostrows@redhat.com>

## Summary

Initializes Audit Annotations with `kaudit.WithAuditAnnotations`.

It is necessary to be done by some early middleware. I verified that with the change we annotate the workspace.

## Hint

Previous PR lost reference and couldn't be changed anymore.

## Related issue(s)
https://github.com/kcp-dev/kcp/pull/1113
https://github.com/kcp-dev/kcp/issues/862

### DIFF
--- a/pkg/server/handler.go
+++ b/pkg/server/handler.go
@@ -152,6 +152,18 @@ func WithClusterScope(apiHandler http.Handler) http.HandlerFunc {
 	}
 }
 
+// WithAuditAnnotation initializes audit annotations in the context. Without
+// initialization kaudit.AddAuditAnnotation isn't preserved.
+func WithAuditAnnotation(handler http.Handler) http.HandlerFunc {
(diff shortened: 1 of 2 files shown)


pkg/server/handler.go
```diff
@@ -152,6 +152,18 @@ func WithClusterScope(apiHandler http.Handler) http.HandlerFunc {
 	}
 }
 
+// WithAuditAnnotation initializes audit annotations in the context. Without
+// initialization kaudit.AddAuditAnnotation isn't preserved.
+func WithAuditAnnotation(handler http.Handler) http.HandlerFunc {
```
sttts, line 157: why not call the original directly?

sttts: nvmd. The upstream func wants more args. This looks fine.

sttts (approved):

PR #1125, opened 2022-05-23
apis/bindings: allow absolute workspace references

sttts:
```yaml
reference:
  workspace:
    path: root:foo:bar
```


pkg/apis/apis/v1alpha1/types.go
```diff
@@ -82,13 +82,18 @@ type ExportReference struct {
...
 	// +kube:validation:MinLength=1
 	// +kubebuilder:validation:Pattern:="^[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
-	WorkspaceName string `json:"name"`
+	WorkspaceName string `json:"name,omitempty"`
+
+	// absolute is an absolute reference to a workspace, e.g. root:org:ws. The workspace must
+	// be some ancestor or a child of some ancestor.
+	// +optional
+	// +kubebuilder:validation:Pattern:="^root(:[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$"
+	LogicalCluster string `json:"absolute,omitempty"`
```
ncdc, line 96: Would it be easier and less confusing if we had only one field for this instead of two?

sttts: they don't have the same meaning. We could drop the WorkspaceName.

shawn-hurley: From a user of this field, I think a single one would make sense, and LogicalCluster makes a ton of sense!

sttts: was not sure we want to use that term towards users. Today it's really a deep Kubernetes-fork detail.

shawn-hurley: What FullyQualifiedWorkspace or something like that?

ncdc: I know we've talked previously about trying to find some way to unify the 2 concepts (logical cluster, workspace) so that users can reason about them consistently and easily. I also know that a workspace has significantly more semantic meaning and behavior than a logical cluster does.

We've also talked about making the logical cluster name more like a unix path, i.e. instead of showing `root:org:ws` to the user, we show them `/org/ws`. I'm not suggesting we necessarily make any changes now, but we could call this something like "path" or "workspace path" perhaps.

In the immediate term, I think I prefer a single field that is the absolute logical cluster name. This is an alpha API, so we can always change the field name for something better in a future revision. WDYT?

ncdc: Sold!

sttts:
@ncdc @shawn-hurley this is ready

ncdc (approved):

Issue #1062, opened 2022-05-16
Stop hardcoding the default secret token names when syncing to a downstream cluster


jmprusi:
**Is your feature request related to a problem? Please describe.**
KCP syncs the secret tokens created by the default KCP SA account to the downstream cluster, but modifies its name with a hardcoded one "kcp-default-token":

* `kube-default-XXXX` gets changed downstream to `kcp-default-token`, where "XXXX" is a random suffix set by the k8s controller.

This was done for two reasons:
* Avoid any possible clash with the existing default service accounts tokens in the downstream cluster.
* Have a "constant" name, so other components can rely on it. (Syncer deployment mutator)

Replacing the "kube-default-" suffix by "kcp-default-" should be Ok, but not removing the random suffix.

**Describe the solution you'd like**

Figure out the necessary changes on the mutators part to discover the currently synced token instead of relying on a hardcoded one (kcp-default-token), applying the minimal transformation possible (`kube-...` -> `kcp-...`)

**Additional context**
related to https://github.com/kcp-dev/kcp/issues/1058

PR #1127, opened 2022-05-23
Syncer: Deployment Mutator use the full secret name instead of a hardcoded one 

jmprusi:
The Deployment Mutator will look for a ServiceAccount token, and use its full name for syncing downstream, instead of hardcoding one secret name (kcp-default-token).

* only the default token is prefixed with: "kcp-", to avoid possible conflicts with the downstream default tokens.
* the token secret type is set to `Opaque`, instead of ServiceAccount type
* the token annotations for a `ServiceAccount` are removed

closes https://github.com/kcp-dev/kcp/issues/1062

related to: https://github.com/kcp-dev/kcp/issues/1058 https://github.com/kcp-dev/kcp/issues/280

### DIFF
--- a/pkg/syncer/spec/mutators/deployment.go
+++ b/pkg/syncer/spec/mutators/deployment.go
@@ -79,6 +79,10 @@ func (dm *DeploymentMutator) Mutate(obj *unstructured.Unstructured) error {
 	for _, secret := range secretList {
 		// Find the SA token that matches the service account name.
 		if val, ok := secret.GetAnnotations()[corev1.ServiceAccountNameKey]; ok && val == desiredServiceAccountName {
+			if desiredServiceAccountName == "default" {
+				desiredSecretName = "kcp-" + secret.GetName()
+				break
+			}
 			desiredSecretName = secret.GetName()
 			break
 		}
--- a/pkg/syncer/spec/mutators/deployment.go
+++ b/pkg/syncer/spec/mutators/deployment.go
@@ -17,7 +17,10 @@ limitations under the License.
 package mutators
 
 import (
+	"fmt"
 	"net/url"
+
+	"github.com/kcp-dev/logicalcluster"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -27,8 +30,11 @@ import (
 	utilspointer "k8s.io/utils/pointer"
 )
 
+type ListSecretFunc func(clusterName logicalcluster.Name, namespace string) ([]*unstructured.Unstructured, error)
+
 type DeploymentMutator struct {
 	upstreamURL *url.URL
+	listSecrets ListSecretFunc
 }
 
 func (dm *DeploymentMutator) GVR() schema.GroupVersionResource {
@@ -39,34 +45,54 @@ func (dm *DeploymentMutator) GVR() schema.GroupVersionResource {
 	}
 }
 
-func NewDeploymentMutator(upstreamURL *url.URL) *DeploymentMutator {
+func NewDeploymentMutator(upstreamURL *url.URL, secretLister ListSecretFunc) *DeploymentMutator {
 	return &DeploymentMutator{
 		upstreamURL: upstreamURL,
+		listSecrets: secretLister,
 	}
 }
 
 // Mutate applies the mutator changes to the object.
-func (dm *DeploymentMutator) Mutate(downstreamObj *unstructured.Unstructured) error {
+func (dm *DeploymentMutator) Mutate(obj *unstructured.Unstructured) error {
 	var deployment appsv1.Deployment
 	err := runtime.DefaultUnstructuredConverter.FromUnstructured(
-		downstreamObj.UnstructuredContent(),
+		obj.UnstructuredContent(),
 		&deployment)
 	if err != nil {
 		return err
 	}
+	upstreamLogicalName := logicalcluster.From(obj)
 
-	// If the deployment has no serviceaccount defined or it is "default", that means that it is using the default one
-	// and we need will need to override it with our own, "kcp-default"
-	//
-	// If the deployment has a serviceaccount defined, that means we need to synchronize the created service account
-	// down to the workloadcluster, so we don't modify it, and expect the scheduler to do the job.
-	//
-	// Alias for the template.spec to improve readability.
 	templateSpec := &deployment.Spec.Template.Spec
+
+	desiredServiceAccountName := "default"
+	if templateSpec.ServiceAccountName != "" && templateSpec.ServiceAccountName != "default" {
+		desiredServiceAccountName = templateSpec.ServiceAccountName
+	}
+
+	secretList, err := dm.listSecrets(upstreamLogicalName, deployment.Namespace)
+	if err != nil {
+		return fmt.Errorf("error listing secrets for workspace %s: %w", upstreamLogicalName.String(), err)
+	}
+
+	desiredSecretName := ""
+	for _, secret := range secretList {
+		// Find the SA token that matches the service account name.
+		if val, ok := secret.GetAnnotations()[corev1.ServiceAccountNameKey]; ok && val == desiredServiceAccountName {
+			desiredSecretName = secret.GetName()
+			break
+		}
+	}
+
+	if desiredSecretName == "" {
+		return fmt.Errorf("couldn't find a token upstream for the serviceaccount %s/%s in workspace %s", desiredServiceAccountName, deployment.Namespace, upstreamLogicalName.String())
+	}
 
 	// Setting AutomountServiceAccountToken to false allow us to control the ServiceAccount
 	// VolumeMount and Volume definitions.
 	templateSpec.AutomountServiceAccountToken = utilspointer.BoolPtr(false)
+	// Set to empty the serviceAccountName on podTemplate as we are not syncing the serviceAccount down to the workload cluster.
+	templateSpec.ServiceAccountName = ""
 
 	kcpExternalHost := dm.upstreamURL.Hostname()
 	kcpExternalPort := dm.upstreamURL.Port()
@@ -94,13 +120,9 @@ func (dm *DeploymentMutator) Mutate(downstreamObj *unstructured.Unstructured) er
 				DefaultMode: utilspointer.Int32Ptr(420),
 				Sources: []corev1.VolumeProjection{
 					{
-						// TODO(jmprusi): Investigate if instead of using a secret directly we should use the serviceaccount
-						//                in order to get a bound token. We will need to investigate this
-						//                as the pcluster keeps rewriting the serviceaccount and its tokens values rendering
-						//                them non-valid for KCP. (Also it removes the ClusterName included in the JWT token)
 						Secret: &corev1.SecretProjection{
 							LocalObjectReference: corev1.LocalObjectReference{
-								Name: "kcp-default-token",
+								Name: desiredSecretName,
 							},
 							Items: []corev1.KeyToPath{
 								{
@@ -174,7 +196,7 @@ func (dm *DeploymentMutator) Mutate(downstreamObj *unstructured.Unstructured) er
 	}
 
 	// Set the changes back into the obj.
-	downstreamObj.SetUnstructuredContent(unstructured)
+	obj.SetUnstructuredContent(unstructured)
 
 	return nil
 }
--- a/pkg/syncer/spec/mutators/deployment.go
+++ b/pkg/syncer/spec/mutators/deployment.go
@@ -19,6 +19,7 @@ package mutators
 import (
 	"fmt"
 	"net/url"
+	"sort"
 
 	"github.com/kcp-dev/logicalcluster"
 
@@ -74,6 +75,15 @@ func (dm *DeploymentMutator) Mutate(obj *unstructured.Unstructured) error {
 	if err != nil {
 		return fmt.Errorf("error listing secrets for workspace %s: %w", upstreamLogicalName.String(), err)
 	}
+
+	// In order to avoid triggering a deployment update on resyncs, we need to make sure that the list
+	// of secrets is sorted by creationTimsestamp. So if the user creates a new token for a given serviceaccount
+	// the first one will be picked always.
+	sort.Slice(secretList, func(i, j int) bool {
+		iCreationTimestamp := secretList[i].GetCreationTimestamp()
+		jCreationTimestamp := secretList[j].GetCreationTimestamp()
+		return iCreationTimestamp.Before(&jCreationTimestamp)
+	})
 
 	desiredSecretName := ""
 	for _, secret := range secretList {
--- a/pkg/syncer/spec/mutators/deployment_test.go
+++ b/pkg/syncer/spec/mutators/deployment_test.go
@@ -22,6 +22,7 @@ import (
 	"net/url"
 	"testing"
 
+	"github.com/kcp-dev/logicalcluster"
 	"github.com/stretchr/testify/require"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -43,7 +44,7 @@ var kcpApiAccessVolume = corev1.Volume{
 				{
 					Secret: &corev1.SecretProjection{
 						LocalObjectReference: corev1.LocalObjectReference{
-							Name: "kcp-default-token",
+							Name: "default-token-1234",
 						},
 						Items: []corev1.KeyToPath{
 							{
@@ -81,20 +82,43 @@ var kcpApiAccessVolumeMount = corev1.VolumeMount{
 	ReadOnly:  true,
 }
 
-func TestMutate(t *testing.T) {
+func TestDeploymentMutate(t *testing.T) {
 	for _, c := range []struct {
 		desc                                   string
+		upstreamSecrets                        []*corev1.Secret
 		originalDeployment, expectedDeployment *appsv1.Deployment
 		config                                 *rest.Config
 	}{{
 		desc: "Deployment without Envs or volumes is mutated.",
+		upstreamSecrets: []*corev1.Secret{
+			{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "Secret",
+					APIVersion: "v1",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        "default-token-1234",
+					Namespace:   "namespace",
+					ClusterName: "root:default:testing",
+					Annotations: map[string]string{
+						"kubernetes.io/service-account.name": "default",
+					},
+				},
+				Data: map[string][]byte{
+					"token":     []byte("token"),
+					"namespace": []byte("namespace"),
+				},
+			},
+		},
 		originalDeployment: &appsv1.Deployment{
 			TypeMeta: metav1.TypeMeta{
 				Kind:       "Deployment",
 				APIVersion: "apps/v1",
 			},
 			ObjectMeta: metav1.ObjectMeta{
-				Name: "test-deployment",
+				Name:        "test-deployment",
+				Namespace:   "namespace",
+				ClusterName: "root:default:testing",
 			},
 			Spec: appsv1.DeploymentSpec{
 				Replicas: new(int32),
@@ -116,7 +140,9 @@ func TestMutate(t *testing.T) {
 				APIVersion: "apps/v1",
 			},
 			ObjectMeta: metav1.ObjectMeta{
-				Name: "test-deployment",
+				Name:        "test-deployment",
+				Namespace:   "namespace",
+				ClusterName: "root:default:testing",
 			},
 			Spec: appsv1.DeploymentSpec{
 				Replicas: new(int32),
@@ -158,13 +184,35 @@ func TestMutate(t *testing.T) {
 		},
 	}, {
 		desc: "Deployment with one env var gets mutated but the already existing env var remains the same",
+		upstreamSecrets: []*corev1.Secret{
+			{
+				TypeMeta: metav1.TypeMeta{
+					Kind:       "Secret",
+					APIVersion: "v1",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        "default-token-1234",
+					Namespace:   "namespace",
+					ClusterName: "root:default:testing",
+					Annotations: map[string]string{
+						"kubernetes.io/service-account.name": "default",
+					},
+				},
+				Data: map[string][]byte{
+					"token":     []byte("token"),
+					"namespace": []byte("namespace"),
+				},
+			},
+		},
 		originalDeployment: &appsv1.Deployment{
 			TypeMeta: metav1.TypeMeta{
 				Kind:       "Deployment",
 				APIVersion: "apps/v1",
 			},
 			ObjectMeta: metav1.ObjectMeta{
-				Name: "test-deployment",
+				Name:        "test-deployment",
+				Namespace:   "namespace",
+				ClusterName: "root:default:testing",
 			},
 			Spec: appsv1.DeploymentSpec{
 				Replicas: new(int32),
@@ -192,7 +240,9 @@ func TestMutate(t *testing.T) {
 				APIVersion: "apps/v1",
 			},
 			ObjectMeta: metav1.ObjectMeta{
-				Name: "test-deployment",
+				Name:        "test-deployment",
+				Namespace:   "namespace",
+				ClusterName: "root:default:testing",
 			},
 			Spec: appsv1.DeploymentSpec{
 				Replicas: new(int32),
@@ -238,13 +288,35 @@ func TestMutate(t *testing.T) {
 		},
 	},
 		{desc: "Deployment with an env var named KUBERNETES_SERVICE_PORT gets mutated and it is overridden and not duplicated",
+			upstreamSecrets: []*corev1.Secret{
+				{
+					TypeMeta: metav1.TypeMeta{
+						Kind:       "Secret",
+						APIVersion: "v1",
+					},
+					ObjectMeta: metav1.ObjectMeta{
+						Name:        "default-token-1234",
+						Namespace:   "namespace",
+						ClusterName: "root:default:testing",
+						Annotations: map[string]string{
+							"kubernetes.io/service-account.name": "default",
+						},
+					},
+					Data: map[string][]byte{
+						"token":     []byte("token"),
+						"namespace": []byte("namespace"),
+					},
+				},
+			},
 			originalDeployment: &appsv1.Deployment{
 				TypeMeta: metav1.TypeMeta{
 					Kind:       "Deployment",
 					APIVersion: "apps/v1",
 				},
 				ObjectMeta: metav1.ObjectMeta{
-					Name: "test-deployment",
+					Name:        "test-deployment",
+					Namespace:   "namespace",
+					ClusterName: "root:default:testing",
 				},
 				Spec: appsv1.DeploymentSpec{
 					Replicas: new(int32),
@@ -272,7 +344,9 @@ func TestMutate(t *testing.T) {
 					APIVersion: "apps/v1",
 				},
 				ObjectMeta: metav1.ObjectMeta{
-					Name: "test-deployment",
+					Name:        "test-deployment",
+					Namespace:   "namespace",
+					ClusterName: "root:default:testing",
 				},
 				Spec: appsv1.DeploymentSpec{
 					Replicas: new(int32),
@@ -313,13 +387,35 @@ func TestMutate(t *testing.T) {
 				Host: "https://4.5.6.7:12345",
 			}},
 		{desc: "Deployment with an existing VolumeMount named kcp-api-access gets mutated and it is overridden and not duplicated",
+			upstreamSecrets: []*corev1.Secret{
+				{
+					TypeMeta: metav1.TypeMeta{
+						Kind:       "Secret",
+						APIVersion: "v1",
+					},
+					ObjectMeta: metav1.ObjectMeta{
+						Name:        "default-token-1234",
+						Namespace:   "namespace",
+						ClusterName: "root:default:testing",
+						Annotations: map[string]string{
+							"kubernetes.io/service-account.name": "default",
+						},
+					},
+					Data: map[string][]byte{
+						"token":     []byte("token"),
+						"namespace": []byte("namespace"),
+					},
+				},
+			},
 			originalDeployment: &appsv1.Deployment{
 				TypeMeta: metav1.TypeMeta{
 					Kind:       "Deployment",
 					APIVersion: "apps/v1",
 				},
 				ObjectMeta: metav1.ObjectMeta{
-					Name: "test-deployment",
+					Name:        "test-deployment",
+					Namespace:   "namespace",
+					ClusterName: "root:default:testing",
 				},
 				Spec: appsv1.DeploymentSpec{
 					Replicas: new(int32),
@@ -354,7 +450,9 @@ func TestMutate(t *testing.T) {
 					APIVersion: "apps/v1",
 				},
 				ObjectMeta: metav1.ObjectMeta{
-					Name: "test-deployment",
+					Name:        "test-deployment",
+					Namespace:   "namespace",
+					ClusterName: "root:default:testing",
 				},
 				Spec: appsv1.DeploymentSpec{
 					Replicas: new(int32),
@@ -394,15 +492,36 @@ func TestMutate(t *testing.T) {
 			config: &rest.Config{
 				Host: "https://4.5.6.7:12345",
 			}},
-
 		{desc: "Deployment with an existing Volume named kcp-api-access gets mutated and it is overridden and not duplicated",
+			upstreamSecrets: []*corev1.Secret{
+				{
+					TypeMeta: metav1.TypeMeta{
+						Kind:       "Secret",
+						APIVersion: "v1",
+					},
+					ObjectMeta: metav1.ObjectMeta{
+						Name:        "default-token-1234",
+						Namespace:   "namespace",
+						ClusterName: "root:default:testing",
+						Annotations: map[string]string{
+							"kubernetes.io/service-account.name": "default",
+						},
+					},
+					Data: map[string][]byte{
+						"token":     []byte("token"),
+						"namespace": []byte("namespace"),
+					},
+				},
+			},
 			originalDeployment: &appsv1.Deployment{
 				TypeMeta: metav1.TypeMeta{
 					Kind:       "Deployment",
 					APIVersion: "apps/v1",
 				},
 				ObjectMeta: metav1.ObjectMeta{
-					Name: "test-deployment",
+					Name:        "test-deployment",
+					Namespace:   "namespace",
+					ClusterName: "root:default:testing",
 				},
 				Spec: appsv1.DeploymentSpec{
 					Replicas: new(int32),
@@ -447,7 +566,9 @@ func TestMutate(t *testing.T) {
 					APIVersion: "apps/v1",
 				},
 				ObjectMeta: metav1.ObjectMeta{
-					Name: "test-deployment",
+					Name:        "test-deployment",
+					Namespace:   "namespace",
+					ClusterName: "root:default:testing",
 				},
 				Spec: appsv1.DeploymentSpec{
 					Replicas: new(int32),
@@ -492,7 +613,17 @@ func TestMutate(t *testing.T) {
 			t.Run(c.desc, func(t *testing.T) {
 				upstreamURL, err := url.Parse(c.config.Host)
 				require.NoError(t, err)
-				dm := NewDeploymentMutator(upstreamURL)
+
+				dm := NewDeploymentMutator(upstreamURL, func(upstreamLogicalCluster logicalcluster.Name, namespace string) ([]*unstructured.Unstructured, error) {
+					unstructuredObjects := make([]*unstructured.Unstructured, 0, len(c.upstreamSecrets))
+					for _, obj := range c.upstreamSecrets {
+						unstObj, err := toUnstructured(obj)
+						require.NoError(t, err)
+						unstructuredObjects = append(unstructuredObjects, unstObj)
+					}
+					return unstructuredObjects, nil
+				})
+
 				unstrOriginalDeployment, err := toUnstructured(c.originalDeployment)
 				require.NoError(t, err, "toRuntimeObject() = %v", err)
 
--- a/pkg/syncer/spec/mutators/deployment_test.go
+++ b/pkg/syncer/spec/mutators/deployment_test.go
@@ -44,7 +44,7 @@ var kcpApiAccessVolume = corev1.Volume{
 				{
 					Secret: &corev1.SecretProjection{
 						LocalObjectReference: corev1.LocalObjectReference{
-							Name: "default-token-1234",
+							Name: "kcp-default-token-1234",
 						},
 						Items: []corev1.KeyToPath{
 							{
--- a/pkg/syncer/spec/mutators/secrets.go
+++ b/pkg/syncer/spec/mutators/secrets.go
@@ -19,7 +19,6 @@ package mutators
 import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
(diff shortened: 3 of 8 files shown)


stevekuznetsov (commented):
Implementation seems good, comments on testing.

pkg/syncer/spec/mutators/secrets.go
```diff
@@ -49,8 +51,13 @@ func (sm *SecretMutator) Mutate(downstreamObj *unstructured.Unstructured) error
 	}
 
 	// We need to transform the default kcp token into an Opaque secret, in order to avoid the pcluster to rewrite it.
-	if secret.Name == "kcp-default-token" {
+	// and we remove the annotations that point to the kcp serviceaccount name/uid
+	if strings.HasPrefix(secret.Name, "kcp-default-token-") {
 		secret.Type = corev1.SecretTypeOpaque
+		annotations := secret.GetAnnotations()
+		delete(annotations, "kubernetes.io/service-account.name")
```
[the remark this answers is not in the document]
jmprusi: done

test/e2e/syncer/syncer_test.go
```diff
@@ -88,21 +89,38 @@ func TestSyncerLifecycle(t *testing.T) {
 	require.NoError(t, err)
 
 	// TODO(marun) The name mapping should be defined for reuse outside of the transformName method in pkg/syncer
-	secretName := "kcp-default-token"
-	configMapName := "kcp-root-ca.crt"
+	var downstreamSecretName string
+	t.Logf("Waiting for upstream service account secret to be created...")
```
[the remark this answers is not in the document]
jmprusi: I see value on understanding what part is broken, configmap/secret/volume, makes it easier to debug tests... So I will go with the first option, extract the secretName from the downstream Deployment.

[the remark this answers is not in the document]
jmprusi: Ok, let me try

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -64,9 +70,50 @@ func (dm *DeploymentMutator) Mutate(downstreamObj *unstructured.Unstructured) er
 	// Alias for the template.spec to improve readability.
 	templateSpec := &deployment.Spec.Template.Spec
 
+	desiredServiceAccountName := "default"
+	if templateSpec.ServiceAccountName != "" && templateSpec.ServiceAccountName != "default" {
+		desiredServiceAccountName = templateSpec.ServiceAccountName
+	}
+
+	secretList, err := dm.secretLister.List(labels.Everything())
```
ncdc, line 78: dm.secretLister comes from the upstream (kcp) informer, so we need to exclude secrets for other logical clusters. We're still working on scoped listers, so in the meantime, I'd recommend adding an indexer on secrets to index by cluster name, then list from the secret indexer (not lister) using the custom index.

jmprusi: can you take a look at the implementation? thanks!

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -64,9 +75,50 @@ func (dm *DeploymentMutator) Mutate(downstreamObj *unstructured.Unstructured) er
...
 
+	desiredServiceAccountName := "default"
+	if templateSpec.ServiceAccountName != "" && templateSpec.ServiceAccountName != "default" {
+		desiredServiceAccountName = templateSpec.ServiceAccountName
+	}
+
+	secretList, err := dm.secretLister.ByIndex(dm.indexName, dm.clusterName.String())
+	if err != nil {
+		klog.Errorf("Failed to list secrets: %v", err)
+		return err
```
ncdc, line 86: We generally want to log or return, but not both. (There are some times where it's valid to do both, but they're pretty limited.) I would recommend `return fmt.Errorf("error listing secrets for cluster %s: %w", dm.clusterName, err)`

pkg/syncer/spec/spec_process.go
```diff
@@ -351,10 +352,10 @@ func transformName(syncedObject *unstructured.Unstructured) {
 	if syncedObject.GroupVersionKind() == configMapGVR && syncedObject.GetName() == "kube-root-ca.crt" {
 		syncedObject.SetName("kcp-root-ca.crt")
 	}
-	// TODO(jmprusi): We are rewriting the name of the object into a non random one so we can reference it from the deployment transformer
-	//                but this means that means than more than one default-token-XXXX object will overwrite the same "kcp-default-token"
-	//				  object. This must be fixed.
-	if syncedObject.GroupVersionKind() == secretGVR && strings.Contains(syncedObject.GetName(), "default-token-") {
-		syncedObject.SetName("kcp-default-token")
+	if syncedObject.GroupVersionKind() == secretGVR && strings.Contains(syncedObject.GetName(), "-token-") {
```
ncdc, line 355: Let's remove the check for `-token-`. If it's a service account secret, it will have the annotation - just check for that.

pkg/syncer/spec/spec_process.go
```diff
@@ -351,10 +352,10 @@ func transformName(syncedObject *unstructured.Unstructured) {
...
+		// Only transform the name if the object is service-account token.
+		if _, ok := syncedObject.GetAnnotations()[core.ServiceAccountNameKey]; ok {
+			syncedObject.SetName("kcp-" + syncedObject.GetName())
```
ncdc, line 358: Could you please remind me why we need to change the name of the token secret in the pcluster?

test/e2e/syncer/syncer_test.go
```diff
@@ -151,7 +121,12 @@ func TestSyncerLifecycle(t *testing.T) {
 			deployment, err = downstreamKubeClient.AppsV1().Deployments(downstreamNamespaceName).Get(ctx, upstreamDeployment.Name, metav1.GetOptions{})
 			require.NoError(t, err)
 			klog.Info(toYaml(deployment))
-			return expectedAvailableReplicas == deployment.Status.AvailableReplicas
+			if expectedAvailableReplicas == deployment.Status.AvailableReplicas {
+				return true
+			}
+			dumpPodEvents(downstreamKubeClient, downstreamNamespaceName, ctx, t)
```
ncdc, line 127: Curious: what were you debugging here?

pkg/syncer/spec/spec_process.go
jmprusi: Ok, that's something we should discuss:

As you know, KCP and Kubernetes create a "default" service account on each namespace (this is going to change in the next k8s release afaik), and that service account has a token created like "default-token-XXXX" where "XXXX" is a random value.

So, what I do here is to add the "kcp-" suffix to the name of the token, to make it easier to be distinguished from the pcluster default tokens, and also, eliminating the possible "conflict" of names (perhaps, this is slightly farfetched).

One possibility would be to just rename the default token, and don't rename any other token created by the user on kcp, but this means controlling for the "special" case (default-token-XXXX)... 

The other idea is just, to stop renaming any token, and hope🤞🏼 we don't have conflicts downstream.

test/e2e/syncer/syncer_test.go
jmprusi: this is related to the previous discussion https://github.com/kcp-dev/kcp/pull/1127#discussion_r879707982 with @stevekuznetsov 

log the pod events in order to detect why the deployment has not the desired number of replicas, this way we don't have to wait for configmaps/tokens to be replicated and we can detect other situations where the deployment is failing.

pkg/syncer/spec/mutators/deployment_test.go
```diff
@@ -510,6 +592,74 @@ func TestMutate(t *testing.T) {
 	}
 }
 
+type fakeIndexer struct {
```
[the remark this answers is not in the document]
jmprusi: done

pkg/syncer/spec/spec_process.go
jmprusi: done

pkg/syncer/spec/spec_process.go
ncdc: I think there is a risk that we could have a conflict (at least as long as automatic token secret creation is enabled, which is pre Kubernetes 1.24), so it's probably best to do some renaming.

We also need to make sure that the renamed name does not exceed the maximum allowed name length for a secret, which follows the DNS subdomain rules (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names).

The most foolproof thing is probably to try to prepend `kcp-`, but if that exceeds the maximum allowed length, then do `kcp-<original value minus 8 characters>-<7 character hash of the original name>` or something along those lines. WDYT @sttts @jmprusi?

Also, could you please create a helper function to do the name transformation, and use it both here and in the deployment mutator?

pkg/syncer/spec/mutators/secrets.go
```diff
@@ -48,18 +49,23 @@ func (sm *SecretMutator) Mutate(downstreamObj *unstructured.Unstructured) error
 		return err
 	}
 
-	// We need to transform the default kcp token into an Opaque secret, in order to avoid the pcluster to rewrite it.
-	if secret.Name == "kcp-default-token" {
+	// We need transform the serviceaccount tokens into an Opaque secret, in order to avoid the pcluster to rewrite them,
```
ncdc, line 52: Can you add unit test(s) for this file?

pkg/syncer/spec/spec_process.go
jmprusi: Makes sense, on it.

jmprusi: done! please take a look at it

pkg/syncer/spec/spec_process_test.go
```diff
@@ -484,9 +484,18 @@ func TestSyncerProcess(t *testing.T) {
 				"state.internal.workloads.kcp.dev/us-west1": "Sync",
 			}, nil),
 			gvr: schema.GroupVersionResource{Group: "apps", Version: "v1", Resource: "deployments"},
-			fromResource: deployment("theDeployment", "test", "root:org:ws", map[string]string{
-				"state.internal.workloads.kcp.dev/us-west1": "Sync",
-			}, nil, nil),
+			fromResources: []runtime.Object{
```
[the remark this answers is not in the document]
jmprusi: I don't have a strong opinion, so we can change that, but can we do it in another PR? :)

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -222,3 +272,16 @@ func updateVolumeMount(volumeMounts []corev1.VolumeMount, overrideVolumeMount co
 
 	return volumeMounts
 }
+
+// PrefixSecretName adds a prefix to the secret name to avoid conflicts while handling the max length of 63 characters.
```
ncdc, line 276: Isn't the max length 253?

pkg/syncer/spec/mutators/secrets.go
jmprusi: done

pkg/syncer/spec/mutators/deployment.go
jmprusi: done

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -222,3 +272,16 @@ func updateVolumeMount(volumeMounts []corev1.VolumeMount, overrideVolumeMount co
...
+func PrefixSecretName(name string) string {
+	prefixedSecretName := secretNamePrefix + name
+	if len(prefixedSecretName) > 253 {
```
ncdc, line 279: It's probably better to use `validation.DNS1123SubdomainMaxLength`

pkg/syncer/spec/mutators/deployment.go
ncdc: Comment still says 63

pkg/syncer/spec/mutators/deployment_test.go
```diff
@@ -510,6 +619,45 @@ func TestMutate(t *testing.T) {
...
+		expectedName string
+	}{
+		{
+			desc:         "short name, we just expect it to be prefixed",
+			name:         "test-secret",
+			expectedName: "kcp-test-secret",
+		},
+		{
+			desc:         "long name of 253, expect a prefixed valid name of 253 characters",
+			name:         "ke2wlbx5hrmx41fu0ak55a3d0p81dasqcbzk0fz0p50t572aimct5k7d3f029196g5ryx54je6v09evvccclblg2aqdhvkfemkfvxxhrn66wj6hzvgfsmdb79bqaavmdp5kqzx8bx90pwp9n8hazvqzrbcbyrx77kyldyzacoqh7e7o2tvz3eja9spjn2f5opyxg0ui804548v9dh03zbgt57p1ddbeur8z6z017eu4vmff8kppj79ugkxb4qniva79ggetl7ykshf0nfbhpuu8sjlorym9bkh87h71lyxmf0t7j8ujsa881v53apwh3yc25dud17mps00sztziktres",
```
ncdc, line 635: Can you make this 253 chars?

test/e2e/syncer/syncer_test.go
```diff
@@ -238,8 +213,18 @@ func TestSyncerLifecycle(t *testing.T) {
 			return false
 		}
 		return true
-	}, wait.ForeverTestTimeout, time.Millisecond*100, "downstream deployment %s/%s was not synced", downstreamNamespaceName, upstreamDeployment.Name)
+	}, 60*time.Second, time.Millisecond*100, "downstream deployment %s/%s was not synced", downstreamNamespaceName, upstreamDeployment.Name)
```
ncdc, line 216: Why'd you change this to 60 seconds?

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -222,3 +272,16 @@ func updateVolumeMount(volumeMounts []corev1.VolumeMount, overrideVolumeMount co
...
+		// Hash the original secret name to avoid conflicts with the shortened secret name.
+		hashFromName := sha256.Sum224([]byte(name))
+		// build the new secret name with the shortened secret name and the hash of the original secret name, but we
+		// only get the first 6 characters of the hash (1bytes = 2chars), this should get us a max length of 253.
```
ncdc, line 283: Do we need more of the hash to minimize the likelihood of any collision? cc @stevekuznetsov

test/e2e/syncer/syncer_test.go
jmprusi: it's a wip commit, im trying to find why some tests are flaky

pkg/syncer/spec/mutators/deployment.go
jmprusi: fixed

test/e2e/syncer/syncer_test.go
ncdc: The flakes are most likely unrelated. I think they came from #1082.

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -17,18 +17,30 @@ limitations under the License.
...
 )
 
+const secretNamePrefix = "kcp-"
+
 type DeploymentMutator struct {
-	upstreamURL *url.URL
+	clusterName  logicalcluster.Name
+	upstreamURL  *url.URL
+	secretLister cache.Indexer
+	indexName    string
```
sttts, line 43: what is this index? byWorkspace?

pkg/syncer/spec/mutators/secrets.go
```diff
@@ -48,18 +49,23 @@ func (sm *SecretMutator) Mutate(downstreamObj *unstructured.Unstructured) error
...
+	// and we remove the annotations that point to the kcp serviceaccount name/uid.
+	if _, ok := secret.Annotations[core.ServiceAccountNameKey]; ok {
```
sttts, line 54: why decode the secret? Nothing in here needs that.

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -39,9 +51,12 @@ func (dm *DeploymentMutator) GVR() schema.GroupVersionResource {
 	}
 }
 
-func NewDeploymentMutator(upstreamURL *url.URL) *DeploymentMutator {
+func NewDeploymentMutator(upstreamURL *url.URL, secretLister cache.Indexer, clusterName logicalcluster.Name, indexName string) *DeploymentMutator {
 	return &DeploymentMutator{
-		upstreamURL: upstreamURL,
+		clusterName:  clusterName,
```
sttts, line 56: why store a clusterName. The downstream object has that info, doesn't it?

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -64,9 +79,49 @@ func (dm *DeploymentMutator) Mutate(downstreamObj *unstructured.Unstructured) er
...
+
+	secretList, err := dm.secretLister.ByIndex(dm.indexName, dm.clusterName.String())
+	if err != nil {
+		return fmt.Errorf("error listing secrets for cluster %s: %w", dm.clusterName, err)
+	}
+
+	desiredSecretName := ""
+	for _, secretObj := range secretList {
+		var secret *corev1.Secret
+		switch obj := secretObj.(type) {
```
sttts, line 95: not needed. A lister always returns the right objects.

pkg/syncer/spec/mutators/deployment.go
sttts: Collision propability with 6 alpha-numerical-case-sensitive chars:

Bildschirmfoto 2022-05-27 um 13 09 25

sttts: with 10 secrets

pkg/syncer/spec/spec_controller.go
```diff
@@ -157,6 +151,22 @@ func NewSpecSyncer(gvrs []schema.GroupVersionResource, workloadClusterLogicalClu
 		klog.V(2).InfoS("Set up downstream informer", "clusterName", workloadClusterLogicalClusterName, "pcluster", workloadClusterName, "gvr", gvr.String())
 	}
 
+	secretMutator := specmutators.NewSecretMutator()
+
+	secretIndexer := upstreamInformers.ForResource(schema.GroupVersionResource{Group: "", Version: "v1", Resource: "secrets"}).Informer().GetIndexer()
+	deploymentMutator := specmutators.NewDeploymentMutator(upstreamURL, secretIndexer, workloadClusterLogicalClusterName, byWorkspace)
```
sttts, line 157: am against this. Why do we break the patterns we have prepared in the code and build one-off code again?

sttts, line 157: why pass the indexer? Just make the index func a convention in the syncer. We will get scoping anyway soon and it will go away.

sttts: workloadClusterLogicalClusterName seems to be wrong. This is about user secrets, not about the WorkloadCluster.

pkg/syncer/spec/mutators/deployment.go
sttts: The indexers we have here are dynamic indexers, aren't they? So always unstructured.

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -64,9 +79,49 @@ func (dm *DeploymentMutator) Mutate(downstreamObj *unstructured.Unstructured) er
...
+		case *corev1.Secret:
+			secret = obj
+		case *unstructured.Unstructured:
+			unstructured := secretObj.(*unstructured.Unstructured).UnstructuredContent()
+			err = runtime.DefaultUnstructuredConverter.
```
sttts, line 100: look for the annotation first before decoding.

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -64,9 +79,49 @@ func (dm *DeploymentMutator) Mutate(downstreamObj *unstructured.Unstructured) er
...
+		}
+
+		// Find the SA token that matches the service account name.
+		if val, ok := secret.Annotations[corev1.ServiceAccountNameKey]; ok && val == desiredServiceAccountName {
+			desiredSecretName = PrefixSecretName(secret.Name)
+		}
+	}
+
+	if desiredSecretName == "" {
+		return fmt.Errorf("couldn't find a token upstream for the serviceaccount %s", desiredServiceAccountName)
```
sttts, line 117: qualify with namespace or whatever else is helpful here.

pkg/syncer/spec/mutators/deployment_test.go
```diff
@@ -490,9 +578,31 @@ func TestMutate(t *testing.T) {
...
 				upstreamURL, err := url.Parse(c.config.Host)
 				require.NoError(t, err)
-				dm := NewDeploymentMutator(upstreamURL)
+				fakeClient := fake.NewSimpleClientset(toObjects(c.upstreamSecrets)...)
+				sharedInformer := informers.NewSharedInformerFactory(fakeClient, 0)
+				informer, err := sharedInformer.ForResource(schema.GroupVersionResource{Version: "v1", Resource: "secrets"})
+				require.NoError(t, err)
+				indexer := informer.Informer().GetIndexer()
+				err = indexer.AddIndexers(cache.Indexers{
+					"by-workspace": func(obj interface{}) ([]string, error) {
```
sttts, line 592: use the existing index func byWorkspace

pkg/syncer/spec/mutators/deployment_test.go
```diff
@@ -510,6 +620,45 @@ func TestMutate(t *testing.T) {
...
+		{
+			desc:         "long name of 253, expect a prefixed valid name of 253 characters",
+			name:         "ke2wlbx5hrmx41fu0ak55a3d0p81dasqcbzka0fz0p50t572aimct5k7d3f029196g5ryx54je6v09evvccclblg2aqdhvkfemkfvxxhrn66wj6hzvgfsmdb79bqaavmdp5kqzx8bx90pwp9n8hazvqzrbcbyrx77kyldyzacoqh7e7o2tvz3eja9spjn2f5opyxg0ui804548v9dh03zbgt57p1ddbeur8z6z017eu4vmff8kppj79ugkxbs",
+			expectedName: "kcp-ke2wlbx5hrmx41fu0ak55a3d0p81dasqcbzka0fz0p50t572aimct5k7d3f029196g5ryx54je6v09evvccclblg2aqdhvkfemkfvxxhrn66wj6hzvgfsmdb79bqaavmdp5kqzx8bx90pwp9n8hazvqzrbcbyrx77kyldyzacoqh7e7o2tvz3eja9spjn2f5opyxg0ui804548v9dh03zbgt57p1ddbeur8z6z017eu4vmff8k-c08535",
+		},
+	} {
+		{
+			t.Run(c.desc, func(t *testing.T) {
+				actual := PrefixSecretName(c.name)
+				if !apiequality.Semantic.DeepEqual(actual, c.expectedName) {
```
sttts, line 643: deep-equal for strings?

sttts: even semantic deep-equal

pkg/syncer/spec/mutators/deployment_test.go
sttts: an idea: don't pass an index to the mutator, but pass a List func. That's is much easier to mock, no need for indexes and stuff and fake-clients.

pkg/syncer/spec/spec_process.go
```diff
@@ -351,10 +353,10 @@ func transformName(syncedObject *unstructured.Unstructured) {
 	if syncedObject.GroupVersionKind() == configMapGVR && syncedObject.GetName() == "kube-root-ca.crt" {
 		syncedObject.SetName("kcp-root-ca.crt")
 	}
-	// TODO(jmprusi): We are rewriting the name of the object into a non random one so we can reference it from the deployment transformer
-	//                but this means that means than more than one default-token-XXXX object will overwrite the same "kcp-default-token"
-	//				  object. This must be fixed.
-	if syncedObject.GroupVersionKind() == secretGVR && strings.Contains(syncedObject.GetName(), "default-token-") {
-		syncedObject.SetName("kcp-default-token")
+	if syncedObject.GroupVersionKind() == secretGVR {
```
sttts, line 356: GVR != GVK

pkg/syncer/spec/mutators/deployment.go
sttts: it's even wrong. There is no one clusterName anymore. The syncer speaks with the vw and that has many workspaces.

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -64,9 +79,49 @@ func (dm *DeploymentMutator) Mutate(downstreamObj *unstructured.Unstructured) er
...
+			if err != nil {
+				klog.Errorf("Failed to convert unstructured object to secret: %v", err)
+				return err
+			}
+		default:
+			return fmt.Errorf("unknown object type: %T", secretObj)
+		}
+
+		// Find the SA token that matches the service account name.
+		if val, ok := secret.Annotations[corev1.ServiceAccountNameKey]; ok && val == desiredServiceAccountName {
```
sttts, line 111: idea: an index by this annotation would turn the list+loop into one Get.

pkg/syncer/spec/mutators/secrets.go
```diff
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/kubernetes/pkg/apis/core"
```
sttts, line 24: don't use the internal api packages of kube, but k8s.io/api. Everything in the syncer must work against k8s.io/api and other non k8s.io/kubernetes repos.

pkg/syncer/spec/mutators/secrets.go
```diff
@@ -48,18 +49,23 @@ func (sm *SecretMutator) Mutate(downstreamObj *unstructured.Unstructured) error
...
-	// We need to transform the default kcp token into an Opaque secret, in order to avoid the pcluster to rewrite it.
-	if secret.Name == "kcp-default-token" {
+	// We need transform the serviceaccount tokens into an Opaque secret, in order to avoid the pcluster to rewrite them,
+	// and we remove the annotations that point to the kcp serviceaccount name/uid.
+	if _, ok := secret.Annotations[core.ServiceAccountNameKey]; ok {
 		secret.Type = corev1.SecretTypeOpaque
+		annotations := secret.GetAnnotations()
+		delete(annotations, core.ServiceAccountNameKey)
+		delete(annotations, core.ServiceAccountUIDKey)
+		secret.SetAnnotations(annotations)
```
sttts, line 59: if you use a secret (i.e. non-unstructured), you can mutate the Annotation directly. No need to get and set.

pkg/syncer/spec/mutators/secrets.go
jmprusi: done

pkg/syncer/spec/mutators/deployment_test.go
jmprusi: done

pkg/syncer/spec/mutators/deployment.go
jmprusi: I get a *corev1.Secret when it comes from tests, that's why I need to handle that case.

ncdc: @jmprusi is there any way to avoid doing that? We'd rather not have special casing main code that has portions that are only there to accommodate test behavior.

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -39,14 +51,15 @@ func (dm *DeploymentMutator) GVR() schema.GroupVersionResource {
...
 	return &DeploymentMutator{
-		upstreamURL: upstreamURL,
+		upstreamURL:  upstreamURL,
+		secretLister: upstreamSecretLister,
 	}
 }
 
 // Mutate applies the mutator changes to the object.
-func (dm *DeploymentMutator) Mutate(downstreamObj *unstructured.Unstructured) error {
+func (dm *DeploymentMutator) Mutate(downstreamObj *unstructured.Unstructured, upstreamLogicalName logicalcluster.Name) error {
```
sttts, line 62: upstreamClusterName

pkg/syncer/spec/mutators/deployment.go
sttts: still open

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -222,3 +256,16 @@ func updateVolumeMount(volumeMounts []corev1.VolumeMount, overrideVolumeMount co
 
 	return volumeMounts
 }
+
+// PrefixSecretName adds a prefix to the secret name to avoid conflicts while handling the max length of 63 characters.
+func PrefixSecretName(name string) string {
```
sttts, line 261: `SecretNamePrefix` – no verb for side-effect free funcs

sttts: Why do we add a prefix and not a suffix as it is common in Kube?

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -222,3 +256,16 @@ func updateVolumeMount(volumeMounts []corev1.VolumeMount, overrideVolumeMount co
...
+	prefixedSecretName := secretNamePrefix + name
+	if len(prefixedSecretName) > validation.DNS1123SubdomainMaxLength {
+		// Hash the original secret name to avoid conflicts with the shortened secret name.
+		hashFromName := sha256.Sum224([]byte(name))
```
sttts, line 265: I am confused. We have a name and hash it in some way to get a prefix? Which conflicts should this resolve? Be clear in the go doc. Is it with some downstream objects? Then we could just put a static suffix, no need for some magic hash.

pkg/syncer/spec/mutators/deployment_test.go
```diff
@@ -510,6 +609,45 @@ func TestMutate(t *testing.T) {
...
+				}
+			})
+		}
+	}
+}
+
+func toObjects(secrets []corev1.Secret) []runtime.Object {
+	objs := make([]runtime.Object, 0, len(secrets))
+	for _, secret := range secrets {
+		objs = append(objs, secret.DeepCopy())
```
sttts, line 646: no random deepcopy in some helper please

pkg/syncer/spec/mutators/deployment_test.go
```diff
@@ -490,13 +567,35 @@ func TestMutate(t *testing.T) {
...
+				upstreamSecretIndexer := informer.Informer().GetIndexer()
+				err = upstreamSecretIndexer.AddIndexers(cache.Indexers{
+					ByWorkspaceIndexName: func(obj interface{}) ([]string, error) {
+						return []string{"root:default:testing"}, nil
+					},
+				})
+				require.NoError(t, err)
+
+				sharedInformer.Start(ctx.Done())
+				sharedInformer.WaitForCacheSync(ctx.Done())
```
sttts, line 592: commented before. get rid of all this informer mock code by putting a `List(...) []Secrets` func into the mutator, not a complete lister.

test/e2e/syncer/syncer_test.go
```diff
@@ -87,36 +87,6 @@ func TestSyncerLifecycle(t *testing.T) {
...
-		if apierrors.IsNotFound(err) {
-			return false
-		}
-		if err != nil {
-			t.Errorf("saw an error waiting for downstream configmap %s/%s to be created: %v", downstreamNamespaceName, configMapName, err)
-			return false
-		}
-		return true
-	}, wait.ForeverTestTimeout, time.Millisecond*100, "downstream configmap %s/%s was not created", downstreamNamespaceName, configMapName)
-
```
sttts, line 119: these removals surprise me. They exist to get early feedback in the e2e about broken syncing. Don't we still sync objects?

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -17,18 +17,30 @@ limitations under the License.
...
 
+const (
+	secretNamePrefix     = "kcp-"
+	ByWorkspaceIndexName = "syncer-spec-byWorkspace-index" // will go away with scoping
+)
+
 type DeploymentMutator struct {
-	upstreamURL *url.URL
+	upstreamURL  *url.URL
+	secretLister cache.Indexer
```
sttts, line 43: this is not a lister

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -64,9 +77,34 @@ func (dm *DeploymentMutator) Mutate(downstreamObj *unstructured.Unstructured) er
...
+	}
+
+	if desiredSecretName == "" {
+		return fmt.Errorf("couldn't find a token upstream for the serviceaccount %s", desiredServiceAccountName)
+	}
+
 	// Setting AutomountServiceAccountToken to false allow us to control the ServiceAccount
 	// VolumeMount and Volume definitions.
 	templateSpec.AutomountServiceAccountToken = utilspointer.BoolPtr(false)
+	// Set to empty the serviceAccountName on podTemplate
```
sttts, line 106: don't write what is done, but why

pkg/syncer/spec/mutators/deployment.go
sttts: please also put upstream or downstream into the name

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -64,9 +77,34 @@ func (dm *DeploymentMutator) Mutate(downstreamObj *unstructured.Unstructured) er
...
+	}
+
+	desiredSecretName := ""
+	for _, secretObj := range secretList {
+		secret := secretObj.(*unstructured.Unstructured)
+		// Find the SA token that matches the service account name.
+		if val, ok := secret.GetAnnotations()[corev1.ServiceAccountNameKey]; ok && val == desiredServiceAccountName {
+			desiredSecretName = PrefixSecretName(secret.GetName())
+		}
+	}
```
sttts, line 97: I am missing namespace logic. Why is the cluster name enough to find the right secret? Can't it be from another namespace?

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -64,9 +77,34 @@ func (dm *DeploymentMutator) Mutate(downstreamObj *unstructured.Unstructured) er
...
+	if err != nil {
+		return fmt.Errorf("error listing secrets for cluster %s: %w", deployment.ClusterName, err)
+	}
+
+	desiredSecretName := ""
+	for _, secretObj := range secretList {
+		secret := secretObj.(*unstructured.Unstructured)
+		// Find the SA token that matches the service account name.
+		if val, ok := secret.GetAnnotations()[corev1.ServiceAccountNameKey]; ok && val == desiredServiceAccountName {
+			desiredSecretName = PrefixSecretName(secret.GetName())
```
sttts, line 95: break?

test/e2e/syncer/syncer_test.go
jmprusi: part of the discussion here: https://github.com/kcp-dev/kcp/pull/1127#discussion_r879525593

pkg/syncer/spec/mutators/deployment.go
jmprusi: We can get rid of this, and just avoid prefixing all the secrets as the only ones that could be conflicting are the "default-token-XXXX" as those are the only secrets created on the downstream namespaces by default, and even that is going away with the next release of k8s, so the possibility of a conflict is almost non-existant.

If we don't prefix, we don't need to handle the >253 chars.

pkg/syncer/spec/mutators/deployment_test.go
jmprusi: done

pkg/syncer/spec/mutators/secrets.go
jmprusi: done

pkg/syncer/spec/mutators/deployment.go
jmprusi: done

pkg/syncer/spec/mutators/deployment.go
jmprusi: done

pkg/syncer/spec/mutators/deployment.go
jmprusi: done

pkg/syncer/spec/mutators/deployment_test.go
jmprusi: done

pkg/syncer/spec/mutators/deployment_test.go
jmprusi: done

test/e2e/syncer/syncer_test.go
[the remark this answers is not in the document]
sttts: We used to have e2e tests for the syncer that started with upstream object creation, then magic happened behind the scenes (or didn't 🤷‍♂️), and then the tests checked for the final end result of a running deployment. If those tests break, you are lost. Hence, we need intermediate steps for the whole process. So I disagree that these checks are not needed. Hours of debugging time was wasted in the past to understand when something went wrong, e.g. during flakes.

I agree with @stevekuznetsov's comment that we should choose intermediate checks through that are not fragile.

test/e2e/syncer/syncer_test.go
sttts: See https://github.com/kcp-dev/kcp/pull/1127#discussion_r885907503. tl/dr: I disagree that these checks are useless. Instead make them non-fragile, but keep them. The machinery behind the scenes is heavy and if it breaks you are lost without having intermediate steps.

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -17,18 +17,30 @@ limitations under the License.
...
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/tools/cache"
 	utilspointer "k8s.io/utils/pointer"
 )
 
+const (
+	ByWorkspaceAndNamespaceIndexName = "syncer-spec-WorkspaceNamespace-index" // will go away with scoping
+)
```
sttts, line 37: as NewSecretLister, move it where the mutator is instantiated.

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -222,3 +275,16 @@ func updateVolumeMount(volumeMounts []corev1.VolumeMount, overrideVolumeMount co
...
+	if !ok {
+		return []string{}, fmt.Errorf("obj is supposed to be a metav1.Object, but is %T", obj)
+	}
+	lcluster := logicalcluster.From(metaObj)
+	return []string{WorkspaceAndNamespaceIndexKey(lcluster, metaObj.GetNamespace())}, nil
+}
+
+func WorkspaceAndNamespaceIndexKey(logicalcluster logicalcluster.Name, namespace string) string {
+	return logicalcluster.String() + "/" + namespace
+}
```
sttts, line 290: same, move up

pkg/syncer/spec/mutators/deployment.go
sttts: As discussed in Slack, if we can add the special case for default-token-XYZ, i.e. to map that to another name, I think we are fine. Rewriting all secrets definitely is not a good approach.

Having said that: have we checked that a deployment has zero chance to mount the pcluster default service account secret (or any other secret not coming from kcp) ?

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -39,34 +51,79 @@ func (dm *DeploymentMutator) GVR() schema.GroupVersionResource {
...
+
+			unstrSecret := elem.(*unstructured.Unstructured)
+			if err = runtime.DefaultUnstructuredConverter.FromUnstructured(unstrSecret.Object, &secret); err != nil {
+				return nil, fmt.Errorf("error converting secret %s/%s: %w", namespace, unstrSecret.GetName(), err)
+			}
+			secrets = append(secrets, &secret)
+		}
+		return secrets, nil
+	}
+}
```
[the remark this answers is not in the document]
jmprusi: done

pkg/syncer/spec/spec_controller.go
```diff
@@ -223,3 +235,44 @@ func (c *Controller) processNextWorkItem(ctx context.Context) bool {
...
+			}
+			secrets = append(secrets, &secret)
+		}
+
+		// In order to avoid triggering a deployment update on resyncs, we need to make sure that the list
+		// of secrets is sorted by creationTimsestamp. So if the user creates a new token for a given serviceaccount
+		// the first one will be picked always.
+		sort.Slice(secrets, func(i, j int) bool {
+			return secrets[i].CreationTimestamp.Before(&secrets[j].CreationTimestamp)
+		})
```
sttts, line 262: this logic does not belong here. A lister does not guarantee order.

pkg/syncer/spec/spec_controller.go
```diff
@@ -223,3 +235,44 @@ func (c *Controller) processNextWorkItem(ctx context.Context) bool {
...
+		if err != nil {
+			return nil, fmt.Errorf("error listing secrets for workspace %s: %w", clusterName, err)
+		}
+
+		secrets := make([]*corev1.Secret, 0, len(secretList))
+		for _, elem := range secretList {
+			var secret corev1.Secret
+
+			unstrSecret := elem.(*unstructured.Unstructured)
+			if err = runtime.DefaultUnstructuredConverter.FromUnstructured(unstrSecret.Object, &secret); err != nil {
```
sttts, line 251: feeling uncomfortable decoding a (potentially long) list of secrets, just to find one with an annotation.

sttts: I see nothing in the consuming code that cannot be done on unstructured.

pkg/syncer/spec/spec_process.go
```diff
@@ -343,18 +345,19 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
...
-		syncedObject.SetName("kcp-root-ca.crt")
+	if syncedObject.GroupVersionKind() == configMapGVK && syncedObject.GetName() == "kube-root-ca.crt" {
+		return "kcp-root-ca.crt"
 	}
-	// TODO(jmprusi): We are rewriting the name of the object into a non random one so we can reference it from the deployment transformer
-	//                but this means that means than more than one default-token-XXXX object will overwrite the same "kcp-default-token"
-	//				  object. This must be fixed.
-	if syncedObject.GroupVersionKind() == secretGVR && strings.Contains(syncedObject.GetName(), "default-token-") {
-		syncedObject.SetName("kcp-default-token")
+	if syncedObject.GroupVersionKind() == secretGVK && strings.Contains(syncedObject.GetName(), "-token-") {
```
sttts, line 356: not `HasPrefix("defaut-token-")`?

pkg/syncer/spec/spec_process.go
```diff
@@ -36,6 +36,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clusters"
 	"k8s.io/klog/v2"
+	"k8s.io/kubernetes/pkg/apis/core"
```
sttts, line 39: commented before: never use pkg/apis from kube directly, but always `k8s.io/api`.

pkg/syncer/spec/spec_process.go
jmprusi: done

pkg/syncer/spec/spec_process_test.go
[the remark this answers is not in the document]
sttts: I agree with Steve. But follow-up is fine.

pkg/syncer/spec/mutators/deployment.go
jmprusi: that code is gone now, done

PR #1128, opened 2022-05-23
hack: check for missing parameters correctly

stevekuznetsov:
`nounset` is an extremely valuable flag for ensuring that a shell script
does what is expected. When checking for missing parameters, simply use
the default substitution syntax so that the flag is not tripped.

---

```
$ head -n 21 hack/go-install.sh | tail -n 1
set -o nounset
$ hack/go-install.sh 
must provide module as first parameter
$ hack/go-install.sh  1
must provide binary name as second parameter
$ hack/go-install.sh  1 2
must provide version as third parameter
$ hack/go-install.sh  1 2 3
GOBIN is not set. Must set GOBIN to install the bin in a specified directory.
```

/assign @ncdc

### DIFF
--- a/hack/go-install.sh
+++ b/hack/go-install.sh
@@ -21,22 +21,22 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-if [ -z "${1}" ]; then
+if [[ -z "${1:-}" ]]; then
   echo "must provide module as first parameter"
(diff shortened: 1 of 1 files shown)


ncdc:
FYI @rgolangh

ncdc (approved):

PR #1140, opened 2022-05-24
reconciler: clusterworkspace: enforce labels

stevekuznetsov:
We cannot enforce that labels match the values in a status update, as a
client cannot update metadata and status in the same call. Therefore, we
move this logic from a mutating admission plugin to the controller.

/assign @sttts

### DIFF
--- a/pkg/admission/clusterworkspace/admission.go
+++ b/pkg/admission/clusterworkspace/admission.go
@@ -21,12 +21,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"strings"
 
 	"k8s.io/apimachinery/pkg/api/validation"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/util/sets"
 	labelvalidation "k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/apiserver/pkg/admission"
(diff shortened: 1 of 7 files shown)


sttts (approved):

pkg/admission/clusterworkspace/admission.go
```diff
@@ -89,6 +86,13 @@ func (o *clusterWorkspace) Validate(ctx context.Context, a admission.Attributes,
 		return fmt.Errorf("failed to convert unstructured to ClusterWorkspace: %w", err)
 	}
 
+	for i, initializer := range cw.Status.Initializers {
+		key := string(tenancyv1alpha1.ClusterWorkspaceInitializerLabelPrefix + initializer)
+		if len(key) > labelvalidation.LabelValueMaxLength {
+			return admission.NewForbidden(a, fmt.Errorf("status.initializers[%d] must be shorter than %d characters", i, labelvalidation.LabelValueMaxLength-len(tenancyv1alpha1.ClusterWorkspaceInitializerLabelPrefix)))
```
sttts, line 92: This must (also) go into the OpenAPI schema of the ClusterWorkspaceType.

sttts:
Controller lgtm. Please add the OpenAPI change in the type.

stevekuznetsov:
@sttts updated

stevekuznetsov:
@sttts  updated, thanks!

pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_controller.go
```diff
@@ -240,6 +241,35 @@ func (c *Controller) process(ctx context.Context, key string) error {
 	previous := obj
 	obj = obj.DeepCopy()
 
+	reconcileMetadata(obj)
+
+	// If the object being reconciled changed as a result, update it.
+	if !equality.Semantic.DeepEqual(previous.ObjectMeta, obj.ObjectMeta) {
```
shawn-hurley, line 247: This might be a silly question but if we are working on labels what is the value of checking the entire ObjectMetas are equal?

shawn-hurley (commented):
LGTM, just one question

pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_controller.go
stevekuznetsov: No real reason, certainly could change `reconcileMetadata` and this block to operate strictly on labels

sttts (approved):

PR #1145, opened 2022-05-25
Better error when failing to create a sub workspace

### LINKED ISSUE #831: kubectl plugin should return friendly errors when trying to run create/list/etc in a workspace that doesn't support nested workspaces
We support a hierarchy of workspaces - root -> org1 -> workspaceA, and sometimes users might feel like they can create an arbitrarily deep nesting of workspaces, e.g. under workspaceA -> foo -> bar -> baz. This is not possible by default unless you create the right type of workspace at each level (e.g. Organization instead of Universal). It can be confusing when you try to create a nested workspace or get a list of workspaces from a workspace that doesn't support nesting - you get errors such as

```
# try to list workspaces from a Universal workspace
kubectl kcp ws list
error: workspaces.tenancy.kcp.dev "test" is forbidden: "root:some-org" workspace access not permitted

# try to create a nested workspace inside a Universal workspace
k kcp ws create inner1
error: the server could not find the requested resource (post clusterworkspaces.tenancy.kcp.dev)
```

The kubectl plugin code should return errors that are more user-friendly. For example, if you try to list workspaces from a workspace that doesn't support nesting, the plugin could tell that to the user - `workspace "foo" does not support nested workspaces` or something to that effect.

This could be achieved by checking discovery as one possible solution.

N/A

**Additional context**
N/A

rgolangh:
Return a nicer error instead of the `resource not found`

Signed-off-by: Roy Golan <rgolan@redhat.com>

## Related issue(s)

Fixes #831

### DIFF
--- a/pkg/cliplugins/workspace/plugin/kubeconfig.go
+++ b/pkg/cliplugins/workspace/plugin/kubeconfig.go
@@ -309,6 +309,13 @@ func (kc *KubeConfig) CreateWorkspace(ctx context.Context, workspaceName string,
 			Type: workspaceType,
 		},
 	}, metav1.CreateOptions{})
+	if apierrors.IsNotFound(err) {
+		// STOP THE BLEEDING: currently, kcp forwards workspace resource request to the workspace virtual apiserver
+		// indpenedently whether the CRD is installed in the workspace. Universal workspaces though don't have that
+		// resource, but the virtual apiserver return 404 in that case, confusingly for clients.
+		// This hack avoids a message confusing for the user.
+		return fmt.Errorf("creating a workspace under a Universal type workspace is not supported")
+	}
 	if apierrors.IsAlreadyExists(err) && ignoreExisting {
 		preExisting = true
 		ws, err = kc.personalClient.Cluster(currentClusterName).TenancyV1beta1().Workspaces().Get(ctx, workspaceName, metav1.GetOptions{})


pkg/cliplugins/helpers/helpers.go
```diff
@@ -62,3 +65,20 @@ func IsValid(cluster logicalcluster.Name) bool {
 
 	return cluster.HasPrefix(v1alpha1.RootCluster) || cluster.HasPrefix(logicalcluster.New("system"))
 }
+
+func ClusterHasResource(gvk schema.GroupVersionKind ,discoveryClient discovery.DiscoveryInterface) (bool, error) {
+	klog.Info(fmt.Sprintf("looking for grouversion %v", gvk.GroupVersion()))
```
stevekuznetsov, line 70: Use `Infof` :)

rgolangh: I want to remove all those log prints anyway. But indeed will for next time.

rgolangh: for some reason the Cluster(currentClusterName) seem to always have the `Workspace` crd. 
I expected that this discovery client would belong to the current workspace. Am I wrong here?

stevekuznetsov: We install the `ClusterWorkspace` CRD in every workspace, IIRC

rgolangh: `kubectl api-resources | grep Workspace`  will show nothing in a Universal workspace

rgolangh: you are right there are clusterworkspaces

sttts (approved):

ncdc:
Hit #1158, unrelated, merging

Issue #1008, opened 2022-05-06
Replace all .ClusterName / GetClusterName() access with logicalcluster package functionality


ncdc:
ObjectMeta.ClusterName is gone in Kubernetes 1.24. We need to replace our usage of it in kcp-dev/logicalcluster, kcp-dev/kubernetes, and kcp-dev/kcp. We should instead use logicalcluster.From(obj).String() if we need to get it from an object. And we'll need to figure out the best patterns to use for setting it.

ncdc:
This is done
/close

PR #1146, opened 2022-05-25
remove getting cluster name from object meta 

shawn-hurley:
## Summary
Removing getting logicalcluster from objects metadata, and instead use logicalcluster.From.

Handeling Setting the value, will be completed in a followup PR.
## Related issue(s)

Partial fix for #1008


stevekuznetsov (approved):

stevekuznetsov:
Changes look reasonable, but you need `make imports`

Issue #324, opened 2021-12-13
Move CI from GitHub Actions to Prow


ncdc:
We have discussed in Slack a few times about moving to Prow. Many of us are more familiar with Prow and it is consistent with how Kubernetes/kubernetes-sigs/OpenShift projects do CI, which will likely be important if we ever donate this project to the CNCF.

Please vote 👍 or 👎 if you think we should move from GitHub Actions to Prow for our CI jobs.

If the decision is 👍, then let's add this to our scheduled list of tasks so we can properly plan for it.

stevekuznetsov:
Some of the other things I've noticed with GitHub actions which Prow fixes for us (but is not the only tool to do so) are:

 - base PR on top of the branch it's merging into for testing
 - ensure PR is tested against latest branch before merging
 - support for jUnit or similar per-test views
 - support for an in-browser artifact viewer
 - able to trigger new jobs on existing PRs when new jobs are added or existing jobs change

pweil-:
@ncdc - looks like this is ready to be added to the schedule somewhere.  If there is no urgency maybe we can stick `help-wanted` on it to see if anyone is interested or get it into P4's tech objectives.

ncdc:
@pweil- this would potentially be a long-term thing, unless GH actions stop serving our needs sooner. I'm a bit hesitant on adding help-wanted, as it requires knowledge of and access to infra where Prow is currently running (either for OpenShift or Kube).

stevekuznetsov:
Items left TODO:

 - [ ] finalize image builds, they are published but we need them to be 1:1 with existing ones (cc @kylape )
 - [ ] add jobs etc for release branches
 - [ ] figure out where jUnit files go for multiple-runs tests ...
 - [ ] get the e2e-shared-server setup/run into a reproducible `make` target and get it running in prow

sttts:
- [ ] https://github.com/kcp-dev/kcp/issues/1209

kylape:
When this issue is ready to be closed, does that mean we can remove the Github Action tasks that build images?

ncdc:
If we're moving to Prow for image builds, yes

PR #1148, opened 2022-05-25
makefile: update things to be consistent & simple

stevekuznetsov:
We should not require more than a simple `make` target to run any test.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -67,15 +67,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-        with:
-          path: kcp
       - uses: actions/setup-go@v3
         with:
           go-version: v1.17
       - name: Check dependencies
         run: |
-          cd kcp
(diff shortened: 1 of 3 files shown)


sttts (approved):

.github/workflows/ci.yaml
```diff
@@ -55,27 +55,19 @@ jobs:
           go-version: v1.17
       - name: Check imports
         run: |
-          cd kcp
-          make imports
-          if  ! git diff --exit-code; then
-            echo "imports are out of date, run make imports"
-            exit 1
-          fi
+          make verify-imports
```
ncdc, line 58: Do you need to remove `path: kcp` from L52?

hack/verify-imports.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 The KCP Authors.
```
ncdc, line 3:
```suggestion
# Copyright 2022 The KCP Authors.
```

stevekuznetsov:
Part of #324

PR #1161, opened 2022-05-27
make: add a target for all of our tools

stevekuznetsov:
This will allow us to put it into a build cache, but still own the
definition of them.


ncdc (approved):

stevekuznetsov:
Part of #324

PR #1165, opened 2022-05-27
e2e/virtual: avoid data race

sttts:
Golang slice semantics 🤷‍♂️


sttts:
Actually, the func create a new slice every time. We should be safe.

Issue #748, opened 2022-03-18
Make CRDs and APIBindings conflict on each other for overlapping GVRs


sttts:
Follow-up #702:

CRDs and APIBindings should not have overlapping GVRs, both because only one can be served obviously, but also because both share the same etcd key space and hence there is some risk (💣) that one steals data of the other.

The idea here is to create an admission plugin on both resources that forbids one or the other when GVRs overlap. This also means that that admission plugin might stop binding of the APIBinding to a new export revision that got new GVRs.

Related to #417

ncdc:
> because both share the same etcd key space and hence there is some risk (💣) that one steals data of the other

Note, with #877 now merged, CRs from APIBindings and CRs from normal CRDs are stored distinctly in etcd. So this concern is no longer an issue.

sttts:
Disagree that is no issue anymore. It is less security-critical, but the user-experience is still broken.

ncdc:
Yes, you're right

PR #1170, opened 2022-05-30
adds crdNoOverlappingGVRAdmission admission 

p0lyn0mial:
## Summary
introduces an admission plugin for ensuring CRD's Group and Resource fields for logical clusters are unique and don't overlap with bound CRDs.

TODO:
- change the binding controller to check for conflicting GRVs (separate PR)
## Related issue(s)

xref: https://github.com/kcp-dev/kcp/issues/748

### DIFF
--- a/pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
+++ b/pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
@@ -1,0 +1,140 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package crdnooverlappinggvr
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"github.com/kcp-dev/logicalcluster"
+
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apiserver/pkg/admission"
+	"k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/client-go/tools/cache"
+
+	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
+	kcpinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
+	"github.com/kcp-dev/kcp/pkg/reconciler/apis/apibinding"
+)
+
+const (
+	PluginName  = "apis.kcp.dev/CRDNoOverlappingGVR"
+	byWorkspace = PluginName + "-byWorkspace"
+)
+
+func Register(plugins *admission.Plugins) {
+	plugins.Register(PluginName,
+		func(_ io.Reader) (admission.Interface, error) {
+			return &crdNoOverlappingGVRAdmission{
+				Handler: admission.NewHandler(admission.Create),
+			}, nil
+		})
+}
+
+type crdNoOverlappingGVRAdmission struct {
+	*admission.Handler
+
+	apiBindingIndexer          cache.Indexer
+	apiBindingIndexerInitError error
+}
+
+// Ensure that the required admission interfaces are implemented.
+var _ = admission.ValidationInterface(&crdNoOverlappingGVRAdmission{})
+var _ = admission.InitializationValidator(&crdNoOverlappingGVRAdmission{})
+
+func (p *crdNoOverlappingGVRAdmission) SetKcpInformers(informers kcpinformers.SharedInformerFactory) {
+	if err := informers.Apis().V1alpha1().APIBindings().Informer().AddIndexers(cache.Indexers{byWorkspace: indexByWorkspace}); err != nil {
+		p.apiBindingIndexerInitError = err
+		return
+	}
+
+	// just in case the plugin gets init multiple times in case of an error
+	p.apiBindingIndexerInitError = nil
+	p.SetReadyFunc(informers.Apis().V1alpha1().APIBindings().Informer().HasSynced)
+	p.apiBindingIndexer = informers.Apis().V1alpha1().APIBindings().Informer().GetIndexer()
+}
+
+func (p *crdNoOverlappingGVRAdmission) ValidateInitialization() error {
+	if p.apiBindingIndexerInitError != nil {
+		return fmt.Errorf(PluginName+" plugin failed to initialize %q APIBindings indexer, err = %v", byWorkspace, p.apiBindingIndexerInitError)
+	}
+	if p.apiBindingIndexer == nil {
+		return fmt.Errorf(PluginName + " plugin needs an APIBindings indexer")
+	}
+	return nil
+}
+
+// Validate checks if the given CRD's Group and Resource don't overlap with bound CRDs
+func (p *crdNoOverlappingGVRAdmission) Validate(ctx context.Context, a admission.Attributes, _ admission.ObjectInterfaces) error {
+	if a.GetResource().GroupResource() != apiextensions.Resource("customresourcedefinitions") {
+		return nil
+	}
+	if a.GetKind().GroupKind() != apiextensions.Kind("CustomResourceDefinition") {
+		return nil
+	}
+	clusterName, err := request.ClusterNameFrom(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve cluster from context: %w", err)
+	}
+	// ignore CRDs targeting system and non-root workspaces
+	if clusterName == apibinding.ShadowWorkspaceName || clusterName == logicalcluster.New("system:admin") {
+		return nil
+	}
+
+	crd, ok := a.GetObject().(*apiextensions.CustomResourceDefinition)
+	if !ok {
+		return fmt.Errorf("unexpected type %T", a.GetObject())
+	}
+	apiBindingsForCurrentClusterName, err := p.listAPIBindingsFor(clusterName)
+	if err != nil {
+		return err
+	}
+	for _, apiBindingForCurrentClusterName := range apiBindingsForCurrentClusterName {
+		for _, boundResource := range apiBindingForCurrentClusterName.Status.BoundResources {
+			if boundResource.Group == crd.Spec.Group && boundResource.Resource == crd.Spec.Names.Plural {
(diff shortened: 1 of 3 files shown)


pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
```diff
@@ -0,0 +1,131 @@
...
+		if err != nil {
+			return err
+		}
+
+		for _, apiExportSchemaName := range apiExport.Spec.LatestResourceSchemas {
+			apiResourceSchema, err := p.getAPIResourceSchema(apiExportClusterName, apiExportSchemaName)
+			if err != nil {
+				return err
+			}
+			if apiResourceSchema.Spec.Group == crd.Spec.Group && apiResourceSchema.Spec.Names.Plural == crd.Spec.Names.Plural {
```
p0lyn0mial, line 93: I think comparing `Group` and `Resource` is enough, we don't have to take `Versions` into consideration.

pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return err
+	}
+	for _, apiBindingForCurrentClusterName := range apiBindingsForCurrentClusterName {
+		apiExportClusterName, err := getAPIExportClusterName(apiBindingForCurrentClusterName)
+		if err != nil {
+			// since all exports needs to be checked we need to stop in case of an error
+			return err
+		}
+		apiExport, err := p.getAPIExport(apiExportClusterName, apiBindingForCurrentClusterName.Spec.Reference.Workspace.ExportName)
```
p0lyn0mial, line 83: I think that `CRDs` from an APIExport are the ones that are eventually created in the "current" cluster.

pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
```diff
@@ -0,0 +1,131 @@
...
+			informers.Apis().V1alpha1().APIResourceSchemas().Informer().HasSynced()
+	})
+	p.apiBindingLister = informers.Apis().V1alpha1().APIBindings().Lister()
+	p.apiExportLister = informers.Apis().V1alpha1().APIExports().Lister()
+	p.apiResourceSchemaLister = informers.Apis().V1alpha1().APIResourceSchemas().Lister()
+}
+
+// Validate checks if the given CRD's Group and Resource don't overlap with other CRDs exposed via APIBindings
+func (p *crdNoOverlappingGVRAdmission) Validate(ctx context.Context, a admission.Attributes, _ admission.ObjectInterfaces) error {
+	if a.GetResource().GroupResource() != apiextensions.Resource("customresourcedefinitions") {
```
p0lyn0mial, line 57: I think I need to find a way to skip `CRDs` created by the APIBinding controller

p0lyn0mial: or actually, we should make a distinction between CRDs created for the binding and the ones created by an admin

pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
sttts: yes, all versions should be owned either by the CRD or the binding.

pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
sttts: try to not depend on the APIExport. The APIBinding itself (the status) must be enough for this decision.

Background: APIExport might live on a different shard and we should not break if the shard is unavailable (network partition). If something is missing for the logic in the APIBinding status, we should add it.

pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
sttts: the ones from APIBinding live in the `system:bounds-crds` logical cluster.

pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
p0lyn0mial: oh, ic, does it mean we should reject the request if APIBinding is still progressing?

sttts: The status of the APIBinding is what counts. This admission plugin also has to stop binding to succeed (with some good condition message).

pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
p0lyn0mial: there seem to be no way of identifying the target cluster from a CRD in the logical cluster

sttts: I think the APIBinding CRDs in `system:bound-crds` you can ignore here as long as you have the GVRs for them from APIBinding.status.

The `system:admin` CRDs matter too here (today that's all from https://github.com/kcp-dev/kcp/tree/main/config/crds, but most of them will turn into APIBindings soon).

pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	if a.GetKind().GroupKind() != apiextensions.Kind("CustomResourceDefinition") {
+		return nil
+	}
+	clusterName, err := request.ClusterNameFrom(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve cluster from context: %w", err)
+	}
+	// ignore CRDs targeting system workspaces
+	if clusterName == apibinding.ShadowWorkspaceName || clusterName == logicalcluster.New("system:admin") {
```
sttts, line 61: maybe even `!clusterName.HasPrefix(tenancyv1alpha1.RootCluster)`

pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+	}
+	if len(notReadyAPIBindings) > 0 {
+		return admission.NewForbidden(a, fmt.Errorf("cannot create %q crd because the following apiBindings werern't ready and werern't checked for conflicting GVRs = %v", crd.Name, notReadyAPIBindings))
+	}
+	return nil
+}
+
+func (p *crdNoOverlappingGVRAdmission) listAPIBindingsFor(clusterName logicalcluster.Name) ([]*apisv1alpha1.APIBinding, error) {
+	list, err := p.apiBindingLister.List(labels.Everything())
```
sttts, line 93: use an index as in https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/scheduling/placement/placement_indexes.go#L30.

pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+		for _, boundResource := range apiBindingForCurrentClusterName.Status.BoundResources {
+			if boundResource.Group == crd.Spec.Group && boundResource.Resource == crd.Spec.Names.Plural {
+				return admission.NewForbidden(a, fmt.Errorf("cannot create %q crd with %q group and %q resource because it overlaps with a bound crd for %q apiBinding in %q logical cluster",
+					crd.Name, crd.Spec.Group, crd.Spec.Names.Plural, apiBindingForCurrentClusterName.Name, clusterName))
+			}
+		}
+	}
+	if len(notReadyAPIBindings) > 0 {
+		return admission.NewForbidden(a, fmt.Errorf("cannot create %q crd because the following apiBindings werern't ready and werern't checked for conflicting GVRs = %v", crd.Name, notReadyAPIBindings))
```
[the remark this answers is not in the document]
sttts: why fail at all in this case?

pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	if a.GetKind().GroupKind() != apiextensions.Kind("CustomResourceDefinition") {
+		return nil
+	}
+	clusterName, err := request.ClusterNameFrom(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve cluster from context: %w", err)
+	}
+	// ignore CRDs targeting system and non-root workspaces
+	if clusterName == apibinding.ShadowWorkspaceName || clusterName == logicalcluster.New("system:admin") || !clusterName.HasPrefix(tenancyv1alpha1.RootCluster) {
```
sttts, line 99: nit: the last test implies the other two. Am fine with either, the first two or the last :)

pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
```diff
@@ -0,0 +1,150 @@
...
+		for _, boundResource := range apiBindingForCurrentClusterName.Status.BoundResources {
+			if boundResource.Group == crd.Spec.Group && boundResource.Resource == crd.Spec.Names.Plural {
+				return admission.NewForbidden(a, fmt.Errorf("cannot create %q CustomResourceDefinition with %q group and %q resource because it overlaps with a bound CustomResourceDefinition for %q APIBinding in %q logical cluster",
+					crd.Name, crd.Spec.Group, crd.Spec.Names.Plural, apiBindingForCurrentClusterName.Name, clusterName))
+			}
+		}
+	}
+	if len(notReadyAPIBindings) > 0 {
+		return admission.NewForbidden(a, fmt.Errorf("cannot create %q CustomResourceDefinition because the following APIBinding werern't ready and werern't checked for conflicting GVRs = %v", crd.Name, notReadyAPIBindings))
+	}
```
sttts, line 126: still think this whole if clause is not right. Being in !BindingUpToDate is normal, not a reason to stop CRD creation.

p0lyn0mial: seems like we need to trade off availability over consistency. My intention was to make sure we have seen all bounded CRDs before admitting a CRD. I guess you prefer the former or is there any other condition we should check? (`!InitialBindingCompleted`?)

sttts: resources not bound via APIBinding status do not exist :)

sttts: Same with `!InitialBindingCompleted`. At that point there are no bound resources yet. So no conflict.

There is the second part with the APIBinding controller that it needs similar logic for the inverse check.

p0lyn0mial: yeah, at this point it is a best-effort admission

p0lyn0mial:
@sttts updated, PTAL

p0lyn0mial:
run it locally and it gave me `error: plugins [apis.kcp.dev/CRDNoOverlappingGVR] registered are not in RecommendedPluginOrder`

PR #1171, opened 2022-05-30
makefile: update path to include output binaries

stevekuznetsov:
There's no real reason to expect every single user to manage their path
individually when calling `make`, and it furthermore never makes sense
to use an out-of-date binary that we create, so our output should be a
prefix to the path.


stevekuznetsov:
Part of #324

Issue #1151, opened 2022-05-26
Add authorization checks (SAR) for APIExport virtual workspace


shawn-hurley:
We want to make sure that the user's access the Virtual Workspace for an APIExport, that they have permissions.

I believe that that this means the user has access to the sub resource '/content' on the given workspace type being used.

It sounds like this to me, is the check that we will run for every interaction with a given virtual workspace.

ncdc:
This is more going to be a ServiceAccount accessing the endpoint, but that's nit-picky 😄 

The client will request `/services/apiexport/$clusterContainingAPIExport/$apiExportName/clusters/...`. We need to perform a SAR that the client is authorized for that request. The SAR for this is probably something like

```
authorizer.AttributesRecord{
		APIGroup:        "apis.kcp.dev",
		APIVersion:      "v1alpha1",
		Resource:        "apiexports",
		Name:            "$apiExportName",
		ResourceRequest: true,
		User:            user,
		Verb:            "<not sure but something other than get/list/etc>",
	}
```

sttts:
Alternative:
```
authorizer.AttributesRecord{
		APIGroup:        "apis.kcp.dev",
		APIVersion:      "v1alpha1",
		Resource:        "apiexports",
		SubResource: "objects"
		Name:            "$apiExportName",
		ResourceRequest: true,
		User:            user,
		Verb:            "access",
	}
```

shawn-hurley:
I think it might be good to make sure that I know what the goal is. 

IIUC, this Virutal Workspace is for the ServiceProvider to watch/list/update the resources that they are exporting.

I think, that a quick win of the subresource of Content, with a followup on the more "andriod permission" model when binding, should allow more granular control. I could see something where when someone "accepts" these permissions they are added in the workspace and we can forward the request to the underlying workspace. 

What do folks think?

ncdc:
@shawn-hurley can you fill out an AttributesRecord with your proposal?

shawn-hurley:
```authorizer.AttributesRecord{
		APIGroup:        "apis.kcp.dev",
		APIVersion:      "v1alpha1",
		Resource:        "certificates",
		SubResource: "",
		Name:            "<actual-object-name>",
		ResourceRequest: true,
		User:            user,
		Verb:            "<actual-verb>",
	}

ncdc:
[...]

PR #1186, opened 2022-06-01
Add authorizers in virtual workspaces

davidfestal:
## Summary

Add authorizers in virtual workspaces

## Related issue(s)

This is a pre-requisite for issues  #1172 and #1151

### DIFF
--- a/pkg/virtual/framework/dynamic/types.go
+++ b/pkg/virtual/framework/dynamic/types.go
@@ -19,6 +19,7 @@ package dynamic
 import (
 	"context"
 
+	"k8s.io/apiserver/pkg/authorization/authorizer"
 	genericapiserver "k8s.io/apiserver/pkg/server"
 
 	"github.com/kcp-dev/kcp/pkg/virtual/framework"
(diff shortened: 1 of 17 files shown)


pkg/virtual/framework/dynamic/types.go
```diff
@@ -51,3 +53,10 @@ func (vw *DynamicVirtualWorkspace) IsReady() error {
 func (vw *DynamicVirtualWorkspace) ResolveRootPath(urlPath string, context context.Context) (accepted bool, prefixToStrip string, completedContext context.Context) {
 	return vw.RootPathResolver(urlPath, context)
 }
+
+func (vw *DynamicVirtualWorkspace) Authorize(ctx context.Context, a authorizer.Attributes) (authorized authorizer.Decision, reason string, err error) {
+	if vw.Authorizer != nil {
+		return vw.Authorizer(ctx, a)
+	}
+	return authorizer.DecisionAllow, "", nil
```
shawn-hurley, line 61: Should this be NoOpinion?

shawn-hurley (approved):
I think the default for the authorizers should be NoOpinion if not set. 

This might break some things that rely on this, I wonder if that is worth the price to pay?

pkg/virtual/framework/dynamic/types.go
davidfestal: So to avoid breaking anything we'd add an explicit AllowEverything authorizers in all the virtual workspace instances ?

sttts: Also a vw without authorizer should forbid every request.

davidfestal: Changed in https://github.com/kcp-dev/kcp/pull/1186/commits/cf98b013c8b5f369eb3cfa950de041360f7f0e88

sttts (approved):

PR #1206, opened 2022-06-03
clusterworkspace: qualify initializer names

### LINKED ISSUE #1169: ClusterWorkspaceTypes are insecure by design
**Describe the bug**

Assume ClusterWorkspaceType creation is no privileged operation. Then ClusterWorkspaceTypes with arbitrary initializers can be used to get access to arbitrary ClusterWorkspaces:

1. create a ClusterWorkspaceType X in your org with a system ClusterWorkspaceType (i.e. X.Name == Y.Name), and an initializer I that is also set in Y.
2. wildcard requests are filtered by initializer labels and not by logical cluster. I.e. a controller with permissions against X will be able to see **all** workspaces (via the vw) created with type X.Name (== Y.Name) in the system, i.e. including those outside X's workspace.

**Solution**

Enforce initializer names in a ClusterWorkspaceType that include the logical cluster.

stevekuznetsov:
We need globally unique identifiers for the initializing entities, or
our model is insecure. By adding the workspace path from which an
initializer runs to the name of the initializer, we get such an
identifier.

In addition, we change the labelling mechanisms to use a hash of these
values in order to ensure that we continue to create unique and correct
label keys and values.

Fixes #1169 
/assign @sttts 
/cc @ncdc @kylape 

This is a backwards-incompatible change to the ClusterWorkspace and ClusterWorkspaceType CRDs.


config/crds/tenancy.kcp.dev_clusterworkspacetypes.yaml
```diff
@@ -53,9 +53,24 @@ spec:
...
+                      description: name is a unique string corresponding to a cluster
+                        workspace initialization controller for the given type of
+                        workspaces.
+                      maxLength: 32
+                      pattern: ^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$
+                      type: string
+                    path:
+                      description: path is an absolute reference to the workspace
+                        that owns this initializer, e.g. root:org:ws. The workspace
+                        must be some ancestor or a child of some ancestor.
```
sttts, line 67: Copy and pasta text? For the type I would expect that it is just the clusterName of this object. I.e. we only have to store it on the ClusterWorkspace because there the initializers can come from different workspaces.

pkg/admission/clusterworkspacetypeexists/admission.go
```diff
@@ -143,14 +143,8 @@ func (o *clusterWorkspaceTypeExists) Admit(ctx context.Context, a admission.Attr
...
 	// add initializers from type to workspace
-	existing := sets.NewString()
-	for _, i := range cw.Status.Initializers {
-		existing.Insert(string(i))
-	}
 	for _, i := range cwt.Spec.Initializers {
-		if !existing.Has(string(i)) {
-			cw.Status.Initializers = append(cw.Status.Initializers, i)
-		}
+		cw.Status.Initializers = initialization.EnsureInitializerPresent(i.Name, i.Path, cw.Status.Initializers)
```
sttts, line 147: s/i.Path/i.ClusterName/

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -131,10 +131,20 @@ type ClusterWorkspaceTypeList struct {
...
+type ClusterWorkspaceInitializer struct {
+	// name is a unique string corresponding to a cluster workspace
+	// initialization controller for the given type of workspaces.
+	//
+	// +kubebuilder:validation:MaxLength=32
+	// +kubebuilder:validation:Pattern=`^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*/)?(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$`
+	Name string `json:"name"`
+
+	// path is an absolute reference to the workspace that owns this initializer, e.g. root:org:ws.
+	// The workspace must be some ancestor or a child of some ancestor.
```
sttts, line 143: copy and pasta.

As written above, this path in the type is unnecessary (and undesired). The type initializer path is just the lcluster of the type.

In the ClusterWorkspace we have to store the path though (and replicate into the labels somehow).

pkg/apis/tenancy/initialization/utils.go
```diff
@@ -0,0 +1,71 @@
...
+		initializers = append(initializers[:removeAt], initializers[removeAt+1:]...)
+	}
+	return initializers
+}
+
+// InitializerToLabel transforms an initializer into a key-value pair to add to a label set. We use a hash
+// to create a unique identifier from this information, prefixing the hash in order to create a value which
+// is unlikely to collide, and adding the full hash as a value in order to make it difficult to forge the pair.
+func InitializerToLabel(initializer tenancyv1alpha1.ClusterWorkspaceInitializer) (string, string) {
+	hash := fmt.Sprintf("%x", sha256.Sum224([]byte(initializer.Path+initializer.Name)))
```
sttts, line 68: is sha224 considered strong these days (I don't know, just curious) ?

stevekuznetsov: I believe everything up to SHA256 has been cracked, but are we trying to make this secure, or simply inconvenient?

pkg/apis/tenancy/v1alpha1/types.go
stevekuznetsov: We need to sync on this - I don't think I understand what you mean by "lcluster" if it's not the path ... ?

pkg/apis/tenancy/initialization/utils.go
sttts: https://en.wikipedia.org/wiki/SHA-2#Cryptanalysis_and_validation – there are some attacks, but what I read it is still considered mostly safe.

pkg/apis/tenancy/v1alpha1/types.go
sttts: lcluster = logical cluster = since recently also called workspace path.

sttts: The "ancestor or a child ..." is the piece that looks like copy&paste.

config/crds/tenancy.kcp.dev_clusterworkspacetypes.yaml
stevekuznetsov: If we have one central API Binding application controller that's simply configured differently per ClusterWorkspaceType, this invariant is already broken/

stevekuznetsov:
I have a commit locally that makes the changes @sttts suggested (the path for an initializer in a ClusterWorkspace is implicit in the ClusterName of the ClusterWorkspaceType that was used to create the ClusterWorkspace) but hilariously today we *do* have global initializers for bootstrapping so .... that is nonfunctional :)

pkg/apis/tenancy/initialization/utils.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"crypto/sha256"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/util/validation"
+
+	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
+)
+
+func InitializerPresent(name, path string, initializers []tenancyv1alpha1.ClusterWorkspaceInitializer) bool {
```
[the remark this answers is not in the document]
sttts: In other words, technically path is a logical cluster, path is just the user-facing abstraction for tenancy.

pkg/apis/tenancy/initialization/utils.go
```diff
@@ -0,0 +1,71 @@
...
+	if removeAt != -1 {
+		initializers = append(initializers[:removeAt], initializers[removeAt+1:]...)
+	}
+	return initializers
+}
+
+// InitializerToLabel transforms an initializer into a key-value pair to add to a label set. We use a hash
+// to create a unique identifier from this information, prefixing the hash in order to create a value which
+// is unlikely to collide, and adding the full hash as a value in order to make it difficult to forge the pair.
+func InitializerToLabel(initializer tenancyv1alpha1.ClusterWorkspaceInitializer) (string, string) {
```
sttts, line 67: is this an API? Then it is misplaced in this package.

pkg/reconciler/tenancy/bootstrap/bootstrap_controller.go
```diff
@@ -69,8 +69,9 @@ func NewController(
 		syncChecks: []cache.InformerSynced{
 			workspaceInformer.Informer().HasSynced,
 		},
-		workspaceType: workspaceType,
-		bootstrap:     bootstrap,
+		workspaceType:   workspaceType,
+		initializerPath: initializerPath,
```
sttts, line 73: we talked about this. IMO wrong direction. But we agreed to explore our direction in a doc.

sttts: Or in other words: this controller should have one and only one initializer path, namely `root`, because it is only used in ClusterWorkspaceTypes in `root`. This initializer is not an API to be used by anybody else than the core system.

pkg/virtual/initializingworkspaces/builder/build.go
```diff
@@ -79,23 +79,28 @@ func BuildVirtualWorkspace(
 			withoutRootPathPrefix := strings.TrimPrefix(urlPath, rootPathPrefix)
 
 			// Incoming requests to this virtual workspace will look like:
-			//  /services/initializingworkspaces/<initializer>/clusters/*/apis/workload.kcp.dev/v1alpha1/workloadclusters
+			//  /services/initializingworkspaces/<workspace>/<initializer>/clusters/*/apis/workload.kcp.dev/v1alpha1/workloadclusters
```
sttts, line 82: `s/<workspace>/<typeWorkspace>/` (ErrTooManyWorkspacesAround)

pkg/virtual/initializingworkspaces/builder/build.go
```diff
@@ -168,7 +173,11 @@ type apiSetRetriever struct {
 }
 
 func (a *apiSetRetriever) GetAPIDefinitionSet(ctx context.Context, key dynamiccontext.APIDomainKey) (apis apidefinition.APIDefinitionSet, apisExist bool, err error) {
-	initializerName := string(key)
+	parts := strings.Split(string(key), "|")
+	if len(parts) != 2 {
+		return nil, false, fmt.Errorf("invalid API domain key, expected workspace|initializerName: %q", key)
+	}
+	initializerWorkspace, initializerName := parts[0], parts[1]
```
sttts, line 180: typeWorkspace is IMO a better name. The initializer is not a thing, and not even unique by itself.

pkg/virtual/initializingworkspaces/builder/build.go
stevekuznetsov: Different mental model.

stevekuznetsov:
@sttts updated the method signatures not to have path as a string, but did not update the API object to use `logicalcluster.Name` since this does not seem to be intended to be embedded.

stevekuznetsov:
```
    virtualworkspace_test.go:174: Create workspaces that using the new types, which will get stuck in initializing
    virtualworkspace_test.go:179: 
        	Error Trace:	virtualworkspace_test.go:179
        	            	clusterworkspaces.tenancy.kcp.dev "e2e-workspace-" is forbidden: spec.type "Alphadkorbpxplz" does not exist
        	Test:       	TestInitializingWorkspacesVirtualWorkspaceAccess
```

That's a new, fun flake. Hm.

sttts:
> That's a new, fun flake. Hm.

Admission and authz are by design eventually consistent. If that's the root cause, all is fine.

sttts:
the flake is fixed in main.

stevekuznetsov:
Superseded by https://github.com/kcp-dev/kcp/pull/1264
/close

PR #1207, opened 2022-06-03
ci: turn off GitHub CI for jobs we run in Prow

stevekuznetsov:

Part of #324 
/assign @ncdc @sttts


sttts (dismissed):

sttts:
What is special about the shared server one?

stevekuznetsov:
@sttts shared server requires that I get the actual running of it figured out reproducibly from `make` and the images wired through etc. Runtime should be 10000% tunable based on how many resources we give the jobs.

sttts:
Ack. I think we can move forward here, but would like to see the multiple-run job to be below 20min. Now it's 35min or so, much more than the others.

ncdc:
/lgtm
/approve

c'mon tide 😄

ncdc:
Manually merging

PR #1215, opened 2022-06-06
contrib: rename demo to demos-unmaintained

sttts:
We don't have the capacity to keep demos running that are built against alpha APIs. Hence, we better make that clear to the ppl trying kcp.


ncdc:
Do you need to remove the new files that were added (kubeconfigs from local demo runs)?

sttts:
Fixed.

/hold cancel

PR #1217, opened 2022-06-06
APIBinding reconciler checks for overlapping CRDs in a logical cluster

### LINKED ISSUE #748: Make CRDs and APIBindings conflict on each other for overlapping GVRs

p0lyn0mial:
## Summary
changes the APIBinding controller to check for CRDs with overlapping GVRs in a logical cluster. Initially, we wanted to implement it as an admission plugging. Since it might involve network calls we decided to move it to the controller.
## Related issue(s)

Fixes #748

### DIFF
--- a/pkg/reconciler/apis/apibinding/apibinding_controller.go
+++ b/pkg/reconciler/apis/apibinding/apibinding_controller.go
@@ -155,6 +155,12 @@ func NewController(
 		},
 	})
 
+	if err := crdInformer.Informer().AddIndexers(cache.Indexers{
+		indexByWorkspace: indexByWorkspaceFunc,
+	}); err != nil {
+		return nil, err
+	}
+
 	apiResourceSchemaInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc:    func(obj interface{}) { c.enqueueAPIResourceSchema(obj) },
 		UpdateFunc: func(_, obj interface{}) { c.enqueueAPIResourceSchema(obj) },
--- a/pkg/reconciler/apis/apibinding/apibinding_indexes.go
+++ b/pkg/reconciler/apis/apibinding/apibinding_indexes.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/kcp-dev/logicalcluster"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/clusters"
 
 	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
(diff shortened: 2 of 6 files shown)


pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -194,15 +194,16 @@ func (c *controller) reconcileBinding(ctx context.Context, apiBinding *apisv1alp
...
+		checker := &conflictChecker{
 			listAPIBindings:      c.listAPIBindings,
 			getAPIExport:         c.getAPIExport,
 			getAPIResourceSchema: c.getAPIResourceSchema,
 			getCRD:               c.getCRD,
+			crdIndexer:           c.crdIndexer,
 		}
 
-		if err := nameConflictChecker.checkForConflicts(crd, apiBinding); err != nil {
+		if err := checker.checkForConflicts(crd, apiBinding); err != nil {
```
p0lyn0mial, line 206: it behaves the same as if there was a conflict name. In the future, we could continue and report at the end. That would allow us to make some progress in case something conflicts.

pkg/reconciler/apis/apibinding/apibinding_indexes.go
```diff
@@ -86,3 +87,15 @@ func indexAPIBindingsByIdentityGroupResourceFunc(obj interface{}) ([]string, err
 func IdentityGroupResourceKeyFunc(identity, group, resource string) string {
 	return fmt.Sprintf("%s/%s/%s", identity, group, resource)
 }
+
+const indexCustomResourceDefinitionByWorkspace = "customResourceDefinitionByWorkspace"
```
sttts, line 91: it must be a unique name, by controller. We do `controllerName + "ByWorkspace"` elsewhere, until a more generic variant lands. Compare https://github.com/kcp-dev/apimachinery/pull/33 and its discussion.

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -155,6 +155,12 @@ func NewController(
 		},
 	})
 
+	if err := crdInformer.Informer().AddIndexers(cache.Indexers{
+		indexCustomResourceDefinitionByWorkspace: indexCustomResourceDefinitionByWorkspaceFunc,
```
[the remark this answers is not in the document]
p0lyn0mial: right, good point.

sttts:
Lgtm. Please squash and I tag.

/approve

p0lyn0mial:
@sttts pushed, have a look, before merging I'd to squash.

Issue #1193, opened 2022-06-02
Missing e2e tests for syncing in connection to finalizers


sttts:
From https://github.com/kcp-dev/kcp/issues/914:

- [ ] e2e tests for syncer finalizers implemenation
- [ ] e2e tests for per location external soft finalizers (finalizers.workloads.kcp.dev/<workload-cluster-id>)

PR #1228, opened 2022-06-07
Enable syncer finalizers (k8s & virtual) by default.

jmprusi:
## Summary

This PR enables the syncer finalizers, k8s-based and virtual ones by default, and the required per-location deletion timestamps. Those features were before feature-gated: https://github.com/kcp-dev/kcp/pull/942

closes https://github.com/kcp-dev/kcp/issues/1193

### DIFF
--- a/pkg/reconciler/workload/resource/resource_reconcile.go
+++ b/pkg/reconciler/workload/resource/resource_reconcile.go
@@ -68,6 +68,11 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName logical
 
 	annotationPatch, labelPatch := computePlacement(ns, obj)
 
+	// If the object DeletionTimestamp is set, we should set all locations deletion timestamps annotations to the same value.
+	if obj.GetDeletionTimestamp() != nil {
+		annotationPatch = propagateDeletionTimestamp(obj, annotationPatch)
+	}
+
 	// create patch
 	if len(labelPatch) == 0 && len(annotationPatch) == 0 {
 		return nil
@@ -109,6 +114,21 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName logical
 	}
 	klog.V(2).Infof("Patched cluster assignment for %q %s|%s/%s: labels=%v, annotations=%v", gvr, lclusterName, ns.Name, obj.GetName(), string(labelsString), string(annotationsString))
 	return nil
+}
+
+func propagateDeletionTimestamp(obj metav1.Object, annotationPatch map[string]interface{}) map[string]interface{} {
+	klog.V(3).Infof("Resource is being deleted; setting the deletion per locations timestamps for %s|%s/%s", logicalcluster.From(obj).String(), obj.GetNamespace(), obj.GetName())
+	objAnnotations := obj.GetAnnotations()
+	objLocations, _ := locations(objAnnotations, obj.GetLabels(), false)
+	if annotationPatch == nil {
+		annotationPatch = make(map[string]interface{})
+	}
+	for location := range objLocations {
+		if val, ok := objAnnotations[workloadv1alpha1.InternalClusterDeletionTimestampAnnotationPrefix+location]; !ok || val == "" {
+			annotationPatch[workloadv1alpha1.InternalClusterDeletionTimestampAnnotationPrefix+location] = obj.GetDeletionTimestamp().String()
+		}
+	}
+	return annotationPatch
 }
 
 // computePlacement computes the patch against annotations and labels. Nil means to remove the key.
--- a/pkg/reconciler/workload/resource/resource_reconcile.go
+++ b/pkg/reconciler/workload/resource/resource_reconcile.go
@@ -81,7 +81,7 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName logical
 		}
 	}
 	if len(annotationPatch) > 0 {
-		if err := unstructured.SetNestedField(patch, labelPatch, "metadata", "annotations"); err != nil {
+		if err := unstructured.SetNestedField(patch, annotationPatch, "metadata", "annotations"); err != nil {
 			klog.Errorf("unexpected unstructured error: %v", err)
 			return err // should never happen
 		}
--- a/pkg/syncer/spec/spec_process_test.go
+++ b/pkg/syncer/spec/spec_process_test.go
@@ -478,7 +478,7 @@ func TestSyncerProcess(t *testing.T) {
 		expectActionsOnFrom []clienttesting.Action
 		expectActionsOnTo   []clienttesting.Action
 	}{
-		"SpecSyncer upsert": {
+		"SpecSyncer sync deployment to downstream, upstream gets patched with the finalizer and the object is created downstream": {
 			upstreamLogicalCluster: "root:org:ws",
 			fromNamespace: namespace("test", "root:org:ws", map[string]string{
 				"state.internal.workload.kcp.dev/us-west1": "Sync",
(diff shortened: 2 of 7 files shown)


pkg/syncer/spec/spec_process_test.go
```diff
@@ -494,7 +494,7 @@ func TestSyncerProcess(t *testing.T) {
 					}),
 				deployment("theDeployment", "test", "root:org:ws", map[string]string{
 					"state.internal.workload.kcp.dev/us-west1": "Sync",
-				}, nil, nil),
+				}, nil, []string{"workload.kcp.dev/syncer-us-west1"}),
```
davidfestal, line 497: Could you add: " - syncer finalizer already there" in the test description ?

pkg/syncer/spec/spec_process_test.go
```diff
@@ -559,7 +559,7 @@ func TestSyncerProcess(t *testing.T) {
 				),
 			},
 		},
-		"SpecSyncer with AdvancedScheduling, sync downstream deployment": {
+		"SpecSyncer sync downstream deployment, upstream gets patched with the finalizer and the object is created downstream": {
```
davidfestal, line 562: Could we move this test case before the first one, and harmonize the start of test case name for both upsert test cases ?

> SpecSyncer upsert

or

> SpecSyncer sync to downstream

?

pkg/syncer/spec/spec_process_test.go
```diff
@@ -695,7 +694,7 @@ func TestSyncerProcess(t *testing.T) {
 				),
 			},
 		},
-		"SpecSyncer with AdvancedScheduling, deletion: object exist downstream": {
```
davidfestal, line 698: Can we gather the deletion test cases for the non-advanced use case so that they follow each other ?

pkg/syncer/status/status_process_test.go
```diff
@@ -383,7 +383,9 @@ func TestSyncerProcess(t *testing.T) {
 			workloadClusterName:                 "us-west1",
 
 			expectActionsOnFrom: []clienttesting.Action{},
-			expectActionsOnTo:   []clienttesting.Action{},
+			expectActionsOnTo: []clienttesting.Action{
+				getDeploymentAction("theDeployment", "test"),
```
davidfestal, line 387: Why is this action expected now (that wasn't before) ?

test/e2e/syncer/syncer_test.go
```diff
@@ -229,6 +247,69 @@ func TestSyncerLifecycle(t *testing.T) {
...
+	annotations["finalizers.workload.kcp.dev/"+syncerFixture.SyncerConfig.WorkloadClusterName] = "external-controller-finalizer"
+	upstreamDeployment.SetAnnotations(annotations)
+	_, err = upstreamKubeClient.AppsV1().Deployments(upstreamNamespace.Name).Update(ctx, upstreamDeployment, metav1.UpdateOptions{})
+	require.NoError(t, err)
+
+	t.Logf("Deleting upstream deployment %s/%s", upstreamNamespace.Name, upstreamDeployment.Name)
+	err = upstreamKubeClient.AppsV1().Deployments(upstreamNamespace.Name).Delete(ctx, upstreamDeployment.Name, metav1.DeleteOptions{GracePeriodSeconds: pointer.Int64(0)})
+	require.NoError(t, err)
+
+	t.Logf("Checking if upstream deployment %s/%s is getting deleted, shouldn't as there is a virtual finalizer", upstreamNamespace.Name, upstreamDeployment.Name)
```
davidfestal, line 269: Isn't it the real syncer finalizer that prevents the upstream object to be deleted (and not the virtual) ?

test/e2e/syncer/syncer_test.go
```diff
@@ -229,6 +247,69 @@ func TestSyncerLifecycle(t *testing.T) {
...
+				t.Errorf("Deployment %s/%s was deleted", upstreamNamespace.Name, upstreamDeployment.Name)
+			} else {
+				t.Errorf("saw an error trying to get the upstream deployment %s/%s, err: %v", upstreamNamespace.Name, upstreamDeployment.Name, err)
+			}
+			return true
+		}
+		return false
+	}, 5*time.Second, time.Second, "upstream Deployment got deleted or there was an error", upstreamNamespace.Name, upstreamDeployment.Name)
+
+	t.Logf("Checking if the downstream deployment %s/%s is deleted or not (shouldn't as there's a virtual finalizer)", upstreamNamespace.Name, upstreamDeployment.Name)
```
davidfestal, line 282: Should we test that the deletion annotation timestamp has been added to the upstream deployment here (with Eventually) ?

test/e2e/syncer/syncer_test.go
```diff
@@ -229,6 +247,69 @@ func TestSyncerLifecycle(t *testing.T) {
...
+	_, err = downstreamKubeClient.AppsV1().Deployments(downstreamNamespaceName).Get(ctx, upstreamDeployment.Name, metav1.GetOptions{})
```
davidfestal, line 283: Should we use Never here ?

pkg/reconciler/workload/resource/resource_reconcile.go
```diff
@@ -68,6 +68,22 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName logical
 
 	annotationPatch, labelPatch := computePlacement(ns, obj)
 
+	// If the object DeletionTimestamp is set, we should set all locations deletion timestamps annotations to the same value.
+	if obj.GetDeletionTimestamp() != nil {
+		klog.V(2).Infof("Resource is being deleted; setting the deletion per locations timestamps for %s|%s/%s", lclusterName, obj.GetNamespace(), obj.GetName())
```
sttts, line 73: V(3). Our rule of thumb: one V(2) line per reconcile.

pkg/reconciler/workload/resource/resource_reconcile.go
```diff
@@ -68,6 +68,22 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName logical
...
+
+		if annotationPatch == nil {
+			annotationPatch = make(map[string]interface{})
+		}
+		for location := range objLocations {
+			if val, ok := objAnnotations[workloadv1alpha1.InternalClusterDeletionTimestampAnnotationPrefix+location]; !ok || val == "" {
+				annotationPatch[workloadv1alpha1.InternalClusterDeletionTimestampAnnotationPrefix+location] = obj.GetDeletionTimestamp().String()
+			}
+		}
+	}
```
sttts, line 85: can we move that into `annotationPatch, labelPatch := propagateDeletionTimestamp(obj, annotationPath, labelPatch)` ?

sttts: and add a unit test for that func

sttts: question: can you annotate an object that is going to be deleted?

test/e2e/syncer/syncer_test.go
jmprusi: Without the virtual finalizer, the syncer will proceed to remove the object downstream and its finalizer. From the k8s point of view, yes it's the finalizer... but from the point of view of the user/external controller, it's the virtual finalizer.

pkg/syncer/status/status_process_test.go
jmprusi: As we removed some of the functions that were gated before: https://github.com/kcp-dev/kcp/pull/1228/files#diff-a4fd3bdfe5e149994aadff9dd072c7cbe8975e673dfcd4c2b7eba60f3c131482L103

Now the EnsureUpstreamFinalizerRemoved is called, doing a `Get` of the object.

pkg/syncer/spec/spec_process_test.go
jmprusi: done

pkg/syncer/spec/spec_process_test.go
jmprusi: done

pkg/syncer/spec/spec_process_test.go
jmprusi: done

test/e2e/syncer/syncer_test.go
jmprusi: done

test/e2e/syncer/syncer_test.go
jmprusi: done

pkg/reconciler/workload/resource/resource_reconcile.go
jmprusi: done

pkg/syncer/status/status_process_test.go
davidfestal: Ah, yes. Thanks for the detailed explanation that reminds me something !

sttts:
@davidfestal lgmt'ed the non-controller changes. Approving.

/lgtm
/approve

Issue #1061, opened 2022-05-16
Epic: Add support for quota


ncdc:
# Demo Objective
- [x] Set quota in a namespace in a workspace and have the accounting only apply to content in that workspace+namespace
- [x] Make sure quota works correctly for built-in types
- [x] Make sure quota works correctly for normal CRs, especially when 2 different workspaces both define the same GroupResource
- [x] Make sure quota works correctly for bound CRs
- [x] Find a way to do quota for cluster-scoped resources e.g. namespaces
    - [x] https://github.com/kcp-dev/kubernetes/pull/97
    - [x] #1846 

====================== 0.6 cut line

- [X] ~APIExport owner can define quota for instances of APIResourceSchemas they own, and the quota is applied in the bound workspace~ not part of the quota implementation / not specific to quota; moving to separate issue if needed.
    - [X] ~A workspace owner with "cluster-admin" is not able to manipulate this quota~ not part of the quota implementation / not specific to quota; moved to #1840

# Stories
- [x] #1236 
- [x] Add support + e2e for quota for normal CRs
- [x] Add support + e2e for quota for bound CRs
- [x] Add support + e2e for quota for cluster-scoped resources

maleck13:
So I think a key thing here is that a service offering an API can define how many of a certain object is allowed per workspace. I imagine this would happen when the APIBinding is created or possibly via an initializer if it is something that needs to be done for every workspace

sttts:
In addition there must be authz that stops the admin user in the workspace to tweak the quota for the APIBinding objects. We at least need a plan for that, implementation can come later.

sttts:
I completely miss a discussion about cluster scoped resources. Afaik Kube cannot restrict them at all. But we will need that. OpenShift has ClusterResourceQuota.

ncdc:
> I completely miss a discussion about cluster scoped resources. Afaik Kube cannot restrict them at all. But we will need that. OpenShift has ClusterResourceQuota.

Recording for posterity (we've chatted about this but not written it here): ClusterResourceQuota lets you quota namespace-scoped resources per cluster (e.g. "only allow 10 ConfigMaps in the entire cluster"). It does not quota cluster-scoped resources.

https://docs.google.com/document/d/15LR7fecEwh0Fupo5m3Yh3rtGy81eEGSZhh8heHqttAQ/edit is a design document for a short-term approach for quota of cluster-scoped resources.

ncdc:
Not yet. We need new epics/issues

PR #1236, opened 2022-06-08
Add initial per-workspace quota support

ncdc:
## Summary
Add initial per-workspace quota support. I have only tested object count with configmaps. I assume other resources will work too, but I haven't tested them.

Follow-up PRs will add support for quotas on namespaces in a workspace, making sure cpu/memory quota is enforced, etc.

## TODO
- [x] Address FIXMEs
- [x] Godoc
- [x] e2e

## Dependencies
- [x] https://github.com/kcp-dev/kubernetes/pull/75

## Related issue(s)

Part of #1061


ncdc:
```
$ k get resourcequotas -oyaml
apiVersion: v1
items:
- apiVersion: v1
  kind: ResourceQuota
  metadata:
    clusterName: root:default:qtest
    creationTimestamp: "2022-06-07T20:50:05Z"
    labels:
      now: "1654635898"
    name: test1
    namespace: default
    resourceVersion: "447"
    uid: be07f38b-7e36-4755-90fb-fe00a341894a
  spec:
    hard:
      configmaps: "2"
  status:
    hard:
      configmaps: "2"
    used:
      configmaps: "1"
kind: List
metadata:
  resourceVersion: ""
$ k create cm test1
configmap/test1 created
$ k create cm test2
error: failed to create configmap: configmaps "test2" is forbidden: exceeded quota: test1, requested: configmaps=1, used: configmaps=2, limited: configmaps=2

$ k ws
Current workspace is "root:default:qtest".
$ k ws use ..
Current workspace is "root:default".
$ k ws use qtest2
Current workspace is "root:default:qtest2" (type "Universal").
$ k get resourcequotas -oyaml
apiVersion: v1
items:
- apiVersion: v1
  kind: ResourceQuota
  metadata:
    clusterName: root:default:qtest2
    creationTimestamp: "2022-06-08T14:41:21Z"
    name: test2
    namespace: default
    resourceVersion: "439"
    uid: 1cfbdcfc-50d7-4871-96e3-27b83e5165dc
  spec:
    hard:
      configmaps: "3"
  status:
    hard:
      configmaps: "3"
    used:
      configmaps: "2"
kind: List
metadata:
  resourceVersion: ""
$ k get cm
NAME               DATA   AGE
kube-root-ca.crt   1      13m
test1              0      12m
$ k create cm abc1
configmap/abc1 created
$ k create cm abc2
error: failed to create configmap: configmaps "abc2" is forbidden: exceeded quota: test2, requested: configmaps=1, used: configmaps=3, limited: configmaps=3
```

pkg/admission/initializers/initializer.go
```diff
@@ -44,6 +47,26 @@ func (i *kcpInformersInitializer) Initialize(plugin admission.Interface) {
 	}
 }
 
+// NewKubeInformersInitializer returns an admission plugin initializer that injects
+// kube shared informer factories into admission plugins.
+func NewKubeInformersInitializer(
```
sttts, line 52: I assume the upstream one is not exposed?

pkg/admission/initializers/interfaces.go
```diff
@@ -29,6 +30,12 @@ type WantsKcpInformers interface {
 	SetKcpInformers(informers kcpinformers.SharedInformerFactory)
 }
 
+// WantsKubeInformers interface should be implemented by admission plugins that want to have a kube informer factory
+// injected.
+type WantsKubeInformers interface {
+	SetKubeInformers(informers informers.SharedInformerFactory)
+}
```
sttts, line 37: this definitely exists upstream

sttts: WantsExternalKubeInformerFactory

pkg/admission/kubequota/kubequota_admission.go
```diff
@@ -0,0 +1,264 @@
...
+func (k *KubeResourceQuota) getOrCreateDelegate(clusterName logicalcluster.Name) (*resourcequota.QuotaAdmission, error) {
+	k.lock.Lock()
+	defer k.lock.Unlock()
+
+	delegate := k.delegates[clusterName]
+	if delegate != nil {
+		return delegate, nil
+	}
+
+	// TODO cherry pick https://github.com/kubernetes/kubernetes/pull/110246 so we can get the right stop channel wired.
```
sttts, line 159: any way to make this code break here when we pick or rebase?

sttts: I don't see how the upstream PR gives you the right channel. Where should it come from? It must be connected to workspace deletion here?

pkg/admission/kubequota/kubequota_admission.go
```diff
@@ -0,0 +1,264 @@
...
+	cluster, err := genericapirequest.ValidClusterFrom(ctx)
+	if err != nil {
+		return err
+	}
+
+	var delegate *resourcequota.QuotaAdmission
+
+	k.lock.RLock()
+	delegate = k.delegates[cluster.Name]
+	k.lock.RUnlock()
```
sttts, line 135: would move this happy path logic into getOrCreateDelegate

pkg/admission/kubequota/kubequota_admission.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+}
+
+func (k *KubeResourceQuota) SetQuotaConfiguration(quotaConfiguration quota.Configuration) {
+	k.quotaConfiguration = quotaConfiguration
+}
+
+// SingleClusterKubeSharedInformerFactoryForResourceQuotas embeds an informers.SharedInformerFactory instance, overriding
+// .Core().V1().ResourceQuotas() to be scoped to a single logical cluster. All other informers are untouched.
+type SingleClusterKubeSharedInformerFactoryForResourceQuotas struct {
```
sttts, line 233: why public?

ScopedResourceQuotaInformerFactory?

pkg/reconciler/kubequota/kubequota_controller.go
```diff
@@ -0,0 +1,352 @@
...
+	// TODO(ncdc): could this be a partial metadata factory for everything except the few types that require full
+	// object details?
+	dynamicInformerFactory := dynamicinformer.NewFilteredDynamicSharedInformerFactoryWithOptions(
+		dynamicClient,
+		"",
+		nil,
+		cache.WithResyncPeriod(c.fullResyncPeriod),
+		cache.WithIndexers(cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}),
+	)
+	dynamicInformerFactory.Start(ctx.Done())
```
sttts, line 308: this seriously worries me, IMO a nogo for the PR. At the very least push the plumbing outside into main server.go. Nested informer creation is bad.

pkg/reconciler/kubequota/kubequota_controller.go
```diff
@@ -0,0 +1,352 @@
...
+
+	resourceQuotaController, err := resourcequota.NewController(resourceQuotaControllerOptions)
+	if err != nil {
+		return err
+	}
+
+	go resourceQuotaController.Run(ctx, c.workersPerLogicalCluster)
+
+	// Periodically the quota controller to detect new resource types
+	go resourceQuotaController.Sync(discoveryFunc, 30*time.Second, ctx.Done())
```
sttts, line 349: this is also by workspace? This does not scale :-/

pkg/admission/initializers/interfaces.go
ncdc: Fixed in squash commit

pkg/admission/kubequota/kubequota_admission.go
ncdc: Fixed in squash commit

pkg/admission/kubequota/kubequota_admission.go
ncdc: That's what I wrote about on the following line. What is the best way to do this, do you think? We could

- Use a new custom admission initializer that injects something that makes it easy to know when a cw is deleted
- Make a "mini controller" with a queue in this file that handles deletion events
- ... ?

pkg/admission/kubequota/kubequota_admission.go
ncdc: Fixed in squash commit. Unexported this as well as the 2 types below.

pkg/admission/initializers/interfaces.go
ncdc: Actually, this requires instantiating one of https://github.com/kcp-dev/kubernetes/blob/5bb0eeecf2cf172e4fc8b414c1953e64634a2a6a/staging/src/k8s.io/apiserver/pkg/admission/initializer/initializer.go#L27-L70, is that ok?

pkg/reconciler/kubequota/kubequota_controller.go
```diff
@@ -0,0 +1,352 @@
...
+	dynamicClusterClient dynamic.ClusterInterface
+	informersStarted     <-chan struct{}
+
+	// TODO(ncdc): see if these are interpreted correctly from upstream.
+	// quotaRecalculationPeriod controls how often a full quota recalculation is performed
+	quotaRecalculationPeriod time.Duration
+	// fullResyncPeriod controls how often the dynamic informers do a full resync
+	fullResyncPeriod time.Duration
+
+	workersPerLogicalCluster int
```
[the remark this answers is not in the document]
ncdc: Yes, the upstream controller runs 5 workers for the entire cluster by default. We'll need to run per workspace. We can drop this down to 1 to start and make it configurable? And long term, come up with a more scalable solution. WDYT?

pkg/reconciler/kubequota/kubequota_controller.go
ncdc: Agreed, this was to get it working.

pkg/reconciler/kubequota/kubequota_controller.go
ncdc: Yes, same comments as above & in Slack. Can come up with a scaling plan after we get this in, if that's ok.

pkg/admission/initializers/interfaces.go
ncdc: I got rid of our custom interface, but retained the custom impl. I don't want to init the upstream combo initializer.

pkg/admission/kubequota/kubequota_clusterworkspace_monitor.go
```diff
@@ -0,0 +1,125 @@
...
+	"k8s.io/client-go/tools/clusters"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+
+	tenancyinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/tenancy/v1alpha1"
+)
+
+const clusterWorkspaceDeletionMonitorControllerName = "kcp-kubequota-cluster-workspace-deletion-monitor"
+
+// clusterWorkspaceDeletionMonitor monitors ClusterWorkspaces and terminates QuotaAdmission for a logical cluster
```
stevekuznetsov, line 37: This could be a finalizer? Do we care if we miss an event here and leak a controller?

pkg/admission/kubequota/kubequota_clusterworkspace_monitor.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	// Clear rate limiting stats on key
+	m.queue.Forget(key)
+
+	return true
+}
+
+func (m *clusterWorkspaceDeletionMonitor) process(key string) error {
+	// e.g. root:org<separator>ws
```
stevekuznetsov, line 116: I found we lack the helpers to go in both directions here, might be useful to add to the logicalcluster lib later?

pkg/indexes/indexes.go
```diff
@@ -0,0 +1,22 @@
+/*
```
stevekuznetsov, line 1: Isn't the plural of index indices? :nerd_face:

pkg/server/controllers.go
```diff
@@ -829,6 +831,51 @@ func (s *Server) installVirtualWorkspaceURLsController(ctx context.Context, conf
...
+		replenishmentPeriod,
+		workersPerLogicalCluster,
+		s.syncedCh,
+	)
+
+	if err := server.AddPostStartHook(controllerName, func(hookContext genericapiserver.PostStartHookContext) error {
+		if err := s.waitForSync(hookContext.StopCh); err != nil {
+			klog.Errorf("failed to finish post-start-hook %s: %v", controllerName, err)
+			// nolint:nilerr
+			return nil // don't klog.Fatal. This only happens when context is cancelled.
```
stevekuznetsov, line 870: You could `errors.Is(context.Cancelled)` if you wanted to be sure

stevekuznetsov (commented):
The shape of this seems workable. I agree that O(workspaces) controllers and 5x goroutines per workspace seems like a bad time, /shrug

pkg/reconciler/kubequota/delegating_event_handler.go
```diff
@@ -0,0 +1,97 @@
...
+	return groupResourceHandlers[clusterName]
+}
+
+func (d *delegatingEventHandler) registerEventHandler(resource schema.GroupResource, clusterName logicalcluster.Name, h cache.ResourceEventHandler) {
+	d.lock.Lock()
+	defer d.lock.Unlock()
+
+	groupResourceHandlers, ok := d.eventHandlers[resource]
+	if !ok {
+		panic("bad")
```
stevekuznetsov, line 93: lol

pkg/reconciler/kubequota/kubequota_controller.go
```diff
@@ -0,0 +1,423 @@
...
+		},
+	}
+
+	clusterWorkspacesInformer.Informer().AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj interface{}) {
+				c.enqueue(obj)
+			},
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				c.enqueue(oldObj)
```
stevekuznetsov, line 129: Why old?

pkg/reconciler/kubequota/kubequota_controller.go
```diff
@@ -0,0 +1,423 @@
...
+				c.enqueue(obj)
+			},
+		},
+	)
+
+	// This sets up a single global event handler on ResourceQuotas. When a resource quota controller is created for
+	// a logical cluster, it registers its own event handler, which is stored in Controller.resourceQuotaEventHandlers.
+	// When an event comes in, this handler looks up the actual handler for the object's logical cluster and invokes it.
+	resourceQuotaInformer.Informer().AddEventHandlerWithResyncPeriod(
+		cache.ResourceEventHandlerFuncs{
```
stevekuznetsov, line 141: This looks a lot like `delegatingEventHandler` - what's different?

stevekuznetsov (commented):
Still seems reasonable to me with the obvious caveats already being pointed out :)

pkg/reconciler/kubequota/kubequota_controller.go
ncdc: Shouldn't matter, I don't think. enqueue just converts it to a key. I can do newObj if that seems better?

MikeSpreitzer:
@MikeSpreitzer

pkg/reconciler/kubequota/delegating_event_handler.go
ncdc: Fixed

ncdc:
Working on some fixes

pkg/server/server.go
```diff
@@ -333,13 +356,13 @@ func (s *Server) Run(ctx context.Context) error {
...
 	}
-	ddsif := informer.NewDynamicDiscoverySharedInformerFactory(
+	s.dynamicDiscoverySharedInformerFactory = informer.NewDynamicDiscoverySharedInformerFactory(
 		s.kcpSharedInformerFactory.Tenancy().V1alpha1().ClusterWorkspaces().Lister(),
 		kubeClusterClient.DiscoveryClient,
 		metadataClusterClient.Cluster(logicalcluster.Wildcard),
 		func(obj interface{}) bool { return true }, s.options.Extra.DiscoveryPollInterval,
 	)
-	if err := ddsif.AddIndexers(cache.Indexers{byWorkspace: indexByWorkspace}); err != nil {
+	if err := s.dynamicDiscoverySharedInformerFactory.AddIndexers(indexers.NamespaceScoped()); err != nil {
```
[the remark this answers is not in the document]
ncdc: It's still there. See indexers.go

ncdc:
e2e-multiple failed on:
```
  controller_test.go:231: Recreate the CRD, and then quickly a namespace and a CR whose CRD was just recreated
I0624 16:35:49.871453   40377 bootstrap.go:135] Bootstrapped CRD root:e2e-org-q78cm:e2e-workspace-ptgwm|sheriffs.f58e12e4-a87e-47cd-908b-195d1333c9cf.io after 9.194251ms
    controller_test.go:239: Now also the sheriff should be scheduled
    controller_test.go:240: 
        	Error Trace:	controller_test.go:240
        	            				controller_test.go:297
        	Test:       	TestNamespaceScheduler/GVRs_are_removed,_and_then_quickly_re-added_to_a_new_workspace
        	Messages:   	failed to see sheriff scheduled
E0624 16:36:22.085592   40377 controller_test.go:243] failed to get sheriff: context canceled
    --- FAIL: TestNamespaceScheduler/GVRs_are_removed,_and_then_quickly_re-added_to_a_new_workspace (46.09s)
```

ncdc:
/hold
while I figure out these failures

ncdc:
First 2 commits are #1372

ncdc:
Depends on https://github.com/kcp-dev/kubernetes/pull/82

I think I have cleared all the flakiness in my newly added quota e2e test. I'm able to run the tests for that package with `-count 20` and they all pass. Hopefully that'll be the case in a full e2e run too 😄.

/unhold

ncdc:
e2e-sharded seems to have flaked on `TestWorkspaceDeletionController/nested_worksapce_cleanup_when_an_org_workspace_is_deleted`. See https://kubernetes.slack.com/archives/C021U8WSAFK/p1658264901447379 for more details.

hack/bump-k8s.sh
```diff
@@ -34,6 +34,6 @@ BRANCH=${BRANCH:-feature-logical-clusters-1.23}
 current_version="$( GOPROXY=direct go mod edit -json | jq '.Replace[] | select(.Old.Path=="k8s.io/kubernetes") | .New.Version' --raw-output )"
 
 # equivalent to go mod edit -replace
-sed -i'' -e "s|${current_version}|${BRANCH}|g" -E -e "s,=> github.com/[^/]+/[a-zA-Z0-9_-]+,=> github.com/${GITHUB_USER}/${GITHUB_REPO},g" go.mod
+sed -i '' -e "s|${current_version}|${BRANCH}|g" -E -e "s,=> github.com/[^/]+/[a-zA-Z0-9_-]+,=> github.com/${GITHUB_USER}/${GITHUB_REPO},g" go.mod
```
stevekuznetsov, line 37: I thought this was supposed to work for both of us? Damn it. Can we do a conditional here based on the platform lol?

pkg/admission/initializers/initializer.go
stevekuznetsov: Does this need anything ... ?

pkg/admission/kubequota/kubequota_admission.go
```diff
@@ -0,0 +1,359 @@
...
+	if err != nil {
+		return err
+	}
+
+	delegate, err := k.getOrCreateDelegate(cluster.Name)
+	if err != nil {
+		return err
+	}
+
+	// hackyA := &hackyAdmissionAttributesForBuiltInTypes{
```
stevekuznetsov, line 148: What's going on here?

pkg/admission/kubequota/kubequota_admission.go
```diff
@@ -0,0 +1,359 @@
...
+	// 	Attributes: a,
+	// }
+
+	return delegate.Validate(ctx, a, o)
+}
+
+// TODO(ncdc): find a way to support the special evaluators for pods/pvcs/services.
```
stevekuznetsov, line 155: Important? I know you said you'd changed the approach here ...

pkg/admission/kubequota/kubequota_admission.go
```diff
@@ -0,0 +1,359 @@
...
+	k.lock.Lock()
+	defer k.lock.Unlock()
+
+	delegate = k.delegates[clusterName]
+	if delegate != nil {
+		return delegate, nil
+	}
+
+	// Set up a context that is cancelable and that is bounded by k.ServerDone
+	ctx, cancel := context.WithCancel(context.Background())
```
stevekuznetsov, line 243: Do we not have some higher context here for the server that's not `Background()`?

pkg/admission/kubequota/kubequota_admission.go
```diff
@@ -0,0 +1,359 @@
...
+	go func(done <-chan struct{}) {
+		<-done
+		cancel()
+	}(k.serverDone)
+
+	quotaAdmission, err := resourcequota.NewResourceQuota(k.userSuppliedConfiguration, 5, ctx.Done())
```
stevekuznetsov, line 249: Is that 5 workers per?

pkg/admission/kubequota/kubequota_clusterworkspace_monitor.go
stevekuznetsov: I think the comment stands?

pkg/informer/informer.go
```diff
@@ -491,3 +562,17 @@ func (d *DynamicDiscoverySharedInformerFactory) calculateInformersLockHeld(lates
 
 	return toAdd, toRemove
 }
+
+// Subscribe registers ch for informer/discovery change notifications.
+func (d *DynamicDiscoverySharedInformerFactory) Subscribe(id string, ch chan<- struct{}) {
```
stevekuznetsov, line 567: If we let users pass us their end of the channel, we need to handle being blocked on sending a notification. Alternatively, this can *return* them a channel (and we can choose to make it buffered, but then also need to handle the case where we can't notify them within some timeout)

pkg/reconciler/kubequota/kubequota_controller.go
stevekuznetsov: Key should be the same, so no worries

pkg/reconciler/kubequota/kubequota_controller.go
stevekuznetsov: We have generics soon ;)

pkg/reconciler/kubequota/kubequota_controller.go
```diff
@@ -0,0 +1,384 @@
...
+	discoveryChangedCh := make(chan struct{}, 1)
+
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-discoveryChangedCh:
+				klog.V(4).Infof("%s: got API change notification", clusterName)
+				resourceQuotaController.KCPUpdateMonitors(ctx, clusterName, c.dynamicDiscoverySharedInformerFactory.DiscoveryData)
```
stevekuznetsov, line 374: Yeah esp since this is going to wait for the informers to sync in the critical path between consuming discovery events we should ensure the send-side of discoveryChangedCh can handle timeouts or something

pkg/reconciler/kubequota/single_cluster_resource_quota_informer.go
```diff
@@ -0,0 +1,187 @@
...
+	// the following:
+	//
+	// 		namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	// 		if err != nil {
+	// 			return err
+	// 		}
+	// 		resourceQuota, err := rq.rqLister.ResourceQuotas(namespace).Get(name)
+	//
+	// And in this case, when you split key into namespace and name, "name" is actually a clusterAwareName, meaning
+	// the below construction of key is correct.
```
stevekuznetsov, line 174: :popcorn:

pkg/admission/kubequota/kubequota_admission.go
ncdc: That's actually what the goroutine right after this is doing. It's taking k.serverDone (which is actually the `ctx` passed to `Server.Run(ctx)` where `Server` is the main kcp server.... although ...

Looking at this again, I think I'm only passing this ctx's Done ch to resourcequota.NewResourceQuota, so I could probably scrap this and just do `resourcequota.NewResourceQuota(k.userSuppliedConfiguration, 5, k.serverDone)`, right?

ncdc: Oh wait, nevermind on that. I need to be able to call `cancel()` per instance when a ClusterWorkspace is deleted.

ncdc: But the goroutine as written probably isn't ideal - it lives as long as the server does, and ideally it terminates when a ClusterWorkspace is deleted too.

pkg/reconciler/kubequota/kubequota_controller.go
ncdc: My thinking is:
- Subscriber should use a buffered channel (as in the previous comment, happy to change this to return a new buffered channel to a subscriber)
- Sender (ddsif) uses the select pattern to try to send to the channel, with a default clause to avoid any blocks (and therefore dropping the notification)
- Because the channel is buffered, when the ddsif wants to send a notification, one of the following two statements must be true:
    1. The channel has capacity to receive, so the send to the channel will succeed, and then the subscriber in theory will pick it up
    2. The channel is full, so it's safe to drop the send to the channel (the subscriber hasn't picked up the notification yet, so there is no need to notify it again; i.e., deduplicate)

pkg/admission/kubequota/kubequota_admission.go
ncdc: Comment removed in next push

hack/bump-k8s.sh
sttts:
```suggestion
sed -i -e "s|${current_version}|${BRANCH}|g" -E -e "s,=> github.com/[^/]+/[a-zA-Z0-9_-]+,=> github.com/${GITHUB_USER}/${GITHUB_REPO},g" go.mod
```
Why do you think the `-i''` does not work? Seems to work here as long as `-e` is following, and you can leave out the`''`, both with sed and gsed.

pkg/informer/informer.go
```diff
@@ -454,6 +457,56 @@ func (d *DynamicDiscoverySharedInformerFactory) updateInformers() {
...
+	return discoveryData
+}
+
+// DiscoveryData implements resourcequota.NamespacedResourcesFunc and is intended to be used by the quota subsystem.
+func (d *DynamicDiscoverySharedInformerFactory) DiscoveryData() ([]*metav1.APIResourceList, error) {
+	d.informersLock.RLock()
+
+	ret := make([]*metav1.APIResourceList, len(d.discoveryData))
+	for i, apiResourceList := range d.discoveryData {
+		ret[i] = apiResourceList.DeepCopy()
```
sttts, line 521: why deepcopy?

pkg/informer/informer.go
```diff
@@ -454,6 +457,56 @@ func (d *DynamicDiscoverySharedInformerFactory) updateInformers() {
...
+	}
+
+	d.informersLock.RUnlock()
```
sttts, line 524: why not the defer variant if it is possible? Seems to be simpler and safer.

pkg/informer/informer.go
```diff
@@ -454,6 +457,56 @@ func (d *DynamicDiscoverySharedInformerFactory) updateInformers() {
 		delete(d.informerStops, gvr)
 		delete(d.startedInformers, gvr)
 	}
+
+	d.discoveryData = gvrsToDiscoveryData(latest)
+}
+
+func gvrsToDiscoveryData(gvrs map[schema.GroupVersionResource]struct{}) []*metav1.APIResourceList {
```
sttts, line 481: so this is actually fake discovery data? Can we make this clear in the go doc?

pkg/informer/informer.go
sttts: this seems brittle. This means we drop events as soon as ch has no capacity at the moment. We cannot do that.

IMO we should not use channels at all. It's a poor type for APIs. Make this a function call. Document that the consumer must make sure to return quickly and not block. The consumer will know how to make that work. It has the context to do that. Here we don't.

pkg/informer/informer.go
```diff
@@ -544,3 +562,17 @@ func (d *DynamicDiscoverySharedInformerFactory) calculateInformersLockHeld(lates
 
 	return toAdd, toRemove
 }
+
+// Subscribe registers ch for informer/discovery change notifications.
+func (d *DynamicDiscoverySharedInformerFactory) Subscribe(id string, ch chan<- struct{}) {
+	d.subscribersLock.Lock()
```
sttts, line 567: defer

pkg/informer/informer.go
```diff
@@ -544,3 +562,17 @@ func (d *DynamicDiscoverySharedInformerFactory) calculateInformersLockHeld(lates
...
+	d.subscribers[id] = ch
+	d.subscribersLock.Unlock()
+}
+
+// Unsubscribe removes the channel associated with id from future informer/discovery change notifications.
+func (d *DynamicDiscoverySharedInformerFactory) Unsubscribe(id string) {
+	d.subscribersLock.Lock()
```
sttts, line 574: defer

pkg/informer/informer.go
```diff
@@ -156,13 +156,12 @@ func NewDynamicDiscoverySharedInformerFactory(
 		}
 	}
 
+	// When CRDs are added or deleted, send a notification that we might need to add/remove informers. We don't
+	// need to handle update events.
```
sttts, line 160: technically there is a time between creation and being Established. Does this mean we ignore that here? (we can, am not against it. But we should be clear in the go doc).

pkg/admission/initializers/initializer.go
```diff
@@ -44,6 +47,26 @@ func (i *kcpInformersInitializer) Initialize(plugin admission.Interface) {
...
+type kubeInformersInitializer struct {
+	kubeInformers informers.SharedInformerFactory
+}
+
+func (i *kubeInformersInitializer) Initialize(plugin admission.Interface) {
+	if wants, ok := plugin.(initializer.WantsExternalKubeInformerFactory); ok {
+		wants.SetExternalKubeInformerFactory(i.kubeInformers)
+	}
+}
+
```
sttts, line 69: this is preexisting. Why another initializer? Compae `WantsExternalKubeInformerFactory`.

pkg/admission/initializers/initializer.go
```diff
@@ -139,3 +162,39 @@ func (i *shardExternalURLInitializer) Initialize(plugin admission.Interface) {
...
+type kubeQuotaConfigurationInitializer struct {
+	quotaConfiguration quota.Configuration
+}
+
+func (i *kubeQuotaConfigurationInitializer) Initialize(plugin admission.Interface) {
+	if wants, ok := plugin.(initializer.WantsQuotaConfiguration); ok {
+		wants.SetQuotaConfiguration(i.quotaConfiguration)
+	}
+}
```
sttts, line 182: this is preexisting `initializer.WantsQuotaConfiguration`

pkg/admission/kubequota/kubequota_admission.go
```diff
@@ -0,0 +1,292 @@
...
+	quotaConfiguration       quota.Configuration
+	serverDone               <-chan struct{}
+
+	// Manually set
+	userSuppliedConfiguration *resourcequotaapi.Configuration
+
+	lock      sync.RWMutex
+	delegates map[logicalcluster.Name]*stoppableQuotaAdmission
+
+	once sync.Once
```
sttts, line 99: once what?

pkg/admission/kubequota/kubequota_admission.go
```diff
@@ -0,0 +1,292 @@
...
+
+	return delegate.Validate(ctx, a, o)
+}
+
+// getOrCreateDelegate creates a resourcequota.QuotaAdmission plugin for clusterName.
+func (k *KubeResourceQuota) getOrCreateDelegate(clusterName logicalcluster.Name) (*stoppableQuotaAdmission, error) {
+	var delegate *stoppableQuotaAdmission
+
+	k.lock.RLock()
+	delegate = k.delegates[clusterName]
```
sttts, line 156: `:=` and drop the var

pkg/admission/kubequota/kubequota_admission.go
sttts: our workspaces are smaller than kube clusters. Can we reduce 5 ?

pkg/admission/kubequota/kubequota_admission.go
```diff
@@ -0,0 +1,292 @@
...
+		stopFunc:       cancel,
+	}
+
+	delegate.SetExternalKubeInformerFactory(&scopedResourceQuotaInformerFactory{
+		SharedInformerFactory: k.kubeInformers,
+		clusterName:           clusterName,
+	})
+
+	delegate.SetExternalKubeClientSet(k.kubeClusterClient.Cluster(clusterName))
+
```
sttts, line 198: nit: empty lines are free, but don't help here with readability :)

pkg/admission/kubequota/kubequota_admission.go
```diff
@@ -0,0 +1,292 @@
...
+		select {
+		case <-ctx.Done():
+		case <-k.serverDone:
+			cancel()
+		}
+	}()
+
+	quotaAdmission, err := resourcequota.NewResourceQuota(k.userSuppliedConfiguration, 5, ctx.Done())
+	if err != nil {
+		return nil, err
```
sttts, line 184: call cancel?

pkg/admission/kubequota/kubequota_admission.go
```diff
@@ -0,0 +1,292 @@
...
+		SharedInformerFactory: k.kubeInformers,
+		clusterName:           clusterName,
+	})
+
+	delegate.SetExternalKubeClientSet(k.kubeClusterClient.Cluster(clusterName))
+
+	delegate.SetQuotaConfiguration(k.quotaConfiguration)
+
+	if err := delegate.ValidateInitialization(); err != nil {
+		return nil, err
```
sttts, line 202: call cancel?

pkg/admission/kubequota/kubequota_admission.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+
+	k.delegates[clusterName] = delegate
+
+	return delegate, nil
+}
+
+type stoppableQuotaAdmission struct {
+	*resourcequota.QuotaAdmission
+	stopFunc func()
```
sttts, line 212: rename to `stop` and drop `stop` method

pkg/admission/kubequota/kubequota_admission.go
```diff
@@ -0,0 +1,292 @@
...
+func (s *stoppableQuotaAdmission) stop() {
+	s.stopFunc()
+}
+
+func (k *KubeResourceQuota) stopQuotaAdmissionForCluster(clusterName logicalcluster.Name) {
+	k.lock.Lock()
+	defer k.lock.Unlock()
+
+	delegate := k.delegates[clusterName]
+
```
sttts, line 224: drop empty line

pkg/admission/kubequota/kubequota_clusterworkspace_monitor.go
sttts: finalizer on the object to manage in-memory objects? We cannot do that. It will break in HA. Finalizers are for API life-cycle, not process life-cycle.

pkg/reconciler/kubequota/delegating_event_handler.go
```diff
@@ -0,0 +1,103 @@
...
+
+		informer.AddEventHandler(
+			cache.ResourceEventHandlerFuncs{
+				AddFunc: func(obj interface{}) {
+					h := d.getEventHandler(resource, obj)
+					if h == nil {
+						return
+					}
+					h.OnAdd(obj)
+				},
```
sttts, line 65: nit:
```
if h := d.getEventHandler(resource, obj); h != nil { h.OnAdd(obj) }
```

pkg/reconciler/kubequota/kubequota_controller.go
```diff
@@ -0,0 +1,384 @@
...
+		},
+
+		listCRDs: func() ([]*apiextensionsv1.CustomResourceDefinition, error) {
+			return crdInformer.Lister().List(labels.Everything())
+		},
+	}
+
+	clusterWorkspacesInformer.Informer().AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj interface{}) {
```
sttts, line 128: nit: `AddFunc: c.enqueue`

pkg/reconciler/kubequota/kubequota_controller.go
```diff
@@ -0,0 +1,384 @@
...
+}
+
+// getResourceQuotaEventHandler returns the ResourceQuota cache.ResourceEventHandler for obj's logical cluster.
+func (c *Controller) getResourceQuotaEventHandler(obj interface{}) cache.ResourceEventHandler {
+	cluster := clusterNameForObj(obj)
+	if cluster.Empty() {
+		return nil
+	}
+
+	c.lock.RLock()
```
sttts, line 197: defer

pkg/reconciler/kubequota/kubequota_controller.go
```diff
@@ -0,0 +1,384 @@
...
+	// listerFuncForResource := generic.ListerFuncForResourceFunc(scopedInformerFactory.ForResource)
+	// quotaConfiguration := install.NewQuotaConfigurationForControllers(listerFuncForResource)
+	quotaConfiguration := generic.NewConfiguration(nil, install.DefaultIgnoredResources())
+
+	resourceQuotaControllerOptions := &resourcequota.ControllerOptions{
+		QuotaClient: resourceQuotaControllerClient.CoreV1(),
+		ResourceQuotaInformer: &SingleClusterResourceQuotaInformer{
+			clusterName:     clusterName,
+			delegate:        c.resourceQuotaInformer,
+			registerHandler: c.registerResourceQuotaEventHandlerLocked,
```
sttts, line 336: have a hard time to see why it is guaranteed here that the lock is taken. Leaking a `*Locked` func into another object is suspicious.

pkg/reconciler/kubequota/kubequota_controller.go
sttts: See my comments. Ddsif should not care about consumers and how they implement this. Make it a contract that the handler func does not block.

pkg/server/controllers.go
sttts: please don't change this here. This code is everywhere. Either change all locations or non. We know there is only this error that can happen. PostStartHooks have braindead handling of errors. Returning an error is of no use, but harmful (= distracts from the actual problem during startup).

test/e2e/fixtures/apifixtures/sheriffs.go
```diff
@@ -166,21 +166,26 @@ func CreateSheriff(
...
-		}, metav1.CreateOptions{}); err != nil {
+		if _, err := dynamicClusterClient.Cluster(clusterName).Resource(sheriffsGVR).Namespace("default").Create(ctx, NewSheriff(group, name), metav1.CreateOptions{}); err != nil {
 			return false, fmt.Sprintf("failed to create Sheriff %s|%s: %v", clusterName, name, err.Error())
 		}
 		return true, ""
 	}, wait.ForeverTestTimeout, time.Millisecond*100, "error creating Sheriff %s|%s", clusterName, name)
 }
 
+// NewSheriff returns a new *unstructured.Unstructured for a Sheriff with the given group and name.
+func NewSheriff(group, name string) *unstructured.Unstructured {
```
sttts, line 177: please no shared helpers for fixtures. Use whatever builder pattern you like locally. But global helpers are of no value and make code hard to reason about. People will extend these and build a shared fixture hell where every test depends on many other.

test/e2e/quota/quota_test.go
```diff
@@ -0,0 +1,339 @@
...
+
+	kubeClusterClient, err := kubernetes.NewClusterForConfig(cfg)
+	require.NoError(t, err, "error creating kube cluster client")
+
+	orgClusterName := framework.NewOrganizationFixture(t, server)
+
+	// Create more than 1 workspace with the same quota restrictions to validate that after we create the first workspace
+	// and fill its quota to capacity, subsequent workspaces have independent quota.
+	for i := 0; i < 3; i++ {
+		t.Logf("Creating workspace %d", i)
```
sttts, line 69: use `NewWorkspaceFixture(...., WithName(fmt.Sprintf(...)))`. And this func already logs the name.

sttts: and while at it, make it `WithName(fmt string, flags ... interface{})`

test/e2e/quota/quota_test.go
```diff
@@ -0,0 +1,339 @@
...
+
+			_, err = kubeClusterClient.Cluster(userClusterName).CoreV1().ResourceQuotas("default").Create(ctx, quota, metav1.CreateOptions{})
+			require.NoError(t, err, "error creating quota")
+
+			t.Logf("Waiting for quota to show 0 used Services")
+			framework.Eventually(t, func() (bool, string) {
+				quota, err = kubeClusterClient.Cluster(userClusterName).CoreV1().ResourceQuotas("default").Get(ctx, "quota", metav1.GetOptions{})
+				if err != nil {
+					return false, fmt.Sprintf("Error getting ws quota %s|default/quota: %v", userClusterName, err)
+				}
```
sttts, line 216: shouldn't we fail hard here with `require.NoError`? I cannot see why any error here is acceptable.

test/e2e/quota/quota_test.go
```diff
@@ -0,0 +1,339 @@
...
+				},
+			}
+
+			_, err = kcpClusterClient.Cluster(userClusterName).ApisV1alpha1().APIBindings().Create(ctx, binding, metav1.CreateOptions{})
+			require.NoError(t, err)
+
+			t.Logf("Wait for binding to be ready")
+			framework.Eventually(t, func() (bool, string) {
+				binding, err := kcpClusterClient.Cluster(userClusterName).ApisV1alpha1().APIBindings().Get(ctx, binding.Name, metav1.GetOptions{})
+				if err != nil {
```
sttts, line 181: `require.NoError`. I don't think any error here is acceptable.

test/e2e/quota/quota_test.go
```diff
@@ -0,0 +1,339 @@
...
+		}
+
+		t.Logf("Creating ws quota")
+		ws1Quota, err = kubeClusterClient.Cluster(ws).CoreV1().ResourceQuotas("default").Create(ctx, ws1Quota, metav1.CreateOptions{})
+		require.NoError(t, err, "error creating ws quota")
+
+		t.Logf("Waiting for ws quota to show used configmaps (kube-root-ca.crt)")
+		framework.Eventually(t, func() (bool, string) {
+			ws1Quota, err = kubeClusterClient.Cluster(ws).CoreV1().ResourceQuotas("default").Get(ctx, "quota", metav1.GetOptions{})
+			if err != nil {
```
sttts, line 90: require.NoError. I cannot see that any error here is acceptable.

hack/bump-k8s.sh
stevekuznetsov: Because I'm on Linux and added the space to make it work for me last time I used the script :)

pkg/informer/informer.go
stevekuznetsov: FWIW this is how watch server works, so while brittle it's likely workable?

pkg/admission/kubequota/kubequota_admission.go
ncdc: Have lowered to 1 for now (in next push)

pkg/informer/informer.go
ncdc: I would prefer to keep this as-is, for a few of reasons:

1. The event dropping is on purpose. I want to deduplicate notifications, and this is easily achieved with a buffer of size 1.
2. The consumer in this case is not guaranteed to return quickly. It is doing a WaitForCacheSync. It currently waits "forever" (until either server shutdown, or another notification comes in (coming in next push))
3. I need to be able to cancel the consumer's WaitForCacheSync, which I can't easily do with a single function call

pkg/informer/informer.go
ncdc: To avoid consumers mutating the cached data

pkg/informer/informer.go
ncdc: I remember Steve saying it's faster without the defer, but I'll switch it in the next push

pkg/informer/informer.go
ncdc: Will do in next push

pkg/informer/informer.go
```diff
@@ -459,6 +465,18 @@ func (d *DynamicDiscoverySharedInformerFactory) updateInformers() {
 	}
 
 	d.discoveryData = gvrsToDiscoveryData(latest)
+
+	d.subscribersLock.Lock()
```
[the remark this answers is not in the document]
ncdc: Next push

pkg/informer/informer.go
ncdc: Next push

pkg/informer/informer.go
ncdc: Next push

pkg/admission/initializers/initializer.go
ncdc: I am using upstream's `WantsExternalKubeInformerFactory` but if we want to reuse upstream's actual initializer, we have to instantiate one of https://github.com/kcp-dev/kubernetes/blob/feature-logical-clusters-1.24/staging/src/k8s.io/apiserver/pkg/admission/initializer/initializer.go, which didn't seem like the best approach.

pkg/admission/initializers/initializer.go
ncdc: Similar to the preceding comment, it didn't seem like the best approach to instantiate one of https://github.com/kcp-dev/kubernetes/blob/feature-logical-clusters-1.24/pkg/kubeapiserver/admission/initializer.go.

pkg/admission/kubequota/kubequota_admission.go
ncdc: Renamed

pkg/admission/kubequota/kubequota_admission.go
ncdc: Fixed

pkg/admission/kubequota/kubequota_admission.go
ncdc: Fixed

pkg/admission/kubequota/kubequota_admission.go
ncdc: Adjusted

pkg/admission/kubequota/kubequota_admission.go
ncdc: Fixed

pkg/admission/kubequota/kubequota_admission.go
ncdc: Fixed

pkg/admission/kubequota/kubequota_admission.go
ncdc: I prefer to have a blank line before the return

pkg/reconciler/kubequota/delegating_event_handler.go
ncdc: Fixed

pkg/reconciler/kubequota/kubequota_controller.go
ncdc: Yeah, I know what you mean. Here is the flow:

- kcp kubequota Controller.process
- it grabs the lock
- calls startQuotaForClusterWorkspace
- calls resourcequota.NewController(resourceQuotaControllerOptions)
- calls options.ResourceQuotaInformer.Informer().AddEventHandlerWithResyncPeriod(
- this is to our SingleClusterResourceQuotaInformer/delegatingInformer
- calls c.registerResourceQuotaEventHandlerLocked

But, let me make add a "level above" and do something similar to `scopingGenericSharedInformerFactory`. Stay tuned. 😄

pkg/admission/initializers/initializer.go
ncdc: Nevermind, I didn't realize this was set up by default in the kube code 😄 Removing.

pkg/admission/initializers/initializer.go
ncdc: That combines a RESTMapper with the QuotaConfiguration. Not ideal for us.

Also, this is constructed in https://github.com/kcp-dev/kubernetes/blob/feature-logical-clusters-1.24/pkg/genericcontrolplane/admission/config.go, which I don't think is used anywhere. But it constructs `quotainstall.NewQuotaConfigurationForAdmission()`, which includes the custom pod/pvc/service evaluators which we can't use right now. I create a different configuration in kcp's server.go.

test/e2e/quota/quota_test.go
ncdc: Will change

ncdc:
Just pushed an update. Have rebased & resolved most of the code review comments. Have not changed the channel notification strategy between ddsif and quota. Please see my new comments there. We should probably discuss live tomorrow/soon. Thanks for the feedback!

pkg/reconciler/kubequota/kubequota_controller.go
sttts: 👀

ncdc:
e2e-multiple-runs flaked on #1634

pkg/informer/informer.go
sttts: defer please if possible

pkg/informer/informer.go
sttts: still valid

pkg/informer/informer.go
sttts: > The event dropping is on purpose. I want to deduplicate notifications, and this is easily achieved with a buffer of size 1.

It's not the duty of the informer to handle that.

> The consumer in this case is not guaranteed to return quickly. It is doing a WaitForCacheSync. It currently waits "forever" (until either server shutdown, or another notification comes in (coming in next push))

It is not the duty of the informer to know what the subscribers do or don't.

> I need to be able to cancel the consumer's WaitForCacheSync, which I can't easily do with a single function call

Why?

sttts: > FWIW this is how watch server works, so while brittle it's likely workable?

It's not about whether it technically works or not in other context. It's about making complexity local. The current interface does not archive that.

hack/bump-k8s.sh
sttts: What am I missing? Works like a charm.

Bildschirmfoto 2022-07-28 um 09 46 12

pkg/admission/initializers/initializer.go
sttts: ack

pkg/informer/informer.go
```diff
@@ -68,6 +70,11 @@ type DynamicDiscoverySharedInformerFactory struct {
 	informers        map[schema.GroupVersionResource]informers.GenericInformer
 	startedInformers map[schema.GroupVersionResource]bool
 	informerStops    map[schema.GroupVersionResource]chan struct{}
+	discoveryData    []*metav1.APIResourceList
+
+	// Support subscribers (e.g. quota) that want to know when informers/discovery have changed.
+	subscribersLock sync.Mutex
+	subscribers     map[string]chan<- struct{}
```
sttts, line 77: my comment still stands. This is a bad interface. Use a function call please and document that these must return instantly.

pkg/informer/informer.go
sttts: Was merely talking about `DiscoveryData()`. Somebody will build upon it and spend lots of time to find out eventually that the APIResources are not complete. Either we make them complete, or we should call `DiscoveryData()` more like `FakeDiscoveryData()`.

sttts: or even `QuotaDiscoveryData()`

pkg/informer/informer.go
sttts: Can't this be part of the signature spec? DeepCopy in advance has smell.

pkg/reconciler/kubequota/delegating_event_handler.go
```diff
@@ -0,0 +1,123 @@
...
+				h.OnDelete(obj)
+			}
+		},
+	}
+}
+
+// getEventHandler returns a cache.ResourceEventHandler for resource and the logicalcluster.Name for obj.
+func (d *delegatingEventHandler) getEventHandler(resource schema.GroupResource, obj interface{}) cache.ResourceEventHandler {
+	clusterName := clusterNameForObj(obj)
+	if clusterName.Empty() {
```
sttts, line 110: am curious when this could happen. Wouldn't the same logic below keep working without this special casing?

pkg/reconciler/kubequota/kubequota_controller.go
```diff
@@ -0,0 +1,366 @@
...
+	// 		DeleteFunc: func(obj interface{}) {
+	// 			h := c.getResourceQuotaEventHandler(obj)
+	// 			if h == nil {
+	// 				return
+	// 			}
+	// 			h.OnDelete(obj)
+	// 		},
+	// 	},
+	// 	quotaRecalculationPeriod,
+	// )
```
sttts, line 161: debugging code?

pkg/informer/informer.go
sttts: This also support my logic: the go doc here says nothing about dropping event. I as subscriber would be highly surprised if I lose these events just because I am a little too slow. This asks for races that are super hard to debug. All this wouldn't be the case with a simple func based callback and local deduplication/dropping logic in the consumer side.

pkg/server/controllers.go
```diff
@@ -888,6 +889,65 @@ func (s *Server) installVirtualWorkspaceURLsController(ctx context.Context, conf
...
+
+	kubeClusterClient, err := kubernetes.NewClusterForConfig(config)
+	if err != nil {
+		return err
+	}
+
+	// TODO(ncdc): should we make these configurable?
+	const (
+		quotaResyncPeriod        = 5 * time.Minute
+		replenishmentPeriod      = 12 * time.Hour
```
sttts, line 908: am curious, does this means that every 12h we might have a storm? Could one smear that over time for different workspaces?

pkg/server/server.go
```diff
@@ -104,6 +106,9 @@ type ExtraConfig struct {
 	// misc
 	preHandlerChainMux *handlerChainMuxes
 
+	// lifecycle
```
sttts, line 109: nit: wouldn't mind to call this misc :)

pkg/admission/kubequota/kubequota_clusterworkspace_monitor.go
ncdc: Keeping as a "controller". It works 😄. Happy to iterate in the future if we decide to make any changes.

pkg/admission/kubequota/kubequota_clusterworkspace_monitor.go
ncdc: Maybe later, but it's highly dependent on the format of the key. I eventually want to "sync" all of this key-related stuff, but we're not there yet.

pkg/indexes/indexes.go
ncdc: Yes, but ... we say indexes a ton when programming 😄

pkg/reconciler/kubequota/kubequota_controller.go
ncdc: Got rid of this.

pkg/reconciler/kubequota/kubequota_controller.go
ncdc: Resolving this thread as it is essentially a dupe of the previous one (which I'll keep open until I address the 🛑)

pkg/informer/informer.go
ncdc: What do you mean by part of the signature spec?

pkg/informer/informer.go
ncdc: Will change the name, thanks

pkg/informer/informer.go
ncdc: How about if we only send notifications for CRDs that are Established & for deletes?

pkg/informer/informer.go
```diff
@@ -565,18 +565,28 @@ func (d *DynamicDiscoverySharedInformerFactory) calculateInformersLockHeld(lates
 	return toAdd, toRemove
 }
 
-// Subscribe registers ch for informer/discovery change notifications.
-func (d *DynamicDiscoverySharedInformerFactory) Subscribe(id string, ch chan<- struct{}) {
+// Subscribe registers ch for informer/discovery change notifications, returning a buffered channel of size 1. When
+// this factory notifies subscribers, it will drop a notification if it's unable to send to a channel. In this case,
+// it means the channel is full, and the subscriber has not yet read the pending notification.
+func (d *DynamicDiscoverySharedInformerFactory) Subscribe(id string) <-chan struct{} {
```
sttts, line 571: :+1:

pkg/reconciler/kubequota/kubequota_controller.go
ncdc: No, was stuff I didn't want to lose in case some refactoring didn't work. Will be removing.

pkg/reconciler/kubequota/delegating_event_handler.go
ncdc: Ideally never, but probably was an overly aggressive optimization to avoid needing to grab the lock. Will simplify if you'd prefer?

pkg/server/controllers.go
ncdc: It would be based on when the workspace was "seen" by the controller. So for a long-running kcp, it would be spread based on when workspaces were created. But at startup, presumably it would see all the workspaces in a short time span, and would probably be more stormy. Follow-up issue ok?

pkg/server/server.go
ncdc: fixed

PR #1240, opened 2022-06-09
turn the watch cache on

p0lyn0mial:
## Summary
requires https://github.com/kcp-dev/kubernetes/pull/73

## Related issue(s)

Fixes #


test/e2e/watchcache/watchcache_enabled_test.go
```diff
@@ -0,0 +1,101 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kubernetesclientset "k8s.io/client-go/kubernetes"
+
+	"github.com/kcp-dev/kcp/test/e2e/fixtures/wildwest"
+	wildwestv1alpha1 "github.com/kcp-dev/kcp/test/e2e/fixtures/wildwest/apis/wildwest/v1alpha1"
+	wildwestclientset "github.com/kcp-dev/kcp/test/e2e/fixtures/wildwest/client/clientset/versioned"
+	"github.com/kcp-dev/kcp/test/e2e/framework"
+)
+
+func TestWatchCacheEnabled(t *testing.T) {
```
sttts, line 39: use extensive `t.Logf` steps in the shape of a story line.

test/e2e/watchcache/watchcache_enabled_test.go
```diff
@@ -0,0 +1,101 @@
...
+	server := framework.SharedKcpServer(t)
+	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
+	org := framework.NewOrganizationFixture(t, server)
+	cluster := framework.NewWorkspaceFixture(t, server, org)
+	cfg := server.DefaultConfig(t)
+	kubeClusterClient, err := kubernetesclientset.NewClusterForConfig(cfg)
+	require.NoError(t, err)
+
+	// create cowboys.wildwest.dev CRD and an object
```
sttts, line 50: can we have the same for APIBindings and native resources?

test/e2e/watchcache/watchcache_enabled_test.go
```diff
@@ -0,0 +1,101 @@
...
+	rsp := kubeClusterClient.RESTClient().Get().AbsPath("/metrics").Do(ctx)
+	raw, err := rsp.Raw()
+	require.NoError(t, err)
+	scanner := bufio.NewScanner(strings.NewReader(string(raw)))
+	var totalCacheHits, cowboysCacheHit int
+	for scanner.Scan() {
+		txt := scanner.Text()
+		if strings.Contains(txt, "apiserver_cache_list_total") {
+			totalCacheHits++
+			if strings.Contains(txt, `resource_prefix="/wildwest.dev/cowboys/customresources`) {
```
sttts, line 84: oh, this means we need work in metrics. It does not distinguish between workspaces and not between CRDs and APIBindings, etc.

sttts: Also note that this test is run in parallel with other e2e also using the cowboys. Can we use a custom API group here, e.g. with a uuid as group name?

p0lyn0mial: I think we should be able to distinguish by 
```
resource_prefix="/group/resource/idenity"}
resource_prefix="/group/resource/customresources"}
```

we could pick up a different `gr`

test/e2e/watchcache/watchcache_enabled_test.go
p0lyn0mial: yes, for native we could check ```/core/secrets```

test/e2e/watchcache/watchcache_enabled_test.go
p0lyn0mial: does it make sense?

sttts: actually, are ok if some other controller increases the metrics? Then we shouldn't care about unique GRs.

test/e2e/watchcache/watchcache_enabled_test.go
sttts: still open.

PR #1246, opened 2022-06-09
test: make framework.Eventually more idiomatic

stevekuznetsov:
As Go developers we know how to identify and pass around errors. Just do
that.

/assign @sttts 
/cc @ncdc


sttts:
Would call this not an improvement. `framework.EventuallyNoError` would be. Using errors for non-errors is wrong.

sttts:
/hold

seeing so many 3 liner wrappers added. This is the wrong direction.

sttts:
Maybe rebase? Retest does not help.

Issue #1187, opened 2022-06-01
Flake: TestSyncerVirtualWorkspace/access_kcp_resources_through_syncer_virtual_workspace


davidfestal:
https://github.com/kcp-dev/kcp/runs/6694296978?check_suite_focus=true

```
--- FAIL: TestSyncerVirtualWorkspace (10.74s)
    util.go:114: Saving test artifacts for test "TestSyncerVirtualWorkspace" under "/tmp/e2e/TestSyncerVirtualWorkspace/2586416673".
    fixture.go:156: Starting kcp servers...
    kcp.go:216: running: /home/runner/work/kcp/kcp/bin/kcp start --root-directory /tmp/e2e/TestSyncerVirtualWorkspace/2586416673/data/kcp/shared --secure-port=44419 --embedded-etcd-client-port=43445 --embedded-etcd-peer-port=40229 --embedded-etcd-wal-size-bytes=5000 --kubeconfig-path=admin.kubeconfig --discovery-poll-interval=5s --token-auth-file /tmp/e2e/TestSyncerVirtualWorkspace/2586416673/data/auth-tokens.csv
    kcp.go:469: success contacting https://10.1.0.249:44419/readyz
    kcp.go:469: success contacting https://10.1.0.249:44419/livez
    fixture.go:183: Started kcp servers after 10.343907826s
    fixture.go:246: Created organization workspace root:e2e-org-m566l
    --- FAIL: TestSyncerVirtualWorkspace/access_kcp_resources_through_syncer_virtual_workspace (59.91s)
        fixture.go:309: Created Universal workspace root:e2e-org-m566l:e2e-workspace-vjgcp
        virtualworkspace_test.go:468: Deploying syncer into workspace root:e2e-org-m566l:e2e-workspace-vjgcp
        util.go:114: Saving test artifacts for test "TestSyncerVirtualWorkspace/access_kcp_resources_through_syncer_virtual_workspace" under "/tmp/e2e/TestSyncerVirtualWorkspace/access_kcp_resources_through_syncer_virtual_workspace/4181382460".
        fixture.go:359: Configuring workspace root:e2e-org-m566l:e2e-workspace-vjgcp for syncing
[...]

PR #1247, opened 2022-06-10
e2e/virtual/syncer: fixing race with resource controller in TestSyncerVirtualWorkspace

sttts:
Fixes https://github.com/kcp-dev/kcp/issues/1187.

### DIFF
--- a/test/e2e/virtual/syncer/virtualworkspace_test.go
+++ b/test/e2e/virtual/syncer/virtualworkspace_test.go
@@ -42,6 +42,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
+	"sigs.k8s.io/yaml"
 
 	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
 	"github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/util/conditions"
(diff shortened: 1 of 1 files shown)


test/e2e/virtual/syncer/virtualworkspace_test.go
```diff
@@ -380,14 +381,15 @@ func TestSyncerVirtualWorkspace(t *testing.T) {
...
-				for _, cs := range virtualWorkspaceCowboys.Items {
-					if logicalcluster.From(&cs) == wildwestClusterName {
-						require.Equal(t, "luckyluke", cs.Name)
-					}
-				}
+				t.Log("Wait for resource controller to schedule cowboy and then show up via virtual workspace wildcard request")
+				var cowboys *wildwestv1alpha1.CowboyList
+				require.Eventually(t, func() bool {
+					cowboys, err = vwClusterClient.Cluster(logicalcluster.Wildcard).WildwestV1alpha1().Cowboys("").List(ctx, metav1.ListOptions{})
+					require.NoError(t, err)
```
ncdc, line 388: Can you do a require inside an Eventually without it failing?

sttts: we have that everywhere. If it does not work, we should generally fix it.

ncdc: Compare
```go
func TestAndy(t *testing.T) {
	i := 0

	require.Eventually(t, func() bool {
		i++
		require.True(t, i%2 == 0)
		return true
	}, time.Second, 100*time.Millisecond)
}
```

with

```go
func TestAndy(t *testing.T) {
	i := 0

	require.Eventually(t, func() bool {
		i++
		return i%2 == 0
	}, time.Second, 100*time.Millisecond)
}
```

Both should eventually succeed, right? But only the 2nd one does.

sttts: Why should the first succeed? It should not.

PR #1252, opened 2022-06-10
ci: delete demos.yaml

ncdc:
## Summary
Demos are no longer maintained

## Related issue(s)

Fixes #


ncdc:
Tide?

ncdc:
Not sure what's up with tide. Manually merging as this is only removing an optional job we weren't using any more

PR #1258, opened 2022-06-13
Placement API

qiujian16:
add API definition
add API to boostrap under location feature gate

## Related issue(s)

Fixes #


pkg/apis/scheduling/v1alpha1/types_placement.go
```diff
@@ -0,0 +1,128 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	conditionsv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/apis/conditions/v1alpha1"
+)
+
+// placement defines a selection rule to choose ONE location for MULTIPLE namespaces in a workspace.
+//
+// placement is in Pending state initially, when a location is selected by the placement, the placement
```
sttts, line 27:
```suggestion
// placement is in Pending state initially. When a location is selected by the placement, the placement
```

pkg/apis/scheduling/v1alpha1/types_placement.go
```diff
@@ -0,0 +1,128 @@
...
+// turns to UnBound state. In Pending or UnBound state, the selection rule can be updated to select another location.
```
sttts, line 28:
```suggestion
// turns to Unbound state. In Pending or Unbound state, the selection rule can be updated to select another location.
```

pkg/apis/scheduling/v1alpha1/types_placement.go
```diff
@@ -0,0 +1,128 @@
...
+// When the user annotates a namespace with the key of "scheudling.kcp.dev/placement", the namespace will
+// pick one placement, and this placement is transfered to Bound state. The spec of the placement is immutable
+// in Bound state. The placement will turns back to Unbound state when no namespace uses this
```
sttts, line 31: have a little problems with marking it read-only. If this object comes from GitOps, that tool will see errors. We should rather make it mutable, but use a condition that says that the status is not up-to-datae.

pkg/apis/scheduling/v1alpha1/types_placement.go
```diff
@@ -0,0 +1,128 @@
...
+
+	// namespaceSelector is a label selector to select ns. It match all ns by default, but can be specified to
+	// a certain set of ns.
+	NamespaceSelector metav1.LabelSelector `json:"namespaceSelector,omitempty"`
+
+	// locationWorkspace is an absolute reference to a workspace for the location. If it is not set, the workspace of
+	// APIBinding will be used.
+	// +optional
+	// +kubebuilder:validation:Pattern:="^root(:[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$"
+	LocationWorkspace *string `json:"locationWorkspace,omitempty"`
```
sttts, line 70: do we have any requirements here? In other context, we require this to be a sibling, an ancestor or sibling of ancestor.

sttts: Reason: such a rule forces publicly consumed workspaces to be as high in the hierarchy as required by its consumers.

pkg/apis/scheduling/v1alpha1/types_placement.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+type PlacementStatus struct {
+	// phase is the current phase of the placement
+	//
+	// +required
+	// +kubebuilder:Required
+	// +kubebuilder:validation:Enum="";Pending;Bound;UnBound
```
sttts, line 78: default: Pending

pkg/apis/scheduling/v1alpha1/types_placement.go
sttts: oh, we actually have that:
> The workspace must be some ancestor or a child of some ancestor.
:+1:

pkg/apis/scheduling/v1alpha1/types_placement.go
```diff
@@ -0,0 +1,128 @@
...
+	Conditions conditionsv1alpha1.Conditions `json:"conditions,omitempty"`
+}
+
+// LocationReference describes a loaction that are provided in the specified Workspace.
+type LocationReference struct {
+	// path is an absolute reference to a workspace, e.g. root:org:ws. The workspace must
+	// be some ancestor or a child of some ancestor.
+	// +optional
+	// +kubebuilder:validation:Pattern:="^root(:[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$"
+	Path string `json:"path,omitempty"`
```
sttts, line 95: is this actually required? Same for APIBinding of course. Or do we default it to be the local workspace? Worth to check in code. Defaulting seems to be nice UX.

pkg/apis/scheduling/v1alpha1/types_placement.go
```diff
@@ -0,0 +1,128 @@
...
+
+type PlacementPhase string
+
+const (
+	// PlacementPending is the phase that the location has not been selected for this placement.
+	PlacementPending = "Pending"
+
+	// PlacementUnBound is the phase that the location has been selected by the placement, but
+	// no namespace is bound to this placement yet.
+	PlacementUnBound = "UnBound"
```
sttts, line 113:
```suggestion
	PlacementUnBound = "Unbound"
```

pkg/apis/scheduling/v1alpha1/types_placement.go
sttts: Went ahead and implemented it https://github.com/kcp-dev/kcp/pull/1262.

pkg/apis/scheduling/v1alpha1/types_placement.go
sttts: nit: omitempty string does not have to be a pointer to be optional.

pkg/apis/scheduling/v1alpha1/types_placement.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// locationResource is the group-version-resource of the instances that are subject to the locations to select.
+	//
+	// +required
+	// +kubebuilder:Required
+	LocationResource GroupVersionResource `json:"locationResource"`
+
+	// namespaceSelector is a label selector to select ns. It match all ns by default, but can be specified to
+	// a certain set of ns.
+	NamespaceSelector metav1.LabelSelector `json:"namespaceSelector,omitempty"`
```
sttts, line 64: omitempty is probably not effective here because LabelSelector is a struct. Must be a pointer for that.

sttts:
Few nits left, but this looks good.

We have two choices to move forward:
1. put this under another feature gate (LocationAPI is on by default), and merge early
2. wait for the implementing before merging the API.

pkg/apis/scheduling/v1alpha1/types_placement.go
sttts: then it must be a pointer

sttts: From the docs:
> An empty label selector matches all objects. A null label selector matches no objects.

So yes, must be a pointer with omitempty.

pkg/apis/scheduling/v1alpha1/types_placement.go
```diff
@@ -24,12 +24,12 @@ import (
...
-// turns to UnBound state. In Pending or UnBound state, the selection rule can be updated to select another location.
-// When the user annotates a namespace with the key of "scheudling.kcp.dev/placement", the namespace will
-// pick one placement, and this placement is transfered to Bound state. The spec of the placement is immutable
-// in Bound state. The placement will turns back to Unbound state when no namespace uses this
-// placement any more.
+// placement is in Pending state initially. When a location is selected by the placement, the placement
+// turns to Unbound state. In Pending or Unbound state, the selection rule can be updated to select another location.
+// When the a namespace is annotated by another controller or user with the key of "scheudling.kcp.dev/placement",
+// the namespace will pick one placement, and this placement is transfered to Bound state. Any update to spec of the placement
+// is ignored in Bound state and reflected in the conditions. The placement will turns back to Unbound state when no namespace
```
sttts, line 31: turn back

Issue #1149, opened 2022-05-25
Epic: ClusterWorkspaceType Take 2


sttts:
# Action Items

- [ ] add virtual apiserver to access ClusterWorkspaces that are initializing https://github.com/kcp-dev/kcp/pull/1141
- [ ] allow workspace content access through the virtual apiserver
- [ ] add authz such that we can allow a service account in the workspace of the ClusterWorkspaceType to access the workspaces that are initializing
- [ ] search for ClusterWorkspaceTypes not only in the current workspace, but all ancestors (= update the current cluster workspace creation admission)
- [ ] rethink how a ClusterWorkspaceType defined in some ancestor can be extended with further initializers further down in the hierarchy.
- [ ] rethink how we can ensure that lower-levels of the hierarchies cannot change in which workspace types an existing type can be created (e.g. we don't want Organization workspaces in Universal. This can break invariants expected by the type owner).

sttts:
Informers are by shard. Every VW needs a way to communicate shard-specific URLs. The controllers authors have to make sure that some instance serves each of those URLs.

sttts:
Having said that, we have an open design issue that we hit already in a couple of places: the parent workspace is not necessarily in the same shard or region as the workspace itself. This means that controller responsible for a workspace need access to the defining ClusterWorkspace in the parent. We likely need a replica of that object in the shard the actual workspace is defined. Otherwise there is a strange skew that risks availability.

In other words: we want what `.` is in unix file-system.

stevekuznetsov:
@sttts re: 

>  fully qualify `allowedChildWorkspaceTypes`,`allowedParentWorkspaceTypes` and `defaultChildWorkspaceType`, while adding admission checks to validate their values

You mentioned admission checks... what is the policy we want there?

stevekuznetsov:
OK, re: fully qualifying the types - we want to do this, and when a type is created these references must be checked after type aliases are resolved, in order to ensure that:

- `root:org:team:MyApp` cannot be a child of `root:Organization` unless `root:Organization` allows `Any` child, or `root:org:team:MyApp` is an alias for a child type that `root:Organization` does allow
- `root:Organization` cannot be a child of `root:org:team:MyApp` unless `root:Organization` allows `Any` parent, or `root:org:team:MyApp` is an alias for a parent type that `root:Organization` does allow

The CWT controller needs to resolve this and post a condition. CW created with the CWT are denied until this condition is true.

sttts:
[...]

PR #1264, opened 2022-06-13
clusterworkspaces: make initializers implicit

stevekuznetsov:
Creating a one-to-one link between a ClusterWorkspaceType and the
initializer that provides it with the functionality it confers allows us
to reason about the system more simply. The composition approach to
types we are working towards will let users leverage this one-to-one
relationship when building composite types as well.

Part of #1149 
/assign @sttts 
/cc @ncdc


sttts:
Nothing to add. Just done well.

/lgtm

PR #1287, opened 2022-06-20
cmd/kcp: non-zero return code on error

sttts:
We always exited with zero.

### DIFF
--- a/cmd/kcp/kcp.go
+++ b/cmd/kcp/kcp.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/errors"
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	"k8s.io/client-go/rest"
+	"k8s.io/component-base/cli"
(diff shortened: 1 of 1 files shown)


s-urbaniak:
good catch 👍

cmd/kcp/kcp.go
```diff
@@ -147,5 +147,6 @@ func main() {
 
 	if err := cmd.Execute(); err != nil {
 		fmt.Fprintf(os.Stderr, "error: %v\n", err)
+		os.Exit(1)
```
s-urbaniak, line 150: Could `k8s.io/component-base/cli#Run` be of help here?

s-urbaniak (changes requested):
actually, yes, I believe `os.Exit(cli.Run(cmd))` will be much more concise

cmd/kcp/kcp.go
sttts: fixed

sttts:
Unrelated flake.

PR #1291, opened 2022-06-20
reconciler/apiexport: don't klog.Fatal

sttts:
Fixed old code that exits the process.

### DIFF
--- a/pkg/reconciler/apis/apiexport/apiexport_controller.go
+++ b/pkg/reconciler/apis/apiexport/apiexport_controller.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	jsonpatch "github.com/evanphx/json-patch"
+	"github.com/google/go-cmp/cmp"
 	"github.com/kcp-dev/logicalcluster"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


s-urbaniak (approved):
/lgtm

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -322,10 +323,6 @@ func (c *controller) patchIfNeeded(ctx context.Context, old, obj *apisv1alpha1.A
 	specOrObjectMetaChanged := !equality.Semantic.DeepEqual(old.Spec, obj.Spec) || !equality.Semantic.DeepEqual(old.ObjectMeta, obj.ObjectMeta)
 	statusChanged := !equality.Semantic.DeepEqual(old.Status, obj.Status)
 
-	if specOrObjectMetaChanged && statusChanged {
```
s-urbaniak, line 325: Let's put as a comment on the inner `apiExportForPatch` declaration that the logic will guarantee that only the objectmeta/spec change will be patched first and on the next iteration the spec change.

s-urbaniak (changes requested):
Let's make it a bit more clear to the future reader in comments how the logic about reconcilation works (in case of programming error objectmeta/spec change will be patched first, in the next iteration the status change).

PR #1294, opened 2022-06-20
server/apiextensions: use indexes and fixes

sttts:
- fix meta data lookup to only be for wildcard requests
- fix cache mutation
- use indexes.

### DIFF
--- a/pkg/server/apiextensions.go
+++ b/pkg/server/apiextensions.go
@@ -321,7 +321,7 @@ func (c *apiBindingAwareCRDLister) Refresh(crd *apiextensionsv1.CustomResourceDe
 	}
 
 	// Start with a shallow copy
-	refreshed := shallowCopyCRD(updatedCRD)
+	refreshed := shallowCopyCRDAndDeepCopyAnnotations(updatedCRD)
 
 	// If crd has the identity annotation, make sure it's added to refreshed
 	if identity := crd.Annotations[apisv1alpha1.AnnotationAPIIdentityKey]; identity != "" {
@@ -330,7 +330,7 @@ func (c *apiBindingAwareCRDLister) Refresh(crd *apiextensionsv1.CustomResourceDe
 
 	// If crd was only partial metadata, make sure refreshed is too
 	if _, partialMetadata := crd.Annotations[annotationKeyPartialMetadata]; partialMetadata {
-		partialMetadataCRD(refreshed)
+		makePartialMetadataCRD(refreshed)
 
 		if strings.HasSuffix(string(crd.UID), ".wildcard.partial-metadata") {
 			refreshed.UID = crd.UID
@@ -384,8 +384,8 @@ func (c *apiBindingAwareCRDLister) Get(ctx context.Context, name string) (*apiex
 	}
 
 	if partialMetadataRequest {
-		crd = shallowCopyCRD(crd)
-		partialMetadataCRD(crd)
+		crd = shallowCopyCRDAndDeepCopyAnnotations(crd)
+		makePartialMetadataCRD(crd)
 
 		if clusterName == logicalcluster.Wildcard {
 			crd.UID = types.UID(name + ".wildcard.partial-metadata")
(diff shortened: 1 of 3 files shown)


pkg/server/apiextensions.go
```diff
@@ -235,16 +227,12 @@ func (c *apiBindingAwareCRDLister) List(ctx context.Context, selector labels.Sel
 		seen.Insert(crdName(kcpSystemCRDs[i]))
 	}
 
-	apiBindings, err := c.apiBindingLister.List(labels.Everything())
+	objs, err := c.apiBindingIndexer.ByIndex(byWorkspace, clusterName.String())
```
p0lyn0mial, line 230: okay, so it replaces `logicalcluster.From(apiBinding) != clusterName`

pkg/server/apiextensions.go
```diff
@@ -257,7 +245,7 @@ func (c *apiBindingAwareCRDLister) List(ctx context.Context, selector labels.Sel
 				continue
 			}
 
-			if !matchesSelector(crd) {
+			if !selector.Matches(labels.Set(crd.Labels)) {
```
p0lyn0mial, line 248: we don't have to check if the selector is empty anymore?

pkg/server/apiextensions.go
```diff
@@ -280,13 +268,14 @@ func (c *apiBindingAwareCRDLister) List(ctx context.Context, selector labels.Sel
...
 	if err != nil {
 		return nil, err
 	}
-	for i := range crds {
-		crd := crds[i]
-		if logicalcluster.From(crd) != clusterName {
+	for _, obj := range objs {
+		crd := obj.(*apiextensionsv1.CustomResourceDefinition)
+
+		if !selector.Matches(labels.Set(crd.Labels)) {
```
p0lyn0mial, line 278: okay, ic, the same labels needs to be applied here.

pkg/server/apiextensions.go
```diff
@@ -375,18 +364,19 @@ func (c *apiBindingAwareCRDLister) Get(ctx context.Context, name string) (*apiex
 		// Not a system CRD, so check in priority order: identity, wildcard, "normal" single cluster
 
 		identity := IdentityFromContext(ctx)
-
-		switch {
-		case identity != "":
+		if identity != "" {
```
p0lyn0mial, line 367: this `if, else` statement looks okay to me

pkg/server/apiextensions.go
```diff
@@ -446,47 +434,46 @@ func decorateCRDWithBinding(in *apiextensionsv1.CustomResourceDefinition, identi
...
 // metadata.
-func partialMetadataCRD(crd *apiextensionsv1.CustomResourceDefinition) {
+func makePartialMetadataCRD(crd *apiextensionsv1.CustomResourceDefinition) {
 	crd.Annotations[annotationKeyPartialMetadata] = ""
 
 	// set minimal schema that prunes everything but ObjectMeta
-	for _, v := range crd.Spec.Versions {
-		v.Schema = &apiextensionsv1.CustomResourceValidation{
+	old := crd.Spec.Versions
+	crd.Spec.Versions = make([]apiextensionsv1.CustomResourceDefinitionVersion, len(old))
```
p0lyn0mial, line 444: since this is a slice, I think it should be `make([]apiextensionsv1.CustomResourceDefinitionVersion, 0, len(old))`

p0lyn0mial: ah, nvm, you index down below

pkg/server/apiextensions.go
```diff
@@ -332,7 +321,7 @@ func (c *apiBindingAwareCRDLister) Refresh(crd *apiextensionsv1.CustomResourceDe
 	}
 
 	// Start with a shallow copy
-	refreshed := shallowCopyCRD(updatedCRD)
+	refreshed := shallowCopyCRDAndDeepCopyAnnotations(updatedCRD)
```
p0lyn0mial, line 324: w00t? why are we modifying a live object? can we ensure no one else is actually using the obj?

p0lyn0mial: ic, we acquire a lock

pkg/server/apiextensions.go
p0lyn0mial: ic

PR #1299, opened 2022-06-20
APIBinding/Workspace deletion improvement

sttts:
- observability for the user: add initializer and subset of conditions to v1beta1 Workspace
- deletion takes long. We need more reconcilers: run 10 of them
- set finalizers via admission – this might fix lots of flakes. The finalizers are so important for consistency. We better enforce them via admission rather than relying on controllers and queue backup to be not too big.
- improve logs: show reasons
- remove cache syncing (done by server already}
- reconciler/clusterworkspacedeletion: fix hotlooping due to WorkspaceContentDeleted condition


pkg/reconciler/tenancy/clusterworkspacedeletion/deletion/workspace_resource_deletor.go
```diff
@@ -463,16 +454,18 @@ func (d *workspacedResourcesDeleter) deleteAllContent(ctx context.Context, ws *t
...
+
+	if len(contentDeletedMessages) > 0 {
 		conditions.MarkFalse(
 			ws,
 			tenancyv1alpha1.WorkspaceContentDeleted,
-			"SomeFinalizersRemain",
+			"SomeResourcesRemain",
 			conditionsv1alpha1.ConditionSeverityError,
-			fmt.Sprintf("Some content in the workspace has finalizers remaining: %s", strings.Join(remainingByFinalizer, ", ")),
+			strings.Join(contentDeletedMessages, "; "),
```
qiujian16: is this to ensure the condition message stable?

sttts: per iteration one must never call `conditions.MarkFalse/True` twice because it will hotloop. Hence, I join both messages here and do one call only.

sttts: no, it's by design. It just compares reason+message and update the timestamp if one changes. But of course, if you call the func multiple times with different args (even if it converges to the same values every time) it will think there are updates.

sttts:
syncer flake.

/retest

Issue #1280, opened 2022-06-16
 Rework downstream namespace names


jmprusi:
# Introduction

We need to improve how the current downstream namespaces names are generated:

- Shorter namespace names, focus on not going over 32 chars (more chars causes issues with generated ingresses/routes as it hits 63 char limits)
- Have the workspace as part of the namespace name.
- Consistent/reproducible. 
- Allow the same k8s cluster to be added as multiple workloadClusters and multiple syncers.
- Avoid possible clashes with other namespaces.

Kubernetes Namespaces need to follow those requirements:
- Namespaces follow the [RFC 1123](https://tools.ietf.org/html/rfc1123) 
  - contain at most 63 characters
  - contain only lowercase alphanumeric characters or '-'
  - start with an alphanumeric character
  - end with an alphanumeric character

### The namespaceLocator struct

Currently the kcp syncer serialises in json some information on an annotation "" in the downstream namespace that contains:

- upstreamNamespace
- workspace

the changes introduced in https://github.com/kcp-dev/kcp/pull/1309 expand the annotation to include:

- upstreamNamespace
- workspace
- SyncTarget workspace
- SyncTarget Name
- SyncTarget UID

This information is used by the Spec Syncer and the Status Syncer to figure what namespace to use.

### The downstream Namespaces:

Previously the downstream namespaces were 59 characters long:

`kcp0124d7647eb6a00b1fcb6f2252201601634989dd79deb7375c373973`

After the PR:

`kcp-2r7hmup1y2r1`

This is done by hashing the namespaceLocator with `sha256.Sum224()`, convert the result to base36 and take 12 characters of the resulting string. 

This should unblock/fix some issues with external controllers adding the namespace to routes/ingresses.

### Spec Syncer

The Spec syncer is in charge of syncing down the KCP resources and creating if needed the downstream namespaces.

Before the PR https://github.com/kcp-dev/kcp/pull/1309 the syncer tried to create the downstream namespace by hashing the desired namespace locator, even if it existed before.

After the changes, the Spec Syncer tries to figure out if a downstream namespaces matches the desired namespacelocator (using the downstream informers). If that's the case it uses that namespace. If not, hashes the desired namespaceLocator and creates the downstream namespace.

The PR adds some basic conflict (namespace exists with different namespacelocator, etc)

### Status Syncer

[...]

PR #1309, opened 2022-06-20
Reworking namespaces

jmprusi:
This PR changes the pattern used to generate the downstream namespace. It generates a shorter string while maintaining a big enough random part to avoid conflicts.

related: https://github.com/kcp-dev/kcp/issues/1280


pkg/syncer/shared/namespace_test.go
```diff
@@ -0,0 +1,92 @@
...
+			// Make sure we don't produce namespace that are too long,
+			// as other components use the namespace as part of their
+			// generated resources (e.g. Ingresses, Routes).
+			// See issue: https://github.com/kcp-dev/kcp/issues/1188
+			if len(got) > 16 {
+				t.Errorf("namespace name is too long: %s", got)
+			}
+			// Make sure we have no collisions.
+			for _, ns := range previousNamespaces {
+				require.NotEqual(t, got, ns)
```
sttts, line 87: not a really a helpful test, is it?

jmprusi: not really.. I can drop that "previousNamespaces" check.

pkg/syncer/shared/namespace.go
```diff
@@ -49,11 +51,18 @@ func LocatorFromAnnotations(annotations map[string]string) (*NamespaceLocator, e
 
 // PhysicalClusterNamespaceName encodes the NamespaceLocator to a new
 // namespace name for use on a physical cluster. The encoding is repeatable.
-func PhysicalClusterNamespaceName(l NamespaceLocator) (string, error) {
+func PhysicalClusterNamespaceName(l NamespaceLocator, workloadClusterName string) (string, error) {
```
sttts, line 54: is this WorkloadCluster name or workload ClusterName?

jmprusi: WorkloadCluster Name

sttts: would drop the "Name" suffix to make it clear.

Note that the name of the workload cluster is not unique in kcp. We need more here.

jmprusi:
> How do we know that
> 
> 1. the syncer controllers only map once, and then use the annotation afterwards? Has this at least been verified manually?

* The Spec Syncer maps every time (so uses the physicalNamespace each time on process())
* The Status Syncer relies on the namespaceLocation annotation

> 2. what is the behaviour if two namespace collide? Where is it handled?

Not handling this right now. what do you propose?

pkg/syncer/shared/namespace.go
jmprusi: I was under the impression that NamespaceLocator(Workspace + Namespace) + WorkloadCluster should be a unique identifier... there's some case I'm missing?

sttts:
> The Spec Syncer maps every time (so uses the physicalNamespace each time on process())
> (This will likely leave dangling namespaces right now...)

Why not use a downstream indexer to find the namespace?

pkg/syncer/shared/namespace.go
sttts: yes and no. The WorkloadCluster lives in a different location workspace. Per `NamespaceLocator(Workspace, ..)` there is only one such location workspace, but on a pcluster this does not help as multiple syncer can point to different location workspaces with the same WorkloadCluster names.

pkg/syncer/shared/namespace.go
```diff
@@ -20,40 +20,75 @@ import (
...
 type NamespaceLocator struct {
-	LogicalCluster logicalcluster.Name `json:"logical-cluster"`
-	Namespace      string              `json:"namespace"`
+	Workspace       logicalcluster.Name `json:"logical-cluster"`
+	SyncerWorkspace logicalcluster.Name `json:"syncer-workspace"`
+	Namespace       string              `json:"namespace"`
 }
 
 func LocatorFromAnnotations(annotations map[string]string) (*NamespaceLocator, error) {
-	annotation := annotations[NamespaceLocatorAnnotation]
```
stevekuznetsov, line 39: What's the benefit of moving this out of one annotation?

pkg/syncer/shared/namespace.go
```diff
@@ -20,40 +20,75 @@ import (
...
-	if len(annotation) == 0 {
+	if len(annotations) == 0 {
 		return nil, nil
 	}
+
 	var locator NamespaceLocator
-	if err := json.Unmarshal([]byte(annotation), &locator); err != nil {
-		return nil, err
+	if val, ok := annotations[NamespaceAnnotation]; ok && val != "" {
```
stevekuznetsov, line 49:
```go

for key, into := range map[string]*string{
    NamespaceAnnotation: &locator.Namespace,
    WorkspaceAnnotation: &locator.Workspace,
// ...
} {
    if val, ok := annotations[key]; ok && val != "" {
        *into = val
    } else {
        return nil, fmt.Errorf("whatever")
    }
}

File: pkg/syncer/shared/namespace.go [hunk 5]
```diff
@@ -20,40 +20,75 @@ import (
...
 // PhysicalClusterNamespaceName encodes the NamespaceLocator to a new
 // namespace name for use on a physical cluster. The encoding is repeatable.
 func PhysicalClusterNamespaceName(l NamespaceLocator) (string, error) {
 	b, err := json.Marshal(l)
 	if err != nil {
 		return "", err
 	}
-	hash := sha256.Sum224(b)
-	return fmt.Sprintf("kcp%x", hash), nil
+	// Append the logicalCluster marshaled bytes with the workloadClusterName to get a hashed value
```
stevekuznetsov, line 85: What does this comment mean ... ? The hash is not of the logical cluster, and there is no appending happening anywhere

stevekuznetsov (commented):
JSON in one annotation makes it simpler to store & load data. Not sure adding N annotations per field (what happens when they get complex?) is a good idea.

pkg/syncer/shared/namespace.go
jmprusi: This was a leftover comment. fixed.

pkg/syncer/shared/namespace.go
jmprusi: I've kept the annotation for now, after playing with the idea, I don't see too much value.

pkg/syncer/shared/namespace.go
```diff
@@ -31,14 +35,25 @@ const (
 // NamespaceLocator stores a logical cluster and namespace and is used
 // as the source for the mapped namespace name in a physical cluster.
 type NamespaceLocator struct {
-	LogicalCluster logicalcluster.Name `json:"logical-cluster"`
-	Namespace      string              `json:"namespace"`
+	Workspace          logicalcluster.Name `json:"workspace,omitempty"`
+	WorkloadCluster    logicalcluster.Name `json:"workloadCluster,omitempty"`
+	WorkloadClusterUID types.UID           `json:"workloadClusterUID,omitempty"`
```
sttts, line 40: WorkloadCluster will become SyncTarget in some days.

pkg/syncer/shared/namespace.go
```diff
@@ -31,14 +35,25 @@ const (
...
+	}
 }
 
 func LocatorFromAnnotations(annotations map[string]string) (*NamespaceLocator, error) {
-	annotation := annotations[NamespaceLocatorAnnotation]
-	if len(annotation) == 0 {
-		return nil, nil
+	annotation, ok := annotations[NamespaceLocatorAnnotation]
+	if !ok || len(annotation) == 0 {
+		return nil, fmt.Errorf("no namespace locator annotation found")
```
sttts, line 56: not a good signature. This is not an error, i.e. some exceptional state, but something normal. I.e. either the old semantics are good, or `return nil, false, nil` vs. `return locator, true, nil`.

pkg/syncer/spec/spec_process.go
```diff
@@ -114,14 +115,30 @@ func (c *Controller) process(ctx context.Context, gvr schema.GroupVersionResourc
...
-	})
-	if err != nil {
-		klog.Errorf("Error hashing namespace %s|%s: %v", clusterName, upstreamNamespace, err)
-		return nil // ignore error, shouldn't happen
+	namespaceGvr := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "namespaces"}
+	desiredNSLocator := shared.NewNamespaceLocator(clusterName, c.workloadClusterLogicalClusterName, c.workloadClusterUID, c.workloadClusterName, upstreamNamespace)
+	downstreamNamespaces := c.downstreamInformers.ForResource(namespaceGvr).Informer().GetIndexer().List()
+
+	var downstreamNamespace string
+	for _, namespace := range downstreamNamespaces {
```
sttts, line 123: now you are using the indexer as a non-index. The index can give you the correct namespace in constant time. See indexer.ByIndex.

pkg/syncer/spec/spec_process.go
```diff
@@ -114,14 +115,30 @@ func (c *Controller) process(ctx context.Context, gvr schema.GroupVersionResourc
...
+		namespace := namespace.(*unstructured.Unstructured)
+		nsLocator, err := shared.LocatorFromAnnotations(namespace.GetAnnotations())
+		if err != nil {
+			continue
+		}
+		if reflect.DeepEqual(desiredNSLocator, *nsLocator) {
```
sttts, line 129: why do this on the locator data type? Why not use the string, i.e. the annotation directly

pkg/syncer/spec/spec_process.go
jmprusi: Yes... the idea is for the Spec Syncer to be able to figure out the namespace without having to calculate the hash, this way we will be able to change the pattern and just rely on the downstream namespaceLocator, but yes, we need to lookup between all the namespaces. 

This could be improved by checking if the namespace from the hashed namespaceLocator exists (physical...), and then just use that one, if not, look for it on the list of namespaces and try to find another namespace with the desired namespaceLocator.

sttts: we cannot look at a list of namespaces. We have to use an index on namespaces here. Not sure we talk about the same.

pkg/syncer/spec/spec_process.go
jmprusi: done

jmprusi:
<!DOCTYPE html>

/test e2e-multiple-runs
--


<br class="Apple-interchange-newline">

pkg/syncer/spec/spec_process.go
```diff
@@ -114,14 +115,33 @@ func (c *Controller) process(ctx context.Context, gvr schema.GroupVersionResourc
...
+	}
+	var downstreamNamespace string
+
+	if len(downstreamNamespaces) == 1 {
+		namespace := downstreamNamespaces[0].(*unstructured.Unstructured)
+		klog.V(4).Infof("Found downstream namespace %s for upstream namespace %s", namespace.GetName(), upstreamNamespace)
+		downstreamNamespace = namespace.GetName()
+	} else if len(downstreamNamespaces) > 1 {
+		// This should never happen unless there's some namespace collision.
+		return fmt.Errorf("(namespace collision) found multiple downstream namespaces for upstream namespace %s", upstreamNamespace)
```
sttts, line 137: return the namespaces in question.

pkg/syncer/spec/spec_process.go
```diff
@@ -357,3 +387,15 @@ func getTransformedName(syncedObject *unstructured.Unstructured) string {
 	}
 	return syncedObject.GetName()
 }
+
+func (c *Controller) getDownstreamNamespace(namespaceName string) (*unstructured.Unstructured, error) {
+	namespaceGVR := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "namespaces"}
+	downstreamNamespace, exists, err := c.downstreamInformers.ForResource(namespaceGVR).Informer().GetIndexer().GetByKey(namespaceName)
```
sttts, line 393: no need to go through the indexer for simple names.

pkg/syncer/spec/spec_process.go
```diff
@@ -181,17 +199,29 @@ func (c *Controller) ensureDownstreamNamespaceExists(ctx context.Context, downst
...
-	if _, err := namespaces.Create(ctx, newNamespace, metav1.CreateOptions{}); err != nil {
-		// An already exists error is ok - it means something else beat us to creating the namespace.
-		if !k8serrors.IsAlreadyExists(err) {
-			// Any other error is not good, though.
-			// TODO bubble this up as a condition somewhere.
-			klog.Errorf("Error while creating namespace %q: %v", downstreamNamespace, err)
+	// Check if the namespace already exists, if not create it.
+	namespace, err := c.getDownstreamNamespace(newNamespace.GetName())
+	if apierrors.IsNotFound(err) {
+		// Not found, so let's create the namespace.
```
sttts, line 205: superfluous comment

pkg/syncer/spec/spec_process.go
```diff
@@ -181,17 +199,29 @@ func (c *Controller) ensureDownstreamNamespaceExists(ctx context.Context, downst
...
+		if _, err := namespaces.Create(ctx, newNamespace, metav1.CreateOptions{}); err != nil {
```
sttts, line 206: handle AlreadyExist. You informer lookup will have latency to be correct.

pkg/syncer/spec/spec_process.go
```diff
@@ -181,17 +199,29 @@ func (c *Controller) ensureDownstreamNamespaceExists(ctx context.Context, downst
...
+		return err
+	}
+
+	// The namespace exists, so check if it has the correct namespace locator.
+	nsLocator, exists, err := shared.LocatorFromAnnotations(namespace.GetAnnotations())
+	if err != nil {
+		return fmt.Errorf("(possible namespace collision) namespace %s already exists, but found an error when trying to decode the annotation: %w", newNamespace.GetName(), err)
+	}
+	if !exists {
+		return fmt.Errorf("(namespace collision) namespace %s has no namespace locator", namespace.GetName())
```
sttts, line 221: this will hotloop, won't it?

pkg/syncer/spec/spec_process.go
```diff
@@ -357,3 +387,15 @@ func getTransformedName(syncedObject *unstructured.Unstructured) string {
 	}
 	return syncedObject.GetName()
 }
+
+func (c *Controller) getDownstreamNamespace(namespaceName string) (*unstructured.Unstructured, error) {
```
sttts, line 391: I don't see what this func add to a direct informer lookup.

pkg/syncer/shared/namespace.go
sttts: a structure like `{"syncTarget": {"path": ...,"name": ...,"uid", ...}, "workspace": ..., "namespace": ...}` would make that read much easier.

sttts:
Some smaller things left, but it looks much better now!

pkg/syncer/spec/spec_process.go
jmprusi: Log error and return nil? There's some way to bubble up the error somewhere visible? as the user feedback will be zero if not...

pkg/syncer/spec/spec_process.go
jmprusi: not sure what to do there, as if it already exists... we should loop again to make sure that the namespaceLocator matches what we want... so instead of repeating the logic there I just loop again.

sttts: no, just assume your informer check few lines above was based on stale informers. The downstream namespace actually exists already.

pkg/syncer/spec/spec_process.go
jmprusi: If we created another namespace with the same nslocator, we will hit: https://github.com/jmprusi/kcp/blob/jmprusi/reworking-namespaces/pkg/syncer/spec/spec_process.go#L13

pkg/syncer/spec/spec_process.go
jmprusi: if we create another namespace with the same nsLocator, we will hit https://github.com/kcp-dev/kcp/pull/1309/files#diff-d0397d2587a97f7f7d490a7e6f83776c15aebd162c8ee61a4ebe19ad161777aaR135

pkg/syncer/spec/spec_process.go
jmprusi: done

pkg/syncer/spec/spec_process.go
jmprusi: done

pkg/syncer/spec/spec_process.go
jmprusi: done

pkg/syncer/spec/spec_controller.go
```diff
@@ -43,6 +44,7 @@ import (
 
 const (
 	controllerName                   = "kcp-workload-syncer-spec"
+	byNamespaceLocatorIndexName      = "syncer-spec-ByNamespaceLocator-index"
```
sttts, line 47: nit: all of these are indexes. No need for the `-index` postfix.

pkg/syncer/spec/spec_process.go
```diff
@@ -114,14 +114,37 @@ func (c *Controller) process(ctx context.Context, gvr schema.GroupVersionResourc
 	}
 	clusterName, name := clusters.SplitClusterAwareKey(clusterAwareName)
 
-	// to downstream
-	downstreamNamespace, err := shared.PhysicalClusterNamespaceName(shared.NamespaceLocator{
-		LogicalCluster: clusterName,
-		Namespace:      upstreamNamespace,
-	})
+	namespaceGvr := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "namespaces"}
```
sttts, line 117: (nit, style) always define variables/constants as near as possible to their use.

pkg/syncer/spec/spec_process.go
```diff
@@ -181,17 +202,33 @@ func (c *Controller) ensureDownstreamNamespaceExists(ctx context.Context, downst
...
-			// Any other error is not good, though.
-			// TODO bubble this up as a condition somewhere.
-			klog.Errorf("Error while creating namespace %q: %v", downstreamNamespace, err)
+	// Check if the namespace already exists, if not create it.
+	namespace, err := c.downstreamInformers.ForResource(schema.GroupVersionResource{Group: "", Version: "v1", Resource: "namespaces"}).Lister().Get(newNamespace.GetName())
+	if err != nil && apierrors.IsNotFound(err) {
+		if _, err := namespaces.Create(ctx, newNamespace, metav1.CreateOptions{}); err != nil {
+			if apierrors.IsAlreadyExists(err) {
+				klog.Infof("Using already existing downstream namespace %s for upstream namespace %s|%s", newNamespace.GetName(), desiredNSLocator.Workspace, desiredNSLocator.Namespace)
+				return nil
```
sttts, line 211: You have to fall through to the "The namespace exists, so check if it has the correct namespace locator." code.

jmprusi: I reverted this change. If I get an error, I throw an error to enqueue the item again; if something created the namespace before us, we would pick it in the next run by reading the namespaceLocator. Seems easier than adding another Get here.

pkg/syncer/status/status_controller.go
```diff
@@ -50,11 +51,12 @@ type Controller struct {
 
 	syncTargetName               string
 	syncTargetLogicalClusterName logicalcluster.Name
```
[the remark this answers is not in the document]
jmprusi: done

pkg/syncer/spec/spec_process.go
jmprusi: done

pkg/syncer/spec/spec_controller.go
jmprusi: done

PR #1321, opened 2022-06-21
admission: record the user creating a workspace

stevekuznetsov:

Part of #1149 
/assign @sttts

### DIFF
--- a/pkg/admission/clusterworkspacetypeexists/admission.go
+++ b/pkg/admission/clusterworkspacetypeexists/admission.go
@@ -18,12 +18,14 @@ package clusterworkspacetypeexists
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 
 	"github.com/kcp-dev/logicalcluster"
 
+	authenticationv1 "k8s.io/api/authentication/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


pkg/admission/clusterworkspacetypeexists/admission.go
```diff
@@ -99,6 +101,29 @@ func (o *clusterWorkspaceTypeExists) Admit(ctx context.Context, a admission.Attr
...
+				Username: user.GetName(),
+				UID:      user.GetUID(),
+				Groups:   user.GetGroups(),
+			}
+			extra := map[string]authenticationv1.ExtraValue{}
+			for k, v := range user.GetExtra() {
+				extra[k] = v
+			}
+			info.Extra = extra
+			rawInfo, err := json.Marshal(info)
```
stevekuznetsov, line 116: This thing does not seem to be in the scheme? So JSON here is fine?

pkg/admission/clusterworkspacetypeexists/admission.go
```diff
@@ -99,6 +101,29 @@ func (o *clusterWorkspaceTypeExists) Admit(ctx context.Context, a admission.Attr
 	}
 
 	if a.GetOperation() == admission.Create {
+		// ensure the user's UID is recorded in annotations
+		if user, ok := genericapirequest.UserFrom(ctx); ok {
```
sttts, line 105: is this different from `a.GetUser()` ?

stevekuznetsov: I did not know about that - let me use it instead

PR #1322, opened 2022-06-21
Fix partial meta request determination logic

ncdc:
## Summary
The accept header can have multiple clauses, but we were only supporting
one. The Kubernetes metadataSharedInformerFactory generates requests
with the accept header having multiple clauses, and our previous logic
didn't handle them correctly, leading to thinking the request was not
for partial metadata, when in fact it was supposed to be.

## Related issue(s)

Discovered while working on #1061


ncdc:
e2e-shared-server failure is #1324

ncdc:
This PR introduces a weird bug with APIBinding deletion & finalizers on CRs.

ncdc:
@stevekuznetsov ready for another review

PR #1323, opened 2022-06-21
virtualworkspace: add a proxy for workspace content

stevekuznetsov:
virtualworkspace: add a proxy for workspace content

---

Part of #1149 
Depends on #1321 
/assign @sttts

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -497,7 +497,7 @@ func (s *Server) Run(ctx context.Context) error {
 	}
 
 	if s.options.Virtual.Enabled {
-		if err := s.installVirtualWorkspaces(ctx, server, kubeClusterClient, dynamicClusterClient, kcpClusterClient, genericConfig.Authentication, genericConfig.ExternalAddress, preHandlerChainMux); err != nil {
+		if err := s.installVirtualWorkspaces(ctx, genericConfig.LoopbackClientConfig, server, kubeClusterClient, dynamicClusterClient, kcpClusterClient, genericConfig.Authentication, genericConfig.ExternalAddress, preHandlerChainMux); err != nil {
 			return err
 		}
 	} else if err := s.installVirtualWorkspacesRedirect(ctx, preHandlerChainMux); err != nil {
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -497,7 +497,7 @@ func (s *Server) Run(ctx context.Context) error {
 	}
 
 	if s.options.Virtual.Enabled {
-		if err := s.installVirtualWorkspaces(ctx, genericConfig.LoopbackClientConfig, server, kubeClusterClient, dynamicClusterClient, kcpClusterClient, genericConfig.Authentication, genericConfig.ExternalAddress, preHandlerChainMux); err != nil {
+		if err := s.installVirtualWorkspaces(ctx, controllerConfig, server, genericConfig.Authentication, genericConfig.ExternalAddress, preHandlerChainMux); err != nil {
 			return err
 		}
 	} else if err := s.installVirtualWorkspacesRedirect(ctx, preHandlerChainMux); err != nil {
--- a/pkg/virtual/framework/types.go
+++ b/pkg/virtual/framework/types.go
@@ -37,6 +37,15 @@ func (r RootPathResolverFunc) ResolveRootPath(urlPath string, context context.Co
 var _ RootPathResolver = RootPathResolverFunc(nil)
 
 type RootPathResolver interface {
+	// ResolveRootPath returns whether the request should be accepted and served by a given VirtualWorkspace. If accepted,
+	// the prefixToStrip is the prefix that is in-front of the kube-like API surface (including `/clusters/<something>`). E.g. for
+	//
+	//   /services/initializingworkspaces/<initializer>/clusters/<something>/apis/workload.kcp.dev/v1alpha1/workloadclusters
+	//
+	// the prefixToStrip is `/services/initializingworkspaces/<initializer/clusters/<something>`.
+	//
+	// Depending on virtual workspace type, the returned context holds e.g. the APIDomainKey, the logical cluster or
+	// other values like scope, if that is part of the URL path.
 	ResolveRootPath(urlPath string, context context.Context) (accepted bool, prefixToStrip string, completedContext context.Context)
 }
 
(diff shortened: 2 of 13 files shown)


pkg/server/server.go
```diff
@@ -497,7 +497,7 @@ func (s *Server) Run(ctx context.Context) error {
 	}
 
 	if s.options.Virtual.Enabled {
-		if err := s.installVirtualWorkspaces(ctx, server, kubeClusterClient, dynamicClusterClient, kcpClusterClient, genericConfig.Authentication, genericConfig.ExternalAddress, preHandlerChainMux); err != nil {
+		if err := s.installVirtualWorkspaces(ctx, genericConfig.LoopbackClientConfig, server, kubeClusterClient, dynamicClusterClient, kcpClusterClient, genericConfig.Authentication, genericConfig.ExternalAddress, preHandlerChainMux); err != nil {
```
sttts, line 500: use controllerConfig.

Also we should not pass config and clients. All the clients should be instantiated in installVirtualWorkspaces, with the right user-agent, as we do it with controllers.

pkg/virtual/framework/types.go
```diff
@@ -37,6 +37,13 @@ func (r RootPathResolverFunc) ResolveRootPath(urlPath string, context context.Co
 var _ RootPathResolver = RootPathResolverFunc(nil)
 
 type RootPathResolver interface {
+	// ResolveRootPath returns whether the request should be accepted and served by a given VirtualWorkspace. If accepted,
+	// the prefixToStrip is the prefix that is in-front of the kube-like API surface (including `/clusters/<something>`). E.g. for
+	//
+	//   /services/initializingworkspaces/<initializer>/clusters/<something>/apis/workload.kcp.dev/v1alpha1/workloadclusters
+	//
+	// the prefixToStrip is `/services/initializingworkspaces/<initializer/clusters/<something>`. The returned context holds the APIDomainKey and
+	// the cluster name (e.g. `<something>` here).
```
s-urbaniak, line 46: love this 👍 we should have more of this in the virtual workspace interface type

davidfestal, line 46: > The returned context holds the APIDomainKey and
	// the cluster name (e.g. `<something>` here).

Just a detail: that's not true of all virtual workspaces, but only of the `DynamicVirtualWorkspace`-based implementations.

pkg/virtual/framework/types.go
```diff
@@ -30,10 +30,49 @@ import (
...
+
+type Name string
+
+func (n Name) GetName() string {
+	return string(n)
+}
+
+var _ Namer = Name("")
+
+type Namer interface {
```
davidfestal, line 72: Are all these individual interfaces really necessary ?
I'm not convinced about them making the code clearer.

sttts:
TestCRDCrossLogicalClusterListPartialObjectMetadata flake.

sttts:
LGTM.

Waiting for @s-urbaniak 's lgtm for my commits.

pkg/virtual/framework/types.go
sttts: Can you give a better wording?

davidfestal: I'd rather add this sentence in the doc of the `DynamicVirtualWorkspace`
And here we could have something more generic like:

```
The returned context holds whatever information has been extracted from the resolved path, for example the
name of a resource or the logical cluster name.
```

pkg/virtual/framework/types.go
davidfestal: @sttts wdyt ?

sttts: alternative suggestion, last commit: https://github.com/kcp-dev/kcp/pull/1328 (follow-up)

pkg/virtual/framework/types.go
sttts: That's very unspecific. Isn't there anything required like the APIDomainKey? Who will consume it?

pkg/virtual/framework/types.go
sttts: about the interfaces, I tend to agree. These add a level of indirection (= complexity) without value.

sttts: But this is clearly taste and not blocking here.

pkg/virtual/framework/types.go
davidfestal: The virtual workspace request handler will use the values addedin this context:
- for `workspaces` virtual workspace:
  - the containing workspace
  - the scope
- For the APIExport + Syncer, ... (Based on the DynamicVirtualWorkspace):
  - the APIDomainKey
  - the logical cluster
- For a virtual workspace based on a simple Handler:
  - it can be any information the Resolving action grabs while parsing the accepted virtual workspace root path.

pkg/virtual/framework/types.go
davidfestal: I don't want to block the PR for this.

stevekuznetsov: Before these changes, every single VirtualWorkspace would hold the name, ready checker, root resolver, etc - and, every single one of them had to explicitly use these subordinate objects to implement functionality. Convince me copy-pasting this:

```go

func (vw *DynamicVirtualWorkspace) GetName() string {
	return vw.Name
}

func (vw *DynamicVirtualWorkspace) IsReady() error {
	return vw.Ready()
}

func (vw *DynamicVirtualWorkspace) ResolveRootPath(urlPath string, context context.Context) (accepted bool, prefixToStrip string, completedContext context.Context) {
	return vw.RootPathResolver(urlPath, context)
}

func (vw *DynamicVirtualWorkspace) Authorize(ctx context.Context, a authorizer.Attributes) (authorized authorizer.Decision, reason string, err error) {
	if vw.Authorizer != nil {
		return vw.Authorizer(ctx, a)
	}
	return authorizer.DecisionNoOpinion, "", nil
}
```

Is somehow better than composition?

stevekuznetsov: So IMHO the interfaces are already currently implicitly there, since we have e.g. `RootResolverFunc` but no `RootResolver`. We can remove the interface and make it less clear that `RootResolverFunc` is used as a component to compose `VirtualWorkspce` (implicit here is not bad necessarily) but IMHO writing clearly that:

1. a `VirtualWorkspace` needs to be a `RootResolver`
2. a `RootResolverFunc` is a `RootResolver`

Adds nothing new on top, just makes the current structure explicit & clear.

PR #1328, opened 2022-06-22
pkg/virtual: externalize VW name and remove extra informers

sttts:
- name how a VW is registered externally, don't make it know its name and register it like that. The former is much more idiomatic.
- remove unused extra informer code, an anti-pattern we don't want to encourage. Informers should be instatiated in external plumbing.


stevekuznetsov (approved):

stevekuznetsov (commented):
Lots of mechanical bits here that are tricky to review but on a first pass this looks good!

/hold

PR #1329, opened 2022-06-22
virtualworkspaces: propagate clarifying changes

stevekuznetsov:

Follow-up to #1323 
/assign @sttts @davidfestal

### DIFF
--- a/pkg/virtual/framework/dynamic/types.go
+++ b/pkg/virtual/framework/dynamic/types.go
@@ -17,8 +17,6 @@ limitations under the License.
 package dynamic
 
 import (
-	"context"
-
 	"k8s.io/apiserver/pkg/authorization/authorizer"
 	genericapiserver "k8s.io/apiserver/pkg/server"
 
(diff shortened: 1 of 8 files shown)


davidfestal:
I understood your explanation (on channel) about the pattern of the existing code not being the right one *in principle*.

But still I'm not really convinced that this change makes the implementation / readability of new Virtual Workspaces builders easier: It becomes required to add `ReadyFunc`, `RootPathResolverFunc`, ... where they were not necessary in the previous approach.
One goal of the initial design was that the virtual workspace builders would be, as much as possible, both readable and self-contained.

Sure, your changes simplifies, and remove a questionable pattern from the implementation of basic VirtualWorkspace templates in the framework (DynamicVirtualWorkspace, ...) but there shouldn't be that many of them afaict.

Just giving my thought here, but I wouldn't block the PR if other like this better.

pkg/virtual/framework/dynamic/types.go
```diff
@@ -31,32 +29,13 @@ var _ framework.VirtualWorkspace = (*DynamicVirtualWorkspace)(nil)
...
-	Ready            framework.ReadyFunc
+	framework.Namer
+	framework.RootPathResolver
+	authorizer.Authorizer
+	framework.ReadyChecker
 
 	// BootstrapAPISetManagement creates, initializes and returns an apidefinition.APIDefinitionSetGetter.
 	// Usually it would also set up some logic that will call the apiserver.CreateServingInfoFor() method
 	// to add an apidefinition.APIDefinition in the apidefinition.APIDefinitionSetGetter on some event.
 	BootstrapAPISetManagement func(mainConfig genericapiserver.CompletedConfig) (apidefinition.APIDefinitionSetGetter, error)
```
shawn-hurley, line 40: I would just as soon make this a typed function as well?

pkg/virtual/framework/dynamic/types.go
```diff
@@ -31,32 +29,13 @@ var _ framework.VirtualWorkspace = (*DynamicVirtualWorkspace)(nil)
 // DynamicVirtualWorkspace is an implementation of a framework.VirtualWorkspace which can dynamically serve resources,
 // based on API definitions (including an OpenAPI v3 schema), and a Rest storage provider.
 type DynamicVirtualWorkspace struct {
-	Name             string
-	RootPathResolver framework.RootPathResolverFunc
-	Authorizer       authorizer.AuthorizerFunc
-	Ready            framework.ReadyFunc
+	framework.Namer
+	framework.RootPathResolver
```
shawn-hurley, line 33: Not to throw so bikeshedding here, but RootPathResolver does not IMO encapsulate all the requirements of this function. 

This is required to also determine the APIDomainKey and set it for dynamic but the FixedGroupVersions does not need this. 

I would be wondering if two functions might be better here?

stevekuznetsov, line 33: @davidfestal after: now anything that implements this interface is valid. It's less restrictive.

stevekuznetsov: Valid concern but it seems orthogonal to this PR.

pkg/virtual/framework/dynamic/types.go
stevekuznetsov: We didn't have any shared implementations for this that we wanted to re-use between workspaces, so I wasn't keen to tackle it for now.

pkg/virtual/framework/dynamic/types.go
```diff
@@ -31,32 +29,13 @@ var _ framework.VirtualWorkspace = (*DynamicVirtualWorkspace)(nil)
 // DynamicVirtualWorkspace is an implementation of a framework.VirtualWorkspace which can dynamically serve resources,
 // based on API definitions (including an OpenAPI v3 schema), and a Rest storage provider.
 type DynamicVirtualWorkspace struct {
-	Name             string
-	RootPathResolver framework.RootPathResolverFunc
```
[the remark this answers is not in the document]
davidfestal: OK

davidfestal:
> @davidfestal my diffs make absolutely no change to the `VirtualWorkspace` interface. You can implement it any way you like. 100% the opposite of what you say - the previous implementation _required_ that there be a `RootResolverFunc()` for example as a member of the `structs` here. My changes allow you to layer in any implementation you like. They are _less_ restrictive.

My point was that it doesn't make the builders more _readable_ and _self-contained_, but I also understand your very valid point and now I can can buy it, for sure.

sttts:
lgtm

Please rebase.

stevekuznetsov:
@sttts updated to rebase on top of yours and ready for another pass

PR #1343, opened 2022-06-24
pkg/admission: add reserved metadata admission plugin

### LINKED ISSUE #1138: Lock down kcp labels and annotations
We use labels and annotations for many security sensitive purposes. We need admission to protect those to be only touched by system components.

s-urbaniak:
## Summary
This adds a new plugin locking down reserved system labels and annotations.

## Related issue(s)

Fixes #1138

Accompanying document about kcp labels/annotations: https://docs.google.com/spreadsheets/d/1DfcEqKG3koLOveddpLwebvULK2imOGJnpe0nwWZmhAY/edit#gid=0

**Question:** We marked a lot of labels/annotations as internal. I suggest I create a separate PR with the naming changes.

/cc @sttts @ncdc

### DIFF
--- a/pkg/admission/plugins.go
+++ b/pkg/admission/plugins.go
@@ -50,6 +50,7 @@ import (
 	workspacenamespacelifecycle "github.com/kcp-dev/kcp/pkg/admission/namespacelifecycle"
 	"github.com/kcp-dev/kcp/pkg/admission/reservedcrdannotations"
 	"github.com/kcp-dev/kcp/pkg/admission/reservedcrdgroups"
+	"github.com/kcp-dev/kcp/pkg/admission/reservedmetadata"
 	kcpvalidatingwebhook "github.com/kcp-dev/kcp/pkg/admission/validatingwebhook"
 )
 
@@ -67,6 +68,7 @@ var AllOrderedPlugins = beforeWebhooks(kubeapiserveroptions.AllOrderedPlugins,
 	reservedcrdannotations.PluginName,
 	reservedcrdgroups.PluginName,
 	crdnooverlappinggvr.PluginName,
+	reservedmetadata.PluginName,
 )
 
 func beforeWebhooks(recommended []string, plugins ...string) []string {
@@ -96,6 +98,7 @@ func RegisterAllKcpAdmissionPlugins(plugins *admission.Plugins) {
 	reservedcrdannotations.Register(plugins)
 	reservedcrdgroups.Register(plugins)
 	crdnooverlappinggvr.Register(plugins)
+	reservedmetadata.Register(plugins)
 }
 
 var defaultOnPluginsInKcp = sets.NewString(
--- a/pkg/admission/reservedmetadata/admission.go
+++ b/pkg/admission/reservedmetadata/admission.go
@@ -1,0 +1,150 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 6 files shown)


pkg/admission/reservedmetadata/admission.go
```diff
@@ -0,0 +1,117 @@
...
+
+var (
+	annotationAllowList = []string{
+		"scheduling.kcp.dev/placement",
+		"internal.scheduling.kcp.dev/negotiation-workspace",
+	}
+
+	labelAllowList = []string{
+		"experimental.workload.kcp.dev/scheduling-disabled",
+		"state.internal.workload.kcp.dev/*",
```
sttts, line 44: are these necessary for e2e to work? Internal APIs should not show up in these lists.

pkg/admission/plugins.go
```diff
@@ -67,6 +68,7 @@ var AllOrderedPlugins = beforeWebhooks(kubeapiserveroptions.AllOrderedPlugins,
 	reservedcrdannotations.PluginName,
 	reservedcrdgroups.PluginName,
 	crdnooverlappinggvr.PluginName,
+	reservedmetadata.PluginName,
```
sttts, line 71: how do we ensure that admission can set some of the annotation/labels, but the user cannot?

pkg/admission/reservedmetadata/admission.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if err := assertViolation(newMeta.GetLabels(), oldMeta.GetLabels(), labelAllowList); err != nil {
+		return admission.NewForbidden(a, fmt.Errorf("label violation: %v", err))
+	}
+
+	return nil
+}
+
+func assertViolation(new, old map[string]string, allowList []string) error {
+	if reflect.DeepEqual(new, old) {
```
sttts, line 98: this is not enough. We have to check per labels/annotation whether it got added/changed/removed.

pkg/admission/reservedmetadata/admission.go
s-urbaniak: Agreed, however this has been marked as mutable by coordination controller in the sheet. I can remove and see if e2e passes (pretty sure it will as everything runs as system:masters).

sttts: oh this then means that we have to change the name to non-internal.

sttts: and it needs a story how we authorize its change.

pkg/admission/reservedmetadata/admission.go
s-urbaniak: this has to compute a real delta between old and new

s-urbaniak: done using a very naive approach (iterating over new and old)

pkg/admission/plugins.go
s-urbaniak: as discussed OOB we'll tackle this part once front-proxy changes are merged.

pkg/admission/reservedmetadata/admission.go
s-urbaniak: as discussed, we just left `experimental.workload.kcp.dev/scheduling-disabled` in the allow list of labels, removed everything else.

pkg/admission/reservedmetadata/admission.go
```diff
@@ -0,0 +1,150 @@
...
+			result = append(result, kLeft)
+		}
+	}
+
+	for k := range onlyOnRight {
+		result = append(result, k)
+	}
+
+	return result
+}
```
sttts, line 132: this can be done without any allocation with two loops

PR #1345, opened 2022-06-24
reconciler/workload/namespace: stop hotloop

sttts:


ncdc:
Definitely seems to help!

/lgtm
/approve

ncdc:
I pulled this in to my quota branch and it's maybe fixed a cause of a loop, but I'm still seeing some hotloop...

PR #1352, opened 2022-06-24
Fix hotloop in placement due to bad patch

### LINKED ISSUE #1351: hotloop in placement reconciler
The patch hand-crafted in https://github.com/kcp-dev/kcp/blob/02dcb4ae2cc87e00d1e59c730a71c6df125a4a39/pkg/reconciler/scheduling/placement/placement_reconcile_scheduling.go#L91 is not always correct, leading to no-op changes and hot-looping because kcp constantly wants to remove an annotation, but it never succeeds.

ncdc:
## Summary
Fix hotloop in placement due to bad patch

## Related issue(s)

Fixes #1351

### DIFF
--- a/pkg/reconciler/scheduling/placement/placement_reconcile_scheduling.go
+++ b/pkg/reconciler/scheduling/placement/placement_reconcile_scheduling.go
@@ -79,31 +79,46 @@ func (r *placementReconciler) reconcile(ctx context.Context, ns *corev1.Namespac
 	}
 	deletePlacementAnnotation := func() (reconcileStatus, *corev1.Namespace, error) {
 		var deletePlacementAnnotation bool
-		var deleteNegotationWorkspaceAnnoation bool
+		var deleteNegotationWorkspaceAnnotation bool
 
 		if _, found := ns.Annotations[schedulingv1alpha1.PlacementAnnotationKey]; found {
 			deletePlacementAnnotation = true
 		}
 		if _, found := ns.Annotations[schedulingv1alpha1.InternalNegotiationWorkspaceAnnotationKey]; found {
-			deleteNegotationWorkspaceAnnoation = true
+			deleteNegotationWorkspaceAnnotation = true
 		}
-		if deletePlacementAnnotation || deleteNegotationWorkspaceAnnoation {
+		if deletePlacementAnnotation || deleteNegotationWorkspaceAnnotation {
 			klog.V(4).Infof("Removing placement from namespace %s|%s, no api bindings", ns.Name)
 
-			nsCopy := ns.DeepCopy()
-
-			if deletePlacementAnnotation {
-				delete(nsCopy.Annotations, schedulingv1alpha1.PlacementAnnotationKey)
-			}
-			if deleteNegotationWorkspaceAnnoation {
-				delete(nsCopy.Annotations, schedulingv1alpha1.InternalNegotiationWorkspaceAnnotationKey)
-			}
-
-			oldData, err := json.Marshal(ns)
+			oldData, err := json.Marshal(&corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: ns.Annotations,
+				},
+			})
 			if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/reconciler/scheduling/placement/placement_reconcile_scheduling.go
```diff
@@ -77,18 +78,41 @@ func (r *placementReconciler) reconcile(ctx context.Context, ns *corev1.Namespac
 		return reconcileStatusStop, nil, err
 	}
 	deletePlacementAnnotation := func() (reconcileStatus, *corev1.Namespace, error) {
-		needsPatch := false
+		var deletePlacement bool
+		var deleteInternal bool
```
sttts, line 82: bad name. deleteNegotationWorkspaceAnnoation

ncdc: 😄 fixed

pkg/reconciler/scheduling/placement/placement_reconcile_scheduling_test.go
```diff
@@ -88,7 +88,7 @@ func TestPlacementReconciler(t *testing.T) {
 			apibindings: map[logicalcluster.Name][]*apisv1alpha1.APIBinding{logicalcluster.New("root:org:ws"): {
 				bound(validExport(binding("kubernetes", "root:org:negotiation-workspace"))),
 			}},
-			wantPatch:           `{"metadata":{"annotations":{"scheduling.kcp.dev/placement":null}}}`,
+			wantPatch:           `{"metadata":{"annotations":null}}`,
```
sttts, line 91: 🔴  this is a blocker. We cannot deletion all annotation without uid+rv pre-conditions.

ncdc: I just had that thought. Will add

pkg/reconciler/scheduling/placement/placement_reconcile_scheduling.go
```diff
@@ -77,18 +78,56 @@ func (r *placementReconciler) reconcile(ctx context.Context, ns *corev1.Namespac
...
+				if k == schedulingv1alpha1.PlacementAnnotationKey && deletePlacementAnnotation {
+					continue
+				}
+
+				if k == schedulingv1alpha1.InternalNegotiationWorkspaceAnnotationKey && deleteNegotationWorkspaceAnnotation {
+					continue
+				}
+
+				newAnnotations[k] = v
+			}
```
sttts, line 113: is this different to two deletes?

pkg/reconciler/scheduling/placement/placement_reconcile_scheduling.go
```diff
@@ -77,18 +78,56 @@ func (r *placementReconciler) reconcile(ctx context.Context, ns *corev1.Namespac
...
+			})
+			if err != nil {
+				return reconcileStatusStop, nil, err
+			}
+			patchBytes, err := jsonpatch.CreateMergePatch(oldData, newData)
+			if err != nil {
+				return reconcileStatusStop, nil, err
+			}
+
+			if _, err := r.patchNamespace(ctx, clusterName, ns.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{}); err != nil {
```
sttts, line 130: looks correct. But this code always hurts my eyes. 30 lines to create a patch.

pkg/reconciler/scheduling/placement/placement_reconcile_scheduling.go
ncdc: I can simplify it. Was trying to avoid 2 deep copies

pkg/reconciler/scheduling/placement/placement_reconcile_scheduling.go
stevekuznetsov: Better than three bug-fix commits for typos in a patch string ... generics should help us with the old/new marshalling and patch creation?

PR #1355, opened 2022-06-25
e2e/syncer: make the test app sleep, not restart with AlreadyExists

sttts:
Before this the first run of the container created the configmap, the second and following failed with AlreadyExists and panic'ed, leading to unavailable deployment under bad timing.


sttts:
@jmprusi

sttts:
Stop the bleeding.Merging.

PR #1357, opened 2022-06-25
e2e: cope with asynchronously served (bound) CRDs

sttts:
As in kube, CRDs can be slightly delayed because they are served informer based.


sttts:
@ncdc

PR #1360, opened 2022-06-27
virtual: clean up post-start errors and var names

sttts:
Post-start-hooks should never return an error. They should just block. An error leads to stack trace output. Why? Legacy reasons 🤷

With this the terminal is not filled with stack traces on ctrl-c on startup.


s-urbaniak:
This is a bummer can we contribute a better godoc of `PostStartHookFunc` to upstream?

PR #1365, opened 2022-06-27
server/apiextensions: stop serving non-wildcard identity requests

sttts:
Per workspace no identity is necessary as GVR is unique. This mode of access was never expected to work.


sttts:
e2e-sharded

PR #1370, opened 2022-06-27
Changes to the downstream namespace generation

jmprusi:
This PR changes the pattern used to generate the downstream namespace. It generates a shorter string while maintaining a big enough random part to avoid conflicts.

related: https://github.com/kcp-dev/kcp/issues/1280

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/googleapis/gnostic v0.5.5
 	github.com/kcp-dev/kcp/pkg/apis v0.0.0-00010101000000-000000000000
 	github.com/kcp-dev/logicalcluster v1.1.0
+	github.com/martinlindhe/base36 v1.1.1 // indirect
 	github.com/muesli/reflow v0.1.0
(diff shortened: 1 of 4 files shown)


ncdc:
@jmprusi what is the status of this vs. #1309? Are you closing 1309, or ... ?

jmprusi:
@ncdc I'm splitting it as I'm running into a problem with the namespace locator struct. I prefer  to unblock some of the folks that are running into issues with the namespace length, and take some time to fix the other changes.

stevekuznetsov:
Seems reasonable but will break previous syncer deployments, right?

jmprusi:
@sttts yes, I have another Branch with those changes, I've run into a bug and I wanted to get this change into the release, but as @stevekuznetsov points out, this change will cause the duplication of all the resources... as the status syncer will read the value from the namespace locator, so it will work, but the spec syncer will create a new namespace using the new pattern. 

We can document this as a required step or wait for the final version, that doesn't require a "migration"...

Let's wait for the full functionality?

jmprusi:
let's keep the #1309 with the full functionality to make it easier to migrate from a previous version.

PR #1372, opened 2022-06-27
Various dynamic discovery shared informer factory updates

ncdc:
## Summary
- Add global indexers for cluster- and namespace-scoped resources.
- Install these indexers by default into the shared informer factories
that need them (i.e. most of them).
- Make the dynamic discovery shared informer factory a field on Server to
prep for using it
- Remove IndexerFor method - require callers to get the informer, then
call .Informer().GetIndexer() (only had 1 caller).
- Remove gvrs field that was redundant - go by informers field instead.
- Move event handler configuration to where the informer is actually
created (informerForResourceLockHeld) because the upcoming quota monitor code
creates informers by calling InformerForResource (indirectly).
- Add Start(stopCh <-chan struct{}) function to support the way kube quota
 starts newly created informers.
- Rename previous Start method to StartPolling.
- Protect handlers with lock

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/indexers/indexers.go
+++ b/pkg/indexers/indexers.go
@@ -1,0 +1,67 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package indexers
+
+import (
+	"github.com/kcp-dev/logicalcluster"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clusters"
+)
+
+const (
+	// ByLogicalCluster is the name for the index that indexes by an object's logical cluster.
+	ByLogicalCluster = "kcp-global-byLogicalCluster"
+	// ByLogicalClusterAndNamespace is the name for the index that indexes by an object's logical cluster and namespace.
+	ByLogicalClusterAndNamespace = "kcp-global-byLogicalClusterAndNamespace"
+)
+
+// ClusterScoped returns cache.Indexers appropriate for cluster-scoped resources.
+func ClusterScoped() cache.Indexers {
+	return cache.Indexers{
+		ByLogicalCluster: IndexByLogicalCluster,
+	}
+}
+
+// NamespaceScoped returns cache.Indexers appropriate for namespace-scoped resources.
+func NamespaceScoped() cache.Indexers {
+	return cache.Indexers{
+		ByLogicalCluster:             IndexByLogicalCluster,
+		ByLogicalClusterAndNamespace: IndexByLogicalClusterAndNamespace,
+	}
+}
+
+// IndexByLogicalCluster is an index function that indexes by an object's logical cluster.
+func IndexByLogicalCluster(obj interface{}) ([]string, error) {
+	a, err := meta.Accessor(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	return []string{logicalcluster.From(a).String()}, nil
+}
+
+// IndexByLogicalClusterAndNamespace is an index function that indexes by an object's logical cluster and namespace.
+func IndexByLogicalClusterAndNamespace(obj interface{}) ([]string, error) {
+	a, err := meta.Accessor(obj)
+	if err != nil {
+		return nil, err
+	}
+
+	return []string{clusters.ToClusterAwareKey(logicalcluster.From(a), a.GetNamespace())}, nil
+}
--- a/pkg/informer/informer.go
+++ b/pkg/informer/informer.go
@@ -21,11 +21,13 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/kcp-dev/logicalcluster"
 
 	corev1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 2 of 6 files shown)


pkg/indexers/indexers.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/kcp-dev/logicalcluster"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clusters"
+)
+
+const (
+	// ByLogicalCluster is the name for the index that indexes by an object's logical cluster.
+	ByLogicalCluster = "kcp-global-byLogicalCluster"
```
sttts, line 29: we call this ByWorkspace everywhere else.

What is kcp-global? A new convention? Why not call it just `byWorkspace` and `byWorkspaceAndNamespace` ? I think we really want an indexers package with helpers to add those indexers without falling over each other when they are added multiple times.

sttts: oh, this is it :)

sttts: Then, can we add the mentioned helper here instead of just referencing `ClusterScoped` and `NamespaceScoped`?

pkg/informer/informer.go
```diff
@@ -108,10 +109,47 @@ func (d *DynamicDiscoverySharedInformerFactory) informerForResourceLockHeld(gvr
...
+		FilterFunc: d.filterFunc,
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj interface{}) {
+				d.handlerLock.RLock()
+
+				for _, h := range d.handlers {
+					h.OnAdd(gvr, obj)
+				}
+
+				d.handlerLock.RUnlock()
```
sttts, line 122: this is not good. Calling out to foreign code under a lock is a nogo. Why not make `d.handlers` update copy on write?

pkg/informer/informer.go
```diff
@@ -53,32 +55,29 @@ type DynamicDiscoverySharedInformerFactory struct {
...
-func (d *DynamicDiscoverySharedInformerFactory) IndexerFor(gvr schema.GroupVersionResource) cache.Indexer {
-	return d.InformerForResource(gvr).Informer().GetIndexer()
+	mu               sync.RWMutex
+	informers        map[schema.GroupVersionResource]informers.GenericInformer
+	startedInformers map[schema.GroupVersionResource]bool
+	informerStops    map[schema.GroupVersionResource]chan struct{}
+	terminating      bool
 }
 
 // InformerForResource returns the GenericInformer for gvr, creating it if needed.
```
sttts, line 72: doc that is must be started

pkg/informer/informer.go
```diff
@@ -214,7 +253,7 @@ func (d *DynamicDiscoverySharedInformerFactory) AddIndexers(indexers cache.Index
 	return nil
 }
 
-func (d *DynamicDiscoverySharedInformerFactory) Start(ctx context.Context) {
```
sttts, line 217: add doc that this blocks

pkg/indexers/indexers.go
ncdc: Yes. And once the new informers&listers are ready, we'll do a full cleanup.

pkg/informer/informer.go
ncdc: Thanks for the suggestion - will fix

pkg/informer/informer.go
```diff
@@ -374,21 +391,46 @@ func (d *DynamicDiscoverySharedInformerFactory) discoverTypes(ctx context.Contex
...
 		delete(d.informerStops, gvr)
+		delete(d.startedInformers, gvr)
 	}
 
-	d.gvrs = latest
-
 	return nil
 }
 
+func (d *DynamicDiscoverySharedInformerFactory) Start(stop <-chan struct{}) {
```
sttts, line 400: nit: stopCh

pkg/informer/informer.go
```diff
@@ -374,21 +391,46 @@ func (d *DynamicDiscoverySharedInformerFactory) discoverTypes(ctx context.Contex
...
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	for gvr, informer := range d.informers {
+		if !d.startedInformers[gvr] {
+			go informer.Informer().Run(stop)
```
sttts, line 406: am not happy with the lifecycle semantic of this informer. When stop closed, and this could be long before the whole ddsif context is done, we leak some stopped informer here. Either we have to remove it from the ddsif, or we should just ignore the stop channel and create a fresh one and place it in the stop channel map.

pkg/informer/informer.go
```diff
@@ -129,10 +167,9 @@ func (d *DynamicDiscoverySharedInformerFactory) Listers() (listers map[schema.Gr
 		return
 	}
 
-	for gvr := range d.gvrs {
+	for gvr, informer := range d.informers {
```
sttts, line 170: if we are adding cluster scoped resources now, should we maybe have a `NamespacedListers()`, `ClusterScopedListers()` ?

pkg/informer/informer.go
```diff
@@ -54,32 +57,30 @@ type DynamicDiscoverySharedInformerFactory struct {
...
 	indexers        cache.Indexers
 
-	mu            sync.RWMutex // guards gvrs
-	gvrs          map[schema.GroupVersionResource]struct{}
-	informers     map[schema.GroupVersionResource]informers.GenericInformer
-	informerStops map[schema.GroupVersionResource]chan struct{}
-	terminating   bool
-}
+	handlersLock sync.Mutex
+	handlers     atomic.Value
```
sttts, line 65: a lock plus atomic? This looks very odd now. I could see either or.

ncdc: Otherwise you risk data loss with 2 simultaneous writers

pkg/informer/informer.go
```diff
@@ -374,21 +391,46 @@ func (d *DynamicDiscoverySharedInformerFactory) discoverTypes(ctx context.Contex
 		klog.V(4).Infof("Removing dynamic informer from maps for %q", gvr)
 		delete(d.informers, gvr)
 		delete(d.informerStops, gvr)
+		delete(d.startedInformers, gvr)
 	}
 
-	d.gvrs = latest
```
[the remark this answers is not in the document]
ncdc: Recording for posterity from our Meet chat: d.gvrs came first, before we tracked d.informers (so we could stop them). Don't think there was a reason other than that.

pkg/informer/informer.go
```diff
@@ -109,10 +113,35 @@ func (d *DynamicDiscoverySharedInformerFactory) informerForResourceLockHeld(gvr
 		nil,
 	)
 
+	inf.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: d.filterFunc,
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj interface{}) {
+				for _, h := range d.handlers.Load().([]GVREventHandler) {
```
sttts, line 120: :+1:

Issue #1069, opened 2022-05-17
Epic: User home workspaces


ncdc:
# Introduction

We've been brainstorming the idea of replacing the current implementation of personal workspaces (+ pretty names) with user home workspaces. The current implementation has some shortcomings:

- Input doesn't necessarily equal output. When you request a personal workspace named "foo", if the parent workspace already has another workspace named "foo," the new workspace is named something like "foo-a7fje." If you run `kubectl workspace list`, you'll see "foo" even though underneath it has a different name.
- The duality created by ^ causes confusion because `kubectl get workspaces` and `kubectl workspace list` can show different sets of data
- There is a lot of of complex code (virtual workspace) to support personal workspaces in this manner. We could reduce complexity by using OOTB RBAC
- It's not easy to grant additional access to a personal workspace

One issue we've been struggling with is what is the starting point for a user coming to kcp? Assuming kcp is configured for OIDC, does a user have a "home organization" or anything like that? If we can say that it's a requirement for there to be a globally unique identifier per user, than we can say their starting workspace is a descendent of the `root:users:...` hierarchy.

# Demo Objective

Show that users can manage their own personal workspaces inside a their dedicated `Home` workspace. 

# Demo Steps

- [x] The user should be able to `kubectl get workspace '~'` in the root workspace, and this would trigger the creation of his own dedicated home workspace (something like `root:users:jd:jdoe`) and return the corresponding workspace resource when available and ready.
- [x] When the user home workspace is created on-the-fly (during first access), it will delay the CLI until the user home workspace is created (through some HTTP 429 "RetryAfter" response), and the initial request will be sent to the newly-created home workspace transparently.
- [x] The user is `cluster-admin` in his home workspace. By default he would create workspaces of Universal type in his home workspace.
- [x] We should allow any KCP shard to perform this on-the-fly creation of the user home workspace: we should not have to point to the shard that contains the bucket cluster workspace.  

=== Nice to have, but not the priority ===

- [x] CLI plugin support: Support the `kubectl ws ~` notation to go to switch to the home workspace, with the `kubectl ws` as an alias.
- [ ] ~Sharding integration: Integration in the front-proxy: the front-proxy should redirect to the shard that is expected to contains the bucket workspace in which in user home workspace take place.~ if we need more here, we will create a new issue

=== stretch goals for 0.6.0, very likely 0.7.0 ===

- [x] Allow users to create individual objects than workspaces at the home level
[...]

PR #1373, opened 2022-06-27
Home workspaces proof PR

davidfestal:
## Related issue(s)

Fixes #


config/home/clusterworkspacetype-use-clusterrole.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:kcp:universal-clusterworkspacetype-use
+rules:
+- apiGroups: ["tenancy.kcp.dev"]
+  resources: ["clusterworkspacetypes"]
+  resourceNames: ["universal"]
+  verbs: ["use"]
```
sttts, line 9: don't think that's how it works now. No roles+bindings necessary anymore deep in the hierarchy, only in root.

config/home/clusterworkspacetype-use-clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: system:kcp:authenticated:universal-clusterworkspacetype-use
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:kcp:universal-clusterworkspacetype-use
+subjects:
+- apiGroup: rbac.authorization.k8s.io
+  kind: Group
+  name: system:authenticated
```
sttts, line 12: same

config/homebucket/clusterworkspacetype-use-home-clusterrole.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:kcp:home-clusterworkspacetype-use
+rules:
+- apiGroups: ["tenancy.kcp.dev"]
+  resources: ["clusterworkspacetypes"]
+  resourceNames: ["home"]
+  verbs: ["use"]
```
sttts, line 9: same

config/homebucket/clusterworkspacetype-use-home-clusterrolebinding.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
sttts, line 1: in general, we have a naming convention for these files. The resource comes first, not last.

config/root/clusterworkspace-home-get-clusterrole.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: system:kcp:home-clusterworkspace-get
+rules:
+- apiGroups: ["tenancy.kcp.dev"]
+  resources: ["clusterworkspaces/workspace"]
+  resourceNames: ["~"]
+  verbs: ["get"]
```
sttts, line 9: just to be on the same page: this stays "get" only, correct? I.e. you cannot list?

config/root/clusterworkspacetype-home.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: ClusterWorkspaceType
+metadata:
+  name: home
+spec:
+  initializer: true
+  defaultChildWorkspaceType: Universal
+  allowedChildWorkspaceTypes:
+  - Any
+  allowedParentWorkspaceTypes:
+  - Homebucket
```
sttts, line 11: nit: HomeBucket

config/root/clusterworkspacetype-homebucket.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: homebucket
+spec:
+  initializer: true
+  allowedChildWorkspaceTypes:
+  - Homebucket
+  - Home
+  allowedParentWorkspaceTypes:
+  - Homebucket
+  - Homeroot
+
```
sttts, line 13: nit: no new line

pkg/server/server.go
```diff
@@ -390,7 +391,10 @@ func (s *Server) Run(ctx context.Context) error {
 					"shard": {Cluster: "shard"},
 				},
 				CurrentContext: "shard",
-			}); err != nil {
+			},
+			[]string{user.AllAuthenticated},
+			[]string{user.AllAuthenticated},
```
sttts, line 407: is there a follow-up to turn these into flags?

config/root/clusterworkspacetype-homebucket.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: homebucket
+spec:
+  initializer: true
+  allowedChildWorkspaceTypes:
+  - Homebucket
+  - Home
+  allowedParentWorkspaceTypes:
+  - Homebucket
+  - Homeroot
```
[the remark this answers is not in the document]
davidfestal: I cannot do that. Upper case letters inside the type name are not possible (due to implementation of `ClusterWorkspaceType`s)

config/root/clusterworkspacetype-home.yaml
davidfestal: I cannot do that. Upper case letters inside the type name are not possible (due to implementation of `ClusterWorkspaceType`s)

sttts: I see `[A-Z][a-zA-Z0-9]+` in the OpenAPI.

sttts: Not sure we enforce that only the first letter is upper-case. Do we? cc @stevekuznetsov

pkg/virtual/workspaces/registry/rest.go
```diff
@@ -599,16 +599,19 @@ func (s *REST) Create(ctx context.Context, obj runtime.Object, createValidation
 		return nil, kerrors.NewInvalid(tenancyv1beta1.SchemeGroupVersion.WithKind("Workspace").GroupKind(), obj.GetObjectKind().GroupVersionKind().String(), []*field.Error{})
 	}
 
+	if workspace.Spec.Type.Name == "" && workspace.Spec.Type.Path == "" {
+		workspace.Spec.Type.Name = "Universal"
+		workspace.Spec.Type.Path = "root"
+	}
```
sttts, line 605: no, this is not a static default anymore. It depends on the parent workspace, which has a default type for children.

davidfestal: I know, but this would probably have to be done in a followup PR, when we completely remove the need for a SAR here, after changing the CTWexists admission plugin to use the owner annotation content.

pkg/virtual/workspaces/builder/build.go
```diff
@@ -47,7 +48,7 @@ import (
 	"github.com/kcp-dev/kcp/pkg/virtual/workspaces/registry"
 )
 
-func BuildVirtualWorkspace(rootPathPrefix string, wildcardsClusterWorkspaces workspaceinformer.ClusterWorkspaceInformer, wildcardsRbacInformers rbacinformers.Interface, kubeClusterClient kubernetes.ClusterInterface, kcpClusterClient kcpclient.ClusterInterface) framework.VirtualWorkspace {
+func BuildVirtualWorkspace(kubeClientConfig *clientrest.Config, rootPathPrefix string, wildcardsClusterWorkspaces workspaceinformer.ClusterWorkspaceInformer, wildcardsRbacInformers rbacinformers.Interface, kubeClusterClient kubernetes.ClusterInterface, kcpClusterClient kcpclient.ClusterInterface) framework.VirtualWorkspace {
```
sttts, line 51: which rest config is that? A cluster config? Then please call it like that. Why is there also a kubeClusterClient?

pkg/virtual/workspaces/registry/rest.go
```diff
@@ -286,14 +299,22 @@ func (s *REST) authorizeForUser(ctx context.Context, orgClusterName logicalclust
 
 	}
 
+	impersonatedClusterClient, err := s.impersonatedkubeClusterClient(&kuser.DefaultInfo{
+		Name:   currentUser.GetName(),
+		Groups: []string{kuser.SystemPrivilegedGroup},
```
sttts, line 304: what is this doing? Without a clear explanation this look frightening.

pkg/server/handler.go
```diff
@@ -209,6 +211,11 @@ func WithWorkspaceProjection(apiHandler http.Handler) http.HandlerFunc {
 			return
 		}
 
+		if cluster.Name == tenancyv1alpha1.RootCluster && req.URL.Path == getHomeWorkspaceRequestPath {
```
sttts, line 214: what is this doing here?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
+/*
+Copyright 2021 The KCP Authors.
```
sttts, line 2: 22

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+	tenancyv1beta1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1beta1"
+	"github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/util/conditions"
+	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	kcpexternalversions "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
+)
+
+const (
+	workspaceFullPathIndexName = "workspace-full-path"
+	homeOwnerRBACName          = "home-owner"
+	homeOwnerRBACPrefix        = "home-owner-"
```
sttts, line 59: what is an RBAC? what are these constants?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+var (
+	homeWorkspaceScheme = runtime.NewScheme()
+	homeWorkspaceCodecs = serializer.NewCodecFactory(homeWorkspaceScheme)
+)
+
+func init() {
+	_ = tenancyv1beta1.AddToScheme(homeWorkspaceScheme)
+}
+
+func WithHomeWorkspaces(apiHandler http.Handler, a authorizer.Authorizer, kubeClusterClient kubernetes.ClusterInterface, kcpClusterClient kcpclient.ClusterInterface, kubeSharedInformerFactory coreexternalversions.SharedInformerFactory, kcpSharedInformerFactory kcpexternalversions.SharedInformerFactory, creationDelay int, homePrefix logicalcluster.Name, bucketLevels, bucketSize int) http.Handler {
```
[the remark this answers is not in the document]
sttts: what is a delay with int type?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+	_ = addWorkspaceFullPathIndexer(clusterWorkspaceInformer)
+
+	return &homeWorkspaceHandler{
+		apiHandler:    apiHandler,
+		authz:         a,
+		creationDelay: creationDelay,
+		homePrefix:    homePrefix,
+		bucketLevels:  bucketLevels,
+		bucketSize:    bucketSize,
+		getLocalClusterWorkspace: func(fullName logicalcluster.Name) (*tenancyv1alpha1.ClusterWorkspace, error) {
```
sttts, line 85: what is a fullName?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+			if err != nil {
+				return nil, err
+			}
+			if len(result) == 0 {
+				return nil, nil
+			}
+			return result[0].(*tenancyv1alpha1.ClusterWorkspace), nil
+		},
+		searchLocalClusterRole: func(workspace logicalcluster.Name, name string) (bool, error) {
+			_, ok, err := crInformer.GetIndexer().GetByKey(clusters.ToClusterAwareKey(workspace, name))
```
sttts, line 96: use a lister, not an indexer for by-name. It will be typed, and it is what we do everywhere else.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+			_, err := kcpClusterClient.Cluster(workspace).TenancyV1alpha1().ClusterWorkspaces().Create(ctx, cw, metav1.CreateOptions{})
+			return err
+		},
+		getClusterWorkspace: func(ctx context.Context, workspace logicalcluster.Name, name string) (*tenancyv1alpha1.ClusterWorkspace, error) {
+			return kcpClusterClient.Cluster(workspace).TenancyV1alpha1().ClusterWorkspaces().Get(ctx, name, metav1.GetOptions{})
+		},
+	}
+}
+
+func addWorkspaceFullPathIndexer(clusterWorkspaceInformer cache.SharedIndexInformer) error {
```
sttts, line 126: move to indexes.go

sttts: why is this index needed in the first place? it is doing the same as looking up by `clusters.ToClusterAwareKey(clusterName, clusterWorkspaceName)`

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+			}
+
+			cluster := logicalcluster.From(o)
+			return []string{cluster.Join(o.Name).String()}, nil
+		},
+	})
+}
+
+type searchFunc func(workspace logicalcluster.Name, name string) (bool, error)
+type mockableFunctions struct {
```
sttts, line 141: surprising test code in production code file.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+	getClusterWorkspace    func(ctx context.Context, workspace logicalcluster.Name, name string) (*tenancyv1alpha1.ClusterWorkspace, error)
+
+	getLocalClusterWorkspace      func(fullName logicalcluster.Name) (*tenancyv1alpha1.ClusterWorkspace, error)
+	searchLocalClusterRole        searchFunc
+	searchLocalClusterRoleBinding searchFunc
+	synced                        func() bool
+
+	homePrefix               logicalcluster.Name
+	bucketLevels, bucketSize int
+	creationDelay            int
```
sttts, line 166: time.Duration. And what is this?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+		responsewriters.InternalError(rw, req, err)
+		return
+	}
+	requestInfo, ok := request.RequestInfoFrom(ctx)
+	if !ok {
+		responsewriters.InternalError(rw, req, errors.New("No request Info"))
+		return
+	}
+
+	if workspaceName == tenancyv1alpha1.RootCluster &&
```
sttts, line 198: a 10,000ft comment what the next 50 lines do wouldn't harm

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+		requestInfo.IsResourceRequest &&
+		requestInfo.Verb == "get" &&
+		requestInfo.APIGroup == tenancyv1beta1.SchemeGroupVersion.Group &&
+		requestInfo.Resource == "workspaces" &&
+		requestInfo.Name == "~" {
+		var homeWorkspace *tenancyv1beta1.Workspace
+		homeWorkspaceFullName := h.getHomeWorkspaceName(user.GetName())
+		homeClusterWorkspace, err := h.getLocalClusterWorkspace(homeWorkspaceFullName)
```
sttts, line 206: fullName and local name are no known concepts. What are these? home `logicalcluster.Name` and `ClusterWorkspace.Name`?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+		if err != nil {
+			responsewriters.InternalError(rw, req, err)
```
sttts, line 208: this is bad for something lister based, asks for races.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+		if homeClusterWorkspace != nil {
+			// The home workspace for the user exists, so let's return it to the user if he request to see its definition
+
+			homeWorkspace = &tenancyv1beta1.Workspace{}
+			projection.ProjectClusterWorkspaceToWorkspace(homeClusterWorkspace, homeWorkspace)
+		} else {
+			// Test if the user has the right to see his Home workspace even though it doesn't exists
+			// => test the get verb on the clusterworkspaces/workspace subresource named ~ in the root workspace.
+
+			attributes := homeWorkspaceAuthorizerAttributes(user, "get")
```
sttts, line 220: why don't we authz the previous case where the home workspace was found?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+				responsewriters.InternalError(rw, req, err)
+				return
+			}
+			if decision != authorizer.DecisionAllow {
+				responsewriters.Forbidden(ctx, attributes, rw, req, reason, homeWorkspaceCodecs)
+				return
+			}
+
+			parent, name := homeWorkspaceFullName.Split()
+			homeWorkspaceURL := &url.URL{Scheme: "https", Host: req.Host, Path: homeWorkspaceFullName.Path()}
```
sttts, line 232: this is wrong. `req.Host` is the shard if I am not mistaken. This must be the front-proxy.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+				},
+			}
+		}
+
+		responsewriters.WriteObjectNegotiated(homeWorkspaceCodecs, negotiation.DefaultEndpointRestrictions, tenancyv1beta1.SchemeGroupVersion, rw, req, http.StatusOK, homeWorkspace)
+		return
+	}
+
+	needsAutomaticCreation, workspaceType := h.needsAutomaticCreation(workspaceName)
+	if needsAutomaticCreation {
```
sttts, line 255: (style) return early for `!needsAutomaticCreation`

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+		isHome := workspaceType == "Home"
+		var retryAfterSeconds int
+		var retryLater bool
+		var localCheckError error
+		var foundLocally bool
```
sttts, line 260: why all these vars???

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+		foundLocally, retryLater, retryAfterSeconds, localCheckError = h.checkWorkspaceLocally(workspaceName, isHome, user.GetName())
```
sttts, line 261: s/localCheckError/err/

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+			responsewriters.InternalError(rw, req, localCheckError)
+			return
+		}
+
+		if !foundLocally {
+			// Home or bucket workspace not found in the local informer
+			// Let's try to create it
+
+			// But first check we have the right to do so.
+			attributes := homeWorkspaceAuthorizerAttributes(user, "create")
```
sttts, line 272: move to the Authorize call

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+
+			if isHome && workspaceName != h.getHomeWorkspaceName(user.GetName()) {
+				// If we're checking a home workspace or home bucket workspace, but not of the consistent user, let's refuse.
+				responsewriters.Forbidden(ctx, attributes, rw, req, "home workspace can only be created by the user of the home workspace", homeWorkspaceCodecs)
```
sttts, line 276: we cannot leak that detailed information. See the error constants in pkg/authorization.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+			// Return a 429 status asking the client to try again after the creationDelay
+			rw.Header().Set("Retry-After", fmt.Sprintf("%d", retryAfterSeconds))
+			http.Error(rw, "Creating the home workspace", http.StatusTooManyRequests)
+			return
+		}
+	}
+	h.apiHandler.ServeHTTP(rw, req)
+}
+
+var reHomeWorkspaceChar = regexp.MustCompile("[^a-z0-9-]")
```
sttts, line 311: go doc. What is this?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+
+// getHomeWorkspaceName returns the full name of the home workspace for a given user
+// This full name includes home bucket workspaces whose name is based on the user name
+// sha1 hash.
+func (h *homeWorkspaceHandler) getHomeWorkspaceName(userName string) logicalcluster.Name {
+	// bucketLevels <= 5
+	// bucketSize <= 4
```
sttts, line 318: what does this mean?

pkg/virtual/workspaces/registry/rest.go
davidfestal: This is part of Splitted PR https://github.com/kcp-dev/kcp/pull/1456

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+	switch {
+	case length <= homeRootWorkspaceLength:
+		return
+	case length <= lastBucketWorkspaceLength:
+		return true, "Homebucket"
+	case length == homeWorkspaceLength:
+		return true, "Home"
+	}
+	return
+}
```
sttts, line 366: this function is fragile. Please only use method of the logicalcluster. It has everything necessary here.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+	userName = reHomeWorkspaceChar.ReplaceAllLiteralString(userName, "-")
+	userName = strings.TrimLeftFunc(userName, func(r rune) bool {
+		return r <= '9'
+	})
+	userName = strings.TrimRightFunc(userName, func(r rune) bool {
+		return r == '-'
+	})
+
+	return result.Join(userName)
+}
```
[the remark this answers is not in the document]
davidfestal: Aren't they good enough ?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+	return
+}
+
+// checkWorkspaceLocally checks whether a workspace is known on the current shard
+// in local informers.
+// For home workspaces, we also check:
+//   - if related RBAC resources are there,
+//   - if the workspace phase is READY
+// and if not answer to retry later.
+func (h *homeWorkspaceHandler) checkWorkspaceLocally(workspaceName logicalcluster.Name, isHome bool, userName string) (found, retryLater bool, retryAfterSeconds int, checkError error) {
```
sttts, line 374: am not a fan of "check*" functions with 60 lines of random checks.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+// tryToCreate tries to create the home or home bucket workspace, with an external client,
+// if it has not been found in the local informers.
+// Of course it can be that it has been created in the meantime (either concurrently on the current shard or
+// on another shard). We don't error in this case.
+// If it cannpot be created because the parent home bucket doesn't exist, create the parent first and
+// retry the creation of the current workspace later.
+// When creating a Home workspace, also create the related RBAC resources.
+func (h *homeWorkspaceHandler) tryToCreate(ctx context.Context, userName string, workspaceToCheck logicalcluster.Name, workspaceType tenancyv1alpha1.ClusterWorkspaceTypeName) (retryLater bool, retryAfterSeconds int, createError error) {
+	if h.mocks.tryToCreate != nil {
+		return h.mocks.tryToCreate(ctx, userName, workspaceToCheck, workspaceType)
```
sttts, line 447: h.mocks rings alarm bells. Why do we need a mocks sub-struct?!

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+	}
+
+	// Try to create it in the parent workspace
+	parent, name := workspaceToCheck.Split()
+	createError = h.createClusterWorkspace(ctx, parent, &tenancyv1alpha1.ClusterWorkspace{
```
sttts, line 452: err please. this doesn't look like go

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+		Spec: tenancyv1alpha1.ClusterWorkspaceSpec{
+			Type: tenancyv1alpha1.ClusterWorkspaceTypeReference{Path: "root", Name: workspaceType},
+		},
+	})
+
+	workspaceExists := createError == nil || kerrors.IsAlreadyExists(createError)
+	if workspaceExists {
```
sttts, line 462:
```
if err := h.createClusterWorkspace( ....
); err == nil || IsAlreadyExistts(err) {
...
}

File: pkg/server/home_workspaces.go [hunk 40]
```diff
@@ -0,0 +1,661 @@
...
+}
+
+// tryToCreate tries to create the home or home bucket workspace, with an external client,
+// if it has not been found in the local informers.
+// Of course it can be that it has been created in the meantime (either concurrently on the current shard or
+// on another shard). We don't error in this case.
+// If it cannpot be created because the parent home bucket doesn't exist, create the parent first and
+// retry the creation of the current workspace later.
+// When creating a Home workspace, also create the related RBAC resources.
+func (h *homeWorkspaceHandler) tryToCreate(ctx context.Context, userName string, workspaceToCheck logicalcluster.Name, workspaceType tenancyv1alpha1.ClusterWorkspaceTypeName) (retryLater bool, retryAfterSeconds int, createError error) {
```
sttts, line 445: s/createError/err/

And please (nearly) always do explicit `return ..., err`. Relying on return variable name is good for documentation, but fragile.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+				return
+			}
+		}
+		// Retry sooner than the creation delay, because it's probably a question of
+		// letting the local informer cache being updated.
+		// Retrying quicly should be sufficient to it.
+		retryAfterSeconds = 1
+		retryLater = true
+		createError = nil
+		return
```
sttts, line 475: `return true, 1, nil`. Avoid named return values please. The code is fragile and hard to read.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+
+	if kerrors.IsForbidden(createError) {
+		if parent == h.homePrefix {
+			// If the parent itself is the home root, which is created by default => return the error.
+			return
+		}
+
+		grandParent, parentName := parent.Split()
+		// Check in the grand parent if the parent itself exists
+		if parent, err2 := h.getClusterWorkspace(ctx, grandParent, parentName); err2 == nil {
```
sttts, line 493: err

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+			retryLater = true
+			retryAfterSeconds = h.creationDelay
+			return found, retryLater, retryAfterSeconds, checkError
+		}
+	}
+
+	if userName == kuser.APIServerUser {
+		// if request user is system:apiserver, don't wait for the RBAC resources to be setup
+		// (the system:apiserver user request is probably being creating those RBAC resources)
+		return found, retryLater, retryAfterSeconds, checkError
```
sttts, line 417: who is system:apiserver in our case?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+			result, err := clusterWorkspaceInformer.GetIndexer().ByIndex(workspaceFullPathIndexName, fullName.String())
+			if err != nil {
+				return nil, err
+			}
+			if len(result) == 0 {
+				return nil, nil
+			}
+			return result[0].(*tenancyv1alpha1.ClusterWorkspace), nil
+		},
+		searchLocalClusterRole: func(workspace logicalcluster.Name, name string) (bool, error) {
```
sttts, line 95: what is a search func? Aren't that totally normal getters?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+		}
+	}
+
+	return true, nil
+}
+
+// createHomeRBAC uses an external client to create the RBAC resources related to a Home workspace
+// Note: it only create a cluster role or cluster role binding if it doesn't exist.
+// For now it doesn't try to detect and update (or recreate) RBAC resources that might be obsolete.
+func (h *homeWorkspaceHandler) createHomeRBAC(ctx context.Context, userName string, homeWorkspace logicalcluster.Name) error {
```
sttts, line 558: did we ever think about custom authz for home workspaces? All this complexity is super ugly.

pkg/virtual/workspaces/builder/build.go
davidfestal: it's just the config passed to all VirtualWorkspaces.

Let me name it `cfg`, the same as @stevekuznetsov did in the `initializingworkspaces` VW:
```
func BuildVirtualWorkspace(
	cfg *rest.Config,
	rootPathPrefix string,
	dynamicClusterClient dynamic.ClusterInterface,
	kubeClusterClient kubernetes.ClusterInterface,
	wildcardKcpInformers kcpinformer.SharedInformerFactory,
) (map[string]framework.VirtualWorkspace, error) {
```

davidfestal: it's simply named "config" in the upper calls.

davidfestal: Fixed in commit https://github.com/kcp-dev/kcp/pull/1456/commits/001ae2f79dcd0fa6e0610661143936c0bc82898e or PR #1456

pkg/virtual/workspaces/registry/rest.go
davidfestal: Fixed in commit https://github.com/kcp-dev/kcp/pull/1456/commits/001ae2f79dcd0fa6e0610661143936c0bc82898e or PR #1456

pkg/server/home_workspaces.go
davidfestal: Not sure I understand your problem. Is it mainly naming ?

pkg/server/home_workspaces.go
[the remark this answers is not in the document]
sttts: I didn't read them, just saying :)

pkg/server/home_workspaces.go
davidfestal: We could, but I was thinking of it this as "the right to see the definition of the home workspace when it still doesn't exist".
When the home workspace has been created with the appropriate permissions to the user, he should in any case have access to it, so should also have the right to know where it is.

pkg/server/home_workspaces.go
davidfestal: We also use it when returning the `Forbidden` error in case of userName mismatch. This is why it is created a bit earlier.

pkg/server/home_workspaces.go
davidfestal: Cannot do this, since `createError` should be returned later on in some cases.

pkg/server/handler.go
davidfestal: BTW fixed in commit a1e36e31d467daafd64d6cf57f61268794951ea3

pkg/server/home_workspaces.go
davidfestal: Fixed in commit a1e36e31d467daafd64d6cf57f61268794951ea3

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+	"github.com/kcp-dev/kcp/pkg/apis/tenancy/projection"
+	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
+	tenancyv1beta1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1beta1"
+	"github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/util/conditions"
+	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	kcpexternalversions "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
+)
+
+const (
+	workspaceFullPathIndexName = "workspace-full-path"
```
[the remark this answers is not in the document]
davidfestal: I removed the indexer which was useless.
Fixed in commit  eac1f8863bdcae327ee37064046c35c46e2bc562

pkg/server/home_workspaces.go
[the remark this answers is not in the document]
davidfestal: Fixed in commit 1f37994d81146bb38fbecd97319d1a5a24eba461

Delay is in seconds.

pkg/server/home_workspaces.go
davidfestal: I removed all full name mentions and replaced with either `workspaceName` or `logicalClusterName` (or `lcluster`) in commit 1f37994d81146bb38fbecd97319d1a5a24eba461

pkg/server/home_workspaces.go
davidfestal: > use a lister, not an indexer for by-name. It will be typed, and it is what we do everywhere else.

Done in commit eac1f8863bdcae327ee37064046c35c46e2bc562

pkg/server/home_workspaces.go
davidfestal: I removed the indexer which was useless, as you pointed.
Fixed in commit  eac1f8863bdcae327ee37064046c35c46e2bc562

pkg/server/home_workspaces.go
davidfestal: This is expected to be in seconds: the number of seconds we'd ask before a retry when the workspace has to be created or waited for readiness.
I renamed it to make it explicit it is a number of seconds in commit 1f37994d81146bb38fbecd97319d1a5a24eba461

pkg/server/home_workspaces.go
davidfestal: I renamed those names to remove the confusion in commit 1f37994d81146bb38fbecd97319d1a5a24eba461

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+		homeClusterWorkspace, err := h.getLocalClusterWorkspace(homeWorkspaceFullName)
+		if err != nil {
+			responsewriters.InternalError(rw, req, err)
+			return
+		}
+		if homeClusterWorkspace != nil {
+			// The home workspace for the user exists, so let's return it to the user if he request to see its definition
+
+			homeWorkspace = &tenancyv1beta1.Workspace{}
+			projection.ProjectClusterWorkspaceToWorkspace(homeClusterWorkspace, homeWorkspace)
```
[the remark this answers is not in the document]
davidfestal: Refactored in commit 1f37994d81146bb38fbecd97319d1a5a24eba461

pkg/server/home_workspaces.go
davidfestal: Done in commit 1f37994d81146bb38fbecd97319d1a5a24eba461

pkg/server/home_workspaces.go
davidfestal: Fixed n commit 1f37994d81146bb38fbecd97319d1a5a24eba461

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+				request.WithCluster(ctx, request.Cluster{Name: tenancyv1alpha1.RootCluster}),
+				attributes); err != nil {
+				responsewriters.InternalError(rw, req, err)
+				return
+			} else if decision != authorizer.DecisionAllow {
+				responsewriters.Forbidden(ctx, attributes, rw, req, reason, homeWorkspaceCodecs)
+				return
+			}
+
+			var createError error
```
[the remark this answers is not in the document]
davidfestal: Fixed in commit 1f37994d81146bb38fbecd97319d1a5a24eba461

pkg/server/home_workspaces.go
davidfestal: remaned as `searchForReadyWorkspaceInLocalInformers` in commit 1f37994d81146bb38fbecd97319d1a5a24eba461

pkg/server/home_workspaces.go
davidfestal: Done in commit 1f37994d81146bb38fbecd97319d1a5a24eba461

pkg/server/home_workspaces.go
davidfestal: Left this return variable name, but I don't use it in the code anymore.

pkg/server/home_workspaces.go
davidfestal: Change done in commit 1f37994d81146bb38fbecd97319d1a5a24eba461

pkg/server/home_workspaces.go
davidfestal: Changed in commit 1f37994d81146bb38fbecd97319d1a5a24eba461

pkg/server/home_workspaces.go
davidfestal: removed the (quite unuseful) type here.

pkg/server/home_workspaces.go
davidfestal: Added a comment to explain this in commit https://github.com/kcp-dev/kcp/pull/1373/commits/73e74f2f2e0057639e39be70bfe58c4c9b7b0561

pkg/server/home_workspaces.go
davidfestal: Renamed and godoc'd in commit 3fa6113f512a3182c4f8712db50917fbdea6cafa

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,661 @@
...
+	// Return the error
+	return
+}
+
+// searchHomeWorkspaceRBACLocally searches for the expected RBAC resources associated to a Home workspace
+// in the local informers.
+func (h *homeWorkspaceHandler) searchHomeWorkspaceRBACLocally(homeWorkspace logicalcluster.Name, userName string) (found bool, err error) {
+	if h.mocks.searchHomeWorkspaceRBACLocally != nil {
+		return h.mocks.searchHomeWorkspaceRBACLocally(homeWorkspace, userName)
+	}
```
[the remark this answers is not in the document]
davidfestal: Removed in commit https://github.com/kcp-dev/kcp/pull/1373/commits/f235675419d83e032e6d496d13b19e48a5bdc5ab

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,707 @@
...
+func buildLocalInformersAccess(kubeSharedInformerFactory coreexternalversions.SharedInformerFactory, kcpSharedInformerFactory kcpexternalversions.SharedInformerFactory) localInformersAccess {
+	clusterWorkspaceInformer := kcpSharedInformerFactory.Tenancy().V1alpha1().ClusterWorkspaces().Informer()
+	crInformer := kubeSharedInformerFactory.Rbac().V1().ClusterRoles().Informer()
+	crbInformer := kubeSharedInformerFactory.Rbac().V1().ClusterRoleBindings().Informer()
+	clusterWorkspaceLister := kcpSharedInformerFactory.Tenancy().V1alpha1().ClusterWorkspaces().Lister()
+	crLister := kubeSharedInformerFactory.Rbac().V1().ClusterRoles().Lister()
+	crbLister := kubeSharedInformerFactory.Rbac().V1().ClusterRoleBindings().Lister()
+
+	return localInformersAccess{
+		getClusterWorkspace: func(logicalCluster logicalcluster.Name) (*tenancyv1alpha1.ClusterWorkspace, error) {
```
sttts, line 140: don't do that. Nobody expect special NotFound logic in a getter.

pkg/virtual/workspaces/registry/rest.go
sttts: This will be fixed through the soft impersonation as discussed.

pkg/server/home_workspaces.go
sttts: > so should also have the right to know where it is.

Am not convinced a user should.

pkg/server/home_workspaces.go
sttts: if these are required for correct function, you must check this and return an error

pkg/server/home_workspaces.go
sttts: check is like "doAllTheWork", it's not a sign of good factoring

pkg/server/home_workspaces.go
sttts: will go away when making home owner admin

pkg/server/home_workspaces.go
davidfestal: He can look into his local kubeconfig to get the URL.

pkg/server/home_workspaces.go
davidfestal: My plan was to test that in the related CLI options.
And I didn't want the signature to return 2 values...

An intermediate option would be to set the values to the Max value if it is upper then the max value.
And complement that with a decent check and error message in the KCP CLI options.
WDYM ?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,664 @@
...
+			parentLogicalCluster, workspaceName := logicalCluster.Split()
+			result, err := clusterWorkspaceLister.Get(clusters.ToClusterAwareKey(parentLogicalCluster, workspaceName))
```
stevekuznetsov, line 145: nit: this is just


```go
return clusterWorkspaceLister.Get(clusters.ToClusterAwareKey(parentLogicalCluster, workspaceName))
```

pkg/server/home_workspaces.go
davidfestal: Fixed in commit b631fc4caa9b4c4da8883dff9b282836eb6dcf25

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,664 @@
...
+				return nil, err
+			}
+			return result, nil
+		},
+		searchClusterRole: func(workspace logicalcluster.Name, name string) (bool, error) {
+			_, err := crLister.Get(clusters.ToClusterAwareKey(workspace, name))
+			if kerrors.IsNotFound(err) {
+				return false, nil
+			}
+			if err != nil {
```
stevekuznetsov, line 156: nit: this is just `return err == nil, err`

pkg/server/home_workspaces.go
davidfestal: Done in commit 0cb62c3aa30411ef67a7350b9ab85dec02c3e210

pkg/server/home_workspaces.go
davidfestal: Done in commit 0cb62c3aa30411ef67a7350b9ab85dec02c3e210

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,664 @@
...
+			_, err := crbLister.Get(clusters.ToClusterAwareKey(workspace, name))
+			if kerrors.IsNotFound(err) {
+				return false, nil
+			}
+			if err != nil {
+				return false, err
+			}
+			return true, nil
+		},
+		synced: func() bool {
```
stevekuznetsov, line 171: So in the case that we misconfigure the informers and this never syncs, how can an admin tell?

stevekuznetsov: Alternatively - can we add a `WaitForNamedCacheSync` to server startup for these somehow?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,664 @@
...
+	bucketLevels, bucketSize int
+	creationDelaySeconds     int
+
+	authz authorizer.Authorizer
+
+	kcp            externalKubeClientsAccess
+	localInformers localInformersAccess
+}
+
+type homeWorkspaceFeatureLogic struct {
```
stevekuznetsov, line 193: Why is this a struct instead of an interface? What is the utility of injecting each of these separately? What are the cases where we inject some but not all? Are there dependencies between choices for injection here?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,664 @@
...
+		responsewriters.InternalError(rw, req, err)
+		return
+	}
+	if sets.NewString(user.GetGroups()...).Has(kuser.SystemPrivilegedGroup) {
+		// If we are the system privileged group, it might be a call from the virtual workspace
+		// in which case we also search the user in the soft impersonation header.
+		if impersonated, err := softimpersonation.UserInfoFromRequestHeader(req); err != nil {
+			responsewriters.InternalError(rw, req, err)
+			return
+		} else if impersonated != nil {
```
stevekuznetsov, line 250: This doesn't handle the case where `impersonated ==nil && err == nil`

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,664 @@
...
+			user = impersonated
+		}
+	}
+	requestInfo, ok := request.RequestInfoFrom(ctx)
+	if !ok {
+		responsewriters.InternalError(rw, req, errors.New("No request Info"))
```
stevekuznetsov, line 256: Are we sending back `metav1.Status` for these errors?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,664 @@
...
+			// We don't need to check any permission before returning the home workspace definition since,
+			// once it has been created, a home workspace is owned by the user.
+
+			homeWorkspace := &tenancyv1beta1.Workspace{}
+			projection.ProjectClusterWorkspaceToWorkspace(homeClusterWorkspace, homeWorkspace)
+			responsewriters.WriteObjectNegotiated(homeWorkspaceCodecs, negotiation.DefaultEndpointRestrictions, tenancyv1beta1.SchemeGroupVersion, rw, req, http.StatusOK, homeWorkspace)
+			return
+		}
+
+		// Test if the user has the right to see his Home workspace even though it doesn't exists
```
stevekuznetsov, line 288: If the user always has a right to see their home workspace, once it exists, what does it mean to check their permissions here? In what case would this check fail?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,664 @@
...
+		}
+		if decision != authorizer.DecisionAllow {
+			responsewriters.Forbidden(ctx, attributes, rw, req, reason, homeWorkspaceCodecs)
+			return
+		}
+
+		// TODO: The way we build this URL here is not fully compatible with sharding long-term:
+		// - short term (what is done now): use genericConfig.ExternalAddress
+		// - medium term: shards will know their name eventually, and with that we can lookup the external URL of the shard
+		// - long-term: create home workspace already on `~` access in order to make scheduling happen and then we know the real external URL.
```
stevekuznetsov, line 305: So what creates this today, if it's not this code?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,664 @@
...
+	}
+
+	needsAutomaticCreation, workspaceType := h.needsAutomaticCreation(lcluster.Name)
+	if !needsAutomaticCreation {
+		h.apiHandler.ServeHTTP(rw, req)
+		return
+	}
+
+	isHome := workspaceType == "Home"
+	if foundLocally, retryAfterSeconds, err := h.searchForReadyWorkspaceInLocalInformers(lcluster.Name, isHome, user.GetName()); foundLocally {
```
stevekuznetsov, line 336: Technically since you're making every method here a field, they can all be `nil` ... unclear how to handle that but we might be making life hard for ourselves

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,664 @@
...
+	if !logicalClusterName.HasPrefix(h.homePrefix) {
+		return false, ""
+	}
+
+	if logicalClusterName == h.homePrefix {
+		return false, ""
+	}
+
+	levelsToHomePrefix := 0
+	for lcluster := logicalClusterName; lcluster != h.homePrefix; lcluster, _ = lcluster.Split() {
```
stevekuznetsov, line 439: a) couldn't this just count the number of `:` in the name?
b) seems like a useful helper for the `logicalcluster` library

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,664 @@
...
+		levelsToHomePrefix++
+	}
+
+	switch {
+	case levelsToHomePrefix <= h.bucketLevels:
+		return true, "Homebucket"
```
stevekuznetsov, line 445: Should these be constants?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,664 @@
...
+	attributes := homeWorkspaceAuthorizerAttributes(user, "create")
+
+	if isHome && lcluster.Name != h.getHomeLogicalClusterName(user.GetName()) {
+		// If we're checking a home workspace or home bucket workspace, but not of the consistent user, let's refuse.
+		klog.Errorf("failed to authorize user %q to create a home workspace %q: home workspace can only be created by the user of the home workspace", user.GetName(), lcluster.Name)
+		responsewriters.Forbidden(ctx, attributes, rw, req, authorization.WorkspaceAcccessNotPermittedReason, homeWorkspaceCodecs)
+		return
+	}
+
+	// Test if the user has the right to create his Home workspace when it doesn't exists
```
stevekuznetsov, line 363: When is this not true ?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,664 @@
...
+	// => test the create verb on the clusterworkspaces/workspace subresource named ~ in the root workspace.
+
+	if decision, reason, err := h.authz.Authorize(
+		request.WithCluster(ctx, request.Cluster{Name: tenancyv1alpha1.RootCluster}),
+		attributes); err != nil {
+		responsewriters.InternalError(rw, req, err)
```
stevekuznetsov, line 369: What kinds of errors will this contain? Does it leak information?

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,664 @@
...
+		http.Error(rw, "Creating the home workspace", http.StatusTooManyRequests)
+		return
+	}
+}
+
+// reHomeWorkspaceNameDisallowedtChars is the regexp that defines what characters
+// are disallowed in a home workspace name.
+// Home workspace name is derived from the user name, with disallowed characters
+// replaced.
+var reHomeWorkspaceNameDisallowedtChars = regexp.MustCompile("[^a-z0-9-]")
```
stevekuznetsov, line 391:
```suggestion
var reHomeWorkspaceNameDisallowedChars = regexp.MustCompile("[^a-z0-9-]")
```

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,664 @@
...
+
+// getHomeLogicalClusterName returns the logicalcluster name of the home workspace for a given user
+// The home workspace logical cluster ancestors are home bucket workspaces whose name is based
+// on the user name sha1 hash.
+func (h *homeWorkspaceHandler) getHomeLogicalClusterName(userName string) logicalcluster.Name {
+	// bucketLevels <= 5
```
stevekuznetsov, line 397: What's the meaning of configuring these values? When would an admin choose a different value for either the levels or size? What implication would it have? Does an admin have the right information to choose this? Can it change over time?

pkg/server/home_workspaces.go
davidfestal: Right, I removed a condition in a previous fix and missed the possible simplification

pkg/server/home_workspaces.go
davidfestal: That's expected, and in this case we don't want the user to be changed

pkg/server/home_workspaces.go
davidfestal: What would it bring ?
Why would it be more required here than [there](https://github.com/kcp-dev/kcp/blob/main/pkg/proxy/handler.go#L47) for example ?

pkg/server/home_workspaces.go
davidfestal: We're precisely in the case when the home workspace doesn't exist yet.
We manage a dedicated permission for:

*being able to see where the home workspace will be (== what it's logical cluster will be) even _**before it exists**_*

When the Home Workspace doesn't not exist, we want to gate the ability:
- to know where the Home workspace will be
- to create the HomeWorkspace automatically on first access.

The 2 are granted by default for authenticated users.

See this already-merged PR for more details: https://github.com/kcp-dev/kcp/pull/1455/files#diff-e4317fcacf05b074783ba5e267a2d0fafd48d0069d9fb8ed0e453b39f3bbb2a8

pkg/server/home_workspaces.go
davidfestal: see answer to comment https://github.com/kcp-dev/kcp/pull/1373/files/0faa774d210b4af08718c56528ea795db4223091#r916265720

pkg/server/home_workspaces.go
davidfestal: Fixed in commit ae9862f8e0de72cb7366e5289b4991dda8a1f72e

pkg/server/home_workspaces.go
davidfestal: To avoid any information leak, I replaced the internal error by a Forbidden error with the typical generic message
Done in commit ae9862f8e0de72cb7366e5289b4991dda8a1f72e

pkg/server/home_workspaces.go
davidfestal: If you create the handler through `homeWorkspaceHandlerBuilder.build()`, which is the way we use, you will have all the fields of the `homeWorkspaceFeatureLogic` struct filled with the implementation, unless you explicitly reset one field to nil. 

Only some unit tests reset some of the function fields to overridden mocks in order to make cut the dependency between tested functions.

pkg/server/home_workspaces.go
davidfestal: > Why is this a struct instead of an interface?

This struct contains the functions we want to be possibly overridden in unit tests, in order to cut dependency between functions, and have 1 test == one function.

> What is the utility of injecting each of these separately?

They are all injected as a whole in the `homeWorkspaceHandlerBuilder.build()` method.

> What are the cases where we inject some but not all?

In tests there are cases where we want to override one but not the others

More generally, this allows easier unit testing of individual methods, and is already the result of answering @sttts PR comment after a first try.
IMHO it allows combining the the easiness of testing with the readability of the main handler logic.

However if you have a better approach, contributions are welcomed ... in a followup PR if possible ;-)

pkg/server/home_workspaces.go
davidfestal: see this comment answer for more context: https://github.com/kcp-dev/kcp/pull/1373#discussion_r916303112

pkg/server/handler.go
```diff
@@ -209,6 +211,15 @@ func WithWorkspaceProjection(apiHandler http.Handler) http.HandlerFunc {
...
```
sttts, line 214: the flags come in follow-up? (if yes, add a follow-up task to the epic please).

pkg/server/home_workspaces.go
sttts: `WaitForNamedCacheSync` is already there. This check here is because the server serves before all informers are up. It's the same behaviour many admission plugins and also authz have.

pkg/server/home_workspaces.go
sttts: all this is implementation detail of this package, not an interface to the outside. So I think either way is fine. If this was an API a interface would be natural.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,656 @@
...
+		return
+	}
+	if sets.NewString(user.GetGroups()...).Has(kuser.SystemPrivilegedGroup) {
+		// If we are the system privileged group, it might be a call from the virtual workspace
+		// in which case we also search the user in the soft impersonation header.
+		if impersonated, err := softimpersonation.UserInfoFromRequestHeader(req); err != nil {
+			responsewriters.InternalError(rw, req, err)
+			return
+		} else if impersonated != nil {
+			user = impersonated
```
sttts, line 241: nit: would call this `effectiveUser`.

pkg/server/home_workspaces.go
sttts: this is used everywhere in kube in similar situations. But I agree that a Status would be cleaner. I guess nobody cares because this is for cases that never happen (only when plumbing is wrong).

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,656 @@
...
+			return
+		}
+		if homeClusterWorkspace != nil {
+			// The home workspace for the user exists, so let's return it to the user if he requests to see its definition.
+			// We don't need to check any permission before returning the home workspace definition since,
+			// once it has been created, a home workspace is owned by the user.
+
+			homeWorkspace := &tenancyv1beta1.Workspace{}
+			projection.ProjectClusterWorkspaceToWorkspace(homeClusterWorkspace, homeWorkspace)
+			responsewriters.WriteObjectNegotiated(homeWorkspaceCodecs, negotiation.DefaultEndpointRestrictions, tenancyv1beta1.SchemeGroupVersion, rw, req, http.StatusOK, homeWorkspace)
```
sttts, line 274: there is no authz here, correct?

pkg/server/home_workspaces.go
sttts: I think the gate here is fine and expected. I wonder though whether we want additional authz for the other case as well, not in the sense of gating, but just to ensure it is really the workspace of this user and nobody else due to a conflict.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,656 @@
...
+			return
+		}
+
+		// Test if the user has the right to see his Home workspace even though it doesn't exists
+		// => test the get verb on the clusterworkspaces/workspace subresource named ~ in the root workspace.
+
+		attributes := homeWorkspaceAuthorizerAttributes(user, "get")
+		decision, reason, err := h.authz.Authorize(ctx, attributes)
+		if err != nil {
+			klog.Errorf("failed to authorize user %q to get a home workspace %q: %w", user.GetName(), lcluster.Name, err)
```
sttts, line 284: unconditional klogs in the handler chain are always questionable. Better call `runtime.HandleError` as that throttles the output.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,656 @@
...
+				ClusterName: parent.String(),
+			},
+			Spec: tenancyv1beta1.WorkspaceSpec{
+				Type: tenancyv1alpha1.ClusterWorkspaceTypeReference{
+					Name: tenancyv1alpha1.ClusterWorkspaceTypeName("Home"),
+					Path: tenancyv1alpha1.RootCluster.String(),
+				},
+			},
+			Status: tenancyv1beta1.WorkspaceStatus{
+				URL: homeWorkspaceURL.String(),
```
sttts, line 312: what's the phase? 🤡

pkg/server/home_workspaces.go
sttts: Assuming non-nil is fine and totally common in Golang. I wished Golang had a non-nil-pointer type, but not in this life.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,656 @@
...
+	if foundLocally, retryAfterSeconds, err := h.searchForReadyWorkspaceInLocalInformers(lcluster.Name, isHome, user.GetName()); foundLocally {
+		if retryAfterSeconds > 0 {
+			// Return a 429 status asking the client to try again after the creationDelay
+			rw.Header().Set("Retry-After", fmt.Sprintf("%d", retryAfterSeconds))
+			http.Error(rw, "Creating the home workspace", http.StatusTooManyRequests)
+			return
+		}
+		h.apiHandler.ServeHTTP(rw, req)
+		return
+	} else if err != nil {
```
sttts, line 336: (nit, style): handle error case first. Easier read flow.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,656 @@
...
+		responsewriters.Forbidden(ctx, attributes, rw, req, authorization.WorkspaceAcccessNotPermittedReason, homeWorkspaceCodecs)
+		return
+	}
+
+	// Test if the user has the right to create his Home workspace when it doesn't exists
+	// => test the create verb on the clusterworkspaces/workspace subresource named ~ in the root workspace.
+
+	if decision, reason, err := h.authz.Authorize(
+		request.WithCluster(ctx, request.Cluster{Name: tenancyv1alpha1.RootCluster}),
+		attributes); err != nil {
```
sttts, line 359: (nit, style): put a comma and newline after attributes. Give better flow, easier to read.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,656 @@
...
+//
+// When the Home workspace is still not Ready, the handler returns a Retry-After response with a delay in seconds that is configurable
+// (creationDelaySeconds), so that client-go clients will automatically retry the request after this delay.
+//
+// - homePrefix is the workspace that will contains all the user home workspaces, partitioned by bucket workspaces
+// - bucketLevels is the number of bucket workspaces met before reaching a home workspace from the homePefix workspace
+// - bucketSize is the number of chars comprising each bucket.
+//
+// Bucket workspace names are calculated based on the user name hash.
+func WithHomeWorkspaces(apiHandler http.Handler, a authorizer.Authorizer, kubeClusterClient kubernetes.ClusterInterface, kcpClusterClient kcpclient.ClusterInterface, kubeSharedInformerFactory coreexternalversions.SharedInformerFactory, kcpSharedInformerFactory kcpexternalversions.SharedInformerFactory, externalHost string, creationDelaySeconds int, homePrefix logicalcluster.Name, bucketLevels, bucketSize int) http.Handler {
```
sttts, line 86: hard to read, make it a multiline signature

pkg/server/home_workspaces.go
sttts: please move these restriction to the constructor doc if it's not already the case, and validate there.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,656 @@
...
+		bucketCharInteger := binary.BigEndian.Uint32(bytes[bucketBytesStart : bucketBytesStart+4])
+		for bucketCharIndex := 0; bucketCharIndex < h.bucketSize; bucketCharIndex++ {
+			bucketChar := byte('a') + byte(bucketCharInteger%26)
+			bucketBytes[bucketCharIndex] = bucketChar
+			bucketCharInteger /= 26
+		}
+		result = result.Join(string(bucketBytes))
+	}
+
+	userName = reHomeWorkspaceNameDisallowedChars.ReplaceAllLiteralString(userName, "-")
```
sttts, line 406: this risks name conflicts of users.1
1. do we handle conflict securely? It is maybe ok to accept conflict and 0.0001% of users get broken. But it would not be ok to give two users access to the same workspace. Storing the full username on the workspace object as annotation could help.
2. any sensible way to deconflict? Is it worth it?

pkg/server/home_workspaces.go
sttts: 🔴

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,656 @@
...
+	})
+	userName = strings.TrimRightFunc(userName, func(r rune) bool {
+		return r == '-'
+	})
+
+	return result.Join(userName)
+}
+
+// needsAutomaticCreation deduces, from the logical cluster name,
+// according the the expected home root and home bucket level number,
```
sttts, line 418:
```suggestion
// according to the expected home root and home bucket level number,
```

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,656 @@
...
+	}
+}
+
+// searchForReadyWorkspaceInLocalInformers checks whether a workspace is known on the current shard
+// in local informers.
+// For home workspaces, we also check:
+//   - if related RBAC resources are there,
+//   - if the workspace phase is READY
+// and if not answer to retry later.
+func searchForReadyWorkspaceInLocalInformers(h *homeWorkspaceHandler, logicalClusterName logicalcluster.Name, isHome bool, userName string) (found bool, retryAfterSeconds int, err error) {
```
sttts, line 451: found is actually "foundAndReadyAndWithRBAC" or "foundAndComplete"

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,656 @@
...
+	if err != nil && !kerrors.IsNotFound(err) {
+		return false, 0, err
+	}
+	if workspace == nil {
+		return false, 0, nil
+	}
+
+	// Workspace has been found in local informer: check its status.
+	if workspaceUnschedulable(workspace) {
+		// The workspace cannot be scheduled => return the error.
```
sttts, line 462: useless comment

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,656 @@
...
+	}
+
+	if workspace.Status.Phase != tenancyv1alpha1.ClusterWorkspacePhaseReady {
+		// We have to wait for the workspace to be Ready before allowing actions in it,
+		// but only for the the home workspaces.
+		// Waiting for home buckets to be ready is done in the `tryToCreate` function.
+
+		// TODO (david): For now we accept all requests in this workspace when it's initializing...
+		// TODO (david): In the future, we might have a way to identify requests that are done through the
+		// initializer virtual workspace
```
sttts, line 478: the users have the right to look into their initializing workspaces, and fix initialization if necessary. So we have to allow access. The initializing VW is just another access point for non-owners of the workspace.

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,656 @@
...
+
+	if err == nil || kerrors.IsAlreadyExists(err) {
+		if workspaceType == "Home" {
+			if err := h.createHomeWorkspaceRBACResources(ctx, userName, logicalClusterName); err != nil {
+				return 0, err
+			}
+		}
+		// Retry sooner than the creation delay, because it's probably a question of
+		// letting the local informer cache being updated.
+		// Retrying quicly (after 1 second) should be sufficient to it.
```
sttts, line 525:
```suggestion
		// Retrying quickly (after 1 second) should be sufficient to it.
```

pkg/server/home_workspaces.go
```diff
@@ -0,0 +1,656 @@
...
+	clusterWorkspaceLister := kcpSharedInformerFactory.Tenancy().V1alpha1().ClusterWorkspaces().Lister()
+	crLister := kubeSharedInformerFactory.Rbac().V1().ClusterRoles().Lister()
+	crbLister := kubeSharedInformerFactory.Rbac().V1().ClusterRoleBindings().Lister()
+
+	return localInformersAccess{
+		getClusterWorkspace: func(logicalCluster logicalcluster.Name) (*tenancyv1alpha1.ClusterWorkspace, error) {
+			parentLogicalCluster, workspaceName := logicalCluster.Split()
+			return clusterWorkspaceLister.Get(clusters.ToClusterAwareKey(parentLogicalCluster, workspaceName))
+		},
+		searchClusterRole: func(workspace logicalcluster.Name, name string) (bool, error) {
```
sttts, line 147: am still confused about these function name. Make them normal informer getters. Would be one (two) surprise(s) less.

sttts: done

pkg/server/home_workspaces.go
sttts: done

pkg/server/home_workspaces.go
sttts: ok, I think this is fine. It is not about workspace access (content), but only about seeing the object.

pkg/server/home_workspaces.go
sttts: done

pkg/server/home_workspaces.go
sttts: done

pkg/server/home_workspaces.go
sttts: done

pkg/server/home_workspaces.go
sttts: done

pkg/server/home_workspaces.go
sttts: done 1

pkg/server/home_workspaces.go
sttts: have only seen this occurence of that need. If we find 3 others let's talk about `logicalcluster.Name.Levels() int`

pkg/server/home_workspaces.go
sttts: done

pkg/server/home_workspaces.go
sttts: done

pkg/server/home_workspaces.go
sttts: done

pkg/server/home_workspaces.go
```diff
@@ -83,7 +85,22 @@ func init() {
...
+	a authorizer.Authorizer,
+	kubeClusterClient kubernetes.ClusterInterface,
+	kcpClusterClient kcpclient.ClusterInterface,
+	kubeSharedInformerFactory coreexternalversions.SharedInformerFactory,
+	kcpSharedInformerFactory kcpexternalversions.SharedInformerFactory,
+	externalHost string,
+	creationDelaySeconds int,
+	homePrefix logicalcluster.Name,
+	bucketLevels,
+	bucketSize int,
```
s-urbaniak, line 99: nit: not all parameters are documented, also let's document the rationale of limiting bucketLevels/bucketSize.

sttts: @davidfestal follow-up

sttts: added to https://github.com/kcp-dev/kcp/issues/1069

sttts:
/retest

Placement flake.

sttts:
TestSyncerVirtualWorkspace flake

/retest

sttts:
/test e2e-multiple-runs

Prow is drunk.

sttts:
e2e-multiple-runs fails on artifacts collection it seems. Fails after PASS 😞

pkg/server/home_workspaces.go
stevekuznetsov: @davidfestal @sttts I can count on many hands the number of times that making such an assumption ends up being incorrect. Prow ends up with panics and we have a service outage. Not fun. Golang doesn't have constructors, @davidfestal , so there's no typesafe way to ensure this.

pkg/server/home_workspaces.go
stevekuznetsov: @davidfestal @sttts if this cannot change over time we need to be EXTREMELY clear and careful so admins choose a correct value here.

Issue #1361, opened 2022-06-27
virtual/workspaces: authorizer not implemented


sttts:
The virtual workspace apiserver for workspaces logs:
```
E0627 09:52:20.737738   38729 build.go:94] the authorizer for the 'workdspaces' virtual workspace is not implemented !
```

ncdc:
Move the SARs inside the workspaces virtual workspace code and make them an authorizer

PR #1374, opened 2022-06-27
Workspaces VW: Fix obsolete permission management...

davidfestal:
## Summary

Fix obsolete permission management in the `workspaces` virtual workspace.

#### Current situation

The `workspaces` virtual workspace is still based on the old approach where user workspaces used to be created in top-level organization workspaces.
It checks permissions against the `clusterworkspaces/content` resource named with the name of the current workspace (typically an organization) in the parent workspace of the current organization workspace.

- for  `get`, `list` and `watch`, it checks the `access` verb
- for `create`, it checks the `member` verb

Additionally to delete a workspace it also checks the `delete` verb on the `clusterworkspaces/workspace` resource of the current organization workspace.

This is both incompatible with:
- the new approach that non-admin user workspaces should not be created in top-level orgs in the future 
- the Home workspaces that can be created on-the-fly (as well as their home bucket parent workspaces) when first accessed. 

#### Implemented changes

In the case of workspaces that are children of a top-level organization, we keep the same mechanism as before, to keep compatibility for a while for user workspaces already created in top-level organizations) 

In all other cases, SARs in the `workspaces` virtual workspace are now always **checked against the `clusterworkspaces/workspace` resource in the current workspace** (not the parent anymore).

- for `get`, `list`, `watch` requests, it checks the `get` verb
- for `create` requests, it checks the `create` verb
- for `delete` requests, it checks the `get` verb, as well as the `delete` verb on resource name of the child workspace which is being deleted.

Due to the KCP authorizer architecture, these SAR will also check the `access` verb on the `clusterworkspaces/content` resource in the parent workspaces (through the `workspaceContentAuthorizer`) as well as basic membership of the top-level org through the `topLevelOrgAccessAuthorizer`. So we don't loose any security checks previously done. 

## Related issue(s)

Prerequisite PR for the [Home workspace EPIC](#1069)

### DIFF
--- a/pkg/virtual/workspaces/registry/rest.go
+++ b/pkg/virtual/workspaces/registry/rest.go
@@ -258,8 +258,55 @@ func (s *REST) authorizeOrgForUser(ctx context.Context, orgClusterName logicalcl
 	return nil
 }
 
+func (s *REST) authorizeForUser(ctx context.Context, orgClusterName logicalcluster.Name, user user.Info, verb string, resourceName string) error {
+	// Root org access is implicit for every user. For non-root orgs, we need to check for
+	// verb permissions against the clusterworkspaces/workspace sub-resource.
+	if orgClusterName == tenancyv1alpha1.RootCluster || sets.NewString(user.GetGroups()...).Has("system:masters") {
+		return nil
+	}
+
+	if parent, _ := orgClusterName.Split(); parent == tenancyv1alpha1.RootCluster {
+		if verb == "get" {
+			return s.authorizeOrgForUser(ctx, orgClusterName, user, "access")
+		}
+		if verb == "create" {
+			return s.authorizeOrgForUser(ctx, orgClusterName, user, "member")
+		}
+	}
+
+	// check for <verb> permission on the ClusterWorkspace workspace subresource for the <resourceName>
+	authz, err := s.delegatedAuthz(orgClusterName, s.kubeClusterClient)
+	if err != nil {
+		klog.Errorf("failed to get delegated authorizer for logical cluster %s", user, orgClusterName)
+		return kerrors.NewForbidden(tenancyv1beta1.Resource("workspaces"), resourceName, fmt.Errorf("%q workspace %q in workspace %q is not allowed", verb, resourceName, orgClusterName))
+	}
+	deleteWorkspaceAttr := authorizer.AttributesRecord{
+		User:            user,
+		Verb:            verb,
+		APIGroup:        tenancyv1alpha1.SchemeGroupVersion.Group,
+		APIVersion:      tenancyv1alpha1.SchemeGroupVersion.Version,
+		Resource:        "clusterworkspaces",
+		Subresource:     "workspace",
+		Name:            resourceName,
+		ResourceRequest: true,
+	}
+	if decision, reason, err := authz.Authorize(ctx, deleteWorkspaceAttr); err != nil {
+		klog.Errorf("failed to authorize user %q to %q clusterworkspaces/workspace name %q in %s", user.GetName(), verb, resourceName, orgClusterName)
+		return kerrors.NewForbidden(tenancyv1beta1.Resource("workspaces"), "", fmt.Errorf("%q workspace %q in workspace %q is not allowed", verb, resourceName, orgClusterName))
+	} else if decision != authorizer.DecisionAllow {
+		klog.Errorf("user %q lacks (%s) clusterworkspaces/workspace %s permission for %q in %s: %s", user.GetName(), decisions[decision], verb, resourceName, orgClusterName, reason)
+		return kerrors.NewForbidden(tenancyv1beta1.Resource("workspaces"), resourceName, fmt.Errorf("%q workspace %q in workspace %q is not allowed", verb, resourceName, orgClusterName))
+	}
+
+	return nil
+}
+
 func shouldUsePersonalScope(scope string, orgClusterName logicalcluster.Name) bool {
-	return scope == PersonalScope && orgClusterName != tenancyv1alpha1.RootCluster
+	parent, _ := orgClusterName.Split()
+
+	// Let's use personal scopes only in the old case when personal user workspaces used to be created
+	// in top-level organizations
+	return scope == PersonalScope && orgClusterName != tenancyv1alpha1.RootCluster && parent == tenancyv1alpha1.RootCluster
 }
 
 // List retrieves a list of Workspaces that match label.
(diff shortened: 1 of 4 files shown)


sttts:
What does fix mean? Remove? What changes?

sttts:
Also a fix without a changed/extended test is odd.

davidfestal:
> What does fix mean? Remove? What changes?

I updated the description

sttts:
> for get, list, watch requests, it checks the get verb

why not check for get, list, watch verbs?

sttts:
> for delete requests, it checks the get verb, as well as the delete verb on resource name of the child workspace which is being deleted.

why also get?

sttts:
> In the case of workspaces that are children of a top-level organization, we keep the same mechanism as before, to keep compatibility for a while for user workspaces already created in top-level organizations)

Does both work, old and new? Is there an issue for follow-ups?

davidfestal:
@sttts Those choices reflect the same level of checks that was implemented previously, but with the new permissions scheme.

> > for get, list, watch requests, it checks the get verb
> 
> why not check for get, list, watch verbs?

Those 3 actions were just checking for `access` permission to the workspace content. Nothing more. Checking the `get` verb on the `clusterworkspaces/workspace` seems what is the most similar.
But sure I can come back to something more standard, and just check the verb that corresponds to the action.

> 
> > for delete requests, it checks the get verb, as well as the delete verb on resource name of the child workspace which is being deleted.
> 
> why also get?

Here as well, generic get permission to `clusterworkspaces/workspace` in the current workspace was tested at the very start (similar to testing the `access` to the parent in the previous model), even before trying to get the workspace internal name. Then the `delete` is tested against the named workspace that has to be deleted.
But I agree that it can probably be simplified.

davidfestal:
> > In the case of workspaces that are children of a top-level organization, we keep the same mechanism as before, to keep compatibility for a while for user workspaces already created in top-level organizations)
> 
> Does both work, old and new? Is there an issue for follow-ups?

Yes, both work for now, and a followup issue exists here: https://github.com/kcp-dev/kcp/issues/1388

davidfestal:
> We should also think what default permissions are (folllow-up). I think reading workspaces is probably fine for everybody.

Followup issue created here: https://github.com/kcp-dev/kcp/issues/1389

pkg/virtual/workspaces/registry/rest.go
```diff
@@ -257,8 +257,55 @@ func (s *REST) authorizeOrgForUser(ctx context.Context, orgClusterName logicalcl
 	return nil
 }
 
+func (s *REST) authorizeForUser(ctx context.Context, orgClusterName logicalcluster.Name, user user.Info, verb string, resourceName string) error {
+	// Root org access is implicit for every user. For non-root orgs, we need to check for
+	// verb permissions against the clusterworkspaces/workspace sub-resource.
+	if orgClusterName == tenancyv1alpha1.RootCluster || sets.NewString(user.GetGroups()...).Has("system:masters") {
```
s-urbaniak, line 263: please use the https://github.com/kubernetes/kubernetes/blob/ab4aa182a5808c9e61413fadc051d99d6572b0ab/staging/src/k8s.io/apiserver/pkg/authentication/user/user.go#L71 constant

pkg/virtual/workspaces/registry/rest.go
```diff
@@ -257,8 +257,55 @@ func (s *REST) authorizeOrgForUser(ctx context.Context, orgClusterName logicalcl
 	return nil
 }
 
+func (s *REST) authorizeForUser(ctx context.Context, orgClusterName logicalcluster.Name, user user.Info, verb string, resourceName string) error {
```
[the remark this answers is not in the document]
davidfestal: I agree. This would be done in a followup issue: https://github.com/kcp-dev/kcp/issues/1361

pkg/virtual/workspaces/registry/rest.go
davidfestal: Fixed in commit cd4ae7e7873e38fa9d93b0513c7bd763f070894b

sttts:
I think we are on the same page after Slack, but just for reference:

> Those 3 actions were just checking for access permission to the workspace content. Nothing more. Checking the get verb on the clusterworkspaces/workspace seems what is the most similar.
> But sure I can come back to something more standard, and just check the verb that corresponds to the action.

access is a virtual verb on a virtual resource. Workspace is a projection (like scale) that we happen to project as non-sub-resource. So the later is not virtual from the point of view of the user. Hence, I think get, list, watch etc. verbs seem to be right.

davidfestal:
> I think we are on the same page after Slack, but just for reference:
> 
> > Those 3 actions were just checking for access permission to the workspace content. Nothing more. Checking the get verb on the clusterworkspaces/workspace seems what is the most similar.
> > But sure I can come back to something more standard, and just check the verb that corresponds to the action.
> 
> access is a virtual verb on a virtual resource. Workspace is a projection (like scale) that we happen to project as non-sub-resource. So the later is not virtual from the point of view of the user. Hence, I think get, list, watch etc. verbs seem to be right.

Fixed in commit 59aae901eaa13a9735a14cad1451d059676eb588

pkg/virtual/workspaces/registry/rest.go
```diff
@@ -258,8 +258,61 @@ func (s *REST) authorizeOrgForUser(ctx context.Context, orgClusterName logicalcl
...
+	if parent, _ := orgClusterName.Split(); parent == tenancyv1alpha1.RootCluster {
+		switch verb {
+		case "get", "watch", "list":
+			return s.deprecatedAuthorizeOrgForUser(ctx, orgClusterName, user, "access")
+		case "create":
+			return s.deprecatedAuthorizeOrgForUser(ctx, orgClusterName, user, "member")
+		case "delete":
+			if err := s.deprecatedAuthorizeOrgForUser(ctx, orgClusterName, user, "access"); err != nil {
+				return err
+			}
```
s-urbaniak, line 279: what about `update` and `patch` verbs?

davidfestal: For now we don't support `update` or `patch` workspaces through the `workspaces` virtual workspace.

pkg/virtual/workspaces/registry/rest.go
```diff
@@ -265,12 +265,18 @@ func (s *REST) authorizeForUser(ctx context.Context, orgClusterName logicalclust
...
-		}
-		if verb == "create" {
-			return s.authorizeOrgForUser(ctx, orgClusterName, user, "member")
+		switch verb {
+		case "get", "watch", "list":
+			return s.deprecatedAuthorizeOrgForUser(ctx, orgClusterName, user, "access")
+		case "create":
+			return s.deprecatedAuthorizeOrgForUser(ctx, orgClusterName, user, "member")
+		case "delete":
+			if err := s.deprecatedAuthorizeOrgForUser(ctx, orgClusterName, user, "access"); err != nil {
```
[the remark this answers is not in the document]
davidfestal: Intentional indeed: In the old model we should both check for access/content in the parent, and delete/workspace on the current workspace.

Let me comment.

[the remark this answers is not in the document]
davidfestal: Fixed in commit 4a61344b263670e7841203d63197b8506d8a47a6

pkg/virtual/workspaces/registry/rest.go
davidfestal: Fixed in commit c05889057dc7234922ebf7b016a036c20e0b39a3

sttts:
Based on @s-urbaniak 's previous lgtm and the last fixes

/lgtm

PR #1376, opened 2022-06-28
Wrap clients

varshaprasad96:
## Summary
This PR modifies the cluster client calls to accept a scoped context and wrap an existing config with the cluster round tripper. 
The first commit modifies the calls for `apiexport_controller`. Since a few helpers need to be modified too, the second commit only makes modifications in files which are affected by change in method signatures for helpers.

To summarize, the following changes are needed:
1. Wrap existing config with clusterroundtripper.
2. Scope context in all client calls.
3. Since discovery clients create a new context internally, and an external context cannot be passed, it requires modifying the host of the rest config.


pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -172,11 +173,11 @@ func NewController(
 type controller struct {
 	queue workqueue.RateLimitingInterface
 
-	kcpClusterClient kcpclient.ClusterInterface
+	kcpClusterClient kcpclient.Interface
```
sttts, line 176: now that the ClusterInterface type goes away we need convention even more about the variable names. Do we agree that we always call them `*ClusterClient` and never ever `*Client`, and not even `*Clients`?

varshaprasad96:
cc: @ncdc @sttts @stevekuznetsov The first commit contains all the changes required to make scoped client calls in apiexport_controller. If the changes sounds reasonable, then we can proceed with modifying all the client calls in kcp. This PR is just an example to test if things work well with our approach.

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -43,6 +43,7 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 
+	kcp "github.com/kcp-dev/apimachinery/pkg/client"
```
sttts, line 46: is this convention to use `kcp` as alias? Would be fine with me. We just should be super consistent. Or maybe `kcpclient`?

test/e2e/apibinding/apibinding_authorizer_test.go
```diff
@@ -57,23 +60,27 @@ func TestAPIBindingAuthorizer(t *testing.T) {
...
-	kcpClients, err := clientset.NewClusterForConfig(cfg)
+	kcpClients, err := clientset.NewForConfig(cfg)
 	require.NoError(t, err, "failed to construct kcp cluster client for server")
 
-	dynamicClients, err := dynamic.NewClusterForConfig(cfg)
+	dynamicClients, err := kcpdynamic.NewClusterDynamicClientForConfig(cfgCopy)
 	require.NoError(t, err, "failed to construct dynamic cluster client for server")
 
-	kubeClusterClient, err := kubernetes.NewClusterForConfig(cfg)
+	kubeclient, err := kubernetes.NewForConfig(cfg)
```
sttts, line 74: please kubeClusterClient.

test/e2e/apibinding/apibinding_authorizer_test.go
```diff
@@ -92,19 +99,24 @@ func TestAPIBindingAuthorizer(t *testing.T) {
 			},
 		}
 
-		_, err = kcpClients.Cluster(consumerWorkspace).ApisV1alpha1().APIBindings().Create(ctx, apiBinding, metav1.CreateOptions{})
+		_, err = kcpClients.ApisV1alpha1().APIBindings().Create(kcp.WithCluster(ctx, consumerWorkspace), apiBinding, metav1.CreateOptions{})
 		require.NoError(t, err)
+
+		// create a copy of config for discovery clients
+		modifiedClient, err := getKcpClient(cfgCopy, consumerWorkspace)
+		require.NoError(t, err)
```
sttts, line 107: be super specific what this client is. Here: `consumerWorkspaceClient`.

test/e2e/apibinding/apibinding_authorizer_test.go
```diff
@@ -251,7 +269,20 @@ func testCRUDOperations(ctx context.Context, t *testing.T, consumer1Workspace lo
...
 	}
-	_, err := cowboyClient.Create(ctx, cowboy, metav1.CreateOptions{})
+	_, err := cowboyClient.Create(kcp.WithCluster(ctx, consumer1Workspace), cowboy, metav1.CreateOptions{})
 	require.NoError(t, err, "error creating cowboy in consumer workspace %q", consumer1Workspace)
 
 }
+
+// getKcpClient modifies the provided config's hose to point towards the given cluster
+// and returns a client to access objects in that cluster.
+func getKcpClient(config *rest.Config, clusterName logicalcluster.Name) (*clientset.Clientset, error) {
```
varshaprasad96, line 279: Would be nice to move this helper in a common place. This is frequently used for creating modified kcp clients. We need a similar one for k8s client too.

Till we figure out a way to scope discovery client calls, would it be reasonable to scaffold this helper for custom clients for users?

sttts, line 279: this is the wrong factoring IMO. This should be `ConfigWithCluster` giving another rest config.

Then it can go where we have WithCluster now.

test/e2e/apibinding/apibinding_test.go
```diff
@@ -66,14 +68,19 @@ func TestAPIBinding(t *testing.T) {
 	consumer3Workspace := framework.NewWorkspaceFixture(t, server, orgClusterName)
 
 	cfg := server.DefaultConfig(t)
+	cfgCopy := rest.CopyConfig(cfg)
 
-	kcpClients, err := clientset.NewClusterForConfig(cfg)
+	cfg.Wrap(func(rt http.RoundTripper) http.RoundTripper {
+		return kcp.NewClusterRoundTripper(rt)
+	})
```
sttts, line 75: `cfg.Wrap(kcp.NewClusterRoundTripper)`.

This has to become a central helper: `func NewClusterConfig(*rest.Config) *rest.Config`.

test/e2e/virtual/initializingworkspaces/virtualworkspace_test.go
```diff
@@ -240,11 +245,12 @@ func TestInitializingWorkspacesVirtualWorkspaceAccess(t *testing.T) {
 	}
 
 	t.Log("Set up RBAC to allow future calls to succeed")
+	scopedCtx := kcp.WithCluster(ctx, clusterName)
```
sttts, line 248: these are not good. Keep them explicit. Don't alias. It harms readability of code as this aliasing is never uniform but always one-offs.

test/e2e/virtual/initializingworkspaces/virtualworkspace_test.go
```diff
@@ -109,19 +111,22 @@ func TestInitializingWorkspacesVirtualWorkspaceAccess(t *testing.T) {
 	t.Cleanup(cancelFunc)
 
 	sourceConfig := source.DefaultConfig(t)
+	sourceConfig.Wrap(func(rt http.RoundTripper) http.RoundTripper {
+		return kcp.NewClusterRoundTripper(rt)
+	})
```
sttts, line 116: please update `server.DefaultConfig(t)` to return a wrapped config.

sttts: and rename `DefaultConfig` to `ClusterConfig` IMO. "Default" was what Maru chose for "cluster config", i.e. an qualified config.

test/e2e/virtual/apiexport/virtualworkspace_test.go
```diff
@@ -60,44 +62,50 @@ func TestAPIExportVirtualWorkspace(t *testing.T) {
 
 	cfg := server.DefaultConfig(t)
 
-	kcpClients, err := clientset.NewClusterForConfig(cfg)
+	// copy of original config to keep it for other uses
+	cfgCopy := rest.CopyConfig(cfg)
+
+	cfg.Wrap(func(rt http.RoundTripper) http.RoundTripper {
+		return kcp.NewClusterRoundTripper(rt)
+	})
```
sttts, line 70: no aliases please. Make `DefaultConfig` do the right thing.

ncdc: What do you mean?

BTW, because discovery doesn't take in a context, we can't use the wrapper for it. That's why we keep cfgCopy around.

pkg/reconciler/apis/apiexport/apiexport_controller.go
varshaprasad96: `kcpclient` is a common alias while using the generated clients for kcp apis. Have modified it to be `kcpclienthelper` at all places instead to maintain uniformity.

test/e2e/apibinding/apibinding_authorizer_test.go
varshaprasad96: Done.

test/e2e/virtual/initializingworkspaces/virtualworkspace_test.go
varshaprasad96: We would still need the `DefaultConfig` which is the unmodified config for discovery clients. That's why haven't modified the default/source config anywhere and kept a copy around instead.

pkg/reconciler/apis/apiexport/apiexport_controller.go
varshaprasad96: I'm fine with having it `ClusterClient` everywhere, that's why didn't change the naming convention which we had earlier and kept it as is to avoid confusion.

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -81,13 +82,13 @@ func NewController(
 			return namespaceInformer.Lister().Get(clusters.ToClusterAwareKey(clusterName, name))
 		},
 		createNamespace: func(ctx context.Context, clusterName logicalcluster.Name, ns *corev1.Namespace) error {
-			_, err := kubeClusterClient.Cluster(clusterName).CoreV1().Namespaces().Create(ctx, ns, metav1.CreateOptions{})
+			_, err := kubeClusterClient.CoreV1().Namespaces().Create(kcpclienthelper.WithCluster(ctx, clusterName), ns, metav1.CreateOptions{})
```
sttts, line 85: that's a long alis, making these lines much longer. Anything short we can come up with as a convention?

varshaprasad96: How about `kcphelper`? I'm open to other alternatives too. Another apt one would be `kcpapimachinery` but that would made the lines really long.

pkg/server/util.go
```diff
@@ -0,0 +1,45 @@
...
+
+	kcpclienthelper "github.com/kcp-dev/apimachinery/pkg/client"
+	"github.com/kcp-dev/logicalcluster"
+
+	"k8s.io/client-go/rest"
+)
+
+// NewClusterConfig wraps an existing config's roundtripper
+// with a custom cluster aware roundtripper.
+func NewClusterConfig(cfg *rest.Config) *rest.Config {
```
varshaprasad96, line 30: Made a helper here. An apt location for this would probably be `kcp-dev/apimachinery`? But leaving here for now.

test/e2e/virtual/initializingworkspaces/virtualworkspace_test.go
```diff
@@ -115,7 +117,7 @@ func TestInitializingWorkspacesVirtualWorkspaceAccess(t *testing.T) {
 	sourceKcpClusterClient, err := kcpclient.NewClusterForConfig(sourceConfig)
 	require.NoError(t, err)
 
-	kubeClusterClient, err := kubernetes.NewClusterForConfig(sourceConfig)
+	kubeClusterClient, err := kubernetes.NewForConfig(serverutil.NewClusterConfig(sourceConfig))
```
varshaprasad96, line 120: Either we could wrap like this, or if this seems convoluted we could have:
```go
clusterCfg := serverutil.NewClusterConfig(sourceConfig)
```

And have `clusterCfg` as a convention for round tripper wrapped ones.
This way our convention would be:
`cfg`/ `sourceCfg` / `DefaultCfg` - refer to existing configs as is.
`clusterCfg` - refers to any of the above mentioned configs wrapped with round tripper.

Does that sound fine?
cc: @stevekuznetsov

varshaprasad96:
Modified the PR to include bare minimal changes to get client calls in apiexport controller and some relative tests scoped. Once we have the conventions finalized, will create a PR which fixes all other controllers and their tests.

pkg/reconciler/apis/apiexport/apiexport_controller.go
varshaprasad96: Addressed in https://github.com/kcp-dev/kcp/pull/1376#discussion_r911498030

test/e2e/virtual/apiexport/virtualworkspace_test.go
```diff
@@ -60,13 +62,15 @@ func TestAPIExportVirtualWorkspace(t *testing.T) {
 
 	cfg := server.DefaultConfig(t)
 
-	kcpClients, err := clientset.NewClusterForConfig(cfg)
+	clusterCfg := serverutil.NewClusterConfig(cfg)
```
varshaprasad96, line 65: This is to make sure `clusterCfg` is scoped always, based on https://github.com/kcp-dev/kcp/pull/1376/files#r911498030

pkg/server/util.go
sttts: Yes, please move to apimachinery. But looks good!

test/e2e/virtual/initializingworkspaces/virtualworkspace_test.go
sttts: :+1:

pkg/reconciler/apis/apiexport/apiexport_controller.go
ncdc: @varshaprasad96 it looks like the comment you linked is about the rest.Config, not the context?

ncdc: What about logicalcluster.Wrap(ctx, clusterName) / logicalcluster.FromContext(ctx)?

pkg/server/util.go
```diff
@@ -0,0 +1,45 @@
...
+	})
+
+	return copyCfg
+}
+
+// ConfigWithCluster modifies the config host path to include the
+// cluster endpoint.
+func ConfigWithCluster(cfg *rest.Config, clusterName logicalcluster.Name) *rest.Config {
+	r := rest.CopyConfig(cfg)
+	r.Host = r.Host + "/" + clusterName.Path()
```
stevekuznetsov, line 43: nit: `clusterName.Path()` has a leading `/`

test/e2e/virtual/apiexport/virtualworkspace_test.go
stevekuznetsov: Isn't every single caller of `server.DefaultConfig(t)` going to want this? Why not make it the default behavior? Or return both for the few cases where the "raw" one is needed as well for discovery-scoped clients? Or is there any ay to make the roundtripper (`NewClusterConfig()`) compatible with the host-prefix (`ConfigWithCluster()`)?

ncdc: We could explore moving this inside `server.DefaultConfig`. And maybe add another method on `server` to get the raw/unwrapped one.

We'll hopefully not need both routes when upstream Discovery supports a ctx

test/e2e/apibinding/apibinding_authorizer_test.go
```diff
@@ -31,11 +32,13 @@ import (
 	"k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/restmapper"
 
 	"github.com/kcp-dev/kcp/config/helpers"
 	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
 	clientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	serverutil "github.com/kcp-dev/kcp/pkg/server"
```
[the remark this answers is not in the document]
varshaprasad96: Just for uniformity across test files to have `serverutil`

test/e2e/virtual/apiexport/virtualworkspace_test.go
varshaprasad96: Modified this to be other way around. `server.DefaultConfig` remains as is, and instead `server.ClusterConfig` gives us a wrapped one. Didn't want to modify the default config till we are sure that discovery calls will be scoped and artifact generation is fixed.

varshaprasad96:
Addressed review comments, waiting on https://github.com/kcp-dev/apimachinery/pull/40/files to merge.

varshaprasad96:
@ncdc @stevekuznetsov @sttts Addressed all reviews. Is it ready to merge?

test/e2e/framework/kcp.go
```diff
@@ -395,6 +397,10 @@ func (c *kcpServer) DefaultConfig(t *testing.T) *rest.Config {
 	return rest.AddUserAgent(rest.CopyConfig(cfg), t.Name())
 }
 
+func (c *kcpServer) ClusterConfig(r *rest.Config) *rest.Config {
```
stevekuznetsov, line 400: I started a thread in Slack for this, but IMHO the round-tripper (with some extra logic) could handle the case where a cluster is provided statically, which would mean that constructing discovery clients would not need "raw" config and there would be no need to have any distinction between `DefaultConfig()` and `ClusterConfig()` - the fewer choices a dev needs to make, the fewer options for mistakes to be made, and the less complex everything ends up. I think we should take a stab at that.

stevekuznetsov (approved):

PR #1381, opened 2022-06-28
Makefile: wait for test server admin.kubeconfig before starting tests

sttts:
Our tests expect the admin.kubeconfig and will fail if it does not exist.


Makefile
```diff
@@ -187,6 +187,7 @@ test-e2e-shared: build-all
 	TEST_IMAGE=$$(KO_DOCKER_REPO=kind.local ko build --platform=linux/$(ARCH) ./test/e2e/fixtures/kcp-test-image) && test -n "$${TEST_IMAGE}"; \
 	NO_GORUN=1 ./bin/test-server --log-file-path="$(LOG_DIR)/kcp.log" $(TEST_SERVER_ARGS) 2>&1 & PID=$$!; echo "PID $$PID"; \
 	trap 'kill -TERM $$PID' TERM INT EXIT; \
+	while [ ! -f .kcp/admin.kubeconfig ]; do sleep 1; done; \
```
p0lyn0mial, line 190: out of curiosity: could you tell me where this file is created?

p0lyn0mial: ic, we wait to avoid races, thx.

PR #1382, opened 2022-06-28
apibinding/conflictchecker: use APIBinding.status.boundResources, not APIExport

sttts:
Never require APIExport outside of core bind process. APIBinding status must be sufficient.

### DIFF
--- a/pkg/reconciler/apis/apibinding/conflict_checker.go
+++ b/pkg/reconciler/apis/apibinding/conflict_checker.go
@@ -30,7 +30,6 @@ import (
 
 type conflictChecker struct {
 	listAPIBindings      func(clusterName logicalcluster.Name) ([]*apisv1alpha1.APIBinding, error)
-	getAPIExport         func(clusterName logicalcluster.Name, name string) (*apisv1alpha1.APIExport, error)
 	getAPIResourceSchema func(clusterName logicalcluster.Name, name string) (*apisv1alpha1.APIResourceSchema, error)
 	getCRD               func(clusterName logicalcluster.Name, name string) (*apiextensionsv1.CustomResourceDefinition, error)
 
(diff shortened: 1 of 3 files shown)


pkg/reconciler/apis/apibinding/conflict_checker.go
```diff
@@ -53,33 +53,13 @@ func (ncc *conflictChecker) getBoundCRDs(apiBindingToExclude *apisv1alpha1.APIBi
...
-			if err != nil {
-				return err
-			}
-
-			if !boundSchemaUIDs.Has(string(schema.UID)) {
-				continue
-			}
-
-			crd, err := ncc.getCRD(ShadowWorkspaceName, string(schema.UID))
+		for _, bs := range apiBinding.Status.BoundResources {
```
p0lyn0mial, line 61: mhm, I think the apibinding controller sets the `BoundResources` after the checker says it is okay. Thus relying only on `apiBinding.Status.BoundResources` seems wrong, no?

sttts: But this is about "the other" bound resources, not of the current object.

p0lyn0mial: right, maybe it is actually even better. In case of a conflict (concurrent APIExports) we will be able at least bind one CRD. I think previously we would reject both.

pkg/reconciler/apis/apibinding/conflict_checker.go
```diff
@@ -53,33 +53,13 @@ func (ncc *conflictChecker) getBoundCRDs(apiBindingToExclude *apisv1alpha1.APIBi
 		if apiBinding.Name == apiBindingToExclude.Name {
 			continue
 		}
-
-		apiExportClusterName, err := getAPIExportClusterName(apiBinding)
-		if err != nil {
-			return err
-		}
-
-		apiExport, err := ncc.getAPIExport(apiExportClusterName, apiBinding.Spec.Reference.Workspace.ExportName)
```
p0lyn0mial, line 62: remove `getAPIExport` from the `conflictChecker` struct.

pkg/reconciler/apis/apibinding/conflict_checker.go
```diff
@@ -53,33 +53,13 @@ func (ncc *conflictChecker) getBoundCRDs(apiBindingToExclude *apisv1alpha1.APIBi
...
-				return err
-			}
-
-			if !boundSchemaUIDs.Has(string(schema.UID)) {
-				continue
-			}
-
-			crd, err := ncc.getCRD(ShadowWorkspaceName, string(schema.UID))
+		for _, bs := range apiBinding.Status.BoundResources {
+			crd, err := ncc.getCRD(ShadowWorkspaceName, bs.Schema.UID)
```
stevekuznetsov, line 62: Why more `bs`? `resource` or `boundResource` would work ... we should prefer names that mean something

pkg/reconciler/apis/apibinding/conflict_checker.go
sttts: yes, chance for something to break is N times higher

pkg/reconciler/apis/apibinding/conflict_checker.go
sttts: brs? :) but yes, bs is bad here as it usually means bytes.

sttts:
@p0lyn0mial ready. ptal

PR #1383, opened 2022-06-28
informer: do not klog.Fatalf

sttts:
Adding a retry loop.

### DIFF
--- a/pkg/informer/informer.go
+++ b/pkg/informer/informer.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/dynamicinformer"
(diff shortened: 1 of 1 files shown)


pkg/informer/informer.go
```diff
@@ -216,11 +217,15 @@ func (d *DynamicDiscoverySharedInformerFactory) AddIndexers(indexers cache.Index
 
 func (d *DynamicDiscoverySharedInformerFactory) Start(ctx context.Context) {
 	// Immediately discover types and start informing.
-	// TODO: Feed any failure to discover types into /readyz, instead of
-	// panicking.
-	if err := d.discoverTypes(ctx); err != nil {
-		// TODO(sttts): don't klog.Fatal, but return an error
-		klog.Fatalf("Error discovering initial types: %v", err)
+	if err := wait.PollImmediateInfiniteWithContext(ctx, time.Second, func(ctx context.Context) (bool, error) {
```
p0lyn0mial, line 220: should we tear down any informers when the ctx has been closed?

sttts: we do. See the defer.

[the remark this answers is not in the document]
sttts: don't think. We only error out on context close.

p0lyn0mial: defer will be executed only if `PollImmediateInfiniteWithContext` ends with success, so we might not clean up once the ctx has been closed and we had some errors.

sttts: but nothing is started before PollImmediateInfiniteWithContext succeeds

stevekuznetsov (approved):

PR #1393, opened 2022-06-29
virtual: generalize label selector wrapper and get rid of panic

sttts:
Little restructuring to avoid a panic in a func without error return type.

### DIFF
--- a/pkg/virtual/framework/forwardingregistry/wrappers.go
+++ b/pkg/virtual/framework/forwardingregistry/wrappers.go
@@ -29,23 +29,21 @@ import (
 	"k8s.io/apimachinery/pkg/watch"
 )
 
-func WithLabelSelector(labelSelector map[string]string) StorageWrapper {
-	return func(resource schema.GroupResource, storage *StoreFuncs) *StoreFuncs {
-		requirements, selectable := labels.SelectorFromSet(labels.Set(labelSelector)).Requirements()
-		if !selectable {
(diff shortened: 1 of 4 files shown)


pkg/virtual/framework/forwardingregistry/wrappers.go
```diff
@@ -29,23 +29,21 @@ import (
...
-	return func(resource schema.GroupResource, storage *StoreFuncs) *StoreFuncs {
-		requirements, selectable := labels.SelectorFromSet(labels.Set(labelSelector)).Requirements()
-		if !selectable {
-			// we can't return an error here since this ends up inside of the k8s apiserver code where
-			// no errors are expected, so the best we can do is panic - this is likely ok since there's
-			// no real way that the syncer virtual workspace would ever create an unselectable selector
-			panic(fmt.Sprintf("creating a new store with an unselectable set: %v", labelSelector))
-		}
+func WithStaticLabelSelector(labelSelector labels.Requirements) StorageWrapper {
+	return WithLabelSelector(func(ctx context.Context) labels.Requirements {
```
p0lyn0mial, line 33: `WithLabelSelector` seems to be used only by `WithStaticLabelSelector` which provides an empty `ctx`. Should we change the signature of `WithLabelSelector` method ?

sttts: No, it will be needed by follow-up.

PR #1316, opened 2022-06-21
System CRDs -> APIBindings

sttts:
All system CRDs with exception of group `apis.kcp.dev` are turned into APIExports in the root workspace, named after their API group.

One exception: `ClusterWorkspaceShard` is in its own APIExport `shards.tenancy.kcp.dev` because these resources only live in the root workspace, and we want to bind to `tenancy.kcp.dev` everywhere else.

Resources in `apis.kcp.dev` are protected through admission which stop non-system:masters to change status.

PR #1395, opened 2022-06-29
server/apiextensions: disable non-system-CRD non-identity wildcard requests

sttts:
Let's see whether this already gets green pre-#1316.


p0lyn0mial:
why those requests were allowed in the past?

ncdc:
I think the test failures might be real failures (or tests that need to be adjusted)

sttts:
Rebased

sttts:
Rebased

Issue #1398, opened 2022-06-29
Inconsistent requirements for ClusterWorkspaceType name


MatousJobanek:
**Describe the bug**

I've been playing with CWTs and found a small complication with the expected name formats. As expected, the name of the actual CWT resource needs to meet the standard format for `metadata.name` field. So let's say that I create CWT called `foo`. However, when I want to create a workspace of the type `foo`:
```
k ws create foo-workspace --type foo
```
then it fails with 
```
Error: ClusterWorkspace.tenancy.kcp.dev "foo-workspace" is invalid: spec.type.name: Invalid value: "foo": spec.type.name in body should match '^[A-Z][a-zA-Z0-9]+$'
```
because it expects that the name should start with a capital https://github.com/kcp-dev/kcp/blob/main/pkg/apis/tenancy/v1alpha1/types.go#L135.
When I try to use `--type Foo` then it fails (as expected) that it cannot find that CWT as there is no such resource with that name.

**To Reproduce**
1. Create CWT `foo`
2. run `k ws create foo-workspace --type foo`

**Expected behavior**
It should accept names that start with lower-case

stevekuznetsov:
I can't reproduce this.

```
$ cat /tmp/cwt.yaml 
apiVersion: tenancy.kcp.dev/v1alpha1
kind: ClusterWorkspaceType
metadata:
  name: foo
spec:
  defaultChildWorkspaceType: Universal
  allowedChildWorkspaceTypes:
  - Any
  allowedParentWorkspaceTypes:
  - Any
$ kubectl kcp workspace use root
Current workspace is "root".
$ kubectl apply -f /tmp/cwt.yaml 
clusterworkspacetype.tenancy.kcp.dev/foo created
$ kubectl kcp workspace create foo-workspace --type Foo
Workspace "foo-workspace" (type root:Foo) created. Waiting for being ready.
$ kubectl kcp workspace use foo-workspace
Current workspace is "root:foo-workspace" (type "root:Foo").
```

stevekuznetsov:
test

MatousJobanek:
May I know why the name has to start with capital? Couldn't we just use lower-case names as is used across the whole k8s?
This format - that I create CWT with a lower-case name, but the reference in ClusterWorksapce starts with capital - is a bit confusing TBH.

MatousJobanek:
ok, now I see that you used `--type Foo` and when I try it it works:
```
$ k ws create foo-worksapce --type Foo
[...]

PR #1399, opened 2022-06-29
accept ClusterWorkspaceTypeName starting with lower-case

MatousJobanek:
## Summary
To be able to use custom CWTs (not only the build-in ones) the format of the `ClusterWorkspaceTypeName` should accept names that start with lower-case letters.
I modified the regex pattern to accept both types of letters lower-case and upper-case (to support also the existing build-in CWTs).

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/1398


stevekuznetsov:
This is not the correct fix. When a user does `--type Foo`, the workspace plugin should look for `clusterworkspace.tenancy.kcp.dev foo`.

/hold

MatousJobanek:
Closing as it's not valid anymore

PR #1403, opened 2022-06-29
Update OWNERS to fix my name

davidfestal:
I was added with the wrong name here.


stevekuznetsov:
No need to lock you out of approvals while tests flake...

PR #1411, opened 2022-06-30
authorizer/apibinding: prefix users and groups in MaximalPermissionPolicy

sttts:
```
// maximalPermissionPolicy will allow for a service provider to set an upper bound on what is allowed
// for a consumer of this API. If the policy is not set, no upper bound is applied,
// i.e the consuming users can do whatever the user workspace allows the user to do.
//
// The policy consists of RBAC (Cluster)Roles and (Cluster)Bindings. A request of an user in
// a workspace that binds to this APIExport via an APIBinding is additional checked against
// these rules, with the user name and the groups prefixed with `apis.kcp.dev:binding:`.
//
// For example: assume a user `adam` with groups `system:authenticated` and `a-team` binds to
// this APIExport in another workspace root:org:ws. Then a request in that workspace
// against a resource of this APIExport is authorized as every other request in that workspace,
// but in addition the RBAC policy here in the APIExport workspace has to grant access to the
// user `apis.kcp.dev:binding:adam` with the groups `apis.kcp.dev:binding:system:authenticated`
// and `apis.kcp.dev:binding:a-team`.
//
// +optional
MaximalPermissionPolicy *MaximalPermissionPolicy `json:"maximalPermissionPolicy,omitempty"`
```


davidfestal:
- 1 test is failing. Seems that the test `TestCreateWorkspaceInOrganizationNotAllowed` has to be updated to provide the valid Kind so that permissions can still be tested as the test initially intended. 
- Seems Codegen should be run again.

Apart from this LGTM

PR #1413, opened 2022-06-30
e2e: prepare for system:masters root shard kubeconfig

sttts:
The fake downstream logical clusters need direct shard access with a system:masters user. This PR adds `--root-kubeconfig` and `--root-shard-admin-context` to the test binary.


stevekuznetsov (approved):

stevekuznetsov:
Looks like you need to chase the refactor but the approach looks good.

sttts:
Replaced with @s-urbaniak 's work to fix up the e2e test to use the special root shard config where necessary (and only there).

PR #1416, opened 2022-06-30
capitalize ClusterWorkspaceType names provided in CLI

MatousJobanek:
## Summary
When dumb people like me forget that the name of the CWT has to start with a capital, then the logic capitalizes it.

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/1398

### DIFF
--- a/pkg/cliplugins/workspace/plugin/kubeconfig.go
+++ b/pkg/cliplugins/workspace/plugin/kubeconfig.go
@@ -27,7 +27,8 @@ import (
 	"time"
 
 	"github.com/kcp-dev/logicalcluster"
-
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	metav1beta1 "k8s.io/apimachinery/pkg/apis/meta/v1beta1"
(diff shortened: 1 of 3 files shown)


pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -305,6 +306,22 @@ func (kc *KubeConfig) currentWorkspace(ctx context.Context, host string, workspa
...
+			Path: currentClusterName,
+		}
+	default:
+		return tenancyv1alpha1.ClusterWorkspaceTypeReference{
+			Name: tenancyv1alpha1.ClusterWorkspaceTypeName(cases.Title(language.Und, cases.NoLower).String(workspaceType[separatorIndex+1:])),
+			Path: workspaceType[:separatorIndex],
+		}
+	}
+}
+
```
sttts, line 324: We had other plans with the `--type` flags that overlap with this:

```
$ kubectl ws create --type Foo bar
```

should walk up the hierarchy looking for CWT `foo`. In constrast `--type root:Foo` is an explicit type without the need for this traversal.

I wouldn't mind to automatically derive the right capitalisation in this process.

sttts: While doing this, how much do we actually stick to the capitalisation. When seeing @davidfestal's Rootbucket, or RootBucket, or RoOTBuckeT, I am not sure our current route is good.

ncdc:
Closing based on https://github.com/kcp-dev/kcp/issues/1398#issuecomment-1175182078
/close

PR #1420, opened 2022-07-01
reconciler/apis/apibinding: make phase=Bound permanent

sttts:
Even during rebinding, the APIBinding is not stopping working, i.e. it is Bound at all points in time.

### DIFF
--- a/cmd/test-server/main.go
+++ b/cmd/test-server/main.go
@@ -18,12 +18,11 @@ package main
 
 import (
 	"context"
+	"flag"
 	"fmt"
 	"os"
 	"os/exec"
 	"strings"
-
-	flag "github.com/spf13/pflag"
 
 	genericapiserver "k8s.io/apiserver/pkg/server"
 
--- a/pkg/reconciler/apis/apibinding/apibinding_reconcile.go
+++ b/pkg/reconciler/apis/apibinding/apibinding_reconcile.go
@@ -55,7 +55,14 @@ func (c *controller) reconcile(ctx context.Context, apiBinding *apisv1alpha1.API
 	case apisv1alpha1.APIBindingPhaseBinding:
 		return c.reconcileBinding(ctx, apiBinding)
 	case apisv1alpha1.APIBindingPhaseBound:
-		return c.reconcileBound(ctx, apiBinding)
+		needsRebind, err := c.reconcileBound(ctx, apiBinding)
+		if err != nil {
+			return err
+		}
+		if needsRebind {
+			return c.reconcileBinding(ctx, apiBinding)
+		}
+		return nil
 	default:
 		klog.Errorf("Invalid phase %q for APIBinding %s|%s", apiBinding.Status.Phase, logicalcluster.From(apiBinding).String(), apiBinding.Name)
 		return nil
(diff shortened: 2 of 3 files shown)


pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -228,15 +228,16 @@ func (c *controller) enqueueAPIExport(obj interface{}) {
...
 	if err != nil {
 		runtime.HandleError(err)
 		return
 	}
 
-	for _, apiBinding := range bindingsForExport {
-		c.enqueueAPIBinding(apiBinding)
+	for _, obj := range bindingsForExport {
+		apiBinding := obj.(*apisv1alpha1.APIBinding)
+		klog.V(2).Infof("Queueing APIBinding %s|%s because of APIExport %s|%s", logicalcluster.From(apiBinding), apiBinding.Name, key)
```
ncdc, line 239: Mismatch between % signs & number of variable args. Do you want to convert key to cluster|name?

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -270,15 +271,28 @@ func (c *controller) enqueueAPIResourceSchema(obj interface{}) {
...
 	apiExports, err := c.apiExportsIndexer.ByIndex(indexAPIExportsByAPIResourceSchema, key)
 	if err != nil {
 		runtime.HandleError(err)
 		return
 	}
 
-	for _, apiExport := range apiExports {
-		c.enqueueAPIExport(apiExport)
+	for _, obj := range apiExports {
+		apiExport := obj.(*apisv1alpha1.APIExport)
```
ncdc, line 281: It looks like the reimplementation of the content inside the loop, here and in enqueueAPIExport, is solely to be able to log why an APIBinding was enqueued? If so, what if we changed `enqueueAPIBinding` to `enqueueAPIBinding(obj interface{}, why string)` and have the reason appended? Could do similar for the other enqueue functions as needed.

Not terribly critical by any means, but I'd like to avoid the reimplementations if possible.

sttts: some log string prefix arg would help

sttts: done

pkg/reconciler/apis/apibinding/apibinding_controller.go
sttts: done

sttts:
@ncdc ptal

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -209,34 +209,33 @@ type controller struct {
...
 	bindingsForExport, err := c.apiBindingsIndexer.ByIndex(indexAPIBindingsByWorkspaceExport, key)
 	if err != nil {
 		runtime.HandleError(err)
 		return
 	}
 
-	for _, apiBinding := range bindingsForExport {
-		c.enqueueAPIBinding(apiBinding)
+	for _, obj := range bindingsForExport {
+		c.enqueueAPIBinding(obj, fmt.Sprintf("%s because of APIExport %s", logSuffix, key))
```
ncdc, line 238:
```suggestion
		c.enqueueAPIBinding(obj, fmt.Sprintf("%sbecause of APIExport %s", logSuffix, key))
```

?

sttts:
Here the syncer tests break. Probably there is something in them that make use of the old behaviour. Will check next week.

PR #1423, opened 2022-07-01
apis: add omitempty to optional fields

stevekuznetsov:

Very rough approximation of where necessary:

```
$ git grep -A 1 '+optional' | grep -v optional | grep -v omitempty | grep json
```

### DIFF
--- a/config/root-phase0/apiexport-apiresource.kcp.dev.yaml
+++ b/config/root-phase0/apiexport-apiresource.kcp.dev.yaml
@@ -4,7 +4,6 @@ metadata:
   creationTimestamp: null
   name: apiresource.kcp.dev
 spec:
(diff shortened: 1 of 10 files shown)


stevekuznetsov:
That home-brewed unstructured conversion is really annoying. Every time the type changes, that test breaks.

@ncdc need another LGTM

stevekuznetsov:
@sttts @ncdc could use another LGTM

sttts:
codegen

stevekuznetsov:
WTF I am running it (clean) locally ...

stevekuznetsov:
AH. Hoisted by my own petard. The files change, but the semantic meaning of the objects inside does not. Sigh.

PR #1428, opened 2022-07-01
Generate informers/listers using kcp-dev/code-generator, proof them on some controllers

fabianvf:
## Summary
Generate informers and listers for kcp types (into pkg/clusterclient).
Currently only updated the virtualworkspaceurls and apibindings controllers to use them.

## Related issue(s)
Depends on https://github.com/kcp-dev/code-generator/pull/16/
Fixes #


ncdc:
Can you update the appropriate file in `hack` to regenerate the cluster aware informers + listers?

fabianvf:
> Can you update the appropriate file in `hack` to regenerate the cluster aware informers + listers?

I think I did, was there more? https://github.com/kcp-dev/kcp/pull/1428/files#diff-8c83ae16a2a45b86c331dc336bfae20cd6c585ac3f8e7f09a5a2de2e13551305

I can update it to do so more properly but figured it would make sense to wait for https://github.com/kcp-dev/code-generator/pull/16 to merge, then pull it in as a proper dependency

ncdc:
Gah, sorry, I swear I didn't see it the first time I looked!

hack/tools.go
```diff
@@ -21,6 +21,7 @@ package tools
 
 // This package imports things required by this repository, to force `go mod` to see them as dependencies
 import (
+	_ "github.com/kcp-dev/code-generator"
```
fabianvf, line 24: can't do this until https://github.com/kcp-dev/code-generator/pull/32 makes it in

pkg/server/server.go
```diff
@@ -390,11 +399,15 @@ func (s *Server) Run(ctx context.Context) error {
 
 		go s.kcpSharedInformerFactory.Apis().V1alpha1().APIExports().Informer().Run(ctx.StopCh)
 		go s.kcpSharedInformerFactory.Apis().V1alpha1().APIBindings().Informer().Run(ctx.StopCh)
+		go s.kcpClusterSharedInformerFactory.Apis().V1alpha1().APIExports().Informer().Run(ctx.StopCh)
+		go s.kcpClusterSharedInformerFactory.Apis().V1alpha1().APIBindings().Informer().Run(ctx.StopCh)
```
sttts, line 403: are these temporary to move incrementally over?

fabianvf: yes, if we just replace the kcpSharedInformerFactory with the new ones we'd need to update every controller at once, this way we can do it as a series of smaller PRs and remove the old stuff once nothing is using it anymore.

sttts: what about kube informers? Are the newly generated ones compatible enough to inject them deep into kube?

fabianvf: after this PR they will all generate https://github.com/kcp-dev/code-generator/pull/38

ncdc:
Replaced by #1813

PR #1422, opened 2022-07-01
config: generate APIResourceSchemas and APIExports robustly

stevekuznetsov:
The generator is now reentrant and idempotent, and handles pruning.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

/assign @sttts 
/cc @ncdc @davidfestal

PR #1429, opened 2022-07-01
tenancy: fully qualify type references

stevekuznetsov:
config: generate APIResourceSchemas and APIExports robustly

The generator is now reentrant and idempotent, and handles pruning.

---

tenancy: all of clusterworkspacetype.spec is immutable

Initializers and type extensions were already immutable. We need
child/parent/default type references to be, as well. Additional labels
should be removed soon. Easier to make entirety of spec immutable rather
than allowing labels to be changed.

---

tenancy: fully-qualify type relationships

---

tenancy: roll up conditions into ready

---

clusterworkspacetype: be explicitly permissive

Now that we have fully qualified references, finding some magic value
that is sensible for users and passes validation is getting silly. Just
use a boolean field.

---

clusterwokspacetype: make default nullable

Otherwise you can't omit it and the field is not optional.

---

Part of #1149 
Depends on #1422 
/assign @sttts 
/cc @ncdc @davidfestal 

Follow-ups:

- [ ] permissive defaulting @sttts 
- [ ] union type check for any vs. types @sttts 
- [ ] CWT controller must ensure that allowed children and parent sets are subsets of extensions, if present
- [ ] nested loop over aliases when checking allowed children/parents @stevekuznetsov 
- [ ] fix requeuing and recomputation @stevekuznetsov


config/crds/tenancy.kcp.dev_clusterworkspacetypes.yaml
```diff
@@ -53,37 +53,83 @@ spec:
                 description: additionalWorkspaceLabels are a set of labels that will
                   be added to a ClusterWorkspace on creation.
                 type: object
+              allowAnyChildWorkspaceTypes:
+                description: allowAnyChildWorkspaceTypes permits all ClusterWorkspaceTypes
+                  to be created in a workspace of this type.
+                type: boolean
+              allowAnyParentWorkspaceTypes:
```
sttts, line 60: I like this. It allows us sensible defaults (through admission 🤷‍♂️). Just a sum type would make this nicer:
```
allowedChildren:
  any: {}
  types: [{path: ..., name: ...}]
```

stevekuznetsov:
@sttts updated to use a sum type

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -546,3 +549,15 @@ const (
...
+
+var (
+	// AnyWorkspaceTypeReference is used in allowed child and parent type lists to denote that any type is permissible.
+	AnyWorkspaceTypeReference = ClusterWorkspaceTypeReference{
+		Name: AnyWorkspaceType,
+	}
+	// RootWorkspaceTypeReference is a reference to the root logical cluster, which has no cluster workspace type
+	RootWorkspaceTypeReference = ClusterWorkspaceTypeReference{
+		Name: RootWorkspaceType,
+		Path: RootCluster.String(),
```
sttts, line 578: nit: technically would have expected empty string here as another special case

pkg/reconciler/tenancy/clusterworkspacetype/clusterworkspacetype_controller_reconcile.go
```diff
@@ -411,3 +428,48 @@ func (s referenceSet) UnsortedList() []tenancyv1alpha1.ClusterWorkspaceTypeRefer
 	}
 	return res
 }
+
+// resolveTypeRelationships ensures that:
+// - the default child type and allowed child types allow this owning type as a parent
+//   - the default child type is an allowed child type
+// - the allowed parent types allow this owning type as a child
+func (c *controller) resolveTypeRelationships(cwt *tenancyv1alpha1.ClusterWorkspaceType) error {
```
sttts, line 436: nit: resolve is a strange word for what this func is doing. It is rather validating than resolving. Resolve would either return something (non-error) or mutate.

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -204,23 +204,34 @@ type ClusterWorkspaceTypeSpec struct {
...
+	// permissive with child types.
 	//
 	// +optional
 	AllowedChildWorkspaceTypes []ClusterWorkspaceTypeReference `json:"allowedChildWorkspaceTypes,omitempty"`
 
+	// allowAnyParentWorkspaceTypes permits this type to be created in workspaces
+	// of all other ClusterWorkspaceTypes.
+	//
+	// +optional
+	AllowAnyParentWorkspaceTypes bool `json:"allowAnyParentWorkspaceTypes,omitempty"`
```
sttts: my comment still stands: the intention is good, but this API now stutters. Make it
```
allowedParents:
  anyType: true # either this
  types: [...] # or this
```

config/root/clusterworkspacetype-team.yaml
```diff
@@ -7,7 +7,9 @@ spec:
   defaultChildWorkspaceType:
     name: Universal
     path: root
-  allowAnyChildWorkspaceTypes: true
-  allowedParentWorkspaceTypes:
+  allowedChildren:
+    any: true
```
sttts, line 11: isn't (can't) this be defaulted?

pkg/admission/clusterworkspacetypeexists/admission.go
```diff
@@ -196,17 +196,23 @@ func (o *clusterWorkspaceTypeExists) resolveValidType(parentClusterName logicalc
...
-	if !cwt.Spec.AllowAnyParentWorkspaceTypes && !setContainsAnyType(cwt.Spec.AllowedParentWorkspaceTypes, parentCwt.Status.TypeAliases) {
+	if !selectorMatchesAnyType(cwt.Spec.AllowedParents, parentCwt.Status.TypeAliases) {
 		return nil, fmt.Errorf("cluster workspace %q (of type %s) does not allow for parent workspaces of type %s", workspaceName, ref.String(), parentRef.String())
 	}
 	return cwt, nil
 }
 
-func setContainsAnyType(set, queries []tenancyv1alpha1.ClusterWorkspaceTypeReference) bool {
-	for _, allowed := range set {
+func selectorMatchesAnyType(selector *tenancyv1alpha1.ClusterWorkspaceTypeSelector, queries []tenancyv1alpha1.ClusterWorkspaceTypeReference) bool {
```
sttts, line 208: ignore my comment in the other commit that this surprised. Better now.

pkg/apis/tenancy/v1alpha1/types.go
```diff
@@ -205,27 +205,25 @@ type ClusterWorkspaceTypeSpec struct {
...
+	// workspace of this type.
 	//
 	// By default, no type is allowed. This means no other workspace can be nested
-	// within a workspace of the given type. Use allowAnyChildWorkspaceTypes to be
+	// within a workspace of the given type. Use allowedChildren.any=true to be
 	// permissive with child types.
 	//
 	// +optional
-	AllowedChildWorkspaceTypes []ClusterWorkspaceTypeReference `json:"allowedChildWorkspaceTypes,omitempty"`
+	AllowedChildren *ClusterWorkspaceTypeSelector `json:"allowedChildren,omitempty"`
```
sttts, line 216: can we switch the default? We should be permissive in my eyes, and you nice new types allow that. OpenAPI can do.

config/crds/tenancy.kcp.dev_clusterworkspacetypes.yaml-patch
```diff
@@ -11,6 +11,3 @@
         - root
         - system
         - any
-- op: add
-  path: /spec/versions/name=v1alpha1/schema/openAPIV3Schema/properties/spec/properties/allowedChildWorkspaceTypes/default
-  value: []
```
sttts, line 16: add something to make any and types mutual exclusive.

pkg/apis/tenancy/v1alpha1/types.go
sttts: same below

config/crds/tenancy.kcp.dev_clusterworkspacetypes.yaml
```diff
@@ -61,9 +61,22 @@ spec:
...
                 items:
-                  description: ClusterWorkspaceTypeName is a name of a ClusterWorkspaceType
-                  pattern: ^[A-Z][a-zA-Z0-9]+$
-                  type: string
+                  description: ClusterWorkspaceTypeReference is a globally unique,
+                    fully qualified reference to a cluster workspace type.
+                  properties:
+                    name:
+                      description: name is the name of the ClusterWorkspaceType
+                      pattern: ^[A-Z][a-zA-Z0-9]+$
```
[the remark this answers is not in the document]
stevekuznetsov: Let's make it orthogonal to this please, smaller blast radius.

pkg/admission/clusterworkspacetypeexists/admission.go
```diff
@@ -197,23 +197,16 @@ func (o *clusterWorkspaceTypeExists) resolveValidType(parentClusterName logicalc
...
 		return nil, fmt.Errorf("parent cluster workspace %q (of type %s) does not allow for child workspaces of type %s", parentClusterName, parentRef.String(), ref.String())
 	}
-	if !setContainsAnyType(cwt.Spec.AllowedParentWorkspaceTypes, parentCwt.Status.TypeAliases) {
+	if !cwt.Spec.AllowAnyParentWorkspaceTypes && !setContainsAnyType(cwt.Spec.AllowedParentWorkspaceTypes, parentCwt.Status.TypeAliases) {
 		return nil, fmt.Errorf("cluster workspace %q (of type %s) does not allow for parent workspaces of type %s", workspaceName, ref.String(), parentRef.String())
 	}
 	return cwt, nil
 }
 
 func setContainsAnyType(set, queries []tenancyv1alpha1.ClusterWorkspaceTypeReference) bool {
```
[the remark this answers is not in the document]
stevekuznetsov: Similar to `sets.String().HasAny()` - how would you prefer this to be clarified?

config/root/clusterworkspacetype-team.yaml
stevekuznetsov: Isn't it? Defaults to false, keeps the behavior we had in the past.

PR #1439, opened 2022-07-04
syncer: avoid fatal WorkloadCluster get

sttts:
Client requests without retries are dangerous. This one was even unnecessary.


s-urbaniak:
good catch

/lgtm

sttts:
Stop the bleeding.

PR #1440, opened 2022-07-04
reconciler/apibinding: flake: resources should only be bound when established

sttts:
Before this PR, between CRD creation for a binding and it actually being served there was always some time where the apiextensions-apiserver would return 404 (it does that until a CRD is established). Hence, we must return that new CRD from the lister before we have seen it established, and with that we must not mark a new schema UID bound before the CRD is established.


sttts:
Rebased.

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -318,6 +317,7 @@ func (c *controller) reconcileBinding(ctx context.Context, apiBinding *apisv1alp
 			c.deletedCRDTracker.Remove(crd.Name)
 
 			needToWaitForRequeueWhenEstablished = append(needToWaitForRequeueWhenEstablished, schemaName)
+			continue
```
p0lyn0mial, line 320: okay, I think I understand, skip it until it becomes `established`

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -358,13 +360,23 @@ func (c *controller) reconcileBinding(ctx context.Context, apiBinding *apisv1alp
...
+		found := false
+		for i, r := range apiBinding.Status.BoundResources {
+			if r.Group == schema.Spec.Group && r.Resource == schema.Spec.Names.Plural {
+				apiBinding.Status.BoundResources[i] = newBoundResource
+				found = true
+				break
+			}
+		}
+		if !found {
+			apiBinding.Status.BoundResources = append(apiBinding.Status.BoundResources, newBoundResource)
```
p0lyn0mial, line 373: and here add it only if it was established.

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
p0lyn0mial: a CRD is marked as established by the `api-extensions` server

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -209,34 +209,33 @@ type controller struct {
...
 	bindingsForExport, err := c.apiBindingsIndexer.ByIndex(indexAPIBindingsByWorkspaceExport, key)
 	if err != nil {
 		runtime.HandleError(err)
 		return
 	}
 
-	for _, apiBinding := range bindingsForExport {
-		c.enqueueAPIBinding(apiBinding)
+	for _, obj := range bindingsForExport {
+		c.enqueueAPIBinding(obj, fmt.Sprintf(" because of APIExport %s%s", key, logSuffix))
```
p0lyn0mial, line 238: okay, you added an empty space, so it should be readable.

pkg/server/server.go
```diff
@@ -416,8 +416,27 @@ func (s *Server) Run(ctx context.Context) error {
 		s.kcpSharedInformerFactory.Start(ctx.StopCh)
 		s.rootKcpSharedInformerFactory.Start(ctx.StopCh)
 
-		s.kcpSharedInformerFactory.WaitForCacheSync(ctx.StopCh)
-		s.rootKcpSharedInformerFactory.WaitForCacheSync(ctx.StopCh)
+		if err := wait.PollImmediateInfiniteWithContext(goContext(ctx), time.Millisecond*500, func(ctx context.Context) (bool, error) {
```
p0lyn0mial, line 419: why not start the informers and then wait until they sync in a post-startup hook?

pkg/server/server.go
```diff
@@ -416,8 +416,27 @@ func (s *Server) Run(ctx context.Context) error {
...
+			for t, started := range startedTypes {
+				if !started {
+					klog.V(3).Infof("failed to sync %s informer", t)
+					return false, nil
+				}
+			}
+			startedTypes = s.rootKcpSharedInformerFactory.WaitForCacheSync(ctx.Done())
+			for t, started := range startedTypes {
+				if !started {
+					klog.V(3).Infof("failed to sync %s informer", t)
```
p0lyn0mial, line 430: I think the default log level is `V(2)` - failing to sync an informer seems important so I'd increase the log level.

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
sttts: yes

sttts: We should not show a resource as bound until that has happened.

pkg/reconciler/apis/apibinding/apibinding_controller.go
sttts: yes, a pattern we should use everywhere IMO. X was enqueued because Y changed because Z changed.

pkg/server/server.go
sttts: this is about the context closing. Was falling a couple of time seeing the post-start-hook continue as if the informers had synced.

p0lyn0mial, line 419: apply the same to (line 351):
```
s.kubeSharedInformerFactory.WaitForCacheSync(ctx.StopCh)
s.apiextensionsSharedInformerFactory.WaitForCacheSync(ctx.StopCh)
s.rootKubeSharedInformerFactory.WaitForCacheSync(ctx.StopCh)
```

sttts: true

sttts: fixed

PR #1441, opened 2022-07-04
test-servers: enable audit logs

sttts:
Enables us to 
- trace the whole life-cycle of every object. I.e. which state at which time, changed by which actor. Everything.
- identify hotlooping or just which component produces most load.

### DIFF
--- a/cmd/test-server/kcp/audit-policy.yaml
+++ b/cmd/test-server/kcp/audit-policy.yaml
@@ -1,0 +1,30 @@
+apiVersion: audit.k8s.io/v1
+kind: Policy
+omitStages:
+- RequestReceived
+omitManagedFields: true
+rules:
+- level: None
+  nonResourceURLs:
+  - "/api*"
+  - "/version"
+
+- level: Metadata
+  resources:
+  - group: ""
+    resources: ["secrets", "configmaps"]
+  - group: "authorization.k8s.io"
+    resources: ["subjectaccessreviews"]
+
+- level: Metadata
+  verbs: ["list", "watch"]
+
+- level: Metadata
+  verbs: ["get", "delete"]
+  omitStages:
+  - ResponseStarted
+
+- level: RequestResponse
+  verbs: ["create", "update", "patch"]
+  omitStages:
+  - ResponseStarted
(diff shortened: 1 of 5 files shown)


s-urbaniak:
nice!

sttts:
Fixed the rebase error 🤞

sttts:
Reduced log by a factor of 3. Rougly we should have 170mb of audit logs, but compressable to around 10%.

cmd/test-server/kcp/audit-policy.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+- level: Metadata
+  verbs: ["get", "delete"]
+  omitStages:
+  - ResponseStarted
+
+- level: RequestResponse
+  verbs: ["create", "update", "patch"]
+  omitStages:
+  - ResponseStarted
```
s-urbaniak, line 30: can we shortly describe why these policy settings are being used, why they are useful in tests?

pkg/server/controllers.go
```diff
@@ -124,7 +124,7 @@ func (s *Server) installKubeNamespaceController(ctx context.Context, config *res
 		metadata,
 		discoverResourcesFn,
 		s.kubeSharedInformerFactory.Core().V1().Namespaces(),
-		time.Duration(30)*time.Second,
+		time.Duration(5)*time.Minute,
```
s-urbaniak, line 127: why does this need to be bumped 10 times?

s-urbaniak: maybe a small comment?

s-urbaniak: nm, the commit message does say it all

cmd/test-server/kcp/audit-policy.yaml
s-urbaniak: not a blocker

sttts: you can trace the whole life-cycle of every object. I.e. which state at which time, changed by which actor. Everything.

sttts: added to description

sttts:
/ignore ci/prow/e2e-multiple-runs

Auditing is not enabled in e2e-multiple-runs. We have too many flakes in there.

sttts:
/override ci/prow/e2e-multiple-runs

ci/prow/e2e-multiple-runs does not use auditing, only shared and sharded tests do.

PR #1447, opened 2022-07-05
Bump kube: to pickup multi-value indexer fix #79

sttts:
Picking https://github.com/kcp-dev/kubernetes/pull/79. Most likely the root cause of various 404s for APIBindings.


sttts:
Stop the bleeding. Looks like there is more.

PR #1453, opened 2022-07-06
server: bootstrap root phase-1 after controllers are starting

sttts:
Phase-1 is all optional for the controllers. We can start controllers first and with that allow phase-1 resources to depend on controller behaviour.


davidfestal (approved):
Thanks !

PR #1456, opened 2022-07-06
Impersonation in `workspaces` virtual workspace SubjectAccessReview requests

davidfestal:
## Summary

The `SubjectAccessReview`s issued by the `workspaces` virtual workspace should be "softly" impersonated as the user.
By "softly" we mean that we should carry the user in the request so that the server have a way to know it, but still be the VW privileged user in KCP authorization (to have the right to create a SAR everywhere).

This is based on the following soft impersonation work: https://github.com/kcp-dev/kcp/pull/1468

This is required since the SAR call, which is systematically the first call to the KCP shard in the implementation of the REST endpoint, may trigger the creation of the user Home workspace, and then the user name
needs to be the one of the Home workspace owner.

## Related issue(s)

This PR is a pre-requisite for the [Home workspaces EPIC](https://github.com/kcp-dev/kcp/issues/1069)

### DIFF
--- a/pkg/virtual/workspaces/registry/rest.go
+++ b/pkg/virtual/workspaces/registry/rest.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/apiserver/pkg/registry/rest"
 	rbacinformers "k8s.io/client-go/informers/rbac/v1"
 	"k8s.io/client-go/kubernetes"
+	clientrest "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	"k8s.io/kube-openapi/pkg/util/sets"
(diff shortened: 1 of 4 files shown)


pkg/virtual/workspaces/registry/rest.go
```diff
@@ -286,14 +299,29 @@ func (s *REST) authorizeForUser(ctx context.Context, orgClusterName logicalclust
...
 	}
 
+	// We need to impersonate the name of the user here, because the user Home workspace
+	// might be created on-the-fly when receiving the SAR call.
+	// And to automatically create the Home workspace, we need to be the right user.
+	// Additionally, we need to keep the `system:masters` group while impersonating,
+	// in order to have the permission to create the SAR in all cases.
+	// Note: This is not a security risk since we known that the virtual workspace user
+	// is already accessing the KCP shard as `system:masters`.
+	impersonatedClusterClient, err := s.impersonatedkubeClusterClient(&kuser.DefaultInfo{
```
sttts, line 309: why doesn't this mean privilege escalation, i.e. if the SAR triggers a home workspace creation for a user who must not create a home workspace?

PR #1460, opened 2022-07-06
docs: link to virtual workspaces demo

markmc:
Since this demo is referenced as the canonical example of this use case, make it easier to find.


ncdc:
/ok-to-test
/lgtm
/approve

Thanks!!!

ncdc:
This is a docs-only change, and our ci-docs-only config is missing e2e-sharded. That's why it's not running on this PR (even though it's required). I've opened #1466 to fix that.

ncdc:
docs-only PR, merging

PR #1462, opened 2022-07-06
apis/tenancy/ClusterWorkspaceTypes: defaulting and union type constraints for allowed children and parents

sttts:


config/crds/tenancy.kcp.dev_clusterworkspacetypes.yaml-patch
```diff
@@ -11,3 +11,25 @@
...
+- op: add
+  path: /spec/versions/name=v1alpha1/schema/openAPIV3Schema/properties/spec/properties/allowedParents/oneOf
+  value:
+  - required: ["any"]
+  - required: ["types"]
+
+- op: add
+  path: /spec/versions/name=v1alpha1/schema/openAPIV3Schema/properties/spec/properties/allowedChildren/default
+  value:
+    any: bool
```
stevekuznetsov, line 30: Should this be `any: true`?

config/crds/tenancy.kcp.dev_clusterworkspacetypes.yaml-patch
```diff
@@ -11,3 +11,25 @@
...
+
+- op: add
+  path: /spec/versions/name=v1alpha1/schema/openAPIV3Schema/properties/spec/properties/allowedParents/default
+  value:
+    any: bool
```
stevekuznetsov, line 35: Should this be `any: true`?

sttts: yes

sttts: fixed

sttts:
The types are still wrong.

stevekuznetsov:
/lgtm cancel
/hold

We can't default to `any: true`, since we will now be ensuring that:

```
clusterworkspacetype.tenancy.kcp.dev root:A allows children {any: false, types: {...}}
clusterworkspacetype.tenancy.kcp.dev root:B extends root:A, does not specify allowed chlidren
```

With this PR, `root:B` gets `{any: true}` defaulted, which is *not* a subset of `root:A`. What the user meant to do when not specifying the field was to inherit the restrictions from `root:A`, but with a defaulted field this is now impossible.

/cc @csams

sttts:
@stevekuznetsov not sure the comment is already clarified in Slack. As written here I don't get it.

I can see how defaulting of the default is not desirably potentially. But for allowedParents and allowedChildren it's a noop. If we honor all invariants by the aliases, this will not restrict anything.

Note: we don't default `any: true`, we default `{any: true, types: nil}`. That's an important difference.

sttts:
Replaced with https://github.com/kcp-dev/kcp/pull/1474

PR #1472, opened 2022-07-08
Update placement doc

qiujian16:

## Related issue(s)

Fixes #

### DIFF
--- a/docs/locations-and-scheduling.md
+++ b/docs/locations-and-scheduling.md
@@ -1,4 +1,4 @@
-# Locations and Scheduling
+# Placement, Locations and Scheduling
 
 KCP implements *Compute as a Service* via a concept of Transparent Multi Cluster (TMC). TMC means that
 Kubernetes clusters are attached to a kcp installation to execute workload objects from the users'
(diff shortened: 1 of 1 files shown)


docs/locations-and-scheduling.md
```diff
@@ -29,77 +29,106 @@ The APIs used for Compute as a Service are:
...
+  1. Network connectivity: a workload in one `SyncTarget` can access service in another `SyncTarget` with a predefined DNS format.
+  2. Storage compatibility: a storage in one `SyncTarget` can be easily moved to another `SyncTarget`.
+
+- `Placement` in `scheduling.kcp.dev/v1alpha1` – represents a selection rule to choose ONE `Location` via location labels, and bind
+  the selected location to MULTIPLE namespaces in a user workspace. Users can create multiple `Placement`s to select multiple `Locations`
+  for namespaces.
+
+  `Placement` are visible and writable to users. A default `Placement` is automatically created when a workload `APIBinding` is
+  created on the user workspace, which randomly select a `Location` and bind to all namespaces in this workspace. User can mutate or delete
+  the default placement.
```
sttts, line 44: mention the annotation to disable that?

docs/locations-and-scheduling.md
```diff
@@ -29,77 +29,106 @@ The APIs used for Compute as a Service are:
   Locations are visible to users, but owned by the compute service team, i.e. read-only to the users and only projected
   into their workspaces for visibility. A placement decision references a location by name.
 
+  `SyncTarget`s in a `Location` are transparent to the user. Workload should be able to seamless move from one `SyncTarget` to another
+  within a `Location`. It is compute service's responsibility to ensure that:
+
+  1. Network connectivity: a workload in one `SyncTarget` can access service in another `SyncTarget` with a predefined DNS format.
```
sttts, line 35: rather than "a DNS format" it is the Kubernetes service DNS format.

Maybe in general, mention that we strive for kube conformance for workloads in a location in the sense that for the user it looks like ONE cluster.

docs/locations-and-scheduling.md
```diff
@@ -29,77 +29,106 @@ The APIs used for Compute as a Service are:
...
+      cloud: aws
+  namespaceSelector:
+    matchLabels:
+      app: foo
+  locationWorkspace: root:default:location-ws
+```
 
-The location strings are of the form `<locationClusterName>+<locationName>+<syncTargetIdentifier>`.
+A matched location will be selected for this `Placement` at first, which makes the `Placement` turns from `Pending` to `Unbound`. Then if there is at
+least one matched ns, the ns will be annotated with `scheduling.kcp.dev/placement` and the placement turns from `Unbound` to `Bound`. After this, a
```
sttts, line 88:
```suggestion
least one matching ns, the ns will be annotated with `scheduling.kcp.dev/placement` and the placement turns from `Unbound` to `Bound`. After this, a
```

docs/locations-and-scheduling.md
```diff
@@ -29,77 +29,106 @@ The APIs used for Compute as a Service are:
...
+`SyncTarget` will be selected from the location picked by the placement.  `state.internal.workload.kcp.dev/<cluster-id>` label with value of `Sync` will
+be set if a valid `SyncTarget` is selected.
 
-Example:
+The use can create another placement targeted to a different location for this ns, e.g.
```
sttts, line 92:
```suggestion
The user can create another placement targeted to a different location for this ns, e.g. 
```

docs/locations-and-scheduling.md
```diff
@@ -29,77 +29,106 @@ The APIs used for Compute as a Service are:
...
+  - matchLabels:
+      cloud: gce
+  namespaceSelector:
+    matchLabels:
+      app: foo
+  locationWorkspace: root:default:location-ws
 ```
 
-Note: multiple sync targets can be bound at the same time.
+which wiil finally result in another `state.internal.workload.kcp.dev/<cluster-id>` label added to the ns, and the ns will have two different
```
sttts, line 109:
```suggestion
which will result in another `state.internal.workload.kcp.dev/<cluster-id>` label added to the ns, and the ns will have two different
```

docs/locations-and-scheduling.md
```diff
@@ -29,77 +29,106 @@ The APIs used for Compute as a Service are:
...
+`state.internal.workload.kcp.dev/<cluster-id>` label.
 
-The placement state machine is (to be) protected against mutation by the user via admission. The user interface
-to influence the placement decisions is (will be) the `Placement` object.
+Placement is in the `Ready` status condition when
 
-### Resource Syncing
+1. selected location matchs the `Placement` spec.
```
sttts, line 114:
```suggestion
1. selected location matches the `Placement` spec.
```

docs/locations-and-scheduling.md
```diff
@@ -29,77 +29,106 @@ The APIs used for Compute as a Service are:
...
-controller will pick up the namespace and
+#### Sync target removing
 
-1. set the `scheduling.kcp.dev/placement` annotation state to `Bound` and
-2. set the `state.internal.workload.kcp.dev/<cluster-id>` label to `Sync`.
+A sync target will be removed when:
+
+1. corresponding `Placement` is deleted.
+2. corresponding `Placement` is not in `Ready` condition.
+3. corresponding `SyncTarget` is evicting/not Ready/deleted
```
sttts, line 123: am curious: do we really evict on "not ready" of the sync target? We have to rethink this topic. In kube, I believe there is some larger timeout (10min?) of unreadiness. Here we talk about clusters. I think we should have an even longer grace period, maybe even infinite if the user wants it, i.e. this must likely become user configurable.

It's also related to https://kubernetes.io/blog/2022/05/20/kubernetes-1-24-non-graceful-node-shutdown-alpha/.

docs/locations-and-scheduling.md
```diff
@@ -29,77 +29,106 @@ The APIs used for Compute as a Service are:
...
   into their workspaces for visibility. A placement decision references a location by name.
 
+  `SyncTarget`s in a `Location` are transparent to the user. Workload should be able to seamless move from one `SyncTarget` to another
+  within a `Location`. It is compute service's responsibility to ensure that:
+
+  1. Network connectivity: a workload in one `SyncTarget` can access service in another `SyncTarget` with a predefined DNS format.
+  2. Storage compatibility: a storage in one `SyncTarget` can be easily moved to another `SyncTarget`.
+
+- `Placement` in `scheduling.kcp.dev/v1alpha1` – represents a selection rule to choose ONE `Location` via location labels, and bind
+  the selected location to MULTIPLE namespaces in a user workspace. Users can create multiple `Placement`s to select multiple `Locations`
```
robszumski, line 39: Is the point here to indicate that multiple namespaces can end up in different locations if desired? Eg, is this more accurate/clear:

> For Workspaces with multiple Namespaces, users can create multiple `Placement`s to assign specific Namespace(s) to specific `Locations`.

robszumski (changes requested):
This is a great update, thank you. 

One nit: replace all the `ns` with `Namespace`

docs/locations-and-scheduling.md
qiujian16: we currently do this. We could introduce a taint/toleration for clusters something like this? https://github.com/open-cluster-management-io/enhancements/tree/main/enhancements/sig-architecture/20-taint-toleration

sttts: We probably want taint/toleration eventually. But I don't think it has to do with the eviction topic, has it? Do we currently schedule away from invalid clusters? We definitely need more controller for that.

sttts: Can you create an issue to come back to this? I tend to even see this as a time-bomb. I think it's far too aggressive and could be fatal if a cluster blibs for reasons like cloud provider hickups.

PR #1473, opened 2022-07-08
ddsif: replace discovery polling

### LINKED ISSUE #1350: Dynamic discovery shared informer factory: fix problems :smile:
(1) The "dynamic discovery" nature won't scale beyond a small number of workspaces.

We are currently iterating through every workspace, getting discovery for it (live http call), collecting all the namespace-scoped GVRs that support list+watch, creating a big union of all of the GVRs (to dedupe), and then ensuring informers are started for any newly-seen GVRs, and stopped for any now defunct GVRs.

We don't need to do this, because every GVR in the system is the deduped sum of
a. built-in types from upstream kube
b. built-in types from kcp (from `system:system-crds`)
c. bound types from apiexports/apibindings (from `system:bound-crds`)
d. "normal" CRDs from any/all workspaces

We hopefully can switch to a mechanism where we list/watch all CRDs and recalculate informers (additions, removals) based on CRD events.

Maybe we could even make it the intersection of all the GVRs we know about, restricted to only those resources that are being synced to clusters. In other words, once we move `--resources` from the syncer CLI flags to the WorkloadCluster/SyncTarget spec, we'll know everything being synced, and we can perform this calculation. If it's mostly deployments/secrets/configmaps/pvcs being synced, this has the potential to reduce unnecessary load (informers + labeling resources we don't really need to).

(2) It only supports namespace-scoped resources because it was initially built for labeling resources for scheduling, but other consumers are wanting to use it (quota, APIExport claims/labeling)

We should find a way to support both use cases (driving placement labeling, and other non placement needs), or split this into multiple factories maybe?

ncdc:
## Summary
Replace the polling mechanism that wouldn't scale as the number of
ClusterWorkspaces increased with one that is based on a fixed-set of
built-in types + a dynamic set of informers driven by all the unique
group-resources across all the CRDs in the system.

## Related issue(s)

Fixes #1350

### DIFF
--- a/pkg/informer/informer.go
+++ b/pkg/informer/informer.go
@@ -28,80 +28,182 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
-	"k8s.io/apimachinery/pkg/labels"
+	apiextensionsinformers "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/informers"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 
-	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
-	tenancylisters "github.com/kcp-dev/kcp/pkg/client/listers/tenancy/v1alpha1"
+	metadataclient "github.com/kcp-dev/kcp/pkg/metadata"
 )
 
 const (
 	resyncPeriod = 10 * time.Hour
-)
 
-type clusterDiscovery interface {
-	WithCluster(name logicalcluster.Name) discovery.DiscoveryInterface
-}
+	byGroupFirstFoundVersionResourceIndex = "byGroup-firstFoundVersion-resource"
+)
 
 // DynamicDiscoverySharedInformerFactory is a SharedInformerFactory that
 // dynamically discovers new types and begins informing on them.
 type DynamicDiscoverySharedInformerFactory struct {
-	workspaceLister tenancylisters.ClusterWorkspaceLister
-	disco           clusterDiscovery
-	dynamicClient   dynamic.Interface
-	filterFunc      func(interface{}) bool
-	pollInterval    time.Duration
-	indexers        cache.Indexers
+	dynamicClient     dynamic.Interface
+	filterFunc        func(interface{}) bool
+	indexers          cache.Indexers
+	crdIndexer        cache.Indexer
+	crdInformerSynced cache.InformerSynced
 
 	// handlersLock protects multiple writers racing to update handlers.
 	handlersLock sync.Mutex
 	handlers     atomic.Value
 
-	mu               sync.RWMutex
+	// updateCh receives notifications for all CRD add/update/delete events, so we can start new informers and stop
+	// informers no longer needed.
+	updateCh chan struct{}
+
+	informersLock    sync.RWMutex
 	informers        map[schema.GroupVersionResource]informers.GenericInformer
 	startedInformers map[schema.GroupVersionResource]bool
 	informerStops    map[schema.GroupVersionResource]chan struct{}
-	terminating      bool
+}
+
+// NewDynamicDiscoverySharedInformerFactory returns a factory for shared
+// informers that discovers new types and informs on updates to resources of
+// those types.
+func NewDynamicDiscoverySharedInformerFactory(
+	cfg *rest.Config,
+	filterFunc func(obj interface{}) bool,
+	crdInformer apiextensionsinformers.CustomResourceDefinitionInformer,
+	indexers cache.Indexers,
+) (*DynamicDiscoverySharedInformerFactory, error) {
+	cfg = rest.AddUserAgent(rest.CopyConfig(cfg), "kcp-partial-metadata-informers")
+
+	metadataClusterClient, err := metadataclient.NewDynamicMetadataClusterClientForConfig(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	f := &DynamicDiscoverySharedInformerFactory{
+		dynamicClient:     metadataClusterClient.Cluster(logicalcluster.Wildcard),
+		filterFunc:        filterFunc,
+		indexers:          indexers,
+		crdIndexer:        crdInformer.Informer().GetIndexer(),
+		crdInformerSynced: crdInformer.Informer().HasSynced,
+
+		// Use a buffered channel of size 1 to allow enqueuing 1 update notification
+		updateCh: make(chan struct{}, 1),
+
+		informers:        make(map[schema.GroupVersionResource]informers.GenericInformer),
+		startedInformers: make(map[schema.GroupVersionResource]bool),
+		informerStops:    make(map[schema.GroupVersionResource]chan struct{}),
+	}
+
+	f.handlers.Store([]GVREventHandler{})
+
+	// Add an index function that indexes a CRD by its group/firstServedVersion/resource. We only need the first
+	// served version because this shared informer factory is expected to be using a wildcard client for partial
+	// metadata only. In this instance, version does not matter, because a wildcard partial metadata list request
+	// for CRs always serves all CRs for the group-resource, regardless of storage version.
+	if err := crdInformer.Informer().AddIndexers(cache.Indexers{
+		byGroupFirstFoundVersionResourceIndex: func(obj interface{}) ([]string, error) {
+			crd, ok := obj.(*apiextensionsv1.CustomResourceDefinition)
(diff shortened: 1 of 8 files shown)


pkg/informer/informer.go
```diff
@@ -24,92 +24,182 @@ import (
...
+		startedInformers: make(map[schema.GroupVersionResource]bool),
+		informerStops:    make(map[schema.GroupVersionResource]chan struct{}),
+		terminatedCh:     make(chan struct{}),
+	}
+
+	f.handlers.Store([]GVREventHandler{})
+
+	// Add an index function that indexes a CRD by its group/firstServedVersion/resource. We only need the first
+	// served version because this shared informer factory is expected to be using a wildcard client for partial
+	// metadata only. In this instance, version does not matter, because a wildcard partial metadata list request
```
stevekuznetsov, line 93: because a wildcard partial metadata list request ... what?

pkg/informer/informer.go
```diff
@@ -24,92 +24,182 @@ import (
...
+			firstServedVersion := ""
+			for _, version := range crd.Spec.Versions {
+				if !version.Served {
+					continue
+				}
+				firstServedVersion = version.Name
+				break
+			}
+
+			if firstServedVersion == "" {
```
stevekuznetsov, line 110: When is this true?

pkg/informer/informer.go
```diff
@@ -134,14 +220,10 @@ func (d *DynamicDiscoverySharedInformerFactory) informerForResourceLockHeld(gvr
 		},
 	})
 
-	if err := inf.Informer().AddIndexers(d.indexers); err != nil {
```
stevekuznetsov, line 137: What's the implication of setting these in the constructor vs using this AddIndexers call?

pkg/informer/informer.go
```diff
@@ -244,140 +300,103 @@ func (d *DynamicDiscoverySharedInformerFactory) AddEventHandler(handler GVREvent
...
-				}
-			}
-		}
+		d.informersLock.Unlock()
 	}()
-}
 
-func (d *DynamicDiscoverySharedInformerFactory) discoverTypes(ctx context.Context) error {
-	latest := map[schema.GroupVersionResource]struct{}{}
+	wait.UntilWithContext(ctx, func(ctx context.Context) {
```
stevekuznetsov, line 316: What's the intent with `UntilWithContext` here? Couldn't you just:


```go
for {
    select {
    case <-ctx.Done():
        return
    case <-d.updateCh:
    }
    d.updateInformers()
}
```

pkg/informer/informer.go
ncdc: I'm tempted to have the constructor take in a rest.Config instead of a client, so we can force the construction the way we need it...

The only way this factory is _currently_ used is with a wildcard partial metadata client. For CRs, when we receive a wildcard partial metadata list, we return all CRs across all clusters, regardless of storage version. See https://github.com/kcp-dev/kubernetes/blob/005ea1354ed5c15eb1f98c1d3c3a7bc6a84c5e17/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/kcp_rest_options_getter.go#L30-L38 for more details.

For the built in types, we aren't currently mixing any storage versions of anything that I'm aware of...

pkg/informer/informer.go
ncdc: Probably never, but this was a "just in case"

pkg/informer/informer.go
ncdc: You can't have an error when doing it in the constructor. You can when using AddIndexers. There is no functional difference.

pkg/informer/informer.go
ncdc: I'm basically trying to mimic "ServerPreferredGroupVersions"

stevekuznetsov: Thanks for the explanation - mind updating the comment? My first response here was just that you left the paragraph mid-sentence without getting to a point :)

pkg/informer/informer.go
stevekuznetsov: Gotcha.

pkg/informer/informer.go
ncdc: heh, got sidetracked. Thanks for noticing!

pkg/informer/informer.go
ncdc: Added a comment, PTAL

pkg/informer/informer.go
ncdc: Finished my sentence, PTAL

ncdc:
I can't get e2e-sharded to pass with this PR. It's a different test that fails each time. I can get them to pass locally with the sharded test server. Maybe these changes are making things flakier (i.e. a real bug)?

ncdc:
/hold

Testing out this + #1486 together

pkg/informer/informer.go
```diff
@@ -24,92 +24,179 @@ import (
...
 	handlers     atomic.Value
 
-	mu               sync.RWMutex
+	updateCh chan struct{}
+
+	informersLock    sync.RWMutex
 	informers        map[schema.GroupVersionResource]informers.GenericInformer
 	startedInformers map[schema.GroupVersionResource]bool
 	informerStops    map[schema.GroupVersionResource]chan struct{}
 	terminating      bool
```
sttts, line 63: is terminating protected by the lock? Looks like, but the name suggests differently.

pkg/informer/informer.go
```diff
@@ -24,92 +24,179 @@ import (
...
+	filterFunc    func(interface{}) bool
+	indexers      cache.Indexers
+	crdIndexer    cache.Indexer
 
 	// handlersLock protects multiple writers racing to update handlers.
 	handlersLock sync.Mutex
 	handlers     atomic.Value
 
-	mu               sync.RWMutex
+	updateCh chan struct{}
```
sttts, line 57: go doc

pkg/informer/informer.go
```diff
@@ -24,92 +24,179 @@ import (
...
 )
 
 const (
 	resyncPeriod = 10 * time.Hour
-)
 
-type clusterDiscovery interface {
-	WithCluster(name logicalcluster.Name) discovery.DiscoveryInterface
-}
+	groupFirstFoundVersionResourceIndex = "group-firstFoundVersion-resource"
```
sttts, line 42: everywhere else we call these `byFooIndex = "byFoo"`

pkg/informer/informer.go
```diff
@@ -244,140 +301,106 @@ func (d *DynamicDiscoverySharedInformerFactory) AddEventHandler(handler GVREvent
...
+		klog.V(4).InfoS("Waiting for notification")
+		select {
+		case <-ctx.Done():
+			return
+		case <-d.updateCh:
+		}
 
-	// Get a list of all the logical cluster names. We'll get discovery from all of them, union all the GVRs, and use
-	// that union for the informer.
+		klog.V(4).InfoS("Notification received")
```
sttts, line 328: this is now spamming CI logs, isn't it? We run in 4.

pkg/informer/informer.go
```diff
@@ -244,140 +301,106 @@ func (d *DynamicDiscoverySharedInformerFactory) AddEventHandler(handler GVREvent
...
-	// TODO(ncdc): this may not scale well. Watchable discovery or something like that
-	// is a better long term solution.
-	workspaces, err := d.workspaceLister.List(labels.Everything())
-	if err != nil {
-		return err
-	}
-	for i := range workspaces {
-		logicalClusterName := logicalcluster.From(workspaces[i]).Join(workspaces[i].Name).String()
+func (d *DynamicDiscoverySharedInformerFactory) updateInformers() {
+	klog.V(2).InfoS("Determining dynamic informer additions and removals")
```
sttts, line 334: log on changes, not for every event

pkg/informer/informer.go
```diff
@@ -244,140 +301,106 @@ func (d *DynamicDiscoverySharedInformerFactory) AddEventHandler(handler GVREvent
...
+		gvrFor("rbac.authorization.k8s.io", "v1", "clusterroles"):                        {},
+		gvrFor("rbac.authorization.k8s.io", "v1", "clusterrolebindings"):                 {},
+		gvrFor("rbac.authorization.k8s.io", "v1", "roles"):                               {},
+		gvrFor("rbac.authorization.k8s.io", "v1", "rolebindings"):                        {},
+		gvrFor("flowcontrol.apiserver.k8s.io", "v1beta2", "flowschemas"):                 {},
+		gvrFor("flowcontrol.apiserver.k8s.io", "v1beta2", "prioritylevelconfigurations"): {},
+		gvrFor("events.k8s.io", "v1", "events"):                                          {},
+		gvrFor("admissionregistration.k8s.io", "v1", "mutatingwebhookconfigurations"):    {},
+		gvrFor("admissionregistration.k8s.io", "v1", "validatingwebhookconfigurations"):  {},
+		gvrFor("apiextensions.k8s.io", "v1", "customresourcedefinitions"):                {},
```
sttts, line 364: would like to see a unit test that compares this list to the server schema minus some disallow list.

pkg/informer/informer.go
ncdc: Should be at most once per second. I'll lower to 5

pkg/server/options/flags.go
```diff
@@ -128,7 +128,7 @@ var (
 		"tracing-config-file", // File with apiserver tracing configuration.
 
 		// KCP flags
-		"discovery-poll-interval",     // Polling interval for dynamic discovery informers.
+		"enable-sharding",             // Enable delegating to peer kcp shards.
```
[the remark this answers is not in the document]
ncdc: must have been, will fix

PR #1478, opened 2022-07-09
Add shard-base-url flag to default clusterworkspaceshards

csams:
Add `shard-base-url` flag to make the `BaseURL` default value configurable for `ClusterWorkspaceShards`

### DIFF
--- a/pkg/admission/clusterworkspaceshard/admission.go
+++ b/pkg/admission/clusterworkspaceshard/admission.go
@@ -52,7 +52,7 @@ func Register(plugins *admission.Plugins) {
 type clusterWorkspaceShard struct {
 	*admission.Handler
 
-	shardBaseURLProvider    func() string
+	shardBaseURL            string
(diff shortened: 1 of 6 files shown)


pkg/admission/clusterworkspaceshard/admission.go
```diff
@@ -125,6 +138,10 @@ func (o *clusterWorkspaceShard) Admit(ctx context.Context, a admission.Attribute
 	return nil
 }
 
+func (o *clusterWorkspaceShard) SetShardBaseURLProvider(shardBaseURLProvider func() string) {
+	o.shardBaseURLProvider = shardBaseURLProvider
```
sttts, line 142: I don't think you need this provider hack here. We only needed it because the external address is not known early enough when admission is setup. Just a simple url.URL or string should do.

csams: Updated

PR #1479, opened 2022-07-11
adds a shard-name flag for assigning a name to a shard instance

p0lyn0mial:
## Summary
the flag will be used to:
- create/update the correct ClusterWorkspaceShard 
- to have custom logic to setup the two informers and the merging, i.e. to know whether this is the root shard or the other one.

the default value of the flag is the `root` name.
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/server/options/options.go
+++ b/pkg/server/options/options.go
@@ -55,6 +55,7 @@ type ExtraOptions struct {
 	ProfilerAddress          string
 	ShardKubeconfigFile      string
 	ShardBaseURL             string
+	ShardName                string
 	EnableSharding           bool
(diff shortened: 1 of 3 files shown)


pkg/server/options/options.go
```diff
@@ -92,6 +93,7 @@ func NewOptions(rootDir string) *Options {
 			ProfilerAddress:          "",
 			ShardKubeconfigFile:      "",
 			ShardBaseURL:             "",
+			ShardName:                "",
```
sttts, line 96: you can default here.

p0lyn0mial: done.

PR #1481, opened 2022-07-11
apis/workload: move workload.kcp.dev/skip-default-object-creation here from apis.kcp.dev/v1alpha1

sttts:
Does not have anything to do with apis.kcp.dev/v1alpha1 in general.


sttts:
Pure rebase.

PR #1488, opened 2022-07-11
e2e: improve output of probers

stevekuznetsov:
Now, we will see the list of components that are *not* ready or healthy
when we asked, instead of a massive dump of the whole request. There
doesn't seem to be any way to get the raw metav1.Status back from the Go
client when there's been a non-200 code, so we fall back to some
best-effort string parsing to provide this output.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

/cc @ncdc 
/assign @sttts @davidfestal

PR #1489, opened 2022-07-11
e2e: require two failed pings to fail the test

stevekuznetsov:
To whatever extent our test setup is causing etcd to have a hiccup here
or there, it does not seem to then cause the rest of the tests to have
any issues. While it's useful to know that there are some latent etcd
issues (or whatever) in the tests, we should be more tolerant of probing
failures, since things happen and distributed, eventually consistent
systems are hard.

---

Depends on #1488 
/assign @sttts @ncdc @davidfestal

### DIFF
--- a/test/e2e/framework/kcp.go
+++ b/test/e2e/framework/kcp.go
@@ -45,6 +45,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	apierrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	kubernetesclientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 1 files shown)


test/e2e/framework/kcp.go
```diff
@@ -513,11 +513,27 @@ func (c *kcpServer) monitorEndpoint(client *rest.RESTClient, endpoint string) {
...
 		if err != nil {
-			c.t.Errorf("error contacting %s: %v", endpoint, err)
+			c.t.Errorf("error contacting %s: %v", endpoint, unreadyComponentsFromError(err))
 		}
 	}, 1*time.Second)
 }
 
+// there doesn't seem to be any simple way to get a metav1.Status from the Go client, so we get
+// the content in a string-formatted error, unfortunately.
+func unreadyComponentsFromError(err error) string {
```
sttts, line 536: please give some example output and what you make out of it. This is hard to follow.

stevekuznetsov: Example here: https://github.com/kcp-dev/kcp/pull/1488#issuecomment-1180866176

This happens in like every single e2e test output, did you not notice these big blobs? These are the tricky ones to read when some prober fails unexpectedly ...

PR #1492, opened 2022-07-12
admission/clusterworkspace: move owner annotation code here

sttts:
The owner annotation has nothing to do with types.


qiujian16 (commented):
/lgtm

PR #1493, opened 2022-07-12
server/home: check owner properly

sttts:
- fix home workspace logic to conside the owner annotation as JSON, not just a string
- allow system master to pre-set the owner
- keep the owner username on a workspace when ready
- check the owner on GET `~`

### DIFF
--- a/pkg/server/home_workspaces.go
+++ b/pkg/server/home_workspaces.go
@@ -271,6 +271,7 @@ func (h *homeWorkspaceHandler) ServeHTTP(rw http.ResponseWriter, req *http.Reque
 		// we return the Home workspace definition of the current user, possibly even before its
 		// underlying ClusterWorkspace resource exists.
 
+		getAttributes := homeWorkspaceAuthorizerAttributes(effectiveUser, "get")
 		homeLogicalClusterName := h.getHomeLogicalClusterName(effectiveUser.GetName())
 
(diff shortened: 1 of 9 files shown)


pkg/server/home_workspaces.go
```diff
@@ -500,12 +521,12 @@ func tryToCreate(h *homeWorkspaceHandler, ctx context.Context, userName string,
 					return 0, err
 				}
 
-				if cw.Annotations[tenancyv1alpha1.ClusterWorkspaceOwnerAnnotationKey] != userName {
+				if info, _ := unmarshalOwner(cw); info == nil || info.UID != user.GetUID() || info.Username != user.GetName() {
 					return 0, kerrors.NewForbidden(tenancyv1alpha1.SchemeGroupVersion.WithResource("clusterworkspaces").GroupResource(), cw.Name, errors.New(authorization.WorkspaceAcccessNotPermittedReason))
```
davidfestal, line 525: Should we somehow log here, at least to know the reason of the Forbidden response when troubleshooting ?

davidfestal (approved):

pkg/server/home_workspaces.go
sttts: added

sttts:
Rebased and added collision logging.

PR #1494, opened 2022-07-12
Wipe owner info as soon as possible

sttts:
We don't need the user info on a) ready workspaces b) on user-visible Workspaces.

### DIFF
--- a/pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_reconcile_metadata.go
+++ b/pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_reconcile_metadata.go
@@ -58,6 +58,13 @@ func (r *metaDataReconciler) reconcile(ctx context.Context, workspace *tenancyv1
 		}
 	}
 
+	if workspace.Status.Phase == tenancyv1alpha1.ClusterWorkspacePhaseReady {
+		if _, found := workspace.Annotations[tenancyv1alpha1.ClusterWorkspaceOwnerAnnotationKey]; found {
+			delete(workspace.Annotations, tenancyv1alpha1.ClusterWorkspaceOwnerAnnotationKey)
+			changed = true
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_reconcile_metadata.go
```diff
@@ -58,6 +58,16 @@ func (r *metaDataReconciler) reconcile(ctx context.Context, workspace *tenancyv1
...
 	}
 
+	if workspace.Status.Phase == tenancyv1alpha1.ClusterWorkspacePhaseReady {
+		if _, found := workspace.Annotations[tenancyv1alpha1.ClusterWorkspaceOwnerAnnotationKey]; found && len(workspace.Annotations) == 1 {
+			workspace.Annotations = nil
+			changed = true
+		} else if found {
+			delete(workspace.Annotations, tenancyv1alpha1.ClusterWorkspaceOwnerAnnotationKey)
+			changed = true
+		}
```
s-urbaniak, line 68: just a nit, not sure if this reads better:
```
		if _, found := workspace.Annotations[tenancyv1alpha1.ClusterWorkspaceOwnerAnnotationKey]; found {
			delete(workspace.Annotations, tenancyv1alpha1.ClusterWorkspaceOwnerAnnotationKey)
			changed = true
		}

		if len(workspace.Annotations) == 0 {
			workspace.Annotations = nil
			changed = true
		}
```

s-urbaniak: please gdoc why setting the annotation map to nil is necessary.

sttts: it's not necessary actually. Removed.

PR #1503, opened 2022-07-12
admission/clusteworkspacetypeexists: in-place validation, fixes

sttts:
First step of https://github.com/kcp-dev/kcp/pull/1474.

- non-controller implementation of clusterworktypeexists admission
  - with that fixed child/parent invariant checking
  - mutable types
  - renamed `allowedChildTypes` to `limitAllowedChildren` as they do NOT list the allowed types, but only limit the allowed type
  - turned `any: false` (which was also omitted when false, not good), to explicit `none: true`.
  - this also fixes audit. Looks like something was hotlooping in the type controller https://github.com/kcp-dev/kcp/pull/1441

Note: we can readd conditions and status if we want to, but purely for informational purpose. The 100 lines of alogrithm in admission now defines the behaviour and easy to reason about.


davidfestal (approved):
Approved with a small suggestion

sttts:
user home workspace flake

PR #1504, opened 2022-07-12
Part 2: Scoping controllers in pkg/server

varshaprasad96:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/reconciler/tenancy/bootstrap/bootstrap_controller.go
+++ b/pkg/reconciler/tenancy/bootstrap/bootstrap_controller.go
@@ -42,6 +42,7 @@ import (
 	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
 	tenancyinformer "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/tenancy/v1alpha1"
 	tenancylister "github.com/kcp-dev/kcp/pkg/client/listers/tenancy/v1alpha1"
+	"github.com/kcp-dev/logicalcluster"
 )
 
 const (
@@ -51,10 +52,10 @@ const (
 func NewController(
 	dynamicCLient dynamic.ClusterInterface,
 	crdClusterClient apiextensionclientset.ClusterInterface,
-	kcpClusterClient kcpclient.ClusterInterface,
+	kcpClusterClient kcpclient.Interface,
 	workspaceInformer tenancyinformer.ClusterWorkspaceInformer,
 	workspaceType tenancyv1alpha1.ClusterWorkspaceTypeReference,
-	bootstrap func(context.Context, discovery.DiscoveryInterface, dynamic.Interface, kcpclient.Interface) error,
+	bootstrap func(context.Context, discovery.DiscoveryInterface, dynamic.Interface, kcpclient.Interface, logicalcluster.Name) error,
 ) (*controller, error) {
 	controllerName := fmt.Sprintf("%s-%s", controllerNameBase, workspaceType)
 	queue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), controllerName)
@@ -90,14 +91,14 @@ type controller struct {
 
 	dynamicClient dynamic.ClusterInterface
 	crdClient     apiextensionclientset.ClusterInterface
-	kcpClient     kcpclient.ClusterInterface
+	kcpClient     kcpclient.Interface
 
 	workspaceLister tenancylister.ClusterWorkspaceLister
 
 	syncChecks []cache.InformerSynced
 
 	workspaceType tenancyv1alpha1.ClusterWorkspaceTypeReference
-	bootstrap     func(context.Context, discovery.DiscoveryInterface, dynamic.Interface, kcpclient.Interface) error
+	bootstrap     func(context.Context, discovery.DiscoveryInterface, dynamic.Interface, kcpclient.Interface, logicalcluster.Name) error
 }
 
 func (c *controller) enqueue(obj interface{}) {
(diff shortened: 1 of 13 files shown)


pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_reconcile.go
```diff
@@ -49,7 +49,7 @@ func (c *Controller) reconcile(ctx context.Context, ws *tenancyv1alpha1.ClusterW
 			listShards: c.rootWorkspaceShardLister.List,
 		},
 		&phaseReconciler{
-			getShardWithQuorum: c.kcpClient.Cluster(tenancyv1alpha1.RootCluster).TenancyV1alpha1().ClusterWorkspaceShards().Get,
```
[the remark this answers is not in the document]
ncdc: I agree this is a bit of a backwards change, but it does get us away from maintaining a fork of client-gen...

What if we did something like this?
```go
getShardWithQuorum: func(ctx context.Context, name string, options metav1.GetOptions) (*tenancyv1alpha1.ClusterWorkspaceShard, error) {
    rootCtx := logicalcluster.WithCluster(ctx, tenancyv1alpha1.RootCluster)
    return c.kcpClient.TenancyV1alpha1().ClusterWorkspaceShards().Get(rootCtx, name, options)
}
```

This doesn't fix it, but it does replicate to a degree what we had before - in the `getShardWithQuorum` declaration, you are setting the scope, as opposed to doing it in the `phaseReconciler` where `getShardWithQuorum` is invoked.

The only other option I can think of (that doesn't involve generating a clientset with `.Cluster()`) would be something like a "KCPClientFactory" that could hand out scoped clients, but this would probably be less efficient because they wouldn't share the same underlying http.Client by default, unless we did some extra work there.

pkg/reconciler/tenancy/bootstrap/bootstrap_reconcile.go
```diff
@@ -47,7 +49,13 @@ func (c *controller) reconcile(ctx context.Context, workspace *tenancyv1alpha1.C
 	klog.Infof("Bootstrapping resources for org workspace %s, logical cluster %s", workspace.Name, wsClusterName)
 	bootstrapCtx, cancel := context.WithDeadline(ctx, time.Now().Add(time.Second*30)) // to not block the controller
 	defer cancel()
-	if err := c.bootstrap(bootstrapCtx, c.crdClient.Cluster(wsClusterName).Discovery(), c.dynamicClient.Cluster(wsClusterName), c.kcpClient.Cluster(wsClusterName)); err != nil {
+
+	clusterConfig := kcpclienthelper.ConfigWithCluster(c.config, wsClusterName)
```
sttts, line 53: this is a miss-namer. We usually call something "clusterSomething" if it is NOT attached to a logical cluster. This one is more like `wsConfig`.

pkg/reconciler/tenancy/bootstrap/bootstrap_reconcile.go
```diff
@@ -47,7 +49,13 @@ func (c *controller) reconcile(ctx context.Context, workspace *tenancyv1alpha1.C
...
+	crdWsClient, err := clientset.NewForConfig(clusterConfig)
```
sttts, line 54: please qualify the clientset, i.e. import alias `crdclientset` or so.

pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_controller.go
```diff
@@ -108,7 +108,7 @@ func NewController(
 type Controller struct {
 	queue workqueue.RateLimitingInterface
 
-	kcpClient        kcpclient.ClusterInterface
+	kcpClient        kcpclient.Interface
```
sttts, line 111: kcpClusterClient. Now without proper types we have to be more strict in naming.

pkg/reconciler/tenancy/bootstrap/bootstrap_controller.go
```diff
@@ -49,9 +51,10 @@ const (
 )
 
 func NewController(
-	dynamicCLient dynamic.ClusterInterface,
-	crdClusterClient apiextensionclientset.ClusterInterface,
-	kcpClusterClient kcpclient.ClusterInterface,
+	config *rest.Config,
+	dynamicCLient dynamic.Interface,
```
sttts, line 55: dynamicClusterClient ?

pkg/reconciler/tenancy/bootstrap/bootstrap_controller.go
```diff
@@ -49,9 +51,10 @@ const (
 )
 
 func NewController(
-	dynamicCLient dynamic.ClusterInterface,
-	crdClusterClient apiextensionclientset.ClusterInterface,
-	kcpClusterClient kcpclient.ClusterInterface,
+	config *rest.Config,
```
sttts, line 54: baseConfig ?

pkg/reconciler/tenancy/bootstrap/bootstrap_controller.go
```diff
@@ -85,12 +89,12 @@ func NewController(
...
+	config         *rest.Config
+	queue          workqueue.RateLimitingInterface
 
-	queue workqueue.RateLimitingInterface
-
-	dynamicClient dynamic.ClusterInterface
-	crdClient     apiextensionclientset.ClusterInterface
-	kcpClient     kcpclient.ClusterInterface
+	dynamicClient    dynamic.Interface
+	crdClient        apiextensionclientset.Interface
```
sttts, line 96: crdClusterClient

pkg/server/controllers.go
```diff
@@ -362,18 +362,19 @@ func (s *Server) installWorkloadResourceScheduler(ctx context.Context, config *r
 }
 
 func (s *Server) installWorkspaceScheduler(ctx context.Context, config *rest.Config) error {
-	config = rest.AddUserAgent(rest.CopyConfig(config), "kcp-workspace-scheduler")
-	kcpClusterClient, err := kcpclient.NewClusterForConfig(config)
+	config = kcpclienthelper.NewClusterConfig(rest.AddUserAgent(rest.CopyConfig(config), "kcp-workspace-scheduler"))
```
varshaprasad96, line 365: leaving the variable name `config` as is, since it could point towards lcluster or base depending on context.

pkg/reconciler/tenancy/bootstrap/bootstrap_controller.go
varshaprasad96: @sttts Modified this to be `crdClusterClient` and `dynamicClient`, but thinking about it - this could point towards the base cluster or cluster depending on the context which is passed. Do we follow the `*ClusterClient` convention hereon for all controllers, or shall we keep this as is.

Issue #1482, opened 2022-07-11
Soft Finalizers not honoured when workload scheduling is updated


philbrookes:
**Describe the bug**
When a workload is scheduled to a workload cluster (cluster1), with a soft finalizer created for cluster1 and then the workload scheduling is changed to a different cluster (cluster2). The workload is immediately deleted from cluster1 rather than waiting for the soft finalizer for cluster1 to be removed. 

**To Reproduce**
1. Disable automatic scheduling on a namespace
2. set label to schedule namespace to cluster1
3. create a service resource in that namespace
4. set a soft finalizer on that service for cluster1
5. change the label on the namespace to schedule to cluster2
6. see the service deleted from cluster1

**Expected behavior**
I would expect the service in the above scenario to stay on cluster1 until the soft finalizer is removed.

**Additional context**
This bug is discovered in 0.5.0.

philbrookes:
cc @jmprusi this is a follow on from our discussion on slack

PR #1515, opened 2022-07-13
Don't unschedule a resource if blocked by a cluster or syncer finalizer

jmprusi:
This PR changes the resource reconciler's behavior to avoid removing the sync state from a resource if a finalizer, either syncer or cluster, exists. 

fixes: https://github.com/kcp-dev/kcp/issues/1482

### DIFF
--- a/pkg/reconciler/workload/resource/resource_reconcile.go
+++ b/pkg/reconciler/workload/resource/resource_reconcile.go
@@ -35,6 +35,7 @@ import (
 	"k8s.io/klog/v2"
 
 	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/syncer/shared"
 )
 
 // reconcileResource is responsible for setting the cluster for a resource of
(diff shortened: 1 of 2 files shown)


pkg/reconciler/workload/resource/resource_reconcile.go
```diff
@@ -134,10 +135,25 @@ func computePlacement(ns *corev1.Namespace, obj metav1.Object) (annotationPatch
 	annotationPatch = map[string]interface{}{}
 	labelPatch = map[string]interface{}{}
 	for _, loc := range objLocations.Difference(nsLocations).List() {
-		// location was removed from namespace, but is still on the object
```
sttts, line 137: leave comment?

sttts:
@qiujian16 @jmprusi afaik we have no e2e test that removes something from a sync target. Can we add it?

qiujian16:
I think we have that tests in scheduling test e2e, we evict a cluster which remove the service. I think we want to manually add a soft finalizer on that service and ensure it is not removed?

jmprusi:
> I think we have that tests in scheduling test e2e, we evict a cluster which remove the service. I think we want to manually add a soft finalizer on that service and ensure it is not removed?

Yes, similar to the syncer e2e tests, where we add a soft finalizer and make sure that the resource is still there

PR #1516, opened 2022-07-14
admission/clusterworkspaceexists: use builder pattern in tests

sttts:
800 lines less or so. Brings back our sanity in this package.

### DIFF
--- a/pkg/admission/clusterworkspacetypeexists/admission_test.go
+++ b/pkg/admission/clusterworkspacetypeexists/admission_test.go
@@ -26,7 +26,6 @@ import (
 	"github.com/kcp-dev/logicalcluster"
 	"github.com/stretchr/testify/require"
 
-	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


pkg/admission/clusterworkspacetypeexists/admission_test.go
```diff
@@ -1551,9 +827,70 @@ func (b builder) allowingParent(qualifiedName string) builder {
 	return b
 }
 
+func (b builder) allowingChild(qualifiedName string) builder {
```
s-urbaniak, line 830: no pointer receiver/return? is this by purpose? effectively, we're mutating the builder.

sttts: the only field in builder is a pointer. So this is essentially `type builder *T`. With the later we would use a receiver pointer. But here, double pointers don't make sense.

sttts:
/skip e2e-multiple-runs

This is changing a unit-test only.

PR #1529, opened 2022-07-14
Adding CEL Validation for Permission Claim 

### LINKED ISSUE #1331: validate that the only resources without identity are known core types.
Patch openapi such that if the identity hash is not set, then only the core types can be GRs that are set for a permission claim

shawn-hurley:
## Summary
Adding CEL Validation as an option for the APIServer

For Permission Claims adding validations:
1. When group is empty, then only allow resource in [configmaps,secrets,namespaces,serviceaccounts]
2. If the group is not empty, then make sure we have a specific identity hash.
## Related issue(s)

Fixes #1331


test/e2e/apibinding/apibinding_test.go
```diff
@@ -304,3 +305,102 @@ func apiexportVWConfig(t *testing.T, kubeconfig clientcmdapi.Config, clusterName
 
 	return rest.AddUserAgent(rest.CopyConfig(config), t.Name())
 }
+
+// TestAPIBindingPermissionClaimCELValidation will validate the permission claims for an otherwise valid APIBinding.
+func TestAPIBindingPermissionClaimCELValidation(t *testing.T) {
```
sttts, line 310: wondering, is there a simple evaluator such that one can unit test CEL expressions? This is really heavy to do it in e2e.

config/crds/apis.kcp.dev_apibindings.yaml
```diff
@@ -69,6 +69,12 @@ spec:
                   required:
                   - resource
                   type: object
+                  x-kubernetes-validations:
+                  - message: if claim for core group must only be one of configmaps,
```
sttts, line 73: core group does not really matter here. You can claim Services or Pods. I think the only restriction we can and should have is about the identity.

shawn-hurley: We may want to update this in the future. But if we are providing things like RBAC, that will not be the core group and will not have an identityHash. I believe that for now let's keep as is (most restrictive) and we can open it up as that path is usually more accessible. Thoughts?

Makefile
```diff
@@ -225,8 +227,10 @@ ifdef USE_GOTESTSUM
 test: $(GOTESTSUM)
 endif
 test: WHAT ?= ./...
+# We will need to move into the sub package, of pkg/apis to run those tests.
 test:
 	$(GO_TEST) -race -count $(COUNT) -coverprofile=coverage.txt -covermode=atomic $(TEST_ARGS) $$(go list "$(WHAT)" | grep -v ./test/e2e/)
+	cd pkg/apis && $(GO_TEST_SUB) -race -count $(COUNT) -coverprofile=coverage.txt -covermode=atomic $(TEST_ARGS) $$(go list "$(WHAT)" | grep -v ./test/e2e/)
```
sttts, line 233: no need for the grep

Makefile
```diff
@@ -160,8 +160,10 @@ GOTESTSUM_ARGS += --junitfile=$(ARTIFACT_DIR)/junit.xml
 endif
 
 GO_TEST = go test
+GO_TEST_SUB = go test
 ifdef USE_GOTESTSUM
 GO_TEST = $(GOTESTSUM) $(GOTESTSUM_ARGS) --
+GO_TEST_SUB = ../../${GO_TEST}
```
sttts, line 166: this is hacky. Just reference the toplevel dir somehow, or make TOOLS_DIR absolute

pkg/apis/apis/v1alpha1/api_cel_test.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	if len(format) != 0 {
+		result.ValueValidation = &schema.ValueValidation{
+			Format: format,
+		}
+	}
+	return result
+}
+
+func getValidatorForPermissionClaim(t *testing.T, crdFilePath string, specClaimName string) (*cel.Validator, schema.Structural) {
```
sttts, line 174: we will get more uses of this func. Why not make it generic enough?

I.e. make the json path inside the schema a parameter.

pkg/apis/apis/v1alpha1/api_cel_test.go
```diff
@@ -0,0 +1,216 @@
...
+	s := schema.Structural{
+		Generic: schema.Generic{
+			Type: "object",
+		},
+		Properties: map[string]schema.Structural{
+			"group":        primitiveType("string", ""),
+			"resource":     primitiveType("string", ""),
+			"identityHash": primitiveType("string", ""),
+		},
+	}
```
sttts, line 212: This schema can be derived, some random example:
```
	openapiSchema, err := apiResourceVersion.GetSchema()
	if err != nil {
		return nil, err
	}
	if err := apiextensionsv1.Convert_v1_JSONSchemaProps_To_apiextensions_JSONSchemaProps(openapiSchema, internalSchema, nil); err != nil {
		return nil, fmt.Errorf("failed converting CRD validation to internal version: %w", err)
	}
	structuralSchema, err := structuralschema.NewStructural(internalSchema)
```

pkg/apis/tenancy/initialization/utils_test.go
```diff
@@ -25,14 +25,6 @@ import (
 	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
 )
 
-func TestClusterWorkspaceInitializerLabelPrefix(t *testing.T) {
```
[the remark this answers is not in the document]
sttts: ClusterWorkspaceInitializerLabelPrefix is ending a slash. So yes, this test is not valid anymore.

pkg/apis/apis/v1alpha1/api_cel_test.go
shawn-hurley: One of the big reasons is that I don't think that it needs to do it, and why spend the extra time because I do have other things that I have to get done. 

Are there plans for other CELs? We would have to have some different thing to get the schema, and it would probably need to be broken up. 

I wonder if this is a follow-up?

test/e2e/apibinding/apibinding_test.go
shawn-hurley: updated

pkg/apis/apis/v1alpha1/api_cel_test.go
shawn-hurley: I am sorry this code block is not clear to me what needs to be done.

PermissionClaim does not have the GetSchema() call

Makefile
shawn-hurley: updated

Makefile
shawn-hurley: updated

pkg/apis/apis/v1alpha1/api_cel_test.go
shawn-hurley: same for this will be useful for a more generic getting of schemas and CELs. Will add it as a follow-up when needed.

config/crds/apis.kcp.dev_apibindings.yaml-patch
```diff
@@ -6,3 +6,9 @@
   path: /spec/versions/name=v1alpha1/schema/openAPIV3Schema/properties/spec/properties/reference/properties/workspace/oneOf
   value:
   - required: ["path"]
+- op: add
+  path: /spec/versions/name=v1alpha1/schema/openAPIV3Schema/properties/spec/properties/acceptedPermissionClaims/items/x-kubernetes-validations
+  value:
+    - rule: |
+        (((has(self.group) && self.group == '') || !(has(self.group))) && self.resource in ['configmaps', 'namespaces', 'secrets', 'serviceaccounts']) || ((has(self.group) && self.group != '') && (has(self.identityHash) && self.identityHash != ''))
```
sttts, line 13: why is that a giant expression? can be split up, giving better messages, and would be much easier to read.

config/crds/apis.kcp.dev_apibindings.yaml-patch
```diff
@@ -6,3 +6,9 @@
...
+      message: "if claim for core group must only be one of configmaps, namespaes, secrets, serviceaccounts' or if group is specified, must have an identityHash"
```
sttts, line 14:
```suggestion
      message: "if claim for core group must only be one of configmaps, namespaces, secrets, serviceaccounts' or if group is specified, must have an identityHash"
```

config/crds/apis.kcp.dev_apibindings.yaml-patch
shawn-hurley: I do not have a way to do it because the rules are ANDed together AFAICT

sttts: I don't see why ANDing should stop that from working. Just turn it into a disjunctive normal form.
```
- rule: "(has(self.group) && self.group != '') || self.resource in ['configmaps', 'namespaces', 'secrets', 'serviceaccounts']"
  message: .resource must be one of configmaps, namespaces, secrets, serviceaccounts if .group is empty.
- rule: "!(has(self.group) && self.group != '') || (has(self.identityHash) && self.identityHash != '')
  message: .identityHash must be set if .group is not empty.
```

config/crds/apis.kcp.dev_apibindings.yaml
```diff
@@ -69,6 +69,14 @@ spec:
...
                   - resource
                   type: object
+                  x-kubernetes-validations:
+                  - message: .resource must be one of configmaps, namespaces, secrets,
+                      serviceaccounts if .group is empty.
+                    rule: (has(self.group) && self.group != '') || self.resource in
+                      ['configmaps', 'namespaces', 'secrets', 'serviceaccounts']
+                  - message: .identityHash must be set if .group is not empty.
+                    rule: '!(has(self.group) && self.group != '''') || (has(self.identityHash)
+                      && self.identityHash != '''')'
```
sttts, line 79: much better

shawn-hurley: I honestly have never done logic like this or at least forgot that it was possible. Thanks for teaching!

config/crds/apis.kcp.dev_apibindings.yaml
```diff
@@ -69,6 +69,14 @@ spec:
                   required:
                   - resource
                   type: object
+                  x-kubernetes-validations:
+                  - message: .resource must be one of configmaps, namespaces, secrets,
+                      serviceaccounts if .group is empty.
+                    rule: (has(self.group) && self.group != '') || self.resource in
+                      ['configmaps', 'namespaces', 'secrets', 'serviceaccounts']
+                  - message: .identityHash must be set if .group is not empty.
+                    rule: '!(has(self.group) && self.group != '''') || (has(self.identityHash)
```
stevekuznetsov, line 78: What's up with the `''''`?

config/crds/apis.kcp.dev_apibindings.yaml
sttts: https://en.wikipedia.org/wiki/Conjunctive_normal_form

PR #1530, opened 2022-07-15
server/home: create home workspace with final URL on ~ access

sttts:
Solving this TODO:
```
// TODO: The way we build this URL here is not fully compatible with sharding long-term:
// - short term (what is done now): use genericConfig.ExternalAddress
// - medium term: shards will know their name eventually, and with that we can lookup the external URL of the shard
// - long-term: create home workspace already on `~` access in order to make scheduling happen and then we know the real external URL.
```

### DIFF
--- a/pkg/server/home_workspaces.go
+++ b/pkg/server/home_workspaces.go
@@ -23,7 +23,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"net/url"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


pkg/server/home_workspaces.go
```diff
@@ -291,79 +291,39 @@ func (h *homeWorkspaceHandler) ServeHTTP(rw http.ResponseWriter, req *http.Reque
 			// fall through because either not ready or RBAC objects missing
 		}
 
-		// Test if the user has the right to see his Home workspace even though it doesn't exists
-		// => test the get verb on the clusterworkspaces/workspace subresource named ~ in the root workspace.
+		lcluster.Name = homeLogicalClusterName
```
davidfestal, line 294: Why don't we use the `searchForWorkspaceAndRBACInLocalInformers` function directly here in this case as well, and also send a retry when it should be done (typically when workspace and rbac resources are already created, but Home workspace is still initializing) ?

davidfestal: This probably also works, but I assume is less optimized.

sttts: The current code path is proven to work. This is to unblock 0.6. We can improve later in a follow-up.

Also note that searchForWorkspaceAndRBACInLocalInformers is not waiting for readiness. That was the root cause for the prior PR here. So it is not that simple.

PR #1538, opened 2022-07-15
separate server config initialization from run

deitch:
Signed-off-by: Avi Deitcher <avi@deitcher.net>

## Summary

* Create a `Config`, which is higher level abstraction than `Options`, lower than `Server`
* Moves all of the configuration work from `server.Run()` into `config.Complete()`

The purpose of this is to make `Server` a more flexible, composable and useful component. Currently, `Run()` does an enormous amount of setup, which, in turn, means that you cannot choose which components to use or not. It also makes "kcp-as-a-library" painfully heavy to use.

This structure was suggested by @sttts as a first step. Nothing should change in functionality. Once this is in place, we can start to break it apart further, by having `Config` options that are enabled or not, which, in turn, should enable us to tweak the knobs.

This first PR changes the API to something like:

```go
opts := NewOptions()
completedOpts := opts.Complete()
config := NewConfig(completedOpts)
completedConfig := config.Complete()
server := NewServer(completedConfig)
server.Run()
```

This already isolates `Run()` just to starting things, other than those that were set up to run in post-start-hooks. It isn't 100% isolation yet, but a big step forward.

The next API steps would be (in no particular order):

* plug `server` into other listeners, e.g. an already running http server
* move the `install*Controller()` funcs into config, so that `Server` itself is ignorant of what is installed, which in turn enables:
* different configs, lightweight for the basics, heavier for all of the kcp goodies. Either way, you end up with a `CompletedConfig` which is passed to a `NewServer()`

### DIFF
--- a/docs/developers/library-structure.md
+++ b/docs/developers/library-structure.md
@@ -1,0 +1,39 @@
+# Library Structure
+
+The kcp library is set up and initialized in several phases. This document describes those phases and the responsibilities of each.
+
+```go
+opts := options.NewOptions()
+completedOptions := opts.Complete()
+conf := config.NewConfig(completedOptions)
+completedConfig := conf.Complete()
+srv := server.NewServer(completedConfig)
+srv.Init()
+srv.Run(ctx)
+```
+
+## Options
+
+Options parses basic configuration options, such as flags or environment variables. `NewOptions()` does default parsing
+of environment variables and CLI flags and sets defaults. You then can change options properties
+as you see fit.
+
+With `Options` in hand, you need to run `opts.Complete()`. This ensures that any modifications necessary,
+such as defaults, are in place.
+
+## Config
+
+Config includes all of the structures required for the server to operate properly. It can include handlers,
+callbacks, and other structures. These settings are informed by the values of `Options`, and so
+take `Options` as a parameter.
+
+With `Config` in hand, you need to run `conf.Complete()`. This ensures that any modifications necessary,
+such as defaults, are in place.
+
+## Server
+
+Server is the actual structure that handles the inbound requests, passing them through RBAC, storing data, invoking handlers, etc.
+It is, in effect, the API server.
+
+To create a Server, you must call `NewServer()`, passing it the `CompletedOptions`. Then you initialize it with `Init()`, and finally
+you `Run()` it.
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -26,6 +26,7 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"math/big"
@@ -43,7 +44,9 @@ import (
 )
 
 type Server struct {
-	Dir string
+	Dir        string
+	config     *embed.Config
+	clientInfo *ClientInfo
 }
 
 type ClientInfo struct {
@@ -55,7 +58,41 @@ type ClientInfo struct {
 	TrustedCAFile string
 }
 
-func (s *Server) Run(ctx context.Context, peerPort, clientPort string, listenMetricsURLs []url.URL, walSizeBytes, quotaBackendBytes int64, forceNewCluster bool) (ClientInfo, error) {
+func (s *Server) Run(ctx context.Context) error {
+	klog.Info("Starting embedded etcd server")
+	if s.config == nil {
+		return errors.New("must Init() before Run()")
+	}
+	e, err := embed.StartEtcd(s.config)
+	if err != nil {
+		return err
+	}
+	// Shutdown when context is closed
+	go func() {
+		<-ctx.Done()
+		e.Close()
+	}()
+
+	select {
+	case <-e.Server.ReadyNotify():
+		return nil
+	case <-time.After(60 * time.Second):
+		e.Server.Stop() // trigger a shutdown
+		return fmt.Errorf("server took too long to start")
+	case e := <-e.Err():
+		return e
+	}
+}
+
+func (s *Server) ClientInfo() (ClientInfo, error) {
+	if s.clientInfo == nil {
+		return ClientInfo{}, errors.New("must Init() before retrieving ClientInfo()")
+	}
+	return *s.clientInfo, nil
+}
+
+// Init initialize an embedded etcd server, getting everything ready except for starting it
+func (s *Server) Init(peerPort, clientPort string, listenMetricsURLs []url.URL, walSizeBytes, quotaBackendBytes int64, forceNewCluster bool) (ClientInfo, error) {
 	klog.Info("Creating embedded etcd server")
 	if walSizeBytes != 0 {
 		wal.SegmentSizeBytes = walSizeBytes
@@ -106,36 +143,21 @@ func (s *Server) Run(ctx context.Context, peerPort, clientPort string, listenMet
 		cfg.QuotaBackendBytes = quotaBackendBytes
 	}
 
-	e, err := embed.StartEtcd(cfg)
-	if err != nil {
-		return ClientInfo{}, err
-	}
-	// Shutdown when context is closed
-	go func() {
-		<-ctx.Done()
-		e.Close()
-	}()
-
 	clientConfig, err := cfg.ClientTLSInfo.ClientConfig()
 	if err != nil {
 		return ClientInfo{}, err
 	}
 
-	select {
-	case <-e.Server.ReadyNotify():
-		return ClientInfo{
-			Endpoints:     []string{cfg.ACUrls[0].String()},
-			TLS:           clientConfig,
-			CertFile:      cfg.ClientTLSInfo.CertFile,
-			KeyFile:       cfg.ClientTLSInfo.KeyFile,
-			TrustedCAFile: cfg.ClientTLSInfo.TrustedCAFile,
-		}, nil
-	case <-time.After(60 * time.Second):
-		e.Server.Stop() // trigger a shutdown
-		return ClientInfo{}, fmt.Errorf("server took too long to start")
-	case e := <-e.Err():
-		return ClientInfo{}, e
+	s.config = cfg
+
+	s.clientInfo = &ClientInfo{
+		Endpoints:     []string{cfg.ACUrls[0].String()},
+		TLS:           clientConfig,
+		CertFile:      cfg.ClientTLSInfo.CertFile,
+		KeyFile:       cfg.ClientTLSInfo.KeyFile,
+		TrustedCAFile: cfg.ClientTLSInfo.TrustedCAFile,
 	}
+	return *s.clientInfo, nil
 }
 
 func generateClientAndServerCerts(hosts []string, dir string) error {
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -1,0 +1,418 @@
+package server
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+	"reflect"
+
+	kcpadmissioninitializers "github.com/kcp-dev/kcp/pkg/admission/initializers"
+	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
+	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
+	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	kcpexternalversions "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
+	"github.com/kcp-dev/kcp/pkg/etcd"
+	"github.com/kcp-dev/kcp/pkg/indexers"
+	"github.com/kcp-dev/kcp/pkg/informer"
+	boostrap "github.com/kcp-dev/kcp/pkg/server/bootstrap"
+	"github.com/kcp-dev/kcp/pkg/server/options"
+	"github.com/kcp-dev/kcp/pkg/server/requestinfo"
+	"github.com/kcp-dev/logicalcluster"
+	etcdtypes "go.etcd.io/etcd/client/pkg/v3/types"
+	apiextensionsapiserver "k8s.io/apiextensions-apiserver/pkg/apiserver"
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	apiextensionsexternalversions "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apiserver/pkg/admission"
+	"k8s.io/apiserver/pkg/endpoints/filters"
+	genericapiserver "k8s.io/apiserver/pkg/server"
+	"k8s.io/apiserver/pkg/util/webhook"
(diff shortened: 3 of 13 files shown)


deitch:
There is a decent amount of passthrough from Options -> Config -> Server. I tried to abstract as much as I could into `Config`, and have `Server` hold onto that rather than `Options`, i.e. layered abstractions. I am not convinced that this is the best path, but if it works, I don't care much.

docs/developers/library-structure.md
```diff
@@ -0,0 +1,39 @@
...
+srv := server.NewServer(completedConfig)
+srv.Init()
+srv.Run(ctx)
+```
+
+## Options
+
+Options parses basic configuration options, such as flags or environment variables. `NewOptions()` does default parsing
+of environment variables and CLI flags and sets defaults. You do not need to use `NewOptions()`; you
+can create an `options.Options` struct yourself and populate it.
```
sttts, line 19: would not suggest that. It will break people when new fields are added.

Rather flags and env vars should be optional and everybody should start with `NewOptions` and then override.

pkg/server/server.go
```diff
@@ -123,369 +114,116 @@ type preShutdownHookEntry struct {
 	hook genericapiserver.PreShutdownHookFunc
 }
 
+// Init initializes the KCP api-server. It does not start it
+func (s *Server) Init() error {
+	s.initialized = true
+	return nil
+}
```
sttts, line 121: This is not the PrepareRun pattern.

See what apiserver does. We should follow exactly the same pattern. Otherwise, we confuse the consumer by call the same concept differently and implementing it differently:

```
// preparedGenericAPIServer is a private wrapper that enforces a call of PrepareRun() before Run can be invoked.
type preparedGenericAPIServer struct {
	*GenericAPIServer
}

// PrepareRun does post API installation setup steps. It calls recursively the same function of the delegates.
func (s *GenericAPIServer) PrepareRun() preparedGenericAPIServer {
	s.delegationTarget.PrepareRun()

	if s.openAPIConfig != nil && !s.skipOpenAPIInstallation {
		s.OpenAPIVersionedService, s.StaticOpenAPISpec = routes.OpenAPI{
			Config: s.openAPIConfig,
		}.InstallV2(s.Handler.GoRestfulContainer, s.Handler.NonGoRestfulMux)
		if utilfeature.DefaultFeatureGate.Enabled(features.OpenAPIV3) {
			s.OpenAPIV3VersionedService = routes.OpenAPI{
				Config: s.openAPIConfig,
			}.InstallV3(s.Handler.GoRestfulContainer, s.Handler.NonGoRestfulMux)
		}
	}

	s.installHealthz()
	s.installLivez()

	// as soon as shutdown is initiated, readiness should start failing
	readinessStopCh := s.lifecycleSignals.ShutdownInitiated.Signaled()
	err := s.addReadyzShutdownCheck(readinessStopCh)
	if err != nil {
		klog.Errorf("Failed to install readyz shutdown check %s", err)
	}
	s.installReadyz()

	// Register audit backend preShutdownHook.
	if s.AuditBackend != nil {
		err := s.AddPreShutdownHook("audit-backend", func() error {
			s.AuditBackend.Shutdown()
			return nil
		})
		if err != nil {
			klog.Errorf("Failed to add pre-shutdown hook for audit-backend %s", err)
		}
	}

	return preparedGenericAPIServer{s}
}
```

pkg/server/server.go
```diff
@@ -123,369 +114,116 @@ type preShutdownHookEntry struct {
...
-
-	if err := s.options.Authorization.ApplyTo(genericConfig, s.kubeSharedInformerFactory, s.kcpSharedInformerFactory); err != nil {
-		return err
-	}
-	kcpAdminToken, shardAdminToken, shardAdminTokenHash, err := s.options.AdminAuthentication.ApplyTo(genericConfig)
-	if err != nil {
-		return err
+	if !s.initialized {
+		return errors.New("must initialize server with srv.Init() prior to Run()")
 	}
```
sttts, line 127: PrepareRun uses types for this.

pkg/server/handler.go
```diff
@@ -67,37 +67,19 @@ func init() {
...
-	userAgentContextKey userAgentContextKeyType = iota
-)
-
 // WithAcceptHeader makes the Accept header available for code in the handler chain. It is needed for
 // Wildcard requests, when finding the CRD with a common schema. For PartialObjectMeta requests we cand
 // weaken the schema requirement and allow different schemas across workspaces.
 func WithAcceptHeader(apiHandler http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-		ctx := context.WithValue(req.Context(), acceptHeaderContextKey, req.Header.Get("Accept"))
+		ctx := context.WithValue(req.Context(), AcceptHeaderContextKey, req.Header.Get("Accept"))
```
sttts, line 75: I don't think we want to leak this.

pkg/server/const.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+type (
+	acceptHeaderContextKeyType int
+	userAgentContextKeyType    int
+)
+
+const (
+	// clusterKey is the context key for the request namespace.
+	AcceptHeaderContextKey acceptHeaderContextKeyType = iota
```
sttts, line 15: don't make these public please. They are no means for consumption. If we have to open them, we add `WithAcceptHeader`and `AcceptHeaderFrom`, without leaking the key.

pkg/server/config.go
```diff
@@ -0,0 +1,494 @@
...
+func (c *CompletedConfig) VirtualEnabled() bool {
+	return c.options.Virtual.Enabled
+}
+func (c *CompletedConfig) Virtual() options.Virtual {
+	return c.options.Virtual
+}
+func (c *CompletedConfig) Controllers() options.Controllers {
+	return c.options.Controllers
+}
+func (c *CompletedConfig) Authorization() options.Authorization {
```
sttts, line 111: all these accessors make customization hard and don't add any abstraction.

pkg/server/config.go
```diff
@@ -0,0 +1,494 @@
...
+	"k8s.io/client-go/tools/clusters"
+	"k8s.io/klog/v2"
+	"k8s.io/kubernetes/pkg/genericcontrolplane"
+)
+
+type Config struct {
+	options *options.CompletedOptions
+}
+
+type CompletedConfig struct {
```
sttts, line 53: this is not the completion pattern. Compare:

```
type completedConfig struct {
	GenericConfig genericapiserver.CompletedConfig
	ExtraConfig   *ExtraConfig
}

type CompletedConfig struct {
	// Embed a private pointer that cannot be instantiated outside of this package.
	*completedConfig
}

// Complete fills in any fields not set that are required to have valid data. It's mutating the receiver.
func (c *AppsServerConfig) Complete() completedConfig {
	cfg := completedConfig{
		c.GenericConfig.Complete(),
		&c.ExtraConfig,
	}

	return cfg
}
```

pkg/server/config.go
```diff
@@ -0,0 +1,494 @@
...
+	EtcdServer                            *etcd.Server
+	kcpSharedInformerFactory              kcpexternalversions.SharedInformerFactory
+	kubeSharedInformerFactory             coreexternalversions.SharedInformerFactory
+	apiextensionsSharedInformerFactory    apiextensionsexternalversions.SharedInformerFactory
+	dynamicDiscoverySharedInformerFactory *informer.DynamicDiscoverySharedInformerFactory
+
+	identityConfig        *rest.Config
+	serverChain           *genericcontrolplane.ServerChain
+	server                *genericapiserver.GenericAPIServer
```
sttts, line 62: this is not config level, but server. See my other comment that this struct is actually what Server should be. You lack the config level.

docs/developers/library-structure.md
deitch: Ok, changed.

pkg/server/handler.go
deitch: I guess it doesn't matter. When I tried to move `Config` into its own package, the same way `Options` is, I had to expose it. I gave up on that for now, so we can make it private again.

pkg/server/config.go
deitch: While I do agree, these are necessary if `Config` lives outside of the same package as the thing that uses it (i.e. `Server`). `Options` live in `pkg/server/options`, so if that is one layer below `Config`, which is one layer below `Server`, wouldn't they eventually live in `pkg/server/config`? I don't mind removing them for now, but we end up with an uneven abstraction, which makes it harder for people to grasp.

deitch:
@sttts a lot of your desired pattern still is not clear. Based on your recommended break line, setting up things like `kcpSharedInformerFactory` and `kubeSharedInformerFactory` (and all of the other factories) belong as part of `NewConfig()`. Fine. We make them properties of `Config`:

```go
type Config struct {
	kcpSharedInformerFactory              kcpexternalversions.SharedInformerFactory
	kubeSharedInformerFactory             coreexternalversions.SharedInformerFactory
        // etc etc
}
```

Now `NewServer()` needs access to those factories, so it can use them. But those are properties only of `Config`, whereas `NewServer()` only accepts a `CompletedConfig`. That leaves a few options, none too great:

* embed `Config` in `CompletedConfig` and then use the accessor to retrieve it. This actually is fine - you really are just decorating `Config` - but you would need to use accessors for that, which you do not like (see above).
* duplicate `kcpSharedInformerFactory` and `kubeSharedInformerFactory` (and all of the other properties) in `CompletedConfig`, populating them in `Complete()`. That is not DRY at all, quite messy, almost guaranteed to break something in the future, not to mention not quite doing what `Complete()` is supposed to do.

deitch:
It gets even harder to separate. `dynamicDiscoverySharedInformerFactory` is an informer factory, which, by the recommended pattern, should be created when in `NewConfig()` along with the others.

Except that `dynamicDiscoverySharedInformerFactory` depends on `metadataClusterClient`, which is created by:

```go
	metadataClusterClient, err := metadataclient.NewDynamicMetadataClusterClientForConfig(server.LoopbackClientConfig)
```

So you need `server.LoopbackClientConfig`, which means you need `server`, which only comes later in `NewServer()`.

Oy.

deitch:
One thing that particularly is deserving of focus is when the factories should be initiated. Currently, they are mostly in `NewConfig()`, with one in `NewServer()`. The latter is there because creating it `informer.NewDynamicDiscoverySharedInformerFactory()` needs a parameter `server.LoopbackClientConfig`, where `server` is taken from `serverChain`, both of which definitely should be in `NewServer()`.

This leaves a few possibilities:

* keep the split (most of the factories in `NewConfig()`, this last one in `NewServer()`)
* somehow refactor `informer.NewDynamicDiscoverySharedInformerFactory()` so that it doesn't depend on the info from `server.LoopbackClientConfig`
* create the factories entirely in `NewServer()` and _not_ in `NewConfig()`. Would this even be correct? 

The latter would drag other dependencies into `NewServer()`, but these may make more sense there. For example from `NewCluster()`:

```go
	kcpClusterClient, err := kcpclient.NewClusterForConfig(c.identityConfig) // this is now generic to be used for all kcp API groups
	if err != nil {
		return c, err
	}
	c.kcpClusterClient = kcpClusterClient
	kcpClient := kcpClusterClient.Cluster(logicalcluster.Wildcard)
	c.kcpSharedInformerFactory = kcpexternalversions.NewSharedInformerFactoryWithOptions(
		kcpClient,
		resyncPeriod,
		kcpexternalversions.WithExtraClusterScopedIndexers(indexers.ClusterScoped()),
		kcpexternalversions.WithExtraNamespaceScopedIndexers(indexers.NamespaceScoped()),
	)
```

Moving the above would mean moving `kcpClusterClient` and `kcpClient` (and others) into `NewServer()`. But maybe that make more sense? However, doing so would mean we could not create `apiBindingAwareCRDLister` until `NewServer()`, as well as `apiHandler`, which means `genericConfig.BuildHandlerChainFunc` could not be created, and that definitely belongs in `NewConfig()` 🤷‍♂️

docs/developers/library-structure.md
```diff
@@ -0,0 +1,39 @@
...
+completedConfig := conf.Complete()
+srv := server.NewServer(completedConfig)
+srv.Init()
+srv.Run(ctx)
+```
+
+## Options
+
+Options parses basic configuration options, such as flags or environment variables. `NewOptions()` does default parsing
+of environment variables and CLI flags and sets defaults. You then can change options properties
```
sttts, line 18: this is not entirely correct. `NewOptions` only returns default value. `options.AddFlags` and then `pflag.ParseFlags` will do the command line parsing.

pkg/etcd/etcd.go
```diff
@@ -43,7 +44,9 @@ import (
 )
 
 type Server struct {
-	Dir string
+	Dir        string
+	config     *embed.Config
+	clientInfo *ClientInfo
 }
```
sttts, line 50: if this struct was called `Config`, and `Init` was called `NewServer` and it was returning a `Server` struct, it would match the other structure much better. Also `ClientInfo()` then wouldn't need to check whether `Init()` has been called.

pkg/server/config.go
```diff
@@ -0,0 +1,418 @@
+package server
```
sttts, line 1: all these files need boilerplate

pkg/server/config.go
```diff
@@ -0,0 +1,418 @@
...
+)
+
+type tokenAuth struct {
+	kcp       string
+	shard     string
+	shardHash []byte
+}
+type Config struct {
+	options                    *options.CompletedOptions
+	EtcdServer                 *etcd.Server
```
sttts, line 49: seeing a server struct here supports my comment https://github.com/kcp-dev/kcp/pull/1538/files#r928278498.

pkg/server/config.go
```diff
@@ -0,0 +1,418 @@
...
+	identityConfig             *rest.Config
+	resolveIdentities          func(ctx context.Context) error
+	genericConfig              *genericapiserver.Config
+	apisConfig                 *apis.Config
+	apiExtensionsConfig        *apiextensionsapiserver.Config
+	preHandlerChain            handlerChainMuxes
+	dynamicClusterClient       *dynamic.Cluster
+	apiextensionsClusterClient *apiextensionsclient.Cluster
+	kcpClusterClient           *kcpclient.Cluster
+	kubeClusterClient          *kubernetes.Cluster
```
sttts, line 59: this list needs some better grouping. A block of client related things would make sense. The sub-apiserver config should go to the top.

pkg/server/config.go
```diff
@@ -0,0 +1,418 @@
...
+type tokenAuth struct {
+	kcp       string
+	shard     string
+	shardHash []byte
+}
+type Config struct {
+	options                    *options.CompletedOptions
+	EtcdServer                 *etcd.Server
+	identityConfig             *rest.Config
+	resolveIdentities          func(ctx context.Context) error
```
sttts, line 51:
```suggestion
	// identityConfig is a rest config for kcp cluster clients that initialially only supports the apis.kcp.dev/v1alpha1
	// requests, and reject any other resources until `resolveIdentities` has been called successfully. 
	identityConfig             *rest.Config
	// resolveIdentities resolves the identities for kcp API groups and will eventually inject them into 
	// identityConfig. Call resolveIdentities during server startup until it succeeds.
	resolveIdentities          func(ctx context.Context) error
```

deitch:
This was superseded by #1609

PR #1542, opened 2022-07-15
server: add --shard-external-url

ncdc:
## Summary
kcp deployers optionally need to be able to set the URL external clients
use when connecting to kcp (e.g., via a front proxy). Add a new
--shard-external-url flag that is used as the default value for
ClusterWorkspaceShard.Spec.ExternalURL.

## Related issue(s)

Fixes #


csams (approved):
/lgtm

PR #1554, opened 2022-07-18
cli: take URL from Workspace also in absolute references

sttts:
Take the right URL from the workspace object for e.g. `kubectl ws root:example`.

Note: this does not contain a solution for the root workspace yet. Probably we need a similar solution to the `~` workspace where we fake a `Workspace` object.

### DIFF
--- a/pkg/cliplugins/workspace/plugin/kubeconfig.go
+++ b/pkg/cliplugins/workspace/plugin/kubeconfig.go
@@ -217,15 +217,24 @@ func (kc *KubeConfig) UseWorkspace(ctx context.Context, name string) (err error)
 			return fmt.Errorf("current URL %q does not point to cluster workspace", config.Host)
 		}
 
-		if strings.Contains(name, ":") && logicalcluster.New(name).HasPrefix(tenancyv1alpha1.RootCluster) {
+		if strings.Contains(name, ":") && cluster.HasPrefix(tenancyv1alpha1.RootCluster) {
+			// absolute logical cluster under root:
 			parentClusterName, workspaceName := logicalcluster.New(name).Split()
-			if _, err := kc.personalClient.Cluster(parentClusterName).TenancyV1beta1().Workspaces().Get(ctx, workspaceName, metav1.GetOptions{}); err != nil {
+			ws, err := kc.personalClient.Cluster(parentClusterName).TenancyV1beta1().Workspaces().Get(ctx, workspaceName, metav1.GetOptions{})
+			if err != nil {
 				return err
 			}
-		}
 
-		if strings.Contains(name, ":") || name == tenancyv1alpha1.RootCluster.String() {
-			// absolute logical cluster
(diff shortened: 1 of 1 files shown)


pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -217,15 +217,24 @@ func (kc *KubeConfig) UseWorkspace(ctx context.Context, name string) (err error)
 			return fmt.Errorf("current URL %q does not point to cluster workspace", config.Host)
 		}
 
-		if strings.Contains(name, ":") && logicalcluster.New(name).HasPrefix(tenancyv1alpha1.RootCluster) {
+		if strings.Contains(name, ":") && cluster.HasPrefix(currentClusterName) {
```
sttts, line 220: hmm, bad completion?

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -217,15 +217,24 @@ func (kc *KubeConfig) UseWorkspace(ctx context.Context, name string) (err error)
 			return fmt.Errorf("current URL %q does not point to cluster workspace", config.Host)
 		}
 
-		if strings.Contains(name, ":") && logicalcluster.New(name).HasPrefix(tenancyv1alpha1.RootCluster) {
+		if strings.Contains(name, ":") && cluster.HasPrefix(tenancyv1alpha1.RootCluster) {
+			// absolute logical cluster under root:
 			parentClusterName, workspaceName := logicalcluster.New(name).Split()
-			if _, err := kc.personalClient.Cluster(parentClusterName).TenancyV1beta1().Workspaces().Get(ctx, workspaceName, metav1.GetOptions{}); err != nil {
+			ws, err := kc.personalClient.Cluster(parentClusterName).TenancyV1beta1().Workspaces().Get(ctx, workspaceName, metav1.GetOptions{})
```
p0lyn0mial, line 223: ok, this is the case we hti today.

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -217,15 +217,24 @@ func (kc *KubeConfig) UseWorkspace(ctx context.Context, name string) (err error)
...
+			// intentionally do not check for readiness here
+
+			newServerHost = ws.Status.URL
+			workspaceType = &ws.Spec.Type
+		} else if strings.Contains(name, ":") {
+			// e.g. system:something
+			u.Path = path.Join(u.Path, logicalcluster.New(name).Path())
+			newServerHost = u.String()
+		} else if name == tenancyv1alpha1.RootCluster.String() {
+			// root workspace
```
p0lyn0mial, line 237: it looks exactly the same as the previous case

p0lyn0mial: clarified: in the future, the root case will require additional logic

PR #1558, opened 2022-07-18
Part 3: Make APIResource controller client calls scoped

varshaprasad96:
## Related issue(s)

Fixes #

### DIFF
--- a/cmd/cluster-controller/main.go
+++ b/cmd/cluster-controller/main.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"time"
 
+	kcpclienthelper "github.com/kcp-dev/apimachinery/pkg/client"
 	"github.com/spf13/pflag"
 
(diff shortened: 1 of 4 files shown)


cmd/cluster-controller/main.go
```diff
@@ -83,13 +84,14 @@ func main() {
 		fmt.Fprintln(os.Stderr, err)
 		os.Exit(1)
 	}
+	clusterConfig := kcpclienthelper.NewClusterConfig(config)
```
varshaprasad96, line 87: Creating a separate `clusterConfig` here since we need the unmodified config for informers/listers. This will be removed after informers are scoped. https://github.com/kcp-dev/kcp/compare/main...varshaprasad96:kcp:wrap-apiresource-controller?expand=1#diff-03df9c971de272e090fac0bbcdb3f0bca9fbccc2f168aafff1e3cdf7e33eb915R100

PR #1561, opened 2022-07-19
Documentation: Add Home, Homebucket, and Homeroot workspace info

pweil-:
Per https://kubernetes.slack.com/archives/C021U8WSAFK/p1658217760034109

### DIFF
--- a/docs/workspaces.md
+++ b/docs/workspaces.md
@@ -62,17 +62,63 @@ objects, e.g. like CRDs where each workspace can have its own set of CRDs instal
 
 ## User Home Workspaces
 
-User home workspaces are an optional feature of kcp. If enabled (through `--enable-home-workspaces`), there is a special virtual `Workspace` called `~` in the root workspace. It is used by `kubectl ws` (which similarly to Unix `cd` move the users to their home) to derive the full path to the user home workspace. This full path depends on the user name, the configured number (default 2), the length (default 2) of bucket workspaces and the root prefix (default `root:users`). With that a user `adam` is mapped to e.g. `root:users:a8:f1:adam`. The bucket names are deterministically derived from the user name (via some hash).
+User home workspaces are an optional feature of kcp. If enabled (through `--enable-home-workspaces`), there is a special 
+virtual `Workspace` called `~` in the root workspace. It is used by `kubectl ws` to derive the full path to the user 
+home workspace, similar to how Unix `cd ~` move the users to their home. 
 
-The user home workspace, if enabled, is logically pre-existing for every user, i.e. the user is given that impression. In reality, the home workspace is created on-demand on access, but this creation is not visible from outside.
+The full path for a user's home workspace has a number of parts: `<prefix>:(<bucket>)+<user-name>`. Buckets are used to 
+ensure that ~1000 sub-buckets or users exist in any bucket, for scaling reasons. The bucket names are deterministically 
+derived from the user name (via some hash). Example for user `adam` when using default configuration: 
+`root:users:a8:f1:adam`. 
 
-Only users of the configured home-creator-groups (default `system:authenticated`) will have a home workspace.
+User home workspaces are created on-demand when they are first accessed, but this is not visible to the user, allowing 
+the system to only incur the cost of these workspaces when they are needed. Only users of the configured 
+home-creator-groups (default `system:authenticated`) will have a home workspace.
 
 The following cluster workspace types are created internally to support User Home Workspaces:
 
 * `homeroot`: the workspace that will contain all the Home workspaces, spread accross buckets. - Can contain only Homebucket workspaces
 * `homebucket`: the type of workspaces that will contain a subset of all home workspaces. - Can contain either Homebucket (multi-level bucketing) or Home workspaces
 * `home`: the ClusterWorkspace of home workspaces - can contain any type of workspaces as children (especially universal workspaces)
+
+### Bucket configuration options
+
+The `kcp` administrator can configure:
+
+* `<prefix>`, which defaults to `root:users`
+* bucket depth, which defaults to 2
+* bucket name length, in characters, which defaults to 2
+
+The following outlines valid configuration options. With the default setup, ~5 users or ~700 sub-buckets will be in 
+any bucket. 
+
+> **NOTE**: DO NOT set the bucket size to be longer than 2, as this will adversely impact performance.
+
+User-names have `(26 * [(26 + 10 + 2) * 61] * 36 = 2169648)` permutations, and buckets are made up of lowercase-alpha 
+chars.  Invalid configurations break the scale limit in sub-buckets or users. Valid configurations should target 
+having not more than ~1000 sub-buckets per bucket and at least 5 users per bucket.
+
+**Valid Configurations**
+
+|length|depth|sub-buckets|users|
+|------|-----|-----------|-----|
+|1     |3    |26 * 1 = 26|2169648 / (26)^3 = 124 |
+|1     |4    |26 * 1 = 26|2169648 / (26)^4 = 5 |
+|2     |2    |26 * 26 = 676|2169648 / (26*26)^2 = 5 |
+
+**Invalid Configurations**
+
+These are examples of invalid configurations and are for illustrative purposes only. In nearly all cases, the default values
+will be sufficient.
+
+|length|depth|sub-buckets|users|
+|------|-----|-----------|-----|
+|1     |1    |26 * 1 = 26|2169648 / (26) = 83448 |
+|1     |2    |26 * 1 = 26|2169648 / (26)^2 = 3209 |
+|2     |1    |26 * 26 = 676|2169648 / (26*26) = 3209 |
+|2     |3    |26 * 26 = 676|2169648 / (26*26)^3 = .007 |
+|3     |1    |26 * 26 * 26 = 17576|2169648 / (26*26*26) = 124 |
+|3     |2    |26 * 26 * 26 = 17576|2169648 / (26*26*26)^2 = .007 |
 
 ## Organization Workspaces
 
--- a/docs/workspaces.md
+++ b/docs/workspaces.md
@@ -60,6 +60,16 @@ they have independent behaviour and no cluster workspace sees objects from other
 cluster workspaces. In contrast to namespace in Kubernetes, this includes non-namespaced
 objects, e.g. like CRDs where each workspace can have its own set of CRDs installed.
 
+## User Home Workspaces
+
+TODO: explaining the concept and related virtual existence and automatic creation 
+
+The following cluster workspace types are created internally to support User Home Workspaces:
+
+* `Homeroot`: the workspace that will contain all the Home workspaces, spread accross buckets. - Can contain only Homebucket workspaces
+* `Homebucket`: the type of workspaces that will contains a subset of all home workspaces. - Can contain either Homebucket (multi-level bucketing) or Home workspaces
+* `Home`: the ClusterWorkspace of home workspaces - can contain any type of workspaces as children (especially universal workspaces)
+
 ## Organization Workspaces
 
 Organization workspaces are ClusterWorkspaces of type `Organization`, defined in the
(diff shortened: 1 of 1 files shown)


docs/workspaces.md
```diff
@@ -60,6 +60,12 @@ they have independent behaviour and no cluster workspace sees objects from other
 cluster workspaces. In contrast to namespace in Kubernetes, this includes non-namespaced
 objects, e.g. like CRDs where each workspace can have its own set of CRDs installed.
 
+The following cluster workspace types are created internally to support User Home Workspaces:
+
+* `Home`: the ClusterWorkspace of home workspaces - can contain any type of workspaces as children (especially universal workspaces)
+* `Homebucket`: the type of workspaces that will contains a subset of all home workspaces. - Can contain either Homebucket (multi-level bucketing) or Home workspaces
+* `Homeroot`: the workspace that will contain all the Home workspaces, spread accross buckets. - Can contain only Homebucket workspaces
```
ncdc, line 67: I would reorder with Homeroot first, then Homebucket, then Home. WDYT?

davidfestal: I'm not against it.

davidfestal: Do we want to include this in a dedicated section about user Home workspaces, very explaining the concept and related virtual existence and automatic creation ?

docs/workspaces.md
```diff
@@ -60,6 +60,16 @@ they have independent behaviour and no cluster workspace sees objects from other
...
 objects, e.g. like CRDs where each workspace can have its own set of CRDs installed.
 
+## User Home Workspaces
+
+TODO: explaining the concept and related virtual existence and automatic creation 
+
+The following cluster workspace types are created internally to support User Home Workspaces:
+
+* `Homeroot`: the workspace that will contain all the Home workspaces, spread accross buckets. - Can contain only Homebucket workspaces
+* `Homebucket`: the type of workspaces that will contains a subset of all home workspaces. - Can contain either Homebucket (multi-level bucketing) or Home workspaces
```
sttts, line 70:
```suggestion
* `Homebucket`: the type of workspaces that will contain a subset of all home workspaces. - Can contain either Homebucket (multi-level bucketing) or Home workspaces
```

docs/workspaces.md
```diff
@@ -60,6 +60,16 @@ they have independent behaviour and no cluster workspace sees objects from other
 cluster workspaces. In contrast to namespace in Kubernetes, this includes non-namespaced
 objects, e.g. like CRDs where each workspace can have its own set of CRDs installed.
 
+## User Home Workspaces
+
+TODO: explaining the concept and related virtual existence and automatic creation
```
sttts, line 65: Maybe helpful:
```
		"enable-home-workspaces",                 // Enable the Home Workspaces feature (enabled by default). Home workspaces allow a personal home workspace to provisioned on first access per-user. A user is cluster-admin inside his personal Home workspace.
		"home-workspaces-creation-delay-seconds", // Delay, in seconds, before accessing the Home is retried after its automatic creation. This value is used when sending 'retry-after' responses to the Kubernetes client.
		"home-workspaces-bucket-levels",          // Number of levels of bucket workspaces when bucketing home workspaces
		"home-workspaces-bucket-size",            // Number of characters of bucket workspace names used when bucketing home workspaces
		"home-workspaces-home-creator-groups",    // Groups of users who can have their home workspace created automatically create when first accessing it.
		"home-workspaces-root-prefix",            // Logical cluster name of the workspace that will contains home workspaces for all workspaces.
```

sttts, line 65:
```suggestion
User home workspaces are an optional feature of kcp. If enabled (through `--enable-home-workspaces`), there is a special virtual `Workspace` called `~` in the root workspace. It is used by `kubectl ws` (which similarly to Unix `cd` move the users to their home) to derive the full path to the user home workspace. This full path depends on the user name, the configured number (default 2), the length (default 2) of bucket workspaces and the root prefix (default `root:users`). With that a user `adam` is mapped to e.g. `root:users:a8:f1:adam`. The bucket names are deterministically derived from the user name (via some hash).

The user home workspace, if enabled, is logically pre-existing for every user, i.e. the user is given that impression. In reality, the home workspace is created on-demand on access, but this creation is not visible from outside.

Only users of the configured home-creator-groups (default `system:authenticated`) will have a home workspace.
```

docs/workspaces.md
```diff
@@ -60,6 +60,16 @@ they have independent behaviour and no cluster workspace sees objects from other
 cluster workspaces. In contrast to namespace in Kubernetes, this includes non-namespaced
 objects, e.g. like CRDs where each workspace can have its own set of CRDs installed.
 
+## User Home Workspaces
+
+TODO: explaining the concept and related virtual existence and automatic creation 
+
+The following cluster workspace types are created internally to support User Home Workspaces:
+
+* `Homeroot`: the workspace that will contain all the Home workspaces, spread accross buckets. - Can contain only Homebucket workspaces
```
sttts, line 69:
```suggestion
* `homeroot`: the workspace that will contain all the Home workspaces, spread accross buckets. - Can contain only Homebucket workspaces
```

docs/workspaces.md
```diff
@@ -60,6 +60,16 @@ they have independent behaviour and no cluster workspace sees objects from other
...
 
+## User Home Workspaces
+
+TODO: explaining the concept and related virtual existence and automatic creation 
+
+The following cluster workspace types are created internally to support User Home Workspaces:
+
+* `homeroot`: the workspace that will contain all the Home workspaces, spread accross buckets. - Can contain only Homebucket workspaces
+* `Homebucket`: the type of workspaces that will contain a subset of all home workspaces. - Can contain either Homebucket (multi-level bucketing) or Home workspaces
+* `Home`: the ClusterWorkspace of home workspaces - can contain any type of workspaces as children (especially universal workspaces)
```
sttts, line 71:
```suggestion
* `homebucket`: the type of workspaces that will contain a subset of all home workspaces. - Can contain either Homebucket (multi-level bucketing) or Home workspaces
* `home`: the ClusterWorkspace of home workspaces - can contain any type of workspaces as children (especially universal workspaces)
```

pweil-:
Suggestions accepted, thanks!

pweil-:
@davidfestal assigning you for lgtm

davidfestal (approved):
Very nice ! Thanks !

docs/workspaces.md
```diff
@@ -60,6 +60,20 @@ they have independent behaviour and no cluster workspace sees objects from other
 cluster workspaces. In contrast to namespace in Kubernetes, this includes non-namespaced
 objects, e.g. like CRDs where each workspace can have its own set of CRDs installed.
 
+## User Home Workspaces
+
+User home workspaces are an optional feature of kcp. If enabled (through `--enable-home-workspaces`), there is a special virtual `Workspace` called `~` in the root workspace. It is used by `kubectl ws` (which similarly to Unix `cd` move the users to their home) to derive the full path to the user home workspace. This full path depends on the user name, the configured number (default 2), the length (default 2) of bucket workspaces and the root prefix (default `root:users`). With that a user `adam` is mapped to e.g. `root:users:a8:f1:adam`. The bucket names are deterministically derived from the user name (via some hash).
```
stevekuznetsov, line 65:
```suggestion
User home workspaces are an optional feature of kcp. If enabled (through `--enable-home-workspaces`), there is a special virtual `Workspace` called `~` in the root workspace. It is used by `kubectl ws` to derive the full path to the user home workspace, similar to how Unix `cd ~` moves a user to their home directory. This full path depends on the user name, the configured number (default 2), the length (default 2) of bucket workspaces and the root prefix (default `root:users`). With that a user `adam` is mapped to e.g. `root:users:a8:f1:adam`. The bucket names are deterministically derived from the user name (via some hash).
```

stevekuznetsov:
```
The full path for a user's home workspace has a number of parts: `<prefix>:(<bucket>:)+<user-name>`. Buckets are used to ensure that ~1000 sub-buckets or users exist in any bucket, for scaling reasons. The `kcp` administrator can configure:

- the `<prefix>`, which defaults to `root:users`
- the bucket depth, which defaults to 2
- the bucket length, which defaults to 2

Note: with the current setup, ~60 users or ~2000 sub-buckets will be in any bucket. DO NOT set the bucket size to be longer than 2, as this will adversely impact performance.
```

stevekuznetsov: cc @sttts @davidfestal 

IMHO - these options are pretty esoteric and an enormous foot-gun. No sane admin would ever choose something other than 2,2 - let's remove the options. This is silly.

docs/workspaces.md
```diff
@@ -60,6 +60,20 @@ they have independent behaviour and no cluster workspace sees objects from other
...
+
+The user home workspace, if enabled, is logically pre-existing for every user, i.e. the user is given that impression. In reality, the home workspace is created on-demand on access, but this creation is not visible from outside.
```
stevekuznetsov, line 67: Who does this implementation detail impact? Lets' entirely remove the line.

docs/workspaces.md
```diff
@@ -60,6 +60,20 @@ they have independent behaviour and no cluster workspace sees objects from other
...
+
+Only users of the configured home-creator-groups (default `system:authenticated`) will have a home workspace.
```
stevekuznetsov, line 69: What is a `home-creator-group`? How do I set it?

docs/workspaces.md
stevekuznetsov: e.g. choosing bucket length > 2 leads to a performance issue. So, we have the following options:

User-names have (26 * [(26 + 10 + 2) * 61] * 36 = 2169648) permutations, and buckets are made up of lowercase-alpha chars, so:

- length = 1, depth = 2 -> 26 sub-buckets per bucket; 2169648 / (26) ^2 = 3200 users per bucket
- length = 1, depth = 3 -> 26 * 38 = 988 sub-buckets per bucket; 2169648 / (26 * 38)^3 = 124 users per bucket
- length = 2, depth = 1 -> 26 * 38 = 988 sub-buckets per bucket; 2169648 / (26 * 38)^2 = 2196 users per bucket
- length = 2, depth = 2 -> 26 * 38 = 988 sub-buckets per bucket; 2169648 / (26 * 38)^2 = 3 users per bucket

Every single other option is nonsense. We should not have more than ~1000 sub-buckets per bucket, and we don't want fewer than 3 users per bucket .... 

IMHO we hard-code length = 1, depth = 3, and just go with it. Exposing these options is an illusion of choice.

docs/workspaces.md
sttts: it impacts admins and ppl trying to understand the system. They docs are not user-docs, not even admin docs.

docs/workspaces.md
pweil-: If this doc accurately depicts the current situation and if we agree that removing the option is probably the right thing (or at least worth discussing more) should we call this good and create an issue for proposing the default and removal of config option and include this breakdown?

pweil-: (to be clear: I'm happy to create that issue, just give me the 👍 )

stevekuznetsov: @pweil- I think we should amend the comments in the doc to something like [this](https://github.com/kcp-dev/kcp/pull/1561#discussion_r926886232) to be more clear

docs/workspaces.md
stevekuznetsov: Hm, docs without a target reader are kind of a smell. If we keep the line, @pweil- can we please edit it to something like:

> Home user workspaces are created on-demand when they are first accessed, but this is not visible to the user, allowing the system to only incur the cost of these workspaces when they are needed.

docs/workspaces.md
sttts: For the follow-up: I see value in the flags, but would clearly document the sensible cases in the flag docs, and add the math here (follow-up).

docs/workspaces.md
pweil-: done

docs/workspaces.md
pweil-: follow up: https://github.com/kcp-dev/kcp/issues/1599

I've revamped the whole thing to include better organization, config info, and the example tables.  PTAL 👍

docs/workspaces.md
pweil-: @davidfestal need your advice here.

stevekuznetsov (commented):
LGTM modulo the `home-creator-groups` bit - I think we can merge as-is but would like to see some comment on how someone changes `home-creator-groups` in the future to make that actionable for an admin.

PR #1566, opened 2022-07-19
syncer/namespace-locator: Rename path to workspace

jmprusi:
This PR renames the "path" field to "workspace". 

~~*This is a breaking change that will trigger a resync of all the resources*~~ It's compatible with old namespace locators.


sttts:
/hold

To make it backwards compatible.

PR #1570, opened 2022-07-20
e2e/clusterworkspacedeletion: use correct root shard system:master client for deletion checks

sttts:
The deletion checks have to skip authorization because authz would lead to 403 on deleted workspaces, i.e. you cannot look into the logical cluster then. Hence, we need a `system:master` client. We have that wired for the root shard.


test/e2e/watchcache/watchcache_enabled_test.go
```diff
@@ -208,7 +208,7 @@ func TestWatchCacheEnabledForBuiltinTypes(t *testing.T) {
 		}
 	}
 
-	totalCacheHits, secretsCacheHit := collectCacheHitsFor(ctx, t, server.RootShardConfig(t), "/core/secrets")
+	totalCacheHits, secretsCacheHit := collectCacheHitsFor(ctx, t, server.RootShardSystemMasterBaseConfig(t), "/core/secrets")
```
p0lyn0mial, line 211: mhm, wondering why the `kcp-admin` doesn't work here

p0lyn0mial: shouldn't `kcp-admin` have access to the `/metrics` endpoint?

PR #1571, opened 2022-07-20
add exports to synctarget API for location workspace

qiujian16:

## Related issue(s)

Fixes #


pkg/apis/workload/v1alpha1/synctarget_types.go
```diff
@@ -83,7 +90,7 @@ type SyncTargetStatus struct {
 	Conditions conditionsv1alpha1.Conditions `json:"conditions,omitempty"`
 
 	// +optional
-	SyncedResources []string `json:"syncedResources,omitempty"`
+	SyncedResources []ResourceToSync `json:"syncedResources,omitempty"`
```
sttts, line 93: add go doc, including who owns (= updates) this field. I.e. this is not updated by the syncer, but by the server (= controllers).

pkg/apis/workload/v1alpha1/synctarget_types.go
```diff
@@ -94,6 +101,44 @@ type SyncTargetStatus struct {
...
+	// group is the group of the API. Empty string for the core API group.
+	//
+	// +required
+	Group string `json:"group"`
+
+	// resource is the resource of the API.
+	//
+	// kubebuilder:validation:MinLength=1
+	// +required
+	Resource string `json:"resource"`
```
sttts, line 114: embed apisv1alpha1.GroupResource instead. It also has nice OpenAPI patterns.

pkg/apis/workload/v1alpha1/synctarget_types.go
```diff
@@ -94,6 +101,44 @@ type SyncTargetStatus struct {
...
+
+	// Versions is the versions of the resource.
+	// +optional
+	Versions []string `json:"versions"`
```
sttts, line 118: Do we know what this actually means? Does it mean this list is ordered, and the syncer should take the first version available on the cluster and use that for syncing?

Please update go doc.

qiujian16: change to `Version string`, I think we need to get state for each GVR from syncer.

sttts: but we have this problem of skew. Old kube might only have beta version of some API, newer kube has v1 and beta. Newest kube has only v1. We need a plan here how to cope with that. cc @davidfestal

sttts: and this poses the question who will do the conversion. 

Side-note: @ncdc, @stevekuznetsov and I talked about CEL conversions yesterday for APIResourceSchemas. Video in slack.

pkg/apis/workload/v1alpha1/synctarget_types.go
```diff
@@ -65,6 +66,12 @@ type SyncTargetSpec struct {
 	// will be unassigned from the cluster.
 	// By default, workloads scheduled to the cluster are not evicted.
 	EvictAfter *metav1.Time `json:"evictAfter,omitempty"`
+
+	// Exports defines a set of APIExports supposed to be supported by this SyncTarget. The SyncTarget
+	// will be selected to deploy the workload only when the resource schema on the SyncTarget is compatible
+	// with the resource schema included in the exports.
+	// If it is not set, the kubernetes export in the same workspace will be used by default.
+	Exports []apisv1alpha1.ExportReference `json:"exports,omitempty"`
```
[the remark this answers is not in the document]
davidfestal: I like `supportedAPIExports`

pkg/apis/workload/v1alpha1/synctarget_types.go
sttts: Proposal:
```golang
// versions are the resource versions the syncer can choose to sync depending on
// availability on the downstream cluster. Conversion to the storage version, if necessary, 
// will be done on the kcp side. The versions are ordered according by precedence, i.e. the
// first version is preferred.
//
// +kubebuilder .... MinItems=1
// +required
Versions []string `json:"versions"`
```

pkg/apis/workload/v1alpha1/synctarget_types.go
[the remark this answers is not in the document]
jmprusi: Same: `supportedAPIExports`

pkg/apis/workload/v1alpha1/synctarget_types.go
```diff
@@ -65,6 +66,33 @@ type SyncTargetSpec struct {
...
 	// By default, workloads scheduled to the cluster are not evicted.
 	EvictAfter *metav1.Time `json:"evictAfter,omitempty"`
+
+	// SupportedAPIExports defines a set of APIExports supposed to be supported by this SyncTarget. The SyncTarget
+	// will be selected to deploy the workload only when the resource schema on the SyncTarget is compatible
+	// with the resource schema included in the exports.
+	// If it is not set, the kubernetes export in the same workspace will be used by default.
+	SupportedAPIExports []apisv1alpha1.ExportReference `json:"supportedAPIExports,omitempty"`
+
+	// Cells represent a set of various cell a sync target belongs to. Each cell has a unique type and ID. SyncTargets
```
sttts, line 76:
```suggestion
	// Cells represent a set of various cells a sync target belongs to. Each cell has a unique type and ID. SyncTargets
```

pkg/apis/workload/v1alpha1/synctarget_types.go
```diff
@@ -65,6 +66,33 @@ type SyncTargetSpec struct {
...
+	// in the same cell runs as they are in the same physical cluster in a certain apsect. i.e. if SyncTargets are in
```
sttts, line 77:
```suggestion
	// in the same set of cells runs as they are in the same physical cluster in a certain apsect. i.e. if SyncTargets are in
```

pkg/apis/workload/v1alpha1/synctarget_types.go
```diff
@@ -65,6 +66,33 @@ type SyncTargetSpec struct {
...
+	// a network type cell with the same ID, the workloads on these SyncTargets can connect to each other just like in
+	// one physical cluster. Cells should be added and updated by different cell provider (i.e. a network provider)
```
sttts, line 79:
```suggestion
	// one physical cluster. Cells should be added and updated by different cell providers (i.e. a network provider)
```

pkg/apis/workload/v1alpha1/synctarget_types.go
```diff
@@ -65,6 +66,33 @@ type SyncTargetSpec struct {
...
+	Cells []Cell `json:"cells,omitempty"`
```
sttts, line 80: list-type is map, with type as the key

pkg/apis/workload/v1alpha1/synctarget_types.go
```diff
@@ -94,6 +108,43 @@ type SyncTargetStatus struct {
 	VirtualWorkspaces []VirtualWorkspace `json:"virtualWorkspaces,omitempty"`
 }
 
+type ResourceToSync struct {
+	apisv1alpha1.GroupResource `json:","`
+
+	// versions are the resource versions the syncer can choose to sync depending on
+	// availability on the downstream cluster. Conversion to the storage version, if necessary,
+	// will be done on the kcp side. The versions are ordered according by precedence and the
```
sttts, line 116:
```suggestion
	// will be done on the kcp side. The versions are ordered by precedence and the
```

pkg/apis/workload/v1alpha1/synctarget_types.go
```diff
@@ -94,6 +108,43 @@ type SyncTargetStatus struct {
...
+	// first version compatible is preferred by syncer.
+	// +kubebuilder:validation:MinItems=1
+	// +required
+	// +kubebuilder:Required
+	Versions []string `json:"versions"`
+
+	// IdentityHash is the identity for a given APIExport that the APIResourceSchema belongs to.
```
sttts, line 123:
```suggestion
	// identityHash is the identity for a given APIExport that the APIResourceSchema belongs to.
```

pkg/apis/workload/v1alpha1/synctarget_types.go
```diff
@@ -94,6 +108,43 @@ type SyncTargetStatus struct {
...
+	// The hash can be found on APIExport and APIResourceSchema's status.
+	// It will be empty for core types.
+	// +optional
+	IdentityHash string `json:"identityHash"`
+
+	// State indicate whether the resources schema is compatible to the SyncTarget. It must be updated
```
sttts, line 129:
```suggestion
	// state indicate whether the resources schema is compatible to the SyncTarget. It must be updated
```

pkg/apis/workload/v1alpha1/synctarget_types.go
```diff
@@ -65,6 +66,17 @@ type SyncTargetSpec struct {
 	// will be unassigned from the cluster.
 	// By default, workloads scheduled to the cluster are not evicted.
 	EvictAfter *metav1.Time `json:"evictAfter,omitempty"`
+
+	// SupportedAPIExports defines a set of APIExports supposed to be supported by this SyncTarget. The SyncTarget
+	// will be selected to deploy the workload only when the resource schema on the SyncTarget is compatible
+	// with the resource schema included in the exports.
+	// If it is not set, the kubernetes export in the same workspace will be used by default.
+	SupportedAPIExports []apisv1alpha1.ExportReference `json:"supportedAPIExports,omitempty"`
```
sttts, line 74: How should we do this defaulting? We could default via OpenAPI here.

PR #1572, opened 2022-07-20
wire TemporaryRootShardKcpSharedInformerFactory

p0lyn0mial:
## Summary

wires TemporaryRootShardKcpSharedInformerFactory, essentially when c.Options.Extra.RootShardKubeconfigFile is set then:
- we start a new SharedInformerFactory for the root shard
- we create a KcpClusterClient to the root shard
- we registerer a new shard in the root namespace

   
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -38,6 +38,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/clusters"
 	"k8s.io/kubernetes/pkg/genericcontrolplane"
 	"k8s.io/kubernetes/pkg/genericcontrolplane/aggregator"
@@ -86,6 +87,7 @@ type ExtraConfig struct {
 	KubeClusterClient          kubernetes.ClusterInterface
 	ApiExtensionsClusterClient apiextensionsclient.ClusterInterface
 	KcpClusterClient           kcpclient.ClusterInterface
+	RootShardKcpClusterClient  kcpclient.ClusterInterface
 
 	// misc
 	preHandlerChainMux   *handlerChainMuxes
@@ -99,6 +101,9 @@ type ExtraConfig struct {
 
 	// TODO(p0lyn0mial):  get rid of TemporaryRootShardKcpSharedInformerFactory, in the future
 	//                    we should have multi-shard aware informers
+	//
+	// TODO(p0lyn0mial): wire it to the root shard, this will be needed to get bindings,
+	//                   eventually it will be replaced by replication
 	//
 	// TemporaryRootShardKcpSharedInformerFactory bring data from the root shard
 	TemporaryRootShardKcpSharedInformerFactory kcpexternalversions.SharedInformerFactory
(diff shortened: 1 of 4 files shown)


pkg/server/server.go
```diff
@@ -28,6 +28,7 @@ import (
 	apiextensionsapiserver "k8s.io/apiextensions-apiserver/pkg/apiserver"
```
sttts, line 28: needs better commit name

pkg/server/server.go
```diff
@@ -121,6 +122,9 @@ type ExtraConfig struct {
 	// resource identities for the root shard into the rest.Config used by the client.
 	// Only after it succeeds, the clients can wildcard-list/watch most kcp resources.
 	ResolveRootKcpShardIdentities func(ctx context.Context) error
+
+	// RootShardKcpClusterClient holds a cluster-wide client to the root shard
+	RootShardKcpClusterClient *kcpclient.Cluster
```
sttts, line 127: move up to the clients

pkg/server/server.go
```diff
@@ -115,6 +116,11 @@ type ExtraConfig struct {
 	//
 	// TemporaryRootShardKcpSharedInformerFactory bring data from the root shard
 	TemporaryRootShardKcpSharedInformerFactory kcpexternalversions.SharedInformerFactory
+
+	// ResolveRootKcpShardIdentities is to be called on server start until it succeeds. It injects the kcp
+	// resource identities for the root shard into the rest.Config used by the client.
+	// Only after it succeeds, the clients can wildcard-list/watch most kcp resources.
+	ResolveRootKcpShardIdentities func(ctx context.Context) error
```
sttts, line 124: we have that at the top already

pkg/server/server.go
```diff
@@ -177,10 +183,42 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to load the kubeconfig from: %s, for the root shard, err: %w", c.Options.Extra.RootShardKubeconfigFile, err)
+		}
+		nonIdentityKcpClusterConfig = nonIdentityRootKcpShardSystemAdminConfig                                        // assign to nonIdentityKcpClusterConfig to the root shard
+		nonIdentityRootKcpShardClient, err := kcpclient.NewClusterForConfig(nonIdentityRootKcpShardSystemAdminConfig) // can only used for wildcard requests of apis.kcp.dev
+		if err != nil {
+			return nil, err
+		}
+		var rootKcpShardIdentityConfig *rest.Config
+		rootKcpShardIdentityConfig, c.ResolveRootKcpShardIdentities = boostrap.NewConfigWithWildcardIdentities(nonIdentityRootKcpShardSystemAdminConfig, boostrap.KcpRootGroupExportNames, boostrap.KcpRootGroupResourceExportNames, nonIdentityRootKcpShardClient.Cluster(tenancyv1alpha1.RootCluster))
```
sttts, line 207: I think this is duplicated now

pkg/server/server.go
p0lyn0mial: this is for the root shard only, resolved in a different place. I think we need both.

pkg/server/server.go
```diff
@@ -89,7 +91,13 @@ type ExtraConfig struct {
...
 	// the clients can wildcard-list/watch most kcp resources.
 	resolveIdentities func(ctx context.Context) error
-	identityConfig    *rest.Config
+
+	// ResolveRootKcpShardIdentities is to be called on server start until it succeeds. It injects the kcp
+	// resource identities for the root shard into the rest.Config used by the client.
+	// Only after it succeeds, the clients can wildcard-list/watch most kcp resources.
+	ResolveRootKcpShardIdentities func(ctx context.Context) error
+
+	identityConfig *rest.Config
```
sttts, line 100: am still confused about the new fields. The existing should do exactly that.

p0lyn0mial: ah, I think I know what you mean :)

p0lyn0mial: I think this is what we need https://github.com/kcp-dev/kcp/pull/1620

sttts: can you wire it here?

pkg/server/server.go
```diff
@@ -177,10 +186,41 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to load the kubeconfig from: %s, for the root shard, err: %w", c.Options.Extra.RootShardKubeconfigFile, err)
+		}
+		nonIdentityKcpClusterConfig = nonIdentityRootKcpShardSystemAdminConfig                                        // assign to nonIdentityKcpClusterConfig to the root shard
+		nonIdentityRootKcpShardClient, err := kcpclient.NewClusterForConfig(nonIdentityRootKcpShardSystemAdminConfig) // can only used for wildcard requests of apis.kcp.dev
+		if err != nil {
+			return nil, err
+		}
+		var rootKcpShardIdentityConfig *rest.Config
+		rootKcpShardIdentityConfig, c.ResolveRootKcpShardIdentities = boostrap.NewConfigWithWildcardIdentities(nonIdentityRootKcpShardSystemAdminConfig, boostrap.KcpRootGroupExportNames, boostrap.KcpRootGroupResourceExportNames, nonIdentityRootKcpShardClient.Cluster(tenancyv1alpha1.RootCluster))
```
sttts, line 206: I am lost in this wiring. Don't we have these two cases:
1. *we are the root shard*. Then we need the old, staged flow where we start the local informer only for apis.kcp.dev first, then pick the identities from the APIExports, inject them into the roundtripper and then start the rest.
2. *we are NOT the root shard*. Then we can run `resolveRootKcpShardIdentities` here in-place very early before anything of the code further down runs, i.e. before any of the informers have been started.

pkg/server/server.go
p0lyn0mial: done

pkg/server/server.go
```diff
@@ -177,14 +181,52 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
...
+			kcpexternalversions.WithExtraClusterScopedIndexers(indexers.ClusterScoped()),
+			kcpexternalversions.WithExtraNamespaceScopedIndexers(indexers.NamespaceScoped()),
+		)
+	} else {
+		// create an empty non-functional factory so that code that uses it but doesn't need it, doesn't have to check against the nil value
+		c.TemporaryRootShardKcpSharedInformerFactory = kcpexternalversions.NewSharedInformerFactory(nil, resyncPeriod)
+	}
+
+	if rootKcpShardIdentityRoundTripper != nil {
+		c.identityConfig = rest.CopyConfig(c.GenericConfig.LoopbackClientConfig)
```
p0lyn0mial, line 220: we could move this inside the if clause in L190

pkg/server/server.go
```diff
@@ -177,14 +181,52 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
...
+		c.identityConfig.Wrap(rootKcpShardIdentityRoundTripper)
+	} else {
+		// The informers here are not  used before the informers are actually started (i.e. no race).
```
p0lyn0mial, line 223: and this into the else clause in L214

pkg/server/config.go
```diff
@@ -164,14 +167,47 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
...
+		nonIdentityRootKcpShardSystemAdminConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(&clientcmd.ClientConfigLoadingRules{ExplicitPath: c.Options.Extra.RootShardKubeconfigFile}, &clientcmd.ConfigOverrides{CurrentContext: "system:admin"}).ClientConfig()
+		if err != nil {
+			return nil, fmt.Errorf("failed to load the kubeconfig from: %s, for the root shard, err: %w", c.Options.Extra.RootShardKubeconfigFile, err)
+		}
+		nonIdentityRootKcpShardClient, err := kcpclient.NewClusterForConfig(nonIdentityRootKcpShardSystemAdminConfig) // can only used for wildcard requests of apis.kcp.dev
+		if err != nil {
+			return nil, err
+		}
+		var rootKcpShardIdentityRoundTripper func(rt http.RoundTripper) http.RoundTripper
+		rootKcpShardIdentityRoundTripper, c.resolveIdentities = boostrap.NewWildcardIdentitiesWrappingRoundTripper(boostrap.KcpRootGroupExportNames, boostrap.KcpRootGroupResourceExportNames, nonIdentityRootKcpShardClient.Cluster(tenancyv1alpha1.RootCluster))
```
p0lyn0mial, line 184: I don't want to resolve it here, config is for setting up configuration

pkg/server/server.go
```diff
@@ -172,24 +173,76 @@ func (s *Server) Run(ctx context.Context) error {
 				return nil // don't klog.Fatal. This only happens when context is cancelled.
 			}
 			klog.Infof("Bootstrapped root workspace phase 0")
-		}
```
p0lyn0mial, line 176: now the flow is:
1. common
2. root
3. non-root
4. common
5. root (moving 4 and 5 breaks the server, not sure it can be done in any other way)

pkg/server/config.go
```diff
@@ -86,6 +87,8 @@ type ExtraConfig struct {
 	KubeClusterClient          kubernetes.ClusterInterface
 	ApiExtensionsClusterClient apiextensionsclient.ClusterInterface
 	KcpClusterClient           kcpclient.ClusterInterface
+	// RootShardKcpClusterClient holds a cluster-wide client to the root shard
+	RootShardKcpClusterClient *kcpclient.Cluster
```
sttts, line 91: use ClusterInterface

pkg/server/config.go
```diff
@@ -86,6 +87,8 @@ type ExtraConfig struct {
 	KubeClusterClient          kubernetes.ClusterInterface
 	ApiExtensionsClusterClient apiextensionsclient.ClusterInterface
 	KcpClusterClient           kcpclient.ClusterInterface
+	// RootShardKcpClusterClient holds a cluster-wide client to the root shard
```
sttts, line 90: no need for the comment. Doesn't add anything :)

pkg/server/config.go
sttts: that's an argument indeed

pkg/server/config.go
```diff
@@ -164,14 +167,47 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
...
+		}
+		nonIdentityRootKcpShardClient, err := kcpclient.NewClusterForConfig(nonIdentityRootKcpShardSystemAdminConfig) // can only used for wildcard requests of apis.kcp.dev
+		if err != nil {
+			return nil, err
+		}
+		var rootKcpShardIdentityRoundTripper func(rt http.RoundTripper) http.RoundTripper
+		rootKcpShardIdentityRoundTripper, c.resolveIdentities = boostrap.NewWildcardIdentitiesWrappingRoundTripper(boostrap.KcpRootGroupExportNames, boostrap.KcpRootGroupResourceExportNames, nonIdentityRootKcpShardClient.Cluster(tenancyv1alpha1.RootCluster))
+		rootKcpShardIdentityConfig := rest.CopyConfig(nonIdentityRootKcpShardSystemAdminConfig)
+		rootKcpShardIdentityConfig.Wrap(rootKcpShardIdentityRoundTripper)
+		c.RootShardKcpClusterClient, err = kcpclient.NewClusterForConfig(rootKcpShardIdentityConfig) // this is now generic to be used for all kcp API groups
```
sttts, line 187: comment is not correct. The client can be used after resolving via `c.resolveIdentities`. Would just drop the comment.

pkg/server/config.go
```diff
@@ -164,14 +167,47 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
...
+		}
+		c.TemporaryRootShardKcpSharedInformerFactory = kcpexternalversions.NewSharedInformerFactoryWithOptions(
+			c.RootShardKcpClusterClient.Cluster(logicalcluster.Wildcard),
+			resyncPeriod,
+			kcpexternalversions.WithExtraClusterScopedIndexers(indexers.ClusterScoped()),
+			kcpexternalversions.WithExtraNamespaceScopedIndexers(indexers.NamespaceScoped()),
+		)
+
+		c.identityConfig = rest.CopyConfig(c.GenericConfig.LoopbackClientConfig)
+		c.identityConfig.Wrap(rootKcpShardIdentityRoundTripper)
```
sttts, line 199: `rootKcpShardIdentityRoundTripper` is not a good name. Would drop the `root`. It does not really have anything todo with the root shard other than the source of the identities is the root.

pkg/server/config.go
```diff
@@ -164,14 +167,47 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
...
+	} else {
+		// create an empty non-functional factory so that code that uses it but doesn't need it, doesn't have to check against the nil value
+		c.TemporaryRootShardKcpSharedInformerFactory = kcpexternalversions.NewSharedInformerFactory(nil, resyncPeriod)
```
sttts, line 202: TODO: wire it to the other shard

pkg/server/config.go
```diff
@@ -164,14 +167,47 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
...
+
+		// The informers here are not  used before the informers are actually started (i.e. no race).
```
sttts, line 204:
```suggestion
		// The informers here are not used before the informers are actually started (i.e. no race).
```

PR #1585, opened 2022-07-21
clusterworkspace scheduling: default to assigning to the root shard

p0lyn0mial:
## Summary
if no specific shard was required, we are going to schedule the given ws onto the root shard.
this step is temporary until working with multi-shard env works,
until then we need to assign ws to the root shard otherwise all e2e test will break.
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_reconcile_scheduling.go
+++ b/pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_reconcile_scheduling.go
@@ -87,11 +87,33 @@ func (r *schedulingReconciler) reconcile(ctx context.Context, workspace *tenancy
 			}
 
 			if len(shards) == 0 {
-				// find a shard for this workspace, randomly
 				var err error
 				shards, err = r.listShards(selector)
 				if err != nil {
 					return reconcileStatusStopAndRequeue, err
+				}
+
+				// if no specific shard was required,
+				// we are going to schedule the given ws onto the root shard
+				// this step is temporary until working with multi-shard env works
+				// until then we need to assign ws to the root shard otherwise all e2e test will break
+				//
+				// note if there are no shards just let it run, at the end, we set a proper condition.
+				if len(shards) > 0 && (workspace.Spec.Shard == nil || workspace.Spec.Shard.Selector == nil) {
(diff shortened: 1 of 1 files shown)


pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_reconcile_scheduling.go
```diff
@@ -87,12 +87,32 @@ func (r *schedulingReconciler) reconcile(ctx context.Context, workspace *tenancy
 			}
 
 			if len(shards) == 0 {
-				// find a shard for this workspace, randomly
+				// no specific shard was required,
+				// we are going to schedule the given ws onto the root shard
+				// this step is temporary until working with multi-shard env works
+				// until then we need to assign ws to the root shard otherwise all e2e test will break
```
sttts, line 93: would move the comment down to line 100

pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_reconcile_scheduling.go
```diff
@@ -87,12 +87,32 @@ func (r *schedulingReconciler) reconcile(ctx context.Context, workspace *tenancy
...
+
+				if len(shards) > 0 {
+					// trim the list to contain only the "root" shard so that we always schedule onto it
+					for _, shard := range shards {
+						if shard.Name == "root" {
+							shards = []*tenancyv1alpha1.ClusterWorkspaceShard{shard}
+							break
+						}
+					}
+					if len(shards) == 0 {
```
sttts, line 108: this case surprises me. Do we need it?

p0lyn0mial: I think so, in case there are no shards we fall back to:

https://github.com/kcp-dev/kcp/blob/74a71ccccd2d40d0221c8d17a22495f0b54cf05f/pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_reconcile_scheduling.go#L152

sttts: does this need a change after our Slack discussion?

PR #1586, opened 2022-07-21
docs: tweaks to the Location concept description

markmc:
The "operation concerns" bit was clarified elsewhere by Stefan. 

Suggesting dropping the "kube conformance" bit because it probably needs further clarification so that it doesn't conflict with 
https://github.com/kcp-dev/kcp/blob/main/FAQ.md#will-kcp-be-able-to-pass-the-k8s-conformance-tests-in-cncf-conformance-suites


sttts:
Thanks for the clarification of the docs.

/lgtm
/approve

PR #1587, opened 2022-07-21
bootstrap the root workspace only on the root shard

p0lyn0mial:
## Summary
the root workspace is only available on the root shard, this pr makes sure it will be only created on it.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -417,17 +417,19 @@ func (s *Server) Run(ctx context.Context) error {
 
 		klog.Infof("Finished starting APIExport and APIBinding informers")
 
-		// bootstrap root workspace phase 0, no APIBinding resources yet
-		if err := configrootphase0.Bootstrap(goContext(ctx),
-			kcpClusterClient.Cluster(tenancyv1alpha1.RootCluster),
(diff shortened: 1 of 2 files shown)


pkg/server/server.go
```diff
@@ -417,18 +417,20 @@ func (s *Server) Run(ctx context.Context) error {
...
 
-		// bootstrap root workspace phase 0, no APIBinding resources yet
-		if err := configrootphase0.Bootstrap(goContext(ctx),
-			kcpClusterClient.Cluster(tenancyv1alpha1.RootCluster),
-			apiextensionsClusterClient.Cluster(tenancyv1alpha1.RootCluster).Discovery(),
-			dynamicClusterClient.Cluster(tenancyv1alpha1.RootCluster),
-		); err != nil {
-			// nolint:nilerr
-			return nil // don't klog.Fatal. This only happens when context is cancelled.
+		if s.options.Extra.ShardName == tenancyv1alpha1.RootCluster.String() {
```
sttts, line 420: nit: should we introduce a RootShard constant? This is not really the cluster.

p0lyn0mial: why not

p0lyn0mial: done.

PR #1602, opened 2022-07-22
0.6: ddsif: replace discovery polling

sttts:
Backport of #1473.

Replace the polling mechanism that wouldn't scale as the number of
ClusterWorkspaces increased with one that is based on a fixed-set of
built-in types + a dynamic set of informers driven by all the unique
group-resources across all the CRDs in the system.


sttts:
Stop the bleeding. Manually merging.

PR #1609, opened 2022-07-24
server: restructure with options/config/server pattern

sttts:
Split up `Server.Run` into the usual `Config`, `completeConfig`, `Server`, `Server.Run` steps for modularity and easier customization and use in different binaries.


pkg/admission/initializers/interfaces.go
```diff
@@ -32,13 +32,13 @@ type WantsKcpInformers interface {
...
 type WantsKubeClusterClient interface {
-	SetKubeClusterClient(kubeClusterClient *kubernetes.Cluster)
+	SetKubeClusterClient(kubeClusterClient kubernetes.ClusterInterface)
 }
 
 // WantsKcpClusterClient interface should be implemented by admission plugins
 // that want to have a kcp cluster client injected.
 type WantsKcpClusterClient interface {
-	SetKcpClusterClient(kubeClusterClient *kcpclientset.Cluster)
+	SetKcpClusterClient(kubeClusterClient kcpclientset.ClusterInterface)
```
stevekuznetsov, line 41: This seems like an unrelated change we could merge first/separately?

pkg/cliplugins/workload/plugin/syncer.yaml
```diff
@@ -13,6 +13,15 @@ metadata:
   name: {{.ServiceAccount}}
   namespace: {{.Namespace}}
 ---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{.ServiceAccount}}-token
+  namespace: {{.Namespace}}
+  annotations:
+    kubernetes.io/service-account.name: {{.ServiceAccount}}
```
stevekuznetsov, line 22: I still do not understand this whatsoever - this is making a token on the pcluster, right? We specifically don't want it to be an SA token, or annotated, right? Since if it is, the downstream controller deletes it?

pkg/embeddedetcd/config.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The KCP Authors.
+Copyright 2022 The KCP Authors.
```
stevekuznetsov, line 2: Do we have to do this?

pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_controller.go
```diff
@@ -53,20 +53,20 @@ const (
 func NewController(
 	kcpClusterClient kcpclient.Interface,
 	workspaceInformer tenancyinformer.ClusterWorkspaceInformer,
-	rootWorkspaceShardInformer tenancyinformer.ClusterWorkspaceShardInformer,
+	clusterWorkspaceShardInformer tenancyinformer.ClusterWorkspaceShardInformer,
```
stevekuznetsov, line 56: This seems like an unrelated change?

pkg/server/controllers.go
```diff
@@ -214,19 +208,17 @@ func (s *Server) installKubeServiceAccountTokenController(ctx context.Context, c
 		return fmt.Errorf("error creating service account controller: %w", err)
 	}
 
-	s.AddPostStartHook("kcp-start-kube-service-account-token-controller", func(hookContext genericapiserver.PostStartHookContext) error {
+	return s.AddPostStartHook("kcp-start-kube-service-account-token-controller", func(hookContext genericapiserver.PostStartHookContext) error {
```
stevekuznetsov, line 211: This return pattern seems like another independent change?

pkg/server/options/options.go
```diff
@@ -65,7 +66,7 @@ type ExtraOptions struct {
 
 type completedOptions struct {
 	GenericControlPlane options.CompletedServerRunOptions
-	EmbeddedEtcd        EmbeddedEtcd
+	EmbeddedEtcd        etcdoptions.CompletedObjects
```
stevekuznetsov, line 69: CompletedOptions?

pkg/server/server.go
```diff
@@ -69,193 +71,194 @@ const resyncPeriod = 10 * time.Hour
...
+	kcpAdminToken, shardAdminToken string // nolint:structcheck
+	shardAdminTokenHash            []byte // nolint:structcheck
+
+	// clients
+	DynamicClusterClient       dynamic.ClusterInterface
+	KubeClusterClient          kubernetes.ClusterInterface
+	ApiExtensionsClusterClient apiextensionsclient.ClusterInterface
+	KcpClusterClient           kcpclient.ClusterInterface
+
+	// stuff
```
stevekuznetsov, line 103: ?

pkg/server/server.go
```diff
@@ -69,193 +71,194 @@ const resyncPeriod = 10 * time.Hour
...
-	rootKcpSharedInformerFactory  kcpexternalversions.SharedInformerFactory
-	rootKubeSharedInformerFactory coreexternalversions.SharedInformerFactory
+type ExtraConfig struct {
+	// resolveIdenties is to be called on server start until it succeeds. It injects the kcp
+	// resource identities into the rest.Config used by the client. Only after it succeeds,
+	// the clients can wildcard-list/watch most kcp resources.
+	resolveIdentities func(ctx context.Context) error // nolint:structcheck
+
+	// authentication
+	kcpAdminToken, shardAdminToken string // nolint:structcheck
```
stevekuznetsov, line 94: IMO just turn off the check at the linter level, so when they fix their bugs and we turn it back on we get the full checking?

stevekuznetsov: Not sure we want to chase their false positives across the codebase (is it even deterministic?)

pkg/server/server.go
```diff
@@ -498,12 +528,14 @@ func (s *Server) Run(ctx context.Context) error {
 		}
 
 		return nil
-	})
+	}); err != nil {
+		return err
+	}
 
 	// ========================================================================================================
 	// TODO: split apart everything after this line, into their own commands, optional launched in this process
```
stevekuznetsov, line 536: Can we remove this TODO unless we actually plan to do it?

pkg/server/server.go
```diff
@@ -553,100 +585,66 @@ func (s *Server) Run(ctx context.Context) error {
 		}
 	}
 
-	if s.options.HomeWorkspaces.Enabled {
+	if s.Options.HomeWorkspaces.Enabled {
```
stevekuznetsov, line 588: not your change: but why does enableAll not trigger this one?

pkg/admission/initializers/interfaces.go
sttts: no, it was necessary because we store the ClusterInterface in the config, not the client pointer.

pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_controller.go
sttts: no, we had `rootWorkspaceShardInformer` in the server for reasons that must have disappeared. I removed the `root*` variant of those informers. Hence, also here.

pkg/server/controllers.go
sttts: no. I unified the `Server.AddPostStartHook` signature and lint complained here.

pkg/server/options/options.go
sttts: :D

sttts: fixed

pkg/server/server.go
sttts: If I just bump the version?

pkg/server/server.go
sttts: misc

pkg/server/server.go
sttts: done

pkg/server/server.go
sttts: oh, this will be a follow-up. Part of this PR is to make that possible.

pkg/server/server.go
sttts: it's not a controller, but a feature. I would actually propose to remove `EnableAll` and only allow opt-out (in follow-up).

pkg/embeddedetcd/config.go
sttts: it's technically a new file, only Github shows it as rename.

pkg/embeddedetcd/server.go
```diff
@@ -31,6 +31,9 @@ type Server struct {
 }
 
 func NewServer(config CompletedConfig) *Server {
+	if config.Config == nil {
+		return nil
```
p0lyn0mial, line 35: shouldn't we handle a nil pointer in kcp.go:L134 then?

pkg/server/controllers.go
```diff
@@ -374,19 +362,19 @@ func (s *Server) installWorkspaceScheduler(ctx context.Context, config *rest.Con
 
 	workspaceController, err := clusterworkspace.NewController(
 		kcpClusterClient,
-		s.kcpSharedInformerFactory.Tenancy().V1alpha1().ClusterWorkspaces(),
-		s.rootKcpSharedInformerFactory.Tenancy().V1alpha1().ClusterWorkspaceShards(),
-		s.kcpSharedInformerFactory.Apis().V1alpha1().APIBindings(),
+		s.KcpSharedInformerFactory.Tenancy().V1alpha1().ClusterWorkspaces(),
+		s.KcpSharedInformerFactory.Tenancy().V1alpha1().ClusterWorkspaceShards(),
```
p0lyn0mial, line 366: am I right that `ClusterWorkspaceShards` are only in the root `ws`?

pkg/reconciler/tenancy/clusterworkspaceshard/clusterworkspaceshard_controller.go
```diff
@@ -47,18 +47,18 @@ const (
...
 ) (*Controller, error) {
 	queue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "kcp-workspaceshard")
 
 	c := &Controller{
-		queue:                     queue,
-		kcpClient:                 rootKcpClient,
-		rootWorkspaceShardIndexer: rootWorkspaceShardInformer.Informer().GetIndexer(),
-		rootWorkspaceShardLister:  rootWorkspaceShardInformer.Lister(),
+		queue:                        queue,
+		kcpClient:                    rootKcpClient,
```
p0lyn0mial, line 56: so, this is now `kcpClusterClient`

pkg/server/controllers.go
sttts: yes. But for that we should use informer scoping if we want that.

pkg/embeddedetcd/server.go
sttts: 134 is about the kcp server, not etcd server.

PR #1618, opened 2022-07-26
Refactor ns scheduler for location workspace

qiujian16:
1. add a internal annotation for placement to
schedule synctarget
2. ns controller only update sync label by collecting
the annotation result from the placements

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/apis/workload/v1alpha1/types.go
+++ b/pkg/apis/workload/v1alpha1/types.go
@@ -112,4 +112,8 @@ const (
 	// AnnotationSkipDefaultObjectCreation is the annotation key for an apiexport or apibinding indicating the other default resources
 	// has been created already. If the created default resource is deleted, it will not be recreated.
 	AnnotationSkipDefaultObjectCreation = "workload.kcp.dev/skip-default-object-creation"
+
+	// InternalSyncTargetPlacementAnnotationKey is a internal annotation key on placement API to mark the synctarget scheduled
+	// from this placement. The value is {location workspace}/{syncTarget name}
+	InternalSyncTargetPlacementAnnotationKey = "internal.workload.kcp.dev/synctarget"
 )
--- a/pkg/reconciler/workload/namespace/namespace_reconcile_scheduling.go
+++ b/pkg/reconciler/workload/namespace/namespace_reconcile_scheduling.go
@@ -19,95 +19,34 @@ package namespace
 import (
 	"context"
 	"encoding/json"
-	"math/rand"
 	"strings"
 	"time"
 
 	"github.com/kcp-dev/logicalcluster/v2"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
-	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 
 	schedulingv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/scheduling/v1alpha1"
 	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
-	locationreconciler "github.com/kcp-dev/kcp/pkg/reconciler/scheduling/location"
+	placementreconciler "github.com/kcp-dev/kcp/pkg/reconciler/workload/placement"
 )
 
 const removingGracePeriod = 5 * time.Second
 
-// placementSchedulingReconciler schedules a workload for this ns. It checks the current placement annotation on the ns,
-// and find all valid synctargets.
+// placementSchedulingReconciler sync scheduled synctarget on placement to for this ns.
 type placementSchedulingReconciler struct {
-	listSyncTarget func(clusterName logicalcluster.Name) ([]*workloadv1alpha1.SyncTarget, error)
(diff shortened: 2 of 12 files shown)


pkg/apis/workload/v1alpha1/types.go
```diff
@@ -112,4 +112,8 @@ const (
 	// AnnotationSkipDefaultObjectCreation is the annotation key for an apiexport or apibinding indicating the other default resources
 	// has been created already. If the created default resource is deleted, it will not be recreated.
 	AnnotationSkipDefaultObjectCreation = "workload.kcp.dev/skip-default-object-creation"
+
+	// InternalSyncTargetPlacementAnnotationKey is a internal annotation key on placement API to mark the synctarget scheduled
+	// from this placement. The value is {location workspace}/{syncTarget name}
```
sttts, line 117: Actually this must be the hash of the value. See @jmprusi's open PR.

sttts: can be follow-up depending on which PR merges first.

jmprusi: Let's merge this one first and I will rebase mine with those changes.

pkg/reconciler/workload/placement/placement_controller.go
```diff
@@ -0,0 +1,264 @@
+/*
+Copyright 2021 The KCP Authors.
```
sttts, line 2: 2022

pkg/reconciler/workload/placement/placement_reconcile_scheduling.go
```diff
@@ -0,0 +1,156 @@
...
+	expectedAnnotations := map[string]interface{}{} // nil means to remove the key
+	currentScheduled, foundScheduled := placement.Annotations[workloadv1alpha1.InternalSyncTargetPlacementAnnotationKey]
+
+	// 2. pick all valid synctargets in this placements
+	syncTargetClusterName, syncTargets, err := r.getAllValidSyncTargetsForPlacement(clusterName, placement)
+	if err != nil {
+		return reconcileStatusStop, placement, err
+	}
+
+	// no valid synctarget, clean the annotationt.
```
sttts, line 60:
```suggestion
	// no valid synctarget, clean the annotation.
```

pkg/apis/workload/v1alpha1/types.go
```diff
@@ -112,4 +112,8 @@ const (
...
+	InternalSyncTargetPlacementAnnotationKey = "internal.workload.kcp.dev/synctarget"
```
sttts, line 118:
```suggestion
	InternalSyncTargetPlacementAnnotationKey = "internal.workload.kcp.dev/synctargets"
```

pkg/reconciler/workload/namespace/namespace_reconcile_scheduling.go
```diff
@@ -19,33 +19,28 @@ package namespace
...
 	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
-	locationreconciler "github.com/kcp-dev/kcp/pkg/reconciler/scheduling/location"
+	placementreconciler "github.com/kcp-dev/kcp/pkg/reconciler/workload/placement"
 )
 
 const removingGracePeriod = 5 * time.Second
 
-// placementSchedulingReconciler schedules a workload for this ns. It checks the current placement annotation on the ns,
-// and find all valid synctargets.
+// placementSchedulingReconciler sync scheduled synctarget on placement to for this ns.
```
sttts, line 41:
```suggestion
// placementSchedulingReconciler reconciles the state.workload.kcp.dev/<syncTarget> labels according the
// selected synctarget for each placement stored in the internal.workload.kcp.dev/synctargets annotation.
```

pkg/reconciler/workload/placement/placement_reconcile_scheduling.go
```diff
@@ -0,0 +1,156 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/klog/v2"
+
+	schedulingv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/scheduling/v1alpha1"
+	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
+	locationreconciler "github.com/kcp-dev/kcp/pkg/reconciler/scheduling/location"
+)
+
+// placementSchedulingReconciler schedules a workload for this ns. It checks the current placement annotation on the ns,
+// and find all valid synctargets.
```
sttts, line 40:
```suggestion
// placementSchedulingReconciler schedules namespaces according to the selected locations in the placement
// objects. It considers only valid SyncTargets per locations and updates the internal.workload.kcp.dev/synctarget
// annotation with the selected ones per placement.
```

pkg/apis/workload/v1alpha1/types.go
sttts, line 118: hmm, I thought this is either a) storing a JSON map with the synctarget per placement, or b) we have multiple annotations, one for each workspace.

qiujian16: this annotation is set on placement API. Since each placement only schedule one synctarget (in one selected location), I think we do not need a map. Or if we should set the annotation on the ns, then it should be a JSON map. There is some difference between these two. In the latter approach, if we have two namespaces, they could be scheduled to different synctargetes in a selected location by placement.

sttts:
/approve

@jmprusi for final review.

pkg/apis/workload/v1alpha1/types.go
davidfestal: In the future don't we envision cases when a placement inside a location might result in 2 SyncTargets being scheduled: transparent deployment spreading case for scalability for example ?

sttts: don't think we will have that beyond temporary during migration. At least I would not expect that being done on purpose by the system.

PR #1620, opened 2022-07-26
introduce NewWildcardIdentitiesRoundTripper

p0lyn0mial:
## Summary
NewWildcardIdentitiesRoundTripper creates a HTTP RoundTripper that injected resource identities for
individual group or group resources. Each group or resource is coming from one APIExport whose
names are passed in as a map.

The returned resolve function will get the APIExports and extract the identities.
The resolve func might return an error if the APIExport is not found or for other reasons. Only
after it succeeds a client using the returned RoundTripper can use the group and group resources
with identities.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/server/bootstrap/identity.go
+++ b/pkg/server/bootstrap/identity.go
@@ -84,7 +84,24 @@ func NewConfigWithWildcardIdentities(config *rest.Config,
 	groupExportNames map[string]string,
 	grouptResourceExportNames map[schema.GroupResource]string,
 	kcpClient kcpclient.Interface) (identityConfig *rest.Config, resolve func(ctx context.Context) error) {
+	identityRoundTripper, identityResolver := NewWildcardIdentitiesWrappingRoundTripper(groupExportNames, grouptResourceExportNames, kcpClient)
+	identityConfig = rest.CopyConfig(config)
+	identityConfig.Wrap(identityRoundTripper)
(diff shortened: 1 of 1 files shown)


pkg/server/bootstrap/identity.go
```diff
@@ -84,7 +84,23 @@ func NewConfigWithWildcardIdentities(config *rest.Config,
...
 
+// NewWildcardIdentitiesRoundTripper creates a HTTP RoundTripper that injected resource identities for
+// individual group or group resources. Each group or resource is coming from one APIExport whose
+// names are passed in as a map.
+//
+// The returned resolve function will get the APIExports and extract the identities.
+// The resolve func might return an error if the APIExport is not found or for other reason. Only
+// after it succeeds a client using the returned RoundTripper can use the group and group resources
+// with identities.
+func NewWildcardIdentitiesRoundTripper(groupExportNames map[string]string,
```
sttts, line 101: it's not really the roundtripper, but a roundtripper wrapper, right?

p0lyn0mial: it returns a RoundTripper and a function that allows for wrapping

p0lyn0mial: updated the function name and changed the gdoc.

p0lyn0mial:
/close

pulled into #1572

PR #1628, opened 2022-07-27
Enable avoiding using podman

davidfestal:
## Summary

Allow using the `CreateKindClusters.sh` script without  `podman`, even though podman is installed on the system (for example on linux boxes where you have both `docker` and `podman`)


sttts:
don't care much about this little chance. Overall, IMO this does not really belong into the repository.

PR #1629, opened 2022-07-27
Cleanup Syncer to prepare for Syncer Virtual Workspace transformations

davidfestal:
## Summary

This PR fixes some corner cases or behaviors in the Syncer that would not be compatible with the Syncer Virtual Workspace transformations. More precisely:
- We now [stop reconciling just after adding the Syncer Finalizer](https://github.com/kcp-dev/kcp/pull/1629/commits/af19673c45f1c7e668162919791620d025843f14). The Update of the upstream resource with the syncer finalizer will trigger a new reconcile round anyway. We should avoid updating both the upstream resource and the downstream resource, and thus combine distinct purposes in the same reconciliation round.
- [Don't add the syncer finalizer to a resource already marked for deletion](https://github.com/kcp-dev/kcp/pull/1629/commits/14c391941f063640f48a0350d5b338d905092b93). This will lead to some sort of hot-looping with the VW logic when the deletion process is already started
- [Use upstream informer listers instead the upstream kube client](https://github.com/kcp-dev/kcp/pull/1629/commits/cc96852c4180b913aaacac5d8d310206f98b613d) to get the upstream resource during the reconcile.
- [Fix how upstream status is updated](https://github.com/kcp-dev/kcp/pull/1629/commits/d5adc7a9c5f9c6439cbefd4b4f7d7ebafdfccab2) : we should not update the upstream status based on the downstream resource: this leaks many downstream-related fields to the Syncer virtual workspace, which we don't want.

### DIFF
--- a/pkg/syncer/shared/finalizer.go
+++ b/pkg/syncer/shared/finalizer.go
@@ -23,8 +23,11 @@ import (
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/dynamic/dynamicinformer"
+	"k8s.io/client-go/tools/clusters"
 	"k8s.io/klog/v2"
 
 	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
@@ -36,12 +39,18 @@ const (
 	SyncerFinalizerNamePrefix = "workload.kcp.dev/syncer-"
 )
 
-func EnsureUpstreamFinalizerRemoved(ctx context.Context, gvr schema.GroupVersionResource, upstreamClient dynamic.ClusterInterface, upstreamNamespace, syncTargetName string, logicalClusterName logicalcluster.Name, resourceName string) error {
-	upstreamObj, err := upstreamClient.Cluster(logicalClusterName).Resource(gvr).Namespace(upstreamNamespace).Get(ctx, resourceName, metav1.GetOptions{})
+func EnsureUpstreamFinalizerRemoved(ctx context.Context, gvr schema.GroupVersionResource, upstreamInformers dynamicinformer.DynamicSharedInformerFactory, upstreamClient dynamic.ClusterInterface, upstreamNamespace, syncTargetName string, logicalClusterName logicalcluster.Name, resourceName string) error {
+	upstreamObjFromLister, err := upstreamInformers.ForResource(gvr).Lister().ByNamespace(upstreamNamespace).Get(clusters.ToClusterAwareKey(logicalClusterName, resourceName))
 	if err != nil && !apierrors.IsNotFound(err) {
 		return err
 	}
 	if apierrors.IsNotFound(err) {
+		return nil
+	}
+
+	upstreamObj, ok := upstreamObjFromLister.(*unstructured.Unstructured)
+	if !ok {
+		klog.Errorf("Resource %s|%s/%s expected to be *unstructured.Unstructured, got %T", logicalClusterName.String(), upstreamNamespace, resourceName, upstreamObjFromLister)
 		return nil
 	}
 
@@ -50,6 +59,8 @@ func EnsureUpstreamFinalizerRemoved(ctx context.Context, gvr schema.GroupVersion
 		// Do nothing: the object should not be deleted anymore for this location on the KCP side
 		return nil
 	}
+
+	upstreamObj = upstreamObj.DeepCopy()
 
 	// Remove the syncer finalizer.
 	currentFinalizers := upstreamObj.GetFinalizers()
--- a/pkg/syncer/spec/spec_process.go
+++ b/pkg/syncer/spec/spec_process.go
@@ -174,11 +174,23 @@ func (c *Controller) process(ctx context.Context, gvr schema.GroupVersionResourc
 	}
 
 	// upsert downstream
-	u, ok := obj.(*unstructured.Unstructured)
+	upstreamObj, ok := obj.(*unstructured.Unstructured)
 	if !ok {
 		return fmt.Errorf("object to synchronize is expected to be Unstructured, but is %T", obj)
 	}
-	return c.applyToDownstream(ctx, gvr, downstreamNamespace, u)
+
+	if err := c.ensureDownstreamNamespaceExists(ctx, downstreamNamespace, upstreamObj); err != nil {
+		return err
+	}
+
+	if added, err := c.ensureSyncerFinalizer(ctx, gvr, upstreamObj); added {
+		// The successful update of the upstream resource finalizer will trigger a new reconcile
+		return nil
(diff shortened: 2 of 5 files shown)


pkg/syncer/spec/spec_process.go
```diff
@@ -174,11 +174,23 @@ func (c *Controller) process(ctx context.Context, gvr schema.GroupVersionResourc
...
 	if !ok {
 		return fmt.Errorf("object to synchronize is expected to be Unstructured, but is %T", obj)
 	}
-	return c.applyToDownstream(ctx, gvr, downstreamNamespace, u)
+
+	if err := c.ensureDownstreamNamespaceExists(ctx, downstreamNamespace, upstreamObj); err != nil {
+		return err
+	}
+
+	if addedFinalizerInUpstream, err := c.ensureSyncerFinalizer(ctx, gvr, upstreamObj); addedFinalizerInUpstream {
```
sttts, line 186: nit: `added` is enough. There is enough context to understand.

davidfestal: Fixed

sttts:
what are these corner cases? And why are they not compatible?

pkg/syncer/shared/finalizer.go
```diff
@@ -36,15 +39,23 @@ const (
 	SyncerFinalizerNamePrefix = "workload.kcp.dev/syncer-"
 )
 
-func EnsureUpstreamFinalizerRemoved(ctx context.Context, gvr schema.GroupVersionResource, upstreamClient dynamic.ClusterInterface, upstreamNamespace, syncTargetName string, logicalClusterName logicalcluster.Name, resourceName string) error {
-	upstreamObj, err := upstreamClient.Cluster(logicalClusterName).Resource(gvr).Namespace(upstreamNamespace).Get(ctx, resourceName, metav1.GetOptions{})
+func EnsureUpstreamFinalizerRemoved(ctx context.Context, gvr schema.GroupVersionResource, informers dynamicinformer.DynamicSharedInformerFactory, upstreamClient dynamic.ClusterInterface, upstreamNamespace, syncTargetName string, logicalClusterName logicalcluster.Name, resourceName string) error {
```
sttts, line 42: upstreamInformers

pkg/syncer/shared/finalizer.go
```diff
@@ -36,15 +39,23 @@ const (
...
 		return nil
 	}
 
+	upstreamObj, ok := upstreamObjFromLister.(*unstructured.Unstructured)
+	if !ok {
+		klog.Errorf("Resource %s|%s/%s expected to be *unstructured.Unstructured, got %T", logicalClusterName.String(), upstreamNamespace, resourceName, upstreamObjFromLister)
+		return nil
+	}
+
+	upstreamObj = upstreamObj.DeepCopy()
```
sttts, line 57: do deepcopy further down. This call is super heavy.

pkg/syncer/status/status_process.go
```diff
@@ -130,21 +130,18 @@ func (c *Controller) updateStatusInUpstream(ctx context.Context, gvr schema.Grou
...
-	// TODO: verify that we really only update status, and not some non-status fields in ObjectMeta.
-	//       I believe to remember that we had resources where that happened.
-
-	upstreamObj.SetResourceVersion(existing.GetResourceVersion())
+	existing, ok := existingObj.(*unstructured.Unstructured)
+	if !ok {
+		klog.Errorf("Resource %s|%s/%s expected to be *unstructured.Unstructured, got %T", upstreamLogicalCluster.String(), upstreamNamespace, name, existing)
+		return nil
+	}
+	existing = existing.DeepCopy()
```
sttts, line 137: why deepcopy?

pkg/syncer/status/status_process.go
```diff
@@ -113,38 +113,31 @@ func (c *Controller) process(ctx context.Context, gvr schema.GroupVersionResourc
...
-	transformName(upstreamObj)
-
-	name := upstreamObj.GetName()
-	downstreamStatus, statusExists, err := unstructured.NestedFieldCopy(upstreamObj.UnstructuredContent(), "status")
+	downstreamStatus, statusExists, err := unstructured.NestedFieldCopy(downstreamObj.UnstructuredContent(), "status")
 	if err != nil {
 		return err
 	} else if !statusExists {
-		klog.Infof("Resource doesn't contain a status. Skipping updating status of resource %s|%s/%s from syncTargetName namespace %s", upstreamLogicalCluster, upstreamNamespace, name, downstreamObj.GetNamespace())
+		klog.Infof("Resource doesn't contain a status. Skipping updating status of resource %s|%s/%s from syncTargetName namespace %s", upstreamLogicalCluster, upstreamNamespace, upstreamName, downstreamObj.GetNamespace())
```
sttts, line 122: while touching this, `V(5)` or higher. This irritates people all the time.

pkg/syncer/spec/spec_process_test.go
```diff
@@ -44,7 +44,6 @@ import (
 	"k8s.io/client-go/tools/clusters"
```
sttts, line 44: please keep this in the commits that change the behaviour. Now it is very hard to understand which of the 3 other commits is related.

pkg/syncer/status/status_process_test.go
```diff
@@ -588,12 +586,12 @@ func setupServersideApplyPatchReactor(toClient *dynamicfake.FakeDynamicClient) {
 	})
 }
 
-func setupWatchReactor(resource string, fromClient *dynamicfake.FakeDynamicClient) chan struct{} {
+func setupWatchReactor(resource string, client *dynamicfake.FakeDynamicClient) chan struct{} {
```
sttts, line 589: why unqualify the client?

pkg/syncer/status/status_process.go
davidfestal: It's a mistake

pkg/syncer/shared/finalizer.go
davidfestal: :+1:

pkg/syncer/shared/finalizer.go
davidfestal: Done in the last rebased and force-pushed commits

pkg/syncer/shared/finalizer.go
davidfestal: Done

pkg/syncer/status/status_process.go
davidfestal: Removed

pkg/syncer/status/status_process.go
davidfestal: Done

pkg/syncer/status/status_process_test.go
davidfestal: Because this method is now use on both the fromClient and the toClient of the testCase, since upstream informers are also used to get upstream objects now.

pkg/syncer/spec/spec_process_test.go
davidfestal: I cleaned up the commit history to really have 1 change per commit, and including the related test change.

sttts: :+1:

pkg/syncer/status/status_process_test.go
sttts: ack

PR #1635, opened 2022-07-27
Fix server never starting up sometimes, webhook npe

### LINKED ISSUE #1634: Server never fully bootstrapped
During an e2e test run, kcp never fully bootstrapped, and the test never considered the server ready, so the test failed.

Failed in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kcp-dev_kcp/1236/pull-ci-kcp-dev-kcp-main-e2e-multiple-runs/1552052493460967424.

Probably not super easy to reproduce, but here's what is happening (excluding tasks that are likely unrelated).

1. `kcp-start-informers` post-start hook
    1. Start kube informers & wait for sync
    2. Start apiextensions informers & wait for sync
    3. Bootstrap system CRDs (apiexports, apibindings, apiresourceschemas)
    4. Start APIExports informer
    5. Start APIBindings informer
    6. Wait for APIExports, APIBindings informers to sync
    7. Resolve APIExport identities
    8. Start kcp factory
    9. Wait for kcp factory sync - never gets past here

The problem:
- The APIBinding controller is running as soon as caches are synced for CRDs, APIExports, APIBindings
- It tries to look up APIResourceSchemas, but we haven't waited for its cache sync to be done
- The controller gets lots of 404s looking up APIResourceSchemas - example:

```
E0726 22:30:39.746868   42368 apibinding_reconcile.go:164] Error binding APIBinding system:shard|apiresource.kcp.dev, APIExport root|apiresource.kcp.dev, APIResourceSchema root|v220628-546034da.apiresourceimports.apiresource.kcp.dev: apiresourceschema.apis.kcp.dev "root|v220628-546034da.apiresourceimports.apiresource.kcp.dev" not found                                                                                                                                                        
```

- At some point, some of the bindings go partially through, but at least one is in the following state:
    - It points to an APIExport that has multiple `latestResourceSchemas` in the list
    - The APIBinding controller successfully finds at least 1 APIResourceSchema and starts accumulating .status.boundResources
    - It then encounters at least 1 APIResourceSchema it cannot find via the lister, and returns early
    - The controller patches the APIBinding's .status.conditions to indicate there was an error **and** it also includes some of the bound resources
    - The patch does not set .status.boundAPIExport

Next, at some point there's some part of startup that goes through the generic webhook code, and there's a bug there that doesn't handle .status.boundAPIExport being nil:

https://github.com/kcp-dev/kcp/blob/0f4ef8666851b6546d48f80f6a41d74b11085f67/pkg/admission/webhook/generic_webhook.go#L99

Proposed fixes:
1. Wait for APIResourceSchema cache to be synced alongside the waits for APIExports, APIBindings
2. Handle a nil .status.boundAPIExport

ncdc:
## Summary
- Wait for APIResourceSchemas cache to be synced along with APIExports and APIBindings
- Fix NPE in webhook code

## Related issue(s)

Fixes #1634

### DIFF
--- a/pkg/admission/webhook/generic_webhook.go
+++ b/pkg/admission/webhook/generic_webhook.go
@@ -96,7 +96,7 @@ func (p *WebhookDispatcher) getAPIBindingWorkspace(attr admission.Attributes, cl
 	for _, obj := range objs {
 		apiBinding := obj.(*apisv1alpha1.APIBinding)
 		for _, br := range apiBinding.Status.BoundResources {
-			if apiBinding.Status.BoundAPIExport.Workspace == nil {
+			if apiBinding.Status.BoundAPIExport != nil && apiBinding.Status.BoundAPIExport.Workspace == nil {
 				// this will never happen today. But as soon as we add other reference types (like exports), this log output will remind out of necessary work here.
 				klog.Errorf("APIBinding %s has no referenced workspace", clusterName, apiBinding.Name)
 				continue
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -445,18 +445,20 @@ func (s *Server) Run(ctx context.Context) error {
 
 		go s.KcpSharedInformerFactory.Apis().V1alpha1().APIExports().Informer().Run(ctx.StopCh)
 		go s.KcpSharedInformerFactory.Apis().V1alpha1().APIBindings().Informer().Run(ctx.StopCh)
+		go s.KcpSharedInformerFactory.Apis().V1alpha1().APIResourceSchemas().Informer().Run(ctx.StopCh)
 
 		if err := wait.PollInfiniteWithContext(goContext(ctx), time.Millisecond*100, func(ctx context.Context) (bool, error) {
 			exportsSynced := s.KcpSharedInformerFactory.Apis().V1alpha1().APIExports().Informer().HasSynced()
 			bindingsSynced := s.KcpSharedInformerFactory.Apis().V1alpha1().APIBindings().Informer().HasSynced()
-			return exportsSynced && bindingsSynced, nil
+			schemasSynced := s.KcpSharedInformerFactory.Apis().V1alpha1().APIResourceSchemas().Informer().HasSynced()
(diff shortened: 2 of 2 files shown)


pkg/admission/webhook/generic_webhook.go
```diff
@@ -96,7 +96,7 @@ func (p *WebhookDispatcher) getAPIBindingWorkspace(attr admission.Attributes, cl
 	for _, obj := range objs {
 		apiBinding := obj.(*apisv1alpha1.APIBinding)
 		for _, br := range apiBinding.Status.BoundResources {
-			if apiBinding.Status.BoundAPIExport.Workspace == nil {
+			if apiBinding.Status.BoundAPIExport != nil && apiBinding.Status.BoundAPIExport.Workspace == nil {
```
sttts, line 99: I thought OpenAPI guarantees this?

sttts:
Good find!

/lgtm
/approve

pkg/admission/webhook/generic_webhook.go
ncdc: It's optional, a pointer, omitempty. I don't see anything that does. Not in the -patch

ncdc:
/hold

Fixing some stuff I missed

pkg/admission/webhook/generic_webhook.go
sttts: You are right. What I meant is about the spec reference, not the status:
```
- op: add
  path: /spec/versions/name=v1alpha1/schema/openAPIV3Schema/properties/spec/properties/reference/oneOf
  value:
  - required: ["workspace"]
```

ncdc:
I'm seeing lots of variability with the server maybe starting, maybe not when I add in waiting for APIResourceSchemas to sync. I'd like more time to understand the exact behavior, so in the meantime, I've opened #1639 with just the webhook npe fix.

ncdc:
Closing this for the time being. I may come back to it if it looks like we need it. But it's not mergeable in its current form.
/close

PR #1636, opened 2022-07-27
Part 7: Wrap authorization and VW controllers

varshaprasad96:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/authorization/delegated/authorizer.go
+++ b/pkg/authorization/delegated/authorizer.go
@@ -30,14 +30,14 @@ import (
 	"k8s.io/klog/v2"
 )
 
-type DelegatedAuthorizerFactory func(clusterName logicalcluster.Name, client kubeclient.ClusterInterface) (authorizer.Authorizer, error)
+type DelegatedAuthorizerFactory func(clusterName logicalcluster.Name, client kubeclient.Interface) (authorizer.Authorizer, error)
 
 // NewDelegatedAuthorizer returns a new authorizer for use in e.g. admission plugins that delegates
 // to the kube API server via SubjectAccessReview.
-func NewDelegatedAuthorizer(clusterName logicalcluster.Name, client kubeclient.ClusterInterface) (authorizer.Authorizer, error) {
+func NewDelegatedAuthorizer(clusterName logicalcluster.Name, client kubeclient.Interface) (authorizer.Authorizer, error) {
 	delegatingAuthorizerConfig := &authorizerfactory.DelegatingAuthorizerConfig{
 		SubjectAccessReviewClient: &clusterAwareAuthorizationV1Client{
-			AuthorizationV1Interface: client.Cluster(clusterName).AuthorizationV1(),
+			AuthorizationV1Interface: client.AuthorizationV1(),
 			cluster:                  clusterName,
 		},
 		AllowCacheTTL:       5 * time.Minute,
--- a/pkg/authorization/delegated/authorizer.go
+++ b/pkg/authorization/delegated/authorizer.go
@@ -30,14 +30,14 @@ import (
 	"k8s.io/klog/v2"
 )
 
-type DelegatedAuthorizerFactory func(clusterName logicalcluster.Name, client kubeclient.Interface) (authorizer.Authorizer, error)
+type DelegatedAuthorizerFactory func(clusterName logicalcluster.Name, clusterClient kubeclient.Interface) (authorizer.Authorizer, error)
 
 // NewDelegatedAuthorizer returns a new authorizer for use in e.g. admission plugins that delegates
 // to the kube API server via SubjectAccessReview.
-func NewDelegatedAuthorizer(clusterName logicalcluster.Name, client kubeclient.Interface) (authorizer.Authorizer, error) {
+func NewDelegatedAuthorizer(clusterName logicalcluster.Name, clusterClient kubeclient.Interface) (authorizer.Authorizer, error) {
 	delegatingAuthorizerConfig := &authorizerfactory.DelegatingAuthorizerConfig{
 		SubjectAccessReviewClient: &clusterAwareAuthorizationV1Client{
-			AuthorizationV1Interface: client.AuthorizationV1(),
+			AuthorizationV1Interface: clusterClient.AuthorizationV1(),
 			cluster:                  clusterName,
 		},
 		AllowCacheTTL:       5 * time.Minute,
--- a/pkg/virtual/apiexport/builder/build.go
+++ b/pkg/virtual/apiexport/builder/build.go
@@ -50,7 +50,7 @@ const VirtualWorkspaceName string = "apiexport"
 
 func BuildVirtualWorkspace(
 	rootPathPrefix string,
-	kubeClusterClient kubernetes.ClusterInterface,
+	kubeClusterClient kubernetes.Interface,
 	dynamicClusterClient dynamic.ClusterInterface,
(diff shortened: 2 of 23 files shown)


pkg/authorization/delegated/authorizer.go
```diff
@@ -30,14 +30,14 @@ import (
 	"k8s.io/klog/v2"
 )
 
-type DelegatedAuthorizerFactory func(clusterName logicalcluster.Name, client kubeclient.ClusterInterface) (authorizer.Authorizer, error)
+type DelegatedAuthorizerFactory func(clusterName logicalcluster.Name, client kubeclient.Interface) (authorizer.Authorizer, error)
 
 // NewDelegatedAuthorizer returns a new authorizer for use in e.g. admission plugins that delegates
 // to the kube API server via SubjectAccessReview.
-func NewDelegatedAuthorizer(clusterName logicalcluster.Name, client kubeclient.ClusterInterface) (authorizer.Authorizer, error) {
+func NewDelegatedAuthorizer(clusterName logicalcluster.Name, client kubeclient.Interface) (authorizer.Authorizer, error) {
```
sttts, line 37: clusterClient

pkg/authorization/delegated/authorizer.go
```diff
@@ -30,14 +30,14 @@ import (
 	"k8s.io/klog/v2"
 )
 
-type DelegatedAuthorizerFactory func(clusterName logicalcluster.Name, client kubeclient.ClusterInterface) (authorizer.Authorizer, error)
+type DelegatedAuthorizerFactory func(clusterName logicalcluster.Name, client kubeclient.Interface) (authorizer.Authorizer, error)
```
sttts, line 33: clusterClient

pkg/authorization/delegated/authorizer.go
sttts, line 37: 🔴  this signature is a blocker. What is `clusterName` now used for? We cannot store the clusterName and at the same time expect a context with the same clusterName. That is hard to reason about will definitely break somewhere someday.

sttts: actually something is missing here. Compare line 81. There is already some cluster aware custom client. Does line 81 also need a rewrite with context?

pkg/server/server.go
```diff
@@ -204,6 +205,12 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// separating scoped cluster client
+	scopedClusterKubeClient, err := kubernetes.NewForConfig(kcpclienthelper.NewClusterConfig(c.GenericConfig.LoopbackClientConfig))
```
sttts, line 210: what is a scoped cluster client?

pkg/virtual/apiexport/builder/build.go
```diff
@@ -188,7 +188,7 @@ func BuildVirtualWorkspace(
 	}
 }
 
-func getAuthorizer(client kubernetes.ClusterInterface) authorizer.AuthorizerFunc {
+func getAuthorizer(client kubernetes.Interface) authorizer.AuthorizerFunc {
```
sttts, line 191: clusterClient

pkg/virtual/workspaces/authorization/watch_test.go
```diff
@@ -62,7 +62,7 @@ func newTestWatcher(username string, groups []string, predicate storage.Selectio
 	informers := tenancyInformers.NewSharedInformerFactory(mockClient, controller.NoResyncPeriodFunc())
 	workspaceCache := workspacecache.NewClusterWorkspaceCache(
 		informers.Tenancy().V1alpha1().ClusterWorkspaces().Informer(),
-		&mockClusterClient{mockClient: mockClient},
```
sttts, line 65: is mockClusterClient now unused? Can we remove it?

pkg/virtual/workspaces/builder/build.go
```diff
@@ -145,7 +145,7 @@ func BuildVirtualWorkspace(cfg *clientrest.Config, rootPathPrefix string, wildca
 						return nil, err
 					}
 
-					workspacesRest := registry.NewREST(cfg, kcpClusterClient.Cluster(tenancyv1alpha1.RootCluster).TenancyV1alpha1(), kubeClusterClient, kcpClusterClient, globalClusterWorkspaceCache, crbInformer, orgListener.FilteredClusterWorkspaces)
+					workspacesRest := registry.NewREST(cfg, kubeClusterClient, kcpClusterClient, globalClusterWorkspaceCache, crbInformer, orgListener.FilteredClusterWorkspaces)
```
sttts, line 148: where is the counterpart of the RootCluster here? Where did that go?

pkg/authorization/delegated/authorizer.go
varshaprasad96: `clusterName` is used to scope the RESTClient calls. In Line 81, we are setting the cluster name to the [request](https://pkg.go.dev/k8s.io/client-go/rest#Request) directly. Actually, we aren't expecting the context with the clusterName at all in the DelegatedAuthorizer. If we see here, passing context without cluster name (https://github.com/kcp-dev/kcp/blob/87335b07e67c2be69e3e367a544f14f1dec6df62/pkg/admission/apibinding/apibinding_admission.go#L200) while calling `Authorize` does work.

pkg/server/server.go
varshaprasad96: So we have config's cluster Kube client (https://github.com/kcp-dev/kcp/blob/6e99317ea3a6a9fe10722f8c7fdb1dd50883c461/pkg/server/config.go#L187) but we can't change it right away to accept `Kubernetes.Interface` till informers (https://github.com/kcp-dev/kcp/blob/6e99317ea3a6a9fe10722f8c7fdb1dd50883c461/pkg/server/config.go#L192) are scoped. That's why I created `scopedClusterKubeClient` for now, but it would be removed when we scoped informers.

pkg/virtual/workspaces/authorization/watch_test.go
varshaprasad96: Yes, it was just wrapping the `Kubermetes.Interface` with `.Cluster()`. We don't need it anymore

pkg/virtual/workspaces/builder/build.go
varshaprasad96: `RootClusterClient` wasn't being used at all (https://github.com/kcp-dev/kcp/blob/v0.6.3/pkg/virtual/workspaces/registry/rest.go) and was redundant. So removed it.

pkg/virtual/workspaces/authorization/watch_test.go
sttts, line 65: can you remove mockClusterClient if it is not used anymore?

pkg/virtual/workspaces/builder/build.go
sttts: ack. Left over from early implementation.

sttts:
/approve
Lgtm modulo red tests.

varshaprasad96:
@stevekuznetsov Yes. But before this, a few more changes in other parts of code need to merged. I'll rebase and update after those. Should have made this last in the sequence.

varshaprasad96:
Closing in favor of https://github.com/kcp-dev/kcp/pull/1753 and https://github.com/kcp-dev/kcp/pull/1810

PR #1642, opened 2022-07-29
e2e: reduce global, trivial Sheriff fixture

sttts:
Towards no global fixture libraries with helper inflation.


jmprusi:
```
fixture.go:183: Fixture setup failed: one or more servers did not become ready
```

/test e2e-multiple-runs

PR #1643, opened 2022-07-29
virtual: give VWs an order in order to control precedence

sttts:
To allow shadowing one VW another, e.g. a by-default APIBinding VW shadowed by a PermissionClaim VW: if the user does not claim APIBindings, just the bindings for their own exports are visible. But through a claim all bindings in those workspaces become visible. This needs an order of VW layers.

### DIFF
--- a/pkg/virtual/initializingworkspaces/builder/build.go
+++ b/pkg/virtual/initializingworkspaces/builder/build.go
@@ -55,6 +55,7 @@ import (
 	"github.com/kcp-dev/kcp/pkg/virtual/framework/dynamic/apiserver"
 	dynamiccontext "github.com/kcp-dev/kcp/pkg/virtual/framework/dynamic/context"
 	"github.com/kcp-dev/kcp/pkg/virtual/framework/handler"
+	"github.com/kcp-dev/kcp/pkg/virtual/framework/rootapiserver"
 	"github.com/kcp-dev/kcp/pkg/virtual/initializingworkspaces"
 )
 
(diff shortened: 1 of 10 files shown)


pkg/virtual/initializingworkspaces/builder/build.go
```diff
@@ -295,10 +296,10 @@ func BuildVirtualWorkspace(
 		}),
 	}
 
-	return map[string]framework.VirtualWorkspace{
-		wildcardWorkspacesName: wildcardWorkspaces,
-		workspacesName:         workspaces,
-		workspaceContentName:   workspaceContent,
+	return []rootapiserver.NamedVirtualWorkspace{
```
p0lyn0mial, line 299: is the order given by the fact that we store workspaces in an array?

pkg/virtual/options/options.go
```diff
@@ -92,22 +90,17 @@ func (o *Options) NewVirtualWorkspaces(
...
+func merge(sets ...[]rootapiserver.NamedVirtualWorkspace) ([]rootapiserver.NamedVirtualWorkspace, error) {
+	var workspaces []rootapiserver.NamedVirtualWorkspace
 	for _, set := range sets {
-		for name, w := range set {
-			if _, ok := workspaces[name]; ok {
-				return nil, fmt.Errorf("duplicate virtual workspace %q", name)
-			}
-			workspaces[name] = w
-		}
+		workspaces = append(workspaces, set...)
```
p0lyn0mial, line 103: no need to check for duplicates anymore?

pkg/virtual/initializingworkspaces/builder/build.go
sttts: yes. First comes first serves.

pkg/virtual/options/options.go
sttts: good point. Readding.

sttts: done

PR #1653, opened 2022-08-01
runs e2e-sharded with two shards

p0lyn0mial:
## Related issue(s)

Fixes #


sttts:
unit tests

pkg/admission/clusterworkspaceshard/admission.go
```diff
@@ -51,9 +51,10 @@ func Register(plugins *admission.Plugins) {
 type clusterWorkspaceShard struct {
 	*admission.Handler
 
-	shardBaseURL            string
-	shardExternalURL        string
-	externalAddressProvider func() string
+	shardBaseURL             string
+	shardExternalURL         string
+	shardVirtualWorkspaceURL string
+	externalAddressProvider  func() string
```
sttts, line 57: can you double check: I think in server.go we have a redirect to some external VW URL, maybe including a flag. When @davidfestal removes the personal workspace support, and with your change here, that will probably be not needed anymore very soon.

Issue #1544, opened 2022-07-15
Add known internal types to internalschemas for the virtual workspace to use (RBAC)


shawn-hurley:
The main set of tests that we need to consider is {Cluster}RoleBindings and {Cluster}Roles.

ncdc:
Need to discuss if this should be for all virtual workspace types, or only for APIExport/permission claims

shawn-hurley:
It sounds like the fix is:

1. a way for a given virtual workspace to ask for a set of APIs that it considers to be internal.
2. Add RBAC resources for the APIExport virtual workspace.
3. Update the CEL validation for permission claims (#1529) to allow claiming the RBAC resources.

If a cluster admin is accepting these claims, I would like them to see what rules the services provider will be controlling or needing access to and potentially locking this down. I wonder if this is a follow-up, though.

ncdc:
@robinbobbitt do you still have bandwidth to work on this?

PR #1665, opened 2022-08-02
apis/apis: allow apis.kcp.dev claims without identity

sttts:
apis.kcp.dev does not have an identity.


shawn-hurley (commented):
Probably could add some test cases for these updates? 

Otherwise LGTM

sttts:
Will come with https://github.com/kcp-dev/kcp/pull/1563. Maybe let's wait for that.

ncdc:
The work @robinbobbitt is doing for #1544 might result in us needing to remove this CEL validation entirely, or generate it... maybe worth a discussion

sttts:
Not sure we have to remove it or can extend it. But I agree that the value is limited.

sttts:
Leaving as part of https://github.com/kcp-dev/kcp/pull/1563.

PR #1676, opened 2022-08-03
make internal apis available for apiexport permission claims

### LINKED ISSUE #1544: Add known internal types to internalschemas for the virtual workspace to use (RBAC)

robinbobbitt:
## Summary
- Separates concept of "internal APIs" for syncer vs APIExport virtual workspace
- Makes all discovered non-KCP APIs available as permission claims for APIExport virtual workspace
- Modifies current CEL validations on APIBinding and APIExport to accommodate expanded options
- Does NOT include CustomResourceDefinition as available permission claim as its schema is recursive and causes issues in existing code.. This support could be added in a follow-up if required.

## Related issue(s)

Fixes #1544


pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_controller.go
```diff
@@ -121,17 +136,110 @@ type APIReconciler struct {
...
+				continue
+			}
+
+			raw, err := runtime.DefaultUnstructuredConverter.ToUnstructured(instance)
+			if err != nil {
+				klog.Errorf("error checking status for gvk: %v/%v %v err: %v", gv.Group, gv.Version, apiResource.Kind, err)
+				continue
+			}
+			u := unstructured.Unstructured{Object: raw}
+			_, hasStatus, _ := unstructured.NestedStringMap(u.Object, "status")
```
shawn-hurley, line 186: I don't know, but just asking, would status be set here if it is omitempty?

robinbobbitt: Doh! Great catch. I see LocalSubjectAccessReview is coming out false for this reason. Do you have any other thoughts on how to figure this out?

config/crds/apis.kcp.dev_apiexports.yaml-patch
```diff
@@ -8,7 +8,5 @@
 - op: add
   path: /spec/versions/name=v1alpha1/schema/openAPIV3Schema/properties/spec/properties/permissionClaims/items/x-kubernetes-validations
   value:
-    - rule: "(has(self.group) && self.group != '') || self.resource in ['configmaps', 'namespaces', 'secrets', 'serviceaccounts']"
-      message: .resource must be one of configmaps, namespaces, secrets, serviceaccounts if .group is empty.
-    - rule: "!(has(self.group) && self.group != '') || (has(self.identityHash) && self.identityHash != '')"
-      message: .identityHash must be set if .group is not empty.
\ No newline at end of file
+    - rule: "!(has(self.group) && self.group != '' && !self.group.endsWith('.k8s.io')) || (has(self.identityHash) && self.identityHash != '')"
+      message: .identityHash must be set if .group is not empty or *.k8s.io.
```
sttts, line 12: we cannot have `*.k8s.io` here. Why not list all resources if care about these validations?

pkg/virtual/apiexport/builder/build.go
```diff
@@ -177,6 +180,11 @@ func BuildVirtualWorkspace(
 					}
 				}
 
+				err := apiReconciler.GatherInternalSchemas()
```
sttts, line 183: this is concerning. If I am not mistaken it puts an order on kcp server and virtual workspace server startup, i.e. it's not deterministic during upgrades.

config/crds/apis.kcp.dev_apiexports.yaml-patch
ncdc: Was chatting with @sttts - we should probably just remove this entire validation for the time being, unless we can easily list out every possible group-resource-identityHash.

sttts: discussed briefly with Andy. I agree, we should really remove these validations because we have a hard time to make them complete and correct.

Also notice that every change in resources we allow for claim is technically a breaking change, either in upgrade or in downgrade. Hence, technically we would have to use ratcheting validation here which makes it even more complicated and hard to reason about.

pkg/virtual/apiexport/builder/build.go
ncdc: Just chatted with @sttts about this. We've agreed that a better direction is to have a hard-coded list of internal APIs that we manually curate.

We can take the implementation for `apiReconciler.GatherInternalSchemas()` and move it into an e2e test to check for drift: compare the list you generate via discovery with the hard-coded list, and if they differ, fail. This will allow us to know when we need to update the list during future kube rebases.

pkg/virtual/syncer/controllers/apireconciler/syncer_apireconciler_reconcile.go
```diff
@@ -161,3 +167,62 @@ func gvrString(gvr schema.GroupVersionResource) string {
...
+	openAPIDefinitionsGetters := []common.GetOpenAPIDefinitions{generatedopenapi.GetOpenAPIDefinitions}
+
+	if apis, err := internalapis.CreateAPIResourceSchemas(schemes, openAPIDefinitionsGetters, syncerInternalAPIs...); err != nil {
+		panic(err)
+	} else {
+		syncerSchemas = apis
+	}
+}
+
+// KCPInternalAPIs provides a list of InternalAPI for the APIs that are part of the KCP scheme and will be there in every KCP workspace
```
ncdc, line 185:
```suggestion
// syncerInternalAPIs provides a list of built-in APIs that are available for all workspaces accessed via the syncer virtual workspace.
```

pkg/virtual/apiexport/builder/build.go
robinbobbitt: I wasn't sure where exactly to do this. We obviously don't want to run it on every Reconcile. Perhaps the first Reconcile pays the price, or is there another spot that's later than this but before a Reconcile runs?

config/crds/apis.kcp.dev_apiexports.yaml-patch
robinbobbitt: I can remove these entirely.

pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_controller.go
robinbobbitt: @sttts do you have any suggestions here on another way to do this check?

ncdc, line 186: Instead of this, what if you maintained a separate `map` of maybe gvk to `internalapis.InternalAPI`, and check if `apiResource.Name` has a suffix of `/status`. If so, strip off `/status` to get the name, and then look up the entry in the map (creating if needed) and set `HasStatus=true`?

ncdc: Just submitted another comment w/a possible solution

ncdc: https://github.com/kcp-dev/kcp/pull/1676#discussion_r938072016

robinbobbitt: Thank you! I'll get going on this.

shawn-hurley: This would only get the resources that have status subresources as having status. This might not be an issue, but I wanted to ask to make sure I understand.

ncdc: I don't believe it's possible to have status and not have the status subresource. Have you encountered otherwise?

shawn-hurley: For internal API's I don't think so, for some CRDs, yeah, I have. it might not be a problem just making sure that I understand :)

pkg/virtual/apiexport/builder/build.go
sttts: it's not about about when a first discovery happens. If we depend on something dynamic as discovery we **must** regularly update it. Otherwise the order I mention above becomes a thing, and then we can undeterministic behaviour. That's super hard to debug when it happens in production.

As claims are a critical piece of the system behaviour and of the APIs we offer, we should rather not depend on some dynamic behaviour with a heuristic for filtering, but maintain a static list that we can review on every rebase. A unit or e2e test as @ncdc suggests will help us to detect drift on rebase.

pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_controller.go
sttts: Technically that can happen, sure. The status subresource is totally opt-in. But clearly in practice nobody will use status without the subresource.

robinbobbitt:
Requested changes are in. Please take another look. Thanks!

ncdc:
This looks fantastic, thanks! Mind squashing the commits?

pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_reconcile.go
```diff
@@ -241,10 +256,286 @@ func (c *APIReconciler) getSchemasFromAPIExport(apiExport *apisv1alpha1.APIExpor
...
+	{
+		Names: apiextensionsv1.CustomResourceDefinitionNames{
+			Plural:   "tokenreviews",
+			Singular: "tokenreview",
+			Kind:     "TokenReview",
+		},
+		GroupVersion:  schema.GroupVersion{Group: "authentication.k8s.io", Version: "v1"},
+		Instance:      &authenticationv1.TokenReview{},
+		ResourceScope: apiextensionsv1.ClusterScoped,
+	},
```
sttts, line 405: do we have a list of resources we don't want? This and several more further down look surprising.

sttts:
Looks much better. Well done! 

The few review types look surprising though.

Issue #1518, opened 2022-07-14
To use CWT the user needs to have "access" permissions in workspaces where the CWT exists


MatousJobanek:
**Describe the bug**
Consider that there is a CWT `Foo` in a workspace `root:org:plane`. The workspace contains also a CR & CRB granting `use` permissions to the `system:authenticated` group.
Then a user `bob` wants to create a workspace of the type `Foo` in his home workspace, but this fails:
```
$ k ws create bar --type root:org:plane:Foo
Error: workspaces.tenancy.kcp.dev "bar" is forbidden: use of the cluster workspace type "root:org:plane:Foo" in workspace "root:users:qa:kh:bob" is not allowed
```
the kcp logs contain:
```
E0713 14:19:32.825814  472150 rest.go:629] user "bob" lacks (denied) clusterworkspacetypes "use" permission for "foo" in root:users:qa:kh:bob: workspace access not permitted
```
This is because the user `bob` doesn't have any permission in the `root:org:plane` workspace. When kcp does SAR for the CWT then it expects that the user has `access` permissions in the workspace (and also all parent workspaces).

**To Reproduce**
1. Create CWT `Foo` in `root:org:plane` (with the necessary CR & CRB granting `use` permissions)
2. `k ws create bar --type root:org:plane:Foo`

**Expected behavior**
it should not require `access` permissions in the workspace where CWT is defined.

**Additional context**
https://coreos.slack.com/archives/C020QAF9D46/p1657716975579069

MatousJobanek:
This technically applies to all SARs that are executed for any other resource - for example an APIExport

ncdc:
This might be helped with a generic Export mechanism, where users could export things (CWT, APIExport) into a "catalog" in the root

PR #1680, opened 2022-08-04
authorization: add system:masters-only deep SAR via X-Kcp-Internal-Deep-SubjectAccessReview:true header

sttts:
This will allow us
- to authorize APIBindings to bind without actual APIExport workspace access
- to authorize use of CWTs without actual CWT workspace access (fixes https://github.com/kcp-dev/kcp/issues/1518)
- to authorize permission claim access to fall under MaximalPermissionClaims of the APIExport owner.

TODO:
- [x] ~~clarify whether we want to map the users somehow with a prefix~~ only local (SA) users would need a prefix, see next point.
- [ ] clarify whether we want to keep restricting service accounts to their workspace, and/or wipe their usernames when not having workspace access

### DIFF
--- a/pkg/authorization/deep_sar.go
+++ b/pkg/authorization/deep_sar.go
@@ -1,0 +1,115 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package authorization
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	authorization "k8s.io/api/authorization/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	kuser "k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/apiserver/pkg/endpoints/handlers/responsewriters"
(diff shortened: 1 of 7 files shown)


pkg/authorization/deep_sar.go
```diff
@@ -0,0 +1,115 @@
...
+
+// WithDeepSARConfig returns a clone of the input rest.Config
+// with an additional header making SARs to be deep.
+func WithDeepSARConfig(config *rest.Config) *rest.Config {
+	clone := *config
+	clone.Wrap(func(rt http.RoundTripper) http.RoundTripper {
+		return &withHeaderRoundtripper{
+			RoundTripper: rt,
+			headers: map[string]string{
+				deepSARHeader: "1",
```
sttts, line 50: "1" look strange. "true" probably is more common.

sttts:
This will also help to get rid of the "bind to an ancestor or a child of an ancestor" rule. This is related to the API catalog discussion of binding against an APIExportKey cc @ncdc

pkg/authorization/deep_sar.go
sttts: fixed

pkg/authorization/deep_sar.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+type withHeaderRoundtripper struct {
+	http.RoundTripper
+
+	headers map[string]string
+}
+
+func (rt *withHeaderRoundtripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	req = req.Clone(req.Context())
```
s-urbaniak, line 64: why is this necessary? it looks like a noop.

s-urbaniak: nevermind:
```
	// RoundTrip should not modify the request, except for
	// consuming and closing the Request's Body. RoundTrip may
	// read fields of the request in a separate goroutine. Callers
	// should not mutate or reuse the request until the Response's
	// Body has been closed.
```

s-urbaniak:
in favor of https://github.com/kcp-dev/kcp/pull/1739

Issue #1679, opened 2022-08-04
When synctarget is being deleted and added back all the pods running in the previous syncer namespaces starts to run again


kasturinarra:
**Describe the bug**
When the synctarget is deleted and added back i see that pod comes back and starts running in the previous syncer namespaces

[knarra@knarra ~]$ oc get pods --all-namespaces | grep kuard
kcp-27tnpxttv54u                                   kuard-5bdc44b494-hf62w                                                      1/1     Running             0              13m
kcp-hv00zupix4tz                                   kuard-5bdc44b494-lq5rs                                                      1/1     Running             0              14m

**To Reproduce**
1. Deploy kcp on openshift cluster
2. Add a workload cluster by running the command `kubectl kcp workload sync qecluster1 --syncer-image ghcr.io/kcp-dev/kcp/syncer:8ad1c4a -o - | KUBECONFIG=<pcluster> kubectl apply -f - `
3. Now verify that it has been added successfully using the command `kubectl get synctarget -o wide’
4. Create deployment using the command ` kubectl create deployment kuard --image gcr.io/kuar-demo/kuard-amd64:blue --dry-run=client -o yaml > kuard.yaml`
5. Apply the kuard.yaml using the command `kubectl apply -f kuard.yaml`
6. Now verify that deployment has been synced to downstream cluster by running the command ‘oc get pods --all-namespaces | grep kuard’
[knarra@knarra ~]$ oc get pods --all-namespaces | grep kuard
kcp-27tnpxttv54u                                   kuard-5bdc44b494-twcbt                                                      1/1     Running             0              11s
7. Now add another synctarget by running the command `kubectl kcp workload sync qecluster1 --syncer-image ghcr.io/kcp-dev/kcp/syncer:8ad1c4a -o - | KUBECONFIG=/home/knarra/Downloads/kubeconfig_411 kubectl apply -f -`
8. Wait for the synctarget to be added successfully
9. Now delete the first synctarget by running the command `kubectl delete synctarget qecluster1`
10. Verify that deployment moved to the second synctarget added which is `qecluster2` by running the command `oc get pods --all-namespaces | grep kuard`
[knarra@knarra ~]$ oc get pods --all-namespaces | grep kuard
kcp-1cmkuvs9epei                                   kuard-5bdc44b494-45dhs                                                      1/1     Running     0              11m
[...]

PR #1687, opened 2022-08-05
Syncer: refuse to work on sync target UID discrepancy.

jmprusi:
The changes in this PR are focused on avoiding a Syncer running with an old SyncTarget, this is accomplished by:

- Syncer: 
   - New flag for providing the expected sync target UID: `--sync-target-uid`
   - When starting will check that the provided UID matches the existing sync target UID, if not, fail to start
   - When patching a sync target status for setting the heartbeat, tries to match the expected sync target UID, if not, refuses to patch 
- KCP Kubectl plugin: The generated manifests include the sync target UID flag.
- SyncTarget controller: Appends the desired sync target UID to the VirtualWorkspace URLs.
- Syncer VirtualWorkspace: validates that the sync target name and UID in the request 
match.
  
related to: https://github.com/kcp-dev/kcp/issues/1679

### DIFF
--- a/pkg/cliplugins/workload/plugin/sync.go
+++ b/pkg/cliplugins/workload/plugin/sync.go
@@ -86,7 +86,7 @@ func (c *Config) Sync(
 		defer outputFile.Close() // nolint: errcheck
 	}
 
-	token, syncerID, err := c.enableSyncerForWorkspace(ctx, config, syncTargetName, kcpNamespaceName)
+	token, syncerID, syncTargetUID, err := c.enableSyncerForWorkspace(ctx, config, syncTargetName, kcpNamespaceName)
 	if err != nil {
 		return err
 	}
@@ -115,6 +115,7 @@ func (c *Config) Sync(
 		Namespace:       downstreamNamespace,
 		LogicalCluster:  currentClusterName.String(),
 		SyncTarget:      syncTargetName,
+		SyncTargetUID:   syncTargetUID,
 		Image:           image,
 		Replicas:        replicas,
 		ResourcesToSync: resourcesToSync,
@@ -147,10 +148,10 @@ func getSyncerID(syncTarget *workloadv1alpha1.SyncTarget) string {
 // enableSyncerForWorkspace creates a sync target with the given name and creates a service
 // account for the syncer in the given namespace. The expectation is that the provided config is
 // for a logical cluster (workspace). Returns the token the syncer will use to connect to kcp.
-func (c *Config) enableSyncerForWorkspace(ctx context.Context, config *rest.Config, syncTargetName, namespace string) (string, string, error) {
+func (c *Config) enableSyncerForWorkspace(ctx context.Context, config *rest.Config, syncTargetName, namespace string) (saToken string, syncerID string, syncTargetUID string, err error) {
 	kcpClient, err := kcpclientset.NewForConfig(config)
 	if err != nil {
-		return "", "", fmt.Errorf("failed to create kcp client: %w", err)
+		return "", "", "", fmt.Errorf("failed to create kcp client: %w", err)
 	}
 
 	syncTarget, err := kcpClient.WorkloadV1alpha1().SyncTargets().Get(ctx,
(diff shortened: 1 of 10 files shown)


pkg/virtual/syncer/builder/build.go
```diff
@@ -82,23 +83,36 @@ func BuildVirtualWorkspace(
...
+			parts := strings.SplitN(withoutRootPathPrefix, "/", 4)
+			if len(parts) < 3 || parts[0] == "" || parts[1] == "" || parts[2] == "" {
 				return
 			}
+			workspace := parts[0]
 			workloadCusterName := parts[1]
+			syncTargetUID := parts[2]
 			apiDomainKey := dynamiccontext.APIDomainKey(clusters.ToClusterAwareKey(logicalcluster.New(parts[0]), workloadCusterName))
 
+			syncTarget, err := kcpClusterClient.Cluster(logicalcluster.New(workspace)).WorkloadV1alpha1().SyncTargets().Get(context.TODO(), workloadCusterName, metav1.GetOptions{})
```
sttts, line 98: must be informer based.

jmprusi: done.

jmprusi: logs are still too verbose for my taste.

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -147,18 +148,18 @@ func getSyncerID(syncTarget *workloadv1alpha1.SyncTarget) string {
 // enableSyncerForWorkspace creates a sync target with the given name and creates a service
 // account for the syncer in the given namespace. The expectation is that the provided config is
 // for a logical cluster (workspace). Returns the token the syncer will use to connect to kcp.
-func (c *Config) enableSyncerForWorkspace(ctx context.Context, config *rest.Config, syncTargetName, namespace string) (string, string, error) {
+func (c *Config) enableSyncerForWorkspace(ctx context.Context, config *rest.Config, syncTargetName, namespace string) (string, string, string, error) {
```
sttts, line 151: named return values woul dmake this much easier to read.

sttts:
https://github.com/jmprusi/kcp/pull/14

sttts:
Still feeling a little uncomfortable that the syncer namespace is still not depending on the UID. So if we really consider  SyncTargets with distinct UIDs different, then we should make the namespace UID dependent.

pkg/cliplugins/workload/plugin/sync.go
stevekuznetsov: the implicit behavior is very yuck though ;)

jmprusi:
> Still feeling a little uncomfortable that the syncer namespace is still not depending on the UID. So if we really consider SyncTargets with distinct UIDs different, then we should make the namespace UID dependent.

I think that right now the synctarget UID is already taken into account for the syncer namespace generation:

```
// getSyncerID returns a unique ID for a syncer derived from the name and its UID. It's
// a valid DNS segment and can be used as namespace or object names.
func getSyncerID(syncTarget *workloadv1alpha1.SyncTarget) string {
	syncerHash := sha256.Sum224([]byte(syncTarget.UID))
	base36hash := strings.ToLower(base36.EncodeBytes(syncerHash[:]))
	return fmt.Sprintf("kcp-syncer-%s-%s", syncTarget.Name, base36hash[:8])
}
```

pkg/virtual/syncer/builder/build.go
```diff
@@ -82,23 +83,38 @@ func BuildVirtualWorkspace(
...
 			realPath := "/"
-			if len(parts) > 2 {
-				realPath += parts[2]
+			if len(parts) > 3 {
+				realPath += parts[3]
 			}
 
-			//  /services/syncer/root:org:ws/<sync-target-name>/clusters/*/api/v1/configmaps
-			//                  ┌───────────────────────────────────┘
+			//  /services/syncer/root:org:ws/<sync-target-uid>/<sync-target-name>/clusters/*/api/v1/configmaps
```
sttts, line 116: the order surprises me. Didn't notice earlier. Can we switch it around? ws/name/uid.

sttts: ic, it's just a typo here in the comment. Can be follow-up to fix.

sttts:
With https://github.com/kcp-dev/kcp/pull/1687/files#r940441942

/lgtm

pkg/virtual/syncer/builder/build.go
jmprusi: ok

sttts:
Was wrong. It is not only the comment.

PR #1686, opened 2022-08-04
VW: Fix permission checking in `root`

davidfestal:
## Summary

We should only skip permission checking in the `root` workspace for the `list` verb, so that any user can always list the workspaces he has `access` to and can `get`.

## Related issue(s)

No issues

PR #1688, opened 2022-08-05
Add --batteries-included

sttts:
This adds a mechanism that allows to have certain defaults when running `kcp start`, but gives system integrators a way to disable the "integrated batteries". Canonical example: organization and team cluster workspace types. These were always just meant as examples, nothing set in stone for a service offering.

```
Start the control plane process

The server process listens on port 6443 and will act like a Kubernetes API server. It will initialize any necessary data to the provided start location or as a '.kcp' directory in the current directory. An admin kubeconfig file will be generated at initialization time that may be used to access the control plane.


Usage:
  kcp start [flags]

Most important flags:

      --batteries-included strings
                A list of batteries included (= default objects that might be unwanted in production, but are very helpful in trying out kcp or for development). These are the possible
                values: user.

                - cluster-workspace-types: creates "organization" and "team" ClusterWorkspaceTypes in the root workspace.
                - user:                  creates an additional non-admin user and context named "user" in the admin.kubeconfig

                Prefixing with - or + means to remove from the default set or add to the default set. (default [cluster-workspace-types])
      --etcd-servers strings
                List of etcd servers to connect with (scheme://ip:port), comma separated. By default an embedded etcd server is started. (default [embedded])
      --run-virtual-workspaces
                Run the virtual workspace apiservers in-process (default true)

Use "kcp start options" for a list of all flags available.
```

Based on https://github.com/kcp-dev/kcp/pull/1686.

### DIFF
--- a/pkg/server/options/options.go
+++ b/pkg/server/options/options.go
@@ -38,6 +38,7 @@ import (
 	etcdoptions "github.com/kcp-dev/kcp/pkg/embeddedetcd/options"
 	_ "github.com/kcp-dev/kcp/pkg/features"
 	kcpfeatures "github.com/kcp-dev/kcp/pkg/features"
+	"github.com/kcp-dev/kcp/pkg/server/options/batteries"
 )
 
 type Options struct {
@@ -63,6 +64,8 @@ type ExtraOptions struct {
 	ShardVirtualWorkspaceURL string
 	DiscoveryPollInterval    time.Duration
 	ExperimentalBindFreePort bool
+
+	BatteriesIncluded []string
 }
 
 type completedOptions struct {
@@ -103,6 +106,7 @@ func NewOptions(rootDir string) *Options {
 			ShardName:                "root",
 			DiscoveryPollInterval:    60 * time.Second,
 			ExperimentalBindFreePort: false,
+			BatteriesIncluded:        batteries.Defaults.List(),
 		},
 	}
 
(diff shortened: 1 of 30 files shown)


pkg/server/options/options.go
```diff
@@ -171,6 +175,16 @@ func (o *Options) rawFlags() cliflag.NamedFlagSets {
 	fs.BoolVar(&o.Extra.ExperimentalBindFreePort, "experimental-bind-free-port", o.Extra.ExperimentalBindFreePort, "Bind to a free port. --secure-port must be 0. Use the admin.kubeconfig to extract the chosen port.")
 	fs.MarkHidden("experimental-bind-free-port") // nolint:errcheck
 
+	fs.StringSliceVar(&o.Extra.BatteriesIncluded, "batteries-included", o.Extra.BatteriesIncluded, fmt.Sprintf(
+		`A list of batteries included (= default objects that might be unwanted in production, but are very helpful in trying out kcp or for development). These are the possible values: %s.
```
davidfestal, line 179: > These are the possible values: %s.

I find the message not very clear: the list just after contains the possible values that are not there by default.
Why don't we rather list the default values here, since the detailed list of all batteries (with description) is provided below ?

sttts: no, the list is a complete list. The output of the default comes from pflag, not from us.

davidfestal (approved):

Issue #1694, opened 2022-08-05
Use Structured Logging to Facilitate Better Observability


stevekuznetsov:
# Demo Objective
- [ ] Admin should be able to filter and query `kcp` logs to determine what actions were taken for a specific workspace, a specific object, or as a result of a specific event.
- [ ] End-to-end tests should filter & clarify the logs to provide a dev with only the pertinent logs on failure

# Demo Steps
1. Developer runs an end-to-end test that fails
1. Built-in code determines which logs are applicable and opens `lnav` to a focused view
1. Everyone is happy!

# Action Items

- [ ] Rebase onto 1.24
- [ ] Instrument our code with structured + contextual logging
- [ ] Add helpers to end-to-end tests

stevekuznetsov:
Logging level suggestions from @sttts

- stopping/starting controller - V(0)
- enqueue key - V(2)
- dequeue key - V(4)
- per-key comments
  - processing - V(1)
  - client calls - V(2)
  - something not done - V(4)
- errors - logger.Error()

stevekuznetsov:
For user-facing (terminal) logs, refer to @sttts experiments with [`zap`](https://github.com/sttts/kcp/commit/f2ee2d19084353819442bd78d60f823abff50493) and [`zerolog`](https://github.com/sttts/kcp/commit/1a4555e0403b3f8761f1faef0c276a8836a2c1a4)

stevekuznetsov:
FWIW, finding items that need to be updated:

```
$ git grep -P 'klog(\.V\([0-9]+\))?\.(Info|Warning|Error)(f|S)?'
...
cmd/sharded-test-server/frontproxy.go:  klog.Infof("Waiting for kcp-front-proxy to be up")
cmd/sharded-test-server/frontproxy.go:                  klog.Errorf("Failed to create kcp client: %v", err)
cmd/sharded-test-server/frontproxy.go:                  klog.V(3).Infof("kcp-front-proxy not ready: %v", err)
...
```

ncdc:
Ongoing work; moving to v0.11

PR #1695, opened 2022-08-05
reconciler/{apis,scheduling,tenancy}: use structured, contextual logging

stevekuznetsov:

Part of #1694

### DIFF
--- a/pkg/logging/constants.go
+++ b/pkg/logging/constants.go
@@ -18,10 +18,13 @@ limitations under the License.
 package logging
 
 import (
+	"fmt"
+
 	"github.com/go-logr/logr"
 	"github.com/kcp-dev/logicalcluster/v2"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/cache"
 )
 
 const (
@@ -64,4 +67,13 @@ func From(obj metav1.Object) []interface{} {
 		NameKey,
 		obj.GetName(),
 	}
+}
+
+// Key is like cache.MetaNamespaceKeyFunc, but with a restricted input set, so it can't error.
+func Key(obj metav1.Object) string {
+	key, err := cache.MetaNamespaceKeyFunc(obj)
+	if err != nil {
+		panic(fmt.Errorf("got an error from cache.MetaNamespaceKeyFunc: %w", err))
+	}
+	return key
 }
--- a/pkg/reconciler/apis/apibinding/apibinding_controller.go
+++ b/pkg/reconciler/apis/apibinding/apibinding_controller.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	jsonpatch "github.com/evanphx/json-patch"
+	"github.com/go-logr/logr"
 	"github.com/kcp-dev/logicalcluster/v2"
 
 	"k8s.io/apiextensions-apiserver/pkg/apihelpers"
@@ -48,6 +49,7 @@ import (
 	apisinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/apis/v1alpha1"
 	apislisters "github.com/kcp-dev/kcp/pkg/client/listers/apis/v1alpha1"
 	"github.com/kcp-dev/kcp/pkg/informer"
+	"github.com/kcp-dev/kcp/pkg/logging"
 )
 
 const (
@@ -129,10 +131,11 @@ func NewController(
 		deletedCRDTracker: newLockedStringSet(),
 	}
 
+	logger := logging.WithReconciler(klog.Background(), controllerName)
 	apiBindingInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    func(obj interface{}) { c.enqueueAPIBinding(obj, "") },
-		UpdateFunc: func(_, obj interface{}) { c.enqueueAPIBinding(obj, "") },
-		DeleteFunc: func(obj interface{}) { c.enqueueAPIBinding(obj, "") },
+		AddFunc:    func(obj interface{}) { c.enqueueAPIBinding(obj, logger, "") },
+		UpdateFunc: func(_, obj interface{}) { c.enqueueAPIBinding(obj, logger, "") },
+		DeleteFunc: func(obj interface{}) { c.enqueueAPIBinding(obj, logger, "") },
 	})
 
 	if err := apiBindingInformer.Informer().AddIndexers(cache.Indexers{
@@ -151,8 +154,8 @@ func NewController(
 			return logicalcluster.From(crd) == ShadowWorkspaceName
 		},
 		Handler: cache.ResourceEventHandlerFuncs{
-			AddFunc:    func(obj interface{}) { c.enqueueCRD(obj) },
-			UpdateFunc: func(_, obj interface{}) { c.enqueueCRD(obj) },
+			AddFunc:    func(obj interface{}) { c.enqueueCRD(obj, logger) },
+			UpdateFunc: func(_, obj interface{}) { c.enqueueCRD(obj, logger) },
 			DeleteFunc: func(obj interface{}) {
 				meta, err := meta.Accessor(obj)
 				if err != nil {
@@ -165,7 +168,7 @@ func NewController(
 				// the deleted CRD.
 				c.deletedCRDTracker.Add(meta.GetName())
 
-				c.enqueueCRD(obj)
+				c.enqueueCRD(obj, logger)
 			},
 		},
 	})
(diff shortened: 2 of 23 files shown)


pkg/reconciler/tenancy/clusterworkspacetype/clusterworkspacetype_controller.go
```diff
@@ -185,6 +187,10 @@ func (c *controller) processNextWorkItem(ctx context.Context) bool {
 	// other workers.
 	defer c.queue.Done(key)
 
+	logger := logging.WithQueueKey(klog.FromContext(ctx), key)
+	ctx = klog.NewContext(ctx, logger)
+	logger.Info("processing key")
```
ncdc, line 192: V(4)?

pkg/reconciler/tenancy/clusterworkspacetype/clusterworkspacetype_controller_reconcile.go
```diff
@@ -69,13 +68,7 @@ func (c *controller) updateVirtualWorkspaceURLs(cwt *tenancyv1alpha1.ClusterWork
...
 		if err != nil {
 			// Should never happen
-			klog.Errorf(
-				"Error parsing ClusterWorkspaceShard %s|%s spec.externalURL %q: %v",
-				logicalcluster.From(clusterWorkspaceShard),
-				clusterWorkspaceShard.Name,
-				clusterWorkspaceShard.Spec.ExternalURL,
-			)
-
+			logger.Error(err, "error parsing clusterWorkspaceShard.spec.externalURL")
```
ncdc, line 71: Not sure if we need to add back including the externalURL as a log value?

pkg/reconciler/tenancy/clusterworkspacetype/clusterworkspacetype_controller.go
stevekuznetsov: I remember Stefan suggesting something common here before I left for PTO... let me dig it up.

pkg/reconciler/tenancy/clusterworkspacetype/clusterworkspacetype_controller_reconcile.go
stevekuznetsov: I don't know that you need the value *and* the error, but sure - can't hurt

pkg/reconciler/tenancy/clusterworkspacetype/clusterworkspacetype_controller.go
stevekuznetsov: - stopping/starting controller - V(0)
- enqueue key - V(2)
- dequeue key - V(4)
- per-key comments
  - processing - V(1)
  - client calls - V(2)
  - something not done - V(4)
- errors - logger.Error()

stevekuznetsov:
Sigh, well in 4h we have not merged this yet and I accidentally pushed the rest of my changes to it ... @sttts @ncdc any issues with just ... reviewing here?

pkg/logging/constants.go
```diff
@@ -65,3 +68,12 @@ func From(obj metav1.Object) []interface{} {
 		obj.GetName(),
 	}
 }
+
+// Key is like cache.MetaNamespaceKeyFunc, but with a restricted input set, so it can't error.
```
stevekuznetsov, line 72: I'm open to doing something where you pass in a prefix and get e.g. `{"foo.workspace": "root", "foo.namespace": "ns", "foo.name": "something"}` as well, not entirely clear to me which is better.

ncdc:
Happy to review here. Just update the PR title?

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -119,8 +121,8 @@ func NewController(
 			return apiResourceSchema, err
 		},
 
-		createCRD: func(ctx context.Context, clusterName logicalcluster.Name, crd *apiextensionsv1.CustomResourceDefinition) (*apiextensionsv1.CustomResourceDefinition, error) {
```
stevekuznetsov, line 122: @ncdc unclear how this has evolved but it seemed weird to pass in an object that *has* a cluster, then ignore it and use a different value instead

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -129,10 +131,11 @@ func NewController(
 		deletedCRDTracker: newLockedStringSet(),
 	}
 
+	logger := logging.WithReconciler(klog.Background(), controllerName)
 	apiBindingInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    func(obj interface{}) { c.enqueueAPIBinding(obj, "") },
-		UpdateFunc: func(_, obj interface{}) { c.enqueueAPIBinding(obj, "") },
-		DeleteFunc: func(obj interface{}) { c.enqueueAPIBinding(obj, "") },
+		AddFunc:    func(obj interface{}) { c.enqueueAPIBinding(obj, logger, "") },
```
stevekuznetsov, line 136: @ncdc also open to ideas here, this seemed like the most natural extension of what you did, but unfortunately since the message was also appended to, it was not opaque to the caller

pkg/reconciler/tenancy/clusterworkspacedeletion/deletion/workspace_resource_deletor.go
```diff
@@ -95,13 +95,14 @@ type workspacedResourcesDeleter struct {
...
 // Caller is expected to keep calling this until it succeeds.
 func (d *workspacedResourcesDeleter) Delete(ctx context.Context, workspace *tenancyv1alpha1.ClusterWorkspace) error {
+	logger := klog.FromContext(ctx)
 	var err error
 	// if the workspace was deleted already, don't do anything
 	if workspace.DeletionTimestamp == nil {
 		return nil
 	}
 
-	klog.V(5).Infof("workspace deletion controller - syncworkspace - workspace: %s, finalizer: %s", workspace.Name, WorkspaceFinalizer)
```
stevekuznetsov, line 104: This file was just a textbook case for the use-case of structured logging :)

stevekuznetsov:
/retitle reconciler/{apis,scheduling,tenancy}: use structured, contextual logging

pkg/reconciler/tenancy/clusterworkspaceshard/clusterworkspaceshard_controller.go
```diff
@@ -121,7 +121,7 @@ func (c *Controller) processNextWorkItem(ctx context.Context) bool {
 
 	logger := logging.WithQueueKey(klog.FromContext(ctx), key)
 	ctx = klog.NewContext(ctx, logger)
-	logger.Info("processing key")
+	logger.V(1).Info("processing key")
```
sttts, line 124: V(1) or V(4), seeing both in different controllers.

Do we have written down our guideline for the levels? We discussed in Slack.

pkg/logging/constants.go
sttts: would prefer `From` and `Key` to give the same keys, wdyt?

pkg/reconciler/apis/apibinding/apibinding_controller.go
sttts: why isn't logger part of the controller struct? Passing it around seems to be an anti-pattern.

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -267,20 +270,25 @@ func (c *controller) enqueueAPIExport(obj interface{}, logSuffix string) {
...
-func (c *controller) enqueueCRD(obj interface{}) {
+func (c *controller) enqueueCRD(obj interface{}, logger logr.Logger) {
 	crd, ok := obj.(*apiextensionsv1.CustomResourceDefinition)
 	if !ok {
 		runtime.HandleError(fmt.Errorf("obj is supposed to be a CustomResourceDefinition, but is %T", obj))
 		return
 	}
+	logger = logger.WithValues(
+		"CustomResource", fmt.Sprintf("%s.%s", crd.Spec.Names.Plural, crd.Spec.Group),
+		"CustomResourceDefinition", logging.Key(crd),
```
sttts, line 286: as written above, am not convinced having keys like that has value, or at least log the key **and** the individual ws+ns+name values in some way. Maybe also the kind and group?

pkg/reconciler/tenancy/clusterworkspaceshard/clusterworkspaceshard_controller.go
stevekuznetsov: I wrote your comments into the epic - you said 1

pkg/reconciler/apis/apibinding/apibinding_controller.go
stevekuznetsov: https://github.com/kcp-dev/kcp/pull/1695#discussion_r939332761

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -159,15 +162,12 @@ func (c *controller) reconcileBinding(ctx context.Context, apiBinding *apisv1alp
 	for _, schemaName := range apiExport.Spec.LatestResourceSchemas {
 		schema, err := c.getAPIResourceSchema(apiExportClusterName, schemaName)
 		bindingClusterName := logicalcluster.From(apiBinding)
-		exportClustername := logicalcluster.From(apiExport)
+		logger = logger.WithValues(
+			"APIExport", logging.Key(apiExport),
+			"APIResourceSchema", logging.Key(schema),
```
sttts, line 167: this panics

pkg/reconciler/apis/apibinding/apibinding_controller.go
sttts: please don't use this pattern. **Always** pass the cluster name and always ignore the value in the object. This kind of code here become unreadable in non-trivial cases.

pkg/reconciler/apis/apibinding/apibinding_controller.go
sttts: Nvmd. This reminds me of the stupid explicit contexts in Golang. No context around 🤷‍♂️

pkg/logging/constants.go
sttts: klog and go-logr onl support strings? `foo: {workspace: root, ns: bla name: blub}` would really be the natural representation.

pkg/reconciler/tenancy/clusterworkspaceshard/clusterworkspaceshard_controller.go
sttts: 1 is fine, just noticed that we have multiple variants :)

pkg/logging/constants.go
sttts: there is another issue with either of these approaches: I would argue the main usecase is to filter logs by "show me everything that involves  APIExport `foo|bar`". Can a structured query e.g. in Grafana/Loki filter like that? With the key this maybe the easiest, because a non-structured query just by text could find it.

stevekuznetsov: This is for ancillary objects, not the particular object being queued by the controller. Yes, generally query languages allow for some loose string matching on keys.

pkg/reconciler/tenancy/clusterworkspaceshard/clusterworkspaceshard_controller.go
stevekuznetsov: Will need a second pass to catch it ... or some linter rule :)

pkg/reconciler/apis/apibinding/apibinding_controller.go
stevekuznetsov: Why? It seems entirely non-obvious and confusing to explicitly set a cluster in the object, then ignore it

sttts:
Let's ship it, play with Loki/Grafana, learn and adapt.

/lgtm
/approve

PR #1696, opened 2022-08-05
apiexport: report 0 VW URLs correctly

ncdc:
## Summary
If there are no shards with their virtual workspace URL set, do not
report that an APIExport's virtual workspace URLs are ready.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/reconciler/apis/apiexport/apiexport_reconcile.go
+++ b/pkg/reconciler/apis/apiexport/apiexport_reconcile.go
@@ -203,7 +203,17 @@ func (c *controller) updateVirtualWorkspaceURLs(apiExport *apisv1alpha1.APIExpor
 		})
 	}
 
-	conditions.MarkTrue(apiExport, apisv1alpha1.APIExportVirtualWorkspaceURLsReady)
+	if len(apiExport.Status.VirtualWorkspaces) == 0 {
+		conditions.MarkFalse(
+			apiExport,
+			apisv1alpha1.APIExportVirtualWorkspaceURLsReady,
+			apisv1alpha1.NoViableShardsReason,
+			conditionsv1alpha1.ConditionSeverityError,
+			"Unable to find any shards with virtual workspace URLs",
+		)
+	} else {
+		conditions.MarkTrue(apiExport, apisv1alpha1.APIExportVirtualWorkspaceURLsReady)
+	}
 
 	return nil
 }
--- a/pkg/apis/apis/v1alpha1/types_apiexport.go
+++ b/pkg/apis/apis/v1alpha1/types_apiexport.go
@@ -35,6 +35,7 @@ const (
 	APIExportVirtualWorkspaceURLsReady conditionsv1alpha1.ConditionType = "VirtualWorkspaceURLsReady"
 
 	ErrorGeneratingURLsReason = "ErrorGeneratingURLs"
+	NoViableShardsReason      = "NoViableShards"
 )
 
 // These are for APIExport identity.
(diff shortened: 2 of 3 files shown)


pkg/reconciler/apis/apiexport/apiexport_reconcile.go
```diff
@@ -203,7 +203,17 @@ func (c *controller) updateVirtualWorkspaceURLs(apiExport *apisv1alpha1.APIExpor
...
 	}
 
-	conditions.MarkTrue(apiExport, apisv1alpha1.APIExportVirtualWorkspaceURLsReady)
+	if len(apiExport.Status.VirtualWorkspaces) == 0 {
+		conditions.MarkFalse(
+			apiExport,
+			apisv1alpha1.APIExportVirtualWorkspaceURLsReady,
+			apisv1alpha1.NoViableShardsReason,
+			conditionsv1alpha1.ConditionSeverityError,
+			"Unable to find any shards with virtual workspace URLs",
```
sttts, line 212: how can that happen? If it is just a moment, not worth all this condition logic.

sttts:
This feels like a workaround instead of a root cause fix. Why does this happen? Do we miss defaulting?

pkg/reconciler/apis/apiexport/apiexport_reconcile.go
ncdc: I honestly have no idea. It happened to me locally. I had my server up a running for a while. I eventually created an APIExport. It had the VW URLs ready condition=true, but no URL in status. I don't know the root cause.

sttts: cc @p0lyn0mial 

If it is not critical, would let Lukasz take a look at this next week.

sttts: Am curious, was that old etcd state or fresh?

ncdc: I _think_ I had a preexisting set of data

ncdc: As soon as I restarted kcp (without any code changes), the APIExport got a VW URL

p0lyn0mial: I have just tried with a preexisting set and a `virtualWorkspaceURL` has been assigned.

A `virtualWorkspaceURL` is assigned during bootstrapping of the root workspace. If it hasn't been set (via a flag) it will be defaulted to the `baseURL` by an admission plugin.

Since the root `ClusterWorkspaceShard` doesn't have `annotationCreateOnlyKey` set it will always be `Updated` during bootstrapping (preexisting set). Also, any failure is eventually retired. Thoughts?

p0lyn0mial: I checked the code, nothing stands out.

Is there an easy way of searching CI for a failed test?

We could see how common the issue is with https://github.com/kcp-dev/kcp/pull/1699

sttts:
/hold

until we understand the root cause, or it becomes urgent that we have to stop the bleeding.

pkg/reconciler/apis/apiexport/apiexport_reconcile.go
ncdc: I've only seen this once, on my laptop

p0lyn0mial: would it make sense to merge https://github.com/kcp-dev/kcp/pull/1699 ?

PR #1700, opened 2022-08-08
server: hardens the remote shard registration

p0lyn0mial:
## Summary
changes the way we register the remote shard so that it handles the following cases:

- the resource has already been registered and is up to date
- the resource has been registered but requires an update
- the resource hasn't been added

all the above cases tolerate any availability issues while connecting to the root shard.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/kcp-dev/logicalcluster/v2"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -235,7 +236,6 @@ func (s *Server) Run(ctx context.Context) error {
 			}
 			klog.Infof("Finished starting kcp informers for the root shard")
 
-			klog.Info("Creating ClusterWorkspaceShard resource in the root shard")
 			shard := &tenancyv1alpha1.ClusterWorkspaceShard{
 				ObjectMeta: metav1.ObjectMeta{Name: s.Options.Extra.ShardName},
 				Spec: tenancyv1alpha1.ClusterWorkspaceShardSpec{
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -244,10 +244,53 @@ func (s *Server) Run(ctx context.Context) error {
 					VirtualWorkspaceURL: s.Options.Extra.ShardVirtualWorkspaceURL,
 				},
 			}
-			if _, err := s.RootShardKcpClusterClient.Cluster(tenancyv1alpha1.RootCluster).TenancyV1alpha1().ClusterWorkspaceShards().Create(goContext(ctx), shard, metav1.CreateOptions{}); err != nil {
-				return err
+
+			if err := wait.PollInfiniteWithContext(goContext(ctx), time.Second, func(ctx context.Context) (bool, error) {
```
stevekuznetsov, line 248: Should we use an exponential backoff here?

stevekuznetsov (commented):
Is there any easy way to test this logic?

pkg/server/server.go
```diff
@@ -244,10 +244,53 @@ func (s *Server) Run(ctx context.Context) error {
...
+					_, err := s.RootShardKcpClusterClient.Cluster(tenancyv1alpha1.RootCluster).TenancyV1alpha1().ClusterWorkspaceShards().Update(ctx, existingShard, metav1.UpdateOptions{})
+					if err != nil {
+						klog.Error("failed updating %q ClusterWorkspaceShard in the root shard, err %v", shard.Name, err)
+						return false, nil
+					}
+					klog.Infof("Updated %q ClusterWorkspaceShard resource in the root shard", shard.Name)
+					return true, nil
+				}
+				klog.Infof("ClusterWorkspaceShard %q resource is up to date in the root shard", shard.Name)
+				return true, nil
```
sttts, line 288: all this worth a method?

pkg/server/server.go
```diff
@@ -244,10 +244,53 @@ func (s *Server) Run(ctx context.Context) error {
 					VirtualWorkspaceURL: s.Options.Extra.ShardVirtualWorkspaceURL,
 				},
 			}
-			if _, err := s.RootShardKcpClusterClient.Cluster(tenancyv1alpha1.RootCluster).TenancyV1alpha1().ClusterWorkspaceShards().Create(goContext(ctx), shard, metav1.CreateOptions{}); err != nil {
-				return err
+
+			if err := wait.PollInfiniteWithContext(goContext(ctx), time.Second, func(ctx context.Context) (bool, error) {
+				klog.Infof("Getting %q ClusterWorkspaceShard from the root shard", shard.Name)
```
sttts, line 249:
```suggestion
				klog.Infof("Getting %q ClusterWorkspaceShard from the root workspace", shard.Name)
```

pkg/server/server.go
```diff
@@ -244,10 +244,53 @@ func (s *Server) Run(ctx context.Context) error {
 					VirtualWorkspaceURL: s.Options.Extra.ShardVirtualWorkspaceURL,
```
sttts, line 244: all these are defaulted already?

pkg/server/server.go
```diff
@@ -244,10 +244,53 @@ func (s *Server) Run(ctx context.Context) error {
...
+				}
+				if len(shard.Spec.ExternalURL) > 0 && existingShard.Spec.ExternalURL != shard.Spec.ExternalURL {
+					existingShard.Spec.ExternalURL = shard.Spec.ExternalURL
+					specChanged = true
+				}
+				if len(shard.Spec.VirtualWorkspaceURL) > 0 && existingShard.Spec.VirtualWorkspaceURL != shard.Spec.VirtualWorkspaceURL {
+					existingShard.Spec.VirtualWorkspaceURL = shard.Spec.VirtualWorkspaceURL
+					specChanged = true
+				}
+				if specChanged {
```
sttts, line 278: what if we skipped all this checking and just update spec unconditionally? Would be good enough, wouldn't it?

pkg/server/server.go
p0lyn0mial: I don't think so. We do it on a server startup, so we want it to be quick. Backing off might delay server initialization.

p0lyn0mial:
> Is there any easy way to test this logic?

yes, you could `./bin/sharded-test-server --number-of-shards=2` on your local machine

pkg/server/server.go
p0lyn0mial: ah, I have just realized that the fields would be updated by the admission plugin. So yeah it seems to be safe.

pkg/server/server.go
p0lyn0mial: I'll move it to a separate hook in the next step. So I'd leave it as it is.

stevekuznetsov:
https://github.com/kcp-dev/kcp/pull/1700#issuecomment-1209062432

I meant a programmatic test ... ;)

PR #1704, opened 2022-08-08
Makefile: set UNSAFE_E2E_HACK_DISABLE_ETCD_FSYNC=true in all e2e tests

sttts:
Let's see how much (if at all) faster this is.


stevekuznetsov:
/lgtm
/approve
/hold

@sttts cancel hold when you're happy with what CI did :)

PR #1685, opened 2022-08-04
Completely cleanup support of personal workspaces in top-level organizations

davidfestal:
## Summary

As an outcome of the [User Home Workspaces](https://github.com/kcp-dev/kcp/issues/1388#1069) work, and a follow-up of PR https://github.com/kcp-dev/kcp/pull/1374, we should completely remove the support of personal workspaces in top-level organizations.

## Related issue(s)

Fixes #1388

PR #1706, opened 2022-08-08
Remove the redirect from shard to virtual workspaces

davidfestal:
## Summary

Remove the redirect from shard to virtual workspaces in case of standalone virtual workspace deployment.
Npw that the old "personal workspaces" feature has been removed from both the `workspaces` virtual workspace and the CLI, there's no need to point to virtual workspaces through the KCP server host.
In the future explicit virtual workspace Base URL (carried by the Shard resource) should be used.

So when a virtual workspace is started in standalone mode, there's no need anymore to setup a redirection from the KCP shard host to the virtual workspace host.

## Related issue(s)

This is a followup of PR https://github.com/kcp-dev/kcp/pull/1685

### DIFF
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net/http"
 	_ "net/http/pprof"
+	"net/url"
 
 	"github.com/kcp-dev/logicalcluster/v2"
 
@@ -263,6 +264,14 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
 		return nil, err
 	}
 
+	var shardVirtualWorkspaceURL *url.URL
+	if !opts.Virtual.Enabled && opts.Extra.ShardVirtualWorkspaceURL != "" {
+		shardVirtualWorkspaceURL, err = url.Parse(opts.Extra.ShardVirtualWorkspaceURL)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	// preHandlerChainMux is called before the actual handler chain. Note that BuildHandlerChainFunc below
 	// is called multiple times, but only one of the handler chain will actually be used. Hence, we wrap it
 	// to give handlers below one mux.Handle func to call.
@@ -301,7 +310,7 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
 		*c.preHandlerChainMux = append(*c.preHandlerChainMux, mux)
 		apiHandler = mux
 
-		apiHandler = WithWorkspaceProjection(apiHandler)
+		apiHandler = WithWorkspaceProjection(apiHandler, shardVirtualWorkspaceURL)
 		apiHandler = WithClusterAnnotation(apiHandler)
 		apiHandler = WithAuditAnnotation(apiHandler) // Must run before any audit annotation is made
 		apiHandler = WithClusterScope(apiHandler)
--- a/pkg/server/handler.go
+++ b/pkg/server/handler.go
@@ -201,7 +201,7 @@ func WithClusterAnnotation(handler http.Handler) http.HandlerFunc {
 
 // WithWorkspaceProjection maps the personal virtual workspace "workspaces" resource into the cluster
 // workspace URL space. This means you can do `kubectl get workspaces` from an org workspace.
-func WithWorkspaceProjection(apiHandler http.Handler) http.HandlerFunc {
+func WithWorkspaceProjection(apiHandler http.Handler, shardVirtualWorkspaceURL *url.URL) http.HandlerFunc {
 	toRedirectPath := path.Join("/apis", tenancyv1beta1.SchemeGroupVersion.Group, tenancyv1beta1.SchemeGroupVersion.Version, "workspaces/")
 	getHomeWorkspaceRequestPath := path.Join(toRedirectPath, "~")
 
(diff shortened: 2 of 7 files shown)


pkg/server/handler.go
```diff
@@ -225,6 +226,9 @@ func WithWorkspaceProjection(apiHandler http.Handler) http.HandlerFunc {
 			newPath := path.Join("/services/workspaces", cluster.Name.String(), req.URL.Path)
 			klog.V(4).Infof("Rewriting %s -> %s", path.Join(cluster.Name.Path(), req.URL.Path), newPath)
 			req.URL.Path = newPath
+			if !opts.Virtual.Enabled && opts.Extra.ShardVirtualWorkspaceURL != "" {
+				req.URL.Host = opts.Extra.ShardVirtualWorkspaceURL
```
p0lyn0mial, line 230: I think we should read the shard resource from the DB (via a lister), doing that ensures we will read default values.

pkg/server/handler.go
```diff
@@ -225,6 +226,9 @@ func WithWorkspaceProjection(apiHandler http.Handler) http.HandlerFunc {
 			newPath := path.Join("/services/workspaces", cluster.Name.String(), req.URL.Path)
 			klog.V(4).Infof("Rewriting %s -> %s", path.Join(cluster.Name.Path(), req.URL.Path), newPath)
 			req.URL.Path = newPath
+			if !opts.Virtual.Enabled && opts.Extra.ShardVirtualWorkspaceURL != "" {
```
p0lyn0mial, line 229: since we removed the `installVirtualWorkspacesRedirect` method, should we rewrite here?

pkg/server/handler.go
davidfestal: OK.

davidfestal: After thinking though it again, I'm not sure it's necessary:
There are 2 cases:
- when admission will set it to the shard base URL, that means that we're in the case where it has not been specified through the CLI, probably in the case when virtual workspaces are embedded in the KCP shard. Then I don't need a URL at all since the project should just change the URL PATH (without a redirect)
- when this value is set through the option, it will be used used anyway (not overridden by the admission plugin) => let's use it directly.

pkg/server/handler.go
davidfestal: Done in rebase code.

pkg/server/config.go
```diff
@@ -263,6 +264,14 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
 		return nil, err
 	}
 
+	var shardVirtualWorkspaceURL *url.URL
+	if !opts.Virtual.Enabled && opts.Extra.ShardVirtualWorkspaceURL != "" {
```
p0lyn0mial, line 268: should we return an error if the virtual is enabled and the URL was provided?

pkg/server/handler.go
```diff
@@ -225,6 +225,15 @@ func WithWorkspaceProjection(apiHandler http.Handler) http.HandlerFunc {
 			newPath := path.Join("/services/workspaces", cluster.Name.String(), req.URL.Path)
 			klog.V(4).Infof("Rewriting %s -> %s", path.Join(cluster.Name.Path(), req.URL.Path), newPath)
 			req.URL.Path = newPath
+			if shardVirtualWorkspaceURL != nil {
```
p0lyn0mial, line 228: okay, if the URL wasn't provided we assume the virtual server is in the same process.

test/e2e/virtual/workspaces/virtual_workspace_test.go
```diff
@@ -622,6 +636,7 @@ func testWorkspacesVirtualWorkspaces(t *testing.T, standalone bool) {
 		}, wait.ForeverTestTimeout, time.Millisecond*100, "virtual workspace apiserver not ready")
 	} else {
 		server = framework.SharedKcpServer(t)
+		virtualWorkspaceServerHost = server.BaseConfig(t).Host
```
p0lyn0mial, line 639: only the host part?

p0lyn0mial: unless it already contains the port

pkg/server/config.go
davidfestal: Probably. Not sure it belongs to this PR though. wdyt ?

pkg/server/handler.go
davidfestal: yes

test/e2e/virtual/workspaces/virtual_workspace_test.go
davidfestal: Yes, afaict:

```
	// Host must be a host string, a host:port pair, or a URL to the base of the apiserver.
	// If a URL is given then the (optional) Path of that URL represents a prefix that must
	// be appended to all request URIs used to access the apiserver. This allows a frontend
	// proxy to easily relocate all of the apiserver endpoints.
	Host string
```

davidfestal:
@p0lyn0mial Are my answers OK for you ?

test/e2e/virtual/workspaces/virtual_workspace_test.go
```diff
@@ -183,6 +185,8 @@ var testCases = []struct {
 		work: func(ctx context.Context, t *testing.T, server runningServer) {
 			testData := newTestData()
 
+			user1Client := server.UserKcpClients[0]
```
p0lyn0mial, line 188: does `user1Client` exercise redirection code when the vw is run in a standalone mode?

pkg/server/config.go
p0lyn0mial: Sure, can be done in a separate PR, just saying that this might be a configuration error.

p0lyn0mial:
for approval

/assign @sttts

PR #1708, opened 2022-08-09
✨ Support for local cluster services DNS resolution

### LINKED ISSUE #505: Cross-namespace service DNS resolution in physical clusters
If a pod in namespace one references a service in another namespace, such as `some-service.two.svc.cluster.local`, how will that work in physical clusters where the namespace names are mapped and no longer retain their original names?

### LINKED ISSUE #1465: Same namespace fully qualified service DNS resolution in pcluster
Pods cannot connect to services when using fully qualified service names (e.g. `mysvc.mynamespace.svc.cluster.local`) because kcp does not preserve the original namespace (which is expected). 

This issue is related to https://github.com/kcp-dev/kcp/issues/505. The same problem exists for pods within the same namespace. Application developers can change their code to use non-qualified service names (e.g. `mysvc`) which is not ideal and might not be [optimal](https://pracucci.com/kubernetes-dns-resolution-ndots-options-and-why-it-may-affect-application-performances.html).

Pods connecting to services using fully qualified service names (e.g. `mysvc.mynamespace.svc.cluster.local`) should succeed without any changes to the application code.

None.

**Additional context**

/cc @ncdc

lionelvillard:
## Summary

Configure synced deployments DNS Config to point to kcp DNS resolver mapping local namespaces to physical namespaces for cluster local services.
 
## Related issue(s)

Fixes #505 
Fixes #1465


jmprusi:
@lionelvillard, thanks for the PR! 

A couple of comments:

I see that the PR doesn't take into account the workspace of the Pod that is making the request, so that could lead to conflicts in the namespace resolution, given the following scenario:

1. Pod A running in Namespace A belongs to workspace A
1. Service B in Namespace B belongs to workspace B
1. Service B in Namespace B belongs to workspace A

if POD A tries to resolve the service B in namespace B. Currently, the plugin can respond with a service from another Workspace (B)

To solve this, the plugin has to be aware of the Workspace of the querying POD and filter accordingly.

Also, wiring up the Syncer PODs IP to the dnsConfig means that if the syncer pod gets recreated (update, scale...), all the currently deployed pods/deployments will need to be resynced. Perhaps a service that fronts the syncer deployment with a cluster IP could be a better solution?

Thank you!

lionelvillard:
> Also, wiring up the Syncer PODs IP to the dnsConfig means that if the syncer pod gets recreated (update, scale...), all the currently deployed pods/deployments will need to be resynced. Perhaps a service that fronts the syncer deployment with a cluster IP could be a better solution?

Yes that's a great idea, thanks! 

On that topic I feel a bit uneasy about packaging both the kcp dns proxy and the syncer together due to different scheduling requirements (the dns proxy is critical, the syncer is not). What's your take on this?

jmprusi:
I agree. Coupling the DNS resolution to the syncer doesn't feel right... we were thinking of something along those lines:

* Deploying one CoreDNS pod for each workspace in a pcluster, with an updated map of the namespaces for the given workspace and pointing all the pods in all the workspace namespaces to their correct CoreDNS server. 

What is essential is to avoid extra watches/informers from being used and try to make the CoreDNS watch/hot-reload a configuration (configmap?) that is updated by the syncers on each namespace creation/deletion.

lionelvillard:
> Deploying one CoreDNS pod for each workspace in a pcluster, with an updated map of the namespaces for the given workspace and pointing all the pods in all the workspace namespaces to their correct CoreDNS server.

I would much prefer this solution. It provides better workspaces isolation and can be easily enforced by using network policies. It also solves the source pod issue (we certainly don't want pod watch!) 

> What is essential is to avoid extra watches/informers from being used and try to make the CoreDNS watch/hot-reload a configuration (configmap?) that is updated by the syncers on each namespace creation/deletion.

There are various ways to do this. Using a configmap is simple but require a watch. Another way is to expose a management API (add/remove namespaces) which is quite easy to implement with Caddy.

jmprusi:
> There are various ways to do this. Using a configmap is simple but require a watch. Another way is to expose a management API (add/remove namespaces) which is quite easy to implement with Cadd

Sorry, I meant avoiding extra watches against KCP, local watches should be fine! :)

lionelvillard:
@jmprusi I implemented what we discussed above: 
- the kcp dns resolver now runs in its own pod and gets its configuration from the kcp-dns-nsmap configmap. 
- the syncer makes sure the kcp-dns-nsmap configmap is up-to-date
- the kubectl kcp plugin generates additional resources to deploy the kcp dns deployment. 

PTAL.

lionelvillard:
Is it possible to get a /ok-to-test so I can check the e2e tests have been fixed? I will probably need several iterations. (it's working locally but I only run one configuration).

lionelvillard:
@jmprusi Do you have time to look at this PR?

jmprusi (changes requested):
Thanks! Worked perfectly!   ~~tried it locally and ran into some issues... I will add more reviews comments once I can resolve the local issue.~~

Not fully sure about 2 different images instead of using just one (syncer) launched with different flags, as that requires users to pass two different images...but I guess we can set a proper default image to avoid that?

lionelvillard:
> Thanks! Worked perfectly! ~tried it locally and ran into some issues... I will add more reviews comments once I can resolve the local issue.~
> 

Great!


> Not fully sure about 2 different images instead of using just one (syncer) launched with different flags, as that requires users to pass two different images...but I guess we can set a proper default image to avoid that?

Having 2 different images is IMO cleaner (greater separation of concerns) and I agree the devex can be improved by setting a proper default image (when we get one).

> 
> Also, can we add some debug logging to the DNS plugin? (or document how to enable it)

Will do!

lionelvillard:
@jmprusi I've updated the branch. PTAL! thx!

lionelvillard:
@jmprusi is there anything left to be done for this PR, besides getting a ghcr.io/kcp-dev/kcp url?

jmprusi:
I've run some tests locally and functionality LGTM, can you review @davidfestal and/or @sttts ?

lionelvillard:
I just refactor a bit the CoreDNS plugin to closer match the CoreDNS [rewrite plugin](https://coredns.io/plugins/rewrite/).

docs/syncer.md
```diff
@@ -23,7 +23,7 @@ Current workspace is "root:my-org" (type "root:organization").
 1. Enable the syncer for a new cluster
 
 ```sh
-$ kubectl kcp workload sync <mycluster> --syncer-image <image name> -o syncer.yaml
+$ kubectl kcp workload sync <mycluster> --syncer-image <image name>  --dns-image <image name> -o syncer.yaml
```
sttts, line 26: why is this a different image? Can we merge them?

test/e2e/syncer/syncer_test.go
```diff
@@ -438,6 +449,8 @@ func TestSyncWorkload(t *testing.T) {
 		syncTargetName,
 		"--syncer-image",
 		"ghcr.io/kcp-dev/kcp/syncer-c2e3073d5026a8f7f2c47a50c16bdbec:41ca72b",
+		"--dns-image",
+		"villardl/coredns-6fd63c29c5e6edb631231bbd0936a177:sha256-697d31e773cf457c920f97f5b820ce021db0f39f1c999cc451cb11bff9d55a19", // TODO(LV): temp location
```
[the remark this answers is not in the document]
sttts: better: merge the two images. Having two is not good. Unneeded complexity. If it is hard to have two commands in one image, add a sub-command to the syncer which starts coredns instead.

docs/syncer.md
lionelvillard: It is a different image because CoreDNS needs root access (see https://github.com/coredns/coredns/issues/1767#issuecomment-507213651) whereas the syncer does not. 

Instead of `--syncer-image` and `--dns-image` we could replace them by `--container-repository` and `--container-tag` and use fixed container image name.

test/e2e/syncer/syncer_test.go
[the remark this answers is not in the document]
lionelvillard: if security is not a strong concern I can merge the two images.

docs/syncer.md
sttts: why does root access have anything to do with the image?

lionelvillard: To run non-root containers, the image needs to be built with a non-root user and the k8s deployment needs to be configured to use the uid created in the image (see `securityContext.runAsUser`). Adding a non-root user in the image is not strictly required and could work if the syncer does not access the user directory.

lionelvillard: There might be another solution based on container port mapping.

lionelvillard: container port mapping works, problem solved. I'll use one image.

davidfestal (commented):
Currently, a Syncer may sync resources form various user KCP workspaces.
More precisely from all workspaces which have bound to the compute APIExports related to the Syncer Synctarget.
In the future this situation would be the most usual one. Am I mistaken or is it right that this use-case is not supported ?
If I understand correctly this PR assumes that the only workspace resources synced by the syncer exist in the same KCP workspace that contains the SyncTarget.

Afaict, after discussing it with @jmprusi, supporting this mid-term scenario would mean either:
- installing a dedicated CoreDNS POD in a dedicated namespace for each couple SyncTarget / user KCP workspace. That would mean managing the creation of the CoreDNS POD from the Syncer itself... and a number of questions are raised by this....
- switching back to running the CoreDNS process as a sidecar in every user deployment POD (along with having the related Config map mounted as a volume on it), which possibly we don't want as well.

The second question I have is about adding NetworkPolicies in order to limit the network allowed communications to the CoreDNS POD(s) at their minimal. But I assume this one depends on what direction we would take on the first question.

Do you think we should open one (or 2) new issue(s) or an EPIC for these concerns and just merge this PR ?

cc @sttts

lionelvillard:
> If I understand correctly this PR assumes that the only workspace resources synced by the syncer exist in the same KCP workspace that contains the SyncTarget.

Currently there is one DNS server per syncer. So when the Syncer is connected to multiple user workspaces, they all connect to the same DNS server. I haven't tested this use case yet so I'm most likely missing something.

Moving forward, I would like to merge this PR as, at least, it is useful for the simple one-to-one use case. If you could open follow up issues I'd be happy to take a look at them.

(side note: I'm working on fixing Stefan's concerns, almost done...)

sttts:
> Moving forward, I would like to merge this PR as, at least, it is useful for the simple one-to-one use case

IMO @davidfestal 's concerns are fundamental. The described use-case by David is the normal use-case, not the other way around. We have to have a plan for that before merging. If we have that plan and decide to merge first, so be it. But this conceptual work IMO has to happen first.

If I remember right, Ben Bennet had discussed this in some doc of how to serve the right DNS records for each workspace, e.g. by source IP, or by having a proxy CoreDNS per workspace. @jmprusi can probably provide the link to that discussion.

davidfestal:
> > Moving forward, I would like to merge this PR as, at least, it is useful for the simple one-to-one use case
> 
> IMO @davidfestal 's concerns are fundamental. The described use-case by David is the normal use-case, not the other way around. We have to have a plan for that before merging. If we have that plan and decide to merge first, so be it. But this conceptual work IMO has to happen first.
> 
So it is not excluded to merge this PR, even though the added feature would not be supported in the use-case of Location Workspaces.
But at least merging this **should not break** the ability of the Syncer to sync workload resources from several user workspaces, and have them up and running as it is today.


> If I remember right, Ben Bennet had discussed this in some doc of how to serve the right DNS records for each workspace, e.g. by source IP, or by having a proxy CoreDNS per workspace. @jmprusi can probably provide the link to that discussion.

@sttts @lionelvillard We should plan a design session to revisit the various options that were initially explorerd by @jmprusi and @knobunc

jmprusi:
The document where we explored/discussed some options is: https://docs.google.com/document/d/1tzyYWMFlhgUL4IOTI6gjfBqTobsyXMjgFeqD7Uijn5g

> @sttts @lionelvillard We should plan a design session to revisit the various options that were initially explorerd by @jmprusi and @knobunc

Yes, please, let's arrange a design session.

davidfestal:
So after our design session here is what I propose to do:
- Before merging this PR we should add a check so that the `DeploymentMutator` would replace the `DNSPolicy` and `DNSConfig` only when the deployment comes from the same upstream workspace as where the SyncTarget workspace.
By doing so we gate the scope of this PR to the only supported use-case, avoid breaking the behavior in the Location Workspace use-case.
- After the merge of this PR we should:
  - change the creation of the CoreDNS Deployment and associated resources from a static creation (in the `workload sync` CLI command) to have the syncer itself create it dynamically, one dpeloyment per user-workspace synced to this Syncer.
  - cleanup the CoreDNS Deployment when no more namespaces are related the corresponding user workspace.
This would be detailed in a dedicated issue.

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -47,11 +48,12 @@ func (dm *DeploymentMutator) GVR() schema.GroupVersionResource {
 	}
 }
 
-func NewDeploymentMutator(upstreamURL *url.URL, secretLister ListSecretFunc, dnsIP string) *DeploymentMutator {
+func NewDeploymentMutator(upstreamURL *url.URL, secretLister ListSecretFunc, downstreamLogicalName logicalcluster.Name, dnsIP string) *DeploymentMutator {
```
davidfestal, line 51: > downstreamLogicalName logicalcluster.Name

I'd rather call it `syncTargetLogicalClusterName`. There are no logical clusters or workspaces **downstream** (== in the physical cluster).

davidfestal:
@jmprusi Do you think you could test that now, after the last changes, this PR doesn't break the location workspace use-case anymore (introducing security risks on the physical cluster side) ?

pkg/syncer/spec/mutators/deployment.go
lionelvillard: done

jmprusi:
@lionelvillard can you rebase the commits and clean up the story?

ncdc:
/hold
pending go.mod question. also continuing to review

cmd/syncer/cmd/dns.go
```diff
@@ -0,0 +1,44 @@
...
+
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/kcp-dev/kcp/pkg/dns"
+)
+
+func NewDnsCommand() *cobra.Command {
```
ncdc, line 25: NewDNSCommand (xref https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -568,12 +568,21 @@ type templateArgs struct {
 	// ServiceAccount is the name of the service account to create in the syncer
 	// namespace on the pcluster.
 	ServiceAccount string
+	// DnsServiceAccount is the name of the DNS service account to create in the syncer
+	// namespace on the pcluster.
+	DnsServiceAccount string
```
ncdc, line 573: Global comment: DNS, or dns, but not Dns 😄 (https://github.com/golang/go/wiki/CodeReviewComments#initialisms)

pkg/dns/plugin/nsmap/name.go
```diff
@@ -0,0 +1,108 @@
...
+*/
+
+package nsmap
+
+import (
+	"github.com/miekg/dns"
+)
+
+// stringRewriter rewrites a string
+type stringRewriter interface {
```
ncdc, line 24: We don't need to define this interface here. You can embed a `*remapStringRewriter` in `nameRewriterResponseRule` and `valueRewriterResponseRule`.

https://github.com/golang/go/wiki/CodeReviewComments#interfaces

pkg/dns/plugin/nsmap/name.go
```diff
@@ -0,0 +1,108 @@
+/*
```
ncdc, line 1: Any chance you could add unit tests?

pkg/dns/plugin/nsmap/namespace.go
```diff
@@ -0,0 +1,76 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+)
+
+type namespaceRewriter struct {
+	Namespaces map[string]string
+}
+
+func (m *namespaceRewriter) updateFromConfigmap(cm *corev1.ConfigMap) {
+	klog.Infof("reloading nsmap config map : +%v", cm.Data)
```
ncdc, line 35: Please have this take in a `ctx` so you can get the logger from it. We are not adding new non-contextual `klog` statements if we can avoid it.

pkg/dns/plugin/nsmap/setup.go
```diff
@@ -0,0 +1,47 @@
...
+
+func setup(c *caddy.Controller) error {
+	mapper := &namespaceRewriter{Namespaces: map[string]string{}}
+
+	dnsserver.GetConfig(c).AddPlugin(func(next plugin.Handler) plugin.Handler {
+		return NsMap{Next: next, namespace: mapper}
+	})
+
+	// Start config map watcher and notify the mapper
+	done := make(chan struct{})
```
ncdc, line 35: Please use a context with cancel. Add a logger to it. Then we can use it elsewhere.

pkg/dns/plugin/nsmap/config.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+const (
+	// ConfigMapName is the name of the configmap containing logical to physical namespace mappings
+	ConfigMapName = "config-nsmap"
+)
+
+type OnUpdateFn func(configMap *corev1.ConfigMap)
+
+func StartWatcher(done chan struct{}, callback OnUpdateFn) error {
```
ncdc, line 39:
```suggestion
func StartWatcher(ctx context.Context, callback OnUpdateFn) error {
```

ncdc (commented):
I have to stop for the day. This is as far as I've made it - sorry it's not more. Will resume tomorrow.

pkg/dns/plugin/nsmap/name.go
lionelvillard: This is private code coming for CoreDNS, copied as is. I would rather leave it that way to not introduced bugs.

ncdc: If it's copied from another repo, please put it in the third_party dir

lionelvillard: Let me be more precise: this code is coming from the CoreDNS rewrite plugin and I've trimmed it down to just support exact rewriting. I didn't want to go to far in changing the original code but while I'm at it I can just trim it down even further and remove the interface

ncdc: Thanks. It still needs to be in third_party

pkg/syncer/dns/syncer.go
```diff
@@ -0,0 +1,121 @@
...
+// and update the dns nsmap config map
+func StartNsMapSyncer(ctx context.Context, client *dynamic.DynamicClient, informers dynamicinformer.DynamicSharedInformerFactory, ns string) {
+	nsinformer := informers.ForResource(schema.GroupVersionResource{Version: "v1", Resource: "namespaces"})
+	nslister := nsinformer.Lister()
+
+	cmclient := client.Resource(schema.GroupVersionResource{Version: "v1", Resource: "configmaps"}).Namespace(ns)
+
+	nsinformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			updateConfigMap(ctx, cmclient, nslister, ns)
```
ncdc, line 50: This must be a proper controller with a work queue and worker goroutines. We can't have event handler functions take lots of time (e.g. make network requests to update a configmap).

ncdc (changes requested):
Still reviewing but posting what I have so far

pkg/dns/plugin/nsmap/name.go
ncdc: I would probably suggest including a link in a comment to the stackoverflow post, but that code does not need to be in third_party.

Everything coming from another repo must be identified as to where it originated from (ideally GitHub link w/specific commit), it must retain the original copyright header, and it needs to go in third_party.

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -207,6 +211,25 @@ func (dm *DeploymentMutator) Mutate(obj *unstructured.Unstructured) error {
 		templateSpec.Volumes = append(templateSpec.Volumes, serviceAccountVolume)
 	}
 
+	if dm.syncTargetLogicalClusterName == upstreamLogicalName {
```
ncdc, line 214: When would this be true? False? Why only do this when the deployment's logical cluster == the logical cluster where the SyncTarget lives?

pkg/syncer/syncer.go
```diff
@@ -197,13 +207,28 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 		advancedSchedulingEnabled = true
 	}
 
+	dnsIP := ""
+	// Get the DNS IP. The DNS service cannot be recreated as existing dnsConfig won't be updated.
+	err = wait.PollImmediate(2*time.Second, time.Minute, func() (bool, error) {
```
ncdc, line 212: Would 1 second be ok?

pkg/syncer/syncer.go
```diff
@@ -197,13 +207,28 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
...
+		ips, err := net.LookupIP(cfg.DNSServer)
+		if len(ips) == 0 || err != nil {
+			return false, nil //nolint:nilerr
```
ncdc, line 215: If err is not nil, I would store it in a variable declared outside the loop. Then, if the loop fails, include its details in the error returned on L221

test/e2e/virtual/syncer/virtualworkspace_test.go
```diff
@@ -652,6 +653,11 @@ func TestSyncerVirtualWorkspace(t *testing.T) {
 					require.NoError(t, err)
 					t.Log("Installing test CRDs into sink cluster...")
 					fixturewildwest.Create(t, logicalcluster.Name{}, sinkCrdClient.ApiextensionsV1().CustomResourceDefinitions(), metav1.GroupResource{Group: wildwest.GroupName, Resource: "cowboys"})
+
+					if isFakePCluster {
+						// Only need to install services and ingresses in a non-logical cluster
```
ncdc, line 658:
```suggestion
						// Only need to install services in a non-logical cluster
```

test/e2e/syncer/syncer_test.go
```diff
@@ -65,7 +68,19 @@ func TestSyncerLifecycle(t *testing.T) {
 	// its sync target to go ready. This implicitly validates the syncer
 	// heartbeating and the heartbeat controller setting the sync target ready in
 	// response.
-	syncerFixture := framework.NewSyncerFixture(t, upstreamServer, wsClusterName).Start(t)
+	syncerFixture := framework.NewSyncerFixture(t, upstreamServer, wsClusterName,
+		framework.WithExtraResources("services"),
+		framework.WithDownstreamPreparation(func(config *rest.Config, isFakePCluster bool) {
+			if !isFakePCluster {
+				// Only need to install services and ingresses in a logical cluster
```
ncdc, line 75:
```suggestion
				// Only need to install services in a logical cluster
```

test/e2e/syncer/syncer_test.go
```diff
@@ -499,7 +514,19 @@ func TestCordonUncordonDrain(t *testing.T) {
...
```
ncdc, line 521:
```suggestion
				// Only need to install services in a logical cluster
```

test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -91,7 +92,19 @@ func TestNamespaceScheduler(t *testing.T) {
 				// TODO(marun) Extract the heartbeater out of the syncer for reuse in a test fixture. The namespace
 				// controller just needs ready clusters which can be accomplished without a syncer by having the
 				// heartbeater update the sync target so the heartbeat controller can set the cluster ready.
-				syncerFixture := framework.NewSyncerFixture(t, server, server.clusterName).Start(t)
+				syncerFixture := framework.NewSyncerFixture(t, server, server.clusterName,
+					framework.WithExtraResources("services"),
+					framework.WithDownstreamPreparation(func(config *rest.Config, isFakePCluster bool) {
+						if !isFakePCluster {
+							// Only need to install services and ingresses in a logical cluster
```
ncdc, line 99:
```suggestion
							// Only need to install services in a logical cluster
```

test/e2e/reconciler/cluster/controller_test.go
```diff
@@ -179,14 +180,21 @@ func TestClusterController(t *testing.T) {
...
 				framework.WithDownstreamPreparation(func(config *rest.Config, isFakePCluster bool) {
 					// Always install the crd regardless of whether the target is
 					// logical or not since cowboys is not a native type.
 					sinkCrdClient, err := apiextensionsclientset.NewForConfig(config)
 					require.NoError(t, err)
 					t.Log("Installing test CRDs into sink cluster...")
 					fixturewildwest.Create(t, logicalcluster.Name{}, sinkCrdClient.ApiextensionsV1().CustomResourceDefinitions(), metav1.GroupResource{Group: wildwest.GroupName, Resource: "cowboys"})
+
+					if isFakePCluster {
+						// Only need to install services and ingresses in a non-logical cluster
```
ncdc, line 193:
```suggestion
						// Only need to install services in a non-logical cluster
```

pkg/syncer/spec/mutators/deployment.go
davidfestal: That's because this PR provides only a limited implementation, and does not support the upcoming use-case where user workspaces and location workspaces are distinct.

See the following comment: https://github.com/kcp-dev/kcp/pull/1708#issuecomment-1243915910 and the comments before that led to this decision. The followup issue to support the more general case is here: https://github.com/kcp-dev/kcp/issues/1987

ncdc: Could we please add a comment to the code itself with this information? Otherwise, it's not very clear.

pkg/syncer/dns/syncer.go
```diff
@@ -0,0 +1,121 @@
...
+	"k8s.io/client-go/util/retry"
+	"k8s.io/klog/v2"
+
+	"github.com/kcp-dev/kcp/pkg/dns/plugin/nsmap"
+	"github.com/kcp-dev/kcp/pkg/syncer/shared"
+)
+
+// StartNsMapSyncer watches for namespace changes scoped by the given informer factory
+// and update the dns nsmap config map
+func StartNsMapSyncer(ctx context.Context, client *dynamic.DynamicClient, informers dynamicinformer.DynamicSharedInformerFactory, ns string) {
```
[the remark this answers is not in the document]
lionelvillard: they are not available in this context.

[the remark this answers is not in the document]
ncdc: Why not?

pkg/dns/plugin/nsmap/namespace.go
```diff
@@ -0,0 +1,76 @@
...
+func (m *namespaceRewriter) Rewrite(ctx context.Context, state request.Request) rewrite.ResponseRules {
+	name := state.Name()
+
+	parts := strings.SplitN(name, ".", 3)
+	if len(parts) < 2 {
+		// No dots: fallthrough
+		return nil
+	}
+
+	if len(parts) == 3 && !strings.HasPrefix(parts[2], "svc") { // TODO(LV): custom prefix
```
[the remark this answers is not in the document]
lionelvillard: actually it custom zone. `svc` is not configurable (see https://github.com/kubernetes/dns/blob/master/docs/specification.md)

I removed the TODO

pkg/syncer/spec/mutators/deployment.go
davidfestal: @lionelvillard Could you add a TODO here ?

pkg/syncer/dns/syncer.go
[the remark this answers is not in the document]
lionelvillard: I removed the DNS syncer which is now part of the downstream namespace controller. I didn't change the namespaceInformer and use the typed client to update/create ConfigMaps

pkg/dns/plugin/nsmap/namespace.go
lionelvillard: done.

pkg/dns/plugin/nsmap/setup.go
lionelvillard: done.

pkg/syncer/dns/syncer.go
lionelvillard: done. I moved the code to the downstream namespace controller.

pkg/syncer/spec/mutators/deployment.go
lionelvillard: @davidfestal done!

lionelvillard:
@davidfestal @jmprusi @ncdc this is ready for another round of reviews.

jmprusi (approved):

pkg/dns/plugin/nsmap/name.go
lionelvillard: this file has moved to third_party.

PR #1710, opened 2022-08-09
virtual/workspace: use real authorizer

sttts:
Cleanup, and getting rid of the warning in the logs that no authorizer is defined.

PR #1712, opened 2022-08-09
Replace clusterworkspaces/{content,workspace} subresources in RBAC with workspaces{/content,}

sttts:


pkg/virtual/workspaces/registry/rest_test.go
```diff
@@ -361,64 +358,6 @@ func TestListWorkspacesWithUserPermission(t *testing.T) {
 	applyTest(t, test)
 }
 
-func TestListWorkspacesForbidden(t *testing.T) {
```
davidfestal, line 364: Test are removed here, but I didn't see any test added to the new authorizer added in the `build.go` file. Did I miss it ?
Do you think there should be one ?

davidfestal (commented):
Do we want to mention (in the title or description) the move to a real authorizer.
That seems quite an important change in this PR

sttts:
> Do we want to mention (in the title or description) the move to a real authorizer.

You missed https://github.com/kcp-dev/kcp/pull/1710.

pkg/virtual/workspaces/registry/rest_test.go
sttts: added get/delete to e2e tests. List everybody with access has now.

davidfestal: OK. Thanks !

davidfestal (approved):

PR #1715, opened 2022-08-09
Update gitops repo org

kylape:
The Red Hat gitops repo is updated with the latest upon for every push
to main.  The repo is moving to a different org, so we need to update
the github action accordingly.

### DIFF
--- a/.github/workflows/kcp-image.yaml
+++ b/.github/workflows/kcp-image.yaml
@@ -61,7 +61,7 @@ jobs:
       - name: Check out shared-resources (gitops repo)
         uses: actions/checkout@v3
         with:
-          repository: kcp-dev/shared-resources
+          repository: redhat-cps/shared-resources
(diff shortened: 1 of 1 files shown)


.github/workflows/kcp-image.yaml
```diff
@@ -61,7 +61,7 @@ jobs:
       - name: Check out shared-resources (gitops repo)
         uses: actions/checkout@v3
         with:
-          repository: kcp-dev/shared-resources
+          repository: redhat-cps/shared-resources
```
sttts, line 64: this dependency surprises me. kcp-dev should not reference a random deployment.

sttts:
/hold

This needs a plan how to get rid of this dependency.

kylape:
I'll go ahead and implement a polling script to achieve continuous deployment. I'll follow up with a PR to remove this action altogether.

PR #1718, opened 2022-08-09
Fixes log line formatting

m1kola:
## Summary

Small fix to a log line: it looks like `klog.InfoS` was used instead of `klog.Infof` accidentally.

Currently it prints something like this:

```
"Downstream GVR %q object %s|%s/%s does not exist. Removing finalizer upstream" apps/v1, Resource=deployments="" default="test-deployment"
```


ncdc (approved):

sttts:
:+1:

Issue #1225, opened 2022-06-07
Multi-Release-Epic: Sharding


sttts:
# Demo Objective
- [ ] kcp admin should be able to add shards for added capacity, each with its own etcd server.
- [ ] shards should keep working (to be defined what this means exactly) when other shards die or are unavailable.
- [x] https://github.com/kcp-dev/kcp/pull/1203 the front-proxy knows how to forward workspace requests to shards
- add (hard-coded) replication of certain object between shards:
  - [ ] defining ClusterWorkspace `X` in the parent workspace `parent(X)` on the shard the workspace `X` lives on.
  - [ ] APIExport that are bound by at least one non-local binding

# Action Items

- [x] https://github.com/kcp-dev/kcp/pull/1203 forward requests to the right shard from the front-proxy
- [x] https://github.com/kcp-dev/kcp/pull/1203 sharded-test-server binary
- [x] https://github.com/kcp-dev/kcp/pull/1203 setup single-shard + front-proxy CI job
- [x] @p0lyn0mial verify where we are with watch cache and performant label selector

========= WIP, but won't merge by v0.6 =========

- [x] @p0lyn0mial hardcode second informer with non-local data.
- [x] @p0lyn0mial prototype and experiment ~with informer merging, and~ how to enable existing controller to work with second informer
- [x] get 2-shard e2e-sharded into place in CI, with limited test coverage
- get ClusterWorkspace, APIExports, APIBindings, auth to work
  - [x] @p0lyn0mial APIBindings on shards can bind APIExport on other shard
  - [x] @p0lyn0mial ClusterWorkspaces can be scheduled to second shard

========= up to here for v0.7 ==========

#### add identities configmap in system:shard @p0lyn0mial 

- [x] https://github.com/kcp-dev/kcp/pull/1725: implement a controller that maintains a local cm with export identities
- [x] https://github.com/kcp-dev/kcp/pull/1727 update the identity provider to read data from a local config map
- [x] https://github.com/kcp-dev/kcp/pull/1781 wire the APIIdentityCacheController to the kcp server

#### implement replication controller / cache server
   - [x] @p0lyn0mial  create initial `cmd/cache-server` and `pkg/cache/server/{config.go,server.go}` and `pkg/cache/server/options` plumbing (https://github.com/kcp-dev/kcp/pull/1790)
     - follow options/config/server pattern tightly
     - options should be compatible and easily composable with those in pkg/server/options 
[...]

PR #1725, opened 2022-08-10
adds ApiExportIdentityProviderController

p0lyn0mial:
## Summary
The controller reconciles APIExports for the root APIs and
maintains a config map in the `system:shard` logical cluster
with identities per exports specified in the group or the group resources maps.

The config map is meant to be used by clients/informers to inject the identities
for the given GRs when making requests to the server.

TODO: add UTs

## Related issue(s)

Fixes #


pkg/server/bootstrap/api-export-identity_controller.go
```diff
@@ -0,0 +1,255 @@
...
+	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
+	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
+	apisinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/apis/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/indexers"
+)
+
+const (
+	ControllerName = "kcp-api-export-identity-provider"
+	workKey        = "key"
+	configMapName  = "api-exports-identities"
```
p0lyn0mial, line 48: or even `root-api-export-identities`

pkg/server/bootstrap/api-export-identity_controller.go
```diff
@@ -0,0 +1,255 @@
...
+	configMapInformer coreinformers.ConfigMapInformer,
+) (*controller, error) {
+	queue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), ControllerName)
+
+	c := &controller{
+		queue: queue,
+
+		kubeClusterClient: kubeClusterClient,
+
+		configMapIndexer:             configMapInformer.Informer().GetIndexer(),
```
p0lyn0mial, line 73: this might be expensive, all we really need is just a single configmap

stevekuznetsov: Is this the first instance of configmap watching in our code? IIRC APIExport already uses it, no>

pkg/server/bootstrap/api-export-identity_controller.go
```diff
@@ -0,0 +1,255 @@
...
+	})
+	return c, nil
+}
+
+// Start starts the controller, which stops when ctx.Done() is closed.
+func (c *controller) Start(ctx context.Context, numThreads int) {
+	defer runtime.HandleCrash()
+	defer c.queue.ShutDown()
+
+	klog.Infof("Starting %s controller", ControllerName)
```
stevekuznetsov, line 146: Do you mind using logging from e.g. anything in apis/tenancy to follow the structured/contextual pattern?

pkg/server/bootstrap/api-export-identity_controller.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+	for _, name := range c.groupResourceExportNames {
+		apiExport, found := interestingApiExports[name]
+		if !found {
+			return fmt.Errorf("required %s apiExport hasn't been found", name)
+		}
+		requiredApiExportIdentitiesConfigMap.Data[name] = apiExport.Status.IdentityHash
+	}
+
+	rawConfigMaps, err := c.configMapIndexer.ByIndex(indexers.ByLogicalClusterAndNamespace, clusters.ToClusterAwareKey(configshard.SystemShardCluster, "default"))
```
stevekuznetsov, line 217: What's the meaning of listing everything in the namespace, then iterating to find the one name we need, as opposed to just getting it from the lister?

stevekuznetsov (commented):
Is there an issue/epic/etc somewhere that explains what our goal is for this work?

pkg/server/bootstrap/api-export-identity_controller.go
p0lyn0mial: ah, right, I've just realized I could've used `clusters.ToClusterAwareKey` to construct the name.

pkg/server/bootstrap/api-export-identity_controller.go
p0lyn0mial: nope, I think that `installRootCAConfigMapController` brings the configmaps as well.

pkg/server/bootstrap/api-export-identity_controller.go
p0lyn0mial: use `klog.InfoS`, right ?

ncdc: Not quite - see

https://github.com/kcp-dev/kcp/blob/3af88fc9027fea16972ddfd4a1a76a0010255f38/pkg/reconciler/apis/apibinding/apibinding_controller.go#L134

https://github.com/kcp-dev/kcp/blob/3af88fc9027fea16972ddfd4a1a76a0010255f38/pkg/reconciler/apis/apibinding/apibinding_controller.go#L339-L342

https://github.com/kcp-dev/kcp/blob/3af88fc9027fea16972ddfd4a1a76a0010255f38/pkg/reconciler/apis/apibinding/apibinding_controller.go#L364-L366

https://github.com/kcp-dev/kcp/blob/3af88fc9027fea16972ddfd4a1a76a0010255f38/pkg/reconciler/apis/apibinding/apibinding_controller.go#L382

https://github.com/kcp-dev/kcp/blob/3af88fc9027fea16972ddfd4a1a76a0010255f38/pkg/reconciler/apis/apibinding/apibinding_controller.go#L400-L401

@stevekuznetsov do we have this doc'd?

pkg/server/bootstrap/api-export-identity_controller.go
sttts: apiexport-identity-cache ?

pkg/server/bootstrap/api-export-identity_controller.go
```diff
@@ -0,0 +1,255 @@
...
+//
+// The controller reconciles APIExports for the root APIs and
+// maintains a config map in the system:shard logical cluster
+// with identities per exports specified in the group or the group resources maps.
+//
+// The config map is meant to be used by clients/informers to inject the identities
+// for the given GRs when making requests to the server.
+func NewApiExportIdentityProviderController(
+	groupExportNames map[string]string,
+	groupResourceExportNames map[schema.GroupResource]string,
```
sttts, line 61: why do we need these?

The APIExport has `<prefix>.resource.group` like schema references. Why don't we just put one identity per GR into the configmap? For all APIExport found in the root workspace.

sttts: will simplify lots of code below

p0lyn0mial:
> Is there an issue/epic/etc somewhere that explains what our goal is for this work?

we added an item to https://github.com/kcp-dev/kcp/issues/1225 epic but it isn't well described.
the idea is to have a local cm for non-root shards with identities for the root APIs. This increases our fault tolerance because we don't have to contact a live replica.

pkg/server/bootstrap/api-export-identity_controller.go
p0lyn0mial: I just took the hardcoded list :) 

We could make it more dynamic. 
Are those names stable or should we actually read `apiresourceschemas` to get the resource names?

pkg/server/bootstrap/api-export-identity_controller.go
stevekuznetsov: Not yet, but every other reconciler uses it, so it should be easy enough to copy this time ;)

stevekuznetsov:
> we added an item to https://github.com/kcp-dev/kcp/issues/1225 epic but it isn't well described.

Could you please specify which item in that is implemented with this PR by linking it inline? Context is always appreciated :)

pkg/server/bootstrap/api-export-identity_controller.go
sttts: no need to restrict the list. So just cache everything you find in root workspace.

The schema names have fixes structure. So you can get resource and group from them, yes

pkg/server/bootstrap/api-export-identity_controller.go
```diff
@@ -0,0 +1,255 @@
...
+
+	configshard "github.com/kcp-dev/kcp/config/shard"
+	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
+	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
+	apisinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/apis/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/indexers"
+)
+
+const (
+	ControllerName = "kcp-api-export-identity-provider"
```
sttts, line 46: why is this controller not where all other controller are? pkg/reconcilers

p0lyn0mial:
I've updated the PR, PTAL. 
I'll wire the controller in a separate PR.

pkg/reconciler/identitycache/api_export_identity_controller.go
```diff
@@ -0,0 +1,162 @@
+/*
+Copyright 2022 The KCP Authors.
```
sttts, line 2: pkg/reconciler/apis ?

pkg/reconciler/identitycache/api_export_identity_reconcile.go
```diff
@@ -0,0 +1,70 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: "default",
+			Name:      ConfigMapName,
+		},
+		Data: map[string]string{},
+	}
+	for _, rawApiExport := range rawApiExports {
+		apiExport := rawApiExport.(*apisv1alpha1.APIExport)
+		if apiExport.Status.IdentityHash == "" {
+			return fmt.Errorf("apiExport %s is not ready", apiExport.Name)
```
sttts, line 50: return nil. Nothing we can do here about it. When there is an identity, we get an event.

sttts: actually, not even return. Just continue and keep going.

sttts:
/lgtm
/approve

Import job failed.

PR #1726, opened 2022-08-10
cli/workspace: don't get Workspace in parent on absolute use, but check inside the workspace

sttts:
The former GET request meant that a user needs both `verb=access` access for the target workspace *and* the parent workspace. That was never intended. With this PR we just check discovery inside of the target workspace.

Note the TODO in the PR:

```
In both the cases of `root` and absolute paths here we assume that the current cluster client
is talking to the right external URL. This obviously not guaranteed, and hence we
silently assume that the front-proxy will route to every workspace.
We might want to add permanent redirections to the front-proxy if the external
URL does not match the workspace's shard, and then add redirect support here to
use the right front-proxy URL in the kubeonfig.
```

### DIFF
--- a/pkg/cliplugins/workspace/plugin/kubeconfig.go
+++ b/pkg/cliplugins/workspace/plugin/kubeconfig.go
@@ -213,24 +213,39 @@ func (kc *KubeConfig) UseWorkspace(ctx context.Context, name string) (err error)
 		}
 
 		if strings.Contains(name, ":") && cluster.HasPrefix(tenancyv1alpha1.RootCluster) {
-			// absolute logical cluster under root:
+			// e.g. root:something:something
+
+			// first try to get Workspace from parent to potentially get a 404. A 403 in the parent though is
+			// not a blocker to enter the workspace. We will do discovery as a final check below
 			parentClusterName, workspaceName := logicalcluster.New(name).Split()
-			ws, err := kc.clusterClient.Cluster(parentClusterName).TenancyV1beta1().Workspaces().Get(ctx, workspaceName, metav1.GetOptions{})
-			if err != nil {
-				return err
+			if _, err := kc.clusterClient.Cluster(parentClusterName).TenancyV1beta1().Workspaces().Get(ctx, workspaceName, metav1.GetOptions{}); apierrors.IsNotFound(err) {
+				return fmt.Errorf("workspace %q not found", name)
 			}
 
-			// intentionally do not check for readiness here
+			groups, err := kc.clusterClient.Cluster(cluster).Discovery().ServerGroups()
+			if err != nil && !apierrors.IsForbidden(err) {
+				return err
+			}
+			if apierrors.IsForbidden(err) || len(groups.Groups) == 0 {
+				return fmt.Errorf("access to workspace %s denied", name)
+			}
 
-			newServerHost = ws.Status.URL
-			workspaceType = &ws.Spec.Type
+			// TODO(sttts): in both the cases of `root` and absolute paths here we assume that the current cluster
+			//              client is talking to the right external URL. This obviously not guaranteed, and hence
+			//              we silently assume that the front-proxy will route to every workspace.
+			//			    We might want to add permanent redirections to the front-proxy if the external
+			//              URL does not match the workspace's shard, and then add redirect support here to
+			//              use the right front-proxy URL in the kubeconfig.
+
+			u.Path = path.Join(u.Path, cluster.Path())
+			newServerHost = u.String()
 		} else if strings.Contains(name, ":") {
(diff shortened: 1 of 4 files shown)


stevekuznetsov:
Seems like a reasonable trade-off. Unit tests broken.

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -213,17 +213,24 @@ func (kc *KubeConfig) UseWorkspace(ctx context.Context, name string) (err error)
 		}
 
 		if strings.Contains(name, ":") && cluster.HasPrefix(tenancyv1alpha1.RootCluster) {
-			// absolute logical cluster under root:
-			parentClusterName, workspaceName := logicalcluster.New(name).Split()
-			ws, err := kc.clusterClient.Cluster(parentClusterName).TenancyV1beta1().Workspaces().Get(ctx, workspaceName, metav1.GetOptions{})
-			if err != nil {
+			// e.g. root:something
+			groups, err := kc.clusterClient.Cluster(logicalcluster.New(name)).Discovery().ServerGroups()
```
davidfestal, line 217:
```suggestion
			groups, err := kc.clusterClient.Cluster(cluster).Discovery().ServerGroups()
```

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -213,17 +213,24 @@ func (kc *KubeConfig) UseWorkspace(ctx context.Context, name string) (err error)
...
+			// TODO(sttts): in both the cases of `root` and absolute paths here we assume that the current cluster
+			//              client is talking to the right external URL. This obviously not guaranteed, and hence
+			//              we silently assume that the front-proxy will route to every workspace.
+			//			    We might want to add permanent redirections to the front-proxy if the external
+			//              URL does not match the workspace's shard, and then add redirect support here to
+			//              use the right front-proxy URL in the kubeonfig.
 
-			newServerHost = ws.Status.URL
-			workspaceType = &ws.Spec.Type
+			u.Path = path.Join(u.Path, logicalcluster.New(name).Path())
```
davidfestal, line 232:
```suggestion
			u.Path = path.Join(u.Path, cluster.Path())
```

davidfestal: and the same below (several times)

davidfestal (approved):
Approviing with some minor suggestions

sttts:
there are some unit tests to adapt. On it, probably next week.

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -213,24 +213,31 @@ func (kc *KubeConfig) UseWorkspace(ctx context.Context, name string) (err error)
...
-			ws, err := kc.clusterClient.Cluster(parentClusterName).TenancyV1beta1().Workspaces().Get(ctx, workspaceName, metav1.GetOptions{})
-			if err != nil {
+			// e.g. root:something
+			groups, err := kc.clusterClient.Cluster(cluster).Discovery().ServerGroups()
+			if err != nil && !apierrors.IsForbidden(err) {
 				return err
 			}
+			if apierrors.IsForbidden(err) || len(groups.Groups) == 0 {
+				return fmt.Errorf("access to workspace %s denied or workspace does not exist", name)
+			}
```
s-urbaniak, line 223: doesn't it make sense to differentiate the two cases?

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -213,24 +213,31 @@ func (kc *KubeConfig) UseWorkspace(ctx context.Context, name string) (err error)
...
 
-			// intentionally do not check for readiness here
+			// TODO(sttts): in both the cases of `root` and absolute paths here we assume that the current cluster
+			//              client is talking to the right external URL. This obviously not guaranteed, and hence
+			//              we silently assume that the front-proxy will route to every workspace.
+			//			    We might want to add permanent redirections to the front-proxy if the external
+			//              URL does not match the workspace's shard, and then add redirect support here to
+			//              use the right front-proxy URL in the kubeonfig.
```
s-urbaniak, line 230: s/kubeonfig/kubeconfig

pkg/cliplugins/workspace/plugin/kubeconfig.go
sttts: we cannot. In either case you get 403 from the server.

sttts: actually, partially we can. sec

PR #1727, opened 2022-08-10
the identity provider gets data from a local cm too

p0lyn0mial:
## Summary
the config map is maintained by a controller and contains identities for the root API Exports.
the config map is used by non-root shards and provides higher fault tolerance than reading directly from the root shard.

if the config map doesn't exist or there were some errors during getting it we always fall back to the previous behavior of getting identities from exports directly.
## Related issue(s)

xref: see item `1. add identities configmap in system:shard` from  https://github.com/kcp-dev/kcp/issues/1225

### DIFF
--- a/pkg/server/bootstrap/identity.go
+++ b/pkg/server/bootstrap/identity.go
@@ -26,14 +26,15 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/kcp-dev/logicalcluster/v2"
-
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	errorsutil "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 
+	configshard "github.com/kcp-dev/kcp/config/shard"
+	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
 	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
 )
 
@@ -82,9 +83,10 @@ func (ids *identities) grIdentity(gr schema.GroupResource) (string, bool) {
 // with identities.
 func NewConfigWithWildcardIdentities(config *rest.Config,
 	groupExportNames map[string]string,
-	grouptResourceExportNames map[schema.GroupResource]string,
-	kcpClient kcpclient.Interface) (identityConfig *rest.Config, resolve func(ctx context.Context) error) {
-	identityRoundTripper, identityResolver := NewWildcardIdentitiesWrappingRoundTripper(groupExportNames, grouptResourceExportNames, kcpClient)
+	groupResourceExportNames map[schema.GroupResource]string,
+	rootShardKcpClient kcpclient.ClusterInterface,
+	localShardKubeClusterClient kubernetes.ClusterInterface) (identityConfig *rest.Config, resolve func(ctx context.Context) error) {
+	identityRoundTripper, identityResolver := NewWildcardIdentitiesWrappingRoundTripper(groupExportNames, groupResourceExportNames, rootShardKcpClient, localShardKubeClusterClient)
 	identityConfig = rest.CopyConfig(config)
 	identityConfig.Wrap(identityRoundTripper)
 	return identityConfig, identityResolver
@@ -101,7 +103,8 @@ func NewConfigWithWildcardIdentities(config *rest.Config,
 // with identities.
 func NewWildcardIdentitiesWrappingRoundTripper(groupExportNames map[string]string,
 	groupResourceExportNames map[schema.GroupResource]string,
-	kcpClient kcpclient.Interface) (func(rt http.RoundTripper) http.RoundTripper, func(ctx context.Context) error) {
+	rootShardKcpClient kcpclient.ClusterInterface,
+	localShardKubeClusterClient kubernetes.ClusterInterface) (func(rt http.RoundTripper) http.RoundTripper, func(ctx context.Context) error) {
 	ids := &identities{
 		groupIdentities:         map[string]string{},
 		groupResourceIdentities: map[schema.GroupResource]string{},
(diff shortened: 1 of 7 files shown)


stevekuznetsov:
> xref https://github.com/kcp-dev/kcp/issues/1225

Can we update the specific bullet point in the epic that this implements?

pkg/server/bootstrap/identity.go
```diff
@@ -157,26 +160,48 @@ func wildcardIdentitiesResolver(ids *identities,
...
 			ids.lock.Unlock()
 
-			klog.V(2).Infof("APIExport %s|%s for resource %s.%s has identity %s", logicalcluster.From(apiExport), name, gr.Resource, gr.Group, apiExport.Status.IdentityHash)
+			klog.V(2).Infof("APIExport %s|%s for resource %s.%s has identity %s", tenancyv1alpha1.RootCluster, name, gr.Resource, gr.Group, apiExportIdentity)
 		}
 		return errorsutil.NewAggregate(errs)
 	}
 }
 
+func apiExportIdentityProviderFn(kcpClient kcpclient.Interface, kubeClusterClient kubernetes.ClusterInterface) func(ctx context.Context, apiExportName string) (string, error) {
```
stevekuznetsov, line 183: nit: don't name this `Fn` - we can see it's a function today and it may not be one in the future. `apiExportIdentityResolver` or similar describes what it does

stevekuznetsov: while `kcpClient` and `kubeClusterClient` are technically not incorrect names, it would be nice to know that they are for different shards - perhaps `rootShardClient` and `localShardClient`?

pkg/server/bootstrap/identity.go
```diff
@@ -157,26 +160,48 @@ func wildcardIdentitiesResolver(ids *identities,
...
+	return func(ctx context.Context, apiExportName string) (string, error) {
+		if kubeClusterClient != nil {
+			actualApiExportIdentitiesConfigMap, err := kubeClusterClient.Cluster(configshard.SystemShardCluster).CoreV1().ConfigMaps("default").Get(ctx, configMapName, metav1.GetOptions{})
```
stevekuznetsov, line 186: What makes this the "actual" map? As opposed to a fake map?

pkg/server/bootstrap/identity.go
```diff
@@ -157,26 +160,48 @@ func wildcardIdentitiesResolver(ids *identities,
...
+			if err == nil {
+				apiExportIdentity, found := actualApiExportIdentitiesConfigMap.Data[apiExportName]
+				if !found {
+					return "", fmt.Errorf("an identity entry for APIExport %s wasn't found in %s|%s configmap", apiExportName, configshard.SystemShardCluster, configMapName)
+				}
+				return apiExportIdentity, nil
+			}
+			// in case of any error just fallback and try to get an export directly
```
stevekuznetsov, line 194: L190 returns an error when the export is not found in the CM - in what sense is this fallback?

pkg/server/config.go
```diff
@@ -211,7 +222,7 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
 		if err != nil {
 			return nil, err
 		}
-		c.identityConfig, c.resolveIdentities = boostrap.NewConfigWithWildcardIdentities(c.GenericConfig.LoopbackClientConfig, boostrap.KcpRootGroupExportNames, boostrap.KcpRootGroupResourceExportNames, nonIdentityKcpClusterClient.Cluster(tenancyv1alpha1.RootCluster))
+		c.identityConfig, c.resolveIdentities = boostrap.NewConfigWithWildcardIdentities(c.GenericConfig.LoopbackClientConfig, boostrap.KcpRootGroupExportNames, boostrap.KcpRootGroupResourceExportNames, nonIdentityKcpClusterClient, nil)
```
stevekuznetsov, line 225: In what cases is this local client `nil` and why?

pkg/server/bootstrap/identity.go
p0lyn0mial: sure, np will rename

pkg/server/bootstrap/identity.go
p0lyn0mial: will rename to something else

pkg/server/bootstrap/identity.go
p0lyn0mial: we fall back when the cm wasn't found or we weren't able to get one. 

we could also fallback when an entry wasn't found but I'd let the controller change to update the cm - the resolver will retry.

pkg/server/config.go
p0lyn0mial: it is `nil` on the root shard because we read identities directly from the exports

p0lyn0mial:
>  [stevekuznetsov](https://github.com/stevekuznetsov)
 Can we update the specific bullet point in the epic that this implements?

done, please see if you like it.

pkg/server/bootstrap/identity.go
```diff
@@ -157,26 +160,48 @@ func wildcardIdentitiesResolver(ids *identities,
...
 	}
 }
 
+func apiExportIdentityProvider(rootShardKcpClient kcpclient.Interface, localShardKubeClusterClient kubernetes.ClusterInterface) func(ctx context.Context, apiExportName string) (string, error) {
+	return func(ctx context.Context, apiExportName string) (string, error) {
+		if localShardKubeClusterClient != nil {
+			apiExportIdentitiesConfigMap, err := localShardKubeClusterClient.Cluster(configshard.SystemShardCluster).CoreV1().ConfigMaps("default").Get(ctx, configMapName, metav1.GetOptions{})
+			if err == nil {
+				apiExportIdentity, found := apiExportIdentitiesConfigMap.Data[apiExportName]
+				if !found {
```
sttts, line 189: why not fall through?

pkg/server/bootstrap/identity.go
sttts: this would couple bootstrapping with the controller startup. Am not convinced that is good. Unnecessary complexity. Just fall through on incomplete CM and read from export.

Note: this is a rare case.

p0lyn0mial: no strong opinion here, I can update if we like it.

p0lyn0mial: > this would couple bootstrapping with the controller startup

it wouldn't as we create a config map with data

p0lyn0mial: > Unnecessary complexity

not falling back proves the controller actually works and it might be easier to debug as the source is easier to trace.

pkg/server/bootstrap/identity.go
p0lyn0mial: done

PR #1738, opened 2022-08-10
hack: verify-go-version.sh in all the repo

sttts:
So many places where we reference a go version.

Check that the build go version matches.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ require-%:
 	@if ! command -v $* 1> /dev/null 2>&1; then echo "$* not found in \$$PATH"; exit 1; fi
 
 build: WHAT ?= ./cmd/...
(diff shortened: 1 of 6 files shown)


stevekuznetsov (approved):

Makefile
```diff
@@ -249,6 +249,10 @@ verify-k8s-deps:
 verify-imports:
 	hack/verify-imports.sh
 
+.PHONY: verify-go-imports
+verify-go-versions:
+	hack/verify-go-versions.sh
```
stevekuznetsov, line 254:
```suggestion
	hack/verify-go-version.sh
```

sttts: yes, fixed

sttts:
@stevekuznetsov with 1.18 most image builds fail here with a killed linker. Do we need more memory?

PR #1745, opened 2022-08-15
Permission claim fixes / cleanups

sttts:
Extracted from https://github.com/kcp-dev/kcp/pull/1563/commits.

### DIFF
--- a/pkg/admission/permissionclaims/mutating_permission_claims.go
+++ b/pkg/admission/permissionclaims/mutating_permission_claims.go
@@ -24,7 +24,6 @@ import (
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/apiserver/pkg/admission"
 	genericapirequest "k8s.io/apiserver/pkg/endpoints/request"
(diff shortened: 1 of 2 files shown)


pkg/admission/permissionclaims/mutating_permission_claims.go
```diff
@@ -93,29 +96,31 @@ func (m *mutatingPermissionClaims) Admit(ctx context.Context, a admission.Attrib
...
+			continue
+		}
+		if _, expected := expectedLabels[k]; !expected {
 			delete(labels, k)
 		}
 	}
-
-	// Add back what should be there
 	for k, v := range expectedLabels {
+		if labels == nil {
```
shawn-hurley, line 108: Why do this inside the loop instead of at line 96?

pkg/admission/permissionclaims/mutating_permission_claims.go
```diff
@@ -93,29 +96,31 @@ func (m *mutatingPermissionClaims) Admit(ctx context.Context, a admission.Attrib
...
 	}
-
 	u.SetLabels(labels)
 
 	return nil
 
 }
 
 func (m *mutatingPermissionClaims) Validate(ctx context.Context, a admission.Attributes, o admission.ObjectInterfaces) error {
+	if a.GetSubresource() != "" {
```
shawn-hurley, line 120: Is it always the case that we don't want to validate the labels the subresource? 

I am unaware of a subresource that will let you change metadata, but I thought I would ask.

shawn-hurley (commented):
I have a couple of questions for myself to understand.

pkg/admission/permissionclaims/mutating_permission_claims.go
sttts: because only here it is needed

pkg/admission/permissionclaims/mutating_permission_claims.go
sttts: as admission is best effort in our case this should be safe, even if some day some other subresource is added that allows it (I doubt it).

pkg/reconciler/apis/permissionclaimlabel/permissionclaimlabel_resource_reconcile.go
```diff
@@ -44,19 +44,14 @@ func (c *resourceController) reconcile(ctx context.Context, obj *unstructured.Un
 		return fmt.Errorf("error calculating permission claim labels for GVR %q %s/%s: %w", gvr, obj.GetNamespace(), obj.GetName(), err)
 	}
 
-	labels := obj.GetLabels()
-	if labels == nil {
-		labels = make(map[string]string)
-	}
```
sttts, line 50: background for this commit: it is not good to depend on nil-vs-empty behaviour of `LabelsFor`.

PR #1749, opened 2022-08-15
⚠️ WIP: virtual/initializingworkspaces: serve workspaces instead of clusterworkspaces

sttts:
PoC implementation.

Background: we don't want to give anybody access to ClusterWorkspaces other than the platform provider.

/assign @stevekuznetsov


stevekuznetsov:
@sttts do we still want this?

ncdc:
As discussed on Slack, closing for now, but it'll still be in the repo so we can come back and look at it again for ideas if needed.

PR #1754, opened 2022-08-16
Generic helper for consistent post-reconcile spec and status updates

csams:
## Summary
Unify post-reconcile spec and status updating logic in a type safe way. The `APIBinding` and `APIExport` controllers are updated as examples.

### DIFF
--- a/pkg/reconciler/apis/apibinding/apibinding_controller.go
+++ b/pkg/reconciler/apis/apibinding/apibinding_controller.go
@@ -46,7 +46,7 @@ import (
 	apislisters "github.com/kcp-dev/kcp/pkg/client/listers/apis/v1alpha1"
 	"github.com/kcp-dev/kcp/pkg/informer"
 	"github.com/kcp-dev/kcp/pkg/logging"
-	"github.com/kcp-dev/kcp/pkg/reconciler/postreconcile"
+	committer "github.com/kcp-dev/kcp/pkg/reconciler/committer"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -129,6 +126,7 @@ func NewController(
 		},
 		crdIndexer:        crdInformer.Informer().GetIndexer(),
 		deletedCRDTracker: newLockedStringSet(),
+		postReconcile:     postreconcile.NewPostReconcile[*APIBinding, *APIBindingSpec, *APIBindingStatus](kcpClusterClient.ApisV1alpha1().APIBindings()),
```
sttts, line 129: isn't that more a "commit" phase?

sttts:
Sgtm.

Only bike-shedding: postreconcile => commit.

pkg/reconciler/apis/apibinding/apibinding_controller.go
csams: Makes sense. I updated the naming.

pkg/reconciler/committer/committer.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+
+		logger.V(2).Info(fmt.Sprintf("patching %s", focusType), "patch", string(patchBytes))
+		_, err = patcher.Patch(logicalcluster.WithCluster(ctx, clusterName), obj.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{}, subresources...)
+		if err != nil {
+			return fmt.Errorf("failed to patch %s %s|%s: %w", focusType, clusterName, name, err)
+		}
+
+		// Simultaneous updates of spec and status are never allowed.
+		if specOrObjectMetaChanged && statusChanged {
```
csams, line 115: @stevekuznetsov Shouldn't this go near the top?

stevekuznetsov: yes, that should be as early as possible

PR #1756, opened 2022-08-16
modified cmd.Use for kubectl-workspaces

### LINKED ISSUE #1579: Incomplete Usage message from kubectl kcp workspace plugin
**Describe the bug**
The usage message is incomplete.

```console
(base) mspreitz@mjs12 kcp % kubectl ws --help
Manages KCP workspaces

Usage:
  workspace [list|create|create-context|<workspace>|..|.|-|~|<root:absolute:workspace>] [flags]
...
```

That listing does not mention `use` or `current`.

**To Reproduce**
1. `kubectl ws --help`
2. Observe the incomplete synopsis

**Expected behavior**
The synopsis includes all the available commands, or none of them, or an explicit indication that the list is incomplete.

**Additional context**
N/A

apoorvajagtap:
## Summary: 
As sub-commands like create/create-context required additional details about workspace, whereas rest of the commands did not. So, the usage section `workspace [create|create-context|<workspace>|..|.|-|~|<root:absolute:workspace>] [flags]` might lead to a confusion as some sub-commands aren't specified. Modified the usage section as below, using the generic keywords:
`Usage:
  workspace [command] [flags]
  workspace [command]
` 
Considered keeping (workspace [command]) for commands like `use`.
Let me know, if we should only keep `workspace [command] [flags]` instead of both the statements.

Fixes #1579

### DIFF
--- a/pkg/cliplugins/workspace/cmd/cmd.go
+++ b/pkg/cliplugins/workspace/cmd/cmd.go
@@ -90,7 +90,7 @@ func New(streams genericclioptions.IOStreams) (*cobra.Command, error) {
 	}
 	cmd := &cobra.Command{
 		Aliases:          []string{"ws", "workspaces"},
-		Use:              "workspace [create|create-context|<workspace>|..|.|-|~|<root:absolute:workspace>]",
+		Use:              "workspace [command]",
 		Short:            "Manages KCP workspaces",
 		Example:          fmt.Sprintf(workspaceExample, "kubectl kcp"),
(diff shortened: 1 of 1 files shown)


sttts:
not everything is a subcommand. So I don't think shortening the help text is the right direction.

apoorvajagtap:
Hi @sttts, thanks for the inputs. I've now added use & current as well to the Usage section. Let me know if this fits better or not. Thanks again!

PR #1762, opened 2022-08-16
Start centralizing APIExport indexers

### LINKED ISSUE #1752: APIExports virtual workspace broken when virtual workspaces are run standalone and an APIExport has permission claims
When the virtual workspaces process is run standalone, the virtual workspace for an APIExport with permission claims is broken because `indexers.IndexAPIExportByIdentity` is not configured (this only currently happens when the APIExport controller is instantiated, which does not happen in the virtual workspaces process).

ncdc:
## Summary
Move the definition of APIExport index functions needed across multiple controllers to pkg/indexers

## Related issue(s)

Fixes #1752

### DIFF
--- a/pkg/indexers/util.go
+++ b/pkg/indexers/util.go
@@ -43,3 +43,18 @@ func AppendOrDie(indexers ...cache.Indexers) cache.Indexers {
 	}
 	return ret
 }
+
+// AddIfNotPresentOrDie tries to add everything from toAdd to indexer's indexers that does not already exist. It panics
+// if it encounters an error.
+func AddIfNotPresentOrDie(indexer cache.Indexer, toAdd cache.Indexers) {
+	existing := indexer.GetIndexers()
+	for indexName := range toAdd {
+		if _, exists := existing[indexName]; exists {
+			delete(toAdd, indexName)
+		}
+	}
+
+	if err := indexer.AddIndexers(toAdd); err != nil {
+		panic(fmt.Errorf("error adding indexers: %w", err))
+	}
+}
--- a/pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_controller.go
+++ b/pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_controller.go
@@ -37,6 +37,7 @@ import (
 	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
 	apisinformer "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/apis/v1alpha1"
 	apislisters "github.com/kcp-dev/kcp/pkg/client/listers/apis/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/indexers"
 	"github.com/kcp-dev/kcp/pkg/logging"
 	"github.com/kcp-dev/kcp/pkg/virtual/framework/dynamic/apidefinition"
 	dynamiccontext "github.com/kcp-dev/kcp/pkg/virtual/framework/dynamic/context"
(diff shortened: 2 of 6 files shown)


ncdc:
ok i broke something, fixing

pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_controller.go
```diff
@@ -76,7 +77,8 @@ func NewAPIReconciler(
 	}
 
 	if err := apiExportInformer.Informer().AddIndexers(cache.Indexers{
-		byWorkspace: indexByWorkspace,
+		byWorkspace:                  indexByWorkspace,
+		indexers.APIExportByIdentity: indexers.IndexAPIExportByIdentity,
```
shawn-hurley, line 81: is this potentially already added?

shawn-hurley (commented):
Assuming that this is a stop-gap to allow for a standalone VW server, this looks good to me.

one question but otherwise looks good to me

pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_controller.go
ncdc: that was the bug - refresh & see the latest 😄

ncdc:
fixed my brokenness

ncdc:
This won't cherry-pick cleanly to release-0.7, so I'll manually backport it

stevekuznetsov (approved):

stevekuznetsov:
```
 pkg/indexers/util.go:58:49: non-wrapping format verb for fmt.Errorf. Use `%w` to format errors (errorlint)
		panic(fmt.Errorf("error adding indexers: %v", err))
		                                              ^ 
```

pkg/indexers/util.go
```diff
@@ -43,3 +43,18 @@ func AppendOrDie(indexers ...cache.Indexers) cache.Indexers {
...
 	return ret
 }
+
+// AddIfNotPresentOrDie tries to add everything from toAdd to indexer's indexers that does not already exist. It panics
+// if it encounters an error.
+func AddIfNotPresentOrDie(indexer cache.Indexer, toAdd cache.Indexers) {
+	existing := indexer.GetIndexers()
+	for indexName := range toAdd {
+		if _, exists := existing[indexName]; exists {
+			delete(toAdd, indexName)
```
sttts, line 53: a pity function pointers cannot be compared

ncdc:
Fixed lint

pkg/indexers/util.go
shawn-hurley: This may seem silly, but this would cleanly allow during startup two indexers named the same but do two different things, right?

I feel like a bug like that would be hard to track down. Would it be worth the cost of using reflection to compare the function pointers?

https://pkg.go.dev/reflect#Value.Pointer

shawn-hurley (commented):
This is exactly what I was thinking, just one minor point/question.

ncdc:
> This may seem silly, but this would cleanly allow during startup two indexers named the same but do two different things, right?

It should be removing any indexes in `toAdd` that already exist, so they don't get added additional times.

shawn-hurley:
The indexer function that was added before could do one set of behavior A, and the one added now could do different behavior B. This will currently just always choose A, which I was just suggesting could make debugging why you are not seeing behavior B a little challenging. If you compared the function pointers, I thought it would be worth saying these functions are not the same.

ncdc:
@shawn-hurley I did some research, and there is not a reliable way to compare 2 function pointers (the only guarantee is you can compare them to `nil`).

shawn-hurley:
/lgtm 

Sorry for sending you down the wrong path. I thought the reflection would work.
Can we add a v4 log or something then for when it does happen just to aid in debugging?

PR #1765, opened 2022-08-16
[release-0.7] Start centralizing APIExport indexers

ncdc:
Backport of #1762 from main to release-0.7

### DIFF
--- a/pkg/indexers/util.go
+++ b/pkg/indexers/util.go
@@ -43,3 +43,18 @@ func AppendOrDie(indexers ...cache.Indexers) cache.Indexers {
 	}
 	return ret
 }
+
+// AddIfNotPresentOrDie tries to add everything from toAdd to indexer's indexers that does not already exist. It panics
+// if it encounters an error.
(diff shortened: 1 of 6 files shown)


pkg/indexers/util.go
```diff
@@ -43,3 +43,18 @@ func AppendOrDie(indexers ...cache.Indexers) cache.Indexers {
...
 	return ret
 }
+
+// AddIfNotPresentOrDie tries to add everything from toAdd to indexer's indexers that does not already exist. It panics
+// if it encounters an error.
+func AddIfNotPresentOrDie(indexer cache.Indexer, toAdd cache.Indexers) {
+	existing := indexer.GetIndexers()
+	for indexName := range toAdd {
+		if _, exists := existing[indexName]; exists {
+			delete(toAdd, indexName)
```
shawn-hurley, line 53: What do we think of adding a log line here, just to maybe help in debugging the scenario?

Probably a follow-up in master, though?

shawn-hurley (approved):
/lgtm

I think the question/comment is just a follow-up or something to come back to if we think it is useful.

PR #1773, opened 2022-08-16
Enable cache mutation detection

### LINKED ISSUE #1134: Enable cache mutation detector when running tests
- set `KUBE_CACHE_MUTATION_DETECTOR=1` when running at least e2e tests
- need to do it for the shared server launched in the e2e-shared-server job
- need to do it for private kcp server's launched by the e2e framework

ncdc:
## Summary
Enable cache mutation detection

## Related issue(s)

Fixes #1134

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -182,7 +182,10 @@ endif
 test-e2e: TEST_ARGS ?=
 test-e2e: WHAT ?= ./test/e2e...
 test-e2e: build-all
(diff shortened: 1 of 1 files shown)


sttts:
What is the overhead?

Shouldn't these env var be merely set from prow/github-actions?

stevekuznetsov (approved):

stevekuznetsov:
/hold

Why would we not want this locally?

ncdc:
I think doing this in CI rather than locally probably makes the most sense.

/close

PR #1774, opened 2022-08-16
✨ virtual/workspaces: add owner references to RBAC role and binding

sttts:
When GC is implemented and the TODO is resolved, deletion of workspaces and their permission is more natural. But it's good to already add the owner refs now before we persist objects without.


pkg/virtual/workspaces/registry/rest.go
```diff
@@ -356,6 +425,8 @@ func (s *REST) Delete(ctx context.Context, name string, deleteValidation rest.Va
 	if kerrors.IsNotFound(errorToReturn) {
 		errorToReturn = kerrors.NewNotFound(tenancyv1beta1.Resource("workspaces"), name)
 	}
+
+	// TODO(sttts): remove when GC is implemented
```
hasheddan, line 429: Propagated TODO to https://github.com/kcp-dev/kcp/pull/2113 in the event it ends up being accepted.

Issue #1723, opened 2022-08-10
Delete specific worksapce multi times, server output fake messages


Phaow:
**Describe the bug**
Delete worksapce multi times, server output fake messages since the second time 
If we already delete 1 ws, delete the second time it should report `not found` err info. But actually  the result is still `[workspace.tenancy.kcp.dev](http://workspace.tenancy.kcp.dev/) "xxxx" deleted` makes confusing. 
kcp server version is v0.7.1

**To Reproduce**
1. Create workspace "myws-test" under user home workspace.
2. Delete the "myws-test".
3. Get the workspaces.
4. Delete the "myws-test" the second time.

**Expected behavior**
At Step 4 output should report `not found` err info.

**Actual result**
At Step 4 output stiil report `[workspace.tenancy.kcp.dev](http://workspace.tenancy.kcp.dev/) "myws-test" deleted`.

```
✘$ kubectl ws .
Current workspace is "root:users:rp:pv:rh-sso-xxxx".
$ kubectl ws
$ kubectl ws create myws-test
$ kubectl delete myws-test
workspace.tenancy.kcp.dev "myws-test" deleted
$ kubectl get workspaces
No resources found
$ kubectl delete myws-test
workspace.tenancy.kcp.dev "myws-test" deleted
```

sttts:
Fixed in https://github.com/kcp-dev/kcp/issues/1776.

PR #1776, opened 2022-08-16
virtual/workspaces: do not filter if able to get all

sttts:
Admins or users in their home workspace can see all workspaces. Filtering is not necessary, and until GC is implemented (see https://github.com/kcp-dev/kcp/pull/1774) even wrong.

Fixes https://github.com/kcp-dev/kcp/issues/1723.

### DIFF
--- a/pkg/virtual/workspaces/registry/rest.go
+++ b/pkg/virtual/workspaces/registry/rest.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/apimachinery/pkg/watch"
 	kuser "k8s.io/apiserver/pkg/authentication/user"
+	"k8s.io/apiserver/pkg/authorization/authorizer"
 	apirequest "k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/apiserver/pkg/registry/rest"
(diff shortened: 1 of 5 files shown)


pkg/virtual/workspaces/registry/rest.go
```diff
@@ -370,3 +419,34 @@ func (s *REST) Delete(ctx context.Context, name string, deleteValidation rest.Va
...
+			if ev.Object == nil {
+				w.ch <- ev
+				continue
+			}
+			if cws, ok := ev.Object.(*tenancyv1alpha1.ClusterWorkspace); ok {
+				ws := &tenancyv1beta1.Workspace{}
+				projection.ProjectClusterWorkspaceToWorkspace(cws, ws)
+				ev.Object = ws
+			}
+			w.ch <- ev
```
stevekuznetsov, line 442: I imagine we want to have the same kind of handling here for failing to send within some reasonable amount of time as we do in the watcher itself, noi?

sttts: which is? What can we do here other than panic?

stevekuznetsov (approved):

sttts:
Pure rebase.

stevekuznetsov (approved):

PR #1782, opened 2022-08-17
bootstrap: create the default ns in the system:shard ws

p0lyn0mial:
## Summary
the ns is required to store the apiexport-identity-cache config map

## Related issue(s)

Fixes #

### DIFF
--- a/config/shard/bootstrap.go
+++ b/config/shard/bootstrap.go
@@ -18,19 +18,30 @@ package shard
 
 import (
 	"context"
+	"embed"
 
 	"github.com/kcp-dev/logicalcluster/v2"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
 
 	confighelpers "github.com/kcp-dev/kcp/config/helpers"
 	kcpclientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
 )
 
+//go:embed *.yaml
+var fs embed.FS
+
 // SystemShardCluster is the name of a logical cluster on every shard (including the root shard) that holds essential system resources (like the root APIs).
 var SystemShardCluster = logicalcluster.New("system:shard")
 
 // Bootstrap creates resources required for a shard.
-// As of today creating API bindings for the root APIs is enough.
-func Bootstrap(ctx context.Context, kcpClient kcpclientset.Interface) error {
+// As of today creating API bindings for the root APIs and the default ns is enough.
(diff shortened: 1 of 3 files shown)


config/shard/default-namespace.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: default
+  annotations:
+    "bootstrap.kcp.dev/create-only": "true"
```
p0lyn0mial, line 6: I think it is okay, we won't change the default ns anytime soon

config/shard/bootstrap.go
```diff
@@ -18,19 +18,30 @@ package shard
...
 )
 
+//go:embed *.yaml
+var fs embed.FS
+
 // SystemShardCluster is the name of a logical cluster on every shard (including the root shard) that holds essential system resources (like the root APIs).
 var SystemShardCluster = logicalcluster.New("system:shard")
 
 // Bootstrap creates resources required for a shard.
 // As of today creating API bindings for the root APIs is enough.
```
stevekuznetsov, line 40: out of date?

config/shard/bootstrap.go
```diff
@@ -18,19 +18,30 @@ package shard
...
+	"k8s.io/client-go/dynamic"
 
 	"github.com/kcp-dev/logicalcluster/v2"
 
 	confighelpers "github.com/kcp-dev/kcp/config/helpers"
 	kcpclientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
 )
 
+//go:embed *.yaml
+var fs embed.FS
```
stevekuznetsov, line 34: nit: usually we name these with what's in the files

pkg/server/server.go
```diff
@@ -140,7 +140,11 @@ func (s *Server) Run(ctx context.Context) error {
 		klog.Infof("Finished bootstrapping system CRDs")
 
 		if err := wait.PollInfiniteWithContext(goContext(ctx), time.Second, func(ctx context.Context) (bool, error) {
-			if err := configshard.Bootstrap(ctx, s.KcpClusterClient.Cluster(configshard.SystemShardCluster)); err != nil {
+			if err := configshard.Bootstrap(ctx,
+				s.ApiExtensionsClusterClient.Cluster(configshard.SystemShardCluster).Discovery(),
```
stevekuznetsov, line 144: Does it matter from which clientset we grab the discovery client? I remember at some point the context/round-tripper stuff was different between the two?

p0lyn0mial: yes, it does matter, since the discovery differs across logical clusters. In this case, it is okay because `system:shard` cluster knows about the built-in types like namespaces.

stevekuznetsov (approved):

Issue #342, opened 2021-12-21
Build A MVP Persistent Caching Proxy


sttts:
Certain data in a kcp cluster is critical for operation and SPOF in a multi-region/AZ environment, e.g. org workspaces. It is feasible to build a consistent cache hierarchy which 
1. persists objects to a local etcd or
2. keep data in memory if one makes the proxies highly available through multiple instances
and which serve consistent data by checking freshness of the cache on quorum reads. Using such a setup could give read-only availability to e.g. org workspaces.

Big question: do we run into time-travel problems as we do with pods and kubelets. So we must be super careful when answering consistent reads with potentially stale data in an outage situation. But for certain operation like the personal workspace virtual workspace from @davidfestal, a stale read is good enough.

## Acceptance Criteria

We're just looking for an MVP implementation here:

- shards push data to the proxy
- no auth{n,z} needed
- APIExport  + APIResourceSchema hard-coded as types to push
- dynamically determine which types to serve. For example, the built-in types (APIExports and APIResourceSchema) could be automatically served as CRDs.
- turn on the watch cache in the cache server
- have secondary informers in kcp reconcilers use the cache server

## Items
   - [x] @p0lyn0mial  create initial `cmd/cache-server` and `pkg/cache/server/{config.go,server.go}` and `pkg/cache/server/options` plumbing (https://github.com/kcp-dev/kcp/pull/1790)
     - follow options/config/server pattern tightly
     - options should be compatible and easily composable with those in pkg/server/options 
   - [x] @p0lyn0mial  instantiate `apiextensions-apiserver` and wire CRD lister similar to today's `pkg/server/apiextensions.go` with at least APIExports and APIResourceSchemas
     - use empty schemas, as dumb as possible: APIExport/APIResourceSchema CRDs can be static as in https://github.com/kcp-dev/kcp/blob/8d2ed6a6ba467ddb4dcba0a7e1675cf8382223c3/pkg/virtual/apiexport/schemas/apis.go#L36
     - no admission, no validation, no conversion, no pruning, no SSA (if possible), for now no authorization
     - PRs: https://github.com/kcp-dev/kcp/pull/1811, https://github.com/kcp-dev/kcp/pull/1815, https://github.com/kcp-dev/kubernetes/pull/93
[...]

PR #1790, opened 2022-08-18
cache-server scaffolding

p0lyn0mial:
## Summary
This PR creates an empty server that follows `options/config/server` pattern. 
The server can be run in two modes, as a standalone binary or can be embedded into the existing server.
This is the first part, in the following PRs I'm planning to wire the generic api server, the etcd, and the apiextention server.

```
The cache server hosts data/resources that are required by shards.
It serves as a cache helping to reduce the storage that would have to
be copied onto every shard otherwise.

The actual group of shards that will use this server should be part of 
the topology. For example, it can be used only by shards that are in 
the same geographical region.

On a high level, the server exposes two HTTP paths. The first one is 
used by the shards for getting all resources. The second one is used 
by individual shards to push data they wish to be shared.
There are no limits on the types of data this server hosts. The rule of 
thumb is that they must be common for a larger group of shards. 
For example the root APIs. 
```
## Related issue(s)

Fixes #


stevekuznetsov (approved):

stevekuznetsov:

PR #1800, opened 2022-08-19
sharded-test-server: support standalone vw server

### LINKED ISSUE #1702: Add option to cmd/test-server and/or cmd/sharded-test-server to deploy the virtual workspaces server standalone
This should be optional, and match the way the Helm chart deploys the virtual workspaces server in a standalone container, e.g.

https://github.com/kcp-dev/helm-charts/blob/d3c0a50c6dc3f61421305156b5c7e93fd83c35db/charts/kcp/templates/kcp.yaml#L205-L206

https://github.com/kcp-dev/helm-charts/blob/d3c0a50c6dc3f61421305156b5c7e93fd83c35db/charts/kcp/templates/kcp.yaml#L283-L304

ncdc:
## Summary
When running sharded-test-server, if you now pass --shard-run-virtual-workspaces=false, it will run virtual-workspaces in
standalone mode, and configure the front proxy to map to the virtual-workspaces process for /services.

Note, this does not currently wire anything into CI/e2e tests. Figured we'd want to discuss where to plumb this in.

## Related issue(s)

Fixes #1702


cmd/sharded-test-server/frontproxy.go
```diff
@@ -53,9 +60,9 @@ func startFrontProxy(ctx context.Context, args []string, servingCA *crypto.CA, h
 		lineprefix.Color(color.New(color.FgHiWhite)),
 	)
 
-	if err := ioutil.WriteFile(filepath.Join(workDirPath, ".kcp-front-proxy/mapping.yaml"), []byte(`
+	if err := ioutil.WriteFile(filepath.Join(workDirPath, ".kcp-front-proxy/mapping.yaml"), []byte(fmt.Sprintf(`
```
stevekuznetsov, line 63: Since we're now print-formatting some YAML, would it be difficult to define an anonymous type here and marshal it instead?

cmd/sharded-test-server/main.go
```diff
@@ -107,6 +109,11 @@ func start(proxyFlags, shardFlags []string, logDirPath, workDirPath string, numb
 		return err
 	}
 
+	standaloneVW := sets.NewString(shardFlags...).Has("--run-virtual-workspaces=false")
+	if standaloneVW {
+		shardFlags = append(shardFlags, fmt.Sprintf("--shard-virtual-workspace-url=https://%s:7444", hostIP))
```
stevekuznetsov, line 114: Wouldn't we run one per shard?

ncdc: Yeah that's probably a change I should make

cmd/sharded-test-server/frontproxy.go
ncdc: Will look into it

cmd/sharded-test-server/virtual.go
```diff
@@ -0,0 +1,80 @@
...
+	out := lineprefix.New(
+		lineprefix.Prefix(yellow(prefix)),
+		lineprefix.Color(color.New(color.FgHiYellow)),
+	)
+
+	commandLine := framework.DirectOrGoRunCommand("virtual-workspaces")
+	commandLine = append(
+		commandLine,
+		"--kubeconfig=.kcp-0/admin.kubeconfig",
+		"--context=system:admin",
```
sttts, line 48: @s-urbaniak @csams @kylape reminder: we definitely don't want this medium term, but find a way, probably some special RBAC permission to do wildcard list/watch.

cmd/sharded-test-server/main.go
sttts: yes, one per shard please cc @p0lyn0mial

cmd/sharded-test-server/frontproxy.go
```diff
@@ -53,18 +61,39 @@ func startFrontProxy(ctx context.Context, args []string, servingCA *crypto.CA, h
...
+		},
+		{
+			Path: "/clusters/",
+			// TODO: support multiple shard backend servers
+			Backend:         "https://localhost:6444",
+			BackendServerCA: ".kcp/serving-ca.crt",
+			ProxyClientCert: ".kcp-front-proxy/requestheader.crt",
+			ProxyClientKey:  ".kcp-front-proxy/requestheader.key",
+		},
+	}
```
sttts, line 89: for follow-up: we should really turn this mapping file into flags. The file promises flexibility that is not there.

cmd/sharded-test-server/virtual.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"k8s.io/klog/v2"
+
+	"github.com/kcp-dev/kcp/cmd/test-server/helpers"
+	"github.com/kcp-dev/kcp/test/e2e/framework"
+)
+
+func startVirtual(ctx context.Context, index int, logDirPath string) (<-chan error, error) {
+	logger := klog.FromContext(ctx)
+	logger.Info("starting virtual-workspaces standalone server", "index", index)
```
sttts, line 37: nit: not sure we do that elsewhere, but maybe print the command line with all parameters instead?

cmd/sharded-test-server/frontproxy.go
sttts: Now there is: https://github.com/kcp-dev/kcp/issues/1857

cmd/sharded-test-server/virtual.go
ncdc: https://github.com/kcp-dev/kcp/issues/1858

PR #1810, opened 2022-08-19
🌱 Part 17: Scope pkg/virtual

varshaprasad96:
## Related issue(s)

Fixes #

### DIFF
--- a/config/root-phase0/bootstrap.go
+++ b/config/root-phase0/bootstrap.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"embed"
 
+	"github.com/kcp-dev/logicalcluster/v2"
+
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
@@ -35,8 +37,8 @@ var fs embed.FS
 // Bootstrap creates resources in this package by continuously retrying the list.
 // This is blocking, i.e. it only returns (with error) when the context is closed or with nil when
 // the bootstrapping is successfully completed.
-func Bootstrap(ctx context.Context, kcpClient kcpclientset.Interface, rootDiscoveryClient discovery.DiscoveryInterface, rootDynamicClient dynamic.Interface, batteriesIncluded sets.String) error {
-	if err := confighelpers.BindRootAPIs(ctx, kcpClient, "shards.tenancy.kcp.dev", "tenancy.kcp.dev", "scheduling.kcp.dev", "workload.kcp.dev", "apiresource.kcp.dev"); err != nil {
+func Bootstrap(ctx context.Context, kcpClient kcpclientset.Interface, rootDiscoveryClient discovery.DiscoveryInterface, rootDynamicClient dynamic.Interface, batteriesIncluded sets.String, clusterName logicalcluster.Name) error {
+	if err := confighelpers.BindRootAPIs(logicalcluster.WithCluster(ctx, clusterName), kcpClient, "shards.tenancy.kcp.dev", "tenancy.kcp.dev", "scheduling.kcp.dev", "workload.kcp.dev", "apiresource.kcp.dev"); err != nil {
 		return err
 	}
 	return confighelpers.Bootstrap(ctx, rootDiscoveryClient, rootDynamicClient, batteriesIncluded, fs)
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -23,6 +23,7 @@ import (
 	_ "net/http/pprof"
 	"net/url"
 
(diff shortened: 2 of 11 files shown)


stevekuznetsov:
```
     virtualworkspace_test.go:234: 
        	Error Trace:	virtualworkspace_test.go:234
        	            	dynamic client does not implement ResourceDeleterInterface
        	Test:       	TestAPIExportVirtualWorkspace 
```

config/root-phase0/bootstrap.go
```diff
@@ -35,8 +37,8 @@ var fs embed.FS
 // Bootstrap creates resources in this package by continuously retrying the list.
 // This is blocking, i.e. it only returns (with error) when the context is closed or with nil when
 // the bootstrapping is successfully completed.
-func Bootstrap(ctx context.Context, kcpClient kcpclientset.Interface, rootDiscoveryClient discovery.DiscoveryInterface, rootDynamicClient dynamic.Interface, batteriesIncluded sets.String) error {
-	if err := confighelpers.BindRootAPIs(ctx, kcpClient, "shards.tenancy.kcp.dev", "tenancy.kcp.dev", "scheduling.kcp.dev", "workload.kcp.dev", "apiresource.kcp.dev"); err != nil {
+func Bootstrap(ctx context.Context, kcpClient kcpclientset.Interface, rootDiscoveryClient discovery.DiscoveryInterface, rootDynamicClient dynamic.Interface, batteriesIncluded sets.String, clusterName logicalcluster.Name) error {
+	if err := confighelpers.BindRootAPIs(logicalcluster.WithCluster(ctx, clusterName), kcpClient, "shards.tenancy.kcp.dev", "tenancy.kcp.dev", "scheduling.kcp.dev", "workload.kcp.dev", "apiresource.kcp.dev"); err != nil {
```
stevekuznetsov, line 41: If we are choosing to add this to the context at the "last" possible chance - when making the client calls - we should go ahead and pass the name through all the way. Or, we can a) attach it to the context earlier or b) scope the client before passing it into this bootstrap function.

pkg/server/config.go
```diff
@@ -224,12 +225,18 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
 		}
 		c.identityConfig, c.resolveIdentities = boostrap.NewConfigWithWildcardIdentities(c.GenericConfig.LoopbackClientConfig, boostrap.KcpRootGroupExportNames, boostrap.KcpRootGroupResourceExportNames, nonIdentityKcpClusterClient, nil)
 	}
-	c.KcpClusterClient, err = kcpclient.NewClusterForConfig(c.identityConfig) // this is now generic to be used for all kcp API groups
+	c.KcpClusterClient, err = kcpclient.NewForConfig(kcpclienthelper.NewClusterConfig(c.identityConfig)) // this is now generic to be used for all kcp API groups
```
stevekuznetsov, line 228: We've changed `NewClusterConfig` to not do a copy of the `*rest.Config`, so we will want to explicitly do that when calling the new helper.

config/root-phase0/bootstrap.go
varshaprasad96: The reason for passing it in `Bootstrap` was because we specifically want the context given to `BindRootAPIs` scoped, but not the ones to `confighelpers.Bootstrap()` (https://github.com/kcp-dev/kcp/pull/1810/files#diff-721469ea78064b0433a3e021a00e1b61ae2931a1840265d627d24301babca1e9R46) which is called next. That's why we can't scope it before passing it into bootstrap function. We can scope the client entirely, but that would mean modifying the host of rest config while creating client in `pkg/server/server.go` (https://github.com/kcp-dev/kcp/pull/1810/files#diff-104a11712684e520f0affa95e6a053faeb713306a10f69ee99b9729cc9dd1996R182). I feel that would unnecessarily make the code more messier by defining new kcp client for every Bootstrap function. That is the reason for passing the cluster name and scoping it while Bootstrap.

stevekuznetsov: Gotcha. Is it common that we scope the context early, as opposed to right in the client? If not, I worry that having just one or two places where we do that could be confusing. If that's true, could we perhaps pass it through then to the client calls themselves?

varshaprasad96: @stevekuznetsov I tried refactoring the code to scope the clients where they are called instead of doing it before. In this case its in `confighelpers.BindRootAPIs`. But looks like in certain cases the clients are not supposed to be scoped at all (ie) the context need not be passed with clusterNames. Passing an empty cluster (`logicalcluster.Name{}`) with the context is causing issues while bootstrapping, hence leaving it as is. I can add a comment mentioning that it is intentionally scoped here, is that fine?

varshaprasad96:
Closing this as decision on context based client scoping may not be required.

PR #1813, opened 2022-08-22
🌱 PROOF: use the new listergen+informergen

stevekuznetsov:
Proot for #https://github.com/kcp-dev/kubernetes/pull/89


stevekuznetsov:
FYI @ncdc @sttts we've got two options here - either duplicate the entire shared informer factory to allow us to cut over bit by bit, or do a whole-hog move right now. I'll keep working on the whole move, and we can figure out if it's too risky of a single PR soon.

PR #1815, opened 2022-08-23
cache-server: expose apiresourceschemas and apiexports

p0lyn0mial:
## Summary
exposes schema-less `apiresourceschemas` and `apiexports` resources. It allows us to store any fields we need. Since this server will be used by the shards, schemas will be implicit.

I did the following manual test which indicates storing `"foo": "bar"` with an `apiexport` obj works. 

POST:
```
curl -k -X POST 'https://localhost:6443/apis/apis.kcp.dev/v1alpha1/apiexports' -H 'Content-Type: application/json' -d '{"apiVersion":"apis.kcp.dev/v1alpha1","kind":"APIExport","metadata":{"name":"ab"},"foo":"bar"}'
```

GET:
```
curl -k 'https://localhost:6443/apis/apis.kcp.dev/v1alpha1/apiexports/ab'
{
  "apiVersion": "apis.kcp.dev/v1alpha1",
  "foo": "bar",
  "kind": "APIExport",
  "metadata": {
    "annotations": {
      "kcp.dev/cluster": "system:admin"
    },
    "creationTimestamp": "2022-08-24T08:50:09Z",
    "generation": 1,
    "managedFields": [
      {
        "apiVersion": "apis.kcp.dev/v1alpha1",
        "fieldsType": "FieldsV1",
        "fieldsV1": {
          "f:foo": {}
        },
        "manager": "curl",
        "operation": "Update",
        "time": "2022-08-24T08:50:09Z"
      }
    ],
    "name": "ab",
    "resourceVersion": "11",
    "uid": "b4b9447e-5ea8-4b73-a996-eb4a98eae3e0"
  }
}
```

note for now passing a logical cluster name is not required in the future it might be required along with a shard name.

## Related issue(s)

requires https://github.com/kcp-dev/kubernetes/pull/93

### DIFF
--- a/config/crds/bootstrap.go
+++ b/config/crds/bootstrap.go
@@ -38,6 +38,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
+	"sigs.k8s.io/yaml"
 )
 
 //go:embed *.yaml
@@ -174,6 +175,16 @@ func CreateSingle(ctx context.Context, client apiextensionsv1client.CustomResour
 
 		return crdhelpers.IsCRDConditionTrue(crd, apiextensionsv1.Established), nil
 	})
+}
+
+// Unmarshal YAML-decodes the give embedded file name into the target.
+func Unmarshal(fileName string, crd *apiextensionsv1.CustomResourceDefinition) error {
+	bs, err := raw.ReadFile(fileName)
+	if err != nil {
+		return err
+	}
+
+	return yaml.Unmarshal(bs, crd)
 }
 
 func retryRetryableErrors(f func() error) error {
--- a/pkg/cache/server/bootstrap/bootstrap.go
+++ b/pkg/cache/server/bootstrap/bootstrap.go
@@ -1,0 +1,69 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package bootstrap
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/kcp-dev/logicalcluster/v2"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
+	"k8s.io/utils/pointer"
+
+	configcrds "github.com/kcp-dev/kcp/config/crds"
+)
+
+// SystemCRDLogicalCluster holds a logical cluster name under which we store system-related CRDs.
+// We use the same name as the KCP for symmetry.
+var SystemCRDLogicalCluster = logicalcluster.New("system:system-crds")
+
+func Bootstrap(ctx context.Context, apiExtensionsClusterClient apiextensionsclient.ClusterInterface) error {
+	crds := []*apiextensionsv1.CustomResourceDefinition{}
+	for _, resource := range []string{"apiresourceschemas", "apiexports"} {
+		crd := &apiextensionsv1.CustomResourceDefinition{}
+		if err := configcrds.Unmarshal(fmt.Sprintf("apis.kcp.dev_%s.yaml", resource), crd); err != nil {
+			panic(fmt.Errorf("failed to unmarshal %v resource: %w", resource, err))
+		}
+		for i := range crd.Spec.Versions {
+			v := &crd.Spec.Versions[i]
+			v.Schema = &apiextensionsv1.CustomResourceValidation{
+				OpenAPIV3Schema: &apiextensionsv1.JSONSchemaProps{
+					Type:                   "object",
+					XPreserveUnknownFields: pointer.BoolPtr(true),
+				},
+			} // wipe the schema, we don't need validation
+		}
+		crds = append(crds, crd)
+	}
+
+	logger := klog.FromContext(ctx)
+	return wait.PollInfiniteWithContext(ctx, time.Second, func(ctx context.Context) (bool, error) {
+		for _, crd := range crds {
+			err := configcrds.CreateSingle(ctx, apiExtensionsClusterClient.Cluster(SystemCRDLogicalCluster).ApiextensionsV1().CustomResourceDefinitions(), crd)
+			if err != nil {
+				logger.Error(err, "failed to create CustomResourceDefinition", crd.Name)
+				return false, nil
+			}
+		}
+		return true, nil
+	})
+}
--- a/pkg/cache/server/config.go
+++ b/pkg/cache/server/config.go
@@ -20,10 +20,15 @@ import (
 	"errors"
 	"net/http"
 	"net/url"
+	"time"
+
+	"github.com/kcp-dev/logicalcluster/v2"
 
 	apiextentionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextentionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	apiextensionsapiserver "k8s.io/apiextensions-apiserver/pkg/apiserver"
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	apiextensionsexternalversions "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions"
 	apiextensionsoptions "k8s.io/apiextensions-apiserver/pkg/cmd/server/options"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -39,16 +44,28 @@ import (
 	kcpserver "github.com/kcp-dev/kcp/pkg/server"
 )
 
+const resyncPeriod = 10 * time.Hour
+
 type Config struct {
 	Options       *cacheserveroptions.CompletedOptions
 	ApiExtensions *apiextensionsapiserver.Config
 	EmbeddedEtcd  *embeddedetcd.Config
+
+	ExtraConfig
 }
 
 type completedConfig struct {
 	Options       *cacheserveroptions.CompletedOptions
 	ApiExtensions apiextensionsapiserver.CompletedConfig
 	EmbeddedEtcd  embeddedetcd.CompletedConfig
+
+	ExtraConfig
+}
+
+type ExtraConfig struct {
+	ApiExtensionsClusterClient apiextensionsclient.ClusterInterface
+
+	ApiExtensionsSharedInformerFactory apiextensionsexternalversions.SharedInformerFactory
 }
 
 type CompletedConfig struct {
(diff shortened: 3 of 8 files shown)


p0lyn0mial:
/hold

I still need to test it manually.

pkg/cache/server/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,68 @@
...
+					XPreserveUnknownFields: pointer.BoolPtr(true),
+				},
+			} // wipe the schema, we don't need validation
+		}
+		crds = append(crds, crd)
+	}
+
+	return wait.PollInfiniteWithContext(ctx, time.Second, func(ctx context.Context) (bool, error) {
+		for _, crd := range crds {
+			_, err := apiExtensionsClusterClient.Cluster(SystemCRDLogicalCluster).ApiextensionsV1().CustomResourceDefinitions().Create(ctx, crd, metav1.CreateOptions{})
```
p0lyn0mial, line 60: creating crds with empty schemas is cheap and doesn't require creating an `in-memory` storage implementation.

it can be reworked in the future if we don't like it.

p0lyn0mial: the storage is required for listing `CRDs` and for discovery (not implemented yet).

config/crds/bootstrap.go
```diff
@@ -176,6 +177,16 @@ func CreateSingle(ctx context.Context, client apiextensionsv1client.CustomResour
 	})
 }
 
+// Unmarshal YAML-decodes the give embedded file name into the target.
+func Unmarshal(fileName string, o interface{}) error {
```
stevekuznetsov, line 181: nit: whenever you are writing non-typesafe code it should not be done lightly. in my personal opinion, the two stdlib calls in this function are not complex enough to warrant a helper that exits the type system

pkg/cache/server/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,68 @@
...
+	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/utils/pointer"
+
+	configcrds "github.com/kcp-dev/kcp/config/crds"
+)
+
+var SystemCRDLogicalCluster = logicalcluster.New("system:system-crds")
```
stevekuznetsov, line 37: This is a constant somewhere already, could we import it?

pkg/cache/server/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,68 @@
...
+
+func Bootstrap(ctx context.Context, logger logr.Logger, apiExtensionsClusterClient apiextensionsclient.ClusterInterface) error {
```
stevekuznetsov, line 39: nit: the expectation is that the logger is *in* the context, and you take it out when you need it

pkg/cache/server/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,68 @@
...
+	crds := []*apiextensionsv1.CustomResourceDefinition{}
+	for _, resource := range []string{"apiresourceschemas", "apiexports"} {
+		crd := &apiextensionsv1.CustomResourceDefinition{}
+		if err := configcrds.Unmarshal(fmt.Sprintf("apis.kcp.dev_%s.yaml", resource), &crd); err != nil {
+			panic(fmt.Errorf("failed to unmarshal %v resource: %w", resource, err))
```
stevekuznetsov, line 44: why panic vs return an error?

pkg/cache/server/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+		for i := range crd.Spec.Versions {
+			v := &crd.Spec.Versions[i]
+			v.Schema = &apiextensionsv1.CustomResourceValidation{
+				OpenAPIV3Schema: &apiextensionsv1.JSONSchemaProps{
+					Type:                   "object",
+					XPreserveUnknownFields: pointer.BoolPtr(true),
+				},
+			} // wipe the schema, we don't need validation
```
stevekuznetsov, line 53: Is this the only diff from the other bootstrap functions? could we DRY this out?

pkg/cache/server/config.go
```diff
@@ -125,17 +143,31 @@ func NewConfig(opts *cacheserveroptions.CompletedOptions) (*Config, error) {
 	serverConfig.LoopbackClientConfig.DisableCompression = true
 	clientutils.EnableMultiCluster(serverConfig.LoopbackClientConfig, &serverConfig.Config, "namespaces", "apiservices", "customresourcedefinitions", "clusterroles", "clusterrolebindings", "roles", "rolebindings", "serviceaccounts", "secrets")
 
+	var err error
+	c.ApiExtensionsClusterClient, err = apiextensionsclient.NewClusterForConfig(serverConfig.LoopbackClientConfig)
```
stevekuznetsov, line 147: nit: we're moving to the context-based clients instead of `NewCluster` - can we do that here as well to save Varsha some work in the future? :)

pkg/cache/server/server.go
```diff
@@ -34,9 +37,47 @@ func NewServer(c CompletedConfig) (*Server, error) {
...
+	}); err != nil {
+		return err
+	}
 	return server.GenericAPIServer.PrepareRun().Run(ctx.Done())
 }
+
+// goContext turns the PostStartHookContext into a context.Context for use in routines that may or may not
+// run inside of a post-start-hook. The k8s APIServer wrote the post-start-hook context code before contexts
+// were part of the Go stdlib.
+func goContext(parent genericapiserver.PostStartHookContext) context.Context {
```
stevekuznetsov, line 76: We must have this as a helper somewhere - or if not, we've copy-pasted it 100 times. Could we DRY it out?

stevekuznetsov (commented):
Mostly seems reasonable, some small comments

pkg/cache/server/bootstrap/bootstrap.go
p0lyn0mial: I think we should treat it as a distinct value, I picked up the same logical cluster for symmetry.

pkg/cache/server/bootstrap/bootstrap.go
p0lyn0mial: the logger is modified before calling this function with some contextual data. I don't like the idea of putting it into ctx just to read it here.

pkg/cache/server/bootstrap/bootstrap.go
p0lyn0mial: indicates the programmer error, plus the err is not "handled" differently by the caller.

pkg/cache/server/server.go
p0lyn0mial: we could, I'll try to find a good place :)

config/crds/bootstrap.go
p0lyn0mial: I can change the signature to accept a concrete type.

pkg/cache/server/bootstrap/bootstrap.go
p0lyn0mial: thoughts?

pkg/cache/server/bootstrap/bootstrap.go
p0lyn0mial: we could call `CreateSingle` from this function and retry on any error.

pkg/cache/server/bootstrap/bootstrap.go
stevekuznetsov: A comment to explain that would be great! Makes sense.

pkg/cache/server/bootstrap/bootstrap.go
stevekuznetsov: I agree with you in general and I think passing a logger is not that onerous, but k8s has disagreed with both of us and I think conforming to their expectations (least surprise) would likely benefit us

pkg/cache/server/config.go
stevekuznetsov: Look for any use of `logicalcluster.WithCluster()` e.g.:

```go
	return c.kubeClusterClient.CoreV1().Namespaces().Patch(logicalcluster.WithCluster(ctx, clusterName), name, pt, data, opts, subresources...)
```

pkg/server/server.go
```diff
@@ -473,10 +473,10 @@ func (s *Server) Run(ctx context.Context) error {
 	return delegationChainHead.PrepareRun().Run(ctx.Done())
 }
 
-// goContext turns the PostStartHookContext into a context.Context for use in routines that may or may not
+// GoContext turns the PostStartHookContext into a context.Context for use in routines that may or may not
 // run inside of a post-start-hook. The k8s APIServer wrote the post-start-hook context code before contexts
 // were part of the Go stdlib.
-func goContext(parent genericapiserver.PostStartHookContext) context.Context {
+func GoContext(parent genericapiserver.PostStartHookContext) context.Context {
```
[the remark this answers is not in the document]
stevekuznetsov: You and @sttts both ... but I like copy-pasting something 100x even less than I like creating a utility package ;)

pkg/cache/server/config.go
p0lyn0mial: okay, but I need a `Wildcard` logical cluster for informers, it seems like informers don't support the new way of passing a cluster name.

pkg/server/server.go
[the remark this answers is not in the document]
p0lyn0mial: moved to a util pkg

pkg/cache/server/config.go
stevekuznetsov: Yes, that is the one case where this approach does not work

stevekuznetsov: #1813 fixes that :)

p0lyn0mial: do we need `2` clients until it merges?

stevekuznetsov: Not sure I follow actually - you need a cluster-scoped client for the factory:

```go
# github.com/kcp-dev/kcp/pkg/client/informers
func NewSharedInformerFactoryWithOptions( ... versioned.Interface, ... ) {}
```

Where you can get such a thing from `.Cluster(logicalcluster.Wildcard)` which should be identical to `SetCluster()` on the `*rest.Config`, no?

```go
# github.com/kcp-dev/kcp/pkg/client/clientset/versioned
func (c *Cluster) Cluster(name v2.Name) Interface {}
```


So `NewClusterForConfig(cfg).Cluster(logicalcluster.Wildcard)` would become `NewClientForConfig(kcpclienthelper.SetCluster(rest.CopyConfig(cfg), logicalcluster.Wildcard))`?

p0lyn0mial: something is broken, I think I have to use it along with `NewClusterRoundTripper` but then the path seems to be incorrect `/clusters/*/clusters/system:system-crds/apis/apiextensions.k8s.io/v1/customresourcedefinitions/apiresourceschemas.apis.kcp.dev`

I'd like to proceed with this PR if that's okay. I can revisit using the context-based clients in the future

ncdc:
@p0lyn0mial if you wouldn't mind, please prefix the PR titles with the area they're for (it makes it easier to know at a glance) - thanks!

/retitle cache-server: expose apiresourceschemas and apiexports

pkg/util/util.go
```diff
@@ -0,0 +1,35 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"context"
+
+	genericapiserver "k8s.io/apiserver/pkg/server"
```
stevekuznetsov, line 22: nit: If I had a crystal ball, I think I'd make the prophecy that Stefan would want this helper to end up in this upstream package ... :^)

stevekuznetsov (approved):
Other comments can/will be follow-ups.

pkg/server/server.go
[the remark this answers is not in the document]
sttts: util packages 😱

pkg/util/util.go
sttts: How about moving that into the upstream package?

stevekuznetsov:

PR #1753, opened 2022-08-16
🌱 Part 16: Scope kubeclient in pkg/virtual/apiexport

varshaprasad96:
## Related issue(s)

Fixes #

PR #1819, opened 2022-08-23
Part 19: Scoped cmd/front-proxy

varshaprasad96:
## Related issue(s)

Fixes #

### DIFF
--- a/cmd/virtual-workspaces/command/cmd.go
+++ b/cmd/virtual-workspaces/command/cmd.go
@@ -95,11 +95,7 @@ func Run(ctx context.Context, o *options.Options) error {
 	nonIdentityConfig.Host = u.String()
 
 	// resolve identities for system APIBindings
-	nonIdentityKcpClusterClient, err := kcpclient.NewClusterForConfig(nonIdentityConfig) // can only used for wildcard requests of apis.kcp.dev
-	if err != nil {
-		return err
-	}
-	identityConfig, resolveIdentities := boostrap.NewConfigWithWildcardIdentities(nonIdentityConfig, boostrap.KcpRootGroupExportNames, boostrap.KcpRootGroupResourceExportNames, nonIdentityKcpClusterClient, nil)
+	identityConfig, resolveIdentities := boostrap.NewConfigWithWildcardIdentities(nonIdentityConfig, boostrap.KcpRootGroupExportNames, boostrap.KcpRootGroupResourceExportNames, nil)
 	if err := wait.PollImmediateInfiniteWithContext(ctx, time.Millisecond*500, func(ctx context.Context) (bool, error) {
 		if err := resolveIdentities(ctx); err != nil {
 			klog.V(3).Infof("failed to resolve identities, keeping trying: %v", err)
--- a/pkg/server/bootstrap/identity.go
+++ b/pkg/server/bootstrap/identity.go
@@ -26,6 +26,8 @@ import (
 	"strings"
 	"sync"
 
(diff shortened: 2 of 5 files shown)


cmd/virtual-workspaces/command/cmd.go
```diff
@@ -95,7 +98,7 @@ func Run(ctx context.Context, o *options.Options) error {
 	nonIdentityConfig.Host = u.String()
 
 	// resolve identities for system APIBindings
-	nonIdentityKcpClusterClient, err := kcpclient.NewClusterForConfig(nonIdentityConfig) // can only used for wildcard requests of apis.kcp.dev
+	nonIdentityKcpClusterClient, err := kcpclient.NewForConfig(kcpclienthelper.SetCluster(rest.CopyConfig(nonIdentityConfig), tenancyv1alpha1.RootCluster)) // can only used for wildcard requests of apis.kcp.dev
```
stevekuznetsov, line 101: Was the root cluster implicit before?

pkg/server/bootstrap/identity.go
```diff
@@ -120,7 +120,7 @@ func NewWildcardIdentitiesWrappingRoundTripper(groupExportNames map[string]strin
 		ids.groupResourceIdentities[gr] = ""
 	}
 
-	return injectKcpIdentities(ids), wildcardIdentitiesResolver(ids, groupExportNames, groupResourceExportNames, apiExportIdentityProvider(rootShardKcpClient.Cluster(tenancyv1alpha1.RootCluster), localShardKubeClusterClient))
```
stevekuznetsov, line 123: I'd prefer that we continue to choose *where* we scope the client - as written, after this change the injection will happen with whatever client is passed in, even though it's only ever valid to inject with the root client. Let's perhaps pass the `*rest.Config` here and create the root client here, as well, to continue to provide that behavior.

stevekuznetsov (commented):
Needs `make imports`

pkg/server/bootstrap/identity.go
varshaprasad96: The client being used here is only scoped to root cluster. This change involves passing a root cluster scoped client instead of an unscoped client to bootstrap. Just wondering, what will be the value of passing config here? If we do so, this would mean that instead of passing clients around to controllers we would be passing configs at places where scoping is possible.

cmd/virtual-workspaces/command/cmd.go
varshaprasad96: It was being scoped later in the controller. Modified this to scope it here, since there is no other use of unscoped client .

pkg/server/bootstrap/identity.go
```diff
@@ -198,8 +199,14 @@ func wildcardIdentitiesResolver(ids *identities,
 	}
 }
 
-func apiExportIdentityProvider(rootShardKcpClient kcpclient.Interface, localShardKubeClusterClient kubernetes.ClusterInterface) func(ctx context.Context, apiExportName string) (string, error) {
+func apiExportIdentityProvider(config *rest.Config, localShardKubeClusterClient kubernetes.ClusterInterface) func(ctx context.Context, apiExportName string) (string, error) {
```
stevekuznetsov, line 202: BTW - any reason we're not getting the `kubernetes.ClusterInterface` here?

varshaprasad96: Need https://github.com/kcp-dev/kcp/pull/1753 to get in to scope this.

varshaprasad96: Will do it a follow up after its merged.

PR #1825, opened 2022-08-24
Fix GH issue template labels

ncdc:


ncdc:
@stevekuznetsov does this look correct (the current templates at least aren't setting bug/feature correctly)?

stevekuznetsov (approved):
I think? Let's try it out

PR #1846, opened 2022-08-26
Cluster scoped quota

ncdc:
## Summary
Support quota of cluster-scoped things.

To use:
1. Create a namespace called `admin`
2. Create a ResourceQuota in `admin` with the following annotation: `"experimental.quota.kcp.dev/cluster-scoped": "true"`
3. You can now specify `count/<resource>` of both namespace- and cluster-scoped resources
    1. For namespace-scoped resources, it counts instances across all namespaces in the workspace
    2. For cluster-scoped resources, it counts instances in the workspace 

## Related issue(s)

Requires: https://github.com/kcp-dev/kubernetes/pull/97
Part of: #1061


pkg/reconciler/kubequota/scoped_generic_shared_informer_factory.go
```diff
@@ -192,8 +192,19 @@ type scopedGenericNamespaceLister struct {
 // List lists all instances from the cache.Indexer scoped to a single logical cluster and namespace, and matching
 // selector.
 func (s *scopedGenericNamespaceLister) List(selector labels.Selector) (ret []runtime.Object, err error) {
```
stevekuznetsov, line 194: Who else uses this?

stevekuznetsov (commented):
If I have a cluster-wide quota for configmaps of 5 and a namespace-specific quota of 3, will this correctly fail if I attempt to create a 4th configmap in one namespace? What determines quota precedence when multiple apply?

ncdc:
> If I have a cluster-wide quota for configmaps of 5 and a namespace-specific quota of 3, will this correctly fail if I attempt to create a 4th configmap in one namespace? What determines quota precedence when multiple apply?

I'm pretty sure it fails if any matching ResourceQuotas fail

pkg/reconciler/kubequota/scoped_generic_shared_informer_factory.go
ncdc: Nobody yet, but I could see us using it for GC

PR #1853, opened 2022-08-30
cache-server: provides client-related functionality for dealing with a shard on the HTTP level

p0lyn0mial:
## Summary
`WithDefaultShardRoundTripper` for setting a default shard in the context.
`WithShardRoundTripper` for changing the URL path to contain a shard name defined in the context.

`WithShardInContext` and `ShardFromContext` for writting/reading a shard name to/from the context.

A shard package for holding a shard name in the memory.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cache/client/context.go
+++ b/pkg/cache/client/context.go
@@ -1,0 +1,41 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package client
+
+import (
+	"context"
+
+	"github.com/kcp-dev/kcp/pkg/cache/client/shard"
+)
+
+type shardKey int
+
+const (
+	// shardContextKey is the context key for a request
+	shardContextKey shardKey = iota
+)
+
+// WithShardInContext returns a context with the given shard set
+func WithShardInContext(parent context.Context, shard shard.Name) context.Context {
+	return context.WithValue(parent, shardContextKey, shard)
+}
+
+// ShardFromContext returns the value of the shard key on the ctx,
+// or an empty Name if there is no shard key.
+func ShardFromContext(ctx context.Context) shard.Name {
+	return ctx.Value(shardContextKey).(shard.Name)
+}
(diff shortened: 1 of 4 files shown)


pkg/cache/client/context.go
```diff
@@ -0,0 +1,42 @@
...
+	shardContextKey shardKey = iota
+)
+
+// WithShardInContext returns a context with the given shard set
+func WithShardInContext(parent context.Context, shard shard.Name) context.Context {
+	return context.WithValue(parent, shardContextKey, shard)
+}
+
+// ShardFromContext returns the value of the shard key on the ctx,
+// or an empty Name if there is no shard key.
```
stevekuznetsov, line 38: If we can compare to the empty sentinel value, why do we need to return `ok`?

stevekuznetsov (commented):
One question, otherwise LGTM

pkg/cache/client/context.go
p0lyn0mial: good point, will update

p0lyn0mial:
@stevekuznetsov I slightly changed the PR, PTAL

pkg/cache/client/context.go
```diff
@@ -0,0 +1,45 @@
...
+func ShardFromContext(ctx context.Context) shard.Name {
+	val, ok := ctx.Value(shardContextKey).(shard.Name)
```
stevekuznetsov, line 40: nit: can be a follow-up: can just `return ctx.Value(shardContextKey).(shard.Name)` since 

a) the `shardContextKey` is private to this package, so the only case where `!ok` is if someone were to write code _in this package_ to add something invalid
b) the zero-value is already `shard.New("")`

pkg/cache/client/round_tripper.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// generatePath formats the request path to target the specified shard
+func generatePath(originalPath string, shard clientshard.Name) (string, error) {
+	// if the originalPath already has the shard then the path was already modified and no change needed
+	if strings.Contains(originalPath, shard.Path()) {
+		return originalPath, nil
+	}
+	// if the originalPath already has a shard set just overwrite it to the given one
+	if strings.Contains(originalPath, "/shards") {
```
stevekuznetsov, line 91: Is this specifically not doing a prefix check for some reason?

stevekuznetsov (approved):
Sorry, one last question - feel free to cancel hold

/lgtm
/approve
/hold

pkg/cache/client/context.go
p0lyn0mial: yes, I just don't like when the code panics but sure I can change it.

pkg/cache/client/round_tripper.go
p0lyn0mial: as of today, we could actually check for a prefix, maybe in the future, it will have to be changed.

pkg/cache/client/context.go
p0lyn0mial: done.

pkg/cache/client/round_tripper.go
p0lyn0mial: done

p0lyn0mial:
@stevekuznetsov updated the PR, PTAL

/hold cancel

PR #1854, opened 2022-08-30
virtual: Support SSA patch requests for non-existent objects

### LINKED ISSUE #1852: bug: Server-side apply requests to virtual workspace API server fail for non-existent objects
### Describe the bug

A SSA patch request to the APIExport virtual workspace API server returns 404 when the corresponding object does not already exists.

### Steps To Reproduce

```golang
config := corev1apply.ConfigMap("config", "default").WithData(map[string]string{})
err = client.Patch(ctx, config, ctrl.Apply, ctrl.ForceOwnership, ctrl.FieldOwner("manager"))
```

### Expected Behaviour

SSA patch requests should succeed even when objects do not already exist.

### Additional Context

_No response_

astefanutti:
## Summary

This PR switches update operations made to the virtual workspace API server, to calling create operations on the forwarding API server, for non-existent objects. This enables support for server-side apply patch requests in that case.

## Related issue(s)

Fixes #1852

### DIFF
--- a/pkg/virtual/framework/forwardingregistry/store.go
+++ b/pkg/virtual/framework/forwardingregistry/store.go
@@ -23,6 +23,8 @@ import (
 
 	"github.com/kcp-dev/logicalcluster/v2"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metainternalversion "k8s.io/apimachinery/pkg/apis/meta/internalversion"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -176,10 +178,17 @@ func DefaultDynamicDelegatedStoreFuncs(
 			return nil, false, err
 		}
 
+		requestInfo, _ := genericapirequest.RequestInfoFrom(ctx)
+
 		doUpdate := func() (*unstructured.Unstructured, error) {
 			oldObj, err := s.Get(ctx, name, &metav1.GetOptions{})
 			if err != nil {
-				return nil, err
+				// Continue on 404 for PATCH requests, so server-side apply requests
+				// for non-existent objects can still be processed.
+				if !(requestInfo != nil && requestInfo.Verb == "patch" && apierrors.IsNotFound(err)) {
+					return nil, err
+				}
+				oldObj = nil
 			}
 
 			obj, err := objInfo.UpdatedObject(ctx, oldObj)
@@ -192,10 +201,17 @@ func DefaultDynamicDelegatedStoreFuncs(
 				return nil, fmt.Errorf("not an Unstructured: %T", obj)
 			}
 
+			// If the updated object does not have a CreationTimestamp, it means it doesn't currently exist.
+			// In that case, we switch to calling a create operation on the forwarding registry.
+			// This enables support for server-side apply requests, to create non-existent objects.
+			if exists, err := hasCreationTimestamp(unstructuredObj); err != nil {
+				return nil, err
+			} else if !exists {
+				return delegate.Create(ctx, unstructuredObj, updateToCreateOptions(options), subResources...)
+			}
 			return delegate.Update(ctx, unstructuredObj, *options, subResources...)
 		}
 
-		requestInfo, _ := genericapirequest.RequestInfoFrom(ctx)
 		if requestInfo != nil && requestInfo.Verb == "patch" {
 			var result *unstructured.Unstructured
 			err := retry.RetryOnConflict(patchConflictRetryBackoff, func() error {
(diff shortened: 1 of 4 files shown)


pkg/virtual/framework/forwardingregistry/store.go
```diff
@@ -273,3 +290,29 @@ func clientGetter(dynamicClusterClient dynamic.ClusterInterface, namespaceScoped
...
+}
+
+// hasUID return whether the provided object has a UID.
+func hasUID(obj runtime.Object) (bool, error) {
+	if obj == nil {
+		return false, nil
+	}
+	accessor, err := meta.Accessor(obj)
+	if err != nil {
+		return false, apierrors.NewInternalError(err)
```
sttts, line 312: this is bad style: better move the `NewInternalError` wrapping to the top-level, if it is not already there.

pkg/virtual/framework/forwardingregistry/store.go
```diff
@@ -192,7 +200,16 @@ func DefaultDynamicDelegatedStoreFuncs(
...
 
-			return delegate.Update(ctx, unstructuredObj, *options, subResources...)
+			// If the updated object does not have a UID, it means it doesn't currently exist.
+			// In that case, we switch to calling a create operation on the forwarding registry.
+			// This enables support for server-side apply requests, to create non-existent objects.
+			if hasUID, err := hasUID(unstructuredObj); err != nil {
+				return nil, err
+			} else if !hasUID {
+				return delegate.Create(ctx, unstructuredObj, updateToCreateOptions(options), subResources...)
+			} else {
```
sttts, line 210: nit: drop last else

pkg/virtual/framework/forwardingregistry/store.go
```diff
@@ -273,3 +290,29 @@ func clientGetter(dynamicClusterClient dynamic.ClusterInterface, namespaceScoped
...
 	}
 }
+
+// updateToCreateOptions creates a CreateOptions with the same field values as the provided PatchOptions.
+func updateToCreateOptions(uo *metav1.UpdateOptions) metav1.CreateOptions {
+	co := metav1.CreateOptions{
+		DryRun:          uo.DryRun,
+		FieldManager:    uo.FieldManager,
+		FieldValidation: uo.FieldValidation,
+	}
```
sttts, line 300: this will bite us when UpdateOptiosn is extended. How can we catch that case, e.g. in unit tests. Even some reflection based check would be fine that UpdateOptions has not gotten new fields.

pkg/virtual/framework/forwardingregistry/store.go
```diff
@@ -179,7 +181,13 @@ func DefaultDynamicDelegatedStoreFuncs(
...
 			oldObj, err := s.Get(ctx, name, &metav1.GetOptions{})
 			if err != nil {
-				return nil, err
+				if !apierrors.IsNotFound(err) {
+					return nil, err
+				}
+				// This is required to make the call to UpdatedObjectInfo.UpdatedObject pass.
+				oldObj = &unstructured.Unstructured{
+					Object: make(map[string]interface{}),
+				}
```
sttts, line 190: worth an e2e test?

pkg/virtual/framework/forwardingregistry/store.go
astefanutti: done

pkg/virtual/framework/forwardingregistry/store.go
astefanutti: done

pkg/virtual/framework/forwardingregistry/store.go
astefanutti: Definitely, working on it 👍🏼

astefanutti:
Added unit and e2e tests. Also added support for the `forceAllowCreate` argument to update operation.

astefanutti:
@sttts @davidfestal PTAL.

pkg/virtual/framework/forwardingregistry/store.go
astefanutti: Added with 987273db6243279cb9ffa2b6d8f55dea57a0a78d.

pkg/virtual/framework/forwardingregistry/store.go
```diff
@@ -176,10 +177,20 @@ func DefaultDynamicDelegatedStoreFuncs(
...
 		doUpdate := func() (*unstructured.Unstructured, error) {
 			oldObj, err := s.Get(ctx, name, &metav1.GetOptions{})
 			if err != nil {
-				return nil, err
+				// Continue on 404 when forceAllowCreate is enabled,
+				// or for PATCH requests, so server-side apply requests
+				// for non-existent objects can still be processed.
+				if !apierrors.IsNotFound(err) ||
+					!forceAllowCreate &&
+						!(requestInfo != nil && requestInfo.Verb == "patch") {
```
davidfestal, line 190: What about `patch` requests which are not SSA requests ?
Will they try to create it as well ?

astefanutti: For JSON or SMP patch requests, the corresponding patcher implementation returns a `errors.NewNotFound` error, so the response code is 404 as expected. The patchers execute as part of the `objInfo.UpdatedObject(ctx, oldObj)` call, and their implementations can be found at https://github.com/kubernetes/apiserver/blob/48405445f09218bb8b7a8a81eb21027cfd65761b/pkg/endpoints/handlers/patch.go, with the errors at:
- https://github.com/kubernetes/apiserver/blob/48405445f09218bb8b7a8a81eb21027cfd65761b/pkg/endpoints/handlers/patch.go#L355
- https://github.com/kubernetes/apiserver/blob/48405445f09218bb8b7a8a81eb21027cfd65761b/pkg/endpoints/handlers/patch.go#L451

davidfestal: Thanks for the explanation.
Do you think you could add a comment in the code to explain this ? It's not easy to think of or guess when reading.

astefanutti: Sure, I've added eb9e64dba2c064ebce3578c08ad8c2105ef002d3 that hopefully covers it.

I've also pushed 6d1b96804c3b9efb1a69cc49597fb6cfba6b4647 to have the use case covered in e2e tests.

PR #1871, opened 2022-08-31
🌱 WithClusterScope: should not assign a default cluster name

p0lyn0mial:
## Summary
clients should specify a cluster name before storing something in the db as that leads to more readable code.
the server should not assign a default cluster name.

at the moment requests without a cluster name are rejected by the storage layer.

Things to consider:
reject a request without a name in the filter
will this break some external clients?

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/server/filters/filters.go
+++ b/pkg/server/filters/filters.go
@@ -35,7 +35,6 @@ import (
 	kaudit "k8s.io/apiserver/pkg/audit"
 	"k8s.io/apiserver/pkg/endpoints/handlers/responsewriters"
 	"k8s.io/apiserver/pkg/endpoints/request"
-	"k8s.io/kubernetes/pkg/genericcontrolplane"
 )
 
 type (
@@ -120,9 +119,7 @@ func WithClusterScope(apiHandler http.Handler) http.HandlerFunc {
 			cluster.Wildcard = true
 			// fallthrough
 			cluster.Name = logicalcluster.Wildcard
-		case clusterName.Empty():
-			cluster.Name = genericcontrolplane.LocalAdminCluster
-		default:
+		case !clusterName.Empty():
 			if !reClusterName.MatchString(clusterName.String()) {
 				responsewriters.ErrorNegotiated(
 					apierrors.NewBadRequest(fmt.Sprintf("invalid cluster: %q does not match the regex", clusterName)),
@@ -134,7 +131,6 @@ func WithClusterScope(apiHandler http.Handler) http.HandlerFunc {
 		}
 
 		ctx := request.WithCluster(req.Context(), cluster)
-
 		apiHandler.ServeHTTP(w, req.WithContext(ctx))
 	}
 }
--- a/pkg/server/home_workspaces.go
+++ b/pkg/server/home_workspaces.go
@@ -270,9 +270,11 @@ func (h *homeWorkspaceHandler) ServeHTTP(rw http.ResponseWriter, req *http.Reque
 
 	ctx := req.Context()
 	logger := klog.FromContext(ctx)
-	lcluster, err := request.ValidClusterFrom(ctx)
-	if err != nil {
-		responsewriters.InternalError(rw, req, err)
+	lcluster := request.ClusterFrom(req.Context())
+	if lcluster == nil {
+		// this is not a home workspace request
+		// just pass it to the next handler
+		h.apiHandler.ServeHTTP(rw, req)
 		return
 	}
 	logger = logging.WithCluster(logger, lcluster)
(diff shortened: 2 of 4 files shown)


stevekuznetsov (commented):
This is one of those super ancient decisions that I don't actually have any memory for why we did it. Hard to think through the implications.

@davidfestal thoughts?

ncdc:
IIRC it was to get "unaware" clients (e.g. kube internals) to succeed in their work (e.g. to create default RBAC resources) in some logical cluster.

I'm in favor of proceeding to remove the defaulting and trying to fix any remaining clients that were relying on this default.

p0lyn0mial:
> IIRC it was to get "unaware" clients (e.g. kube internals) to succeed in their work (e.g. to create default RBAC resources) in some logical cluster.

I think it was achieved by providing `multiClusterClientConfigRoundTripper`.

ncdc:
```
E0901 06:16:38.200882   38061 errors.go:77] cluster name is empty in the request context - RequestInfo: &request.RequestInfo{IsResourceRequest:false, Path:"/livez", Verb:"get", APIPrefix:"", APIGroup:"", APIVersion:"", Namespace:"", Resource:"", Subresource:"", Name:"", Parts:[]string(nil)}
```

stevekuznetsov:
(and we've since learned that of course unaware clients doing "something" without being multi-cluster-aware just do nothing important whatsoever in the root logical cluster ... and no importance in supporting that)

pkg/server/home_workspaces.go
```diff
@@ -272,7 +272,7 @@ func (h *homeWorkspaceHandler) ServeHTTP(rw http.ResponseWriter, req *http.Reque
 	logger := klog.FromContext(ctx)
 	lcluster, err := request.ValidClusterFrom(ctx)
 	if err != nil {
-		responsewriters.InternalError(rw, req, err)
+		h.apiHandler.ServeHTTP(rw, req)
```
p0lyn0mial, line 275: it looks like the `home-ws` is first in the chain, capturing `healthz/readyz` as well.

p0lyn0mial:
I almost have it, a few unit tests require fixing.

test/e2e/virtual/workspaces/virtual_workspace_test.go
```diff
@@ -584,6 +584,7 @@ func testWorkspacesVirtualWorkspaces(t *testing.T, standalone bool) {
 		// write kubeconfig to disk, next to kcp kubeconfig
 		kcpAdminConfig, _ := server.RawConfig()
 		var baseCluster = *kcpAdminConfig.Clusters["base"] // shallow copy
+		baseCluster.Server = fmt.Sprintf("%s/clusters/system:admin", baseCluster.Server)
```
p0lyn0mial, line 587: alternatively, we could add something to a handler but having a cluster name here is much cleaner.

sttts:
How is this going? Any blocker?

p0lyn0mial:
> How is this going? Any blocker?

I need to fix a few failing unit tests and this is ready to go.

pkg/server/home_workspaces.go
```diff
@@ -270,9 +270,11 @@ func (h *homeWorkspaceHandler) ServeHTTP(rw http.ResponseWriter, req *http.Reque
 
 	ctx := req.Context()
 	logger := klog.FromContext(ctx)
-	lcluster, err := request.ValidClusterFrom(ctx)
-	if err != nil {
-		responsewriters.InternalError(rw, req, err)
+	lcluster := request.ClusterFrom(req.Context())
```
p0lyn0mial, line 273: I'll follow up with a PR that removes usage of `ValidClusterFrom` from the authorizers

p0lyn0mial:
okay `ci/prow/test` is green, PTAL.

pkg/server/filters/filters.go
```diff
@@ -121,7 +120,7 @@ func WithClusterScope(apiHandler http.Handler) http.HandlerFunc {
 			// fallthrough
 			cluster.Name = logicalcluster.Wildcard
 		case clusterName.Empty():
-			cluster.Name = genericcontrolplane.LocalAdminCluster
```
ncdc, line 124: if we get rid of `cluster.Wildcard`, i don't even think we need the logic in here. instead:

```go
if clusterName != logicalcluster.Wildcard {
  // do regex check
}
cluster.Name = clusterName
```

ncdc: can do in a follow-up if you want

p0lyn0mial: Didn't know we were planning to get rid of `cluster.Wildcard`. I can do that in a follow-up. 
Meanwhile, I removed the case when the `clusterName` was empty. We will make use of the default value.

stevekuznetsov: Yeah I think the same arguments used to not have a .Wildcard for shard name apply here

ncdc (approved):

PR #1872, opened 2022-08-31
home-ws: wait for tenancy binding readiness

### LINKED ISSUE #1693: Home workspace creation failure
In https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kcp-dev_kcp/1673/pull-ci-kcp-dev-kcp-main-e2e-multiple-runs/1554909491965726720, we saw the attempt to create a child workspace in a not-yet-created home workspace hierarchy fail. This test usually succeeds, so it's presumably some sort of issue where informers are slow to catch up, and we are in an unintentional permanent failure state.

The workspaces VW code that handles requests to create v1beta1 workspaces got a 404 when trying to create the underlying v1alpha1 clusterworkspace. The only way that should happen is if the apibinding in the parent workspace isn't ready, I think.

#1692 adds more logging to try to triage further if we see the test failure again

ncdc:
## Summary
If the incoming request is in the home workspaces hierarchy and it's for `clusterworkspaces`, there is a chance the tenancy APIBinding is not yet ready (or the cache is stale). If that is the case, return a retry-after to give the APIBinding cache a
chance to catch up.

## Related issue(s)

Fixes #1693

### DIFF
--- a/pkg/server/home_workspaces.go
+++ b/pkg/server/home_workspaces.go
@@ -49,6 +49,7 @@ import (
 	"k8s.io/klog/v2"
 
 	clusterworkspaceadmission "github.com/kcp-dev/kcp/pkg/admission/clusterworkspace"
+	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
 	"github.com/kcp-dev/kcp/pkg/apis/tenancy/projection"
 	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
 	tenancyv1beta1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1beta1"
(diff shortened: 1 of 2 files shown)


pkg/server/home_workspaces.go
```diff
@@ -351,6 +378,21 @@ func (h *homeWorkspaceHandler) ServeHTTP(rw http.ResponseWriter, req *http.Reque
 				http.Error(rw, "Creating the home workspace", http.StatusTooManyRequests)
 				return
 			}
+
+			if requestInfo.IsResourceRequest && requestInfo.APIGroup == tenancyv1alpha1.SchemeGroupVersion.Group && requestInfo.Resource == "clusterworkspaces" {
```
stevekuznetsov, line 382: This kind of breaks my brain - can we close over this in `searchForWorkspaceAndRBACInLocalInformers`? Otherwise the flow here is pretty hard to follow (and we need to discern specifics about what might or might be true based on what some other helper returned)

ncdc: I don't have the `requestInfo` in `searchForWorkspaceAndRBACInLocalInformers` and it would be a huge amount of work to fix all the test cases that would be broken by adding a param to that function. That's why I did it here.

Issue #1219, opened 2022-06-06
Epic: API Export Permissions on Binding


shawn-hurley:
# Demo Objective
- [x] Service provider should be able to ask for additional permissions when creating an API Export for when the API is bound
- [x] API Binding users must accept or deny the specific permissions when binding to the APIExport
- [x] When accepted and bound, the objects granted access should be projected into the APIExports virtual workspace.

# Demo Steps
1. Service Provider creates an APIExport but asks for permissions to a secrets
1. User binds to the API and accepts the permissions
1. Service provider can access the secret from the virtual workspace.

# Stories
 
- [x] #1221 
- [x] #1244
- [x] When accepting the permissions, the virtual workspace should now have access to watch/list those resources.

------ 0.6 cut line

- [x] #1632

------- 0.7 cut line

- [x] #1544
- [x] #1500

------- 0.7.1 cut line

- [x] Add e2e test cases https://github.com/kcp-dev/kcp/pull/1864
    - When an APIExport removes a permission claim, its apiexport virtual workspace no longer serves that resource
    - When an APIBinding changes an accepted claim to rejected, the apiexport virtual workspace no longer serves that resource (was: #1501)
    - When an APIBinding has not accepted a claim (it's neither accepted nor rejected), the apiexport virtual workspace does not include resources from the binding's workspace
- [x] 💣🔒 **APIExport VW should restrict for a given workspace based on bindings acceptance or rejection** fixed in main
- [x] **Pending/approval life-cycle in APIBindings**
  - [x] add `binding.status.exportPermissionClaims`

------- 0.8 cut line

- [ ] 💣🔒 **APIExport VW for a claim should not override MaximalPermissionPolicy of the owning APIExport of the resource (#1337)** @s-urbaniak 
    - [ ] @ncdc has an e2e ready to go for this, but it will fail until we fix it
- [ ] warning on entering a workspace if there are claims from an APIExport that have not yet been accepted/rejected in an APIBinding (inside of kubectl kcp) @nrb  https://github.com/kcp-dev/kcp/issues/1934

-------- MVP complete – testing-ready --------

  - [ ] `kubectl kcp apis bind | accept | reject` – sketch out a command @dinhxuanvu https://github.com/kcp-dev/kcp/issues/1935
- [ ] 🔒 **APIExport can add authorization and limit what a user can do in their own namespace using permission claims.** https://github.com/kcp-dev/kcp/issues/1840
[...]

Issue #1786, opened 2022-08-17
Confusing error message when trying to create an APIBinding that points to a workspace whose top-level workspace does not exist


ncdc:
If you create an APIBinding with `.spec.reference.workspace.path` to a non-existent top-level workspace (or descendent thereof), you get the following error:

```
apibindings.apis.kcp.dev "foo" is forbidden: unable to create APIImport: missing verb='bind' permission on apiexports
```

This comes from https://github.com/kcp-dev/kcp/blob/1b6cde3443ae9fc2710cd4680258dd30ec8c4eb1/pkg/admission/apibinding/apibinding_admission.go#L203

We are not currently including the reason from the call to `authz.Authorize` on L200.

@stevekuznetsov & I hacked my kcp to include the reason, which ends up being

```
workspace access not permitted
```

This comes from https://github.com/kcp-dev/kcp/blob/1b6cde3443ae9fc2710cd4680258dd30ec8c4eb1/pkg/authorization/toplevel_org_authorizer.go#L142

The top-level workspace doesn't exist, and we return `workspace access not permitted` with no additional details to the user.

Compare this with Kubernetes:
1. when you do have permission to perform the operation, but e.g. the namespace doesn't exist:
```
$ k -n asdf create cm bob
error: failed to create configmap: namespaces "asdf" not found
```
2. when you don't have permission:
```
$ k --token user-1-token -n asdf create cm bob
error: failed to create configmap: configmaps is forbidden: User "user-1" cannot create resource "configmaps" in API group "" in the namespace "asdf"
```

@stevekuznetsov was wondering if we could somehow return an error message like this, in this situation:
```
apibindings.apis.kcp.dev "foo" is forbidden: unable to create APIBinding: missing verb='bind' permission on apiexports in workspace "root:notfound:whatever"
```

This at least shows the workspace where the APIExport is supposed to be (`root:notfound:whatever`). We presumably can make this change in the APIBinding admission.

@s-urbaniak I'm not sure if there are any changes we'd want to make to the authorizers, or if this is purely an APIBinding admission issue...

s-urbaniak:
> I'm not sure if there are any changes we'd want to make to the authorizers, or if this is purely an APIBinding admission issue...

@ncdc to separate the two cases you mentioned:

1. `error: failed to create configmap: namespaces "asdf" not found`

[...]

PR #1882, opened 2022-09-02
✨ pkg/virtual/apiexport: add maximal permission policy authorization

### LINKED ISSUE #1337: Resources accessed through `PermissionClaim`s cannot allow privilege escalations
Resources accessed through `PermissionClaim`s cannot allow privilege escalations - see [this comment](https://github.com/kcp-dev/kcp/issues/1337#issuecomment-1229093358) for details

s-urbaniak:
## Summary

This adds maximal permission policy authorization to the virtual apiexport apiserver. Currently, this is not enforced at all.

## Related issue(s)

Part of #1219
Fixes #1337

/cc @sttts

### DIFF
--- a/docs/authorization.md
+++ b/docs/authorization.md
@@ -55,7 +55,7 @@ They are related in the following way:
 [ASCIIFlow document](https://asciiflow.com/#/share/eJyrVspLzE1VslLydg5QcCwtycgvyqxKLVLSUcpJrATSVkrVMUoVMUpWlpaGOjFKlUCWkaUZkFWSWlEC5MQoKdAAPJrS82hKA9FoQkxMHm2c0YQhgGoViQ5FuJhM3RPIsXgCFvXTdoE855OfnJijEJCfk5lcCVQyB0eAgpSG5Bfo5qSWpeaghQ1GGCGLoEtC%2BMhaE%2BFJDiYBDeOi1MLS1OISqKh%2FUXpiXmZVYklmfh667eH5RdnFBYnJqWie3IIebiDFjgGeCk6ZeSmZeelYXIPpWIhrCIQwJDBRvALWPweLKuf8vJLUPKi%2FtNOL8ksLilFUYhqGatASqOFTSEk3M7CmXfSYwxU1qJatQTWXUCxjgkfT9pDmEuwyJIbCDLxu8g9CjgF055EU1qiBQ4buGTjciBIqpBWQoEDfRPVSEiWOnPLzS4oVihILFFCyDrVtRA1MSGaBlWBQJfDcMoOW9QJqDqW%2BV5GsQhOgmVWkFSkxSrVKtQBJrg9s)
 ## Top-Level Organization authorizer
 
-An top-level organization is a workspace directly under root. When a user accesses a top-level organization or
+A top-level organization is a workspace directly under root. When a user accesses a top-level organization or
 a sub-workspace like `root:org:ws:ws`, this authorizer will check in `root` whether the user has permission
 to the top-level org workspace represented by the `ClusterWorkspace` named `org` in `root` with the following verbs:
 
@@ -163,6 +163,72 @@ can be access only by users that are granted `admin` verb on the `workspaces/con
 parent workspace.
 
 Service accounts declared within a workspace don't have access to initializing workspaces.
+
+## API binding authorizer
+
+If the requested resource type is part of an API binding, then the API binding authorizer verifies that
+the request is not exceeding the maximum permission policy of the related API export.
+Currently, the "local policy" maximum permission policy type is supported.
+
+### Local policy
+
+The local maximum permission policy delegates the decision to the RBAC of the related API export.
+To distinguish between local RBAC role bindings in that workspace and those for this these maximum permission policy,
+every name and group is prefixed with `apis.kcp.dev:binding:`.
+
+Example:
+
+Given an API binding for type `foo` declared in workspace `consumer` that refers to an API export declared in workspace `provider`
+and a user `user-1` having the group `group-1` requesting a `create` of `foo` in the `default` namespace in the `consumer` workspace,
+the API binding authorizer verifies that `user-1` is allowed to execute this request by delegating to `provider`'s RBAC using prefixed attributes.
+
+Here, the API binding authorizer prepends the `apis.kcp.dev:binding:` prefix to the username and all groups the user belongs to.
+Using prefixed attributes prevents RBAC collisions i.e. if `user-1` is granted to execute requests within the `provider` workspace directly.
+
+For the given example RBAC request looks as follows:
+
+- Username: `apis.kcp.dev:binding:user-1`
+- Group: `apis.kcp.dev:binding:group-1`
+- Resource: `foo`
+- Namespace: `default`
+- Workspace: `provider`
+- Verb: `create`
+
+The following role and role binding declared within the `provider` workspace will grant access to the request:
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: foo-creator
+  clusterName: provider
+rules:
+- apiGroups:
+  - foo.api
+  resources:
+  - foos
+  verbs:
+  - create
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: user-1-foo-creator
+  namespace: default
+  clusterName: provider
+subjects:
+- kind: User
+  name: apis.kcp.dev:binding:user-1
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: foo-creator
+```
+
+Note: The same authorization scheme is enforced when executing the request of a claimed resource via the virtual API Export API server,
+i.e. a claimed resource is bound to the same maximal permission policy. Only the actual owner of that resources can go beyond that policy.
+
+TBD: Example
 
 ## Kubernetes Bootstrap Policy authorizer
 
--- a/pkg/authorization/apibinding_authorizer.go
+++ b/pkg/authorization/apibinding_authorizer.go
@@ -22,7 +22,6 @@ import (
 
 	"github.com/kcp-dev/logicalcluster/v2"
 
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kaudit "k8s.io/apiserver/pkg/audit"
 	"k8s.io/apiserver/pkg/authentication/user"
 	"k8s.io/apiserver/pkg/authorization/authorizer"
@@ -33,11 +32,8 @@ import (
 
 	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
 	kcpinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
+	"github.com/kcp-dev/kcp/pkg/indexers"
 	rbacwrapper "github.com/kcp-dev/kcp/pkg/virtual/framework/wrappers/rbac"
-)
-
-const (
-	byWorkspaceIndex = "apiBindingAuthorizer-byWorkspace"
 )
 
 const (
@@ -46,37 +42,13 @@ const (
 	APIBindingContentAuditReason   = APIBindingContentAuditPrefix + "reason"
 )
 
-// NewAPIBindingAccessAuthorizer returns an authorizer that checks if the the request is for a
+// NewAPIBindingAccessAuthorizer returns an authorizer that checks if the request is for a
 // bound resource or not. If the resource is bound we will check the user has RBAC access in the
-// exported resources workspace. If it is not allowed we will return NoDecision, if allowed we
+// exported resource workspace. If it is not allowed we will return NoDecision, if allowed we
 // will call the delegate authorizer.
 func NewAPIBindingAccessAuthorizer(kubeInformers kubernetesinformers.SharedInformerFactory, kcpInformers kcpinformers.SharedInformerFactory, delegate authorizer.Authorizer) (authorizer.Authorizer, error) {
-	if _, found := kcpInformers.Apis().V1alpha1().APIBindings().Informer().GetIndexer().GetIndexers()[byWorkspaceIndex]; !found {
-		err := kcpInformers.Apis().V1alpha1().APIBindings().Informer().AddIndexers(
-			cache.Indexers{
-				byWorkspaceIndex: func(obj interface{}) ([]string, error) {
-					return []string{logicalcluster.From(obj.(metav1.Object)).String()}, nil
-				},
-			},
-		)
-		if err != nil {
-			// nothing we can do here. But this should also never happen. We check for existence before.
-			return nil, fmt.Errorf("failed to add indexer for APIBindings: %w", err)
-		}
-	}
-	if _, found := kcpInformers.Apis().V1alpha1().APIExports().Informer().GetIndexer().GetIndexers()[byWorkspaceIndex]; !found {
-		err := kcpInformers.Apis().V1alpha1().APIExports().Informer().AddIndexers(
-			cache.Indexers{
-				byWorkspaceIndex: func(obj interface{}) ([]string, error) {
-					return []string{logicalcluster.From(obj.(metav1.Object)).String()}, nil
-				},
-			},
-		)
-		if err != nil {
-			// nothing we can do here. But this should also never happen. We check for existence before.
-			return nil, fmt.Errorf("failed to add indexer for APIExports: %w", err)
-		}
-	}
+	apiBindingIndexer := kcpInformers.Apis().V1alpha1().APIBindings().Informer().GetIndexer()
+	apiExportIndexer := kcpInformers.Apis().V1alpha1().APIExports().Informer().GetIndexer()
 
 	// Make sure informer knows what to watch
 	kubeInformers.Rbac().V1().Roles().Lister()
@@ -85,18 +57,31 @@ func NewAPIBindingAccessAuthorizer(kubeInformers kubernetesinformers.SharedInfor
 	kubeInformers.Rbac().V1().ClusterRoleBindings().Lister()
 
 	return &apiBindingAccessAuthorizer{
-		versionedInformers: kubeInformers,
-		apiBindingIndexer:  kcpInformers.Apis().V1alpha1().APIBindings().Informer().GetIndexer(),
-		apiExportIndexer:   kcpInformers.Apis().V1alpha1().APIExports().Informer().GetIndexer(),
-		delegate:           delegate,
+		getAPIBindingReferenceForAttributes: func(attr authorizer.Attributes, clusterName logicalcluster.Name) (*apisv1alpha1.ExportReference, bool, error) {
+			return getAPIBindingReferenceForAttributes(apiBindingIndexer, attr, clusterName)
+		},
+		getAPIExportByReference: func(exportRef *apisv1alpha1.ExportReference) (*apisv1alpha1.APIExport, bool, error) {
+			return getAPIExportByReference(apiExportIndexer, exportRef)
+		},
+		newAuthorizer: func(clusterName logicalcluster.Name) authorizer.Authorizer {
+			clusterKubeInformer := rbacwrapper.FilterInformers(clusterName, kubeInformers.Rbac().V1())
+			return rbac.New(
+				&rbac.RoleGetter{Lister: clusterKubeInformer.Roles().Lister()},
+				&rbac.RoleBindingLister{Lister: clusterKubeInformer.RoleBindings().Lister()},
+				&rbac.ClusterRoleGetter{Lister: clusterKubeInformer.ClusterRoles().Lister()},
+				&rbac.ClusterRoleBindingLister{Lister: clusterKubeInformer.ClusterRoleBindings().Lister()},
+			)
+		},
+		delegate: delegate,
 	}, nil
 }
 
 type apiBindingAccessAuthorizer struct {
-	versionedInformers kubernetesinformers.SharedInformerFactory
-	apiBindingIndexer  cache.Indexer
-	apiExportIndexer   cache.Indexer
-	delegate           authorizer.Authorizer
+	delegate authorizer.Authorizer
+
+	getAPIBindingReferenceForAttributes func(attr authorizer.Attributes, clusterName logicalcluster.Name) (ref *apisv1alpha1.ExportReference, found bool, err error)
+	getAPIExportByReference             func(exportRef *apisv1alpha1.ExportReference) (ref *apisv1alpha1.APIExport, found bool, err error)
+	newAuthorizer                       func(clusterName logicalcluster.Name) authorizer.Authorizer
 }
 
 func (a *apiBindingAccessAuthorizer) Authorize(ctx context.Context, attr authorizer.Attributes) (authorizer.Decision, string, error) {
@@ -113,7 +98,7 @@ func (a *apiBindingAccessAuthorizer) Authorize(ctx context.Context, attr authori
 		return authorizer.DecisionNoOpinion, apiBindingAccessDenied, err
 	}
 
-	bindingLogicalCluster, bound, err := a.getAPIBindingReference(attr, lcluster)
+	bindingLogicalCluster, bound, err := a.getAPIBindingReferenceForAttributes(attr, lcluster)
 	if err != nil {
 		kaudit.AddAuditAnnotations(
 			ctx,
@@ -132,7 +117,7 @@ func (a *apiBindingAccessAuthorizer) Authorize(ctx context.Context, attr authori
 		return a.delegate.Authorize(ctx, attr)
 	}
 
-	apiExport, found, err := a.getAPIExport(bindingLogicalCluster)
+	apiExport, found, err := a.getAPIExportByReference(bindingLogicalCluster)
 	if err != nil {
 		kaudit.AddAuditAnnotations(
 			ctx,
@@ -178,13 +163,7 @@ func (a *apiBindingAccessAuthorizer) Authorize(ctx context.Context, attr authori
 	}
 
 	// If bound, create a rbac authorizer filtered to the cluster.
-	clusterKubeInformer := rbacwrapper.FilterInformers(logicalcluster.From(apiExport), a.versionedInformers.Rbac().V1())
-	clusterAuthorizer := rbac.New(
-		&rbac.RoleGetter{Lister: clusterKubeInformer.Roles().Lister()},
-		&rbac.RoleBindingLister{Lister: clusterKubeInformer.RoleBindings().Lister()},
-		&rbac.ClusterRoleGetter{Lister: clusterKubeInformer.ClusterRoles().Lister()},
-		&rbac.ClusterRoleBindingLister{Lister: clusterKubeInformer.ClusterRoleBindings().Lister()},
-	)
+	clusterAuthorizer := a.newAuthorizer(logicalcluster.From(apiExport))
 	prefixedAttr := deepCopyAttributes(attr)
 	userInfo := prefixedAttr.User.(*user.DefaultInfo)
 	userInfo.Name = apisv1alpha1.MaximalPermissionPolicyRBACUserGroupPrefix + userInfo.Name
@@ -204,7 +183,7 @@ func (a *apiBindingAccessAuthorizer) Authorize(ctx context.Context, attr authori
 
 	kaudit.AddAuditAnnotations(
 		ctx,
-		APIBindingContentAuditDecision, decisionString(dec),
+		APIBindingContentAuditDecision, DecisionString(dec),
 		APIBindingContentAuditReason, fmt.Sprintf("API export cluster %q reason: %v", logicalcluster.From(apiExport), reason),
 	)
 
@@ -215,9 +194,8 @@ func (a *apiBindingAccessAuthorizer) Authorize(ctx context.Context, attr authori
 	return authorizer.DecisionNoOpinion, reason, nil
 }
 
-// TODO [shawn-hurley]: this should be a helper shared.
-func (a *apiBindingAccessAuthorizer) getAPIBindingReference(attr authorizer.Attributes, clusterName logicalcluster.Name) (*apisv1alpha1.ExportReference, bool, error) {
-	objs, err := a.apiBindingIndexer.ByIndex(byWorkspaceIndex, clusterName.String())
+func getAPIBindingReferenceForAttributes(apiBindingIndexer cache.Indexer, attr authorizer.Attributes, clusterName logicalcluster.Name) (*apisv1alpha1.ExportReference, bool, error) {
+	objs, err := apiBindingIndexer.ByIndex(indexers.ByLogicalCluster, clusterName.String())
 	if err != nil {
 		return nil, false, err
 	}
@@ -235,12 +213,11 @@ func (a *apiBindingAccessAuthorizer) getAPIBindingReference(attr authorizer.Attr
 	return nil, false, nil
 }
 
-func (a *apiBindingAccessAuthorizer) getAPIExport(exportRef *apisv1alpha1.ExportReference) (*apisv1alpha1.APIExport, bool, error) {
-	objs, err := a.apiExportIndexer.ByIndex(byWorkspaceIndex, exportRef.Workspace.Path)
+func getAPIExportByReference(apiExportIndexer cache.Indexer, exportRef *apisv1alpha1.ExportReference) (*apisv1alpha1.APIExport, bool, error) {
+	objs, err := apiExportIndexer.ByIndex(indexers.ByLogicalCluster, exportRef.Workspace.Path)
 	if err != nil {
 		return nil, false, err
 	}
-
 	for _, obj := range objs {
 		apiExport := obj.(*apisv1alpha1.APIExport)
 		if apiExport.Name == exportRef.Workspace.ExportName {
--- a/pkg/authorization/audit_logger.go
+++ b/pkg/authorization/audit_logger.go
@@ -1,0 +1,52 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package authorization
+
+import (
+	"context"
+	"fmt"
+
+	kaudit "k8s.io/apiserver/pkg/audit"
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+)
+
+const (
+	auditDecision = "decision"
+	auditReason   = "reason"
+)
+
+// NewAuditLogger returns an authorizer that logs every decision of the delegate authorizer
+// for the given audit prefix key.
+// Note: the prefix key must not contain a trailing slash `/`.
+func NewAuditLogger(auditPrefix string, delegate authorizer.Authorizer) authorizer.Authorizer {
+	return authorizer.AuthorizerFunc(func(ctx context.Context, attr authorizer.Attributes) (authorizer.Decision, string, error) {
+		dec, reason, err := delegate.Authorize(ctx, attr)
+
+		auditReason := reason
+		if err != nil {
+			auditReason = fmt.Sprintf("reason: %q, error: %v", reason, err)
+		}
+
+		kaudit.AddAuditAnnotations(
+			ctx,
+			auditPrefix+"/"+auditDecision, DecisionString(dec),
+			auditPrefix+"/"+auditReason, auditReason,
+		)
+
+		return dec, reason, err
+	})
+}
--- a/pkg/authorization/audit_logger.go
+++ b/pkg/authorization/audit_logger.go
@@ -19,6 +19,7 @@ package authorization
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	kaudit "k8s.io/apiserver/pkg/audit"
 	"k8s.io/apiserver/pkg/authorization/authorizer"
@@ -33,18 +34,22 @@ const (
 // for the given audit prefix key.
 // Note: the prefix key must not contain a trailing slash `/`.
 func NewAuditLogger(auditPrefix string, delegate authorizer.Authorizer) authorizer.Authorizer {
+	if strings.HasSuffix(auditPrefix, "/") {
+		panic(fmt.Sprintf("audit prefix must not have a trailing slash: %q", auditPrefix))
+	}
+
 	return authorizer.AuthorizerFunc(func(ctx context.Context, attr authorizer.Attributes) (authorizer.Decision, string, error) {
 		dec, reason, err := delegate.Authorize(ctx, attr)
 
-		auditReason := reason
+		auditReasonMsg := reason
 		if err != nil {
-			auditReason = fmt.Sprintf("reason: %q, error: %v", reason, err)
+			auditReasonMsg = fmt.Sprintf("reason: %q, error: %v", reason, err)
 		}
 
 		kaudit.AddAuditAnnotations(
 			ctx,
 			auditPrefix+"/"+auditDecision, DecisionString(dec),
-			auditPrefix+"/"+auditReason, auditReason,
+			auditPrefix+"/"+auditReason, auditReasonMsg,
 		)
 
 		return dec, reason, err
--- a/pkg/server/apiextensions.go
+++ b/pkg/server/apiextensions.go
@@ -40,7 +40,6 @@ import (
 	"k8s.io/klog/v2"
 
 	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
-	"github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/util/conditions"
 	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
 	apislisters "github.com/kcp-dev/kcp/pkg/client/listers/apis/v1alpha1"
 	tenancylisters "github.com/kcp-dev/kcp/pkg/client/listers/tenancy/v1alpha1"
@@ -103,9 +102,6 @@ func (c *apiBindingAwareCRDLister) List(ctx context.Context, selector labels.Sel
 	}
 	for _, obj := range objs {
 		apiBinding := obj.(*apisv1alpha1.APIBinding)
-		if !conditions.IsTrue(apiBinding, apisv1alpha1.InitialBindingCompleted) {
-			continue
-		}
 
 		for _, boundResource := range apiBinding.Status.BoundResources {
 			crdKey := clusters.ToClusterAwareKey(apibinding.ShadowWorkspaceName, boundResource.Schema.UID)
@@ -224,15 +220,12 @@ func (c *apiBindingAwareCRDLister) Get(ctx context.Context, name string) (*apiex
 		if clusterName == logicalcluster.Wildcard && identity != "" {
 			// Priority 2: APIBinding CRD
 			crd, err = c.getForIdentityWildcard(name, identity)
-		} else if clusterName != logicalcluster.Wildcard && identity != "" {
-			// identities are only supported on the wildcard cluster
-			return nil, apierrors.NewNotFound(apiextensionsv1.Resource("customresourcedefinitions"), name)
 		} else if clusterName == logicalcluster.Wildcard && partialMetadataRequest {
 			// Priority 3: partial metadata wildcard request
 			crd, err = c.getForWildcardPartialMetadata(name)
 		} else if clusterName != logicalcluster.Wildcard {
 			// Priority 4: normal CRD request
-			crd, err = c.get(clusterName, name)
+			crd, err = c.get(clusterName, name, identity)
 		} else {
 			return nil, apierrors.NewNotFound(apiextensionsv1.Resource("customresourcedefinitions"), name)
 		}
@@ -381,7 +374,7 @@ func (c *apiBindingAwareCRDLister) getSystemCRD(clusterName logicalcluster.Name,
 	return c.crdLister.Get(clusters.ToClusterAwareKey(SystemCRDLogicalCluster, name))
 }
 
-func (c *apiBindingAwareCRDLister) get(clusterName logicalcluster.Name, name string) (*apiextensionsv1.CustomResourceDefinition, error) {
+func (c *apiBindingAwareCRDLister) get(clusterName logicalcluster.Name, name, identity string) (*apiextensionsv1.CustomResourceDefinition, error) {
 	var crd *apiextensionsv1.CustomResourceDefinition
 
 	// Priority 1: see if it comes from any APIBindings
@@ -394,12 +387,12 @@ func (c *apiBindingAwareCRDLister) get(clusterName logicalcluster.Name, name str
 	for _, obj := range objs {
 		apiBinding := obj.(*apisv1alpha1.APIBinding)
 
-		if !conditions.IsTrue(apiBinding, apisv1alpha1.InitialBindingCompleted) {
-			continue
-		}
-
 		for _, boundResource := range apiBinding.Status.BoundResources {
-			if boundResource.Group == group && boundResource.Resource == resource {
+			// identity is empty string if the request is coming from a regular workspace client.
+			// It is set if the request is coming from the virtual apiexport apiserver client.
+			matchingIdentity := identity == "" || boundResource.Schema.IdentityHash == identity
+
+			if boundResource.Group == group && boundResource.Resource == resource && matchingIdentity {
 				crdKey := clusters.ToClusterAwareKey(apibinding.ShadowWorkspaceName, boundResource.Schema.UID)
 				crd, err = c.crdLister.Get(crdKey)
 				if err != nil && apierrors.IsNotFound(err) {
@@ -420,16 +413,18 @@ func (c *apiBindingAwareCRDLister) get(clusterName logicalcluster.Name, name str
 		}
 	}
 
-	// Priority 2: see if it exists in the current logical cluster
-	crdKey := clusters.ToClusterAwareKey(clusterName, name)
-	crd, err = c.crdLister.Get(crdKey)
-	if err != nil && !apierrors.IsNotFound(err) {
-		// something went wrong w/the lister - could only happen if meta.Accessor() fails on an item in the store.
-		return nil, err
-	}
+	if identity == "" {
+		// Priority 2: see if it exists in the current logical cluster
+		crdKey := clusters.ToClusterAwareKey(clusterName, name)
+		crd, err = c.crdLister.Get(crdKey)
+		if err != nil && !apierrors.IsNotFound(err) {
+			// something went wrong w/the lister - could only happen if meta.Accessor() fails on an item in the store.
+			return nil, err
+		}
 
-	if crd != nil {
-		return crd, nil
+		if crd != nil {
+			return crd, nil
+		}
 	}
 
 	return nil, apierrors.NewNotFound(schema.GroupResource{Group: apiextensionsv1.SchemeGroupVersion.Group, Resource: "customresourcedefinitions"}, name)
--- a/pkg/virtual/apiexport/authorizer/content.go
+++ b/pkg/virtual/apiexport/authorizer/content.go
@@ -1,0 +1,106 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package authorizer
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/kcp-dev/logicalcluster/v2"
+
+	kaudit "k8s.io/apiserver/pkg/audit"
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+	"k8s.io/client-go/kubernetes"
+
+	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
+	kcpauth "github.com/kcp-dev/kcp/pkg/authorization"
+	"github.com/kcp-dev/kcp/pkg/authorization/delegated"
+	dynamiccontext "github.com/kcp-dev/kcp/pkg/virtual/framework/dynamic/context"
+)
+
+const (
+	VirtualAPIExportNotPermittedReason = "virtual apiexport access not permitted"
+	VirtualAPIExportPermittedReason    = "virtual apiexport access permitted"
+)
+
+const (
+	VirtualAPIExportContentAuditPrefix   = "virtual.apiexport.content.authorization.kcp.dev/"
+	VirtualAPIExportContentAuditDecision = VirtualAPIExportContentAuditPrefix + "decision"
+	VirtualAPIExportContentAuditReason   = VirtualAPIExportContentAuditPrefix + "reason"
+)
+
+type apiExportsContentAuthorizer struct {
+	newDelegatedAuthorizer func(clusterName string) (authorizer.Authorizer, error)
+	delegate               authorizer.Authorizer
+}
+
+// NewAPIExportsContentAuthorizer creates a new authorizer that checks
+// if the user has access to the `apiexports/content` subresource using the same verb as the requested resource.
+// The given kube cluster client is used to execute a SAR request against the cluster of the current in-flight API export.
+// If the SAR decision allows access, the given delegate authorizer is executed to proceed the authorizer chain,
+// else access is denied.
+func NewAPIExportsContentAuthorizer(delegate authorizer.Authorizer, kubeClusterClient kubernetes.ClusterInterface) authorizer.Authorizer {
+	return &apiExportsContentAuthorizer{
+		newDelegatedAuthorizer: func(clusterName string) (authorizer.Authorizer, error) {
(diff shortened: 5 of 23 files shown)


pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,113 @@
...
+
+type maximalPermissionAuthorizer struct {
+	getAPIExport           func(ctx context.Context, clusterName, apiExportName string) (*apisv1alpha1.APIExport, error)
+	newDelegatedAuthorizer func(clusterName string) (authorizer.Authorizer, error)
+}
+
+func NewMaximalPermissionAuthorizer(deepSARClient kubernetes.ClusterInterface, kcpClusterClient kcpclient.ClusterInterface) authorizer.Authorizer {
+	return &maximalPermissionAuthorizer{
+		getAPIExport: func(ctx context.Context, clusterName, apiExportName string) (*apisv1alpha1.APIExport, error) {
+			return kcpClusterClient.Cluster(logicalcluster.New(clusterName)).ApisV1alpha1().APIExports().Get(ctx, apiExportName, v1.GetOptions{})
```
s-urbaniak, line 32: is the live get a problem vs. indexer @sttts ?

pkg/virtual/apiexport/authorizer/content.go
```diff
@@ -0,0 +1,102 @@
...
+	VirtualAPIExportContentAuditPrefix   = "virtual.apiexport.content.authorization.kcp.dev/"
+	VirtualAPIExportContentAuditDecision = VirtualAPIExportContentAuditPrefix + "decision"
+	VirtualAPIExportContentAuditReason   = VirtualAPIExportContentAuditPrefix + "reason"
+
+	VirtualAPIExportMaximumPermissionPolicyAuditPrefix   = "virtual.apiexport.maxpermissionpolicy.authorization.kcp.dev/"
+	VirtualAPIExportMaximumPermissionPolicyAuditDecision = VirtualAPIExportMaximumPermissionPolicyAuditPrefix + "decision"
+	VirtualAPIExportMaximumPermissionPolicyAuditReason   = VirtualAPIExportMaximumPermissionPolicyAuditPrefix + "reason"
+)
+
+type apiExportsContentAuthorizer struct {
```
sttts, line 48: go doc what this does on a high level

pkg/virtual/apiexport/authorizer/content.go
```diff
@@ -0,0 +1,102 @@
...
+	SARAttributes := authorizer.AttributesRecord{
+		APIGroup:        apisv1alpha1.SchemeGroupVersion.Group,
+		APIVersion:      apisv1alpha1.SchemeGroupVersion.Version,
+		User:            attr.GetUser(),
+		Verb:            attr.GetVerb(),
+		Name:            apiExportName,
+		Resource:        "apiexports",
+		ResourceRequest: true,
+		Subresource:     "content",
+	}
```
sttts, line 94: wondering why this authorizer is needed in the first place. It checks that the APIExport user, i.e. an owner of a permission claim, has content access to the workspace at hand?

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,129 @@
...
+	"k8s.io/client-go/kubernetes"
+
+	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
+	kcpauth "github.com/kcp-dev/kcp/pkg/authorization"
+	"github.com/kcp-dev/kcp/pkg/authorization/delegated"
+	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	dynamiccontext "github.com/kcp-dev/kcp/pkg/virtual/framework/dynamic/context"
+)
+
+type maximalPermissionAuthorizer struct {
```
sttts, line 40: high level comment please what this authorizer does

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,129 @@
...
+
+	apiDomainKey := dynamiccontext.APIDomainKeyFrom(ctx)
+	parts := strings.Split(string(apiDomainKey), "/")
+	if len(parts) < 2 {
+		return authorizer.DecisionNoOpinion, VirtualAPIExportNotPermittedReason, fmt.Errorf("invalid API domain key: %q", apiDomainKey)
+	}
+	apiExportCluster := parts[0]
+	apiExportName := parts[1]
+
+	apiExport, err := a.getAPIExport(ctx, apiExportCluster, apiExportName)
```
sttts, line 77: is this the right APIExport ? The domain key points to the red APIExport B in https://docs.google.com/drawings/d/1_sOiFZReAfypuUDyHS9rwpxbZgJNJuxdvbgXXgu2KAQ/edit, doesn't it?

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil && errors.IsNotFound(err) {
+		return authorizer.DecisionNoOpinion, VirtualAPIExportNotPermittedReason, fmt.Errorf("API export %q not found: %q", apiDomainKey)
+	}
+
+	if apiExport.Spec.MaximalPermissionPolicy == nil {
+		return authorizer.DecisionAllow, "no maximal permission policy defined", nil
+	}
+
+	if apiExport.Spec.MaximalPermissionPolicy.Local == nil {
+		return authorizer.DecisionAllow, "no maximal local permission policy defined", nil
```
sttts, line 87: does the reason end up in the request response? In that case, I would say we should not leak this information. Showing it in audit of course is fine (are there audit logs for virtual apiservers? cc @davidfestal)

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return authorizer.DecisionNoOpinion, VirtualAPIExportNotPermittedReason, err
+	}
+
+	dec, reason, err = authz.Authorize(ctx, prefixedAttr)
+	if err != nil {
+		return authorizer.DecisionNoOpinion, VirtualAPIExportNotPermittedReason, fmt.Errorf("error authorizing in API export cluster %q: %w", apiExportCluster, err)
+	}
+	reason = fmt.Sprintf("maximal permission policy authorization reason in API export cluster %q: %q", apiExportCluster, reason)
+	return
```
sttts, line 108: nit: am not a fan of non-explicit returns, i.e. via named vars. This also risks that one does forget to set certain vars.

pkg/virtual/apiexport/authorizer/content.go
sttts: oh, nvmd. It checks that the "red APIExport B owner" actually has the required access. Makes sense.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,129 @@
...
+type maximalPermissionAuthorizer struct {
+	getAPIExport           func(ctx context.Context, clusterName, apiExportName string) (*apisv1alpha1.APIExport, error)
+	newDelegatedAuthorizer func(clusterName string) (authorizer.Authorizer, error)
+}
+
+func NewMaximalPermissionAuthorizer(deepSARClient kubernetes.ClusterInterface, kcpClusterClient kcpclient.ClusterInterface) authorizer.Authorizer {
+	return &maximalPermissionAuthorizer{
+		getAPIExport: func(ctx context.Context, clusterName, apiExportName string) (*apisv1alpha1.APIExport, error) {
+			return kcpClusterClient.Cluster(logicalcluster.New(clusterName)).ApisV1alpha1().APIExports().Get(ctx, apiExportName, v1.GetOptions{})
+		},
```
sttts, line 49: as hinted at in your question, this should be lister backed.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: yes, indeed 😱 i have to rework this part.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: honestly, i'm not a fain either. let me see if i can factor this and not have lots of duplication.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: ack, wasn't sure as we use live gets somewhere else as well but will factor towards lister 👍

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: In fact, they do. we have https://github.com/kcp-dev/kcp/issues/1786 which asks for more transparent messages towards clients. But I'd say I'll tackle this separately so we can first reach consensus on transparency vs. information leaking separately.

s-urbaniak:
marking as WIP, as this is currently fundamentally wrong.

stevekuznetsov (commented):
Could we please require a markdown section in some authz doc in the repo for this to merge? I'd like us to be able to describe the flow here to users the moment it's available.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: i'll rework audit here completely

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: done

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: will leave messages as is

s-urbaniak: that is, reasons won't be exposed to the user

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: fixed

s-urbaniak:
@stevekuznetsov absolutely 👍 i'll make sure to add this.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: it is, fixed

docs/authorization.md
```diff
@@ -164,6 +164,69 @@ parent workspace.
...
 
+## API binding authorizer
+
+If the requested resource type is part of an API binding, then the API binding authorizer verifies that
+the request is not exceeding the maximum permission policy of the related API export.
+Currently, the "local policy" maximum permission policy type is supported.
+
+### Local policy
+
+The local maximum permission policy delegates the decision to the RBAC of the related API export workspace.
```
sttts, line 175:
```suggestion
The local maximum permission policy delegates the decision to the RBAC of the related API export. To distinguish between local RBAC role bindings in that workspace and those for this these maximum permission policy, every name and group is prefixed with `apis.kcp.dev:binding:`.
```

docs/authorization.md
```diff
@@ -164,6 +164,69 @@ parent workspace.
...
+subjects:
+- kind: User
+  name: apis.kcp.dev:binding:user-1
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: foo-creator
+```
+
+Note: The same authorization scheme is enforced when executing the request via the virtual API Export API server.
```
sttts, line 226:
```suggestion
Note: The same authorization scheme is enforced when executing the request of a claimed resource (... link permission claims ...) via the virtual API Export API server, i.e. a claimed resource is bound to the same maximal permission policy. Only the actual owner of that resources can go beyond that policy.
```

pkg/authorization/apibinding_authorizer.go
```diff
@@ -51,32 +52,22 @@ const (
...
-		)
-		if err != nil {
-			// nothing we can do here. But this should also never happen. We check for existence before.
-			return nil, fmt.Errorf("failed to add indexer for APIExports: %w", err)
-		}
-	}
+	apiBindingIndexer := kcpInformers.Apis().V1alpha1().APIBindings().Informer().GetIndexer()
+	apiExportIndexer := kcpInformers.Apis().V1alpha1().APIExports().Informer().GetIndexer()
+
+	indexers.AddIfNotPresentOrDie(
```
ncdc, line 58: Since you're changing this... every shared informer generally has `indexers.ByLogicalCluster` and `indexers.ByLogicalClusterAndNamespace` preconfigured.

docs/authorization.md
s-urbaniak: it is mentioned later but adding here additionally makes sense 👍 fixed

docs/authorization.md
s-urbaniak: > `... link permission claims ...`

we don't have a markdown for this 😞 that's next on my list but adding the rest.

pkg/authorization/apibinding_authorizer.go
s-urbaniak: oh i missed that one, I believe I can then even change that to:
```
indexers.AddIfNotPresentOrDie(apiBindingIndexer, indexers.ClusterScoped())
indexers.AddIfNotPresentOrDie(apiExportIndexer, indexers.ClusterScoped())
```

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,257 @@
...
+			if err != nil {
+				return nil, err
+			}
+			if len(exports) == 0 {
+				return nil, errors.New("export not found")
+			}
+			return exports[0].(*apisv1alpha1.APIExport), nil
+		},
+		newDeepSARAuthorizer: func(clusterName logicalcluster.Name) (authorizer.Authorizer, error) {
+			return delegated.NewDelegatedAuthorizer(clusterName, deepSARClient)
```
stevekuznetsov, line 86: What do we gain by factoring this out?

pkg/authorization/apibinding_authorizer.go
```diff
@@ -235,12 +215,11 @@ func (a *apiBindingAccessAuthorizer) getAPIBindingReference(attr authorizer.Attr
 	return nil, false, nil
 }
 
-func (a *apiBindingAccessAuthorizer) getAPIExport(exportRef *apisv1alpha1.ExportReference) (*apisv1alpha1.APIExport, bool, error) {
-	objs, err := a.apiExportIndexer.ByIndex(byWorkspaceIndex, exportRef.Workspace.Path)
+func GetAPIExportByReference(apiExportIndexer cache.Indexer, exportRef *apisv1alpha1.ExportReference) (*apisv1alpha1.APIExport, bool, error) {
```
stevekuznetsov, line 218: Why are these two not simply indices? As opposed to doing the broader cluster index and iterating?

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,257 @@
...
+		kaudit.AddAuditAnnotations(
+			ctx,
+			VirtualAPIExportMaximumPermissionPolicyAuditDecision, kcpauth.DecisionNoOpinion,
+			VirtualAPIExportMaximumPermissionPolicyAuditReason, fmt.Sprintf("error getting cluster from request: %v", err),
+		)
+		return authorizer.DecisionNoOpinion, VirtualAPIExportNotPermittedReason, err
+	}
+
+	if lcluster != logicalcluster.Wildcard {
+		_, bound, err := a.getAPIBindingReferenceForAttributes(attr, lcluster)
```
stevekuznetsov, line 105: I'm not sure I follow this logic - why is checking to see if the API is bound part of authz here?

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	identityHash := getPermissionClaimIdentity(apiExport, attr)
+	if identityHash == "" {
+		kaudit.AddAuditAnnotations(
+			ctx,
+			VirtualAPIExportMaximumPermissionPolicyAuditDecision, kcpauth.DecisionAllowed,
+			VirtualAPIExportMaximumPermissionPolicyAuditReason, "claimed resource does not have an identity hash",
+		)
+		return authorizer.DecisionAllow, notEnforcedMaximumPermissionPolicyReason, nil
```
stevekuznetsov, line 157: Why wouldn't we enforce maximal permission policy on resources without an identity hash? example would be a secret, right?

pkg/authorization/apibinding_authorizer.go
```diff
@@ -85,18 +59,31 @@ func NewAPIBindingAccessAuthorizer(kubeInformers kubernetesinformers.SharedInfor
...
 type apiBindingAccessAuthorizer struct {
-	versionedInformers kubernetesinformers.SharedInformerFactory
-	apiBindingIndexer  cache.Indexer
-	apiExportIndexer   cache.Indexer
-	delegate           authorizer.Authorizer
+	delegate authorizer.Authorizer
+
+	getAPIBindingReferenceForAttributes func(attr authorizer.Attributes, clusterName logicalcluster.Name) (ref *apisv1alpha1.ExportReference, found bool, err error)
+	getAPIExportByReference             func(exportRef *apisv1alpha1.ExportReference) (ref *apisv1alpha1.APIExport, found bool, err error)
+	newAuthorizer                       func(clusterName logicalcluster.Name) authorizer.Authorizer
```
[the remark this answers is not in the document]
s-urbaniak: I was pointed to this pattern by @ncdc which is used elsewhere. Yes I want to unit test the authorizers, they are missing coverage.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: same as https://github.com/kcp-dev/kcp/pull/1882#discussion_r966004349

pkg/authorization/apibinding_authorizer.go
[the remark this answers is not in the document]
stevekuznetsov: OK - if it used for unit testing, sounds good. I did not see any unit tests :P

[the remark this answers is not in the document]
sttts: yes, we use that in many places now, mainly controllers to get easy testability. The idea to simplify mocking, i.e. not to force us to have a big interface to mock even if only one func is used.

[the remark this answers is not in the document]
s-urbaniak: > follow-up is adding unit tests

you know what ... i'll add them here :-)

pkg/authorization/apibinding_authorizer.go
```diff
@@ -215,9 +196,8 @@ func (a *apiBindingAccessAuthorizer) Authorize(ctx context.Context, attr authori
 	return authorizer.DecisionNoOpinion, reason, nil
 }
 
-//TODO [shawn-hurley]: this should be a helper shared.
-func (a *apiBindingAccessAuthorizer) getAPIBindingReference(attr authorizer.Attributes, clusterName logicalcluster.Name) (*apisv1alpha1.ExportReference, bool, error) {
-	objs, err := a.apiBindingIndexer.ByIndex(byWorkspaceIndex, clusterName.String())
+func GetAPIBindingReferenceForAttributes(apiBindingIndexer cache.Indexer, attr authorizer.Attributes, clusterName logicalcluster.Name) (*apisv1alpha1.ExportReference, bool, error) {
```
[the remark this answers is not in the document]
s-urbaniak: This assumptions is held in the current API binding authorizer, I didn't question it, @sttts is this an invariant we **should** question?

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: Two cases:
- wildcard request: no unique API binding can be determined => this case is skipped
- cluster request: a unique API binding can be determined => according to @sttts we should only forward authz iff the API binding is bound, it resembles the same logic as in the API binding authorizer

Maybe omitting this here is good enough nevertheless (just follow the "wildcard request" case) @sttts ?

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil && kerrors.IsNotFound(err) {
+		kaudit.AddAuditAnnotations(
+			ctx,
+			VirtualAPIExportMaximumPermissionPolicyAuditDecision, kcpauth.DecisionNoOpinion,
+			VirtualAPIExportMaximumPermissionPolicyAuditReason, fmt.Sprintf("API export %q not found", apiDomainKey),
+		)
+		return authorizer.DecisionNoOpinion, VirtualAPIExportNotPermittedReason, err
+	}
+
+	identityHash := getPermissionClaimIdentity(apiExport, attr)
```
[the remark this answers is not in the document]
s-urbaniak: `getPermissionClaimIdentity` iterates over the claimed resources in the export and matches the GR of the requested resource. If it doesn't find any this returns empty string.

docs/authorization.md
```diff
@@ -164,6 +164,72 @@ parent workspace.
...
+the request is not exceeding the maximum permission policy of the related API export.
+Currently, the "local policy" maximum permission policy type is supported.
+
+### Local policy
+
+The local maximum permission policy delegates the decision to the RBAC of the related API export.
+To distinguish between local RBAC role bindings in that workspace and those for this these maximum permission policy,
+every name and group is prefixed with `apis.kcp.dev:binding:`.
+
+Example:
```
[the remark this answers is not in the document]
s-urbaniak: sounds great 👍 i can definitely add very concrete instructions.

pkg/authorization/apibinding_authorizer.go
s-urbaniak: That was in the original code, I reused it. However, we could build a dedicated index that uses api export's cluster name + api export's bound resource group + export's bound resource.

Are compound index keys idiomatic?

s-urbaniak: @sttts ^ ?

stevekuznetsov: Use type-safe index key formatters to ensure it is used correctly, see https://github.com/kcp-dev/apimachinery/blob/main/pkg/cache/controller.go

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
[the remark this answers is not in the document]
stevekuznetsov: A claimed resource may have no identity hash, this needs a different factoring

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
stevekuznetsov: Right, but we certainly want to have permission policies on those, no?

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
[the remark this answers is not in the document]
s-urbaniak: hmm ... @sttts i need your input on this, i don't have enough knowledge on the design invariants.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: according to my discussions with @sttts maximum permission policy cannot be applied in this case.

stevekuznetsov: Isn't the implication here then that an API provider that claims some k8s resource (say, a particular Secret or ConfigMap) could set a maxmial permission on that resource from the user? I'm confused about the use-cases here

s-urbaniak: You have a point here 🤔 I believe we have to separete still the two request cases:

- _non-wildcard request for a concrete cluster_: in this case a unique API binding can be found and thus an API export whose maximal permission can be asserted, even for native types. This is currently missing here but present in the existing API binding authorizer.
- _wildcard requests_: In this case no uniqe API binding can be found but still we can try to determine an API export via the identity hash of the claimed resource, if it exists. In this case an API export can be found and asserted. However if it cannot be found, contrary to the current logic, we should deny access.

Let me code this down, and also present the different use cases in the community meeting, and also document those as it is indeed tricky.

pkg/authorization/apibinding_authorizer.go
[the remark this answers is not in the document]
sttts: Yes, BoundResources of bindings and CRDs are not overlapping. It's an invariant.

The APIBinding controller writing this is to be run as singleton either inside of non-HA kcp itself or a controller manager with a lease. Then we can trust this invariant.

pkg/authorization/apibinding_authorizer.go
sttts: We have a handful of exports per workspace. Iterating is not an issue here. If we had a hundred or a thousand of those objects we would care.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: > Isn't the implication here then that an API provider that claims some k8s resource (say, a particular Secret or ConfigMap) could set a maxmial permission on that resource from the user?

afaik this is not possible, we cannot set maximal permission policies on k8s resources.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: It was wrong indeed, I removed it. If the request is targetting a concrete cluster (not to be confused with the api export cluster being present of the virtual apiserver URL) we could go the route of checking API bindings as well, however after a paring session with @sttts we figured that this authorizer should focus on the API exports in the first place.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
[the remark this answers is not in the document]
s-urbaniak: > A claimed resource may have no identity hash, this needs a different factoring

Yes precisely, in this case a maximum policy cannot exist. We factored this to allow the request in this case.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: You cannot set a maximal permission policy on i.e. a secret or configmap, only on exported types. To have more clarification we made this an explicit comment.

pkg/authorization/apibinding_authorizer.go
```diff
@@ -33,50 +32,25 @@ import (
...
-		)
-		if err != nil {
-			// nothing we can do here. But this should also never happen. We check for existence before.
-			return nil, fmt.Errorf("failed to add indexer for APIExports: %w", err)
-		}
-	}
+	apiBindingIndexer := kcpInformers.Apis().V1alpha1().APIBindings().Informer().GetIndexer()
+	apiExportIndexer := kcpInformers.Apis().V1alpha1().APIExports().Informer().GetIndexer()
+	indexers.AddIfNotPresentOrDie(apiBindingIndexer, indexers.ClusterScoped())
+	indexers.AddIfNotPresentOrDie(apiExportIndexer, indexers.ClusterScoped())
```
ncdc, line 53: I don't think you need these - see https://github.com/kcp-dev/kcp/blob/ea31363ae8d683175e435421c4597646e1a317e8/pkg/server/config.go#L230-L235

pkg/virtual/apiexport/authorizer/content.go
```diff
@@ -0,0 +1,106 @@
...
+	apiDomainKey := dynamiccontext.APIDomainKeyFrom(ctx)
+	parts := strings.Split(string(apiDomainKey), "/")
+	if len(parts) < 2 {
+		return authorizer.DecisionNoOpinion, VirtualAPIExportNotPermittedReason, fmt.Errorf("invalid API domain key: %q", apiDomainKey)
+	}
+
+	apiExportCluster, apiExportName := parts[0], parts[1]
+	authz, err := a.newDelegatedAuthorizer(apiExportCluster)
+	if err != nil {
+		return authorizer.DecisionNoOpinion, VirtualAPIExportNotPermittedReason, fmt.Errorf("error creating delegated authorizer for API export %q: %w", apiExportCluster, err)
```
ncdc, line 76: the error message says "API export %q" but the %q value is apiExportCluster (?)

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,192 @@
...
+// The check is omitted if the requested resource itself is not associated with another API export.
+//
+// If the request is a cluster request the authorizer skips authorization if the request is not for a bounded resource.
+// If the request is a wildcard request this check is skipped because no unique API binding can be determined.
+func NewMaximalPermissionAuthorizer(deepSARClient kubernetes.ClusterInterface, apiExportInformer apisinformers.APIExportInformer, apiBindingInformer apisinformers.APIBindingInformer) authorizer.Authorizer {
+	apiExportLister := apiExportInformer.Lister()
+	apiExportIndexer := apiExportInformer.Informer().GetIndexer()
+	apiBindingIndexer := apiBindingInformer.Informer().GetIndexer()
+	indexers.AddIfNotPresentOrDie(apiBindingIndexer, indexers.ClusterScoped())
+	indexers.AddIfNotPresentOrDie(apiExportIndexer, indexers.ClusterScoped())
```
ncdc, line 64: Should not be needed

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,192 @@
...
+		getAPIExportsByIdentity: func(identityHash string) ([]*apisv1alpha1.APIExport, error) {
+			exports, err := apiExportIndexer.ByIndex(indexers.APIExportByIdentity, identityHash)
+			if err != nil {
+				return nil, err
+			}
+			var results []*apisv1alpha1.APIExport
+			for _, obj := range exports {
+				results = append(results, obj.(*apisv1alpha1.APIExport))
+			}
+			return results, nil
```
ncdc, line 79: `indexers.ByIndex` offers go generics support to make this a 1-liner 😄

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,192 @@
...
+	if len(parts) < 2 {
+		err := fmt.Errorf("invalid API domain key: %q", apiDomainKey)
+		return authorizer.DecisionNoOpinion, VirtualAPIExportNotPermittedReason, err
+	}
+
+	claimingAPIExportCluster := parts[0]
+	claimingAPIExportName := parts[1]
+
+	claimingAPIExport, err := a.getAPIExport(claimingAPIExportCluster, claimingAPIExportName)
+	if err != nil && kerrors.IsNotFound(err) {
```
ncdc, line 99: can simplify to just `if kerrors.IsNotFound(err)`

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,192 @@
...
+		if claimedAPIExport.Spec.MaximalPermissionPolicy == nil {
+			continue
+		}
+
+		if claimedAPIExport.Spec.MaximalPermissionPolicy.Local == nil {
+			continue
+		}
+
+		prefixedAttr := deepCopyAttributes(attr)
+		userInfo := prefixedAttr.User.(*user.DefaultInfo)
```
ncdc, line 138: My initial reaction here was to ask why we're type asserting to `*user.DefaultInfo`. I then realized it was so we could modify the user and groups to prepend the maximal permission policy prefix. I see that deepCopyAttributes is only used in this file and added in this PR. Would it be possible to make some edits so maybe instead of doing this in two steps (deep copy, then adjust user/groups), we just do it in one, and we can avoid the type assertion?

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: fixed 👍

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: uh, nice! fixing!

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: oh sure thing 👍 this is pure copy+pasta from somewhere else, let me try to enhance this.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: removed

pkg/virtual/apiexport/authorizer/content.go
s-urbaniak: yes, %q makes nice double quotes. however, returning the api export is not necessary as it directly part of the URL.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,192 @@
...
+	getAPIExport            func(clusterName, apiExportName string) (*apisv1alpha1.APIExport, error)
+	newDeepSARAuthorizer    func(clusterName logicalcluster.Name) (authorizer.Authorizer, error)
+	getAPIExportsByIdentity func(identityHash string) ([]*apisv1alpha1.APIExport, error)
+}
+
+// NewMaximalPermissionAuthorizer creates an authorizer that checks the maximal permission policy
+// for the requested resource if the resource is a claimed resource in the requested API export.
+// The check is omitted if the requested resource itself is not associated with another API export.
+//
+// If the request is a cluster request the authorizer skips authorization if the request is not for a bounded resource.
```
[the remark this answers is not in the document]
s-urbaniak: fixed

pkg/virtual/apiexport/authorizer/content.go
ncdc: fwiw, i wasn't asking about %q 😄 i was pointing out that the text says "API export" but the value is the cluster name

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: done

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: done, nice!

pkg/server/apiextensions.go
```diff
@@ -103,9 +102,6 @@ func (c *apiBindingAwareCRDLister) List(ctx context.Context, selector labels.Sel
 	}
 	for _, obj := range objs {
 		apiBinding := obj.(*apisv1alpha1.APIBinding)
-		if !conditions.IsTrue(apiBinding, apisv1alpha1.InitialBindingCompleted) {
```
stevekuznetsov, line 106: `apiBindingAwareCRDLister.List()` shows the CRDs that are either present from literal CRD objects, projected from `kcp`, or present from `APIBindings` - but if the binding has not been bound yet, it is not providing any APIs in the workspace, right? So what is the meaning from removing this block?

pkg/server/apiextensions.go
```diff
@@ -394,12 +387,12 @@ func (c *apiBindingAwareCRDLister) get(clusterName logicalcluster.Name, name str
 	for _, obj := range objs {
 		apiBinding := obj.(*apisv1alpha1.APIBinding)
 
-		if !conditions.IsTrue(apiBinding, apisv1alpha1.InitialBindingCompleted) {
-			continue
-		}
-
 		for _, boundResource := range apiBinding.Status.BoundResources {
-			if boundResource.Group == group && boundResource.Resource == resource {
+			// identity is empty string if the request is coming from a regular workspace client.
```
stevekuznetsov, line 391: IMHO using the presence or absence of stringified data to drive logic is fragile - I just spent a long time reworking assumptions like this in keyfuncs to be more concrete & use the type system to be clearer. Do we *need* this implicit assumption or is there perhaps some other factoring / metadata / etc that could help us make the distinction explicitly?

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,176 @@
...
+
+type maximalPermissionAuthorizer struct {
+	getAPIExport            func(clusterName, apiExportName string) (*apisv1alpha1.APIExport, error)
+	newDeepSARAuthorizer    func(clusterName logicalcluster.Name) (authorizer.Authorizer, error)
+	getAPIExportsByIdentity func(identityHash string) ([]*apisv1alpha1.APIExport, error)
+}
+
+// NewMaximalPermissionAuthorizer creates an authorizer that checks the maximal permission policy
+// for the requested resource if the resource is a claimed resource in the requested API export.
+// The check is omitted if the requested resource itself is not associated with another API export.
```
stevekuznetsov, line 53: What does this mean? A resource must be associated with more than one export for this to fire?

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,176 @@
...
+			return delegated.NewDelegatedAuthorizer(clusterName, deepSARClient)
+		},
+	}
+}
+
+func (a *maximalPermissionAuthorizer) Authorize(ctx context.Context, attr authorizer.Attributes) (authorizer.Decision, string, error) {
+	apiDomainKey := dynamiccontext.APIDomainKeyFrom(ctx)
+	parts := strings.Split(string(apiDomainKey), "/")
+	if len(parts) < 2 {
+		return authorizer.DecisionNoOpinion, VirtualAPIExportNotPermittedReason, fmt.Errorf("invalid API domain key")
```
stevekuznetsov, line 78: Are those user-facing errors?

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,176 @@
...
+		// it's a resource in the claiming API export, hence unclaimed
+		return authorizer.DecisionAllow, "unclaimed resource", nil
+	}
+	if claimedIdentityHash == "" {
+		// it's a native k8s resource (secret, configmap, ...), or a system kcp CRD resource (apis.kcp.dev)
+		// For neither case a maximum permission policy can exist.
+		return authorizer.DecisionAllow, "unclaimable resource", nil
+	}
+
+	claimedAPIExports, err := a.getAPIExportsByIdentity(claimedIdentityHash)
```
stevekuznetsov, line 100: nit: this code is super dense so I think for clarity variable names will make a big difference - I got lost with `claimedAPIExports` - since there's no concept of claiming an export. This is really the list of exports that provide the APIs which then have permission claims on them ... no idea what to call this. `apiExportsProvidingClaimedResources`?

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
```diff
@@ -0,0 +1,176 @@
...
+
+	if len(claimedAPIExports) == 0 {
+		// a claimed identity hash exists but not API export can be found referring to it (potentially eventually consistent).
+		// In this case be safe and deny the request, forcing the caller to retry.
+		return authorizer.DecisionDeny, VirtualAPIExportNotPermittedReason, nil
+	}
+
+	// multiple claimed API exports can share the same identity hash (even in different workspaces).
+	// All maximum permission policies must grant access because at this point no deterministic API export can be picked
+	// (especially in the wildcard case).
```
stevekuznetsov, line 113: I thought the GoDoc above said we don't even bother to check in wildcard?

test/e2e/virtual/apiexport/virtualworkspace_test.go
```diff
@@ -94,7 +94,7 @@ func TestAPIExportVirtualWorkspace(t *testing.T) {
 	framework.AdmitWorkspaceAccess(t, ctx, kubeClusterClient, serviceProviderWorkspace, []string{"user-1", "user-2"}, nil, []string{"access"})
 	framework.AdmitWorkspaceAccess(t, ctx, kubeClusterClient, consumerWorkspace, []string{"user-3"}, nil, []string{"admin"})
 
-	cr, crb := createClusterRoleAndBindings("user-3-binding", "user-3", "User", "apiexports", "today-cowboys", []string{"bind"})
```
stevekuznetsov, line 97: for anyone keeping track, it's the loss of velocity & extra burden caused by generic test helpers which I don't like ;)

test/e2e/virtual/apiexport/authorizer_test.go
```diff
@@ -0,0 +1,339 @@
...
+	framework.Eventually(t, func() (done bool, str string) {
+		sheriffExport, err := user1KcpClient.ApisV1alpha1().APIExports().Get(logicalcluster.WithCluster(ctx, serviceProvider1Workspace), "wild.wild.west", metav1.GetOptions{})
+		if err != nil {
+			return false, fmt.Sprintf("error while waiting to get API export: %v", err)
+		}
+		if conditions.IsTrue(sheriffExport, apisv1alpha1.APIExportIdentityValid) {
+			identityHash = sheriffExport.Status.IdentityHash
+			return true, ""
+		}
+		return false, "waiting for API export identity to be valid"
```
stevekuznetsov, line 116: It can be nice to return the reason & message on the condition here so it's clear from logs what went wrong

test/e2e/virtual/apiexport/authorizer_test.go
```diff
@@ -0,0 +1,339 @@
...
+		if err != nil {
+			return false, fmt.Sprintf("error while waiting to list sherriffs: %v", err)
+		}
+		return true, ""
+	}, wait.ForeverTestTimeout, 100*time.Millisecond, "listing claimed resources failed")
+
+	_, err = user2DynamicVWClient.Cluster(tenantShadowCRDWorkspace).Resource(schema.GroupVersionResource{Version: "v1alpha1", Resource: "cowboys", Group: "wildwest.dev"}).List(ctx, metav1.ListOptions{})
+	require.Error(t, err, "expected error, got none")
+	require.True(t, errors.IsNotFound(err))
+}
```
stevekuznetsov, line 339: bravo on this, holy moly is that a complex test

stevekuznetsov (commented):
Authorizer logic was fairly clear to me but I'm not confident I could figure out if there were missing edges

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: I just checked, yes these bubble down to the user.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: `apiExportsProvidingClaimedResources` sounds very verbose but if it helps i can rename this one 👍

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: non-native speaker alarm, it's much simpler: Does
> The check is omitted if the requested resource itself is not associated with an API export.

sound more clear?

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: I think the brackets can be indeed left out as the code doesn't make a difference any more between wildcard and non-wildcard requests.

pkg/server/apiextensions.go
s-urbaniak: referring to @sttts as this is resulting from pair programming.

pkg/server/apiextensions.go
s-urbaniak: also referring to @sttts 🙏

test/e2e/virtual/apiexport/virtualworkspace_test.go
s-urbaniak: sorry, i don't follow :-) do you mind to elaborate? Generally I am not too happy about my factoring of `createClusterRoleAndBindings` either. Putting those struct initialziations here though makes the code size huge. I am fine doing this though if it seems more idiomatic and leave `createClusterRoleAndBindings` as is, wdyt?

test/e2e/virtual/apiexport/authorizer_test.go
s-urbaniak: uff, yes, this e2e test took most of the coding time. but it literally "tells a story".

test/e2e/virtual/apiexport/authorizer_test.go
s-urbaniak: indeed, changed to
```
return false, fmt.Sprintf("waiting for API export identity to be valid: %+v", conditions.Get(sheriffExport, apisv1alpha1.APIExportIdentityValid))
```

pkg/server/apiextensions.go
sttts: > but if the binding has not been bound yet, it is not providing any APIs in the workspace, right?

The condition is not the source of thruth, the bound resource slice is. The binding can be partial at any point in time (imagine a resource is added to an export, but a CRD is in place leading to a conflict). Making this logic dependent on a condition it not helpful, i.e. we should always just consider bound resources.

pkg/server/apiextensions.go
sttts: we already have this today basically everywhere including in this file, i.e. identity empty vs. identity non-empty. This code here just matches that mattern.

What I agree with is that the comment should not distinguish between clients here, but talk about the request: with-identity and without-identity requests.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
stevekuznetsov: I don't have strong feelings on the particular name :) --> maybe the Germans will like that long one more though :D

test/e2e/virtual/apiexport/virtualworkspace_test.go
stevekuznetsov: We've had a lot of conversations in the team over the last N months about mega-helpers in tests, and I've always said to keep things simple & isolated, or you end up with this type of refactor burden on YOU in this PR, which is not ideal. You should do whatever you think is simpler and quicker for you here.

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: yes, indeed, and i am realizing most authorizers bubble down errors as well. I need to wrap this with a anonymizer+audit logger 🤔

pkg/virtual/apiexport/authorizer/maximal_permission_policy.go
s-urbaniak: removed the brackets and the wildcard case sentence

pkg/authorization/audit_logger.go
```diff
@@ -0,0 +1,52 @@
...
+
+		auditReason := reason
+		if err != nil {
+			auditReason = fmt.Sprintf("reason: %q, error: %v", reason, err)
+		}
+
+		kaudit.AddAuditAnnotations(
+			ctx,
+			auditPrefix+"/"+auditDecision, DecisionString(dec),
+			auditPrefix+"/"+auditReason, auditReason,
```
ncdc, line 47:
```suggestion
			auditPrefix+"/"+auditReason, reason,
```

test/e2e/virtual/apiexport/authorizer_test.go
```diff
@@ -0,0 +1,342 @@
...
+			return false, fmt.Sprintf("waiting on apiexport to be available %v", err.Error())
+		}
+		if len(apiExport.Status.VirtualWorkspaces) > 0 {
+			return true, ""
+		}
+		return false, "waiting on virtual workspace to be ready"
+
+	}, wait.ForeverTestTimeout, 100*time.Millisecond, "waiting on virtual workspace to be ready")
+
+	t.Logf("verify that we cannot list claimed resources")
```
ncdc, line 289:
```suggestion
	t.Logf("verify that user-2 cannot list claimed resources")
```

test/e2e/virtual/apiexport/authorizer_test.go
```diff
@@ -0,0 +1,342 @@
...
+		"apis.kcp.dev:binding:user-2", "User",
+		[]string{"create", "list"},
+		"wild.wild.west", "sheriffs", "",
+	)
+	_, err = kubeClient.RbacV1().ClusterRoles().Create(logicalcluster.WithCluster(ctx, serviceProvider1Workspace), cr, metav1.CreateOptions{})
+	require.NoError(t, err)
+	_, err = kubeClient.RbacV1().ClusterRoleBindings().Create(logicalcluster.WithCluster(ctx, serviceProvider1Workspace), crb, metav1.CreateOptions{})
+	require.NoError(t, err)
+
+	t.Logf("verify that we can lists all claimed resources using a wildcard request")
```
ncdc, line 315:
```suggestion
	t.Logf("verify that user-2 can lists all claimed resources using a wildcard request")
```

test/e2e/virtual/apiexport/authorizer_test.go
```diff
@@ -0,0 +1,342 @@
...
+		for _, gvr := range claimedGVRs {
+			_, err := user2DynamicVWClient.Cluster(logicalcluster.Wildcard).Resource(gvr).List(ctx, metav1.ListOptions{})
+			if err != nil {
+				return false, fmt.Sprintf("error while waiting to list %q: %v", gvr, err)
+			}
+		}
+		return true, ""
+	}, wait.ForeverTestTimeout, 100*time.Millisecond, "listing claimed resources failed")
+
+	t.Logf("verify that we can lists all claimed resources using a request towards the tenant workspace")
```
ncdc, line 330:
```suggestion
	t.Logf("verify that user-2 can list all claimed resources in the tenant workspace")
```

test/e2e/virtual/apiexport/authorizer_test.go
```diff
@@ -0,0 +1,342 @@
...
+	framework.Eventually(t, func() (success bool, reason string) {
+		_, err = user2DynamicVWClient.Cluster(tenantWorkspace).Resource(schema.GroupVersionResource{Version: "v1", Resource: "sheriffs", Group: "wild.wild.west"}).List(ctx, metav1.ListOptions{})
+		if err != nil {
+			return false, fmt.Sprintf("error while waiting to list sherriffs: %v", err)
+		}
+		return true, ""
+	}, wait.ForeverTestTimeout, 100*time.Millisecond, "listing claimed resources failed")
+
+	_, err = user2DynamicVWClient.Cluster(tenantShadowCRDWorkspace).Resource(schema.GroupVersionResource{Version: "v1alpha1", Resource: "cowboys", Group: "wildwest.dev"}).List(ctx, metav1.ListOptions{})
```
ncdc, line 339: please add a t.Logf line before this saying what it's doing

pkg/authorization/audit_logger.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+const (
+	auditDecision = "decision"
+	auditReason   = "reason"
+)
+
+// NewAuditLogger returns an authorizer that logs every decision of the delegate authorizer
+// for the given audit prefix key.
+// Note: the prefix key must not contain a trailing slash `/`.
```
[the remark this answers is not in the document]
s-urbaniak: added 👍

pkg/authorization/audit_logger.go
```diff
@@ -0,0 +1,52 @@
...
+func NewAuditLogger(auditPrefix string, delegate authorizer.Authorizer) authorizer.Authorizer {
+	return authorizer.AuthorizerFunc(func(ctx context.Context, attr authorizer.Attributes) (authorizer.Decision, string, error) {
+		dec, reason, err := delegate.Authorize(ctx, attr)
+
+		auditReason := reason
+		if err != nil {
+			auditReason = fmt.Sprintf("reason: %q, error: %v", reason, err)
+		}
```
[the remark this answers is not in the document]
s-urbaniak: that's intentional, only for the audit log we want to append the error.

pkg/authorization/audit_logger.go
s-urbaniak: same as above

pkg/virtual/apiexport/authorizer/maximal_permission_policy_test.go
```diff
@@ -0,0 +1,333 @@
...
+						},
+					},
+				}, nil
+			},
+
+			expectedDecision: authorizer.DecisionAllow,
+			expectedReason:   `unclaimed resource in API export: "fooExport", workspace :"someWorkspace"`,
+		},
+		{
+			name: "claimed identity without idenity hash",
```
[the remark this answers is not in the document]
s-urbaniak: fixed

test/e2e/virtual/apiexport/authorizer_test.go
```diff
@@ -0,0 +1,342 @@
...
+		Spec: apisv1alpha1.APIBindingSpec{
+			Reference: apisv1alpha1.ExportReference{
+				Workspace: &apisv1alpha1.WorkspaceExportReference{
+					Path:       serviceProvider2Workspace.String(),
+					ExportName: "today-cowboys",
+				},
+			},
+		},
+	}
+	framework.Eventually(t, func() (bool, string) {
```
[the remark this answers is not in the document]
s-urbaniak: This is done in line 228 https://github.com/kcp-dev/kcp/pull/1882/files/674e9f71643a5fe4e8b849cc8cb12d98e7e4f4ca#diff-8d6f5573691c6741b1615098ed0f51a30afa2e1e9a906f19d5cbfceaef754930R228 but I moved it down so it has better locality.

test/e2e/virtual/apiexport/authorizer_test.go
```diff
@@ -0,0 +1,342 @@
...
+		if strings.Contains(condition.Message, `overlaps with "cowboys.wildwest.dev" CustomResourceDefinition`) {
+			return true, ""
+		}
+		return false, fmt.Sprintf("CRD conflict condition not yet met: %q", condition.Message)
+	}, wait.ForeverTestTimeout, time.Millisecond*100, "api binding creation failed")
+
+	user3wildwestClusterClient, err := wildwestclientset.NewForConfig(framework.UserConfig("user-3", rest.CopyConfig(cfg)))
+	require.NoError(t, err, "failed to construct wildwest cluster client for server")
+	cowboy := newCowboy("default", "cowboy1")
+	_, err = user3wildwestClusterClient.WildwestV1alpha1().Cowboys("default").Create(logicalcluster.WithCluster(ctx, tenantWorkspace), cowboy, metav1.CreateOptions{})
```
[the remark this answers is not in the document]
s-urbaniak: done

test/e2e/virtual/apiexport/authorizer_test.go
```diff
@@ -0,0 +1,342 @@
...
+	require.NoError(t, err, "error creating cowboy in tenant workspace %q", tenantWorkspace)
+
+	_, err = user3wildwestClusterClient.WildwestV1alpha1().Cowboys("default").Create(logicalcluster.WithCluster(ctx, tenantShadowCRDWorkspace), cowboy, metav1.CreateOptions{})
```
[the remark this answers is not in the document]
s-urbaniak: done

test/e2e/virtual/apiexport/authorizer_test.go
s-urbaniak: oh yes, fixed

test/e2e/virtual/apiexport/authorizer_test.go
```diff
@@ -0,0 +1,342 @@
...
+	user2ApiExportVWCfg.Host = apiExport.Status.VirtualWorkspaces[0].URL
+	user2DynamicVWClient, err := kcpdynamic.NewClusterDynamicClientForConfig(user2ApiExportVWCfg)
+
+	_, err = user2DynamicVWClient.Cluster(logicalcluster.Wildcard).Resource(schema.GroupVersionResource{Version: "v1", Resource: "sheriffs", Group: "wild.wild.west"}).List(ctx, metav1.ListOptions{})
+	require.ErrorContains(
+		t, err,
+		`sheriffs.wild.wild.west is forbidden: User "user-2" cannot list resource "sheriffs" in API group "wild.wild.west" at the cluster scope: access denied`,
+		"user-2 must not be allowed to list sheriff resources")
+
+	_, err = user2DynamicVWClient.Cluster(logicalcluster.Wildcard).Resource(schema.GroupVersionResource{Version: "v1", Resource: "configmaps"}).List(ctx, metav1.ListOptions{})
```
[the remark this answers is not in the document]
s-urbaniak: done in a couple of lines above but moved down for better locality.

test/e2e/virtual/apiexport/authorizer_test.go
s-urbaniak: fixed

test/e2e/virtual/apiexport/authorizer_test.go
s-urbaniak: fixed

pkg/authorization/audit_logger.go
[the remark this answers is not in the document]
s-urbaniak: oh but it shadowed the const, renamed to `auditReasonMsg`

s-urbaniak:
final commit addresses e2e failure:
```
    apibinding.go:64: 
        	Error Trace:	apibinding.go:64
        	            				authorizer_test.go:132
        	            	apibindings.apis.kcp.dev "root-e2e-org-n65lh-provider-a" is forbidden: unable to create APIImport: missing verb='bind' permission on apiexports
        	Test:       	TestAPIExportAuthorizers
        	Messages:   	error creating APIBinding root:e2e-org-n65lh:tenant|root-e2e-org-n65lh-provider-a
```

We create role binding just before API creation so API binding creation must be retried.

PR #1883, opened 2022-09-02
KCP: stop messing with the RawPath

p0lyn0mial:
Generally, the RawPath is only available when some chars have been encoded. The Path always fully expands the URL.

Since we trim the Path, naturally we would like also to trim the RawPath. To do that correctly we would have to consider encoded slash char as well (%2F).

For example the code breaks on %2Fclusters%2F*/apis/apis.kcp.dev/v1alpha1/apiexports

Because we don't really need to deal with encoded chars and we don't use the RawPath only the Path we could simplify.
This PR simply recomputes the URL after trimming and overwrites it in the HTTP requests.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/server/handler.go
+++ b/pkg/server/handler.go
@@ -129,17 +129,15 @@ func WithClusterScope(apiHandler http.Handler) http.HandlerFunc {
 			}
 			clusterName, path = logicalcluster.New(path[:i]), path[i:]
 			req.URL.Path = path
-			for i := 0; i < 2 && len(req.URL.RawPath) > 1; i++ {
-				slash := strings.Index(req.URL.RawPath[1:], "/")
-				if slash == -1 {
-					responsewriters.ErrorNegotiated(
-						apierrors.NewInternalError(fmt.Errorf("unable to parse cluster when shortening raw path, have clusterName=%q, rawPath=%q", clusterName, req.URL.RawPath)),
-						errorCodecs, schema.GroupVersion{},
-						w, req)
-					return
-				}
-				req.URL.RawPath = req.URL.RawPath[slash:]
+			newURL, err := url.Parse(req.URL.String())
+			if err != nil {
+				responsewriters.ErrorNegotiated(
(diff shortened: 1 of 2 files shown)


pkg/server/handler.go
```diff
@@ -129,17 +129,15 @@ func WithClusterScope(apiHandler http.Handler) http.HandlerFunc {
...
-						apierrors.NewInternalError(fmt.Errorf("unable to parse cluster when shortening raw path, have clusterName=%q, rawPath=%q", clusterName, req.URL.RawPath)),
-						errorCodecs, schema.GroupVersion{},
-						w, req)
-					return
-				}
-				req.URL.RawPath = req.URL.RawPath[slash:]
+			newURL, err := url.Parse(req.URL.String())
+			if err != nil {
+				responsewriters.ErrorNegotiated(
+					apierrors.NewInternalError(fmt.Errorf("unable to resolve %s, err %w", req.URL.Path, err)),
```
p0lyn0mial, line 135: it is InternalError because it is on us. Not sure if we want to expose all the details to an end user.

pkg/server/handler.go
```diff
@@ -129,17 +129,15 @@ func WithClusterScope(apiHandler http.Handler) http.HandlerFunc {
...
-				slash := strings.Index(req.URL.RawPath[1:], "/")
-				if slash == -1 {
-					responsewriters.ErrorNegotiated(
-						apierrors.NewInternalError(fmt.Errorf("unable to parse cluster when shortening raw path, have clusterName=%q, rawPath=%q", clusterName, req.URL.RawPath)),
-						errorCodecs, schema.GroupVersion{},
-						w, req)
-					return
-				}
-				req.URL.RawPath = req.URL.RawPath[slash:]
+			newURL, err := url.Parse(req.URL.String())
```
p0lyn0mial, line 132: since we don't really use the RawPath we could simply overwrite the Path and leave the RawPath untouched.

PR #1884, opened 2022-09-02
Update usage text for virtual-workspaces to be accurate

hasheddan:
## Summary

Updates the usage text for the virtual-workspaces command to be accurate
with current status.

## Related issue(s)

n/a -- good to see you folks :)

### DIFF
--- a/cmd/virtual-workspaces/command/cmd.go
+++ b/cmd/virtual-workspaces/command/cmd.go
@@ -55,8 +55,8 @@ func NewCommand(ctx context.Context, errout io.Writer) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "workspaces",
-		Short: "Launch workspaces virtual workspace apiserver",
-		Long:  "Start a virtual workspace apiserver to managing personal, shared or organization workspaces",
+		Short: "Launch virtual workspace apiservers",
+		Long:  "Start the root virtual workspace apiserver to enable virtual workspace management.",
 
 		RunE: func(c *cobra.Command, args []string) error {
 			if err := opts.Logs.ValidateAndApply(kcpfeatures.DefaultFeatureGate); err != nil {


cmd/virtual-workspaces/command/cmd.go
```diff
@@ -55,8 +55,8 @@ func NewCommand(ctx context.Context, errout io.Writer) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "workspaces",
-		Short: "Launch workspaces virtual workspace apiserver",
-		Long:  "Start a virtual workspace apiserver to managing personal, shared or organization workspaces",
+		Short: "Launch virtual workspace apiserver",
+		Long:  "Start a virtual workspace apiserver to manage personal, shared or organization workspaces",
```
sttts, line 59: actually this line is long outdated.

sttts, line 59:
```suggestion
		Short: "Launch virtual workspace apiservers",
		Long:  "Start virtual workspace apiserver for different purposes.",
```

hasheddan: That should future-proof it for a while :)

sttts: not sure this is so much better. It runs these https://github.com/kcp-dev/kcp/tree/main/pkg/virtual: syncer, workspaces, apiexport, initializingworkspaces. More to come.

hasheddan: @sttts what would you think about making the `Long` description say something about starting the root API server to be a bit more specific?

sttts: Fine with me. Wondering whether we can get the list easily somehow.

sttts: It's probably hidden deep in the constructor code.

hasheddan: > Wondering whether we can get the list easily somehow.

Seems like a great thing to do in a codegen step 👍🏻 I'll update to just mention starting the root API server for now -- happy to open an issue tracking automating listing components in usage if that would be desirable

cmd/virtual-workspaces/command/cmd.go
```diff
@@ -55,8 +55,8 @@ func NewCommand(ctx context.Context, errout io.Writer) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "workspaces",
-		Short: "Launch workspaces virtual workspace apiserver",
-		Long:  "Start a virtual workspace apiserver to managing personal, shared or organization workspaces",
+		Short: "Launch virtual workspace apiservers",
+		Long:  "Start the root virtual workspace apiserver to enable virtual workspace management.",
```
hasheddan, line 59: @sttts updated here -- happy to do more iteration if this feels under-specified or inaccurate

cmd/virtual-workspaces/command/cmd.go
sttts: :+1: to both

cmd/virtual-workspaces/command/cmd.go
sttts: It's not super important clearly. If you run this command manually, you know what you are doing.

davidfestal (approved):

PR #1892, opened 2022-09-06
docs/architecture: remove because totally outdated

sttts:
These docs are horribly outdated and confuse people more than they help.


hardys:
@sttts I know how hard it is to keep this kind of content updated in a fast-moving project, but it's enormously helpful to have architecture overview diagrams etc to help with onboarding (speaking for myself, but also more generally :) ) - are there any plans to revive this in future, or is the overhead just considered too high?

PR #1896, opened 2022-09-06
📝 documents the cache-server

p0lyn0mial:
## Summary
describes the cache-server, atm under `docs/cache-server.md`, could be moved to `kcp/pkg/cache/server` or `docs/compontents`.

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/1865

### DIFF
--- a/docs/cache-server.md
+++ b/docs/cache-server.md
@@ -1,0 +1,131 @@
+### Purpose
+
+The primary purpose of the cache server is to support cross-shard communication.
+Shards need to communicate with each other in order to enable, among other things, migration and replication features.
+Direct shard-to-shard communication is not feasible in a larger setup as it scales with n*(n-1).
+The web of connections would be hard to reason about, maintain and troubleshoot. 
+Thus, the cache server serves as a central place for shards to store and read common data.
+Note, that the final topology can have more than one cache server, i.e. one in some geographic region.
+
+### High-level overview
+
+The cache server is a regular, Kubernetes-style CRUD API server with support of LIST/WATCH semantics. 
+Conceptually it supports two modes of operations. 
+The first mode is in which a shard gets its private space for storing its own data. 
+The second mode is in which a shard can read other shards' data.
+
+The first mode of operation is implemented as a write controller that runs on a shard.
+It holds some state/data from that shard in memory and pushes it to the cache server.
+The controller uses standard informers for getting required resources both from a local kcp instance and remote cache server.
+Before pushing data it has to compare remote data to its local copy and make sure both copies are consistent.
+
+The second mode of operation is implemented as a read controller(s) that runs on some shard. 
+It holds other shards' data in an in-memory cache using an informer, 
+effectively pulling data from the central cache.
+Thanks to having a separate informer for interacting with the cache server 
+a shard can implement a different resiliency strategy.
+For example, it can tolerate the unavailability of the secondary during startup and become ready.
+
+### Running the server
+
+The cache server can be run as a standalone binary or as part of a kcp server.
+
+The standalone binary is in https://github.com/kcp-dev/kcp/tree/main/cmd/cache-server and can be run by issuing `go run ./cmd/cache-server/main.go` command.
+
+To run it as part of a kcp server, pass `--cache-url` flag to the kcp binary.
+
+### Client-side functionality
+
+In order to interact with the cache server from a shard, the https://github.com/kcp-dev/kcp/tree/main/pkg/cache/client 
+repository provides the following client-side functionality:
+
+[ShardRoundTripper](https://github.com/kcp-dev/kcp/blob/b739fa5b5c83fb2c43b631c6234264d5dd1fc6e4/pkg/cache/client/round_tripper.go#L54), 
+a shard aware wrapper around `http.RoundTripper`. It changes the URL path to target a shard from the context.
+
+[DefaultShardRoundTripper](https://github.com/kcp-dev/kcp/blob/b739fa5b5c83fb2c43b631c6234264d5dd1fc6e4/pkg/cache/client/round_tripper.go#L128) 
+is a `http.RoundTripper` that sets a default shard name if not specified in the context.
+
+For example, in order to make a client shard aware, inject the `http.RoundTrippers` to a `rest.Config`
+
+```go
+import (
+  cacheclient "github.com/kcp-dev/kcp/pkg/cache/client"
+  "github.com/kcp-dev/kcp/pkg/cache/client/shard"
+)
+
+// adds shards awareness to a client with a default `shard.Wildcard` name.
+NewClusterForConfig(cacheclient.WithShardRoundTripper(cacheclient.WithDefaultShardRoundTripper(serverConfig.LoopbackClientConfig, shard.Wildcard)))
+
+// and then change the context when you need to access a specific shard and pass is when making a HTTP request
+ctx = cacheclient.WithShardInContext(ctx, shard.New("cache"))
+```
+
+### Authorization/Authentication
+
+Not implemented at the moment
+
+### Built-in resources
+
+Out of the box, the server supports the following resources:
+- `apiresourceschemas`
+- `apiexports`
+
+All those resources are represented as CustomResourceDefinitions and 
+stored in `system:cache:server` shard under `system:system-crds` cluster.
+
+### Adding new resources
+
+Not implemented at the moment.
+Our near-term plan is to maintain a list of hard-coded resources that we want to keep in the cache server. 
+In the future, we will use the ReplicationClam which will describe schemas that need to be exposed by the cache server.
+
+### Deletion of data
+
+Not implemented at the moment.
+Only deleting resources explicitly is possible.
+In the future, some form of automatic removal will be implemented.
+
+### Design details
+
+The cache server is implemented as the `apiextensions-apiserver`. 
+It is based on the same fork used by the kcp server, extended with shard support.
+
+Since the server serves as a secondary replica it doesn't support versioning, validation, pruning, or admission. 
+All resources persisted by the server are deprived of schema. 
+That means the schema is implicit, maintained, and enforced by the shards pushing/pulling data into/from the server.
+
+#### On the HTTP level 
+
+The server exposes the following path:
+
+`/shards/{shard-name}/clusters/{cluster-name}/apis/group/version/namespaces/{namespace-name}/resource/{resource-name}`
+
+Parameters:
+
+`{shard-name}`: a required string holding a shard name, a wildcard is also supported indicating a cross-shard request
+
+`{cluster-name}`: a required string holding a cluster name, a wildcard is also supported indicating a cross-cluster request.
+
+`{namespace-name}`: an optional string holding a namespace the given request is for, not all resources are stored under a namespace.
+
+`{resource-name}`: an optional string holding the name of a resource
+
+For example:
+
+`/shards/*/clusters/*/apis/apis.kcp.dev/v1alpha1/apiexports`: for listing apiexports for all shards and clusters
+
+`/shards/amber/clusters/*/apis/apis.kcp.dev/v1alpha1/apiexports`: for listing apiexports for amber shard for all clusters
+
+`/shards/sapphire/clusters/system:sapphire/apis/apis.kcp.dev/v1alpha1/apiexports`: for listing apiexports for sapphire shard stored in system:sapphire cluster
+
+#### On the storage layer
+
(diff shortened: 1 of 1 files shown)


docs/cache-server.md
```diff
@@ -0,0 +1,125 @@
...
+
+### High-level overview
+
+The cache server is a regular, Kubernetes-style CRUD API server with support of LIST/WATCH semantics. 
+Conceptually it supports two modes of operations. 
+The first mode is in which a shard gets its private space for storing its own data. 
+The second mode is in which a shard can read other shards' data.
+
+The first mode of operation is implemented as a controller that runs within a shard. 
+It maintains an informer for `LIST/WATCH`ing its own resources.
```
stevekuznetsov, line 17: Could we make this section more clear about the flow of data? Controller runs on a shard, holds some state from that shard in memory, and pushes it to the central cache?

docs/cache-server.md
```diff
@@ -0,0 +1,125 @@
+### Purpose
+
+The primary purpose of the cache server is to support cross-shard communication.
+Shards need to communicate with each other in order to enable, among other things, migration and replication features.
+Direct shard-to-shard communication is not feasible in a larger setup.
```
stevekuznetsov, line 5:
```suggestion
Direct shard-to-shard communication is not feasible in a larger setup as it scales with n^2.
```

stevekuznetsov: (or some other way to indicate *why* it's not feasible)

docs/cache-server.md
```diff
@@ -0,0 +1,125 @@
...
+The cache server is a regular, Kubernetes-style CRUD API server with support of LIST/WATCH semantics. 
+Conceptually it supports two modes of operations. 
+The first mode is in which a shard gets its private space for storing its own data. 
+The second mode is in which a shard can read other shards' data.
+
+The first mode of operation is implemented as a controller that runs within a shard. 
+It maintains an informer for `LIST/WATCH`ing its own resources. 
+It compares received data to its local copy and makes sure both copies are consistent.
+
+The second mode of operation is implemented as a separate informer within a shard.
```
stevekuznetsov, line 20: Also here - this is a controller, on some shard, holding other shards' data in an in-memory cache using informer, pulling data from the central cache

docs/cache-server.md
```diff
@@ -0,0 +1,125 @@
+### Purpose
+
+The primary purpose of the cache server is to support cross-shard communication.
+Shards need to communicate with each other in order to enable, among other things, migration and replication features.
+Direct shard-to-shard communication is not feasible in a larger setup.
+The web of connections would be hard to reason about, maintain and troubleshoot. 
+Thus, the cache server serves as a central place for shards to store and read common data.
```
stevekuznetsov, line 7: Central, in a region?

docs/cache-server.md
```diff
@@ -0,0 +1,125 @@
...
+
+[ShardRoundTripper](https://github.com/kcp-dev/kcp/blob/b739fa5b5c83fb2c43b631c6234264d5dd1fc6e4/pkg/cache/client/round_tripper.go#L54), 
+a shard aware wrapper around `http.RoundTripper`. It changes the URL path to target a shard from the context.
+
+[DefaultShardRoundTripper](https://github.com/kcp-dev/kcp/blob/b739fa5b5c83fb2c43b631c6234264d5dd1fc6e4/pkg/cache/client/round_tripper.go#L128) 
+is a `http.RoundTripper` that sets a default shard name if not specified in the context.
+
+For example, in order to make a client shard aware, inject the `http.RoundTrippers` to a `rest.Config`
+
+```
```
stevekuznetsov, line 45:
```suggestion
```go
```

docs/cache-server.md
```diff
@@ -0,0 +1,125 @@
...
+import(
+cacheclient "github.com/kcp-dev/kcp/pkg/cache/client"
```
stevekuznetsov, line 47: run `go fmt` / provide an example snippet here that someone could copy-paste please

docs/cache-server.md
```diff
@@ -0,0 +1,125 @@
...
+
+// and then change the context when you need to access a specific shard and pass is when making a HTTP request
+ctx = cacheclient.WithShardInContext(ctx, shard.New("cache"))
+```
+
+### Authorization/Authentication
+
+Not implemented at the moment
+
+### Build-in resources
```
stevekuznetsov, line 62:
```suggestion
### Built-in resources
```

docs/cache-server.md
```diff
@@ -0,0 +1,125 @@
...
+
+Out of the box, the server supports the following resources:
+- `apiresourceschemas`
+- `apiexports`
+
+All those resources are stored in `system:cache:server` shard under `system:system-crds` cluster.
+
+### Adding new resources
```
stevekuznetsov, line 70: Do we expect the built-in ones to remain built-in forever? Could we instead have a section here on "how do you add data to the cache server?" and lay out the near-term (hard-code) and long-term (replication claim) visions?

docs/cache-server.md
```diff
@@ -0,0 +1,125 @@
...
+Not implemented at the moment.
+Only deleting resources explicitly is possible.
+In the future, some form of automatic removal will be implemented.
+
+### Design details
+
+The cache server is implemented as the `apiextensions-apiserver`. 
+It is based on the same fork used by the kcp server, extended with shard support.
+
+Since the server serves as a secondary replica it doesn't support versioning, validation, pruning, or admission.
```
stevekuznetsov, line 87: Do we expect this to be the case forever? How do we handle normal client-go SDK clients asking for specific versions and getting something else back from the cache server?

stevekuznetsov: Do we even support discovery on cache server? Will we?

docs/cache-server.md
```diff
@@ -0,0 +1,125 @@
...
+
+`{cluster-name}`: a required string holding a cluster name, a wildcard is also supported indicating a cross-cluster request.
+
+`{namespace-name}`: an optional string holding a namespace the given request is for, not all resources are stored under a namespace.
+
+`{resource-name}`: an optional string holding the name of a resource
+
+For example:
+
+`/shards/*/clusters/*/apis/apis.kcp.dev/v1alpha1/apiexports`: for listing apiexports for all shards and clusters
```
stevekuznetsov, line 109: Do we want to leave a section here for figuring out what `*` maps to? How does a client know if/when a constituent shard is missing from `*` or the data they see is incomplete? Even if we don't know exactly how yet, good to call out that this will in the future be a functionality from the server

docs/cache-server.md
p0lyn0mial: > Do we expect the built-in ones to remain built-in forever?

I think so, some resource like `apiresourceschemas`, `apiexports` or `secrets`, `configmaps` will always be required/availiable

docs/cache-server.md
p0lyn0mial: > Do we expect this to be the case forever? How do we handle normal client-go SDK clients asking for specific versions and getting something else back from the cache server?

versioning is implicit, imposed when shards add data to the cache.

docs/cache-server.md
p0lyn0mial: I don't have a complete example. I will add one, once we will write a controller consuming data from the cache.

docs/cache-server.md
p0lyn0mial: clarified.

docs/cache-server.md
p0lyn0mial: > Even if we don't know exactly how yet, good to call out that this will in the future be a functionality from the server

I'm not sure if this will be on the cache server or some other component. I think that shards consuming data from the cache (via an informer) will have the same issue.

p0lyn0mial:
updated the PR, PTAL, thx!

docs/cache-server.md
p0lyn0mial: updated.

docs/cache-server.md
stevekuznetsov: This is fundamentally a question of how stale the data in this particular cache server is - what other component could provide this?

docs/cache-server.md
stevekuznetsov: Are we certain about this?

stevekuznetsov (commented):
LGTM, only two more questions

docs/cache-server.md
p0lyn0mial: oh, when it comes to freshens of data pushed by the other shards, I briefly mentioned it in the "deletion of data" section. 

Currently, we don't delete outdated data. In the future, some form of a TTL or an auxiliary data structure could determine when was the last time a shard pushed something to the cache server in runtime. Based on that we could identify resources or even all data for a particular shard that needs to be removed.

docs/cache-server.md
p0lyn0mial: maybe @sttts could confirm that

docs/cache-server.md
stevekuznetsov: A user performs a `*`-shard, `*`-cluster LIST from the cache server. They expect to get all data. One of the shards has not heartbeat/is not connected to the central cache, so the LIST does not contain data from this shard. The client must have some mechanism for understanding that the results they got from the LIST are not complete, no?

p0lyn0mial: > They expect to get all data. One of the shards has not heartbeat/is not connected to the central cache

this could be addressed by some form of a TTL attached to the shard's resources or an auxiliary data structure that could help determine "the last time this particular shard did something on its data was 1 day ago, we should remove stale data"

a shard not connecting to a cache at all is like a controller in kube not doing its work, you need to monitor the shard/controller. After all, you have no way of knowing what data a shard will push to the registry.

docs/cache-server.md
```diff
@@ -0,0 +1,130 @@
...
+// adds shards awareness to a client with a default `shard.Wildcard` name.
+NewClusterForConfig(cacheclient.WithShardRoundTripper(cacheclient.WithDefaultShardRoundTripper(serverConfig.LoopbackClientConfig, shard.Wildcard)))
+
+// and then change the context when you need to access a specific shard and pass is when making a HTTP request
+ctx = cacheclient.WithShardInContext(ctx, shard.New("cache"))
+```
+
+### Authorization/Authentication
+
+Not implemented at the moment
```
sttts, line 65: do we have an issue for this?

docs/cache-server.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Built-in resources
+
+Out of the box, the server supports the following resources:
+- `apiresourceschemas`
+- `apiexports`
+
+All those resources are stored in `system:cache:server` shard under `system:system-crds` cluster.
```
sttts, line 73: what does this sentence mean? What are these used for? Are you saying we make these (system-)CRDs available by default? What about other CRDs?

And this sentence is just about the CRDs, right, not about the CRs (it sounds like that) ?

docs/cache-server.md
sttts: yes, there is expected to be a consensus (non-technically, but socially during development) what do store in the case and in which version (only one). Eventually, we have to think more about how to transition to a new version. For now the data there is intentionally basically untyped (= schema), i.e. no pruning, no validation, ... If we could, I would even like to avoid any CRD to be created before we can serve something, i.e. much more like KV stores work where you can write what you want and the clients have to agree on keys and values.

docs/cache-server.md
sttts: Totally agree that we have to come back here for staleness information and ttl logic. Would keep it simple for now in the first step, to see whether it is a viable approach. Basically for now we assume that eventually the cache data is complete. We definitely need more after a first proof of concept.

sttts:
Sgtm for now. Leaving lgtm to Steve. This described the first step. We will definitely come back and extend the mechanism and then doc it here.

/approve

docs/cache-server.md
p0lyn0mial: added a todo step to https://github.com/kcp-dev/kcp/issues/1225

docs/cache-server.md
p0lyn0mial: yes, it is about CRDs, will clarify

docs/cache-server.md
stevekuznetsov: Yeah I meant more like a section/header/callout that we expect to handle this case, in the cache server, perhaps with some auxiliary data structures, but at the very least we expect this to be part of the design and handling it is on the cache server

stevekuznetsov: (we can add this later if you'd like)

PR #1915, opened 2022-09-08
verify-go-versions.sh make CONTRIBUTING.md check conditional

hardys:
## Summary

This check breaks the Dockerfile build as the file isn't COPY'd

## Related issue(s)

This was introduced via #1894 but I didn't notice that the `verify-go-versions` Makefile target is run by the default all/build target in the Dockerfile


hardys:
/cc @sttts I'm wondering if `verify-go-versions` should perhaps be part of the lint target instead, then we could avoid all the related copies in the Dockerfile, is there a reason it's run via `build` atm?

But anyway, this fixes the issue caused by my previous PR, sorry for that! :)

PR #1926, opened 2022-09-09
cache-server: adds ShardNameFromObjectRoundTripper

p0lyn0mial:
## Summary
Knows how to read a shard name from an object and store it in the context. 
It should be only used by the internal kube-clients that are not aware of a shard name.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cache/client/round_tripper.go
+++ b/pkg/cache/client/round_tripper.go
@@ -17,10 +17,14 @@ limitations under the License.
 package client
 
 import (
+	"bytes"
 	"net/http"
 	"regexp"
 	"strings"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/rest"
 
 	clientshard "github.com/kcp-dev/kcp/pkg/cache/client/shard"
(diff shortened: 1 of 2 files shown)


pkg/cache/client/round_tripper.go
```diff
@@ -144,3 +148,73 @@ func (c *DefaultShardRoundTripper) RoundTrip(req *http.Request) (*http.Response,
 	}
 	return c.delegate.RoundTrip(req)
 }
+
+// WithShardNameFromObjectRoundTripper wraps an existing config's with ShardNameFromObjectRoundTripper.
```
sttts, line 152: config's ?

sttts, line 152:
```suggestion
// WithShardNameFromObjectRoundTripper wraps an existing config's roudtripper with ShardNameFromObjectRoundTripper.
```

sttts:
Should we reference https://github.com/kcp-dev/kubernetes/pull/100/files#diff-11607ac007eca978cf25bd87a6ec355cd2b4e04dbe0df698a2e892a5b801d8aeR30 instead?

p0lyn0mial:
> Should we reference https://github.com/kcp-dev/kubernetes/pull/100/files#diff-11607ac007eca978cf25bd87a6ec355cd2b4e04dbe0df698a2e892a5b801d8aeR30 instead?

we could, i don't like the idea that the client pkg is required to depend on a pkg that belongs to the server, but if you think otherwise I can change it

p0lyn0mial:
> we could, I don't like the idea that the client pkg is required to depend on a pkg that belongs to the server, but if you think otherwise I can change it

given the above, I'm merging this pr, can be easily changed in the future if somebody doesn't like it.

/hold cancel

stevekuznetsov:

PR #1928, opened 2022-09-09
🌱 cache: makes all internal clients shard and cluster aware

p0lyn0mial:
## Summary
provides an ordered list of HTTP round-trippers that add shard and cluster awareness to all clients that use the loopback config.

requires https://github.com/kcp-dev/kcp/pull/1926 and https://github.com/kcp-dev/kubernetes/pull/100

## Related issue(s)

Fixes #


stevekuznetsov:

Issue #1891, opened 2022-09-05
bug: Infinite loop of reconciles when finalizer applied to deleting resource 


philbrookes:
### Describe the bug

In the scenario where a synctarget is set to use advanced scheduling, and a resource has been created with a finalizer, and that resource is then deleted, the resource will reconcile (adding and removing the deletion timestamp annotation) so rapidly that it is impossible for another controller to remove the finalizer, and so preventing the deletion of this resource.

Tested and confirmed on ingresses, services and deployments, so I assume it affects all resources.

### Steps To Reproduce

1. Create synctarget with advanced scheduling enabled.
2. Create a synced resource with a regular finalizer
3. Delete the resource
4. Watch the resource: (e.g. `kubectl get <resource> <resource-name> -w -o yaml | yq eval ".metadata.resourceVersion" -`)
5. See very rapid updates being applied to the resource
6. This appears to be related to the deletion timestamp (`deletion.internal.workload.kcp.dev/...`) annotation being added and removed rapidly, see the output from: `kubectl get <resource> <ingress-name> -w -o yaml | yq eval ".metadata.annotations" -`

### Expected Behaviour

I would expect the resource to have the deletion timestamp annotation permanently and not be reconciled so aggressively - allowing my controller to remove the finalizer from the resource.

### Additional Context

This is occurring in 0.7.0 and 0.8.0

philbrookes:
After further investigation, this appears to happen regardless of whether or not the sync target has advanced scheduling enabled.

stevekuznetsov:
Sounds like this is not a well-supported use-case today. Filed as a bug for 0.9

PR #1929, opened 2022-09-09
🐛 Only propagate the deletionTimestamp to a location if there is a syncer finalizer for it.

jmprusi:
## Summary

The deletionTimestamp propagation should only happen if there is a syncer finalizer holding that location.

This PR fixes a situation where the syncer and the resource controller will enter a reconciliation storm, due to the resource controller propagating the deletion timestamp of a resource, and the syncer removing its location deletionTimestamp as it has been removed from the downstream cluster.

## Related issue(s)
Fixes https://github.com/kcp-dev/kcp/issues/1891

### DIFF
--- a/pkg/reconciler/workload/resource/resource_reconcile.go
+++ b/pkg/reconciler/workload/resource/resource_reconcile.go
@@ -151,15 +151,28 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName logical
 func propagateDeletionTimestamp(logger logr.Logger, obj metav1.Object, annotationPatch map[string]interface{}) map[string]interface{} {
 	logger.V(3).Info("resource is being deleted; setting the deletion per locations timestamps")
 	objAnnotations := obj.GetAnnotations()
+	objFinalizers := obj.GetFinalizers()
 	objLocations, _ := locations(objAnnotations, obj.GetLabels(), false)
 	if annotationPatch == nil {
 		annotationPatch = make(map[string]interface{})
 	}
 	for location := range objLocations {
+		hasFinalizerForLocation := false
+		for _, finalizer := range objFinalizers {
+			if finalizer == syncershared.SyncerFinalizerNamePrefix+location {
+				hasFinalizerForLocation = true
+				break
+			}
+		}
+		// If the object doesn't have a syncer finalizer for the given location, we don't need to set the deletion timestamp.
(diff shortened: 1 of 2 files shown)


pkg/reconciler/workload/resource/resource_reconcile.go
```diff
@@ -151,15 +151,29 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName logical
...
+		// If the object doesn't have a syncer finalizer for the given location, we don't need to set the deletion timestamp.
+		if !hasFinalizerForLocation {
+			break
+		}
 		if val, ok := objAnnotations[workloadv1alpha1.InternalClusterDeletionTimestampAnnotationPrefix+location]; !ok || val == "" {
 			annotationPatch[workloadv1alpha1.InternalClusterDeletionTimestampAnnotationPrefix+location] = obj.GetDeletionTimestamp().Format(time.RFC3339)
 		}
 	}
+
+	logger.Info("resource", "annotations", obj.GetAnnotations(), "finalizers", obj.GetFinalizers(), "labels", obj.GetLabels(), "patch", annotationPatch)
```
stevekuznetsov, line 176: Logging all annotations & labels is going to be pretty verbose. Can we extract out the important bits?

jmprusi: let's remove it,just added it for debugging.

stevekuznetsov (approved):
/hold

This seems reasonable. Deferring to @davidfestal for merge

davidfestal:
Isn't there a risk that by fixing the problem this way on the resource controller side, we may loose the "intent to remove from a location" information ?
The fact that no syncer finalizer exist for a given SynTarget only means that the corresponding Syncer still didn't take ownership of the resource at this time, but it might be that the Syncer is previsl in the process of adding this finalizer.

I'm wondering whether the fix should rather be done on the Syncer side, to change how we react to the presence of the deletion annotation.

davidfestal:
After discussing about it, it seems that a more proper fix in the resource controller would be, at least as a first step, to avoid scheduling a KCP resource at all (== setting the `RequestState` label) if it has a deletion timestamp.

jmprusi:
Closed in favor of https://github.com/kcp-dev/kcp/pull/1953 as not propagating the deletionTimestamp can have some unintended consequences.

PR #1947, opened 2022-09-12
:seedling: cache: moves common HTTP handlers to a shared pkg

p0lyn0mial:
## Summary
It moves common HTTP handlers to a separate package.
Those handlers are used by the kcp-server and the cache-server.
Since we are going to run the cache from the kcp.
Importing will cause a cycle. 
Thus common functionality is moved to a common package.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -57,6 +57,7 @@ import (
 	"github.com/kcp-dev/kcp/pkg/indexers"
 	"github.com/kcp-dev/kcp/pkg/informer"
 	bootstrap "github.com/kcp-dev/kcp/pkg/server/bootstrap"
+	kcpfilters "github.com/kcp-dev/kcp/pkg/server/filters"
 	kcpserveroptions "github.com/kcp-dev/kcp/pkg/server/options"
 	"github.com/kcp-dev/kcp/pkg/server/options/batteries"
 	"github.com/kcp-dev/kcp/pkg/server/requestinfo"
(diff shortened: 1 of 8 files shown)


sttts:
/hold

Wrong direction. We should not share the handler chain between kcp and the caching server. Why should we?

pkg/server/handler.go
```diff
@@ -184,19 +78,6 @@ func WithAuditAnnotation(handler http.Handler) http.HandlerFunc {
 	})
 }
 
-// WithClusterAnnotation adds the cluster name into the annotation of an audit
-// event. Needs initialized annotations.
-func WithClusterAnnotation(handler http.Handler) http.HandlerFunc {
```
sttts, line 189: this needs a better name. WithAuditEventClusterAnnotation

pkg/server/config.go
```diff
@@ -322,12 +323,12 @@ func NewConfig(opts *kcpserveroptions.CompletedOptions) (*Config, error) {
...
-		apiHandler = WithClusterAnnotation(apiHandler)
+		apiHandler = util.WithClusterAnnotation(apiHandler)
 		apiHandler = WithAuditAnnotation(apiHandler) // Must run before any audit annotation is made
-		apiHandler = WithClusterScope(apiHandler)
+		apiHandler = util.WithClusterScope(apiHandler)
 		apiHandler = WithInClusterServiceAccountRequestRewrite(apiHandler)
-		apiHandler = WithAcceptHeader(apiHandler)
-		apiHandler = WithUserAgent(apiHandler)
+		apiHandler = util.WithAcceptHeader(apiHandler)
+		apiHandler = util.WithUserAgent(apiHandler)
```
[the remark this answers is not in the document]
p0lyn0mial: `WithUserAgent` wasn't required also was able to get rid of `WithAcceptHeader`

pkg/server/handler.go
p0lyn0mial: updated.

p0lyn0mial:
added `lgtm` label manually since I only fixed the boilterplate

stevekuznetsov:

PR #1949, opened 2022-09-12
:seedling: cache: small refactor to make wiring into kcp easier

p0lyn0mial:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cache/server/options/options.go
+++ b/pkg/cache/server/options/options.go
@@ -62,7 +62,7 @@ func (o *CompletedOptions) Validate() []error {
 	return errors
 }
 
-// NewOptions crwates a new Options with default parameters.
+// NewOptions creates a new Options with default parameters.
 func NewOptions(rootDir string) *Options {
 	o := &Options{
 		ServerRunOptions: genericoptions.NewServerRunOptions(),
--- a/pkg/cache/server/server.go
+++ b/pkg/cache/server/server.go
@@ -19,6 +19,7 @@ package server
 import (
 	"context"
 
+	apiextensionsapiserver "k8s.io/apiextensions-apiserver/pkg/apiserver"
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	"k8s.io/klog/v2"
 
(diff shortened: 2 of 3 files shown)


pkg/cache/server/options/options.go
```diff
@@ -74,26 +74,25 @@ func NewOptions(rootDir string) *Options {
...
-	o.Etcd.EnableWatchCache = false
 	return o
 }
 
 func (o *Options) Complete() (*CompletedOptions, error) {
 	if servers := o.Etcd.StorageConfig.Transport.ServerList; len(servers) == 1 && servers[0] == "embedded" {
 		o.EmbeddedEtcd.Enabled = true
 	}
-
+	o.ServerRunOptions.EnablePriorityAndFairness = false
```
sttts, line 87: why here in complete?

pkg/cache/server/options/options.go
```diff
@@ -74,26 +74,25 @@ func NewOptions(rootDir string) *Options {
...
 	// TODO: enable authN/Z stack
 	o.Authentication = nil
 	o.Authorization = nil
+	// TODO: enable the watch cache, it was disabled because
+	//  - we need to pass a shard name so that the watch cache can calculate the key
+	//    we already do that for cluster names (stored in the obj)
+	//  - we need to modify wildcardClusterNameRegex and crdWildcardPartialMetadataClusterNameRegex
+	o.Etcd.EnableWatchCache = false
```
sttts, line 95: why here? Don't think this is completion.

pkg/cache/server/server.go
```diff
@@ -37,11 +37,11 @@ func NewServer(c CompletedConfig) (*Server, error) {
 	return s, nil
 }
 
-func (s *Server) Run(ctx context.Context) error {
+func (s *Server) Complete(ctx context.Context) (*genericapiserver.GenericAPIServer, error) {
```
sttts, line 40: why? This is not completion semantics anymore.

pkg/cache/server/options/options.go
p0lyn0mial: because I need to overwrite the etcd options when the cache-server is run with the kcp-server and then turn off what needs to be turned off, for example:

```
cacheServerOptions := cacheoptions.NewOptions(s.Options.Extra.RootDirectory)

cacheServerEtcdOptions := *s.Options.GenericControlPlane.Etcd
cacheServerOptions.Etcd = &cacheServerEtcdOptions

completedCacheServerOptions, err := cacheServerOptions.Complete()
```

pkg/cache/server/server.go
p0lyn0mial: I need this intermediary step when running the cache-server with the kcp. We can change the name. 

Basically, I will:
```
newCacheServer, err := cacheserver.NewServer(completedCacheServerConfig)
if err != nil {
	return err
}
completedCacheServer, err := newCacheServer.Complete(ctx)
if err != nil {
	return err
}

preparedCacheServer := completedCacheServer.PrepareRun()
if err := s.AddPostStartHook("kcp-start-cache-server", func(ctx genericapiserver.PostStartHookContext) error {
	preparedCacheServer.RunPostStartHooks(ctx.StopCh)
	return nil
}); err != nil {
	return err
}

s.preHandlerChainMux.Handle(virtualcommandoptions.DefaultRootPathPrefix+"/cache/", preparedCacheServer.Handler)
s.preHandlerChainMux.Handle("/shards/", preparedCacheServer.Handler)

```

pkg/cache/server/options/options.go
p0lyn0mial: have a look at https://github.com/kcp-dev/kcp/pull/1954/files#diff-e3fc4d4c96ce1c7a560285f676ceafdb8b2ec8c582113f1ffce8305a2b9bec37R39

sttts: you cannot override non-pointers in Complete.

pkg/cache/server/server.go
```diff
@@ -37,11 +37,11 @@ func NewServer(c CompletedConfig) (*Server, error) {
...
 	logger := klog.FromContext(ctx).WithValues("component", "cache-server")
```
sttts, line 41: this is wrong. You have to take the logger from the context of `hookContext`.

stevekuznetsov:

PR #1953, opened 2022-09-12
🐛 Don't calculate new placements for resources that are being deleted

jmprusi:
# Summary

This PR will take into account if the deletionTimestamp of the resource is set in order to schedule new placements or not. 

This PR fixes a situation where the syncer and the resource controller will enter into a reconciliation storm, due to the resource controller re-adding the sync intention to a location once it has been removed by the syncer.

Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/1891

### DIFF
--- a/pkg/reconciler/workload/resource/resource_reconcile.go
+++ b/pkg/reconciler/workload/resource/resource_reconcile.go
@@ -72,11 +72,14 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName logical
 		return fmt.Errorf("error reconciling resource %s|%s/%s: error getting namespace: %w", lclusterName, obj.GetNamespace(), obj.GetName(), err)
 	}
 
-	annotationPatch, labelPatch := computePlacement(ns, obj)
+	var annotationPatch, labelPatch map[string]interface{}
 
 	// If the object DeletionTimestamp is set, we should set all locations deletion timestamps annotations to the same value.
 	if obj.GetDeletionTimestamp() != nil {
-		annotationPatch = propagateDeletionTimestamp(logger, obj, annotationPatch)
+		annotationPatch = propagateDeletionTimestamp(logger, obj)
+	} else {
+		// We only need to compute the new placements if the resource is not being deleted.
+		annotationPatch, labelPatch = computePlacement(ns, obj)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/reconciler/workload/resource/resource_reconcile.go
```diff
@@ -72,11 +72,14 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName logical
 		return fmt.Errorf("error reconciling resource %s|%s/%s: error getting namespace: %w", lclusterName, obj.GetNamespace(), obj.GetName(), err)
 	}
 
-	annotationPatch, labelPatch := computePlacement(ns, obj)
+	annotationPatch, labelPatch := make(map[string]interface{}), make(map[string]interface{})
```
davidfestal, line 75: I'd rather have just declared vars here without an instantiation, and instantiate them in the next conditional block.

pkg/reconciler/workload/resource/resource_reconcile.go
```diff
@@ -72,11 +72,14 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName logical
...
 
 	// If the object DeletionTimestamp is set, we should set all locations deletion timestamps annotations to the same value.
 	if obj.GetDeletionTimestamp() != nil {
 		annotationPatch = propagateDeletionTimestamp(logger, obj, annotationPatch)
```
davidfestal, line 79:
```suggestion
		annotationPatch = propagateDeletionTimestamp(logger, obj)
```

davidfestal: Seems the `annotationPatch` argument is not needed anymore.

pkg/reconciler/workload/resource/resource_reconcile.go
stevekuznetsov: Letting the map be `nil` will have a NPE when you set fields in it, no?

davidfestal (approved):

pkg/reconciler/workload/resource/resource_reconcile.go
davidfestal: Where do we update the maps in this function ? I may have missed it

PR #1954, opened 2022-09-12
🌱 Wire the cache server into the kcp server

p0lyn0mial:
## Summary
This PR allows for running the cache-server in-process with the kcp server. 
The new server is exposed under `/services/cache/...` path. 
By default, the cache server won't run with the kcp server.
In order to turn it, pass `--run-cache-server=true` flag.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cache/server/config.go
+++ b/pkg/cache/server/config.go
@@ -39,6 +39,7 @@ import (
 	genericoptions "k8s.io/apiserver/pkg/server/options"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	"k8s.io/apiserver/pkg/util/webhook"
+	"k8s.io/client-go/rest"
 	"k8s.io/kubernetes/pkg/genericcontrolplane/clientutils"
 
 	cacheclient "github.com/kcp-dev/kcp/pkg/cache/client"
@@ -87,7 +88,9 @@ func (c *Config) Complete() (CompletedConfig, error) {
 	}}, nil
 }
 
-func NewConfig(opts *cacheserveroptions.CompletedOptions) (*Config, error) {
+// NewConfig returns a new Config for the given options and optional rest.Config that point to the local server.
+// Pass it only when you combine this server with a different one.
+func NewConfig(opts *cacheserveroptions.CompletedOptions, optionalLocalShardRestConfig *rest.Config) (*Config, error) {
 	c := &Config{
 		Options: opts,
 	}
@@ -116,8 +119,15 @@ func NewConfig(opts *cacheserveroptions.CompletedOptions) (*Config, error) {
 	if err := opts.Etcd.ApplyTo(&serverConfig.Config); err != nil {
 		return nil, err
 	}
-	if err := opts.SecureServing.ApplyTo(&serverConfig.Config.SecureServing, &serverConfig.Config.LoopbackClientConfig); err != nil {
-		return nil, err
+	if optionalLocalShardRestConfig == nil {
+		if err := opts.SecureServing.ApplyTo(&serverConfig.Config.SecureServing, &serverConfig.Config.LoopbackClientConfig); err != nil {
+			return nil, err
+		}
+	} else {
+		if err := opts.SecureServing.ApplyTo(&serverConfig.Config.SecureServing, nil); err != nil {
+			return nil, err
+		}
+		serverConfig.LoopbackClientConfig = rest.CopyConfig(optionalLocalShardRestConfig)
 	}
 	if err := opts.Authentication.ApplyTo(&serverConfig.Config.Authentication, serverConfig.SecureServing, serverConfig.OpenAPIConfig); err != nil {
 		return nil, err
--- a/pkg/server/cache.go
+++ b/pkg/server/cache.go
@@ -1,0 +1,65 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package server
+
+import (
+	"context"
+
+	genericapiserver "k8s.io/apiserver/pkg/server"
+	"k8s.io/client-go/rest"
+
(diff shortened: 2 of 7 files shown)


pkg/server/cache.go
```diff
@@ -0,0 +1,141 @@
...
+	"k8s.io/client-go/rest"
+
+	virtualcommandoptions "github.com/kcp-dev/kcp/cmd/virtual-workspaces/options"
+	cacheserver "github.com/kcp-dev/kcp/pkg/cache/server"
+	cacheoptions "github.com/kcp-dev/kcp/pkg/cache/server/options"
+)
+
+// installCacheServer integrates the cache-server into the kcp-server.
+// The new server is available at /services/cache/... HTTP path.
+func (s *Server) installCacheServer(ctx context.Context) error {
```
p0lyn0mial, line 39: @sttts @stevekuznetsov this is how I am going to wire the cache-server into the kcp-server

pkg/server/cache.go
```diff
@@ -0,0 +1,141 @@
...
+
+// withCacheServiceRoundTripper wraps an existing config's with cacheServiceRoundTripper.
+func withCacheServiceRoundTripper(cfg *rest.Config) *rest.Config {
+	cfg.Wrap(func(rt http.RoundTripper) http.RoundTripper {
+		return newCacheServiceRoundTripper(rt)
+	})
+	return cfg
+}
+
+// cacheServiceRoundTripper is a http.RoundTripper that appends "/services/cache" prefix to a request.
```
p0lyn0mial, line 111: I think that the cache-server should add the `/services/cache` prefix. It would always give us the same path, whether we run as a standalone server or part of kcp. WDYT?

pkg/server/server.go
```diff
@@ -467,6 +467,10 @@ func (s *Server) Run(ctx context.Context) error {
 		}
 	}
 
+	if err := s.installCacheServer(ctx); err != nil {
```
p0lyn0mial, line 470: needs to be behind a flag

ncdc:
/retitle 🌱 Wire the cache server into the kcp server

pkg/cache/server/config.go
```diff
@@ -87,7 +88,7 @@ func (c *Config) Complete() (CompletedConfig, error) {
 	}}, nil
 }
 
-func NewConfig(opts *cacheserveroptions.CompletedOptions) (*Config, error) {
+func NewConfig(opts *cacheserveroptions.CompletedOptions, restConfig *rest.Config) (*Config, error) {
```
sttts, line 91: can we qualify that more? Which restConfig is that? To the shard?

pkg/server/cache.go
```diff
@@ -0,0 +1,73 @@
...
+
+	virtualcommandoptions "github.com/kcp-dev/kcp/cmd/virtual-workspaces/options"
+	cacheserver "github.com/kcp-dev/kcp/pkg/cache/server"
+	cacheoptions "github.com/kcp-dev/kcp/pkg/cache/server/options"
+)
+
+// installCacheServer integrates the cache-server into the kcp-server.
+// The new server is available at /services/cache/... HTTP path.
+func (s *Server) installCacheServer(ctx context.Context) error {
+	cacheServerOptions := cacheoptions.NewOptions(s.Options.Extra.RootDirectory)
```
sttts, line 33: this looks like options/config/server skew.

pkg/cache/server/config.go
p0lyn0mial: yes, `optionalShardConfig` when we run in-process, will update the name.

pkg/server/cache.go
p0lyn0mial: you were right, I think it looks better now, PTAL.

pkg/cache/server/config.go
```diff
@@ -87,7 +88,9 @@ func (c *Config) Complete() (CompletedConfig, error) {
 	}}, nil
 }
 
-func NewConfig(opts *cacheserveroptions.CompletedOptions) (*Config, error) {
+// NewConfig returns a new Config for the given options and optional rest.Config that point to the local server.
+// Pass it only when you combine this server with a different one.
+func NewConfig(opts *cacheserveroptions.CompletedOptions, optionalLocalShardRestConfig *rest.Config) (*Config, error) {
```
sttts, line 93: :+1:

pkg/server/options/options.go
```diff
@@ -329,7 +329,18 @@ func (o *Options) Complete() (*CompletedOptions, error) {
 		o.Extra.BatteriesIncluded = bats.List()
 	}
 
-	o.Cache.Complete(completedGenericControlPlane.SecureServing)
+	cacheServerEtcdOptions := *completedGenericControlPlane.Etcd
+	o.Cache.Server.Etcd = &cacheServerEtcdOptions
```
sttts, line 333: nit: technically a completion func should do something like:
```
if o.Cache.Server.Etcd == nil { 
  o.Cache.Server.Etcd = &cacheServerEtcdOptions
}
```
Think about future customizers.

pkg/server/options/cache.go
```diff
@@ -23,42 +23,64 @@ import (
 	"github.com/spf13/pflag"
 
 	genericoptions "k8s.io/apiserver/pkg/server/options"
+
+	cacheoptions "github.com/kcp-dev/kcp/pkg/cache/server/options"
 )
 
+type CacheCompleted struct {
```
sttts, line 30: this must be a private type

pkg/server/options/cache.go
```diff
@@ -23,42 +23,64 @@ import (
...
+	}
+	return errs
+}
+
 type Cache struct {
+	// Server includes options provided by the cache server
+	Server *cacheoptions.Options
+	extra
+}
+type extra struct {
```
sttts, line 51: Extra is usually public. Otherwise, you cannot access it from outside.

pkg/server/options/options.go
p0lyn0mial: mhm, `o.Cache.Server.Etcd ` cannot be nil, it gets populated in `Server: cacheoptions.NewOptions(rootDir),`

stevekuznetsov:
v

stevekuznetsov:

p0lyn0mial:
@sttts please remove the last commit and add the gist to the https://github.com/kcp-dev/kcp/issues/2021 issue. I'd like to merge this PR and refactor it as you suggested in a follow-up PR.

sttts:
Do we have a follow-up for the option-config-server skew refactoring? Do we have TODOs in the code?

p0lyn0mial:
> Do we have a follow-up for the option-config-server skew refactoring? Do we have TODOs in the code?

I have created https://github.com/kcp-dev/kcp/issues/2021 to address that concern.

Issue #209, opened 2021-11-16
Design & Implement Cross-Cluster Listers and Informers


stevekuznetsov:
Right now, a cross-cluster informer can be created by simply passing in a cross-cluster client (using `*`) for the source client in constructor. As the `MetaNamespaceKeyFunc()` is already cluster-aware, all of the internal machinery seems to work correctly.

On the consumer end, though, we need to:
- expose a simple splitter func for the cluster/namespace/name
- expose new interfaces for listing that are first clustered, then namespaced, then named, eg.:

```go
cluster, namespace, name, err := cache.SplitMetaClusterNamespaceKey(key) // or whatever
obj, err := c.lister.Cluster(cluster).ObjectType(namespace).Get(name)
```

/cc @ncdc

ncdc:
client-go's hand-written base code for the cache store and shared index informers all are written based on a hard-coded key function that only knows how to encode an object's namespace (optionally) and name. The generated informers and listers follow from this, therefore only supporting namespace and name.

Do we want to...

1. change the key funcs to assume <cluster, namespace, name> always (i.e. force k/k to support logical clusters)?
2. hack this on top of client-go (i.e. in kcp) with our own generators that can generate informers & listers that support logical clusters?
3. some other option?

@smarterclayton @sttts any thoughts?

stevekuznetsov:
I think we must do something on top - a user of the listers that is not cluster aware should not need to know that the key they are using will contain cluster information, especially since naive clients will be within one lcluster anyway.

ncdc:
Also worth remembering that in our kube fork, we have made a change to `cache.MetaNamespaceKeyFunc` to include the cluster name as part of the name segment. At the very least, this has to be resolved & finalized for going back into upstream kube, even if we don't have direct support for informers/listers that are logical cluster aware.

sttts:
Question is also how to opt into the "*" behaviour. If cross-shard client access is only possible through a virtual workspace, the default behaviour with empty cluster name could be cross-shard. So the user actually opts into queries for specific shards.

stevekuznetsov:
I guess as a straw-man, I expect:

- a new pair of funcs to encode and decode a cluster-aware key
- the ability to configure the func used in the reflector/store
- explicit opt-in to a cross-cluster lister and informer by using the cross-cluster client

Without explicit choice, no client call will ever cross cluster boundaries.

pweil-:
0.8 check in here.  Still on track to land in 0.8?

[...]

PR #1958, opened 2022-09-12
:sparkles:  *: use cluster-aware keyfuncs & splitters 

stevekuznetsov:
go.mod: use newest kcp-dev/apimachinery

---

*: use cluster-aware keyfuncs & splitters

---

Part of #209 
Generated with https://github.com/kcp-dev/code-generator/pull/53

/assign @ncdc @sttts

### DIFF
--- a/pkg/admission/kubequota/kubequota_clusterworkspace_monitor.go
+++ b/pkg/admission/kubequota/kubequota_clusterworkspace_monitor.go
@@ -20,12 +20,12 @@ import (
 	"fmt"
 	"time"
 
+	kcpcache "github.com/kcp-dev/apimachinery/pkg/cache"
 	"github.com/kcp-dev/logicalcluster/v2"
 
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/tools/clusters"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 
@@ -60,7 +60,7 @@ func newClusterWorkspaceDeletionMonitor(
 }
 
 func (m *clusterWorkspaceDeletionMonitor) enqueue(obj interface{}) {
-	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+	key, err := kcpcache.DeletionHandlingMetaClusterNamespaceKeyFunc(obj)
 	if err != nil {
 		runtime.HandleError(err)
 		return
@@ -113,13 +113,11 @@ func (m *clusterWorkspaceDeletionMonitor) processNextWorkItem() bool {
 }
 
 func (m *clusterWorkspaceDeletionMonitor) process(key string) error {
-	_, clusterAwareName, err := cache.SplitMetaNamespaceKey(key)
+	parent, _, name, err := kcpcache.SplitMetaClusterNamespaceKey(key)
 	if err != nil {
 		runtime.HandleError(err)
 		return nil
 	}
-	// e.g. root:org<separator>ws
-	parent, name := clusters.SplitClusterAwareKey(clusterAwareName)
 
 	// turn it into root:org:ws
 	clusterName := parent.Join(name)
--- a/pkg/reconciler/apis/apibindingdeletion/apibinding_deletion_controller.go
+++ b/pkg/reconciler/apis/apibindingdeletion/apibinding_deletion_controller.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	jsonpatch "github.com/evanphx/json-patch"
+	kcpcache "github.com/kcp-dev/apimachinery/pkg/cache"
 	"github.com/kcp-dev/logicalcluster/v2"
 
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -111,7 +112,7 @@ type Controller struct {
 }
 
 func (c *Controller) enqueue(obj interface{}) {
-	key, err := cache.MetaNamespaceKeyFunc(obj)
+	key, err := kcpcache.MetaClusterNamespaceKeyFunc(obj)
 	if err != nil {
 		runtime.HandleError(err)
 		return
--- a/pkg/reconciler/apis/apiexport/apiexport_controller.go
+++ b/pkg/reconciler/apis/apiexport/apiexport_controller.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"time"
 
+	kcpcache "github.com/kcp-dev/apimachinery/pkg/cache"
 	"github.com/kcp-dev/logicalcluster/v2"
 
 	corev1 "k8s.io/api/core/v1"
@@ -173,7 +174,7 @@ type controller struct {
 
 // enqueueAPIBinding enqueues an APIExport .
 func (c *controller) enqueueAPIExport(obj interface{}) {
-	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+	key, err := kcpcache.DeletionHandlingMetaClusterNamespaceKeyFunc(obj)
 	if err != nil {
 		runtime.HandleError(err)
 		return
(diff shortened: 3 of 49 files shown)


pkg/admission/kubequota/kubequota_clusterworkspace_monitor.go
```diff
@@ -113,13 +113,11 @@ func (m *clusterWorkspaceDeletionMonitor) processNextWorkItem() bool {
 }
 
 func (m *clusterWorkspaceDeletionMonitor) process(key string) error {
-	_, clusterAwareName, err := cache.SplitMetaNamespaceKey(key)
+	parent, _, name, err := kcpcache.SplitMetaClusterNamespaceKey(key)
 	if err != nil {
 		runtime.HandleError(err)
 		return nil
 	}
-	// e.g. root:org<separator>ws
```
stevekuznetsov, line 121: Hm, crap - Need to do a lot more work to get comments munged correctly ... would likely need to merge this with any from the first statement, since this `SplitClusterAwareKey` node gets removed entirely

pkg/reconciler/apis/identitycache/api_export_identity_controller.go
```diff
@@ -90,17 +90,16 @@ func NewApiExportIdentityProviderController(
 
 	configMapInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
 		FilterFunc: func(obj interface{}) bool {
-			key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
```
stevekuznetsov, line 93: @ncdc @sttts do we really want `obj interface -> clusterName logicalcluster.Name` transformation to be implemented like this? Why not a `logicalcluster.FromAny(obj interface) (logicalcluster.Name, error)`?

ncdc:
I think there are legit failures from the changes here?

stevekuznetsov:
That's ... concerning. I'll check tomorrow - syncer unit tests likely are from the test fakes?

stevekuznetsov:
OK, rebased on all re-reqs. Passing locally. @ncdc ready for review

pkg/reconciler/apis/apibindingdeletion/apibinding_deletion_controller.go
```diff
@@ -111,7 +112,7 @@ type Controller struct {
 }
 
 func (c *Controller) enqueue(obj interface{}) {
-	key, err := cache.MetaNamespaceKeyFunc(obj)
+	key, err := kcpcache.MetaClusterNamespaceKeyFunc(obj)
```
sttts, line 115: when does this have to be the deletion handling variant?

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -205,20 +206,25 @@ func (c *controller) enqueueAllAPIExports(clusterWorkspaceShard interface{}) {
...
+	apiExports, err := c.apiExportIndexer.ByIndex(indexers.APIExportBySecret, secretKey)
 	if err != nil {
 		runtime.HandleError(err)
 		return
 	}
 
 	logger := logging.WithObject(logging.WithReconciler(klog.Background(), controllerName), obj.(*corev1.Secret))
-	for _, key := range apiExportKeys {
+	for _, apiExport := range apiExports {
+		key, err := kcpcache.DeletionHandlingMetaClusterNamespaceKeyFunc(apiExport)
```
sttts, line 223: deletion handling is strange here.

pkg/reconciler/apis/apibindingdeletion/apibinding_deletion_controller.go
stevekuznetsov: In order to be more certain about the automated refactor code, I did not make any opinionated choice there - a simple change from non-cluster-aware to cluster-aware is the refactor. If we think it's a good idea, though, at least in our code I can move stuff to be deletion handling aware in a follow-up.

pkg/reconciler/apis/apiexport/apiexport_controller.go
stevekuznetsov: Similar to our conversation - if it's always correct to use `DeletionHandling`, let's just do it. It degenerates gracefully to the right thing and there's less mental overhead to figuring out when it is and when it is not necessary.

pkg/reconciler/workload/resource/resource_controller.go
```diff
@@ -279,7 +285,23 @@ func (c *Controller) processResource(ctx context.Context, key string) error {
...
+		return nil
+	}
+	// TODO(skuznets): can we figure out how to not leak this detail up to this code?
+	// I guess once the indexer is using kcpcache.MetaClusterNamespaceKeyFunc, we can just use that formatter ...
+	var indexKey string
+	if namespace != "" {
+		indexKey += namespace + "/"
+	}
+	if !lclusterName.Empty() {
+		indexKey += lclusterName.String() + "|"
```
ncdc, line 301: This is the right format? namespace/cluster|name?

pkg/reconciler/apis/apiexport/apiexport_controller.go
ncdc: There's no harm in doing it. But it's only strictly necessary when processing objects you get directly from AddFunc/UpdateFunc/DeleteFunc. Here, the objects are coming directly from the indexer, so you know they weren't deleted.

pkg/syncer/spec/spec_process.go
```diff
@@ -153,7 +151,7 @@ func (c *Controller) process(ctx context.Context, gvr schema.GroupVersionResourc
 	}
 	if !exists {
 		// deleted upstream => delete downstream
-		klog.Infof("Deleting downstream GVR %q object %s/%s for upstream cluster %q", gvr.String(), upstreamNamespace, name, clusterName)
+		klog.Infof("Deleting downstream GVR %q object %s/%s for upstream cluster %q", gvr.String(), downstreamNamespace, name, clusterName)
```
ncdc, line 164: not part of autogen? (but looks like a good catch)

pkg/reconciler/apis/apibindingdeletion/apibinding_deletion_controller.go
ncdc: DeletionHandling should be used on objects received directly as params to AddFunc/UpdateFunc/DeleteFunc. You don't _have_ to, but it does mean that if the reflector missed the delete event and it relists and the object is gone, you'll get a DeletedFinalStateUnknown, in which case you'll need to use the deletion handling variant to determine the key. And if you want the object in its last known state, you'll need to properly handle the tombstone.

pkg/reconciler/workload/resource/resource_controller.go
stevekuznetsov: That's what our fork of k8s is doing right now, yes ...

test/e2e/reconciler/cluster/controller_test.go
```diff
@@ -140,11 +140,13 @@ func TestClusterController(t *testing.T) {
 					return true
 				}, wait.ForeverTestTimeout, time.Millisecond*100, "expected source to show status change")
 
+				t.Logf("Deleting object in the source")
```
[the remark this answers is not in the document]
stevekuznetsov: No, did I accidentally put that in the wrong commit?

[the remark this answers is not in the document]
ncdc: Yes

pkg/syncer/spec/spec_process.go
stevekuznetsov: Yep, sorry. Wrong commit.

pkg/reconciler/apis/apiexport/apiexport_controller.go
stevekuznetsov: It's a fairly nuanced point and unless we think there's real value from not just using `DeletionHandling` everywhere IMHO it's going to be easier for us to be provably correct & doing the right thing if we just use it everywhere (I *think* you agree, since you suggested we drop the `DeletionHandling` option from our keyfuncs and make it default everywhere)

ncdc: Unless/until we see any performance issues, we know we can't mess up if we use the deletion handling variant 100% of the time, so yes I am in favor of that.

ncdc (approved):

PR #1961, opened 2022-09-13
:seedling: cache: exposes the cache-server under "/services/cache" path

p0lyn0mial:
## Summary
We already have `services/apiexports` for accessing resources exposed by the VW server. 
This PR adds `/services/cache` for accessing resources hosted on the cache-server.
Thanks to that the same path will be exposed no matter if the cache-server is running in a standalone mode or in-process with the kcp-server.

## Related issue(s)

Fixes #


sttts:
/lgtm
7approve

stevekuznetsov:

PR #1970, opened 2022-09-14
:seedling: kcp: adds flags related to the cache server

p0lyn0mial:
The following flags will be exposed to end users

 --cache-url string
       A URL address of a cache server associated with this instance (default localhost)
       
 --run-cache-server
       If set to true it runs the cache server with this instance (default false)

## Related issue(s)

Fixes #


pkg/server/options/cache.go
```diff
@@ -0,0 +1,51 @@
...
+	Enabled bool
+
+	// URL the url address of the cache server
+	URL string
+}
+
+func NewCache() *Cache {
+	return &Cache{
+		URL:     "",
+		Enabled: false,
```
p0lyn0mial, line 36: until we have a controller that uses it.

pkg/server/options/cache.go
```diff
@@ -0,0 +1,62 @@
...
+	fs.StringVar(&c.URL, "cache-url", c.URL, "A URL address of a cache server associated with this instance (default https://localhost:6443)")
+	fs.BoolVar(&c.Enabled, "run-cache-server", c.Enabled, "If set to true it runs the cache server with this instance (default false)")
+}
+
+func (c *Cache) Complete(secureServing *genericoptions.SecureServingOptionsWithLoopback) {
+	bindPort := 6443
+	if secureServing != nil && secureServing.BindPort != bindPort {
+		bindPort = secureServing.BindPort
+	}
+	c.URL = fmt.Sprintf("https://localhost:%v", bindPort)
```
ncdc, line 61: This overrides anything the user sets via the CLI flag?

p0lyn0mial: oh, you are right, I need to fix it.

ncdc:
/approve

will leave to @stevekuznetsov to lgtm

stevekuznetsov (approved):

stevekuznetsov:

PR #1978, opened 2022-09-14
:bug: syncer: bifurcate the namespace controller

stevekuznetsov:
Today, one controller/workqueue is used to drive two separate reconciliation flows: reacting to events in the upstream and downstream informers. The presence of a cluster in the queue key is used to infer which informers the event came from.

This is hard to reason about, and the logic for the two cases is entirely separate, and, furthermore, when we're syncing to a kcp logical cluster, the "downstream" objects have a cluster set and the keys do not follow the expected format.

Instead of inferring things from stringified keys in one queue, we can just split the two controllers and explicitly use cluster-aware or cluster-unaware keyfunctions where necessary.

Needed for #1958 
/assign @jmprusi 
/cc @ncdc

### DIFF
--- a/pkg/syncer/namespace/namespace_upstream_controller.go
+++ b/pkg/syncer/namespace/namespace_upstream_controller.go
@@ -45,15 +45,14 @@ import (
 
 const (
 	byNamespaceLocatorIndexName = "syncer-namespace-ByNamespaceLocator"
-	controllerName              = "kcp-workload-syncer-namespace"
+	upstreamControllerName      = controllerNameRoot + "-upstream"
 )
 
-type Controller struct {
+type UpstreamController struct {
 	queue workqueue.RateLimitingInterface
 
 	deleteDownstreamNamespace                  func(ctx context.Context, namespace string) error
 	upstreamNamespaceExists                    func(clusterName logicalcluster.Name, upstreamNamespaceName string) (bool, error)
-	getDownstreamNamespace                     func(name string) (runtime.Object, error)
 	getDownstreamNamespaceFromNamespaceLocator func(namespaceLocator shared.NamespaceLocator) (runtime.Object, error)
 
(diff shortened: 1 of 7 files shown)


pkg/syncer/namespace/namespace_controller.go
```diff
@@ -40,21 +38,19 @@ import (
...
+	"github.com/kcp-dev/kcp/third_party/keyfunctions"
 )
 
 const (
-	byNamespaceLocatorIndexName = "syncer-namespace-ByNamespaceLocator"
-	controllerName              = "kcp-workload-syncer-namespace"
+	controllerName = "kcp-workload-syncer-namespace"
 )
 
 type Controller struct {
```
ncdc, line 48:
```suggestion
type DownstreamController struct {
```

pkg/syncer/namespace/namespace_upstream_controller.go
```diff
@@ -0,0 +1,204 @@
...
+	upstreamNamespaceExists                    func(clusterName logicalcluster.Name, upstreamNamespaceName string) (bool, error)
+	getDownstreamNamespaceFromNamespaceLocator func(namespaceLocator shared.NamespaceLocator) (runtime.Object, error)
+
+	syncTargetName      string
+	syncTargetWorkspace logicalcluster.Name
+	syncTargetUID       types.UID
+	syncTargetKey       string
+}
+
+func NewNamespaceUpstreamController(
```
ncdc, line 64:
```suggestion
func NewUpstreamController(
```

pkg/syncer/namespace/namespace_controller.go
ncdc: Also rename constructor to NewDownstreamController

ncdc (approved):

Issue #1838, opened 2022-08-26
feat: add a workspace tree subcommand


lburgazzoli:
### Feature Description

It would be nice to have a workspace tree subcommand to visualize the hierarchy of workspaces, i.e.

```
NAME           TYPE           PHASE   URL
------------- ------------- -------- -----
root           universal      Ready   ...
 ├─ org1       organization   Ready   ...
 │   ├─ team1  organization   Ready   ...
 │   └─ team2  organization   Ready   ...
 └─ foo        user           Ready   ...
```

There are some similar tools:
1. https://github.com/ahmetb/kubectl-tree
2. https://github.com/nimakaviani/knative-inspect/

### Proposed Solution

Ass a kubectl-ws tree subcommand 

### Want to contribute?

- [X] I would like to work on this issue.

ncdc:
FYI, as of today, this will require multiple requests (1 per level)

stevekuznetsov:
@ncdc or access to the tenancy APIExport?

ncdc:
That doesn't give you a list of all workspaces hierarchically though

lburgazzoli:
> FYI, as of today, this will require multiple requests (1 per level)

Yes I had a look at the code and it is probably be a little bit convoluted to implement so may need some help, but I'll :)

stevekuznetsov:
@ncdc not sure I understand what you mean - it gives you a cross-cluster list, which you could then reorder in hierarchical order?

ncdc:
hmm, maybe? although when we have 2+ shards, that might require multiple calls?

lburgazzoli:
> I imagine that would be hidden behind the virtual workspace, no?

can you point me to some code ?

lburgazzoli:
> I'm not sure exactly what you mean - but if you request a cross-cluster list of some resource from an APIExport virtual workspace, you will get all of them, without shard distinctions (today).

Just that I'm not yet deeply expert in KCP so some exemplar code could be very helpful :)

PR #1989, opened 2022-09-14
🌱 add CLI `workspace show-tree'

mjudeikis:
## Summary

Workspaces being nested objects its quite hard to keep mental image of the system in your head. 
Simple command to show workspace tree:
```
[mjudeikis@unknown kcp]$ kubectl ws show-tree
.
└── workspaces
    └── root:users:zu:yc
        └── root:users:zu:yc:kcp-admin
            └── root:users:zu:yc:kcp-admin:services

[mjudeikis@unknown kcp]$ kubectl ws use root
Current workspace is "root".
[mjudeikis@unknown kcp]$ kubectl ws show-tree
.
└── workspaces
    ├── root:bob-magic
    ├── root:foo
    │   └── root:foo:bar
    └── root:users
        └── root:users:zu
            └── root:users:zu:yc
                └── root:users:zu:yc:kcp-admin
                    └── root:users:zu:yc:kcp-admin:services
```

## Related issue(s)

Fixes #


ncdc:
@mjudeikis thanks! Please update your description to indicate this fixes #1838 😄

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -673,3 +675,109 @@ func newKCPClusterClient(clientConfig clientcmd.ClientConfig) (kcpclient.Cluster
...
+	_, currentClusterName, err := pluginhelpers.ParseClusterURL(config.Host)
+	if err != nil {
+		return fmt.Errorf("current URL %q does not point to cluster workspace", config.Host)
+	}
+
+	tree := treeprint.New()
+	b := tree.AddBranch("workspaces")
+	var result *v1beta1.WorkspaceList
+
+	if err := wait.PollImmediate(time.Millisecond*100, time.Second*5, func() (bool, error) {
```
stevekuznetsov, line 726: Why is this poll necessary?

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -673,3 +675,109 @@ func newKCPClusterClient(clientConfig clientcmd.ClientConfig) (kcpclient.Cluster
...
+		}
+		return true, nil
+	}); err != nil {
+		return err
+	}
+
+	for _, workspace := range result.Items {
+		_, currentClusterName, err := pluginhelpers.ParseClusterURL(workspace.Status.URL)
+		if err != nil {
+			return fmt.Errorf("current URL %q does not point to cluster workspace", config.Host)
```
stevekuznetsov, line 741: This is not the current URL any longer - but a sub-workspace URL

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -673,3 +675,109 @@ func newKCPClusterClient(clientConfig clientcmd.ClientConfig) (kcpclient.Cluster
...
+		return err
+	}
+	_, currentClusterName, err := pluginhelpers.ParseClusterURL(config.Host)
+	if err != nil {
+		return fmt.Errorf("current URL %q does not point to cluster workspace", config.Host)
+	}
+
+	tree := treeprint.New()
+	b := tree.AddBranch("workspaces")
+	var result *v1beta1.WorkspaceList
```
stevekuznetsov, line 724: nit: you should be able to factor this into one caller, and not need to duplicate the list+recurse logic

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -673,3 +675,109 @@ func newKCPClusterClient(clientConfig clientcmd.ClientConfig) (kcpclient.Cluster
...
+	}
+
+	b := tree.AddBranch(name.String())
+
+	for _, workspace := range result.Items {
+		_, currentClusterName, err := pluginhelpers.ParseClusterURL(workspace.Status.URL)
+		if err != nil {
+			return fmt.Errorf("current URL %q does not point to cluster workspace", workspace.Status.URL)
+		}
+		if !strings.Contains(workspace.Name, ":") {
```
stevekuznetsov, line 775: To my knowledge it's not possible to have `:` in a name, anyway. We shouldn't need to do this check.

stevekuznetsov (commented):
The `tree` command only shows the segment of the path that's new, so it would rock if the command here also only showed the segment that was different from the parent.

mjudeikis:
> @mjudeikis thanks! Please update your description to indicate this fixes #1838 smile

Ou, nice, there was an issue for this :D Will tide up this in next few days. It was quick hack but looks like moving to right direction :)

mjudeikis:
I know this might be sub-optimal. But I still lack some of the knowledge of internals to make it optimal :)
I didn't wanted to drag in any exteral dependecies so looked into vendor tree to see what libs are available, and saddly this one does not have fancy print options. But we can replace if needed later on.

pkg/cliplugins/workspace/plugin/kubeconfig.go
mjudeikis: yap, was me trying to understand how to construct this better. Should be ok now

mjudeikis:
If we want we can add type to this too, :shrug: 
```
[mjudeikis@unknown kcp]$ go run ./cmd/kubectl-workspaces/ tree -f
.
└── [root]  root
    ├── [organization]  root:cert-manager-service
    ├── [universal]  root:compute
    ├── [organization]  root:demo
    │   └── [universal]  root:demo:webapp
    └── [homeroot]  root:users

[mjudeikis@unknown kcp]$ go run ./cmd/kubectl-workspaces/ tree
.
└── root
    ├── cert-manager-service
    ├── compute
    ├── demo
    │   └── webapp
    └── users
```

mjudeikis:
ping @stevekuznetsov @ncdc

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -673,3 +674,90 @@ func newKCPClusterClient(clientConfig clientcmd.ClientConfig) (kcpclient.Cluster
...
+
+// Run outputs the current workspace.
+func (o *TreeOptions) Run(ctx context.Context) error {
+	config, err := o.ClientConfig.ClientConfig()
+	if err != nil {
+		return err
+	}
+	_, currentClusterName, err := pluginhelpers.ParseClusterURL(config.Host)
+	if err != nil {
+		return fmt.Errorf("sub-workspace URL %q does not point to workspace", config.Host)
```
stevekuznetsov, line 722:
```suggestion
		return fmt.Errorf("current config context URL %q does not point to workspace", config.Host)
```

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -673,3 +674,90 @@ func newKCPClusterClient(clientConfig clientcmd.ClientConfig) (kcpclient.Cluster
...
+		}
+		return err
+	}
+
+	var b treeprint.Tree
+	if o.Full {
+		b = tree.AddBranch(name.String())
+	} else {
+		parts := strings.Split(name.String(), ":")
+		b = tree.AddBranch(parts[len(parts)-1])
```
stevekuznetsov, line 749:
```go
name.Base()
```

stevekuznetsov (commented):
Looking good! Just some minor comments

stevekuznetsov (approved):
Sweet! Thanks @mjudeikis

stevekuznetsov:
I like the idea of adding a type to the tree - if you end up thinking it's a feature you would use a lot, perhaps we can do that in a follow-up?

mjudeikis:
> I like the idea of adding a type to the tree - if you end up thinking it's a feature you would use a lot, perhaps we can do that in a follow-up?

Yap, lets see how it goes and if becomes nesessity - add later.

PR #1990, opened 2022-09-14
:seedling: ci: don't use a local registry with kind for syncer e2e tests

markmc:
For test-e2e-shared and test-e2e-sharded, we need a kind cluster and for our images (syncer, test-image) to be available from that cluster.

In github, we use the container-tools/kind-action to set up this kind cluster, which also sets up a local registry by default and configures the kind cluster to use it.

It turns out, if you want to reproduce this in your dev environment, you don't need to have this local registry setup because ko will happily side-load the image into the kind node. This seems pretty convenient.

Since the local registry isn't necessary, and probably not what you want when developing, let's not use it in CI for consistency.

See:
https://github.com/container-tools/kind-action/blob/main/registry.sh https://github.com/ko-build/ko/blob/main/pkg/publish/kind/write.go

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,9 +37,10 @@ jobs:
           go-version: v1.18
       - run: make build
 
-      # Install kind with a local registry
       - uses: container-tools/kind-action@v1
-        name: Kubernetes KinD Cluster w/local registry
+        name: Kubernetes KinD Cluster
+        with:
+          registry: false
 
       - run: |-
           go install github.com/google/ko@latest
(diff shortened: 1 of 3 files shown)


markmc:
> Can you add the `kind load docker-image` bits

We don't seem to use that - we use ko's kind publisher via `KO_DOCKER_REPO=kind.local`

> to the README as well

I don't see where we're building images in README, but I updated CONTRIBUTING and docs/syncer.md

ncdc:
@markmc sorry for the delay in attention on your PR! Would you be willing to rebase & we can get it merged?

markmc:
Will reopen if I ever rebase

Issue #1888, opened 2022-09-05
feature: syncer should sync all resources exposed in syncer virtual workspace.


qiujian16:
### Feature Description

This is part of the epic https://github.com/kcp-dev/kcp/issues/1895 to support location(compute) workspace.

### Proposed Solution

Currently syncer syncs resource based on `resource-to-sync` flag. The syncer should sync resources that are all discovered in virtual workspace.

Syncer also should create apiResourceImport on syncTarget workspace based on `syncTarget.Status.SyncedResources`.

The `resource-to-sync` flag should be deprecated in the future.

### Want to contribute?

- [X] I would like to work on this issue.

PR #1995, opened 2022-09-15
✨ feature: sync resources from syncer virtual server

qiujian16:

## Summary
In addition to sync resource based on resources flag on syncer, also read the syncedResource on synctarget. It requires to start spec/status syncer per gvr dynamically. With this change, user can specify the apiexport on synctarget and do not need to set resource flag on syncer any more.

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/1888


pkg/syncer/apiimporter.go
```diff
@@ -167,8 +165,42 @@ func (i *APIImporter) Stop() {
...
+			},
+		},
+	})
+	if err != nil {
+		klog.Errorf("error creating APIResourceImport: %v", err)
+		return
+	}
+	clusterObj, exists, err := i.clusterIndexer.GetByKey(clusterKey)
+	if err != nil {
+		klog.Errorf("error creating APIResourceImport: %v", err)
```
stevekuznetsov, line 182: Please use structured, contextual logging,

pkg/syncer/resourcesync/controller.go
```diff
@@ -0,0 +1,348 @@
...
+	if syncTarget.GetUID() != c.syncTargetUID {
+		return nil
+	}
+
+	requiredGVRs, err := getAllGVRs(syncTarget)
+	if err != nil {
+		return err
+	}
+
+	errs := []error{}
```
stevekuznetsov, line 222: nit: we generally prefer to not initialize the slice: 

```go
var err []error
```

pkg/syncer/resourcesync/controller.go
```diff
@@ -0,0 +1,348 @@
...
+}
+
+// stopSyncer stop syncers for gvrs not in requiredGVRs
+func (c *Controller) stopSyncers(ctx context.Context, requiredGVRs map[schema.GroupVersionResource]bool) {
+	logger := klog.FromContext(ctx)
+
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	syncersToDelete := []schema.GroupVersionResource{}
```
stevekuznetsov, line 242: nit: we generally prefer to not initialize the slice

pkg/syncer/resourcesync/controller.go
```diff
@@ -0,0 +1,348 @@
...
+	for gvr := range c.syncerMap {
+		if _, ok := requiredGVRs[gvr]; !ok {
+			syncersToDelete = append(syncersToDelete, gvr)
+		}
+	}
+
+	for _, gvr := range syncersToDelete {
```
stevekuznetsov, line 249: What's the point of having this slice if we're just going to iterate on it?

pkg/syncer/resourcesync/controller.go
```diff
@@ -0,0 +1,348 @@
...
+
+	if _, ok := c.syncerMap[gvr]; ok {
+		return nil
+	}
+
+	if len(syncTarget.Status.VirtualWorkspaces) == 0 {
+		return nil
+	}
+
+	syncerVirtualWorkspaceURL := syncTarget.Status.VirtualWorkspaces[0].URL
```
stevekuznetsov, line 272: Is it too early to handle the N>1 case?

pkg/syncer/resourcesync/controller.go
```diff
@@ -0,0 +1,348 @@
...
+
+	upstreamURL, err := url.Parse(syncerVirtualWorkspaceURL)
+	if err != nil {
+		return err
```
stevekuznetsov, line 276: Why will retrying this help?

pkg/syncer/resourcesync/controller.go
```diff
@@ -0,0 +1,348 @@
...
+	}
+
+	// Check whether we're in the Advanced Scheduling feature-gated mode.
+	advancedSchedulingEnabled := false
+	if syncTarget.GetAnnotations()[AdvancedSchedulingFeatureAnnotation] == "true" {
+		logger.WithValues("SyncTarget", c.syncTargetName).Info("Advanced Scheduling feature is enabled")
```
stevekuznetsov, line 282: nit: lower-case the field `syncTarget.Name` (does it not come from the reconcilitaiton keys?) and add this to the logger early.

pkg/syncer/resourcesync/controller.go
```diff
@@ -0,0 +1,348 @@
...
+		advancedSchedulingEnabled = true
+	}
+
+	syncTargetKey := workloadv1alpha1.ToSyncTargetKey(c.syncTargetWorkspace, c.syncTargetName)
+
+	upstreamInformers := dynamicinformer.NewFilteredDynamicSharedInformerFactoryWithOptions(c.upstreamDynamicClusterClient.Cluster(logicalcluster.Wildcard), metav1.NamespaceAll, func(o *metav1.ListOptions) {
```
stevekuznetsov, line 288: Is it at all possible to have one set of informers and to defer filtering? Do we care about the load here?

pkg/syncer/resourcesync/controller.go
```diff
@@ -0,0 +1,348 @@
...
+	go statusSyncer.Start(syncerCtx, c.numSyncerThreads)
+	go downstreamInformers.Start(syncerCtx.Done())
+	go upstreamInformers.Start(syncerCtx.Done())
+
+	c.syncerMap[gvr] = cancel
+	return nil
+}
+
+func getAllGVRs(synctarget *workloadv1alpha1.SyncTarget) (map[schema.GroupVersionResource]bool, error) {
+	// add configmap and secrets by default
```
stevekuznetsov, line 320: Why? We have a list of built-in types that's more than just this, and regardless don't we have e.g. the `kubernetes` export with these bits?

stevekuznetsov (commented):
Are we certain that a reconciler process per synctarget like this is the most efficient way to do it?

pkg/syncer/resourcesync/controller.go
qiujian16: This is from the old code :) https://github.com/kcp-dev/kcp/blob/main/pkg/syncer/syncer.go#L292. I think the reason is we do not sync cluster scoped and we do not sync sa. The kubernetes export does not include configmaps and secrets.

pkg/syncer/resourcesync/controller.go
qiujian16: The filter is to handle the certain synctarget only.

qiujian16:
> Are we certain that a reconciler process per synctarget like this is the most efficient way to do it?

would you explain more on the question? I think the syncer process is to handle a single synctarget.

stevekuznetsov:
@qiujian16 do we think at a fundamental design level this could be done by dynamically managing the lifecycle of informer's handlers in the syncer controllers as-is? E.G. when the spec or status syncer sees that the set of GVRs to manage changes, start & stop handling events on the correct informers. That would allow us to have fewer controllers happening here and likely make the approach a little simpler.

ncdc:
@qiujian16 could you please explain what you're doing here? 😄 I'm curious why you are starting 1 set of controllers per GVR?

I think you could accomplish the same thing by just filtering out "invalid" GVRs when processing resources, instead of starting lots and lots of controllers?

qiujian16:
> @qiujian16 could you please explain what you're doing here? 😄 I'm curious why you are starting 1 set of controllers per GVR?
> 
> I think you could accomplish the same thing by just filtering out "invalid" GVRs when processing resources, instead of starting lots and lots of controllers?

we do not know what is a "invalid" GVR, the objective is the syncer could know from synctarget API (or maybe discovery API from the syncer virtual workspace) what are the GVRs to be synced from the upstream. The GVRs list can be pretty dynamic. so we need to dynamically add/remove upstream/downstream informer per gvr for spec/status sync.

qiujian16:
> @qiujian16 do we think at a fundamental design level this could be done by dynamically managing the lifecycle of informer's handlers in the syncer controllers as-is? E.G. when the spec or status syncer sees that the set of GVRs to manage changes, start & stop handling events on the correct informers. That would allow us to have fewer controllers happening here and likely make the approach a little simpler.

I think we could. We still need to start/stop informer per gvr based on synctarget API. But we could just set eventHandler for the spec/status sync, instead of start spec/status sync per gvr. Does it sound sensible?

qiujian16:
@stevekuznetsov @ncdc PTAL again, change to avoid creating spec/status syncer per gvr

pkg/syncer/apiimporter.go
```diff
@@ -134,44 +134,71 @@ type APIImporter struct {
 func (i *APIImporter) Start(ctx context.Context, pollInterval time.Duration) {
 	defer runtime.HandleCrash()
 
-	i.kcpInformerFactory.Start(ctx.Done())
-	i.kcpInformerFactory.WaitForCacheSync(ctx.Done())
+	logger := logging.WithReconciler(klog.FromContext(ctx), "kcp-syncer-apiimporter")
+	ctx = klog.NewContext(ctx, logger)
 
-	klog.Infof("Starting API Importer for location %s in cluster %s", i.location, i.logicalClusterName)
+	logger.Info("Starting API Importer", "location", i.location, "cluster", i.logicalClusterName)
```
stevekuznetsov, line 140: `logging.WithObject(location)` will help :)

pkg/syncer/apiimporter.go
```diff
@@ -134,44 +134,71 @@ type APIImporter struct {
...
 	<-ctx.Done()
-	i.Stop()
+	i.Stop(ctx)
 }
 
-func (i *APIImporter) Stop() {
-	klog.Infof("Stopping API Importer for location %s in cluster %s", i.location, i.logicalClusterName)
+func (i *APIImporter) Stop(ctx context.Context) {
+	logger := klog.FromContext(ctx)
+	logger.Info("Stopping API Importer", "location", i.location, "cluster", i.logicalClusterName)
```
stevekuznetsov, line 153: If you add the object to the logger in the context, you will not need to add these fields more than once

pkg/syncer/apiimporter.go
```diff
@@ -134,44 +134,71 @@ type APIImporter struct {
...
+			Annotations: map[string]string{
+				logicalcluster.AnnotationKey: i.logicalClusterName.String(),
+			},
+		},
+	})
+	if err != nil {
+		logger.Error(err, "error building clusterKey")
+		return
+	}
+	syncTarget, err := i.syncTargetLister.Get(syncTargetKey)
```
stevekuznetsov, line 185: This will be the wrong key format - see other syncer code for the hard-coded reformat before we get new listers in.

pkg/syncer/apiimporter.go
```diff
@@ -189,22 +216,22 @@ func (i *APIImporter) ImportAPIs(ctx context.Context) {
 			clusterctl.GetGVRForLocationInLogicalClusterIndexKey(i.location, i.logicalClusterName, gvr),
 		)
 		if err != nil {
-			klog.Errorf("error pulling CRDs: %v", err)
+			logger.Error(err, "error pulling CRDs")
 			continue
 		}
 		if len(objs) > 1 {
-			klog.Errorf("There should be only one APIResourceImport of GVR %s for location %s in logical cluster %s, but there was %d", gvr.String(), i.location, i.logicalClusterName, len(objs))
+			logger.Error(fmt.Errorf("failed to get APIResourceImport"), "There should be only one APIResourceImport", "gvr", gvr.String(), "location", i.location, "cluster", i.logicalClusterName, "obejcts", len(objs))
```
stevekuznetsov, line 223: Here also - when you add the location to the logger at the top  level, you never need to add it again

pkg/syncer/resourcesync/controller.go
```diff
@@ -0,0 +1,330 @@
...
+			}
+			return name == syncTargetName && clusterName == syncTargetWorkspace
+		},
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj interface{}) { c.enqueueSyncTarget(obj, logger) },
+			UpdateFunc: func(old, obj interface{}) {
+				oldCluster := old.(*workloadv1alpha1.SyncTarget)
+				newCluster := obj.(*workloadv1alpha1.SyncTarget)
+
+				// only enqueue when syncedResource is changed.
```
stevekuznetsov, line 132: fundamentally is it ok to do this? why? why not always enqueue? what if we miss an event?

pkg/syncer/resourcesync/controller.go
```diff
@@ -0,0 +1,330 @@
...
+		c.queue.AddRateLimited(key)
+		return true
+	}
+
+	c.queue.Forget(key)
+	return true
+}
+
+func (c *Controller) process(ctx context.Context, key string) error {
+	syncTarget, err := c.syncTargetLister.Get(key)
```
stevekuznetsov, line 219: This will be the wrong key format - see other syncer code for the hard-coded reformat before we get new listers in.

pkg/syncer/resourcesync/controller.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+func (c *Controller) startSyncerInformer(ctx context.Context, gvr schema.GroupVersionResource, syncTarget *workloadv1alpha1.SyncTarget) {
+	logger := klog.FromContext(ctx).WithValues("syncTarget", c.syncTargetName).WithValues("gvr", gvr.String())
+
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	if _, ok := c.syncerInformerMap[gvr]; ok {
+		return
```
stevekuznetsov, line 267: how do you handle the case where the set of resources for a synctarget changes.?

pkg/syncer/resourcesync/controller.go
qiujian16: hrm, I am bit confused here...I am using the same pattern as here https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/workload/synctargetexports/synctargetexports_controller.go#L265

pkg/syncer/resourcesync/controller.go
qiujian16: in sync target, we only track per gvr. If a gvr is included in the syncTarget, we start the informer when it is not started.

pkg/syncer/apiimporter.go
qiujian16: the original code is here https://github.com/kcp-dev/kcp/blob/main/pkg/syncer/apiimporter.go#L218-L230.

stevekuznetsov:
```go
	// TODO(skuznets): can we figure out how to not leak this detail up to this code?
	// I guess once the indexer is using kcpcache.MetaClusterNamespaceKeyFunc, we can just use that formatter ...
	var indexKey string
	if upstreamNamespace != "" {
		indexKey += upstreamNamespace + "/"
	}
	if !clusterName.Empty() {
		indexKey += clusterName.String() + "|"
	}
	indexKey += name
	// get the upstream object
	obj, exists, err := c.upstreamInformers.ForResource(gvr).Informer().GetIndexer().GetByKey(indexKey)
```

From `spec_process.go` is the correct incantation.

pkg/syncer/resourcesync/controller.go
stevekuznetsov:
```go
	// TODO(skuznets): can we figure out how to not leak this detail up to this code?
	// I guess once the indexer is using kcpcache.MetaClusterNamespaceKeyFunc, we can just use that formatter ...
	var indexKey string
	if upstreamNamespace != "" {
		indexKey += upstreamNamespace + "/"
	}
	if !clusterName.Empty() {
		indexKey += clusterName.String() + "|"
	}
	indexKey += name
	// get the upstream object
	obj, exists, err := c.upstreamInformers.ForResource(gvr).Informer().GetIndexer().GetByKey(indexKey)
```

From `spec_process.go` is the correct incantation.

stevekuznetsov: Please feel free to change synctargetexports as well

davidfestal:
Some minor comments, but apart from that LGTM

pkg/syncer/resourcesync/controller.go
```diff
@@ -0,0 +1,337 @@
...
+			Resource: "configmaps",
+		}: true,
+		{
+			Version:  "v1",
+			Resource: "secrets",
+		}: true,
+	}
+
+	// TODO(qiujian16) We currently checks the API compaibility on the server side. When we change to check the
+	// compatibility on the syncer side, this part needs to be changed.
```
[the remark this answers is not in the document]
qiujian16: I remembered that we had a discussion that we might expose APIResourceSchema in the virtual workspace so syncer can do the check locallly.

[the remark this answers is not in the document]
davidfestal: I see.

pkg/syncer/resourcesync/controller.go
qiujian16: fixed

davidfestal:
As we discussed during the meeting, do you think we could:
- issue a SSAR before adding a GVR to the map of syncer informers ([here ?](https://github.com/kcp-dev/kcp/pull/1995/files#diff-45a6b4e4e335883be4169968e14be1bbcebfb0127e2f61d634640d580785f0e8R277))
- Trigger a new reconcile (every 5 or 10 minutes) for resources that have not the expected permissions

davidfestal (changes requested):
As we discussed during the meeting, do you think we could:
- issue a SSAR before adding a GVR to the map of syncer informers ([here ?](https://github.com/kcp-dev/kcp/pull/1995/files#diff-45a6b4e4e335883be4169968e14be1bbcebfb0127e2f61d634640d580785f0e8R277))
- Trigger a new reconcile (every 5 or 10 minutes) for resources that have not the expected permissions
- Add the related tests

qiujian16:
> As we discussed during the meeting, do you think we could:
> 
> * Trigger a new reconcile (every 5 or 10 minutes) for resources that have not the expected permissions

I think we already have heartbeat that will trigger the reconcile here every 20s.

qiujian16:
ssar check and related test added.

davidfestal (approved):

PR #1996, opened 2022-09-15
🌱 authorisation: stop using the ValidClusterFrom method

p0lyn0mial:
## Summary
The `ValidClustrFrom` method doesn't allow for a cluster name to be empty.
Since the authorizers handle cases in which a cluster name wasn't provided using the `ClusterFrom` method leads to a more readable code.
The `ValidClusterFrom` returns an error when the name is empty which leads to slightly less readable code.

This is a follow-up from https://github.com/kcp-dev/kcp/pull/1871
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/authorization/toplevel_org_authorizer.go
+++ b/pkg/authorization/toplevel_org_authorizer.go
@@ -86,14 +86,14 @@ func (a *topLevelOrgAccessAuthorizer) Authorize(ctx context.Context, attr author
 		return a.delegate.Authorize(ctx, attr)
 	}
 
-	cluster, err := genericapirequest.ValidClusterFrom(ctx)
-	if err != nil || cluster == nil || cluster.Name.Empty() {
+	cluster := genericapirequest.ClusterFrom(ctx)
+	if cluster == nil || cluster.Name.Empty() {
 		kaudit.AddAuditAnnotations(
 			ctx,
 			TopLevelContentAuditDecision, DecisionNoOpinion,
-			TopLevelContentAuditReason, fmt.Sprintf("error getting cluster from request: %v", err),
+			TopLevelContentAuditReason, "empty cluster name",
 		)
-		return authorizer.DecisionNoOpinion, WorkspaceAccessNotPermittedReason, err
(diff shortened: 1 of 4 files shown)


p0lyn0mial:
We have a few more users of the `ValidClusterFrom` method left (some admission plugins). I will take a look and see if those could be updated as well.

pkg/authorization/toplevel_org_authorizer.go
```diff
@@ -86,8 +86,8 @@ func (a *topLevelOrgAccessAuthorizer) Authorize(ctx context.Context, attr author
 		return a.delegate.Authorize(ctx, attr)
 	}
 
-	cluster, err := genericapirequest.ValidClusterFrom(ctx)
-	if err != nil || cluster == nil || cluster.Name.Empty() {
+	cluster := genericapirequest.ClusterFrom(ctx)
+	if cluster == nil || cluster.Name.Empty() {
 		kaudit.AddAuditAnnotations(
```
s-urbaniak, line 91: this seems not symmetric with the other changes in the authorizer. I suggest to add audit log annotations there as well.

s-urbaniak:
one nit, else lgtm 👍

pkg/authorization/toplevel_org_authorizer.go
p0lyn0mial: okay, I think it only applies to the `LocalAuthorizer`, the other authorizers seem to be adding the annotation when the cluster wasn't found.

stevekuznetsov:
Nice!

pkg/authorization/toplevel_org_authorizer.go
```diff
@@ -86,8 +86,8 @@ func (a *topLevelOrgAccessAuthorizer) Authorize(ctx context.Context, attr author
...
 			ctx,
 			TopLevelContentAuditDecision, DecisionNoOpinion,
```
s-urbaniak, line 93: (next line) since the error is gone, this should now be:
```
TopLevelContentAuditReason, "empty cluster name",
```

p0lyn0mial: updated, thx

Issue #1836, opened 2022-08-25
feat: add --profiler-address to virtual workspace server and front proxy


kylape:
### Feature Description

This would be valuable to do memory profiling on more than just kcp.

### Proposed Solution

Make the --profiler-address flag work on the VW and front proxy commands just like it does in kcp.

### Want to contribute?

- [ ] I would like to work on this issue.

ncdc:
See https://github.com/kcp-dev/kcp/blob/d3c2260fabecaececb5c339c46504a52ea90f57c/pkg/server/config.go#L153-L156

bipuladh:
I would like to take this if no one is working on this.

ncdc:
Please go ahead, thanks!

PR #2023, opened 2022-09-19
:sparkles: Add profiler support to front proxy and virtual workspaces

bipuladh:
## Summary
Adds profiler address as command line argument

Fixes https://github.com/kcp-dev/kcp/issues/1836

Screenshot: 
Screenshot from 2022-09-19 20-44-37

### DIFF
--- a/cmd/virtual-workspaces/command/cmd.go
+++ b/cmd/virtual-workspaces/command/cmd.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"net/http"
 	"net/url"
 	"time"
 
@@ -76,6 +77,7 @@ func NewCommand(ctx context.Context, errout io.Writer) *cobra.Command {
 
 // Run takes the options, starts the API server and waits until stopCh is closed or initial listening fails.
 func Run(ctx context.Context, o *options.Options) error {
+	logger := klog.FromContext(ctx).WithValues("component", "virtual-workspaces")
 	// parse kubeconfig
 	kubeConfig, err := readKubeConfig(o.KubeconfigFile, o.Context)
 	if err != nil {
@@ -96,7 +98,7 @@ func Run(ctx context.Context, o *options.Options) error {
 	identityConfig, resolveIdentities := bootstrap.NewConfigWithWildcardIdentities(nonIdentityConfig, bootstrap.KcpRootGroupExportNames, bootstrap.KcpRootGroupResourceExportNames, nil)
 	if err := wait.PollImmediateInfiniteWithContext(ctx, time.Millisecond*500, func(ctx context.Context) (bool, error) {
 		if err := resolveIdentities(ctx); err != nil {
-			klog.V(3).Infof("failed to resolve identities, keeping trying: %v", err)
+			logger.V(3).Info("failed to resolve identities, keeping trying: ", "err", err)
 			return false, nil
 		}
 		return true, nil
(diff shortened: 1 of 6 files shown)


pkg/virtual/options/options.go
```diff
@@ -72,6 +76,11 @@ func (o *Options) NewVirtualWorkspaces(
 	wildcardKubeInformers kubernetesinformers.SharedInformerFactory,
 	wildcardKcpInformers kcpinformers.SharedInformerFactory,
 ) ([]rootapiserver.NamedVirtualWorkspace, error) {
+
```
stevekuznetsov, line 79: I think the main `kcp` process also calls this - if so, we'd want to make sure that we start the profiler only when we're in the standalone case.

bipuladh: @stevekuznetsov I don't think we would be able to set the profiler address when it's getting called from `kcp` process but I could move the invocation to `cmd.go`.

stevekuznetsov: When we're running the virtual workspace server as part of `kcp`, we should be using the `kcp` options to run the (single) profiler server. When we're running standalone, we should use the built-in k8s APIServer profiling option for the VW to run it.

bipuladh: I am not sure how to use the profiling option of APIServer and specify a particular address as it's config doesn't accept any endpoints as per  https://pkg.go.dev/k8s.io/apiserver/pkg/server?utm_source=gopls#Config.EnableProfiling
@sttts any pointers?

sttts: yes, this belongs into cmd/virtual-workspaces/options.

s-urbaniak:
just a linter check error for the missing err check after the ListenAndServe call, else lgtm from my side as well 👍

cmd/virtual-workspaces/command/cmd.go
```diff
@@ -96,7 +98,7 @@ func Run(ctx context.Context, o *options.Options) error {
 	identityConfig, resolveIdentities := bootstrap.NewConfigWithWildcardIdentities(nonIdentityConfig, bootstrap.KcpRootGroupExportNames, bootstrap.KcpRootGroupResourceExportNames, nil)
 	if err := wait.PollImmediateInfiniteWithContext(ctx, time.Millisecond*500, func(ctx context.Context) (bool, error) {
 		if err := resolveIdentities(ctx); err != nil {
-			klog.V(3).Infof("failed to resolve identities, keeping trying: %v", err)
+			logger.V(3).Info(fmt.Sprintf("failed to resolve identities, keeping trying: %v", err))
```
sttts, line 101: this is not yet what we need here (although the linter passed): look into other uses of logger throughout the code-base. Basically if you need `fmt.Sprintf` it's a sign of non-structured logs. Structured logs have pairs of keys and values behind the string. Here `err` is such a value. @stevekuznetsov is the expert here. Maybe he has some good pointers.

cmd/virtual-workspaces/command/cmd.go
```diff
@@ -161,7 +168,7 @@ func Run(ctx context.Context, o *options.Options) error {
 		return err
 	}
 
-	klog.Infof("Starting virtual workspace apiserver on %s (%s)", rootAPIServerConfig.GenericConfig.ExternalAddress, version.Get().String())
+	logger.Info(fmt.Sprintf("Starting virtual workspace apiserver on %s (%s)", rootAPIServerConfig.GenericConfig.ExternalAddress, version.Get().String()))
```
sttts, line 171: similarly to above

PR #2029, opened 2022-09-20
:bug: pkg/admission/webhook: fix nil panic

s-urbaniak:
## Summary
pkg/admission/webhook: fix nil panic

## Related issue(s)

Fixes #2028

Issue #2028, opened 2022-09-20
bug: 


s-urbaniak:
### Describe the bug

We observe a panic in kcp-stable:

```
I0920 06:18:45.080532       1 httplog.go:131] "HTTP" verb="LIST" URI="/clusters/root:users:xr:vd:rh-sso-hongweiliu:integration-ws/api/v1/namespaces" latency="30.807977ms" userAgent="kcp/v1.24.3+kcp (linux/amd64) kubernetes/7e4e8d8/kcp-workspace-deletion-controller" audit-ID="4114112e-8f37-4b7f-9473-7a27fd3fec2a" srcIP="[::1]:35884" apf_pl="e
xempt" apf_fs="exempt" apf_execution_time="30.636802ms" resp=200
I0920 06:18:45.080828       1 httplog.go:131] "HTTP" verb="LIST" URI="/apis/managed-gitops.redhat.com/v1alpha1/gitopsdeployments" latency="251.607µs" userAgent="kcp/v1.24.3+kcp (linux/amd64) kubernetes/7e4e8d8/kcp-apibinding-controller" audit-ID="f5495de4-e987-47f5-b663-74d171f62557" srcIP="[::1]:35884" apf_pl="exempt" apf_fs="exempt" apf_execution_time="80.766µs" resp=404
I0920 06:18:45.090080       1 httplog.go:131] "HTTP" verb="LIST" URI="/clusters/root:users:zu:bh:rh-sso-mnairn-hcg/api/v1/namespaces/kcp-system/resourcequotas" latency="10.586141ms" userAgent="kcp/v1.24.3+kcp (linux/amd64) kubernetes/7e4e8d8" audit-ID="f7c60c68-b480-4748-8a99-3516ab4ba91d" srcIP="[::1]:35884" apf_pl="exempt" apf_fs="exempt" apf_execution_time="10.37765ms" resp=200
E0920 06:18:45.090508       1 runtime.go:79] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 98168 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x3e2ec60?, 0x6428850})
        /go/pkg/mod/github.com/kcp-dev/kubernetes/staging/src/k8s.io/apimachinery@v0.0.0-20220906141821-de523af4fb02/pkg/util/runtime/runtime.go:75 +0x99
[...]

PR #2031, opened 2022-09-20
:bug: pkg/admission/webhook: differentiate states

s-urbaniak:
## Summary
Only if a bound API export is set but no bound API export workspace we want to log. This fixes it.

## Related issue(s)

Related with #2028
Follow-up for #2029

### DIFF
--- a/pkg/admission/webhook/generic_webhook.go
+++ b/pkg/admission/webhook/generic_webhook.go
@@ -96,12 +96,13 @@ func (p *WebhookDispatcher) getAPIBindingWorkspace(attr admission.Attributes, cl
 	for _, obj := range objs {
 		apiBinding := obj.(*apisv1alpha1.APIBinding)
 		for _, br := range apiBinding.Status.BoundResources {
-			if apiBinding.Status.BoundAPIExport == nil || apiBinding.Status.BoundAPIExport.Workspace == nil {
+			if apiBinding.Status.BoundAPIExport != nil && apiBinding.Status.BoundAPIExport.Workspace == nil {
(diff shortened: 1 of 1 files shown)


pkg/admission/webhook/generic_webhook.go
```diff
@@ -96,12 +96,13 @@ func (p *WebhookDispatcher) getAPIBindingWorkspace(attr admission.Attributes, cl
...
 		for _, br := range apiBinding.Status.BoundResources {
-			if apiBinding.Status.BoundAPIExport == nil || apiBinding.Status.BoundAPIExport.Workspace == nil {
+			if apiBinding.Status.BoundAPIExport != nil && apiBinding.Status.BoundAPIExport.Workspace == nil {
 				// this will never happen today. But as soon as we add other reference types (like exports), this log output will remind out of necessary work here.
 				klog.Errorf("APIBinding %s has no referenced workspace", clusterName, apiBinding.Name)
 				continue
 			}
-			if br.Group == attr.GetResource().Group && br.Resource == attr.GetResource().Resource {
+			if apiBinding.Status.BoundAPIExport != nil && apiBinding.Status.BoundAPIExport.Workspace != nil &&
+				br.Group == attr.GetResource().Group && br.Resource == attr.GetResource().Resource {
```
sttts, line 105: 💣 this potentially means that bound resources will miss admission. That's not good. We have to think and clean up our binding state logic everywhere.

sttts:
@s-urbaniak https://github.com/kcp-dev/kcp/pull/2031#discussion_r975180349 is critical, so we need a follow-up of this PR. To stop the bleeding let's move forward here though.

PR #2002, opened 2022-09-16
:seedling: Front proxy  use contextual logging

hardys:
## Summary
Updates the front-proxy cmd and related sharded-test-server code to align with contextual logging

## Related issue(s)
Part of #1694

PR #2042, opened 2022-09-21
:seedling: pkg/proxy use contextual logging

hardys:
## Summary

Follow up from #2002 with some more incremental changes moving the proxy code to contextual logging

## Related issue(s)
Part of #1694

### DIFF
--- a/hack/logcheck.out
+++ b/hack/logcheck.out
@@ -116,8 +116,6 @@
 /pkg/logging/constants.go:52:9: Key positional arguments are expected to be inlined constant strings. Please replace ReconcilerKey provided with string value.
 /pkg/logging/constants.go:57:9: Key positional arguments are expected to be inlined constant strings. Please replace QueueKeyKey provided with string value.
 /pkg/logging/constants.go:67:9: Additional arguments to WithValues should always be Key Value pairs. Please check if there is any key or value missing.
-/pkg/proxy/index/index_controller.go:195:2: function "Infof" should not be used, convert to contextual logging
-/pkg/proxy/index/index_controller.go:196:8: function "Infof" should not be used, convert to contextual logging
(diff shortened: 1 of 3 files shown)


hack/logcheck.out
```diff
@@ -116,14 +116,6 @@
...
-/pkg/proxy/handler.go:54:4: function "V" should not be used, convert to contextual logging
-/pkg/proxy/handler.go:61:4: function "Infof" should not be used, convert to contextual logging
-/pkg/proxy/handler.go:61:4: function "V" should not be used, convert to contextual logging
-/pkg/proxy/handler.go:71:3: function "Infof" should not be used, convert to contextual logging
-/pkg/proxy/handler.go:71:3: function "V" should not be used, convert to contextual logging
-/pkg/proxy/index/index_controller.go:195:2: function "Infof" should not be used, convert to contextual logging
-/pkg/proxy/index/index_controller.go:196:8: function "Infof" should not be used, convert to contextual logging
 /pkg/proxy/index/index_controller.go:212:2: function "Infof" should not be used, convert to contextual logging
 /pkg/proxy/mapping.go:72:3: function "Infof" should not be used, convert to contextual logging
 /pkg/proxy/mapping.go:72:3: function "V" should not be used, convert to contextual logging
```
hardys, line 121: @stevekuznetsov these remaining cases are in top-level functions that don't have access to a context object - how would you prefer to handle that, leave them or use e.g Background/TODO?

stevekuznetsov: If they are in something below `main()` we can thread through a context ( likely we have one implicitly given interrupt signals, right? ) but if it's a massive hassle in some options validation code, we can just ignore it.

stevekuznetsov (approved):

hack/logcheck.out
hardys: Yeah you're right, they both just need to be wired in from the main - I'll do that and push a follow-up PR, thanks!

PR #2044, opened 2022-09-21
🌱  proxy: pass context for contextual logging

hardys:
## Summary

Follow-up to #2042 - to convert these we have to pass the context in from main

## Related issue(s)
Part of #1694


hardys:
/cc @stevekuznetsov 

Based on the discussion [here](https://github.com/kcp-dev/kcp/pull/2042#discussion_r976352781)

stevekuznetsov (commented):
Where do we create the root logger with the `component` field?

hardys:
> Where do we create the root logger with the `component` field?

It's set [here in the main.go](https://github.com/kcp-dev/kcp/blob/main/cmd/kcp-front-proxy/main.go#L77) (added as part of #2002)

ncdc (approved):

PR #2047, opened 2022-09-22
:seedling: pkg/server: enable audit logging in virtual workspace apiserver

s-urbaniak:
## Summary
This enables audit logging in the virtual workspace apiserver (`/services/.*` handler).

## Related issue(s)

None

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -474,7 +474,7 @@ func (s *Server) Run(ctx context.Context) error {
 	}
 
 	if s.Options.Virtual.Enabled {
-		if err := s.installVirtualWorkspaces(ctx, controllerConfig, delegationChainHead, s.GenericConfig.Authentication, s.GenericConfig.ExternalAddress, s.preHandlerChainMux); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/server/server.go
```diff
@@ -474,7 +474,7 @@ func (s *Server) Run(ctx context.Context) error {
 	}
 
 	if s.Options.Virtual.Enabled {
-		if err := s.installVirtualWorkspaces(ctx, controllerConfig, delegationChainHead, s.GenericConfig.Authentication, s.GenericConfig.ExternalAddress, s.preHandlerChainMux); err != nil {
+		if err := s.installVirtualWorkspaces(ctx, controllerConfig, delegationChainHead, s.GenericConfig.Authentication, s.GenericConfig.ExternalAddress, s.GenericConfig.AuditPolicyRuleEvaluator, s.preHandlerChainMux); err != nil {
```
s-urbaniak, line 477: don't we want to simply pass `s.GenericConfig` here? would make the signature smaller.

sttts: leaking less is better IMO. So I would leave it like this.

PR #2053, opened 2022-09-22
🌱 kcp: controllers don't have to check if caches have been synced during startup

p0lyn0mial:
## Summary
This requirement is enforced by the `waitForSync` method.
All controllers must be gated on that method before they are run.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/reconciler/kubequota/kubequota_controller.go
+++ b/pkg/reconciler/kubequota/kubequota_controller.go
@@ -18,7 +18,6 @@ package kubequota
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -75,8 +74,6 @@ type Controller struct {
 	scopingResourceQuotaInformer        *ScopingResourceQuotaInformer
 	scopingGenericSharedInformerFactory *scopingGenericSharedInformerFactory
 
-	crdsSynced cache.InformerSynced
-
 	// For better testability
 	getClusterWorkspace func(key string) (*tenancyv1alpha1.ClusterWorkspace, error)
 	listCRDs            func() ([]*apiextensionsv1.CustomResourceDefinition, error)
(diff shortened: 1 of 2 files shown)


stevekuznetsov (commented):
What was the reason we didn't do this everywhere in the first place?

ncdc:
This is covered by https://github.com/kcp-dev/kcp/blob/3c4e7937b80aac7b2f20878393e957417702969a/pkg/server/controllers.go#L1107

ncdc:
s.syncedCh is closed when all informers are synced

p0lyn0mial:
> s.syncedCh is closed when all informers are synced

okay, but I think that:

- controllers should be self-contained
- this is now how a standard controller looks like
- this is non-intuitive, for example, [this controller](https://github.com/kcp-dev/kcp/blob/3c4e7937b80aac7b2f20878393e957417702969a/pkg/reconciler/kubequota/kubequota_controller.go#L95) accepts `informersStarted <-chan struct{}` (`s.syncedCh`) but also waits for `crdInformer.Informer().HasSynced` or [the bootstrap controller](https://github.com/kcp-dev/kcp/blob/cf0235c3cc51e8eae17981ebca0424200ab8c3c4/pkg/reconciler/tenancy/bootstrap/bootstrap_controller.go#L77) that waits for `workspaceInformer.Informer().HasSynced`

ncdc:
> controllers should be self-contained

This is the key point up for discussion I'd say. @sttts & I made a decision a while ago to do it the way you currently see in the repo. I am happy to discuss if we want to change the approach.

> this is non-intuitive, for example, [this controller](https://github.com/kcp-dev/kcp/blob/3c4e7937b80aac7b2f20878393e957417702969a/pkg/reconciler/kubequota/kubequota_controller.go#L95) accepts informersStarted <-chan struct{} (s.syncedCh)

This is a one-off requirement of the Kubernetes quota controller. Now that I think about it, since we don't start any quota delegates until after s.syncedCh is closed, we could remove passing it in, and instead pass an always-closed channel when creating delegate controllers.

> but also waits for crdInformer.Informer().HasSynced

I don't recall the specifics, but I think I ran into some issues... we could try removing this and see what happens. Maybe I missed putting in the waitForSync originally, added the specific crdInformer check, and never went back to remove it.

> or [the bootstrap controller](https://github.com/kcp-dev/kcp/blob/cf0235c3cc51e8eae17981ebca0424200ab8c3c4/pkg/reconciler/tenancy/bootstrap/bootstrap_controller.go#L77) that waits for workspaceInformer.Informer().HasSynced

It's probably safe to remove this.

p0lyn0mial:
Ok, updated the PR.

The only controller that makes use of the `HasSynced` method is `ResourceController`. I think it is okay since it uses `DynamicDiscoverySharedInformerFactory`.

Issue #2066, opened 2022-09-23
Replace +kubebuilder:Required with +kubebuilder:validation:Required


ncdc:
The former is not valid/has no effect

PR #2067, opened 2022-09-23
:bug: apis: use correct annotation to require fields

stevekuznetsov:

Fixes https://github.com/kcp-dev/kcp/issues/2066
@ncdc did you expect this to change the output of the code generators?

### DIFF
--- a/pkg/apis/apis/v1alpha1/types_apiexport.go
+++ b/pkg/apis/apis/v1alpha1/types_apiexport.go
@@ -223,7 +223,7 @@ type GroupResource struct {
 	// not provided by an api export.
 	// +kubebuilder:validation:Pattern=`^[a-z][-a-z0-9]*[a-z0-9]$`
 	// +required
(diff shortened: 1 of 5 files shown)


ncdc:
Not sure... here's the controller-tools code
```go
switch defaultMode {
		// if this package isn't set to optional default...
		case "required":
			// ...everything that's not inline, omitempty, or explicitly optional is required
			if !inline && !omitEmpty && field.Markers.Get("kubebuilder:validation:Optional") == nil && field.Markers.Get("optional") == nil {
				props.Required = append(props.Required, fieldName)
			}

		// if this package isn't set to required default...
		case "optional":
			// ...everything that isn't explicitly required is optional
			if field.Markers.Get("kubebuilder:validation:Required") != nil {
				props.Required = append(props.Required, fieldName)
			}
		}
```

ncdc (approved):

PR #2071, opened 2022-09-23
🌱 Add make targets to update and verify go.mod

ncdc:
## Summary
- Add `make modules` to run `go mod tidy` on the repo root and pkg/apis
- Add `make verify-modules` to ensure the everything is up to date

## Related issue(s)

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -274,6 +274,18 @@ verify-imports:
 verify-go-versions:
 	hack/verify-go-versions.sh
 
+.PHONY: modules
+modules: ## Run go mod tidy to ensure modules are up to date
+	go mod tidy
+	cd pkg/apis; go mod tidy
+
+.PHONY: verify-modules
+verify-modules: modules  ## Verify go modules are up to date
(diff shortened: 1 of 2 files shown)


ncdc:
@stevekuznetsov @sttts @davidfestal do you want this as a separate Prow job (similar to boilerplate, codegen, etc.)? Or incorporate it into one of the existing ones?

Bigger question: should we consolidate any of these into a single Prow job / what are the pros/cons of having 1 job vs. many?
- lint
- codegen
- deps
- imports
- boilerplate
- test

ncdc:
hooray for prow jobs getting killed
/retest

stevekuznetsov:
They're just containers in the end, so there's really no overhead to more. I don't know that any of these is costly & hard to run, IMO the risk of a mega-job is re-doing too much work when something flakes, right?

ncdc:
Ok, happy to update openshift-release to add a new job after this merges

PR #2078, opened 2022-09-27
:bug: Revert "syncer: remove finalizer if downstream resource has been deleted"

jmprusi:
## Summary

Revert the latest syncer changes causing issues with the e2e tests and syncer stability.


kasturinarra:
@jmprusi one question here, is there anything that is a user facing change here that needs to be verified to make sure this is working? thanks !!

jmprusi:
@kasturinarra I don't think so, as this reverted a commit that introduced some issues (syncer and scheduler entering into a change loop of resources.)

Issue #1994, opened 2022-09-14
bug: flake: Test_integration_multiple_connections


stevekuznetsov:
### Describe the bug

```
 === FAIL: pkg/tunneler Test_integration_multiple_connections (2.62s)
2022/09/14 16:54:57 http: proxy error: EOF
2022/09/14 16:54:57 http: proxy error: EOF
    integration_test.go:121: Expected Hello world received 
    integration_test.go:121: Expected Hello world received 
```
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kcp-dev_kcp/1985/pull-ci-kcp-dev-kcp-main-test/1570092330344517632

### Steps To Reproduce

```
make test
```

### Expected Behaviour

success

ncdc:
@stevekuznetsov @sttts it may be time to disable this test (`t.Skip`) until we have time to investigate it - it's for something that's behind a feature gate and not fully implemented yet.

PR #2079, opened 2022-09-27
:seedling: Combine Dockerfile kubectl layers to save space

hardys:
## Summary
Combining the chmod with the download avoids creating another layer, which saves about 45MB currently


ncdc (approved):

ncdc:
@hardys in the future, please put a link to the flake when you `/retest` - thanks!

#1994

Issue #1787, opened 2022-08-17
Syncer deletes the kcp-root-ca.crt downstream configmap when its upstream kube-root-ca.crt still exists


ncdc:
The syncer is deleting the kcp-root-ca.crt configmap from the downstream cluster when kube-root-ca.crt still exists upstream. Not sure yet why.

Original description:
~As seen on one of @robinbobbitt's OpenShift clusters, a syncer was up for 4+ days, but it didn't appear to be doing resyncs (which are supposed to be every 10 hours). The user-facing issue is that in kcp workspaces, we have the `kube-root-ca.crt` configmap, which is supposed to be synced down the the physical cluster as `kcp-root-ca.crt`. Some of these configmaps were 5+ days old in the workspace, yet they didn't exist downstream. If we created a new namespace, the configmap was created downstream. If we labeled an existing configmap in the kcp workspace, the syncer immediately picked it up.~

ncdc:
From @robinbobbitt - syncer logs showing the syncer deleting it!

```
I0817 21:03:15.881514       1 spec_process.go:156] Deleting downstream GVR "/v1, Resource=configmaps" object default/kube-root-ca.crt for upstream cluster "root:org:ryb-compute-operator-aug14"
I0817 21:13:38.884529       1 spec_process.go:156] Deleting downstream GVR "/v1, Resource=configmaps" object compute-config/kube-root-ca.crt for upstream cluster "root:org:ryb-compute-operator-aug14"
I0818 12:52:58.643062       1 spec_process.go:156] Deleting downstream GVR "/v1, Resource=configmaps" object compute-config/kcp-root-ca.crt for upstream cluster "root:org:ryb-compute-operator-aug14"
I0818 12:52:58.643535       1 spec_process.go:156] Deleting downstream GVR "/v1, Resource=configmaps" object default/kcp-root-ca.crt for upstream cluster "root:org:ryb-compute-operator-aug14"
I0818 12:53:12.588652       1 spec_process.go:156] Deleting downstream GVR "/v1, Resource=configmaps" object ryb-aug12/kcp-root-ca.crt for upstream cluster "root:org:ryb-compute-operator-aug14"
I0818 12:54:16.892030       1 spec_process.go:156] Deleting downstream GVR "/v1, Resource=secrets" object ryb-aug12/kcp-default-token-q272m for upstream cluster "root:org:ryb-compute-operator-aug14"
[...]

PR #2082, opened 2022-09-27
:bug: Syncer: remove upstream finalizer when the resource is deleted downstream.

jmprusi:
## Summary

When a resource is deleted downstream, the status syncer will remove the finalizer from the upstream resource to better reflect the current state. (This could be triggered by either a user or process deleting the downstream resource) 

* Includes some refactoring of the "EnsureUpstreamFinalizerRemoved" function, as it has been split into two different funcs.

* Additional checks in tests to avoid flakyness.

## Related issue(s)

Related to https://github.com/kcp-dev/kcp/issues/1787


jmprusi:
after discussion we don't want to go into that direction, this issue has a better fix in https://github.com/kcp-dev/kcp/pull/2096

PR #2084, opened 2022-09-27
:seedling: Remove deploy task from github actions

kylape:
This task triggers a deploy to an internal Red Hat environment, but now this is done with a separate component.  We do not want a direct connection to the internal environment like this.


csams (approved):
/lgtm

PR #2087, opened 2022-09-28
🌱 cache: annotate apiexports that will be replicated to the cache server

p0lyn0mial:
## Summary
adds `internal.sharding.kcp.dev/replicate` annotation to all apiexports from the root workspace.

note that this PR has no effect until the replication controller and the cache server is started.

## Related issue(s)


p0lyn0mial:
/close 

in favour of https://github.com/kcp-dev/kcp/pull/2213

PR #2024, opened 2022-09-19
🌱 cache: introduce the replication controller

p0lyn0mial:
## Summary
The new controller copies objects of defined resources that have the `internal.sharding.kcp.dev/replicate` annotation to the cache server.
The replicated object will be placed under the same cluster as the original object.
In addition to that, all replicated objects will be placed under the shard taken from the shardName argument.
For example: shards/{shardName}/clusters/{clusterName}/apis/apis.kcp.dev/v1alpha1/apiexports

As of today the controller replicates the ApiExport resources.
The reconcileAPIExports function handles the following cases:
1. creation of the object in the cache server
   happens when the cached object is not found in c.cacheApiExportsIndexer
2. deletion of the object from the cache server
   happens when either of the following is true:
   - the original/local object was removed
   - the original/local object was not found in c.localApiExport
3. modification of the object to match the original object
   happens when either of the following is true:
    - the original object's metadata doesn't match the cached one.
    - the original object's spec doesn't match the cached one.
    - the original object's status doesn't match the cached one.

TODO:
 - in a follow-up PR I am going to add `APIResourceSchema` to this controller

## Related issue(s)

https://github.com/kcp-dev/kcp/issues/342

PR #2090, opened 2022-09-28
🌱 cache: extend the replication controller to reconcile apiresourceschema

p0lyn0mial:
## Summary
makes the replication controller a bit more generic so that adding new resources is easier.
it also extends the controller to reconcile `ApiResourceSchema` resources.

## Related issue(s)

requires https://github.com/kcp-dev/kcp/pull/2024

### DIFF
--- a/pkg/reconciler/cache/replication/replication_controller.go
+++ b/pkg/reconciler/cache/replication/replication_controller.go
@@ -23,6 +23,8 @@ import (
 
 	kcpcache "github.com/kcp-dev/apimachinery/pkg/cache"
 
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
@@ -33,7 +35,7 @@ import (
 	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
 	cacheclient "github.com/kcp-dev/kcp/pkg/cache/client"
 	"github.com/kcp-dev/kcp/pkg/cache/client/shard"
-	apisinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/apis/v1alpha1"
+	kcpinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
 	apislisters "github.com/kcp-dev/kcp/pkg/client/listers/apis/v1alpha1"
 	"github.com/kcp-dev/kcp/pkg/logging"
 )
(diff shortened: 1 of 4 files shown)


pkg/reconciler/cache/replication/replication_reconcile.go
```diff
@@ -19,102 +19,126 @@ package replication
...
 			return err
 		}
-		unstructuredCacheApiExport.SetKind("APIExport")
-		unstructuredCacheApiExport.SetAPIVersion(gvr.GroupVersion().String())
+		unstructuredCacheObject.SetKind(gvk.Kind)
+		unstructuredCacheObject.SetAPIVersion(gvr.GroupVersion().String())
 	}
-	if localApiExport != nil {
-		unstructuredLocalApiExport, err = toUnstructured(localApiExport)
+	if localObject != nil && !reflect.ValueOf(localObject).IsNil() {
```
p0lyn0mial, line 113: an interface is a two-word data, it holds data and a type, since this code operates on the `interface` type and informers are strongly typed we need to make sure that the data part is not nil.

pkg/reconciler/cache/replication/replication_reconcile.go
```diff
@@ -19,102 +19,126 @@ package replication
...
 		}
 	}
 
-	var unstructuredCacheApiExport *unstructured.Unstructured
-	var unstructuredLocalApiExport *unstructured.Unstructured
-	if cacheApiExport != nil {
-		unstructuredCacheApiExport, err = toUnstructured(cacheApiExport)
+	var unstructuredCacheObject *unstructured.Unstructured
+	var unstructuredLocalObject *unstructured.Unstructured
+	if cacheObject != nil && !reflect.ValueOf(cacheObject).IsNil() {
```
p0lyn0mial, line 105: I think that https://github.com/kcp-dev/kcp/pull/2090/files#r985622615 doesn't apply here but I added in for symmetry

pkg/reconciler/cache/replication/replication_reconcile.go
```diff
@@ -19,102 +19,130 @@ package replication
...
+		return fmt.Errorf("incorrect key: %v, expected group.version.resource::key", gvrKey)
 	}
 	switch keyParts[0] {
 	case apisv1alpha1.SchemeGroupVersion.WithResource("apiexports").String():
-		return c.reconcileAPIExports(ctx, keyParts[1], apisv1alpha1.SchemeGroupVersion.WithResource("apiexports"))
+		return c.reconcileObject(ctx,
+			keyParts[1],
+			apisv1alpha1.SchemeGroupVersion.WithResource("apiexports"),
+			apisv1alpha1.SchemeGroupVersion.WithKind("APIExport"),
+			func(gvr schema.GroupVersionResource, cluster, namespace, name string) (interface{}, error) {
```
stevekuznetsov, line 49: Can we use generics to keep this strongly typed?

p0lyn0mial: mhm, I think it would require refactoring because you cannot define a type parameter on a method (in that case that would be reconcileObject). I'm not sure if this is worth it. What would that give us? This code doesn't need to know about the type because it translates an object to an unstructured object anyway.

pkg/reconciler/cache/replication/replication_controller.go
```diff
@@ -57,38 +59,52 @@ func NewController(
 	shardName string,
 	dynamicCacheClient dynamic.ClusterInterface,
 	dynamicLocalClient dynamic.ClusterInterface,
-	localApiExportInformer apisinformers.APIExportInformer,
-	cacheApiExportInformer apisinformers.APIExportInformer,
+	localKcpInformers kcpinformers.SharedInformerFactory,
```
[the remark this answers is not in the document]
p0lyn0mial: deliberately changed to use the factory because this controller would end up with many parameters.

pkg/reconciler/cache/replication/replication_reconcile.go
p0lyn0mial: I haven't used generics in go yet, I experimented with https://go.dev/play/p/nI3bsSm_MRZ

sttts:
/lgtm
/approve

Red unit tests.

p0lyn0mial:
@stevekuznetsov if you decide to continue with https://github.com/kcp-dev/kcp/pull/2170 this PR will force you to rebase - sorry for that :)

Issue #2088, opened 2022-09-28
bug: cannot build withe the latest code on Mac 


qiujian16:
### Describe the bug

seems relate to this https://github.com/kcp-dev/kcp/pull/2019

it returns error

```
GOOS=""  GOARCH=i38664 go build  -ldflags="-X k8s.io/client-go/pkg/version.gitCommit=c3b6f9ca -X k8s.io/client-go/pkg/version.gitTreeState=clean -X k8s.io/client-go/pkg/version.gitVersion=v1.24.3+kcp-v0.8.0-188-gc3b6f9ca11b85e -X k8s.io/client-go/pkg/version.gitMajor=1 -X k8s.io/client-go/pkg/version.gitMinor=24 -X k8s.io/client-go/pkg/version.buildDate=2022-09-28T08:32:54Z -X k8s.io/component-base/version.gitCommit=c3b6f9ca -X k8s.io/component-base/version.gitTreeState=clean -X k8s.io/component-base/version.gitVersion=v1.24.3+kcp-v0.8.0-188-gc3b6f9ca11b85e -X k8s.io/component-base/version.gitMajor=1 -X k8s.io/component-base/version.gitMinor=24 -X k8s.io/component-base/version.buildDate=2022-09-28T08:32:54Z" -o bin ./cmd/...
go: unsupported GOOS/GOARCH pair darwin/i38664
```

```
 % uname -p
i386
```

os: MacOS Moterey 12.5.1
cpu: Intel Core i7

go env:

```
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
```
### Steps To Reproduce

pull the lates code and run `make build` on Mac

### Expected Behaviour

`make build` should succeed.

PR #2091, opened 2022-09-28
:bug: Allow go to determine ARCH and OS during build

rh-iatanaso:
## Summary
Allow go to determine ARCH and OS during build

## Related issue(s)

Fixes #2088

PR #2093, opened 2022-09-28
:bug: Set ARCH using go env

qiujian16:

## Summary

`make build` fails on mac os due to run ARCH. Change to use ARCH in go env instead

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/2088


ncdc:
Thanks! Going to go with #2091 as that came in slightly earlier 😄 
/close

PR #2099, opened 2022-09-29
✨ pkg/admission/apibinding: verify workspace reference for existence

### LINKED ISSUE #1786: Confusing error message when trying to create an APIBinding that points to a workspace whose top-level workspace does not exist

s-urbaniak:
## Summary
This adds a check to apibinding admission to verify the referenced api binding workspace reference actually exists.
Conincidentally it enhances user facing messages.

@ncdc @stevekuznetsov i believe the addition call is worth it, it augments the workspace existence check and enhances user messaging, wdyt?

also cc @sttts 

## Related issue(s)

Fixes #1786

### DIFF
--- a/pkg/admission/apibinding/apibinding_admission.go
+++ b/pkg/admission/apibinding/apibinding_admission.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/kcp-dev/logicalcluster/v2"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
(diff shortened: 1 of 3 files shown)


s-urbaniak:
closing out, based on the discussion in https://kubernetes.slack.com/archives/C021U8WSAFK/p1664451035006599 this exposes too much information about workspace topology. One could find out if workspaces exist or not.

PR #2103, opened 2022-09-29
:seedling: Remove ingress splitter and deployment splitter which do not work currently...

davidfestal:
## Summary

Remove ingress splitter and deployment splitter since they currently do not work
and should be reimplemented based on
coordination controllers.


ncdc:
There's also an e2e for ingress

ncdc:
Will need to update contextual logging

davidfestal:
> There's also an e2e for ingress

Let's remove it also since anyway it is broken:
```
				//
				//
				// Here, the ingress test stop working because the ingresssplitter controller looks for
				// ingress objects in kcp without taking the identity into consideration. We have to lift
				// ingressplitter be negotiation workspace aware, and while doing that probably also move
				// to vw-based transformations.
				//
				//
				return
```

Issue #1671, opened 2022-08-03
Epic: API evolution


ncdc:
- [ ] Add fields to APIResourceSchema that contain CEL that is called during API version conversion
- [ ] Add CEL converter to apiextensions-apiserver
- [ ] Bound CRDs will need a way to point back at the actual CEL (annotation?)
- [ ] Conversions must go through a single hub version
- [ ] Feature gated

Demo:
- `kubectl apply` yaml with v1alpha1
- `kubectl get` returns v1alpha2, see results of conversion

pweil-:
0.8 Check in here.  I don't see anything scoped per milestone, can we break down this epic more to see what is expected when?  Thanks!

ncdc:
I started looking into this before my PTO. I will get it updated this week. Going to move to 0.9 for the time being.

/milestone v0.9

ncdc:
Started some API brainstorming https://hackmd.io/_EnPemBUTF-b7WFs8o6Gjw

stevekuznetsov:
This has been subsumed by https://github.com/kcp-dev/kcp/issues/2014 

/close

Issue #2014, opened 2022-09-16
feature: Evolve an APIResourceSchema From One Version To Another


stevekuznetsov:
### Feature Description

As a service provider, I want to move my provided API from `v1alpha1` to `v1beta1`, changing the schema to add and/or remove fields, and migrating users automatically.

### Proposed Solution

Lots of open questions.

[Design HackMD](https://hackmd.io/_EnPemBUTF-b7WFs8o6Gjw)
[Design Google Doc](https://docs.google.com/document/d/1k72fgGBNJzbY1Rkxj9_1WksHZR0cNT5OeadCpWuc-zM/edit#heading=h.5hl04ap2sjps)

### Want to contribute?

- [ ] I would like to work on this issue.

PR #2105, opened 2022-09-29
✨ Add conversions + CEL transformations for APIResourceSchemas

ncdc:
## Summary

Add support for CEL-based conversions for APIResourceSchemas.

Currently implemented:
- Added a new `APIConversion` resource
- Its name must match the name of an `APIResourceSchema` exactly
- When an `APIBinding` is processed, if any of the `APIResourceSchemas` it's binding to have `APIConversions`, the controller creates a copy of each `APIConversion` in `system:bound-crds`
    - The name of the `APIConversion` is set to the name of the corresponding CRD (which matches the UID of the `APIResourceSchema`)
- Here is an example `APIConversion`. These are the only supported features in this PR (we can iterate in future PRs):

```yaml
kind: APIConversion
metadata:
  name: rev0002.widgets.example.io
spec:
  conversions:
  - from: v1
    to: v2
    rules:
    - field: .spec.firstName
      destination: .spec.name.first
    - field: .spec.lastName
      destination: .spec.name.last
      transformation: self
    - field: .spec.lastName
      destination: .spec.name.lastUpper
      transformation: self.upperAscii()
  - from: v2
    to: v1
    rules:
    - field: .spec.name.first
      destination: .spec.firstName
    - field: .spec.name.last
      destination: .spec.lastName
    preserve:
    - .spec.someNewField
```
- When an `APIConversion` is created or updated, admission compiles every CEL expression, rejecting upon any errors
    - We need to decide if we want to allow updates; I just realized updates are not carried forward to the copy in `system:bound-crds`

Possible follow-ups in separate PRs:

- [ ] Figure out how to support indexing in maps and lists in CEL rules
- [ ] Figure out how to support indexing in maps and lists in "to" fields
- [ ] See if we can identify invalid evolutions and reject them
- [ ] What happens if a field's type changes / can we support that?
- [ ] Fully spec out preserve fields annotation-based flows

## Related issue(s)

Fixes #

Requires https://github.com/kcp-dev/kubernetes/pull/104

Part of #1671


ncdc:
Nice to see that I at least didn't break anything in the e2es, even if the conversion code isn't being exercised yet 😄

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
```diff
@@ -163,3 +163,88 @@ func (v *APIResourceVersion) SetSchema(schema *apiextensionsv1.JSONSchemaProps)
...
+	// +required
+
+	// rule is a CEL expression specifying the data to use from the source version to populate into the field referenced
+	// by 'to' in the target version. References to fields are relative to the 'from' version, such as 'v1.spec.name'.
+	Rule string `json:"rule"`
+
+	// +required
+
+	// to is a path to the field in the target version, relative to the root of the object, such as 'spec.name.first'.
+	To string `json:"to"`
```
sttts, line 239: can we rename one of the "to" fields? Source, destination, target, ....

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
```diff
@@ -163,3 +163,88 @@ func (v *APIResourceVersion) SetSchema(schema *apiextensionsv1.JSONSchemaProps)
...
+	Preserve []string `json:"preserve,omitempty"`
+}
+
+// APIConversionRule specifies how to convert a single field.
+type APIConversionRule struct {
+	// +required
+
+	// rule is a CEL expression specifying the data to use from the source version to populate into the field referenced
+	// by 'to' in the target version. References to fields are relative to the 'from' version, such as 'v1.spec.name'.
+	Rule string `json:"rule"`
```
sttts, line 234: am not convinced of the factoring. Would rather prefer this:

```yaml
from: v1
to: v2
rule:
   field: <extended-json-path>
   destination: <extended-json-path> # defaults to field
   transformation: <CEL> # default to self
```

This naturally extends to arrays and maps:
```yaml
from: v1
to: v2
rule:
   field: .spec.versions[*].labels
   destination: .spec.somewhere.else.versions[*].versionLabels
   transformation: self
```
Here the sequence of quantifiers `[*]` must match.

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
```diff
@@ -163,3 +163,88 @@ func (v *APIResourceVersion) SetSchema(schema *apiextensionsv1.JSONSchemaProps)
...
+
+	// +optional
+
+	// preserve contains a list of paths to fields to preserve, relative to the root of the object, such as
+	// 'spec.name.first'.
+	Preserve []string `json:"preserve,omitempty"`
+}
+
+// APIConversionRule specifies how to convert a single field.
+type APIConversionRule struct {
```
sttts, line 229: how can I delete fields?

sttts: nvmd. Deletion is not an issue with the compute model.

But wondering: what is the new object before the first rule? Empty? Or equal where types match?

sttts: And if the later, what is the type equality applied here?

ncdc: It's up to us. Currently, `to` starts as a complete copy of `from`. I believe because unknown fields are not preserved, any places where the types don't match, those fields are dropped.

sttts: ic, so basically a unstructured copy that is then pruned.

> any places where the types don't match

This we have to understand very well. What does it mean to not match? Is it basically JSON marshal-unmarshal compatibility?

ncdc: I mistyped. Any places where the field paths don't match, those fields are dropped. e.g. if v1 has spec.firstName and v2 has spec.name.first, the conversion will create an map that looks like this

```yaml
spec:
  firstName: Andy
  name:
    first: Andy
```

and then spec.firstName is pruned (somewhere - haven't tracked down the specific place) so by the time you get it out as v2, the only thing left is spec.name.first

ncdc:
Pushed an update based on refactoring in https://github.com/kcp-dev/kubernetes/pull/104

pkg/server/api_conversion.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+
+	return converter.Convert(in, targetGV)
+}
+
+func (f *deferredConverter) getConverter() (conversion.Converter, error) {
+	f.lock.Lock()
+	defer f.lock.Unlock()
+
+	converter, ok := f.celConverters[string(f.crd.UID)]
```
csams, line 116: If someone makes a mistake with a conversion definition and wants to update it, can this cache be cleared?

pkg/server/api_conversion.go
```diff
@@ -0,0 +1,325 @@
...
+	for i := range list.Items {
+		original := &list.Items[i]
+		converted := original.DeepCopy()
+		converted.SetAPIVersion(targetGV.String())
+
+		originalVersion := original.GetObjectKind().GroupVersionKind().Version
+		compiledRules := c.compiledRules[originalVersion]
+		for _, rule := range compiledRules {
+			if rule.Err != nil {
+				// TODO(ncdc): what to do here?
```
nrb, line 209: I would expect a rule that couldn't compile to be rejected at admission, and that we couldn't get to this function if there were any entries with `Err` populated.

That would require logic to set status on the APIConversion and handle updates, though.

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
```diff
@@ -24,17 +24,17 @@ import (
...
-// A APIResourceSchema is immutable and cannot be deleted if they are referenced by
-// an APIExport in the same workspace.
-//
 // +crd
 // +genclient
 // +genclient:nonNamespaced
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:resource:scope=Cluster,categories=kcp
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
+
```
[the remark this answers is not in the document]
ncdc: Those are incorrect 😄

pkg/admission/apiconversion/apiconversion_admission.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+// Validate validates the creation and updating of APIBinding resources. It also performs a SubjectAccessReview
+// making sure the user is allowed to use the 'bind' verb with the referenced APIExport.
+func (o *apiConversionAdmission) Validate(ctx context.Context, a admission.Attributes, _ admission.ObjectInterfaces) error {
+	if a.GetResource().GroupResource() != apisv1alpha1.Resource("apiconversions") {
+		return nil
+	}
+
+	cluster, err := genericapirequest.ValidClusterFrom(ctx)
```
stevekuznetsov, line 76: Totally unrelated question - I have recently hacked cluster name into admission attributes, necessary for webhook plubming. Can we rally around that and not need this "genericapirequest" stuff?

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
```diff
@@ -163,3 +163,93 @@ func (v *APIResourceVersion) SetSchema(schema *apiextensionsv1.JSONSchemaProps)
...
+	// +required
+	// +listType=map
+	// +listMapKey=destination
+
+	// rules contains field-specific conversion expressions.
+	Rules []APIConversionRule `json:"rules"`
+
+	// +optional
+
+	// preserve contains a list of JSONPath expressions to fields to preserve in the originating version of the object, relative to its, such as
```
stevekuznetsov, line 223: Relative to its what?

pkg/conversion/conversion.go
```diff
@@ -0,0 +1,176 @@
...
+
+	"k8s.io/apiextensions-apiserver/pkg/apiserver/schema"
+	crdcel "k8s.io/apiextensions-apiserver/pkg/apiserver/schema/cel"
+	"k8s.io/apiextensions-apiserver/pkg/apiserver/schema/cel/library"
+	"k8s.io/apiextensions-apiserver/third_party/forked/celopenapi/model"
+
+	"github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
+)
+
+const celCheckFrequency = 100
```
stevekuznetsov, line 35: This needs a unit :)

pkg/conversion/conversion.go
```diff
@@ -0,0 +1,176 @@
...
+
+// CompiledRule contains the compiled cel.Program to convert a single field. If there was a compilation error, that is
```
stevekuznetsov, line 37:
```suggestion
// CompiledRule contains the compiled cel.Program to convert a single field. If there were a compilation error, that would be
```

(if we care about correct mood)

stevekuznetsov (commented):
How are conversion compilation errors surfaced to the user?

pkg/admission/apiconversion/apiconversion_admission.go
ncdc: Sure thing

stevekuznetsov:
Part of https://github.com/kcp-dev/kcp/issues/2014

pkg/admission/apiconversion/apiconversion_admission.go
ncdc: @stevekuznetsov I see that our generic webhook admission impl sets the cluster in the attrs, but I don't see it happening anywhere else. Do we need to plumb this in to the various handlers e.g. createHandler upstream?

ncdc:
@stevekuznetsov 
> How are conversion compilation errors surfaced to the user?

Directly, as errors, implemented via admission

pkg/server/api_conversion.go
ncdc: You would have to create a new APIResourceSchema with a different name (don't have to change its spec though), a new APIConversion with a matching name, and update the APIExport to point at the new schema. We can look at making this easier in the future as a follow-up.

pkg/server/api_conversion.go
ncdc: The APIConversion is rejected at admission if any conversion specifications are invalid. i.e. it's not even created. I guess we should never hit this condition?

ncdc:
Working on adding unit tests

pkg/admission/apiconversion/apiconversion_admission.go
sttts: what about the inverse, remove the getters from the attributes? Am not a big fan of changing an upstream interface.

pkg/admission/apiconversion/apiconversion_admission.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return admission.NewForbidden(a, fmt.Errorf("error determining workspace: %w", err))
+	}
+
+	if cluster.Name == apibinding.SystemBoundCRDsClusterName {
+		// TODO(ncdc): do we also want to validate these conversions? They've already been validated once (in their
+		// original logical cluster).
```
sttts, line 81: we won't copy APIConversions into `system:bound-crds`, will we?

sttts: neither into the bound workspace (I hope 🙏)

ncdc: My current code copies from the source workspace into system:bound-crds, but that is just the initial implementation. Will change it to use what's in the source workspace instead of copying.

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
```diff
@@ -163,3 +163,95 @@ func (v *APIResourceVersion) SetSchema(schema *apiextensionsv1.JSONSchemaProps)
...
+
+// APIConversionSpec contains rules to convert between different API versions in an APIResourceSchema.
+type APIConversionSpec struct {
+	// +required
+	// +listType=map
+	// +listMapKey=from
+	// +listMapKey=to
+
+	// conversions specify rules to convert between different API versions in an APIResourceSchema.
+	Conversions []APIVersionConversion `json:"conversions"`
```
sttts, line 192: <bike-shedding>rules?</bike-shedding> No strong opinion, just another option.

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
```diff
@@ -163,3 +163,95 @@ func (v *APIResourceVersion) SetSchema(schema *apiextensionsv1.JSONSchemaProps)
...
+	// +listType=map
+	// +listMapKey=destination
+
+	// rules contains field-specific conversion expressions.
+	Rules []APIConversionRule `json:"rules"`
+
+	// +optional
+
+	// preserve contains a list of JSONPath expressions to fields to preserve in the originating version
+	// of the object, relative to its root, such as '.spec.name.first'.
```
sttts, line 224: can we specify via a regex?

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
```diff
@@ -163,3 +163,95 @@ func (v *APIResourceVersion) SetSchema(schema *apiextensionsv1.JSONSchemaProps)
...
+	// destination is a JSONPath expression to the field in the target version of the object, relative to
+	// its root, such as '.spec.name.first'.
+	Destination string `json:"destination"`
+
+	// +optional
+
+	// transformation is an optional CEL expression used to execute user-specified rules to transform the
+	// originating field -- identified by 'self' -- to the destination field.
+	Transformation string `json:"transformation,omitempty"`
+}
```
sttts, line 247: am still thinking we are missing something here:

1. assume destination `a.b[*].c` and field `a.b[*]`. Then we can access e.g. a.b[*].old_c` to derive `a.b[42].c`. That's fine. So I guess field must be a prefix of destination.
2. assume destination `a.b[*].c` and field `a.b`. Then we can access e.g. `a.b[42].old_c` to derive `a.b[42].c`, i.e. copy a static value into each item in the list.
3. assume destination `a.b[*].c` and field `a.b`. Then we can access e.g. a.b.old_c` to derive `a.b[42].c`, but we cannot reference `a.b[42].d` at the same time. **We could map `root, a, a.b, a.b[42]` into the CEL context to give access to every level. Any better idea**?
4. we would also like to access the index, or in case of maps the key. We might also want to find a way to put that in the CEL context.

ncdc: This PR doesn't support indexing currently

pkg/cache/server/config.go
```diff
@@ -200,32 +202,30 @@ func NewConfig(opts *cacheserveroptions.CompletedOptions, optionalLocalShardRest
...
-			// error.
-			ServiceResolver:        &unimplementedServiceResolver{},
+			CRDRESTOptionsGetter:   apiextensionsoptions.NewCRDRESTOptionsGetter(*opts.Etcd),
 			MasterCount:            1,
-			AuthResolverWrapper:    webhook.NewDefaultAuthenticationInfoResolverWrapper(nil, nil, rt, nil),
 			Client:                 c.ApiExtensionsClusterClient,
 			Informers:              c.ApiExtensionsSharedInformerFactory,
 			ClusterAwareCRDLister:  &crdClusterLister{lister: c.ApiExtensionsSharedInformerFactory.Apiextensions().V1().CustomResourceDefinitions().Lister()},
 			DisableServerSideApply: true,
+			ConversionFactory:      conversion.NewCRConverterFactory(c.KCPSharedInformerFactory.Apis().V1alpha1().APIConversions()),
```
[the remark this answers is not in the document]
ncdc: No, it's required.

pkg/admission/apiconversion/apiconversion_admission.go
sttts: look into the cache server (and today into the source workspace, yes).

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
sttts: No I mean to give a regex for a JSON path.

pkg/cache/server/config.go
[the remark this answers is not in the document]
sttts: But the cache server should not convert. Do we need this complexity?

pkg/client/clientset/versioned/cluster/typed/apis/v1alpha1/apiconversion.go
```diff
@@ -0,0 +1,72 @@
...
+
+// APIConversionClusterInterface can operate on APIConversions across all clusters,
+// or scope down to one cluster and return a apisv1alpha1client.APIConversionInterface.
+type APIConversionClusterInterface interface {
+	Cluster(logicalcluster.Path) apisv1alpha1client.APIConversionInterface
+	List(ctx context.Context, opts metav1.ListOptions) (*apisv1alpha1.APIConversionList, error)
+	Watch(ctx context.Context, opts metav1.ListOptions) (watch.Interface, error)
+}
+
+type aPIConversionsClusterInterface struct {
```
sttts, line 51: api

ncdc: This is generated code. Will see if I can influence what's generated.

pkg/conversion/cel_converter.go
```diff
@@ -0,0 +1,214 @@
...
+			return nil, fmt.Errorf("failed converting version %s validation to internal version: %w", v.Name, err)
+		}
+
+		structuralSchema, err := structuralschema.NewStructural(internalJSONSchemaProps)
+		if err != nil {
+			return nil, fmt.Errorf("error getting structural schema for version %s: %w", v.Name, err)
+		}
+
+		// Deep copy because PruneDefaults is mutating
+		structuralSchema = structuralSchema.DeepCopy()
```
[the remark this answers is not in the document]
ncdc: Copied from upstream...

pkg/conversion/cel_converter.go
```diff
@@ -0,0 +1,214 @@
...
+				continue
+			}
+
+			toPath := rule.ToPath
+			if toPath[0] == '.' {
+				toPath = toPath[1:]
+			}
+			fields := strings.Split(toPath, ".")
+			if err := unstructured.SetNestedField(converted.Object, fromValue, fields...); err != nil {
+				return nil, fmt.Errorf("error setting target field %q: %w", rule.ToPath, err)
```
sttts, line 123: destination

pkg/conversion/cel_converter.go
```diff
@@ -0,0 +1,214 @@
...
+			}
+
+			if rule.Program == nil {
+				continue
+			}
+
+			// structuralSchema := c.structuralSchemas[originalVersion]
+			bindings := map[string]interface{}{
+				"self": fromValue, // crdcel.UnstructuredToVal(original.Object, structuralSchema),
+			}
```
sttts, line 133: nice, super simple. So we can easily add more values to the context per activation.

pkg/conversion/cel_converter.go
```diff
@@ -0,0 +1,214 @@
...
+			activation, err := interpreter.NewActivation(bindings)
+			if err != nil {
+				return nil, fmt.Errorf("error creating CEL interpreter activation: %w", err)
+			}
+
+			evalResult, evalDetails, err := rule.Program.ContextEval(context.TODO(), activation)
```
sttts, line 139: some context with timeout maybe? E.g. 1 sec.

pkg/conversion/cel_converter.go
```diff
@@ -0,0 +1,214 @@
...
+				return nil, fmt.Errorf("error setting target field %q: %w", rule.ToPath, err)
+			}
+		}
+
+		annotations := converted.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+
+		preserveAnnotation := annotations[targetGV.Version+".conversion.apis.kcp.dev/preserve"]
```
sttts, line 155: kcp.io

pkg/conversion/cel_converter.go
```diff
@@ -0,0 +1,214 @@
...
+		structuralSchema = structuralSchema.DeepCopy()
+
+		if err := structuraldefaulting.PruneDefaults(structuralSchema); err != nil {
+			return nil, fmt.Errorf("error pruning defaults for version %s: %w", v.Name, err)
+		}
+
+		structuralSchemas[v.Name] = structuralSchema
+	}
+
+	compiledRules, err := CompileConversions(apiConversion, structuralSchemas)
```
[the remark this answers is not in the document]
ncdc: I got rid of rule.Err in the most recent push. I now aggregate and return errors all the way back to here, so we don't have to iterate over `compiledRules` any more.

pkg/conversion/conversion_rules.go
```diff
@@ -0,0 +1,208 @@
...
+
+	var propDecls []*expr.Decl
+	propDecls = append(propDecls, decls.NewVar("self", root.ExprType()))
+
+	opts = append(opts, cel.Declarations(propDecls...), cel.HomogeneousAggregateLiterals())
+	opts = append(opts, library.ExtensionLibs...)
+	return env.Extend(opts...)
+}
+
+func schemaIndexByJSONPath(s *structuralschema.Structural, path string) (*structuralschema.Structural, error) {
```
sttts, line 182: "index" irritates me. Why not `schemaByJSONPath`?

sttts: I guess you mean `schema(IndexByJSONPath)`, not `(schemaIndex)ByJSONPath` (what I read)

pkg/conversion/conversion_rules.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		return nil, field.Invalid(rulePath.Child("transformation"), rule.Transformation, fmt.Sprintf("error creating CEL program: %v", err))
+	}
+
+	cr.Program = program
+
+	return cr, nil
+}
+
+func getCELEnv(structuralSchema *structuralschema.Structural) (*cel.Env, error) {
```
sttts, line 141: merely `createCELEnv`. This is not a getter.

pkg/conversion/conversion_rules.go
```diff
@@ -0,0 +1,208 @@
...
+
+	for i, f := range fields {
+		// While iterating, i-1 is the index of the parent field
+		if lastParentIndex >= 0 && (i-1) <= lastParentIndex && cursor.Type != "object" {
+			return nil, fmt.Errorf("expected field %q to be an object", strings.Join(visited, "."))
+		}
+
+		visited = append(visited, f)
+
+		property, exists := cursor.Properties[f]
```
sttts, line 200: also `AdditionalProperties` is possible.

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
```diff
@@ -163,3 +163,95 @@ func (v *APIResourceVersion) SetSchema(schema *apiextensionsv1.JSONSchemaProps)
...
+	Preserve []string `json:"preserve,omitempty"`
+}
+
+// APIConversionRule specifies how to convert a single field.
+type APIConversionRule struct {
+	// +required
+
+	// field is a JSONPath expression to the field in the originating version of the object, relative to its root, such
+	// as '.spec.name.first'.
+	Field string `json:"field"`
```
sttts, line 234: missed this before: here we defnitely need a very restrictive spec, e.g. regex or in admission.

pkg/conversion/converter.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+type Converter struct {
+	crd               *apiextensionsv1.CustomResourceDefinition
+	compiledRules     map[string][]*CompiledRule
+	structuralSchemas map[string]*structuralschema.Structural
+	apiConversion     *apisv1alpha1.APIConversion
+}
+
+func (c *Converter) Convert(list *unstructured.UnstructuredList, targetGV schema.GroupVersion) (*unstructured.UnstructuredList, error) {
```
sttts, line 87: not sure that is documented elsewhere: the input list is never mutated?

pkg/conversion/converter.go
```diff
@@ -0,0 +1,245 @@
...
+			return nil, fmt.Errorf("error setting destination field %q: %w", rule.ToPath, err)
+		}
+	}
+
+	annotations := converted.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+
+	preserveAnnotation := annotations[targetGV.Version+".conversion.apis.kcp.io/preserve"]
```
sttts, line 133: constant please, even in the API package. This is technically an API.

pkg/conversion/deferred_converter.go
```diff
@@ -0,0 +1,103 @@
...
+
+	return converter.Convert(in, targetGV)
+}
+
+func (f *deferredConverter) getConverter() (conversion.CRConverter, error) {
+	f.lock.Lock()
+	defer f.lock.Unlock()
+
+	if f.delegate != nil {
+		return f.delegate, nil
```
sttts, line 63: this surprises. What is a delegate in practice?

sttts: i.c., you cache the result. Maybe add a oneliner comment, or rename to  `cachedConverter` or similar.

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -211,6 +197,23 @@ func NewController(
...
+
+	// APIExport-related indexers
+	indexers.AddIfNotPresentOrDie(apiExportInformer.Informer().GetIndexer(), cache.Indexers{
+		indexers.ByLogicalClusterPathAndName: indexers.IndexByLogicalClusterPathAndName,
+		indexAPIExportsByAPIResourceSchema:   indexAPIExportsByAPIResourceSchemasFunc,
+	})
+	indexers.AddIfNotPresentOrDie(globalAPIExportInformer.Informer().GetIndexer(), cache.Indexers{
+		indexers.ByLogicalClusterPathAndName: indexers.IndexByLogicalClusterPathAndName,
+		indexAPIExportsByAPIResourceSchema:   indexAPIExportsByAPIResourceSchemasFunc,
+	})
```
sttts, line 209: I used to put all indexers first, then all event handlers. I assumed this is a convention we follow.

pkg/conversion/conversion_rules.go
ncdc: Renaming to `getStructuralSchemaForField`

pkg/conversion/conversion_rules.go
ncdc: Fixing

pkg/conversion/conversion_rules.go
ncdc: Fixing

pkg/conversion/converter.go
ncdc: It would not be good if the input list were mutated. There's no guard that prevents it currently. I might need to make a slight change to the upstream code:

```go
convertedObjects, err := c.converter.Convert(list, toGVK.GroupVersion())
// error checking

convertedList := list.DeepCopy()
// loop through convertedObjects and assign each one to the appropriate index in 'list'
```

If `list` is modified by `c.converter.Convert`, results are not guaranteed. I should move the DeepCopy up.

ncdc: https://github.com/kubernetes/kubernetes/pull/114959, will pull into our fork

pkg/conversion/converter.go
ncdc: Fixing

pkg/admission/apiconversion/apiconversion_admission.go
stevekuznetsov: If we don't have it up there, we have to refactor the entire call-chain to pass it through for generic webhook admission.

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
```diff
@@ -24,17 +24,17 @@ import (
...
 // +crd
 // +genclient
 // +genclient:nonNamespaced
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:resource:scope=Cluster,categories=kcp
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
+
+// APIResourceSchema describes a resource, identified by (group, version, resource, schema).
+//
+// A APIResourceSchema is immutable and cannot be deleted if they are referenced by
```
stevekuznetsov, line 36:
```suggestion
// An APIResourceSchema is immutable and cannot be deleted if it is referenced by
```

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
```diff
@@ -163,3 +163,103 @@ func (v *APIResourceVersion) SetSchema(schema *apiextensionsv1.JSONSchemaProps)
...
+	metav1.ObjectMeta `json:"metadata"`
+
+	// Spec holds the desired state.
+	Spec APIConversionSpec `json:"spec"`
+}
+
+// APIConversionSpec contains rules to convert between different API versions in an APIResourceSchema.
+type APIConversionSpec struct {
+	// +required
+	// +listType=map
```
stevekuznetsov, line 194: For types where these attributes were *also* added to struct tags, having them on separate lines caused an error:

```
2023/01/11 07:57:40 OpenAPI code generation error: Failed executing generator: some packages had errors:
Tags in comment and struct should match for member (SubResources) of (github.com/kcp-dev/kcp/pkg/apis/apiresource/v1alpha1.CommonAPIResourceSpec)
```

Why is it correct to do this here?

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
stevekuznetsov: Why? Don't we get something stronger when compiling/validating?

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
```diff
@@ -163,3 +163,103 @@ func (v *APIResourceVersion) SetSchema(schema *apiextensionsv1.JSONSchemaProps)
...
+	// +kubebuilder:validation:MinLength=1
+
+	// field is a JSONPath expression to the field in the originating version of the object, relative to its root, such
+	// as '.spec.name.first'.
+	Field string `json:"field"`
+
+	// +required
+
+	// destination is a JSONPath expression to the field in the target version of the object, relative to
+	// its root, such as '.spec.name.first'.
```
stevekuznetsov, line 247: nit: it might be useful to have one coherent example that all the GoDoc references, for clarity

pkg/conversion/conversion_factory.go
```diff
@@ -0,0 +1,73 @@
...
+	"k8s.io/apiextensions-apiserver/pkg/apiserver/conversion"
+
+	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
+	apisinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/apis/v1alpha1"
+)
+
+// CRConverterFactory instantiates converters that are capable of converting custom resources between different API
+// versions. It supports CEL-based conversion rules from APIConversion resources and the "none" conversion strategy.
+type CRConverterFactory struct {
+	nopConverter     conversion.CRConverter
```
stevekuznetsov, line 34: Is this a costly thing to create ephemerally? Surprised we need a shared instance here

pkg/conversion/conversion_factory.go
```diff
@@ -0,0 +1,73 @@
...
+func NewCRConverterFactory(apiConversionInformer apisinformers.APIConversionClusterInformer) *CRConverterFactory {
+	return &CRConverterFactory{
+		nopConverter: conversion.NewNOPConverter(),
+		getAPIConversion: func(clusterName logicalcluster.Name, name string) (*apisv1alpha1.APIConversion, error) {
+			return apiConversionInformer.Lister().Cluster(clusterName).Get(name)
+		},
+	}
+}
+
+// NewConverter returns the appropriate conversion.Converter based on the CRD. If the CRD identifies as for a wildcard
```
stevekuznetsov, line 51: "If the CRD identifies as for a wildcard" is not super clear

pkg/conversion/converter.go
```diff
@@ -0,0 +1,245 @@
...
+
+type Converter struct {
+	crd               *apiextensionsv1.CustomResourceDefinition
+	compiledRules     map[string][]*CompiledRule
+	structuralSchemas map[string]*structuralschema.Structural
+	apiConversion     *apisv1alpha1.APIConversion
+}
+
+func (c *Converter) Convert(list *unstructured.UnstructuredList, targetGV schema.GroupVersion) (*unstructured.UnstructuredList, error) {
+	ctx := context.TODO()
```
stevekuznetsov, line 88: Isn't this conversion happening in the context of a client request? Can we plumb that?

pkg/conversion/converter.go
```diff
@@ -0,0 +1,245 @@
...
+
+	originalVersion := original.GetObjectKind().GroupVersionKind().Version
+	compiledRules := c.compiledRules[originalVersion]
+	for _, rule := range compiledRules {
+		value, err := evaluateRule(ctx, original, rule)
+		if err != nil {
+			return nil, fmt.Errorf("error converting: %w", err)
+		}
+
+		toPath := rule.ToPath
```
stevekuznetsov, line 117: Do we need to keep the string-formatted value in `rule`? Why not store the array?

pkg/conversion/converter.go
```diff
@@ -0,0 +1,245 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("error preserving fields: %w", err)
+		}
+
+		if len(m) > 0 {
+			encoded, err := json.Marshal(m)
+			if err != nil {
+				return nil, fmt.Errorf("error encoding preserve map: %v", err)
+			} else {
+				annotations[originalVersion+apisv1alpha1.VersionPreservationAnnotationKeySuffix] = string(encoded)
```
stevekuznetsov, line 159: From prior experience with our annotation-stuffing schemes - how can we be certain that targetGV.Version+apisv1alpha1.VersionPreservationAnnotationKeySuffix is a valid annotation key?

pkg/conversion/conversion_factory.go
```diff
@@ -0,0 +1,73 @@
...
+// partial metadata request, the nop converter is used. Otherwise, it returns a CEL-based converter if there is an
+// associated APIConversion, a nop converter if the strategy is "none", or an error otherwise.
+func (f *CRConverterFactory) NewConverter(crd *apiextensionsv1.CustomResourceDefinition) (conversion.CRConverter, error) {
```
stevekuznetsov, line 54: Are we guaranteed that this is only called once per CRD? If not, do we need to share/dedupe the outcome of the deferred converter?

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
ncdc: As discussed on Slack: admission is acceptable.

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
ncdc: Not sure?

pkg/conversion/conversion_factory.go
ncdc: No, it's not costly. Can create on the fly if you'd prefer.

pkg/conversion/converter.go
ncdc: Requires a huge upstream change

pkg/conversion/converter.go
ncdc: You mean during rule compilation?

pkg/conversion/converter.go
ncdc: It should be. If we find issues, we can adjust it.

pkg/conversion/conversion_factory.go
ncdc: Yes, it's called by `getOrCreateServingInfoFor` only if it's not in the storageMap

pkg/admission/apiconversion/apiconversion_admission.go
```diff
@@ -0,0 +1,153 @@
...
+
+type apiConversionAdmission struct {
+	*admission.Handler
+
+	getAPIResourceSchema func(clusterName logicalcluster.Name, name string) (*apisv1alpha1.APIResourceSchema, error)
+}
+
+// Ensure that the required admission interfaces are implemented.
+var (
+	_ = admission.ValidationInterface(&apiConversionAdmission{})
```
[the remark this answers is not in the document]
ncdc: Fixing

pkg/conversion/conversion.go
ncdc: PTAL at what I added

pkg/admission/apiconversion/apiconversion_admission.go
ncdc: Fixed

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
ncdc: As discussed on Meet, the API in the PR is ok to merge, and we can do follow-ups to evolve to support more advanced use cases.

pkg/conversion/cel_converter.go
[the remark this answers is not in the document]
ncdc: Removed

pkg/conversion/conversion_factory.go
ncdc: Adjusted to `If the CRD identifies as for a "wildcard partial metadata request"`

pkg/conversion/converter.go
stevekuznetsov: Doesn't this depend on the length of `originalVersion`? Is that bounded?

pkg/conversion/converter.go
stevekuznetsov: :(

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
stevekuznetsov: Should make sure we don't need the struct-tags?

pkg/conversion/converter.go
ncdc: If it's safer to hash, let's do that. @sttts ?

ncdc: Max length of originalVersion is 63 chars

ncdc: Will change to $prefix/$name - prefix can be 253, name 63 (https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set)

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
vincepri: IIRC These are required if we're using client-go like clients. The struct tags are used in https://github.com/kubernetes/kubernetes/blob/6ce055d62d8608842db18901c948c6982c74a164/staging/src/k8s.io/apimachinery/third_party/forked/golang/json/fields.go which is then used in the strategicpatch package https://github.com/kubernetes/kubernetes/blob/6ce055d62d8608842db18901c948c6982c74a164/staging/src/k8s.io/apimachinery/pkg/util/strategicpatch/meta.go#L54

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
```diff
@@ -163,3 +163,104 @@ func (v *APIResourceVersion) SetSchema(schema *apiextensionsv1.JSONSchemaProps)
...
+// APIResourceSchema. Additionally, to avoid data loss when round-tripping from a version that
+// contains a new field to one that doesn't and back again, you can specify a list of fields to
+// preserve (these are stored in annotations).
+type APIVersionConversion struct {
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Pattern=^v[1-9][0-9]*([a-z]+[1-9][0-9]*)?$
+
+	// from is the source version.
+	From string `json:"from"`
```
vincepri, line 212:
```suggestion
        // from is the source version.
        //
	// +required
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:Pattern=^v[1-9][0-9]*([a-z]+[1-9][0-9]*)?$
	From string `json:"from"`
```

Is there a reason why we don't prefer the pattern I've seen above? Having everything in the same block makes it easier (at least for me, but others might disagree) to parse which rules are associated with which field, and keeps everything together.

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
vincepri: Could we offer globs instead of full regex? Like `spec.newStuff.*`, etc

pkg/conversion/conversion_rules.go
```diff
@@ -0,0 +1,227 @@
...
+			errs = append(errs, err)
+			continue
+		}
+		rulesForVersion = append(rulesForVersion, cr)
+	}
+
+	return rulesForVersion, kerrors.NewAggregate(errs)
+}
+
+func compileRule(rulePath *field.Path, rule apisv1alpha1.APIConversionRule, schema *structuralschema.Structural) (*CompiledRule, error) {
```
vincepri, line 99: add a comment?

pkg/conversion/conversion_rules.go
```diff
@@ -0,0 +1,227 @@
...
+			continue
+		}
+
+		compiledRules[c.From] = rulesForVersion
+	}
+
+	return compiledRules, kerrors.NewAggregate(errs)
+}
+
+func compileConversion(path *field.Path, c *apisv1alpha1.APIVersionConversion, structuralSchemas map[string]*structuralschema.Structural) ([]*CompiledRule, error) {
```
vincepri, line 77: add a comment?

pkg/conversion/conversion_rules.go
```diff
@@ -0,0 +1,227 @@
...
+		rulesForVersion = append(rulesForVersion, cr)
+	}
+
+	return rulesForVersion, kerrors.NewAggregate(errs)
+}
+
+func compileRule(rulePath *field.Path, rule apisv1alpha1.APIConversionRule, schema *structuralschema.Structural) (*CompiledRule, error) {
+	fromPath := rule.Field
+	if fromPath[0] == '.' {
+		fromPath = fromPath[1:]
```
vincepri, line 102: Add reasoning why this is needed?

pkg/conversion/conversion_rules.go
```diff
@@ -0,0 +1,227 @@
...
+	if err != nil {
+		return nil, field.Invalid(rulePath.Child("field"), rule.Field, err.Error())
+	}
+
+	if rule.Transformation == "" {
+		return cr, nil
+	}
+
+	env, err := createCELEnv(schemaForField)
+	if env == nil {
```
vincepri, line 127: Could we try to always return an error when a nil environment is returned?

pkg/conversion/conversion_rules.go
```diff
@@ -0,0 +1,227 @@
...
+
+	// inline local copy of upstream's generateUniqueSelfTypeName()
+	scopedTypeName := fmt.Sprintf("selfType%d", time.Now().Nanosecond())
+
+	ruleTypes, err := model.NewRuleTypes(scopedTypeName, structuralSchema, true, registry)
+	if err != nil {
+		return nil, fmt.Errorf("error creating rule types: %w", err)
+	}
+	if ruleTypes == nil {
+		return nil, nil
```
vincepri, line 169: Should this be returning an error?

pkg/conversion/conversion_rules.go
```diff
@@ -0,0 +1,227 @@
...
+}
+
+func getStructuralSchemaForField(s *structuralschema.Structural, fields ...string) (*structuralschema.Structural, error) {
+	cursor := s
+
+	visited := make([]string, 0, len(fields))
+
+	// spec.name.first ==> 1
+	//  0    1     2
+	lastParentIndex := len(fields) - 2
```
vincepri, line 201: Is this always guaranteed to be at least 2?

pkg/conversion/conversion_rules.go
```diff
@@ -0,0 +1,227 @@
...
+
+	for i, f := range fields {
```
vincepri, line 203: Without knowing how CEL works, this for loop might be a bit hard to understand, in terms of why and how, could we add some comments explaining the logic and reasoning?

vincepri (commented):
~ Half way through

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -265,6 +265,30 @@ func (r *bindingReconciler) reconcile(ctx context.Context, apiBinding *apisv1alp
 			return reconcileStatusContinue, nil
 		}
 
+		// If there are multiple versions, there must be an APIConversion
+		if len(schema.Spec.Versions) > 1 {
+			_, err = r.getAPIConversion(logicalcluster.From(schema), schema.Name)
+			if err != nil {
```
vincepri, line 271:
```suggestion
			if _, err := r.getAPIConversion(logicalcluster.From(schema), schema.Name)
```
?

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
ncdc: Godoc generation. If you keep it all together, the directives are included in the doc

pkg/conversion/converter.go
```diff
@@ -0,0 +1,231 @@
...
+
+	bindings := map[string]interface{}{
+		"self": fromValue,
+	}
+	activation, err := interpreter.NewActivation(bindings)
+	if err != nil {
+		return nil, fmt.Errorf("error creating CEL interpreter activation: %w", err)
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, time.Second)
```
vincepri, line 187: Should this be configurable? Is it necessary for the first iteration of this API?

vincepri (commented):
~ Half way through

pkg/conversion/converter.go
ncdc: @sttts suggested putting a timeout in here instead of `context.TODO()`. 1 second per rule execution seems super generous (probably too much, but maybe CI needs it). I also considered an overall timeout in `convert` and passing that in here. Maybe that's better?

Not sure it should be configurable. This is to protect the apiserver from runaway conversions.

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
ncdc: Potentially in a future PR

pkg/admission/apiconversion/apiconversion_admission.go
```diff
@@ -0,0 +1,144 @@
...
+	_ admission.ValidationInterface     = (*apiConversionAdmission)(nil)
+	_ admission.InitializationValidator = (*apiConversionAdmission)(nil)
+	_ initializers.WantsKcpInformers    = (*apiConversionAdmission)(nil)
+)
+
+// Validate ensures all the conversion rules specified in an APIConversion are correct.
+func (o *apiConversionAdmission) Validate(ctx context.Context, a admission.Attributes, _ admission.ObjectInterfaces) error {
+	if a.GetResource().GroupResource() != apisv1alpha1.Resource("apiconversions") {
+		return nil
+	}
```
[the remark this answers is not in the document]
ncdc: No, because all admission plugins are invoked on every request

pkg/apis/apis/v1alpha1/types_apiresourceschema.go
ncdc: CRDs don't support strategic merge, if that's what we're talking about? Also, none of our other code does this today. Suggest deferring to a wholesale repo follow-up if needed.

pkg/conversion/deferred_converter.go
ncdc: I originally did this because at the time, I was copying the APIConversion from the source ws into system:bound-crds, and there was no guarantee the APIConversion was in the informer's cache when the serving info is created for the bound CRD (the converter is created at that time). But now that I'm not copying, we could consider removing this deferred converter entirely. The only upside of keeping it is that if you create a normal CRD and later you create an APIConversion, this deferredConverter allows the APIConversion to take effect. Otherwise, you have to deleted the CRD and recreate it.

pkg/conversion/converter.go
vincepri: Ah, this is _per-rule_ which yeah makes it very generous indeed; making it global might make sense

pkg/conversion/conversion_rules.go
vincepri: Thanks now it's much more clear!

pkg/conversion/conversion_rules.go
```diff
@@ -0,0 +1,241 @@
...
+	//  0    1     2
+	lastParentIndex := len(fields) - 2
+
+	// Starting with the initial field (e.g. "spec"), try to resolve the next segment (e.g. "name") until we get
+	// to the desired field (e.g. "first").
+	for i, f := range fields {
+		// Verify that each intermediate field is an object. While iterating, i-1 is the index of the parent field.
+		if lastParentIndex >= 0 && (i-1) <= lastParentIndex && cursor.Type != "object" {
+			return nil, fmt.Errorf("expected field %q to be an object", strings.Join(visited, "."))
+		}
```
vincepri, line 216: Assuming I've understood this logic correctly, these checks are basically saying "every field in the path before the last one has to be of type `object`. If that's the case, given that we're iterating through `fields` from outermost to innermost, couldn't we check the `Type` field before doing the `continue` in Properties/AdditionalProperties below?

pkg/conversion/converter.go
```diff
@@ -0,0 +1,271 @@
...
+		},
+		[]string{},
+	)
+)
+
+func NewConverter(
+	crd *apiextensionsv1.CustomResourceDefinition,
+	apiConversion *apisv1alpha1.APIConversion,
+	objectCELTransformationsTimeout time.Duration,
+) (*Converter, error) {
```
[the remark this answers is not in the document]
ncdc: We have multiple types in this package that we're constructing

vincepri (approved):
/lgtm

pkg/admission/apiconversion/apiconversion_admission.go
ncdc: If needed, will do in a follow-up

ncdc: #2621 to track

PR #2106, opened 2022-09-30
:bug: fix- nil context in workspace-use plugin

varshaprasad96:
## Summary
Wrong command context was passed in `ws use` subcommand because of which there was a null pointer error when using it.

Issue:
Workspace use command run from building kubectl-kcp binary gives null pointer error. The passed command context was from parent (workspace command) was getting closed, causing seg fault. 
```
➜  kcp git:(debug/ws-use) ✗ ./bin/kubectl-kcp ws use webapp-back
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x4d6b83d]

goroutine 1 [running]:
golang.org/x/time/rate.(*Limiter).WaitN(0xc000181c20, {0x0, 0x0}, 0x1)
	/Users/vnarsing/go/pkg/mod/golang.org/x/time@v0.0.0-20220210224613-90d013bbcef8/rate/rate.go:234 +0x1bd
```
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cliplugins/workspace/cmd/cmd.go
+++ b/pkg/cliplugins/workspace/cmd/cmd.go
@@ -103,7 +103,7 @@ func New(streams genericclioptions.IOStreams) (*cobra.Command, error) {
 			if err := useWorkspaceOpts.Validate(); err != nil {
 				return err
 			}
-			return useWorkspaceOpts.Run(cmd.Context())
+			return useWorkspaceOpts.Run(c.Context())
 		},
 	}
 	useWorkspaceOpts.BindFlags(useCmd)


ncdc (approved):
Good catch!

kasturinarra:
@varshaprasad96 one quick question, i am trying to run command `kubectl ws use qetest5` and kcp throws error which reads "No resources found" and now i am trying to create the same workspace using the command `kubectl ws create qetest5` and it says workspace already exists, but it actually gets created then, is this expected ? thanks !!

varshaprasad96:
@kasturinarra I don't think this is expected (doesn't seem to be UX friendly). If the command is just creating workspace, then it should not be printing out that it already exists. Can you create an issue with steps to reproduce the scenario, so that its easier to debug?

kasturinarra:
@varshaprasad96 here is the bug https://github.com/kcp-dev/kcp/issues/2232, could you please help take a look when you get sometime ? thanks !!

PR #2111, opened 2022-09-30
:seedling: Deprecate APIResourceImport.Spec.Location

markmc:
This field is used by the syncer and various controllers to identify which SyncTarget owns the import. However, we can do this using the SyncTarget owner reference.
    
Removing this field would decouple the `apiresource` API group from the `workload` API group - i.e. this is currently the only place in the `apiresource` types where we refer to a `workload` concept.
    
Removing it would also remove a point of SyncTarget vs Location vs Cluster confusion.

The syncer and synctargetexports controller uses the Location field to identify the SyncTarget which owns the import. We can use the SyncTarget owner reference instead.

The heartbeat controller is currently using it to trigger reconciliation of a SyncTarget when there are updates to its imports, but that doesn't seem to be necessary at all.


markmc:
Will reopen if I ever rebase

Issue #1964, opened 2022-09-13
bug: user loses access to workspace being deleted if content has finalizers remaining


ncdc:
### Describe the bug

When deleting a workspace, if any of the content in the workspace, its namespaces, its subworkspaces, etc. has finalizers, it may take time for all the content to be deleted. Or sometimes the finalizers might never be removed (for whatever reason).... this is all fine, except our workspace deletion logic deletes the owner ClusterRole and ClusterRoleBinding as soon as the delete-workspace requests succeeds (which usually happens, as it just sets the deletion timestamp).

As a result, the user will be able to see their workspace is terminating/still being deleted, but they won't be able to access any content in the workspace to see why it's "stuck".

### Steps To Reproduce

1. Go to your home workspace
2. Create a child workspace
3. In the child workspace, add a finalizer to a resource
4. Back in the home workspace, delete the child workspace
5. See the workspace is being deleted
6. Try to use the child workspace and then do anything, e.g. `kubectl api-resources`
7. You won't see any APIs

### Expected Behaviour

User still has access to child workspace

hasheddan:
@ncdc I'd be happy to pick this one up

ncdc:
@hasheddan go for it! Happy to discuss first if you'd like for how to approach (I'll admit I haven't thought about it yet).

/assign @hasheddan

hasheddan:
Thanks @ncdc! It looked like there was a relatively lightweight option for providing a solution here, so I went ahead and opened https://github.com/kcp-dev/kcp/pull/2113 which I am happy to discuss on!

PR #2113, opened 2022-10-02
🐛 Only delete ClusterRoles and ClusterRoleBindings when Workspace finalizer is removed

hasheddan:
## Summary

Drops the cleanup of ClusterRoles and ClusterRoleBindings from the
projected Workspace deletion handler. Because Workspace is typically
deleted immediately, but the underlying ClusterWorkspace may take longer
to clean up, or may never get cleaned up, we want to continue to allow
users to access the terminating workspace to observe resources within
it.

Updates the ClusterWorkspace deletion controller to delete ClusterRoles
and ClusterRoleBindings associated with a Workspace as a final step
before removing the Workspace finalizer. This ensures that users are
able to access the contents of a Workspace during termination and only
have access removed when the Workspace is actually removed.

I was able to reproduce the behavior described in https://github.com/kcp-dev/kcp/issues/1964 prior to the patch set in this PR:

```
🤖 (kcp) k ws use '~'
Current workspace is "root:users:py:ov:user".
🤖 (kcp) k ws tree
.
└── user

 🤖 (kcp) k ws create child
Workspace "child" (type root:universal) created. Waiting for it to be ready...
Workspace "child" (type root:universal) is ready to use.
🤖 (kcp) k get ws
NAME    TYPE        PHASE   URL
child   universal   Ready   https://192.168.1.189:6443/clusters/root:users:py:ov:user:child
🤖 (kcp) go run ./cmd/kubectl-workspace/ tree
.
└── user
    └── child

🤖 (kcp) k ws use child
Current workspace is "root:users:py:ov:user:child" (type "root:universal").
🤖 (kcp) k create ns cool-ns
namespace/cool-ns created
🤖 (kcp) k edit ns cool-ns
namespace/cool-ns edited
🤖 (kcp) k get ns
NAME      STATUS   AGE
cool-ns   Active   62s
default   Active   91s
🤖 (kcp) k ws use '~'
Current workspace is "root:users:py:ov:user".
🤖 (kcp) k get ws
NAME    TYPE        PHASE   URL
child   universal   Ready   https://192.168.1.189:6443/clusters/root:users:py:ov:user:child
🤖 (kcp) k delete ns child
Error from server (NotFound): namespaces "child" not found
🤖 (kcp) k delete ws child
workspace.tenancy.kcp.dev "child" deleted
🤖 (kcp) k get ws
NAME    TYPE        PHASE      URL
child   universal   Deleting   https://192.168.1.189:6443/clusters/root:users:py:ov:user:child
🤖 (kcp) k ws use child
error checking APIBindings: apibindings.apis.kcp.dev is forbidden: User "user" cannot list resource "apibindings" in API group "apis.kcp.dev" at the cluster scope: workspace access not permittedCurrent workspace is "root:users:py:ov:user:child" (type "root:universal").
```

After applying patches, behavior matches what is expected:

```
🤖 (kcp) k ws use '~'
Current workspace is "root:users:py:ov:user".
🤖 (kcp) k ws tree
.
└── user

🤖 (kcp) k ws create child
Workspace "child" (type root:universal) created. Waiting for it to be ready...
Workspace "child" (type root:universal) is ready to use.
🤖 (kcp) k ws use child
Current workspace is "root:users:py:ov:user:child" (type "root:universal").
🤖 (kcp) k get ns
NAME      STATUS   AGE
default   Active   26s
🤖 (kcp) k create ns cool-ns
namespace/cool-ns created
🤖 (kcp) k edit ns cool-ns
namespace/cool-ns edited
🤖 (kcp) k get ns
NAME      STATUS   AGE
cool-ns   Active   28s
default   Active   60s
🤖 (kcp) k ws use '~'
Current workspace is "root:users:py:ov:user".
🤖 (kcp) k get ws
NAME    TYPE        PHASE   URL
child   universal   Ready   https://192.168.1.189:6443/clusters/root:users:py:ov:user:child
🤖 (kcp) k delete ws child
workspace.tenancy.kcp.dev "child" deleted
🤖 (kcp) k get ws
NAME    TYPE        PHASE      URL
child   universal   Deleting   https://192.168.1.189:6443/clusters/root:users:py:ov:user:child
🤖 (kcp) k ws use child
Current workspace is "root:users:py:ov:user:child" (type "root:universal").
🤖 (kcp) k get ns
NAME      STATUS        AGE
cool-ns   Terminating   56s
```

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/1964


pkg/apis/tenancy/v1alpha1/helper/helper.go
```diff
@@ -26,6 +26,13 @@ import (
 	"github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
 )
 
+const (
+	WorkspaceNameLabel = "workspaces.kcp.dev/name"
+)
```
hasheddan, line 31: This feels significant enough to me to live in the `v1beta1` package, perhaps alongside the type definitions, but it appears there is a pattern of defining some helpers here so I opted to drop it in here for the time-being. More than happy to move it as desired.

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
```diff
@@ -248,9 +259,19 @@ func (c *Controller) finalizeWorkspace(ctx context.Context, workspace *tenancyv1
...
 
+			lcn := logicalcluster.From(workspace)
+			listOpts := metav1.ListOptions{
+				LabelSelector: helper.WorkspaceLabelSelector(workspace.Name),
+			}
+			if err := c.kubeClusterClient.Cluster(lcn).RbacV1().ClusterRoles().DeleteCollection(ctx, backgroudDeletion, listOpts); err != nil && !kerrors.IsNotFound(err) {
+				return fmt.Errorf("could not delete clusterroles: %w", err)
+			}
+			if err := c.kubeClusterClient.Cluster(lcn).RbacV1().ClusterRoleBindings().DeleteCollection(ctx, backgroudDeletion, listOpts); err != nil && !kerrors.IsNotFound(err) {
+				return fmt.Errorf("could not delete clusterrolebindings: %w", err)
```
hasheddan, line 270: Notable difference from the Workspace delete handler is we actually return errors now if we fail to delete and it is not due to CRs / CRBs not being found. This makes more sense in the context of a controller than in a typical REST handler IMO.

pkg/apis/tenancy/v1alpha1/helper/helper.go
```diff
@@ -26,6 +26,13 @@ import (
 	"github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
 )
 
+const (
+	WorkspaceNameLabel = "workspaces.kcp.dev/name"
```
sttts, line 30: would even go a step further: move it to types.go. Labels without the "internal" prefix are API.

pkg/virtual/workspaces/registry/rest_test.go
```diff
@@ -1139,14 +1140,6 @@ func TestDeleteWorkspace(t *testing.T) {
 			assert.NoError(t, err)
 			assert.Nil(t, response)
 			assert.False(t, deletedNow)
-			crbList, err := kubeClient.Tracker().List(rbacv1.SchemeGroupVersion.WithResource("clusterrolebindings"), rbacv1.SchemeGroupVersion.WithKind("ClusterRoleBinding"), "")
```
davidfestal, line 1142: Shouldn't we here test that the bindings are still there, instead of simply removing the tests ?

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
```diff
@@ -50,7 +53,13 @@ const (
 	controllerName = "kcp-clusterworkspacedeletion"
 )
 
+var (
+	background        = metav1.DeletePropagationBackground
+	backgroudDeletion = metav1.DeleteOptions{PropagationPolicy: &background}
+)
+
 func NewController(
+	kubeClusterClient kubernetesclient.ClusterInterface,
```
davidfestal, line 62: Shouldn't this be an `kubernetesclient.Interface` instead of the `kubernetesclient.ClusterInterface` , as it is done for the `kcpClusterClient` ?

pkg/server/controllers.go
```diff
@@ -312,8 +312,12 @@ func (s *Server) installWorkspaceDeletionController(ctx context.Context, config
 		}
 		return discoveryClient.ServerPreferredResources()
 	}
-
+	kubeClusterClient, err := kubernetesclient.NewClusterForConfig(config)
```
davidfestal, line 315:
```suggestion
	kubeClusterClient, err := kubernetesclient.NewConfig(config)
```

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
```diff
@@ -248,9 +259,23 @@ func (c *Controller) finalizeWorkspace(ctx context.Context, workspace *tenancyv1
...
 
+			lcn := logicalcluster.From(workspace)
+			listOpts := metav1.ListOptions{
+				LabelSelector: helper.WorkspaceLabelSelector(workspace.Name),
+			}
+
+			// TODO(hasheddan): ClusterRole and ClusterRoleBinding cleanup
+			// should be handled by garbage collection when the controller is
+			// implemented.
+			if err := c.kubeClusterClient.Cluster(lcn).RbacV1().ClusterRoles().DeleteCollection(ctx, backgroudDeletion, listOpts); err != nil && !kerrors.IsNotFound(err) {
```
davidfestal, line 270:
```suggestion
			if err := c.kubeClusterClient.RbacV1().ClusterRoles().DeleteCollection(logicalcluster.WithCluster(ctx, lcn), backgroudDeletion, listOpts); err != nil && !kerrors.IsNotFound(err) {
```

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
```diff
@@ -248,9 +259,23 @@ func (c *Controller) finalizeWorkspace(ctx context.Context, workspace *tenancyv1
...
+				return fmt.Errorf("could not delete clusterroles: %w", err)
+			}
+			if err := c.kubeClusterClient.Cluster(lcn).RbacV1().ClusterRoleBindings().DeleteCollection(ctx, backgroudDeletion, listOpts); err != nil && !kerrors.IsNotFound(err) {
```
davidfestal, line 273:
```suggestion
			if err := c.kubeClusterClient.RbacV1().ClusterRoleBindings().DeleteCollection(logicalcluster.WithCluster(ctx, lcn), backgroudDeletion, listOpts); err != nil && !kerrors.IsNotFound(err) {
```

pkg/virtual/workspaces/registry/rest_test.go
hasheddan: @davidfestal good point, will update 👍🏻

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
hasheddan: @davidfestal happy to update to `kubernetesclient.Interface` here -- just to make sure I understand why we would choose to do so: we are fine to use the `kubernetesclient.Interface` here rather than `kubernetesclient.ClusterInterface` because when we enqueue we are using a "cluster-aware" key (`key, err := kcpcache.MetaClusterNamespaceKeyFunc(obj)`), so specifying `.Cluster(lcn)` when performing operations in this controller would be redundant.

Does that sound right? Please feel free to correct anything I have incorrect. If folks would find it useful, I'd love to also add this convention to the developer docs.

stevekuznetsov: There are two cases: 

1. you are operating in a single-cluster context, as in a controller against a compute cluster, you use `cache.MetaNamespaceKeyFunc`, you have a `kubernetes.Interface`.
2. you are operating against kcp in a multi-cluster context, you use `kcpcache.MetaClusterNamespaceKeyFunc`, you have a `kubernetes.ClusterInterface`

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
stevekuznetsov: no

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
stevekuznetsov: This is the second case, so let's keep the implementation the way it is.

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
davidfestal: Can you explain how this case is distinct from the existing use of the `kcpClusterClient` ? as shown in line https://github.com/kcp-dev/kcp/pull/2113/files/c1d4b8380218e20ec36ad913ffdcdaa153acb200#diff-13a6c58d33e07a3a3d8bc6d2fdeda5f798da6150f69dde58f7a51b455fc4bc18R277 ?

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
davidfestal: I might be missing some context here, but I don't see how it is a different use-case than the use of the `kcpClientCluster`. Could explain it ?

pkg/apis/tenancy/v1beta1/types.go
```diff
@@ -23,6 +23,8 @@ import (
 	conditionsv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/apis/conditions/v1alpha1"
 )
 
+const WorkspaceNameLabel = "workspaces.kcp.dev/name"
```
ncdc, line 26: Should probably have godoc?

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
```diff
@@ -248,9 +258,23 @@ func (c *Controller) finalizeWorkspace(ctx context.Context, workspace *tenancyv1
 		if workspace.Finalizers[i] == deletion.WorkspaceFinalizer {
 			workspace.Finalizers = append(workspace.Finalizers[:i], workspace.Finalizers[i+1:]...)
 
+			lcn := logicalcluster.From(workspace)
```
ncdc, line 261: Note: we typically name this type of variable `clusterName` (although we're not 100% consistent in the codebase)

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
```diff
@@ -248,9 +258,23 @@ func (c *Controller) finalizeWorkspace(ctx context.Context, workspace *tenancyv1
...
+			listOpts := metav1.ListOptions{
+				LabelSelector: helper.WorkspaceLabelSelector(workspace.Name),
+			}
+
+			// TODO(hasheddan): ClusterRole and ClusterRoleBinding cleanup
+			// should be handled by garbage collection when the controller is
+			// implemented.
+			if err := c.kubeClusterClient.Cluster(lcn).RbacV1().ClusterRoles().DeleteCollection(ctx, backgroudDeletion, listOpts); err != nil && !apierrors.IsNotFound(err) {
+				return fmt.Errorf("could not delete clusterroles: %w", err)
```
ncdc, line 270: Probably should include the clusterName in the error message

pkg/reconciler/tenancy/clusterworkspacedeletion/clusterworkspace_deletion_controller.go
```diff
@@ -248,9 +258,23 @@ func (c *Controller) finalizeWorkspace(ctx context.Context, workspace *tenancyv1
...
+			}
+			if err := c.kubeClusterClient.Cluster(lcn).RbacV1().ClusterRoleBindings().DeleteCollection(ctx, backgroudDeletion, listOpts); err != nil && !apierrors.IsNotFound(err) {
+				return fmt.Errorf("could not delete clusterrolebindings: %w", err)
```
ncdc, line 273: Probably should include the clusterName in the error message

hasheddan (commented):
@ncdc comments addressed with one note / question below 👍🏻

ncdc (approved):

Issue #1214, opened 2022-06-05
Difficult to understand error message when trying to create child workspace named "root"


MikeSpreitzer:
**Describe the bug**
I tried to make a little mischief.  I was rebuffed, but the error message will not be understandable to someone who is not familiar with the code.

```
(base) mspreitz@mjs12 kcp % kubectl kcp ws default
Current workspace is "root:default" (type "Organization").

(base) mspreitz@mjs12 kcp % kubectl kcp ws create root
error: ClusterWorkspace.tenancy.kcp.dev "root" is invalid: <nil>: Invalid value: "": "metadata.name" must not validate the schema (not)

(base) mspreitz@mjs12 kcp % kubectl kcp ws create rooty
Workspace "rooty" (type "Universal") created. Waiting for being ready.
```

That last command makes it clear that the string appearing last is not an absolute pathname but just the last step.

**To Reproduce**
Shown above.

**Expected behavior**
When the particular step name is bad, the error message should clearly say that.

**Additional context**

sttts:
That's an OpenAPI error message. We could move it from the OpenAPI pattern to admission to produce a nicer message.

PR #2118, opened 2022-10-03
✨ Add reservednames admission plugin to surface clearer error messages

hasheddan:
## Summary

Adds a mutating admission plugin for ClusterWorkspaces and
ClusterWorkspaceTypes to ensure that a more helfpul error message is
returned if a user attempts to create with a reserved name. A mutating
webhook is used due to the fact that we want to catch this prior to
OpenAPI schema validation.

Before
```
🤖 (kcp) k kcp ws create root
error: ClusterWorkspace.tenancy.kcp.dev "root" is invalid: <nil>: Invalid value: "": "metadata.name" must not validate the schema (not)
exit status 1
```
After
```
🤖 (kcp) k kcp ws create root
error: clusterworkspaces.tenancy.kcp.dev "root" is forbidden: cannot use reserved name "root" as "metadata.name"
exit status 1
```

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/1214

### DIFF
--- a/pkg/admission/reservednames/admission.go
+++ b/pkg/admission/reservednames/admission.go
@@ -1,0 +1,107 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reservednames
+
+import (
+	"context"
+	"io"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apiserver/pkg/admission"
+
+	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
+)
+
+// PluginName is the name used to identify this admission webhook.
+const PluginName = "apis.kcp.dev/ReservedNames"
+
+// Register registers the reserved name admission webhook.
+func Register(plugins *admission.Plugins) {
+	plugins.Register(PluginName,
+		func(_ io.Reader) (admission.Interface, error) {
+			return NewReservedNames(), nil
+		})
+}
+
+// A reservedNameFn determines whether the name of the object is reserved.
+type reservedNameFn func(resource schema.GroupResource, kind schema.GroupKind, name string) bool
+
+// newReservedNameFunc builds a reserved name check for the given resource and
+// kind.
+func newReservedNameFn(resource schema.GroupResource, kind schema.GroupKind, reserved ...string) reservedNameFn {
+	res := make(map[string]struct{}, len(reserved))
+	for _, r := range reserved {
+		res[r] = struct{}{}
+	}
+	check := func(name string) bool {
+		_, reserved := res[name]
+		return reserved
+	}
+	return func(objResource schema.GroupResource, objKind schema.GroupKind, objName string) bool {
+		if objResource != resource {
+			return false
+		}
+		if objKind != kind {
+			return false
+		}
+		return check(objName)
+	}
+}
+
+// ReservedNames is an admission plugin for checking reserved object names.
+type ReservedNames struct {
+	*admission.Handler
+
+	reservedNameFns []reservedNameFn
+}
+
+// NewReservedNames constructs a new ReservedNames admission plugin.
+func NewReservedNames() *ReservedNames {
+	return &ReservedNames{
+		Handler: admission.NewHandler(admission.Create),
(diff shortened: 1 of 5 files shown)


pkg/apis/tenancy/v1alpha1/helper/helper.go
```diff
@@ -40,3 +40,12 @@ func QualifiedObjectName(obj metav1.Object) string {
...
 	return fmt.Sprintf("%s|%s", logicalcluster.From(obj), obj.GetName())
 }
+
+// IsReservedName indicates that the provided name is reserved and cannot be
+// used in a user-supplied object.
+func IsReservedName(name string) bool {
+	if name == "root" || name == "system" {
+		return true
+	}
+	return false
```
ncdc, line 50: simplification suggestion: `return name == "root" || name == "system"`

ncdc: Also: reserved clusterworkspace names are `root` and `system`. Reserved clusterworkspacetype names are `system` and `any`. (see the 2 `.yaml-patch` files in `config/crds` for the respective crds).

pkg/admission/reservednames/admission_test.go
```diff
@@ -0,0 +1,105 @@
...
+		},
+		"NotApplicableKind": {
+			attr: createAttr("root", &tenancyv1alpha1.ClusterWorkspace{}, "NotaCWorCWT", "clusterworkspaces"),
+		},
+	}
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			plugin := &reservedNames{
+				Handler: admission.NewHandler(admission.Create, admission.Update),
+				applicableKinds: map[schema.GroupKind]struct{}{
```
ncdc, line 89: Maybe we should make a constructor for the `reservedNames` instance and call that here?

hasheddan (commented):
@ncdc I did a bit of a refactor here to try to make the checks a bit cleaner. Also added some `TODOs` for generating the patches we apply to the OpenAPI schema from the reserved name definitions in code. Happy to open an issue tracking that work if you're on board.

pkg/admission/reservednames/admission.go
```diff
@@ -0,0 +1,106 @@
...
+// Ensure that the required admission interfaces are implemented.
+// NOTE(hasheddan): we must use mutation rather than validation because OpenAPI
+// validation will occur prior to validation webhook.
+var _ = admission.MutationInterface(&ReservedNames{})
+
+// Admit ensures that the object does not violate reserved name constraints.
+func (o *ReservedNames) Admit(ctx context.Context, a admission.Attributes, _ admission.ObjectInterfaces) (err error) {
+	resource, kind := a.GetResource().GroupResource(), a.GetKind().GroupKind()
+	for _, fn := range o.reservedNameFns {
+		if fn(resource, kind, a.GetName()) {
```
ncdc, line 100: Maybe `if isReservedName()` would be more indicative of what it's doing?

pkg/admission/reservednames/admission_test.go
```diff
@@ -0,0 +1,109 @@
...
+			attr: createAttr("root", &tenancyv1alpha1.ClusterWorkspace{}, "ClusterWorkspace", "notacworcwt"),
+		},
+		"NotApplicableKind": {
+			attr: createAttr("root", &tenancyv1alpha1.ClusterWorkspace{}, "NotaCWorCWT", "clusterworkspaces"),
+		},
+	}
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			plugin := NewReservedNames(
+				newReservedNameFn(
```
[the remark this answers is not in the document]
hasheddan: I could go either way -- I think I may be optimizing a bit too heavily for future use cases of reserved names, whereas we really only have these two APIs where it applies right now. I'll switch to hard-coded 👍🏻

pkg/admission/reservednames/admission.go
hasheddan: will do 👍🏻

pkg/admission/reservednames/admission_test.go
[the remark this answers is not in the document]
hasheddan: done 👍🏻

pkg/admission/reservednames/admission.go
```diff
@@ -0,0 +1,107 @@
...
+
+// ReservedNames is an admission plugin for checking reserved object names.
+type ReservedNames struct {
+	*admission.Handler
+
+	reservedNameFns []reservedNameFn
+}
+
+// NewReservedNames constructs a new ReservedNames admission plugin.
+func NewReservedNames(fns ...reservedNameFn) *ReservedNames {
```
ncdc, line 74:
```suggestion
func NewReservedNames() *ReservedNames {
```

pkg/admission/reservednames/admission.go
hasheddan: Went with `isReserved` as name is being passed

pkg/admission/reservednames/admission.go
hasheddan: whoops -- sorry about that, moving a little too fast

pkg/admission/reservednames/admission_test.go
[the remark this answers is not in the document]
ncdc: Remove from here?

ncdc (approved):

PR #2121, opened 2022-10-03
:bug: Detect kubectl plugin via FlagSet Name

### LINKED ISSUE #2119: bug: wrong variable substitution in `kubectl kcp workspace` example
### Describe the bug

The `cliName` is defined as per this logic

https://github.com/kcp-dev/kcp/blob/32fac8a42a32616bc56e653af15b47de022f2165/pkg/cliplugins/workspace/cmd/cmd.go#L72-L75

and is substituted here

https://github.com/kcp-dev/kcp/blob/32fac8a42a32616bc56e653af15b47de022f2165/pkg/cliplugins/workspace/cmd/cmd.go#L81

the doc-generator has no way of deciding what is the path/argument. So it generates the example by substituting `kubectl` instead of `kubectl kcp`. See below

image
 

### Steps To Reproduce

1. Go to `kcp` project root with latest commits
2. Run `make generate-docs`
3. Verify the bug by opening `docs/cli/kcp_workspace.md`

### Expected Behaviour

The example should be generated by substituting `kubectl kcp` instead of `kubectl`.

### Additional Context

This problem was not identified during the testing of PR #2000.

hardys:
## Summary

The current approach uses the binary name, which won't work in the case where code gets run via [hack/generate/cli-doc/gen-cli-doc.go](https://github.com/kcp-dev/kcp/blob/main/hack/generate/cli-doc/gen-cli-doc.go)

Fixes #2119

### DIFF
--- a/cmd/kubectl-kcp/main.go
+++ b/cmd/kubectl-kcp/main.go
@@ -20,10 +20,15 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/spf13/pflag"
+
 	"github.com/kcp-dev/kcp/cmd/kubectl-kcp/cmd"
 )
 
 func main() {
+	flags := pflag.NewFlagSet("kubectl-kcp", pflag.ExitOnError)
+	pflag.CommandLine = flags
+
 	cmd := cmd.KubectlKcpCommand()
 	if err := cmd.Execute(); err != nil {
 		fmt.Fprintf(os.Stderr, "error: %v\n", err)
--- a/cmd/kubectl-workspace/main.go
+++ b/cmd/kubectl-workspace/main.go
@@ -30,7 +30,7 @@ import (
 )
 
 func main() {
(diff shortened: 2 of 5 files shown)


hardys:
/retitle :bug: Detect kubectl plugin via FlagSet Name

ncdc:
Seems reasonable. Any downside?

hardys:
> Seems reasonable. Any downside?

Not that I can think of - this approach means the usage is coupled to the code independent of the binary naming, which is probably preferable if we expect to have usage like [gen-cli-doc.go](https://github.com/kcp-dev/kcp/blob/main/hack/generate/cli-doc/gen-cli-doc.go) outside of the plugin binaries (I wasn't aware of this usage at the time of #1901)

ncdc (approved):

ncdc:
Cool, thanks!

avinal (approved):
Tested, working as expected.

PR #2126, opened 2022-10-04
:bug:  Fix bugs in the resource controller

davidfestal:

## Summary

This PR fixes several bugs in the resource controller:
- We should not add the `RequestState` label to [Sync](url) for a `SyncTarget` when the namespace already has the deletion annotation for this `SyncTarget`. It doesn't make sense to start syncing a resource to a SyncTarget while the intent already expressed on the namespace is to delete all resources from this `SyncTarget`
- Additional small fixes in the resource `computePlacement` function.

See the individual commits for more information
 
## Related issue(s)

Bugs spotted while debugging the Syncer Virtual Transformations e2e test

### DIFF
--- a/pkg/reconciler/workload/resource/resource_controller.go
+++ b/pkg/reconciler/workload/resource/resource_controller.go
@@ -448,7 +448,7 @@ func locations(annotations, labels map[string]string, skipPending bool) (locatio
 	}
 	for k := range annotations {
 		if strings.HasPrefix(k, workloadv1alpha1.InternalClusterDeletionTimestampAnnotationPrefix) {
-			deleting.Insert(strings.TrimPrefix(k, workloadv1alpha1.ClusterResourceStateLabelPrefix))
+			deleting.Insert(strings.TrimPrefix(k, workloadv1alpha1.InternalClusterDeletionTimestampAnnotationPrefix))
 		}
 	}
 	return
--- a/pkg/reconciler/workload/resource/resource_reconcile.go
+++ b/pkg/reconciler/workload/resource/resource_reconcile.go
@@ -85,7 +85,7 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName logical
 	// clean finalizers from removed syncers
 	filteredFinalizers := make([]string, 0, len(obj.GetFinalizers()))
 	for _, f := range obj.GetFinalizers() {
-		logger = logger.WithValues("finalizer", f)
+		logger := logger.WithValues("finalizer", f)
 		if !strings.HasPrefix(f, syncershared.SyncerFinalizerNamePrefix) {
 			filteredFinalizers = append(filteredFinalizers, f)
 			continue
--- a/pkg/reconciler/workload/resource/resource_reconcile.go
+++ b/pkg/reconciler/workload/resource/resource_reconcile.go
@@ -176,8 +176,10 @@ func computePlacement(ns *corev1.Namespace, obj metav1.Object) (annotationPatch 
 	// create merge patch
 	annotationPatch = map[string]interface{}{}
 	labelPatch = map[string]interface{}{}
+
+	// unschedule objects on locations where the namespace is not scheduled
 	for _, loc := range objLocations.Difference(nsLocations).List() {
-		// location was removed from namespace, but is still on the object
+		// That's an inconsistent state, probably due to the namespace deletion reaching its grace period => let's repair it
 		var hasSyncerFinalizer, hasClusterFinalizer bool
 		// Check if there's still the syncer or the cluster finalizer.
 		for _, finalizer := range obj.GetFinalizers() {
@@ -199,7 +201,9 @@ func computePlacement(ns *corev1.Namespace, obj metav1.Object) (annotationPatch 
 			}
 		}
 	}
-	for _, loc := range nsLocations.Intersection(nsLocations).List() {
+
+	// sync deletion timestamps if both namespace and object are scheduled
+	for _, loc := range nsLocations.Intersection(objLocations).List() {
 		if nsTimestamp, found := ns.Annotations[workloadv1alpha1.InternalClusterDeletionTimestampAnnotationPrefix+loc]; found && validRFC3339(nsTimestamp) {
 			objTimestamp, found := obj.GetAnnotations()[workloadv1alpha1.InternalClusterDeletionTimestampAnnotationPrefix+loc]
 			if !found || !validRFC3339(objTimestamp) {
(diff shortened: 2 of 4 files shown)


sttts:
Needs better commit message.

Are these two changes or one?

pkg/reconciler/workload/resource/resource_controller.go
```diff
@@ -448,7 +448,7 @@ func locations(annotations, labels map[string]string, skipPending bool) (locatio
 	}
 	for k := range annotations {
 		if strings.HasPrefix(k, workloadv1alpha1.InternalClusterDeletionTimestampAnnotationPrefix) {
-			deleting.Insert(strings.TrimPrefix(k, workloadv1alpha1.ClusterResourceStateLabelPrefix))
+			deleting.Insert(strings.TrimPrefix(k, workloadv1alpha1.InternalClusterDeletionTimestampAnnotationPrefix))
```
stevekuznetsov, line 461: Do we have a unit test for this fix?

davidfestal: Yes, indirectly. The additional test added in commit 0bcf6f345ecf70f64ad1c28148c2d61b226cdcfd will fail without this fix.
So I didn't think useful to add a new dedicated test for the `locations` function

pkg/reconciler/workload/resource/resource_controller.go
```diff
@@ -159,6 +160,15 @@ func filterNamespace(obj interface{}) bool {
 		logging.WithReconciler(klog.Background(), controllerName).WithValues("namespace", name).V(2).Info("skipping syncing Namespace")
 		return false
 	}
+	// Filter out namespaces which have been synced to a SyncTarget living on a KCP shard (typical e2e test situation)
```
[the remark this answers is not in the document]
davidfestal: Even though this situation might happen most typically e2e tests, this is not limited to it.
It could happen in other cases as well.
Do we really want resources that are targets of a Syncing to become sources of another syncing ?
Do we want to enable this use-case ?
To me it seems better to disable this use-case until we clearly acknowledged its requirement, scoped its limits and explored its impacts.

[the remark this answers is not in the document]
sttts: How could we opt-in the resource controller? Does this controller even do anything if there is no placement?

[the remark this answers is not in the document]
sttts: Or is the problem that there is always a placement? Maybe we have to rethink our CWT structure, i.e. that the e2e does not create a fake pcluster that has workload APIs.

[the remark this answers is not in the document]
stevekuznetsov: @davidfestal I meant - if we have a (working) mechanism to opt in or out of syncing, let's use it?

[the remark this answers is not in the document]
davidfestal: Yes, it does. It reconcile all the namespaces and resources, and ensures that the placement on individual resources match the placement on namespaces. That's really useless work.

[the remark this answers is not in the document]
davidfestal: In addition to that, I've hit, while debugging, a case when such a resource has a `RequestState` label set to `Sync` while its  namespace doesn't. This leads to the resource reconciler patching the resource to suppress it.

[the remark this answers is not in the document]
davidfestal: I split the PR to remove this questionable commit, and created a dedicated PR for it: https://github.com/kcp-dev/kcp/pull/2131 to continue the discussion

pkg/reconciler/workload/resource/resource_reconcile.go
```diff
@@ -199,7 +200,7 @@ func computePlacement(ns *corev1.Namespace, obj metav1.Object) (annotationPatch
 			}
 		}
 	}
-	for _, loc := range nsLocations.Intersection(nsLocations).List() {
+	for _, loc := range nsLocations.Intersection(objLocations).List() {
```
sttts, line 203: a one-liner go doc would help to understand quickly what these loops are doing

pkg/reconciler/workload/resource/resource_reconcile.go
```diff
@@ -208,7 +209,12 @@ func computePlacement(ns *corev1.Namespace, obj metav1.Object) (annotationPatch
 		}
 	}
 	for _, loc := range nsLocations.Difference(objLocations).List() {
-		// location was missing on the object
+		// location is missing on the object
```
sttts, line 212: put it above the loop. What is this loop doing?

pkg/reconciler/workload/resource/resource_reconcile.go
sttts: `// sync deletion timestamps if both namespace and object are scheduled

pkg/reconciler/workload/resource/resource_reconcile.go
sttts: `// set label on unscheduled objects, namespace is not deleting

sttts:
Some nits about better go docs. Don't document the obvious. Document more high level what the loops do instead of every line.

/lgtm
/approve

pkg/reconciler/workload/resource/resource_reconcile.go
```diff
@@ -199,16 +201,22 @@ func computePlacement(ns *corev1.Namespace, obj metav1.Object) (annotationPatch
...
 				annotationPatch[workloadv1alpha1.InternalClusterDeletionTimestampAnnotationPrefix+loc] = nsTimestamp
 			}
 		}
 	}
+	// set label on unscheduled objects if namespace is scheduled and not deleting
 	for _, loc := range nsLocations.Difference(objLocations).List() {
-		// location was missing on the object
+		if nsDeleting.Has(loc) {
+			// if the namespace is removing resources from this location
+			// then do not add the location label on the object
```
sttts, line 217: nit: would claim this comment is not adding anything

pkg/reconciler/workload/resource/resource_reconcile.go
sttts: super-nit: put empty line above. Also for the other loops.

pkg/reconciler/workload/resource/resource_reconcile.go
```diff
@@ -176,8 +176,10 @@ func computePlacement(ns *corev1.Namespace, obj metav1.Object) (annotationPatch
 	// create merge patch
 	annotationPatch = map[string]interface{}{}
 	labelPatch = map[string]interface{}{}
+	// unschedule objects on locations where the namespace is not scheduled
 	for _, loc := range objLocations.Difference(nsLocations).List() {
-		// location was removed from namespace, but is still on the object
+		// Location was removed from namespace, but is still on the object
```
sttts, line 181: this line does not add anything.

kasturinarra:
@davidfestal is there any thing for QE to validate here ? Any user facing change or functionality impact that we need to look at ? thanks !!

PR #2129, opened 2022-10-04
🐛 Syncer AdvancedScheduling fixes: Avoid syncing down the experimental status annotation.

jmprusi:
## Summary

When the Syncer is running in AdvancedScheduling mode, the experimental status annotation is synced to the downstream resource, this PR removes that annotation when syncing the resource downstream.

### DIFF
--- a/pkg/syncer/spec/spec_process.go
+++ b/pkg/syncer/spec/spec_process.go
@@ -342,6 +342,8 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
 	// Strip cluster name annotation
 	downstreamAnnotations := downstreamObj.GetAnnotations()
 	delete(downstreamAnnotations, logicalcluster.AnnotationKey)
+	//TODO(jmprusi): To be removed when switching to the syncer Virtual Workspace transformations.
+	delete(downstreamAnnotations, workloadv1alpha1.InternalClusterStatusAnnotationPrefix+c.syncTargetKey)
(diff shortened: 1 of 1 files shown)


pkg/syncer/spec/spec_process.go
```diff
@@ -333,6 +333,7 @@ func (c *Controller) applyToDownstream(ctx context.Context, gvr schema.GroupVers
 	// Strip cluster name annotation
 	downstreamAnnotations := downstreamObj.GetAnnotations()
 	delete(downstreamAnnotations, logicalcluster.AnnotationKey)
+	delete(downstreamAnnotations, workloadv1alpha1.InternalClusterStatusAnnotationPrefix+c.syncTargetKey)
```
davidfestal, line 336: Could you add a TODO that it could be removed when switching to Syncer Virtual Workspace transformations ?

mikenairn:
@jmprusi I tested this yesterday after our conversation, and it does work better now, but i did observe a few times that the syncer could still miss the status update to the resource on the physical cluster, or not process it for whatever reason, meaning it still can end up being left empty. Does/should the syncer handle missed events at all? Updates are all current;y event based right?

davidfestal:
> Updates

Would be great if you could try to gather the typical use-case / flow when this happen.

pkg/syncer/spec/spec_process.go
jmprusi: done!

davidfestal (approved):

PR #2131, opened 2022-10-04
:bug: Don't schedule downstream resources

davidfestal:
## Summary

The resource controller used to schedule resources which in fact were already the result of syncing (when a SyncTarget cluster is created as a KCP workspace). This is a typical use-case for e2e tests, and these resources should not take part in the resource scheduling: including them in resource scheduling makes the e2e test execution quite confusing.

This is a split of PR https://github.com/kcp-dev/kcp/pull/2126.

[Discussion started in the initial PR](https://github.com/kcp-dev/kcp/pull/2126#discussion_r987082954) should be continued here before merging.

### DIFF
--- a/pkg/reconciler/workload/resource/resource_controller.go
+++ b/pkg/reconciler/workload/resource/resource_controller.go
@@ -28,6 +28,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


davidfestal:
Last comment from @sttts on IM channel:
> we must make the workload controllers so that they don't do anything in workspaces without placements, or without compute bindings. the e2e case is just a special case of that rule

davidfestal:
obsolete

PR #2134, opened 2022-10-05
:seedling: Syncer: Complete the move to structural logging

davidfestal:
## Summary

Complete the move to structural logging in the Syncer.
This has been started in a previous PR, but many places were not finished.

### DIFF
--- a/pkg/logging/tmc.go
+++ b/pkg/logging/tmc.go
@@ -1,0 +1,38 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package logging supplies common constants to ensure consistent use of structured logs.
+package logging
+
+const (
+	// SyncTargetKeyPrefix is the prefix used for all the keys related to a SyncTarget.
+	SyncTargetKeyPrefix = "syncTarget."
+
+	// SyncTargetWorkspaceKey is used to specify a workspace when a log is related to a SyncTarget.
+	SyncTargetWorkspaceKey = SyncTargetKeyPrefix + "workspace"
+	// SyncTargetNamespaceKey is used to specify a namespace when a log is related to a SyncTarget.
+	SyncTargetNamespaceKey = SyncTargetKeyPrefix + "namespace"
+	// SyncTargetNameKey is used to specify a name when a log is related to a SyncTarget.
+	SyncTargetNameKey = SyncTargetKeyPrefix + "name"
+
+	// DownstreamKeyPrefix is the prefix used for all the keys related to a downstream object.
+	DownstreamKeyPrefix = "downstream."
+
+	// DownstreamNamespaceKey is used to specify a namespace when a log is related to a downstream object.
+	DownstreamNamespaceKey = DownstreamKeyPrefix + "namespace"
+	// DownstreamNameKey is used to specify a name when a log is related to a downstream object.
+	DownstreamNameKey = DownstreamKeyPrefix + "name"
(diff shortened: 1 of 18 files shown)


pkg/syncer/namespace/namespace_downstream_controller.go
```diff
@@ -88,32 +89,34 @@ func NewDownstreamController(
...
-func (c *DownstreamController) AddToQueue(obj interface{}, logger logr.Logger) {
+func (c *DownstreamController) AddToQueue(obj interface{}) {
 	key, err := keyfunctions.DeletionHandlingMetaNamespaceKeyFunc(obj) // note: this is *not* a cluster-aware key
 	if err != nil {
 		utilruntime.HandleError(err)
 		return
 	}
 
-	logger.V(4).Info("queueing namespace", "key", key)
+	c.logger.V(2).Info("queueing namespace", logging.QueueKeyKey, key)
```
stevekuznetsov, line 119: Why not `logging.WithQueueKey()`?

pkg/syncer/namespace/namespace_downstream_controller.go
```diff
@@ -88,32 +89,34 @@ func NewDownstreamController(
 		syncTargetWorkspace: syncTargetWorkspace,
 		syncTargetUID:       syncTargetUID,
 		syncTargetKey:       syncTargetKey,
+
+		logger: logging.WithReconciler(syncerLogger, downstreamControllerName),
```
stevekuznetsov, line 93: I don't think we want to do this. The power of a logger per context is that you get to attach information *to a particular logger* and pass that around. For instance, multi-threaded use of this common logger would not allow the controller to add the queue key being processed, the object name, etc, to the specific instance of the logger used by a particular goroutine doing a particular reconciliation. If you're holding on to a "bare" logger and creating new ones from it in every top-level case, then I'd argue why bother with the top-level one at all? When a new dev is working inside of some particular sub-method, how would they know whether to pass one through the context, or subclass one from the shared state?

pkg/logging/tmc.go
```diff
@@ -0,0 +1,38 @@
...
+
+// Package logging supplies common constants to ensure consistent use of structured logs.
+package logging
+
+const (
+	// SyncTargetKeyPrefix is the prefix used for all the keys related to a SyncTarget.
+	SyncTargetKeyPrefix = "syncTarget."
+
+	// SyncTargetWorkspaceKey is used to specify a workspace when a log is related to a SyncTarget.
+	SyncTargetWorkspaceKey = SyncTargetKeyPrefix + "workspace"
```
stevekuznetsov, line 25: In which cases does a client want to add these to the logger, but cannot call `logging.WithObject(logger, syncTarget)`?

pkg/syncer/namespace/namespace_downstream_controller.go
davidfestal: I refactored the whole according to pattern you used in other places in the code, and loggers are not stored anymore in controller structs, but passed through context when possible.

pkg/syncer/namespace/namespace_downstream_controller.go
davidfestal: Fixed

pkg/logging/tmc.go
davidfestal: In all the places where a SyncTargetKey is required without having the Synctarget object at hand.
In addition, I only need to log the key, not the whole object (with apiVersion, etc ...)

stevekuznetsov: >  I only need to log the key, not the whole object (with apiVersion, etc ...)

That's what `WithObject()` does :)

pkg/server/bootstrap/identity.go
```diff
@@ -135,7 +135,7 @@ func wildcardIdentitiesResolver(ids *identities,
 			logger := logging.WithObject(logger, &apisv1alpha1.APIExport{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:        name,
-					Annotations: map[string]string{logging.APIVersionKey: tenancyv1alpha1.RootCluster.String()},
```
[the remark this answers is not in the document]
davidfestal: Oh, I don't know. I just added the `string()` around it when the key was of `Key` type. But now it should be removed.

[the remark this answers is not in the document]
davidfestal: I removed my change in fact. And let the previous code.

Issue #2051, opened 2022-09-22
bug: inconsistent state of API bindings having bound resources but no bound API export


s-urbaniak:
### Describe the bug

We observed a nil panic in https://github.com/kcp-dev/kcp/blob/cd9b93cb654695152b3d4c62ed5e225d35f3456c/pkg/admission/webhook/generic_webhook.go#L105

reported in https://github.com/kcp-dev/kcp/issues/2028.

Having a nil value for `apiBinding.Status.BoundAPIExport` is unexpected as it has bound resources being looped over: https://github.com/kcp-dev/kcp/blob/cd9b93cb654695152b3d4c62ed5e225d35f3456c/pkg/admission/webhook/generic_webhook.go#L98-L107

The current merged fix is to skip if `apiBinding.Status.BoundAPIExport` or `apiBinding.Status.BoundAPIExport.Workspace` is nil but has the danger bound resources missing admission, recognized in https://github.com/kcp-dev/kcp/pull/2031#discussion_r975180349.

As discussed in https://kubernetes.slack.com/archives/C021U8WSAFK/p1663669768445509?thread_ts=1663669357.447249&cid=C021U8WSAFK we have to clarify:

1. which claims apply in the different components (admission, controller, VW) when a rebinding is incomplete.
1. which admission webhooks apply on rebinding?
1. which admission webhooks apply on initial bind that does not complete? 
1. which maximal permission policies apply? same fear as in 3. that we don't authz correctly now.

### Steps To Reproduce

n/a, currently non-deterministic

### Expected Behaviour

correct state for bound resources

PR #2144, opened 2022-10-06
🌱 remove API export reference from API binding status, make API export reference in API binding immutable 

s-urbaniak:
## Summary
This is WIP, checking e2e tests.

I am adding more text to make e2e really happy, also this will fix some things, like API export immutability.
Let's see if this is enough to make the bots happy.

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/2051


pkg/admission/webhook/generic_webhook.go
```diff
@@ -96,14 +96,13 @@ func (p *WebhookDispatcher) getAPIBindingWorkspace(attr admission.Attributes, cl
 	for _, obj := range objs {
 		apiBinding := obj.(*apisv1alpha1.APIBinding)
 		for _, br := range apiBinding.Status.BoundResources {
-			if apiBinding.Status.BoundAPIExport != nil && apiBinding.Status.BoundAPIExport.Workspace == nil {
+			if apiBinding.Spec.Reference.Workspace == nil {
```
sttts, line 99: // this can never happen by OpenAPI validation

test/e2e/apibinding/apibinding_test.go
```diff
@@ -51,6 +54,62 @@ import (
...
+
+	apiBinding, err = kcpClusterClient.ApisV1alpha1().APIBindings().Get(logicalcluster.WithCluster(ctx, consumerWorkspace), apiBinding.Name, metav1.GetOptions{})
+	require.NoError(t, err)
+
+	patchedBinding := apiBinding.DeepCopy()
+	patchedBinding.Spec.Reference.Workspace.ExportName = "other-export"
+	mergePatch, err := jsonpatch.CreateMergePatch(encodeJSON(t, apiBinding), encodeJSON(t, patchedBinding))
+	require.NoError(t, err)
+	_, err = kcpClusterClient.ApisV1alpha1().APIBindings().Patch(logicalcluster.WithCluster(ctx, consumerWorkspace), apiBinding.Name, types.MergePatchType, mergePatch, metav1.PatchOptions{})
+	require.ErrorContains(t, err, "APIExport reference must not be changed")
```
s-urbaniak, line 110: do we rather want this as a unit test? seems a bit heavy-weight here. I saw some helper to use the CEL validator in tests.

pkg/apis/apis/v1alpha1/types_apibinding.go
```diff
@@ -69,6 +69,7 @@ type APIBindingSpec struct {
 	//
 	// +required
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="APIExport reference must not be changed"
```
s-urbaniak, line 72: Do you know if I can I assume that property changes traverse up? It's what I observe in the e2e test but I am not 100% confident.

pkg/reconciler/apis/apibinding/apibinding_reconcile_test.go
```diff
@@ -632,120 +622,6 @@ func TestReconcileBinding(t *testing.T) {
 	}
 }
 
-func TestReconcileBound(t *testing.T) {
```
ncdc, line 635: I don't think we want to remove this because there are still valid scenarios to test

test/e2e/apibinding/apibinding_test.go
ncdc: I'm fine to have it here

ncdc (changes requested):

pkg/apis/apis/v1alpha1/types_apibinding.go
s-urbaniak: i.e. this sets the validation rule on the top-level `reference` property only, not on `reference.path` and also not on `reference.exportName`. My question is: is it enough to have the validation rule on the top-level `reference.path` only?

The e2e tests indicate, yes, but I wasn't sure 100% :D

test/e2e/apibinding/apibinding_test.go
s-urbaniak: ack, i added a unit test as well 👍

pkg/apis/apis/v1alpha1/types_apibinding.go
ncdc: If e2e is working, 👍

pkg/reconciler/apis/apibinding/apibinding_reconcile_test.go
s-urbaniak: re-introduced but removed one test that asserted rebinding upon api export changes which is not possible any more.

pkg/admission/webhook/generic_webhook_test.go
```diff
@@ -115,9 +115,6 @@ func TestDispatch(t *testing.T) {
 				admission.Create,
 			),
 			cluster: "root:org:dest-cluster",
-			expectedHooks: []webhook.WebhookAccessor{
```
ncdc, line 118: Need to keep this. The proper "fix" is to correct the incorrect spec.reference.workspace.path value "foo" - it needs to be "root:org:source-cluster"

s-urbaniak: fixed

ncdc (approved):

PR #2149, opened 2022-10-06
:bug: fix 'make install' on mac

varshaprasad96:
When `make install` is run on Mac it gives the error:

```
ln -sfr /Users/vnarsing/go/bin/kubectl-workspace /Users/vnarsing/go/bin/kubectl-ws
ln: illegal option -- r
usage: ln [-s [-F] | -L | -P] [-f | -i] [-hnv] source_file [target_file]
    ln [-s [-F] | -L | -P] [-f | -i] [-hnv] source_file ... target_dir
    link source_file target_file
make: *** [install] Error 1
```

Since `-r` is not supported in Mac.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,8 +115,8 @@ build-kind-images: build-kind-images-ko
 install: WHAT ?= ./cmd/...
 install:
 	GOOS=$(OS) GOARCH=$(ARCH) go install -ldflags="$(LDFLAGS)" $(WHAT)
-	ln -sfr $(INSTALL_GOBIN)/kubectl-workspace $(INSTALL_GOBIN)/kubectl-ws
-	ln -sfr $(INSTALL_GOBIN)/kubectl-workspace $(INSTALL_GOBIN)/kubectl-workspaces
+	ln -sf $(INSTALL_GOBIN)/kubectl-workspace $(INSTALL_GOBIN)/kubectl-ws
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -115,8 +115,8 @@ build-kind-images: build-kind-images-ko
 install: WHAT ?= ./cmd/...
 install:
 	GOOS=$(OS) GOARCH=$(ARCH) go install -ldflags="$(LDFLAGS)" $(WHAT)
-	ln -sfr $(INSTALL_GOBIN)/kubectl-workspace $(INSTALL_GOBIN)/kubectl-ws
-	ln -sfr $(INSTALL_GOBIN)/kubectl-workspace $(INSTALL_GOBIN)/kubectl-workspaces
+	ln -sf $(INSTALL_GOBIN)/kubectl-workspace $(INSTALL_GOBIN)/kubectl-ws
```
varshaprasad96, line 118: `--relative` creates a symbolic link in linux. We could have separate commands depending on the OS, but wondering if its required here. Can we leave them to be absolute?

ncdc (approved):

Makefile
hardys: Thanks for the fix - an absolute path should be fine I think - I used `-r` thinking we may want the symlinks to remain functional when copied in the Dockerfile, but we ended up runing `ln` [in the Dockerfile](https://github.com/kcp-dev/kcp/blob/main/Dockerfile#L53..L54) instead

Issue #2146, opened 2022-10-06
bug: 'default' ServiceAccount has an invalid secret name on the compute synced namespace


Roming22:
### Describe the bug

* The default secret named `default-token-*****` is renamed to `kcp-default-token-*****` when synchronized to the compute namespace.
* However the `default` service account still references the `default-token-*****` on the compute resource, leading to errors when instanciating PipelineRuns.

### Steps To Reproduce

1. Create a workspace.
2. Bind any APIBinding so that the default namespace is synced to a compute cluster.
3. Go to the compute cluster and find the `kcp-************` namespace for your workspace/namespace.
4. Verify that the secret `kcp-default-token-*****` has been created.
5. Verify that the ServiceAccount references the secret without the `kcp-` prefix

### Expected Behaviour

The ServiceAccount should reference the right secret name.

### Additional Context

This issue is causing a full Pipeline Service outage.

stevekuznetsov:
OK, seems like this is the flow:

 - `kcp` is now creating the `default` service account in the pcluster namespace, with the values from kcp
 - SA controllers in pcluster add in the SA tokens for the `default` SA from the pcluster
 - you end up with an SA in the pcluster that refers to the token secret from `kcp`

stevekuznetsov:
The syncer has `--resources=serviceaccounts` set, so this seems like a foot-gun

stevekuznetsov:
The user did not pass this to the syncer input, `v0.8.2` - the `kubectl kcp workload sync` must have added the defaults for `configmaps/deployments/serviceaccounts/secrets` etc.

Roming22:
We would like to see a release with that fix. We have automation running, so fixing the manifests manually is going to be an issue.

Roming22:
@stevekuznetsov This seems related to https://github.com/kcp-dev/kcp/issues/727.

PR #2150, opened 2022-10-07
:bug: cmd/Sync: remove serviceaccounts from default resource list

jmprusi:
## Summary

Serviceaccounts are currently synced by default, but this shouldn't be the case as it will cause issue overriding downstream SAs.


stevekuznetsov:
@jmprusi can we please remember to link to the issues we're fixing with

Fixes https://github.com/kcp-dev/kcp/issues/2146

ncdc:
No need to pick back to 0.8. We do not maintain multiple previous versions - only the most recent.

PR #2120, opened 2022-10-03
:book: update docs and docs-gen according to hugo syntax

avinal:
## Summary

- [add cobra cli-doc-gen as third party](https://github.com/kcp-dev/kcp/pull/2120/commits/bf26ff3594e963955e87f72d3bc42eb6a8ba0812)
- [hack cli-docs generator to comply hugo syntax](https://github.com/kcp-dev/kcp/pull/2120/commits/16f27463be08e13b7674e5ddc2e2cb9a1d23a1bb)
  - Original code: https://github.com/spf13/cobra/blob/main/doc/md_docs.go
  - Hugo front matter: https://gohugo.io/content-management/front-matter/
- [update kcp docs according to hugo syntax](https://github.com/kcp-dev/kcp/pull/2120/commits/c92a66c4e8354193441fdab4db747a7daa9b9a7e)
  - remove cli documentation see https://github.com/kcp-dev/kcp/issues/2119#issuecomment-1265831212
  - add front matter to all docs
  - fix linting errors in markdown docs

Signed-off-by: Avinal Kumar <avinal@redhat.com>

## Test

Test deployment: https://avinal.space/kcp-dev.github.io/docs (might have lots of broken links, rest works)

PR #2153, opened 2022-10-07
:book: add github actions to push docs 

avinal:
## Summary

This GitHub Actions does the following:
- generates CLI docs
- generates CRD refs
- saves new tags as archive in gh-pages branch for doc versioning
- builds and deploys hugo docs to github pages

## Needs

- depends on PR #2120
- depends on PR #2219

### DIFF
--- a/.github/workflows/docs-gen-and-push.yaml
+++ b/.github/workflows/docs-gen-and-push.yaml
@@ -1,0 +1,116 @@
+name: Generate and push docs
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "v*"
+    paths:
+      - "cmd/**"
+      - "docs/**"
+      - "pkg/**"
+      - "README.md"
+  workflow_dispatch:
+
+jobs:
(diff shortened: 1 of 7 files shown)


.github/workflows/docs-gen-and-push.yaml
```diff
@@ -0,0 +1,55 @@
+name: Generate and push docs
+
+on:
+  push:
+    branches:
+      - main
+      - "release-*"
```
sttts, line 7: is this correct to have so many branches?

avinal: My idea was to have versioned docs if we intend to do so. We can leave it to just `main` for now.

avinal:
Need to fix `hack/verify-go-versions.sh`, it is grepping `go version` from `hugo version` 😆

avinal:
> Need to fix `hack/verify-go-versions.sh`, it is grepping `go-version` from `hugo-version` laughing

fixed ✅

PR #2166, opened 2022-10-11
:seedling: Improve syncer/nsmap local testing experience

markmc:
We have docs on running the syncer locally, but they haven't been updated for #1708 

Also, we can improve our logging to help people figure out when the syncer is failing because of DNS lookups


pkg/syncer/syncer.go
```diff
@@ -198,6 +198,7 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 	dnsIP := ""
 	// Get the DNS IP. The DNS service cannot be recreated as existing dnsConfig won't be updated.
 	err = wait.PollImmediate(2*time.Second, time.Minute, func() (bool, error) {
+		logger.Info(fmt.Sprintf("performing a DNS lookup of %s", cfg.DNSServer))
```
hardys, line 201: To be consistent with the structural logging changes (ref #2134) this should probably either use `WithValues`, or pass multiple args to the `Info` instead of using `Sprintf`?

markmc:
Will reopen if I ever rebase

Issue #2152, opened 2022-10-07
bug: failure to provide identity hash in APIExport / APIBinding permission claims fails silently


hasheddan:
### Describe the bug

I had an `APIExport` that included `permissionClaims` for `deployments/apps`. `Deployments` are not a built-in type, and are only available in a given workspace after setting up a `SyncTarget` to a pcluster (which creates an `APIExport` with name `kubernetes`). 

```
$ k get apiexports kubernetes -o yaml
apiVersion: apis.kcp.dev/v1alpha1
kind: APIExport
metadata:
  annotations:
    kcp.dev/cluster: root
    workload.kcp.dev/skip-default-object-creation: "true"
  creationTimestamp: "2022-10-07T12:17:03Z"
  generation: 3
  name: kubernetes
  resourceVersion: "506"
  uid: e3af3dac-d89a-48ff-8921-5fd0826a7cd3
spec:
  identity:
    secretRef:
      name: kubernetes
      namespace: kcp-system
  latestResourceSchemas:
  - rev-503.deployments.apps
status:
  conditions:
  - lastTransitionTime: "2022-10-07T12:17:03Z"
    status: "True"
    type: IdentityValid
  - lastTransitionTime: "2022-10-07T12:17:03Z"
    status: "True"
    type: VirtualWorkspaceURLsReady
  identityHash: 5fdf7c7aaf407fd1594566869803f565bb84d22156cef5c445d2ee13ac2cfca6
  virtualWorkspaces:
  - url: https://192.168.1.189:6443/services/apiexport/root/kubernetes
```

The `permissionClaims` in my `APIExport` looked like the following:
```yaml
  permissionClaims:
  - group: ""
    resource: secrets
  - group: ""
    resource: namespaces
  - group: ""
    resource: configmaps
  - group: apps
    # (SHOULD HAVE THIS) identityHash: 5fdf7c7aaf407fd1594566869803f565bb84d22156cef5c445d2ee13ac2cfca6
    resource: deployments
```

Because I did not have the identity hash provided, when I created an `APIBinding` for the export, `Deployments` did not show up in the virtual workspace for the exported types.

```
$ k -s https://192.168.1.189:6443/services/apiexport/root/kxp.crossplane.io/clusters/* api-resources
NAME                           SHORTNAMES   APIVERSION                             NAMESPACED   KIND
configmaps                                  v1                                     true         ConfigMap
namespaces                                  v1                                     false        Namespace
[...]

PR #2169, opened 2022-10-11
✨ Check for identityHash in APIExport admission and support multiple versions for APIs in permission claims

### LINKED ISSUE #2161: bug: APIExport controller uses first encountered `GroupResource` match for internal APIs, but separate schemas exist for built-in APIs with multiple versions
### Describe the bug

`APIResourceSchema` for all built-in APIs are generated [here](https://github.com/kcp-dev/kcp/blob/525b29592b773f16f408f805a3f8b044e8e1e8a6/pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_reconcile.go#L293) and then [stored in a slice](https://github.com/kcp-dev/kcp/blob/525b29592b773f16f408f805a3f8b044e8e1e8a6/pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_reconcile.go#L302). When [evaluating whether a permission claim is for an internal API](https://github.com/kcp-dev/kcp/blob/525b29592b773f16f408f805a3f8b044e8e1e8a6/pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_reconcile.go#L111), we [loop through](https://github.com/kcp-dev/kcp/blob/525b29592b773f16f408f805a3f8b044e8e1e8a6/pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_reconcile.go#L280) the generated `APIResourceSchema` and select the [first match by group / resource](https://github.com/kcp-dev/kcp/blob/525b29592b773f16f408f805a3f8b044e8e1e8a6/pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_reconcile.go#L281). However, some built-in APIs (such as `FlowSchema` [`v1beta1`](https://github.com/kcp-dev/kcp/blob/525b29592b773f16f408f805a3f8b044e8e1e8a6/pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_reconcile.go#L444) and [`v1beta2`](https://github.com/kcp-dev/kcp/blob/525b29592b773f16f408f805a3f8b044e8e1e8a6/pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_reconcile.go#L466)) have multiple versions, but have a separate `APIResourceSchema` generated for each version. This means that when we loop through, we always use the schema that was added to the slice first.

### Steps To Reproduce

1. Start kcp.
2. Export `FlowSchemas`.
3. Observe that only `v1beta1` is served.

### Expected Behaviour

I believe what would be expected here is that we do match on `GroupResource`, but that the `APIResourceSchema` returned contains schemas for both versions of the given resource.

### Additional Context

_No response_

hasheddan:
## Summary

This patch set includes two primary changes:

Update the APIResourceSchema generation to consider multiple versions of
an API and append the schemas to one ARS rather than generating a
separate ARS for each.

Adds an APIExport admission plugin that, for the time-being, only
validates that any permissionClaims for types that are not built-in
include an identityHash.

The first can be observed when creating an `APIExport` with `permissionClaims` on `FlowSchemas` or `PriorityLevelConfigurations`.

Before:
```
🤖 (kxp) k -s https://192.168.1.189:6443/services/apiexport/root/kxp.crossplane.io/clusters/* api-versions
apiextensions.crossplane.io/v1
apiextensions.crossplane.io/v1alpha1
flowcontrol.apiserver.k8s.io/v1beta1
pkg.crossplane.io/v1
pkg.crossplane.io/v1alpha1
pkg.crossplane.io/v1beta1
secrets.crossplane.io/v1alpha1
v1
```

After:
```
🤖 (kxp) k -s https://192.168.1.189:6443/services/apiexport/root/kxp.crossplane.io/clusters/* api-versions
apiextensions.crossplane.io/v1
apiextensions.crossplane.io/v1alpha1
flowcontrol.apiserver.k8s.io/v1beta1
flowcontrol.apiserver.k8s.io/v1beta2
pkg.crossplane.io/v1
pkg.crossplane.io/v1alpha1
pkg.crossplane.io/v1beta1
secrets.crossplane.io/v1alpha1
v1
```

The second can be observed by trying to include a `permissionClaim` for something like `Deployment` (not built-in) without an `identityHash`:
```
🤖 (kxp) cat bad-export.yaml 
kind: APIExport
metadata:
  name: data.my.domain
spec:
  latestResourceSchemas:
    - today.widgets.data.my.domain
  permissionClaims:
    - group: ""
      resource: "secrets"
    - group: ""
      resource: "configmaps"
    - group: ""
      resource: "namespaces"
    - group: "apps"
      resource: "deployments"
🤖 (kxp) k apply -f bad-export.yaml 
Error from server (Forbidden): error when creating "bad-export.yaml": apiexports.apis.kcp.dev "data.my.domain" is forbidden: spec.permissionClaims.[3].identityHash: Invalid value: "": identityHash is required for API types that are not built-in
```

## Related issue(s)

Fixes #2161 
xref #2152 (more work to be done on exposing conditions for other failure modes)

### DIFF
--- a/pkg/admission/apiexport/admission.go
+++ b/pkg/admission/apiexport/admission.go
@@ -1,0 +1,96 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package apiexport
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apiserver/pkg/admission"
+
+	"github.com/kcp-dev/kcp/pkg/apis/apis"
+	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
+	builtinapiexport "github.com/kcp-dev/kcp/pkg/virtual/apiexport/schemas/builtin"
+)
+
+// PluginName is the name used to identify this admission webhook.
+const PluginName = "apis.kcp.dev/APIExport"
+
+// Register registers the reserved name admission webhook.
+func Register(plugins *admission.Plugins) {
+	plugins.Register(PluginName,
+		func(_ io.Reader) (admission.Interface, error) {
+			return NewAPIExportAdmission(builtinapiexport.IsBuiltInAPI), nil
+		})
+}
+
+// APIExportAdmission is an admission plugin for checking APIExport validity.
+type APIExportAdmission struct {
+	*admission.Handler
+
+	isBuiltIn func(apisv1alpha1.GroupResource) bool
+}
+
+// NewAPIExportAdmission constructs a new APIExportAdmission admission plugin.
+func NewAPIExportAdmission(isBuiltIn func(apisv1alpha1.GroupResource) bool) *APIExportAdmission {
+	return &APIExportAdmission{
+		Handler:   admission.NewHandler(admission.Create, admission.Update),
+		isBuiltIn: isBuiltIn,
+	}
+}
+
+// Ensure that the required admission interfaces are implemented.
+var _ = admission.ValidationInterface(&APIExportAdmission{})
+
+// Validate ensures that the APIExport is valid.
+func (e *APIExportAdmission) Validate(ctx context.Context, a admission.Attributes, _ admission.ObjectInterfaces) (err error) {
+	if a.GetResource().GroupResource() != apisv1alpha1.Resource("apiexports") {
+		return nil
+	}
+
+	if a.GetKind().GroupKind() != apisv1alpha1.Kind("APIExport") {
+		return nil
+	}
+
+	u, ok := a.GetObject().(*unstructured.Unstructured)
+	if !ok {
+		return fmt.Errorf("unexpected type %T", a.GetObject())
+	}
+	ae := &apisv1alpha1.APIExport{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, ae); err != nil {
+		return fmt.Errorf("failed to convert unstructured to APIExport: %w", err)
+	}
+
+	for i, pc := range ae.Spec.PermissionClaims {
+		if pc.IdentityHash == "" && !e.isBuiltIn(pc.GroupResource) && pc.Group != apis.GroupName {
+			return admission.NewForbidden(a,
+				field.Invalid(
+					field.NewPath("spec").
+						Child("permissionClaims").
+						Index(i).
+						Child("identityHash"),
+					"",
+					"identityHash is required for API types that are not built-in"))
+		}
+	}
+
+	return nil
+}
--- a/pkg/admission/apiexport/admission_test.go
+++ b/pkg/admission/apiexport/admission_test.go
@@ -1,0 +1,231 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package apiexport
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apiserver/pkg/admission"
+	"k8s.io/apiserver/pkg/authentication/user"
+
+	"github.com/kcp-dev/kcp/pkg/admission/helpers"
+	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
+)
+
+func createAttr(name string, obj runtime.Object, kind, resource string) admission.Attributes {
+	return admission.NewAttributesRecord(
+		helpers.ToUnstructuredOrDie(obj),
+		nil,
+		apisv1alpha1.Kind(kind).WithVersion("v1alpha1"),
+		"",
+		name,
+		apisv1alpha1.Resource(resource).WithVersion("v1alpha1"),
+		"",
+		admission.Create,
+		&metav1.CreateOptions{},
+		false,
+		&user.DefaultInfo{},
+	)
+}
+
+func updateAttr(name string, obj runtime.Object, kind, resource string) admission.Attributes {
+	return admission.NewAttributesRecord(
+		helpers.ToUnstructuredOrDie(obj),
+		helpers.ToUnstructuredOrDie(obj),
+		apisv1alpha1.Kind(kind).WithVersion("v1alpha1"),
+		"",
+		name,
+		apisv1alpha1.Resource(resource).WithVersion("v1alpha1"),
+		"",
+		admission.Update,
+		&metav1.UpdateOptions{},
+		false,
+		&user.DefaultInfo{},
+	)
+}
+
+func TestAdmission(t *testing.T) {
+	cases := map[string]struct {
+		attr        admission.Attributes
+		update      bool
+		kind        string
+		resource    string
+		hasIdentity bool
+		isBuiltIn   bool
+		modifyPCs   func([]apisv1alpha1.PermissionClaim) []apisv1alpha1.PermissionClaim
+		want        error
+	}{
+		"NotAPIExportKind": {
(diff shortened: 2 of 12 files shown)


hasheddan:
Need to cleanup some of these e2e tests that I believe are correctly failing now.

pkg/admission/apiexport/admission.go
```diff
@@ -0,0 +1,96 @@
...
+		return fmt.Errorf("failed to convert unstructured to APIExport: %w", err)
+	}
+
+	for i, pc := range ae.Spec.PermissionClaims {
+		if pc.IdentityHash == "" && !e.isBuiltInFn(pc.GroupResource) && pc.Group != apis.GroupName {
+			return admission.NewForbidden(a,
+				field.Invalid(
+					field.NewPath("spec").
+						Child("permissionClaims").
+						Child(fmt.Sprintf("[%d]", i)).
```
ncdc, line 88: You can use .Index here

pkg/admission/apiexport/admission_test.go
```diff
@@ -0,0 +1,217 @@
...
+		apisv1alpha1.Resource(resource).WithVersion("v1alpha1"),
+		"",
+		admission.Update,
+		&metav1.UpdateOptions{},
+		false,
+		&user.DefaultInfo{},
+	)
+}
+
+func TestAdmission(t *testing.T) {
```
ncdc, line 68: Not required but I think we can simplify the length of the test...

- Change the test case struct to something like
```go
update bool
builtIn bool
kind string
resource string
hasIdentity bool
want error
```
- Update the main t.Run logic to create an APIExport based on the testCase data, along with either a create or update attr

I think this will make it shorter if you are interested

hasheddan: Definitely interested -- thanks for the recommendation!

hasheddan: Updated 👍🏻

pkg/admission/apiexport/admission.go
```diff
@@ -0,0 +1,96 @@
...
+		func(_ io.Reader) (admission.Interface, error) {
+			return NewAPIExportAdmission(builtinapiexport.IsBuiltInAPI), nil
+		})
+}
+
+// APIExportAdmission is an admission plugin for checking APIExport validity.
+type APIExportAdmission struct {
+	*admission.Handler
+
+	isBuiltInFn func(apisv1alpha1.GroupResource) bool
```
ncdc, line 49: nit: I would call this `isBuiltIn`

pkg/admission/apiexport/admission.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// NewAPIExportAdmission constructs a new APIExportAdmission admission plugin.
+func NewAPIExportAdmission(isInternal func(apisv1alpha1.GroupResource) bool) *APIExportAdmission {
```
ncdc, line 53:
```suggestion
func NewAPIExportAdmission(isBuiltIn func(apisv1alpha1.GroupResource) bool) *APIExportAdmission {
```

pkg/admission/apiexport/admission_test.go
```diff
@@ -0,0 +1,174 @@
...
+		&metav1.CreateOptions{},
+		false,
+		&user.DefaultInfo{},
+	)
+}
+
+func updateAttr(name string, obj runtime.Object, kind, resource string) admission.Attributes {
+	return admission.NewAttributesRecord(
+		helpers.ToUnstructuredOrDie(obj),
+		nil,
```
ncdc, line 54: Since this is an update, I know it doesn't matter for the code under test, but should we fill in the "old" object here?

pkg/admission/apiexport/admission_test.go
```diff
@@ -0,0 +1,174 @@
...
+}
+
+func TestAdmission(t *testing.T) {
+	cases := map[string]struct {
+		attr        admission.Attributes
+		update      bool
+		kind        string
+		resource    string
+		hasIdentity bool
+		isBuiltIn   func(apisv1alpha1.GroupResource) bool
```
ncdc, line 74: Unless you're intending to vary the return based on the input, I think this can be a normal `bool` and you can set the function pointer once, inside t.Run?

pkg/admission/apiexport/admission_test.go
```diff
@@ -0,0 +1,174 @@
...
+		},
+		"ValidUpdateNonInternal": {
+			update:      true,
+			kind:        "APIExport",
+			resource:    "apiexports",
+			hasIdentity: true,
+			isBuiltIn: func(gr apisv1alpha1.GroupResource) bool {
+				return false
+			},
+		},
```
ncdc, line 136: Missing test case(s) where builtIn=true

pkg/admission/apiexport/admission_test.go
```diff
@@ -0,0 +1,174 @@
...
+			update:   true,
+			kind:     "APIExport",
+			resource: "apiexports",
+			isBuiltIn: func(gr apisv1alpha1.GroupResource) bool {
+				return false
+			},
+			want: field.Invalid(
+				field.NewPath("spec").
+					Child("permissionClaims").
+					Index(0).
```
ncdc, line 115: Should probably add a test case that verifies a nonzero index?

pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_reconcile.go
```diff
@@ -92,23 +79,25 @@ func (c *APIReconciler) reconcile(ctx context.Context, apiExport *apisv1alpha1.A
...
 		if _, found := apiResourceSchemas[gr]; found {
 			if otherClaim, found := claims[gr]; found {
 				logger.Info("permission claim is shadowed by another claim", "claim", pc, "otherClaim", otherClaim)
-			} else {
-				logger.Info("permission claim is shadowed by exported resource", "claim", pc)
+				continue
 			}
-			continue
+
+			logger.Info("permission claim is shadowed by exported resource", "claim", pc)
```
ncdc, line 92: Do we need a continue after this?

pkg/virtual/syncer/schemas/builtin/builtin_test.go
```diff
@@ -0,0 +1,12 @@
+package builtin
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	_ "k8s.io/kubernetes/pkg/apis/core/install"
```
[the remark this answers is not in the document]
ncdc: Sounds like we should?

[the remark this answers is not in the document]
hasheddan: I'm happy to, though in theory this is a change in behavior that is not strictly how all consumers would have to behave today. In practice, kcp is installing the core types into the legacy schema before this `init()` runs, but not all consumers have to. In general I don't love this pattern of registering schemas in `init()` calls, and would lean towards going ahead and moving the anonymous import (as you suggest) to cover the 99% case here.

pkg/admission/apiexport/admission_test.go
hasheddan: Happy to 👍🏻

pkg/virtual/syncer/schemas/builtin/builtin_test.go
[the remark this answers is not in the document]
ncdc: @sttts what do you prefer?

pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_reconcile.go
hasheddan: good catch -- yes 👍🏻

pkg/admission/apiexport/admission_test.go
hasheddan: will do 👍🏻

ncdc:
/lgtm

Would like to hear Stefan's thoughts on the import question

pkg/virtual/syncer/schemas/builtin/builtin_test.go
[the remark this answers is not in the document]
hasheddan: @sttts any thoughts here?

pkg/virtual/syncer/schemas/builtin/builtin.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/kcp-dev/kcp/pkg/virtual/framework/internalapis"
+)
+
+// syncerSchemas contains a list of internal APIs that should be exposed for the
+// syncer of any SyncTarget.
+var SyncerSchemas map[apisv1alpha1.GroupResource]*apisv1alpha1.APIResourceSchema
+
+func init() {
+	genericcontrolplane.Install(legacyscheme.Scheme)
+	schemes := []*runtime.Scheme{legacyscheme.Scheme}
```
sttts, line 39: what is this used for? don't like to see the legacyscheme in our code.

pkg/virtual/syncer/schemas/builtin/builtin_test.go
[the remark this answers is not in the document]
sttts: don't have much context here. But in general it is highly discourage to use the legacyscheme in any way, especially in new code like kcp. Create dedicated schemes and add the group you want via the AddToScheme methods.

pkg/virtual/syncer/schemas/builtin/builtin.go
hasheddan: @sttts this isn’t strictly new in this patch set, just being relocated from the syncer API reconciler. I presume the legacy scheme is being used as the sync target is assumed to be Kubernetes API-compatible. Happy to update this to explicitly add each type in the internal API list below to the scheme.

ncdc: I would recommend doing a follow-up to deal with legacyscheme usage as a whole

hasheddan: @ncdc happy to open a tracking issue and tackle 👍🏻

ncdc:
Approving. Keep merging & iterate on 😄 
/approve

pkg/virtual/syncer/schemas/builtin/builtin.go
ncdc: @hasheddan yes let's do that please, thanks!

PR #2174, opened 2022-10-11
🐛 Fix permissionclaim patch thrashing

ncdc:
## Summary
- Add map keys to APIBinding.Status.AppliedPermissionClaims and .ExportPermissionClaims
- Switched to using a sorted list when calculating applied permission claims so it doesn't keep generating new patches because of different ordering in the list.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/apis/apis/v1alpha1/types_apibinding.go
+++ b/pkg/apis/apis/v1alpha1/types_apibinding.go
@@ -174,11 +174,17 @@ type APIBindingStatus struct {
 	// state in spec.permissionClaims.
 	//
 	// +optional
+	// +listType=map
+	// +listMapKey=group
+	// +listMapKey=resource
 	AppliedPermissionClaims []PermissionClaim `json:"appliedPermissionClaims,omitempty"`
 
 	// exportPermissionClaims records the permissions that the export provider is asking for
 	// the binding to grant.
 	// +optional
+	// +listType=map
+	// +listMapKey=group
+	// +listMapKey=resource
 	ExportPermissionClaims []PermissionClaim `json:"exportPermissionClaims,omitempty"`
 }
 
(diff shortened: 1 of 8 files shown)


csams (approved):
/lgtm

stevekuznetsov (approved):

sttts:
This PR is wrong. It might be a hot-fix, but it's the wrong direction.

pkg/apis/apis/v1alpha1/types_apibinding.go
```diff
@@ -174,11 +174,17 @@ type APIBindingStatus struct {
...
 	// +optional
+	// +listType=map
+	// +listMapKey=group
+	// +listMapKey=resource
 	AppliedPermissionClaims []PermissionClaim `json:"appliedPermissionClaims,omitempty"`
 
 	// exportPermissionClaims records the permissions that the export provider is asking for
 	// the binding to grant.
 	// +optional
+	// +listType=map
```
sttts, line 185: this must be atomic

pkg/apis/apis/v1alpha1/types_apibinding.go
```diff
@@ -174,11 +174,17 @@ type APIBindingStatus struct {
 	// state in spec.permissionClaims.
 	//
 	// +optional
+	// +listType=map
```
sttts, line 177: this must be atomic

sttts: the spec one can be set.

pkg/apis/apis/v1alpha1/types_apibinding.go
sttts: if atomic broke something, then the patch code is wrong.

PR #2176, opened 2022-10-12
Revert "🐛 Fix permissionclaim patch thrashing"

sttts:
Reverts kcp-dev/kcp#2174


ncdc:
Would prefer a new PR that fixes the list type instead of a revert if possible

sttts:
/hold

This is a placeholder for a discussion about a real solution.

Issue #2181, opened 2022-10-12
bug: workloads APIBinding stays failed if fails


mjudeikis:
### Describe the bug

Sometimes on startup APIResourceSchema lister fails in apibinding reconciler:
```
getAPIResourceSchema: func(clusterName logicalcluster.Name, name string) (*apisv1alpha1.APIResourceSchema, error) {
			apiResourceSchema, err := apiResourceSchemaInformer.Lister().Get(clusters.ToClusterAwareKey(clusterName, name))
			if errors.IsNotFound(err) {
				return temporaryRemoteShardApiResourceSchemaInformer.Lister().Get(clusters.ToClusterAwareKey(clusterName, name))
			}
			return apiResourceSchema, err
		},
```
and Binding is marked as Failed (error message was appended while debugging, orginal message says just "internal error")
```
  - lastTransitionTime: "2022-10-12T18:43:53Z"                                                                                                                                                                                                                                                                                                                                                                                                                                                
    message: An internal error occurred with the APIExport apiresourceschema.apis.kcp.dev                                                                                                                                                                                                                                                                                                                                                                                                     
      "root|v220923-836dfac8.synctargets.workload.kcp.dev" not found                                                                                                                                                                                                                                                                                                                                                                                                                          
    reason: InternalError                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
    severity: Error                                                                                                                                                                                                                                                                                                                                                                                                                                                                           
    status: "False"                                                                                                                                                                                                                                                                                                                                                                                                                                                                           
    type: APIExportValid   
```

And after subsequancial restarts/edits status is not cleared down or modified to `status: True`

For this reason APiBinding shows failed, when it is not.

https://github.com/kcp-dev/kcp/blob/4abc675460f553354e5d44cd0eab409936ad14b9/pkg/reconciler/apis/apibinding/apibinding_reconcile.go#L467 

### Steps To Reproduce

1. Wait until one of the APIBindings (for me workloads almost always) fails on startup
2. Observe status never fixed
3. Be sad because of it

### Expected Behaviour

Status be changed after requeue/reconcile

### Additional Context

_No response_

mjudeikis:
```
$ kubectl get apibinding workload.kcp.dev-afhq1 -o yaml                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
[...]

PR #2189, opened 2022-10-14
🐛 Correctly serve APIBindings in the APIExport virtual workspace

### LINKED ISSUE #1421: Make APIBinding readible in APIExport VirtualWorkspace
Consider that we have a service provider called "build". That service provider needs some things to be present in the user's workspace to fully function. Let's say that these resources are:
* namespace
* ConfigMap & Secret
* NetworkPolicies, plus maybe something more

The first thought of how to meet this requirement is to use CWT and the initializer. Initializer creates the APIBinding for the service provider plus all needed resources. So far so good.
However, after some time the logic of the service provider is changed and it needs something more to be created or with different parameters. How should we meet such a requirement?
If we change the logic of the CWT initializer, then it applies the change only for the new workspaces, but not the existing ones.
Instead of using CWTs, we could move the creation of the resources into the service provider logic, but the "build" controller is watching a VirtualWorkspace of its APIExport (within the boundaries of the API that is defined in the APIExport), so it doesn't know when a new workspace is created.

The VirtualWorkspace should include also all APIBindings of the particular APIExport automatically. In that way, the service provider could have a controller watching the APIBindings so it would be notified when a new one is created. Just to be clear, the controller wouldn't do anything with the APIBindings, it would only watch them.

The APIExport could technically claim the APIBindings, but it would get all of them. It would be great if the VirtualWorkspace logic could filter them out so it would get only the ones that were created for the particular APIExport.

ncdc:
## Summary

Fix bug in the APIExport virtual workspace that prevented it from including APIBindings for the APIExport when there wasn't a permission claim for APIBindings.

Switch from checking APIBinding phase for determining what actions to perform to instead always trying to resolve desired state.

Add e2e test for APIBindings with/without permission claims in the APIExport virtual workspace.

## Related issue(s)

Fixes #1421
Fixes: #2181

### DIFF
--- a/pkg/reconciler/apis/apibinding/apibinding_reconcile_test.go
+++ b/pkg/reconciler/apis/apibinding/apibinding_reconcile_test.go
@@ -102,8 +102,8 @@ var (
 			WithWorkspaceReference("org:some-workspace", "conflict").
 			WithBoundResources(
 			new(boundAPIResourceBuilder).
-				WithGroupResource("other.io", "widgets").
-				WithSchema("another.widgets.other.io", "anotherwidgetsuid").
+				WithGroupResource("kcp.dev", "widgets").
+				WithSchema("another.widgets.kcp.dev", "anotherwidgetsuid").
 				BoundAPIResource,
 		)
 
@@ -139,11 +139,11 @@ var (
 
 	someOtherWidgetsAPIResourceSchema = &apisv1alpha1.APIResourceSchema{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "another.widgets.other.io",
+			Name: "another.widgets.kcp.dev",
 			UID:  "anotherwidgetsuid",
 		},
 		Spec: apisv1alpha1.APIResourceSchemaSpec{
-			Group: "other.io",
+			Group: "kcp.dev",
 			Names: apiextensionsv1.CustomResourceDefinitionNames{
 				Plural:   "widgets",
 				Singular: "widget",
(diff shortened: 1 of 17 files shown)


test/e2e/virtual/apiexport/virtualworkspace_test.go
```diff
@@ -311,6 +319,273 @@ func TestAPIExportVirtualWorkspace(t *testing.T) {
...
+		if err == nil {
+			return true, ""
+		}
+		require.False(t, apierrors.IsConflict(err))
+		return false, err.Error()
+	}, wait.ForeverTestTimeout, 100*time.Millisecond, "error updating %s|%s to accept claims", workspace1, binding1.Name)
+
+	t.Logf("Verifying APIExport 1 serves APIBindings 1|1, 1|2, and 2|1")
+	verifyBindings(workspace1, "export1", func(bindings []apisv1alpha1.APIBinding) error {
+		// "workspace1|binding1", "workspace2|binding1", "workspace1|binding2")
```
nrb, line 516: Remove?

ncdc: Can do

nrb:
Looking through issues, it looks like this PR fixes https://github.com/kcp-dev/kcp/issues/1421, is that the case?

ncdc:
> Looking through issues, it looks like this PR fixes #1421, is that the case?

Thanks, couldn't find it quickly

pkg/reconciler/apis/apibinding/apibinding_reconcile_test.go
```diff
@@ -139,11 +139,11 @@ var (
 
 	someOtherWidgetsAPIResourceSchema = &apisv1alpha1.APIResourceSchema{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "another.widgets.other.io",
```
stevekuznetsov, line 142: What's the meaning of this change?

stevekuznetsov (commented):
Change to reconciler structure seems good

pkg/reconciler/apis/apibinding/apibinding_reconcile_test.go
ncdc: This is meant to be the conflicting schema, and it needs to be in the same API group. The conflict test cases passed before because they were relying on the getCRD behavior defined below to return a widgets.kcp.dev CRD (the conflict checker now uses a schema instead of a CRD).

ncdc:
Tunnel flake
/retest

PR #2192, opened 2022-10-15
✨ Publish pkg/client as its own go module

sttts:


pkg/client/go.mod
```diff
@@ -0,0 +1,55 @@
+module github.com/kcp-dev/kcp/pkg/client
+
+go 1.18
+
+require (
```
stevekuznetsov, line 5: We will just need to be judicious about this list :)

stevekuznetsov:
/retest

@sttts lots of failures ... ?

PR #2193, opened 2022-10-15
:bug: virtualworkspace/workspaces: fix generateName

sttts:
Workspaces with `metaname.generateName` have empty name. But the VW code depended on `workspace.name` for roles and bindings and error messages.


hasheddan (approved):
/lgtm

PR #2132, opened 2022-10-05
🌱 cache: run the replication controller when the cache server is enabled

p0lyn0mial:
## Summary
it works!

```
❯ k get apiexports apiresource.kcp.dev -ojson
{
    "apiVersion": "apis.kcp.dev/v1alpha1",
    "kind": "APIExport",
    "metadata": {
        "annotations": {
            "internal.sharding.kcp.dev/replicate": "true",
            "kcp.dev/cluster": "root"
        },
        "creationTimestamp": "2022-10-05T09:12:46Z",
        "generation": 2,
        "name": "apiresource.kcp.dev",
        "resourceVersion": "364",
        "uid": "4c263ff5-a0b3-4107-ba6f-36e55c0e938b"
    },
    "spec": {
        "identity": {
            "secretRef": {
                "name": "apiresource.kcp.dev",
                "namespace": "kcp-system"
            }
        },
        "latestResourceSchemas": [
            "v220628-546034da.apiresourceimports.apiresource.kcp.dev",
            "v220628-546034da.negotiatedapiresources.apiresource.kcp.dev"
        ],
        "maximalPermissionPolicy": {
            "local": {}
        }
    },
    "status": {
        "conditions": [
            {
                "lastTransitionTime": "2022-10-05T09:12:46Z",
                "status": "True",
                "type": "IdentityValid"
            },
            {
                "lastTransitionTime": "2022-10-05T09:12:46Z",
                "status": "True",
                "type": "VirtualWorkspaceURLsReady"
            }
        ],
        "identityHash": "de433c6a7209dc59510af6bacb79e80e603e766aeb6d2a7779bbcc2b0bcdbb4d",
        "virtualWorkspaces": [
            {
                "url": "https://192.168.32.104:6443/services/apiexport/root/apiresource.kcp.dev"
            }
        ]
    }
}
```
replicated resource is almost identical, expect the generation and the RV
```
❯ curl -k 'https://localhost:6443/services/cache/shards/*/clusters/*/apis/apis.kcp.dev/v1alpha1/apiexports'
{
  "apiVersion": "apis.kcp.dev/v1alpha1",
  "items": [
    {
      "apiVersion": "apis.kcp.dev/v1alpha1",
      "kind": "APIExport",
      "metadata": {
        "annotations": {
          "internal.sharding.kcp.dev/replicate": "true",
          "kcp.dev/cluster": "root",
          "kcp.dev/shard": "root"
        },
        "creationTimestamp": "2022-10-05T09:12:46Z",
[...]

PR #2197, opened 2022-10-17
🌱 e2e framework:  introduce KcpConfigOption function

p0lyn0mial:
## Summary
Allows for providing a custom configuration when creating a `PrivateKcpServer`. Could be extended in the future to `SharedKcpServer`

It will be used in https://github.com/kcp-dev/kcp/pull/2132. The `PrivateKcpServer` will be created with `WithScratchDirectories`. The scratch directories will be shared with an instance of `the cache server`

## Related issue(s)

### DIFF
--- a/test/e2e/framework/kcp.go
+++ b/test/e2e/framework/kcp.go
@@ -95,12 +95,22 @@ type kcpFixture struct {
 
 // PrivateKcpServer returns a new kcp server fixture managing a new
 // server process that is not intended to be shared between tests.
-func PrivateKcpServer(t *testing.T, args ...string) RunningServer {
+func PrivateKcpServer(t *testing.T, options ...KcpConfigOption) RunningServer {
 	serverName := "main"
-	f := newKcpFixture(t, kcpConfig{
-		Name: serverName,
-		Args: args,
-	})
+
+	cfg := &kcpConfig{Name: serverName}
+	for _, opt := range options {
+		cfg = opt(cfg)
+	}
+
+	if len(cfg.ArtifactDir) == 0 || len(cfg.DataDir) == 0 {
+		artifactDir, dataDir, err := ScratchDirs(t)
+		require.NoError(t, err, "failed to create scratch dirs: %v", err)
+		cfg.ArtifactDir = artifactDir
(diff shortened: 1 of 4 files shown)


test/e2e/framework/kcp.go
```diff
@@ -213,12 +231,34 @@ type RunningServer interface {
...
+type KcpConfigOption func(*kcpConfig) *kcpConfig
+
+// WithScratchDirectories adds custom scratch directories to a kcp configuration
+func WithScratchDirectories(artifactDir, dataDir string) KcpConfigOption {
+	return func(cfg *kcpConfig) *kcpConfig {
+		cfg.ArtifactDir = artifactDir
+		cfg.DataDir = dataDir
+		return cfg
+	}
+}
```
sttts, line 244: why isn't this the default to have independent artifact and data dirs?

p0lyn0mial: The scratch dirs are provided/generated by default by the test framework. See https://github.com/kcp-dev/kcp/pull/2197/files#diff-6119c8a3bef16c15807ae827fde6e0f6b9e6d893969064b4cc24063755a90fffR106 and https://github.com/kcp-dev/kcp/pull/2197/files#diff-6119c8a3bef16c15807ae827fde6e0f6b9e6d893969064b4cc24063755a90fffR141

p0lyn0mial: The `PrivateKcpServer` was extended so that the scratch dirs can be provided externally by the test.

sttts: if they are created already, is the problem that you cannot access them?

p0lyn0mial: I wanted to create these files before starting my tests, so that the cache-server artifact, data are placed in the same location.

p0lyn0mial: I cannot rely on the `PrivateKcpServer` method because it immediately starts the kcp server.

p0lyn0mial: and in a standalone mode the kcp requires a path to kubeconifig that holds config for the cache-server

PR #2206, opened 2022-10-17
🌱 cmd/syncer: switch to contextual logging

ncdc:
## Summary
cmd/syncer: switch to contextual logging

## Related issue(s)

Part of #1694

### DIFF
--- a/cmd/syncer/cmd/syncer.go
+++ b/cmd/syncer/cmd/syncer.go
@@ -53,7 +53,7 @@ func NewSyncerCommand() *cobra.Command {
 			}
 
 			ctx := genericapiserver.SetupSignalContext()
-			if err := Run(options, ctx); err != nil {
+			if err := Run(ctx, options); err != nil {
 				return err
 			}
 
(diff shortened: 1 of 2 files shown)


cmd/syncer/cmd/syncer.go
```diff
@@ -53,7 +53,11 @@ func NewSyncerCommand() *cobra.Command {
 			}
 
 			ctx := genericapiserver.SetupSignalContext()
-			if err := Run(options, ctx); err != nil {
+
+			logger := klog.Background()
+			ctx = klog.NewContext(ctx, logger)
```
hardys, line 58: We already have a `ctx` from `SetupSignalContext` above, is there a reason we can't just call `klog.FromContext` using that object?

ncdc: You're right, `klog.FromContext()` will return `klog.Background()` if there isn't a logger in the context. Will adjust.

ncdc:
@hardys updated

PR #2214, opened 2022-10-18
:sparkles: Add Upsync controller

bipuladh:
Signed-off-by: Bipul Adhikari <badhikar@redhat.com>

## Summary
Adds Upsync controller
Creates resources in the Upstream workspace for labelled resources.


pkg/syncer/upsync/upsync_controller.go
```diff
@@ -0,0 +1,174 @@
...
+		syncTargetKey:       syncTargetKey,
+	}
+	logger := logging.WithReconciler(syncerLogger, controllerName)
+
+	syncerInformers.AddDownstreamEventHandler(
+		func(gvr schema.GroupVersionResource) cache.ResourceEventHandler {
+			logger.V(2).Info("Set up downstream informer", "gvr", gvr.String())
+			return cache.ResourceEventHandlerFuncs{
+				AddFunc: func(obj interface{}) {
+					if isResourceToBeSynced(obj.(*unstructured.Unstructured)) {
```
stevekuznetsov, line 115: Filtering in an event handler makes your controller edge-based, not level-driven. The state of the object at enqueue time does not necessarily match the state at reconcile time. Either filter via client options in the informer factory, or do it in reconciliation.

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,113 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+)
+
+func (c *Controller) process(ctx context.Context, gvr schema.GroupVersionResource, key string) error {
+	logger := klog.FromContext(ctx)
+
+	downstreamNamespace, downstreamName, err := cache.SplitMetaNamespaceKey(key)
```
stevekuznetsov, line 20:
```suggestion
	downstreamNamespace, downstreamName, err := keyfunctions.SplitMetaNamespaceKey(key)
```

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,113 @@
...
+	if !resourceExists {
+		logger.Info("Downstream object doesn't exist")
+		return nil
+	}
+
+	u, ok := obj.(*unstructured.Unstructured)
+
+	if !ok {
+		// type mismatch
+		logger.Info("Type mismatch of resource object")
```
stevekuznetsov, line 96: Print the unexpected type with `%T`

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,113 @@
...
+
+	return nil
+}
+
+func (c *Controller) createResourceInUpstream(ctx context.Context, gvr schema.GroupVersionResource, upstreamNS string, upstreamLogicalCluster logicalcluster.Name, downstreamObj *unstructured.Unstructured) error {
+	// Make a deepcopy
+	downstreamObject := downstreamObj.DeepCopy()
+
+	// Create the resource
+	if _, err := c.upstreamClient.Cluster(upstreamLogicalCluster).Resource(gvr).Create(ctx, downstreamObject, metav1.CreateOptions{}); err != nil {
```
stevekuznetsov, line 109: Do you need to check to see if it's namespaced?

davidfestal:
@bipuladh if this PR was not in a reviewable state (which I had understood from what you told me), you should probably have created it as a DRAFT PR.

pkg/syncer/upsync/upsync_process.go
davidfestal: Well, for now we'd upsync only cluster-wide resources (namely PVs), but in the future (as soon as we start upsyncing Pods related the synced deployments, we'll need to support namespaced resources.

pkg/syncer/syncer.go
```diff
@@ -227,6 +236,12 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 		return err
 	}
 
+	logger.Info("Creating upsyncer")
+	upSyncer, err := upsync.NewUpSyncer(logger, cfg.SyncTargetWorkspace, cfg.SyncTargetName, syncTargetKey, upstreamDynamicClusterClient, downstreamDynamicClient, upstreamInformers, downstreamFilteredInformers, syncerInformers, syncTarget.GetUID())
```
davidfestal, line 240: Now the the Upsync Virtual Workspace has been merged, instead of using `upstreamDynamicClusterClient`, `upstreamInformers`, etc... based on the `upstreamConfig` wih the Syncer VirtualWorkspace URL, as done here https://github.com/kcp-dev/kcp/blob/6a4456e59b58cc0b6b40f569b169dcc141bb81db/pkg/syncer/syncer.go#L148. But you would create similar ones, but based on  modified config in which you would just replace the string `/services/syncer/` by `/services/upsyncer/` in the host URL.

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,183 @@
...
+	logger = logger.WithValues(DownstreamNamespace, downstreamNamespace, DownstreamName, downstreamName)
+
+	var upstreamLocator *shared.NamespaceLocator
+	var locatorExists bool
+	var downstreamObject interface{}
+	// Namespace scoped resource
+	if downstreamNamespace != "" {
+		var nsObj runtime.Object
+		if nsObj, err = c.downstreamNamespaceLister.Get(downstreamNamespace); err != nil {
+			logger.Error(err, "Error getting downstream Namespace from downstream namespace lister", "ns", downstreamNamespace)
```
davidfestal, line 55: We should not continue here. In case of error, `nsObj` may be `nil`.
We should distinguish between NotFound error and another error.
- In case of NotFound, I assume there's not much we can do, because we can't even find back the upstream namespace to do some cleanup. So just return `nil` to have the reconciliation successful
- In case of another error, just return the error.

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,183 @@
...
+			logger.Info(fmt.Sprintf("Error: downstream ns expected to be metav1.Object got %T", nsObj))
+		}
+
+		upstreamLocator, locatorExists, err = shared.LocatorFromAnnotations(nsMeta.GetAnnotations())
+		if err != nil {
+			logger.Error(err, "Error ")
+		}
+
+		if !locatorExists || upstreamLocator == nil {
+			return nil
```
davidfestal, line 68: Should we log this ?
This is not normal I assume (at least a Warning should be logged)

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+		downstreamObject, err = c.downstreamClient.Resource(gvr).Namespace(downstreamNamespace).Get(ctx, downstreamName, metav1.GetOptions{})
+		if err != nil {
```
davidfestal, line 71: You have to manage the case here and return.
Distinguish between NotFound and a real error.
- If the object is not found, we might want to delete the object upstream (we can do it since it has the same name and we know the upstream namespace through the locator)
- In case of other errors, just return the error.

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,183 @@
...
+			logger.Error(err, "Could not find the resource downstream")
+		}
+	}
+	if downstreamNamespace == "" {
+		downstreamObject, err = c.downstreamClient.Resource(gvr).Get(ctx, downstreamName, metav1.GetOptions{})
+		if err != nil {
+			logger.Error(err, "Could not find the resource downstream")
```
davidfestal, line 78: You have to distinguish between NotFound and other errors.
The NotFound case is not an error. Normally it should trigger a deletion upstream (if the resource exists upstream). For now we cannot do it because the Namespace locator is on the cluster-wide resource itself.
So in this case let's return nil (==success) and add a TODO to remember the fact that we should find a way to cleanup the upstream resource.

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,183 @@
...
+		if upstreamLocator, locatorExists, err = shared.LocatorFromAnnotations(objMeta.GetAnnotations()); err != nil {
+			logger.Error(err, "Error decoding annotations on DS cluster-wide res")
+		}
+
+		if !locatorExists || upstreamLocator == nil {
+			return nil
+		}
+	}
+
+	u, ok := downstreamObject.(*unstructured.Unstructured)
```
davidfestal, line 93: If we know it will be an `unstructured` object, let's convert it at the beginning of the function. So you won't have to convert to `metav1`

bipuladh: updated!

pkg/syncer/upsync/upsync_process.go
bipuladh: updated!

pkg/syncer/upsync/upsync_process.go
bipuladh: logging and throwing an error

pkg/syncer/upsync/upsync_process.go
bipuladh: updated

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,183 @@
...
+		if !locatorExists || upstreamLocator == nil {
+			return nil
+		}
+		downstreamObject, err = c.downstreamClient.Resource(gvr).Namespace(downstreamNamespace).Get(ctx, downstreamName, metav1.GetOptions{})
+		if err != nil {
+			logger.Error(err, "Could not find the resource downstream")
+		}
+	}
+	if downstreamNamespace == "" {
+		downstreamObject, err = c.downstreamClient.Resource(gvr).Get(ctx, downstreamName, metav1.GetOptions{})
```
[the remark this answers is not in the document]
bipuladh: updated

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,183 @@
...
+
+		if !locatorExists || upstreamLocator == nil {
+			return nil
+		}
+		downstreamObject, err = c.downstreamClient.Resource(gvr).Namespace(downstreamNamespace).Get(ctx, downstreamName, metav1.GetOptions{})
+		if err != nil {
+			logger.Error(err, "Could not find the resource downstream")
+		}
+	}
+	if downstreamNamespace == "" {
```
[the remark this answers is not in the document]
bipuladh: updated!

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,183 @@
...
+
+		upstreamLocator, locatorExists, err = shared.LocatorFromAnnotations(nsMeta.GetAnnotations())
+		if err != nil {
+			logger.Error(err, "Error ")
+		}
+
+		if !locatorExists || upstreamLocator == nil {
+			return nil
+		}
+		downstreamObject, err = c.downstreamClient.Resource(gvr).Namespace(downstreamNamespace).Get(ctx, downstreamName, metav1.GetOptions{})
```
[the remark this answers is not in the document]
bipuladh: using namespace lister.

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	if upstreamLocator.SyncTarget.UID != c.syncTargetUID || upstreamLocator.SyncTarget.Workspace != c.syncTargetWorkspace.String() {
+		return nil
+	}
+
+	upstreamNamespace := upstreamLocator.Namespace
+	upstreamWorkspace := upstreamLocator.Workspace
+
+	upstreamResource, err := c.upstreamClient.Cluster(upstreamWorkspace).Resource(gvr).Namespace(upstreamNamespace).Get(ctx, u.GetName(), metav1.GetOptions{})
```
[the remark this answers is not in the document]
bipuladh: Using lister

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,183 @@
...
+	resourceVersionUpstream := ""
+	if k8serror.IsNotFound(err) {
+		// Resource doesn't exist upstream
+		logger.Info("Creating resource upstream", upstreamNamespace, upstreamWorkspace, u)
+		c.createResourceInUpstream(ctx, gvr, upstreamNamespace, upstreamWorkspace, u)
```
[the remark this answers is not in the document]
bipuladh: updated

pkg/syncer/upsync/upsync_controller.go
```diff
@@ -0,0 +1,224 @@
...
+		return false
+	}
+	qk := key.(queueKey)
+	ks := key.(Keysource)
+	keySource := "Downstream"
+	if ks == Upstream {
+		keySource = "Upstream"
+	}
+
+	logger := logging.WithQueueKey(klog.FromContext(ctx), qk.key).WithValues("gvr", qk.gvr.String(), "Key for:", keySource)
```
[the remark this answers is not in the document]
bipuladh: updated

pkg/syncer/upsync/upsync_process.go
bipuladh: updated!

pkg/syncer/upsync/upsync_controller.go
```diff
@@ -0,0 +1,294 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+)
+
+const (
+	controllerName = "kcp-resource-upsycner"
```
lionelvillard, line 47: nit: kcp-resource-upsycner -> kcp-resource-upsyncer

pkg/syncer/upsync/upsync_controller.go
```diff
@@ -0,0 +1,294 @@
+/*
+Copyright 2022 The KCP Authors.
```
lionelvillard, line 2: new files should have 2023.

davidfestal: Done

pkg/syncer/upsync/upsync_process_test.go
```diff
@@ -301,6 +301,74 @@ func TestUpsyncerprocess(t *testing.T) {
...
+			resourceToProcessName: "test-pod",
+			syncTargetName:        "us-west1",
+			expectActionsOnFrom:   []clienttesting.Action{},
+			expectActionsOnTo: []kcptesting.Action{
+				kcptesting.NewDeleteAction(schema.GroupVersionResource{Group: "", Version: "v1", Resource: "pods"}, logicalcluster.NewPath("root:org:ws"), "test", "test-pod"),
+			},
+			isUpstream:    false,
+			includeStatus: false,
+		},
+		"Upsyncer handles deletion of  upstream resources when they have finalizers set": {
```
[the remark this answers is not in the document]
davidfestal: Seems to be just copy paste. I cleaned it up.

pkg/syncer/spec/spec_controller.go
```diff
@@ -175,6 +176,20 @@ func NewSpecSyncer(syncerLogger logr.Logger, syncTargetClusterName logicalcluste
...
+				if d, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+					obj = d.Obj
+				}
+				var unstrObj *unstructured.Unstructured
+				if unstr, ok := obj.(*unstructured.Unstructured); !ok {
+					utilruntime.HandleError(fmt.Errorf("resource should be a *unstructured.Unstructured, but was %T", unstr))
+					return
+				} else {
+					unstrObj = unstr
+				}
```
sttts, line 188:
```suggestion
				unstrObj, ok := obj.(*unstructured.Unstructured)
				if !ok {
					utilruntime.HandleError(fmt.Errorf("resource should be a *unstructured.Unstructured, but was %T", unstr))
					return
				}
```

pkg/syncer/status/status_controller.go
```diff
@@ -116,6 +117,17 @@ func NewStatusSyncer(syncerLogger logr.Logger, syncTargetClusterName logicalclus
 				if gvr == namespaceGVR {
 					return
 				}
+				var unstrObj *unstructured.Unstructured
+				if unstr, ok := obj.(*unstructured.Unstructured); !ok {
+					runtime.HandleError(fmt.Errorf("resource should be a *unstructured.Unstructured, but was %T", unstr))
+					return
+				} else {
+					unstrObj = unstr
+				}
```
sttts, line 126: as above

pkg/syncer/status/status_controller.go
```diff
@@ -133,6 +147,20 @@ func NewStatusSyncer(syncerLogger logr.Logger, syncTargetClusterName logicalclus
...
+				if d, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+					obj = d.Obj
+				}
+				var unstrObj *unstructured.Unstructured
+				if unstr, ok := obj.(*unstructured.Unstructured); !ok {
+					runtime.HandleError(fmt.Errorf("resource should be a *unstructured.Unstructured, but was %T", unstr))
+					return
+				} else {
+					unstrObj = unstr
+				}
```
sttts, line 159: same as above

pkg/syncer/status/status_process.go
```diff
@@ -164,6 +164,11 @@ func (c *Controller) process(ctx context.Context, gvr schema.GroupVersionResourc
 	if !ok {
 		return fmt.Errorf("object to synchronize is expected to be Unstructured, but is %T", obj)
 	}
+	if u.GetLabels()[workloadv1alpha1.ClusterResourceStateLabelPrefix+c.syncTargetKey] == string(workloadv1alpha1.ResourceStateUpsync) {
+		logger.Info("do not update the status in upstream, since the downstreal resource is in Upsync mode")
```
sttts, line 168: verbosity. Usually for "not doing X" we use V(3) or V(4)

pkg/syncer/syncer.go
```diff
@@ -291,6 +292,12 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 		return err
 	}
 
+	logger.Info("Creating resource upsyncer")
```
sttts, line 295: does this need log output?

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,365 @@
...
+	"k8s.io/klog/v2"
+
+	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
+	. "github.com/kcp-dev/kcp/pkg/logging"
+	"github.com/kcp-dev/kcp/pkg/syncer/shared"
+	. "github.com/kcp-dev/kcp/tmc/pkg/logging"
+)
+
+const (
+	ResourceVersionAnnotation = "kcp.io/resource-version"
```
sttts, line 42: what is this?

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,365 @@
...
+			//
+			// TODO(davidfestal): the call below won't work afaict since  the key is a downstream one, not an upstream one.
+			// And in any case we don't have the locator to find out in where the upstream reosurce should be deleted.
+			//
+			// To manage this case we should catch downstream namespace delete events, in the resource handler,
+			// and get the upstream namespace (from the locator there, to then submit the deletion of the upstream value from there)
+			//
+			//			return c.pruneUpstreamResource(ctx, gvr, key)
+			//
+			logger.Error(err, "the downstream namespace doesn't exist anymore.")
```
sttts, line 140: why is this an error?

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,365 @@
...
+		nsObj, err := downstreamNamespaceLister.Get(downstreamNamespace)
+		if k8serror.IsNotFound(err) {
+			// Since the namespace is already deleted downstream we can delete the resource upstream as well
+			//
+			// TODO(davidfestal): the call below won't work afaict since  the key is a downstream one, not an upstream one.
+			// And in any case we don't have the locator to find out in where the upstream reosurce should be deleted.
+			//
+			// To manage this case we should catch downstream namespace delete events, in the resource handler,
+			// and get the upstream namespace (from the locator there, to then submit the deletion of the upstream value from there)
+			//
```
sttts, line 137: what's the impact of this TODO?

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,365 @@
...
+	if downstreamNamespace != "" {
+		downstreamObject, err = downstreamLister.ByNamespace(downstreamNamespace).Get(downstreamName)
+	} else {
+		downstreamObject, err = downstreamLister.Get(downstreamName)
+	}
+	if err != nil && !k8serror.IsNotFound(err) {
+		return err
+	}
+	downstreamResource, ok := downstreamObject.(*unstructured.Unstructured)
+	if !ok {
```
sttts, line 165: we hit this in the IsNotFound case. Intended?

pkg/syncer/upsync/upsync_controller.go
```diff
@@ -0,0 +1,253 @@
...
+
+// NewUpSyncer returns a new controller which upsyncs, through the Usyncer virtual workspace, downstream resources
+// which are part of the upsyncable resource types (fixed limited list for now), and provide
+// the following labels:
+//   - internal.workload.kcp.io/cluster: <sync target key>
+//   - state.workload.kcp.io/<sync target key>: Upsync
+//
+// and optionally, for cluster-wide resources, the `kcp.io/namespace-locator` annotation
+// filled with the information necessary identify the upstream workspace to upsync to.
+func NewUpSyncer(syncerLogger logr.Logger, syncTargetWorkspace logicalcluster.Name,
```
sttts, line 58: elsewhere we call this ClusterName, not Workspace.

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+
+	if downstreamNamespace == "" {
+		locatorHolder = downstreamResource
+	}
+
+	var upstreamLocator *shared.NamespaceLocator
+	if locatorHolder != nil {
+		if locator, locatorExists, err := shared.LocatorFromAnnotations(locatorHolder.GetAnnotations()); err != nil {
+			logger.Error(err, "error getting downstream locator annotation")
```
sttts, line 176: this is logged anyway by returning it

pkg/syncer/upsync/upsync_process.go
```diff
@@ -0,0 +1,365 @@
...
+			preparedResource.SetAnnotations(addResourceVersionAnnotation(resourceVersionDownstream, preparedResource.GetAnnotations()))
+			preparedResource.SetResourceVersion(updatedResource.GetResourceVersion())
+			_, err = c.upstreamClient.Resource(gvr).Cluster(upstreamWorkspace.Path()).Namespace(upstreamNamespace).Update(ctx, preparedResource, metav1.UpdateOptions{})
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	if downstreamResource.GetDeletionTimestamp() != nil {
```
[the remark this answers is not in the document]
sttts: so I guess in line 330 that's the finalizer? When is it removed?

pkg/syncer/syncer.go
davidfestal: We have messages for the other 2 controllers. Why not be consistent ?
These are only output at the start of the Syncer anyway.

pkg/syncer/spec/spec_controller.go
davidfestal: Fixed

pkg/syncer/status/status_controller.go
davidfestal: Fixed

pkg/syncer/status/status_controller.go
davidfestal: Fixed

pkg/syncer/status/status_process.go
davidfestal: Fixed

pkg/syncer/upsync/upsync_controller.go
davidfestal: Fixed

pkg/syncer/upsync/upsync_process.go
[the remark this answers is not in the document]
davidfestal: Yes, the finalizer is added either at creation or at update of the upstream resource.
This second update is to update the resourceVersion annotation only, after both the main reosurce and the status (if needed) have been updated.

The finalizer is removed when the `removeUpstreamResource()` function is called with the `force` argument to `true`. So mainly when the downstream resource is effectively deleted.

davidfestal:
> Main comment: this controller should be two. It would make the controller follow standard patterns.

The watching of upstream events is really an edge case. As you can see, it only watches Add events, and this is mainly only useful to do some cleanup at Syncer restart if some downstream objects were deleted while the Syncer was off, and the upsynced upstream resources are still there and should be cleaned.
No more logic than that.

Does it make sense to create a dedicated controller for that ? Would it make the code cleaner and easier to understand ?I'm dubious about it. I assume we can structure things even better here to avoid the feeling of a non-standard controller, and also comment more.

BTW the same type of situation (watch the other direction for some edge case) exists in the SpecSyncer, this time to manage deletion from downstream: https://github.com/kcp-dev/kcp/blob/main/pkg/syncer/spec/spec_controller.go#L174

pkg/syncer/upsync/upsync_process.go
davidfestal: In full generality it is an error since we cannot cleanup the related synced upstream resource since we don't have the namespace locator to know in which upstream workspace/namespace to remove it.

davidfestal: This might happen when a downstream resource is removed and the containing namespace are both deleted before the reconcilation happens.

pkg/syncer/upsync/upsync_process.go
davidfestal: see answer below.

davidfestal: I also updated the comment.

sttts:
> Does it make sense to create a dedicated controller for that ? 

Definitely.

pkg/syncer/upsync/upsync_controller.go
```diff
@@ -0,0 +1,437 @@
...
+		utilruntime.HandleError(err)
+		return false, nil
+	}
+
+	dirtyStatus := false
+	if dirtyStatusObj, found := c.dirtyStatusKeys.LoadAndDelete(queueKey{
+		gvr: gvr,
+		key: key,
+	}); found {
+		dirtyStatus = dirtyStatusObj.(bool)
```
sttts, line 401: I think on error of the reconciler you should mark the key dirty again if it had been dirty before

pkg/syncer/upsync/upsync_controller.go
```diff
@@ -0,0 +1,437 @@
...
+		getter = upstreamLister.ByCluster(clusterName).ByNamespace(namespace).Get
+	}
+
+	upstreamObj, err := getter(name)
+	if err != nil && !errors.IsNotFound(err) {
+		return false, err
+	}
+
+	var upstreamResource *unstructured.Unstructured
+	var ok bool
```
sttts, line 419: move var as near as possible to where it is used

pkg/syncer/upsync/upsync_controller.go
```diff
@@ -0,0 +1,437 @@
...
+			new, ok := obj.(*unstructured.Unstructured)
+			if !ok {
+				utilruntime.HandleError(fmt.Errorf("resource should be a *unstructured.Unstructured, but was %T", obj))
+				return
+			}
+			if new.GetLabels()[workloadv1alpha1.ClusterResourceStateLabelPrefix+syncTargetKey] != string(workloadv1alpha1.ResourceStateUpsync) {
+				return
+			}
+			status := new.UnstructuredContent()["status"]
+			c.enqueueDownstream(gvr, new, logger, status != nil && !equality.Semantic.DeepEqual(status, emptyMap))
```
sttts, line 140: technically I think an emptyMap could be dirty as well, imagine we have to sync the status because in upstream it is not empty.

pkg/syncer/upsync/upsync_controller.go
```diff
@@ -0,0 +1,437 @@
...
+			c.enqueueUpstream(gvr, upstreamUpsyncedResource, logger, false)
+		}
+	}
+}
+
+func (c *controller) enqueueDownstream(gvr schema.GroupVersionResource, downstreamObj *unstructured.Unstructured, logger logr.Logger, dirtyStatus bool) {
+	downstreamNamespace := downstreamObj.GetNamespace()
+	locatorHolder := downstreamObj
+	if downstreamNamespace != "" {
+		downstreamNamespaceLister, err := c.getDownstreamLister(corev1.SchemeGroupVersion.WithResource("namespaces"))
```
sttts, line 293:
```suggestion
		// get locator from namespace for namespaced objects
		downstreamNamespaceLister, err := c.getDownstreamLister(corev1.SchemeGroupVersion.WithResource("namespaces"))
```

pkg/syncer/upsync/upsync_process_test.go
```diff
@@ -0,0 +1,762 @@
...
+			fromResource: createPod("test-pod", "kcp-33jbiactwhg0", "", map[string]string{
+				"internal.workload.kcp.io/cluster": "6ohB8yeXhwqTQVuBzJRgqcRJTpRjX7yTZu5g5g",
+				workloadv1alpha1.ClusterResourceStateLabelPrefix + "6ohB8yeXhwqTQVuBzJRgqcRJTpRjX7yTZu5g5g": "Upsync",
+			}, nil, nil, "1"),
+			toResource:            nil,
+			resourceToProcessName: "test-pod",
+			syncTargetName:        "us-west1",
+			expectActionsOnFrom:   []clienttesting.Action{},
+			expectActionsOnTo: []kcptesting.Action{
+				kcptesting.NewCreateAction(schema.GroupVersionResource{Group: "", Version: "v1", Resource: "pods"}, logicalcluster.NewPath("root:org:ws"), "test", toUnstructured(t, createPod("test-pod", "test", "", map[string]string{
```
sttts, line 159: a `unmarshalYaml(t, ```...```)` would make this much easier to read.

pkg/syncer/upsync/upsync_reconcile.go
```diff
@@ -0,0 +1,261 @@
...
+	// whether an upsynced upstream resource is up-to-date with the downstream resource.
+	ResourceVersionAnnotation = "workload.kcp.io/rv"
+)
+
+func (c *controller) reconcile(ctx context.Context, upstreamObject *unstructured.Unstructured, gvr schema.GroupVersionResource, upstreamClusterName logicalcluster.Name, upstreamNamespace, upstreamName string, dirtyStatus bool) (bool, error) {
+	logger := klog.FromContext(ctx)
+
+	downstreamNamespace := ""
+	if upstreamNamespace != "" {
+		// retrieve the downstream namespace from the upstream namespace through the namespace locator index
```
sttts, line 53:
```suggestion
		// find downstream namespace through locator index
```

pkg/syncer/upsync/upsync_reconcile.go
```diff
@@ -0,0 +1,261 @@
...
+		namespaceLocator := shared.NewNamespaceLocator(upstreamClusterName, c.syncTargetClusterName, c.syncTargetUID, c.syncTargetName, upstreamNamespace)
+		locatorAnnotationValue, err := json.Marshal(namespaceLocator)
```
sttts, line 55:
```suggestion
		locatorValue, err := json.Marshal(namespaceLocator)
```

pkg/syncer/upsync/upsync_reconcile.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	if err != nil {
+		return false, err
+	}
+
+	downstreamResource, ok := downstreamObject.(*unstructured.Unstructured)
+	if !ok {
+		return false, fmt.Errorf("type mismatch of resource object: received %T", downstreamResource)
+	}
+	resourceVersionDownstream := downstreamResource.GetResourceVersion()
```
sttts, line 109: downstreamRV

pkg/syncer/upsync/upsync_reconcile.go
```diff
@@ -0,0 +1,261 @@
...
+	markedForDeletionDownstream := downstreamResource.GetDeletionTimestamp() != nil
+
+	if upstreamObject == nil && !markedForDeletionDownstream {
```
sttts, line 112: downstreamObject would match upstreamObject

pkg/syncer/upsync/upsync_reconcile.go
```diff
@@ -0,0 +1,261 @@
...
+		// Resource doesn't exist upstream => let's create it upstream unless the corresponding downstream resource has been marked for deletion.
+		logger.Info("Creating resource upstream")
```
sttts, line 114: V(1)

sttts, line 114:
```suggestion
	// (potentially) create object upstream
	if upstreamObject == nil {
		if markedForDeletionDownstream {
			return false, nil
		}
		logger.V(1).Info("Creating resource upstream")
```

pkg/syncer/upsync/upsync_reconcile.go
davidfestal: wdym ?

pkg/syncer/upsync/upsync_reconcile.go
```diff
@@ -0,0 +1,261 @@
...
+		preparedResource.SetAnnotations(addResourceVersionAnnotation(resourceVersionDownstream, preparedResource.GetAnnotations()))
+		preparedResource.SetResourceVersion(updatedResource.GetResourceVersion())
+		_, err = c.upstreamClient.Resource(gvr).Cluster(upstreamClusterName.Path()).Namespace(upstreamNamespace).Update(ctx, preparedResource, metav1.UpdateOptions{})
+		return false, err
+	}
+
+	resourceVersionUpstream := upstreamObject.GetAnnotations()[ResourceVersionAnnotation]
+	if resourceVersionDownstream != resourceVersionUpstream {
+		// Resource version annotation on the upstream resource doesn't match the downstream resource version
+		// => let's update the resource upstream.
```
sttts, line 148:
```suggestion
	// update upstream when annotation RV differs
	resourceVersionUpstream := upstreamObject.GetAnnotations()[ResourceVersionAnnotation]
	if resourceVersionDownstream != resourceVersionUpstream {
```

pkg/syncer/upsync/upsync_reconcile.go
```diff
@@ -0,0 +1,261 @@
...
+		logger.Info("Updating upstream resource")
+		preparedResource := c.prepareResourceForUpstream(ctx, gvr, upstreamNamespace, upstreamClusterName, downstreamResource)
+		if err != nil {
+			return false, err
+		}
+
+		if !dirtyStatus {
+			// if no status needs to be upsynced upstream, then we can set the resource version annotation to the downstream resource version
+			// at the same time as we update the main resource content.
```
sttts, line 157:
```suggestion
		// quick path: status unchanged, only update main resource
		if !dirtyStatus {
```

pkg/syncer/upsync/upsync_reconcile.go
```diff
@@ -0,0 +1,261 @@
...
+			}
+			preparedResource.SetResourceVersion(existingResource.GetResourceVersion())
+			_, err = c.upstreamClient.Resource(gvr).Cluster(upstreamClusterName.Path()).Namespace(upstreamNamespace).Update(ctx, preparedResource, metav1.UpdateOptions{})
+			// If the downstream resource is marked for deletion, let's requeue it to manage the deletion timestamp
+			return markedForDeletionDownstream, err
+		}
+
+		// Status also needs to be upsynced so let's do it in 3 steps:
+		// - update the upstream resource main content
+
```
sttts, line 172:
```suggestion
		// slow path: status changed => we need 3 steps
		// 1. update main resource
```

pkg/syncer/upsync/upsync_reconcile.go
```diff
@@ -0,0 +1,261 @@
...
+		updatedResource, err = c.upstreamClient.Resource(gvr).Cluster(upstreamClusterName.Path()).Namespace(upstreamNamespace).UpdateStatus(ctx, preparedResource, metav1.UpdateOptions{})
+		if err != nil {
+			return false, err
+		}
+
+		// - finally update the main content again to set the resource version annotation to the value of the downstream resource version.
+		preparedResource.SetAnnotations(addResourceVersionAnnotation(resourceVersionDownstream, preparedResource.GetAnnotations()))
+		preparedResource.SetResourceVersion(updatedResource.GetResourceVersion())
+		_, err = c.upstreamClient.Resource(gvr).Cluster(upstreamClusterName.Path()).Namespace(upstreamNamespace).Update(ctx, preparedResource, metav1.UpdateOptions{})
+		// If the downstream resource is marked for deletion, let's requeue it to manage the deletion timestamp
```
sttts, line 195: why not fall through?

pkg/syncer/upsync/upsync_reconcile.go
```diff
@@ -0,0 +1,261 @@
...
+			} else if err != nil {
+				return err
+			}
+		}
+	}
+	if err := c.upstreamClient.Resource(gvr).Cluster(clusterName.Path()).Namespace(namespace).Delete(ctx, name, metav1.DeleteOptions{}); k8serror.IsNotFound(err) {
+		return nil
+	} else {
+		return err
+	}
```
sttts, line 252:
```suggestion
return c.upstreamClient.Resource(gvr).Cluster(clusterName.Path()).Namespace(namespace).Delete(ctx, name, metav1.DeleteOptions{})
```

pkg/syncer/upsync/upsync_reconcile.go
```diff
@@ -0,0 +1,261 @@
...
+	resourceToUpsync.SetManagedFields(nil)
+	resourceToUpsync.SetDeletionTimestamp(nil)
+	resourceToUpsync.SetDeletionGracePeriodSeconds(nil)
+	resourceToUpsync.SetOwnerReferences(nil)
+	resourceToUpsync.SetFinalizers([]string{shared.SyncerFinalizerNamePrefix + c.syncTargetKey})
+
+	return resourceToUpsync
+}
+
+func (c *controller) removeUpstreamResource(ctx context.Context, gvr schema.GroupVersionResource, clusterName logicalcluster.Name, namespace, name string, force bool) error {
```
sttts, line 231: deleteUpstreamResource ? delete is the term in kube.

pkg/syncer/upsync/upsync_reconcile.go
sttts: and move IsNotFound outside where this is called

pkg/syncer/upsync/upsync_controller.go
davidfestal: Done

pkg/syncer/upsync/upsync_controller.go
davidfestal: Fixed in commit fffff6927c99c86702c7709881a77057331950cd

pkg/syncer/upsync/upsync_controller.go
davidfestal: Fixed in commit fffff6927c99c86702c7709881a77057331950cd

pkg/syncer/upsync/upsync_reconcile.go
davidfestal: Fixed  in commit fffff6927c99c86702c7709881a77057331950cd

pkg/syncer/upsync/upsync_reconcile.go
```diff
@@ -0,0 +1,261 @@
...
+	ResourceVersionAnnotation = "workload.kcp.io/rv"
+)
+
+func (c *controller) reconcile(ctx context.Context, upstreamObject *unstructured.Unstructured, gvr schema.GroupVersionResource, upstreamClusterName logicalcluster.Name, upstreamNamespace, upstreamName string, dirtyStatus bool) (bool, error) {
+	logger := klog.FromContext(ctx)
+
+	downstreamNamespace := ""
+	if upstreamNamespace != "" {
+		// retrieve the downstream namespace from the upstream namespace through the namespace locator index
+		namespaceLocator := shared.NewNamespaceLocator(upstreamClusterName, c.syncTargetClusterName, c.syncTargetUID, c.syncTargetName, upstreamNamespace)
```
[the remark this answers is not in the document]
davidfestal: Fixed  in commit fffff6927c99c86702c7709881a77057331950cd

pkg/syncer/upsync/upsync_reconcile.go
davidfestal: Fixed  in commit fffff6927c99c86702c7709881a77057331950cd

pkg/syncer/upsync/upsync_reconcile.go
davidfestal: Fixed in commit fffff6927c99c86702c7709881a77057331950cd

pkg/syncer/upsync/upsync_reconcile.go
davidfestal: Fixed in commit fffff6927c99c86702c7709881a77057331950cd

davidfestal: Fixed

pkg/syncer/upsync/upsync_reconcile.go
davidfestal: Fixed in commit fffff6927c99c86702c7709881a77057331950cd

pkg/syncer/upsync/upsync_reconcile.go
davidfestal: Fixed in commit fffff6927c99c86702c7709881a77057331950cd

pkg/syncer/upsync/upsync_reconcile.go
davidfestal: Fixed in commit fffff6927c99c86702c7709881a77057331950cd

pkg/syncer/upsync/upsync_reconcile.go
davidfestal: Fixed in commit fffff6927c99c86702c7709881a77057331950cd

pkg/syncer/upsync/upsync_controller.go
davidfestal: I agree on a technical POV, but OTOH we don't expect and manage the case where the status (or more generally the content of the upstream resource) would be changed or removed upstream, since upsynced resources are expected to be completely owned by downstream (this will be enforced through an admission controller at the KCP level at some point).
So I'm not sure it really makes sense to take this case into account.

pkg/syncer/upsync/upsync_reconcile.go
davidfestal: It was so at start. But stopping there and using requeue avoid having a bit `if/else` block for both cases (quick path / slow path): we return directly at the end of the quick path and requeue.
So I kept the "return with requeue" at the end of the slow path for consistency.
But I can move back to the previous structure, sure.

pkg/syncer/upsync/upsync_process_test.go
davidfestal: I made a try, but I'm not convinced it makes it so better.
Could we keep that for a distinct PR where we would would change the unit tests of all the controller: Upsyncer, SpecSyncer, StatsSyncer, which all follow the same structure ?

davidfestal: After our dicsussion, I fixed it by refactoring the test to use a small dsl, in order to both increase readability while avoiding more verbosity: see commit 80a228ba11dc7c931ecb334c327e8d9fc7653278

test/e2e/syncer/tunnels_test.go
```diff
@@ -249,21 +262,21 @@ func TestSyncerTunnel(t *testing.T) {
...
-		pods, err := userKcpClient.Cluster(userWsPath).CoreV1().Pods(upstreamNamespaceName).List(ctx, metav1.ListOptions{})
+		pods, err = userKcpClient.Cluster(userWsPath).CoreV1().Pods(upstreamNamespaceName).List(ctx, metav1.ListOptions{
+			LabelSelector: "foo=bar",
+		})
 		if apierrors.IsUnauthorized(err) {
 			return false, fmt.Sprintf("failed to list pods: %v", err)
 		}
 		require.NoError(t, err)
 
+		return pods != nil && len(pods.Items) > 0, "upsynced pods no found"
```
mamachanko, line 275:
```suggestion
		return pods != nil && len(pods.Items) > 0, "upsynced pods not found"
```

pkg/syncer/upsync/upsync_controller.go
sttts: it's unnecessarily brittle now, even has extra code here for that. Why not simple and more reliable? What is the downside? Also note that this is not about modifcations upstream. IMO the code now is just wrong. A downstream objects can get an empty status. You must sync that.

pkg/syncer/upsync/upsync_process_test.go
```diff
@@ -0,0 +1,940 @@
...
+		expectActionsOnDownstream []clienttesting.Action
+		expectActionsOnUpstream   []kcptesting.Action
+		includeStatus             bool
+	}
+	tests := map[string]testCase{
+		"Upsyncer upsyncs namespaced resources": {
+			upstreamLogicalCluster: "root:org:ws",
+			upstreamNamespaceName:  "test",
+			downstreamNamespace: namespace("kcp-33jbiactwhg0").
+				WithLabels(map[string]string{
```
sttts, line 135: much better to read!

pkg/syncer/upsync/upsync_reconcile.go
```diff
@@ -0,0 +1,264 @@
...
+		preparedResource.SetAnnotations(addResourceVersionAnnotation(downstreamRV, preparedResource.GetAnnotations()))
+		preparedResource.SetResourceVersion(updatedResource.GetResourceVersion())
+		_, err = c.upstreamClient.Resource(gvr).Cluster(upstreamClusterName.Path()).Namespace(upstreamNamespace).Update(ctx, preparedResource, metav1.UpdateOptions{})
+		return false, err
+	}
+
+	// update upstream when annotation RV differs
+	resourceVersionUpstream := upstreamObject.GetAnnotations()[ResourceVersionAnnotation]
+	if downstreamRV != resourceVersionUpstream {
+		logger.Info("Updating upstream resource")
```
sttts, line 157: V(1)

pkg/syncer/upsync/upsync_reconcile.go
sttts: k

pkg/syncer/upsync/upsync_controller.go
```diff
@@ -0,0 +1,450 @@
...
+	"github.com/kcp-dev/kcp/pkg/logging"
+	syncerindexers "github.com/kcp-dev/kcp/pkg/syncer/indexers"
+	"github.com/kcp-dev/kcp/pkg/syncer/shared"
+)
+
+const controllerName = "kcp-resource-upsyncer"
+
+var namespaceGVR schema.GroupVersionResource = corev1.SchemeGroupVersion.WithResource("namespaces")
+
+// NewUpSyncer returns a new controller which upsyncs, through the Usyncer virtual workspace, downstream resources
```
mamachanko, line 59:
```suggestion
// NewUpSyncer returns a new controller which upsyncs, through the Upsyncer virtual workspace, downstream resources
```

pkg/syncer/upsync/upsync_controller.go
```diff
@@ -0,0 +1,450 @@
...
+// the following labels:
+//   - internal.workload.kcp.io/cluster: <sync target key>
+//   - state.workload.kcp.io/<sync target key>: Upsync
+//
+// and optionally, for cluster-wide resources, the `kcp.io/namespace-locator` annotation
+// filled with the information necessary identify the upstream workspace to upsync to.
+func NewUpSyncer(syncerLogger logr.Logger, syncTargetClusterName logicalcluster.Name,
+	syncTargetName, syncTargetKey string,
+	upstreamClient kcpdynamic.ClusterInterface, downstreamClient dynamic.Interface,
+	ddsifForUpstreamUpyncer *ddsif.DiscoveringDynamicSharedInformerFactory,
```
mamachanko, line 70:
```suggestion
	ddsifForUpstreamUpsyncer *ddsif.DiscoveringDynamicSharedInformerFactory,
```

pkg/syncer/upsync/upsync_controller.go
```diff
@@ -0,0 +1,450 @@
...
+	syncerindexers "github.com/kcp-dev/kcp/pkg/syncer/indexers"
+	"github.com/kcp-dev/kcp/pkg/syncer/shared"
+)
+
+const controllerName = "kcp-resource-upsyncer"
+
+var namespaceGVR schema.GroupVersionResource = corev1.SchemeGroupVersion.WithResource("namespaces")
+
+// NewUpSyncer returns a new controller which upsyncs, through the Usyncer virtual workspace, downstream resources
+// which are part of the upsyncable resource types (fixed limited list for now), and provide
```
mamachanko, line 60: **question:** What are the upsyncable resource types?

mamachanko: I _think_ upsyncable resources are limited to `pods`, `endpoints` and `persistentvolumeclaims`. Both experimentation and code support my understanding:

https://github.com/bipuladh/kcp/blob/6cf3293ff35c3558e032c0cd3e4a00724c00e9ec/pkg/syncer/syncer.go#L236-L245

pkg/syncer/upsync/upsync_process_test.go
davidfestal: Thanks !

pkg/syncer/upsync/upsync_controller.go
davidfestal: That's right, for now only these 3 resources are upsynced, since they correspond to the use-cases forseeable short-term.
Mid-term / long-term, we should have followup work that would allow customizing this list at the KCP level (and possibly storing it to the Synctarget status), but this still needs further design and discussion, especially to keep security in mind.

We should probably open a dedicated issue for this, to track the first thoughts we had in the regard, and further discuss the details.

sttts: The upsync controller is agnostic which GVRs are upsynced. But we are very opinionated in the wiring into the syncer for the moment. Whether we open that up eventually in some way, we will see.

mamachanko: Thanks @davidfestal @sttts 

> We should probably open a dedicated issue for this, to track the first thoughts we had in the regard, and further discuss the details.

I can take care of that.

pkg/syncer/upsync/upsync_controller.go
davidfestal: I rolled back to syncing empty statuses.

pkg/syncer/upsync/upsync_reconcile.go
davidfestal: Fixed

pkg/syncer/upsync/upsync_controller.go
davidfestal: Fixed

pkg/syncer/upsync/upsync_controller.go
davidfestal: Fixed

pkg/syncer/upsync/upsync_controller.go
davidfestal: I created the related issue here: https://github.com/kcp-dev/kcp/issues/2804

Let's discuss in this new issue.

pdettori: Thanks ! I will copy the Crossplane use case description there.

mamachanko: Thanks for creating the issue @davidfestal you were faster than me 🙇🏻

Issue #1937, opened 2022-09-09
feature: Specify Which Objects Of A Resource A PermissionClaim Applies To


stevekuznetsov:
### Feature Description

Today, and API provider can only request permissions on all objects of a given GR. Similarly, the future reverse permission claims in #1840 will have the same limitation. We need to allow both of these mechanisms to select some small subset of objects.

### Proposed Solution

As an API provider,
I want to specify exactly which objects I need access to in a user's workspace,
So that I can minimize my permissions footprint in their workspace

### Want to contribute?

- [ ] I would like to work on this issue.

### Additional Context

We need to do a user survey to determine the best approaches here. Obvious selection mechanisms are:
- object names
- label selectors
- CEL or JSONPath

However, other options are also possible - "all Secrets that are mounted in a Pod" - but we need to understand better the use-cases of service providers to know if we need to provide those options. A minimal, extensible solution may suffice for this card.

nrb:
A draft question for the survey:

> If you are operating a service in a user's workspace, what access do you need to objects from other API objects?
>
> For example, "I need to manage all `Deployments` for `<label selector>` and all `Secrets` mounted into pods for `Deployments` I create"

I think that covers what we're looking for, but before I send something out I'd like input from @sttts and @ncdc to see if there's anything else we should ask.

maleck13:
I need to manage secrets with tls secrets in them that are referenced from an ingress (they also share a label and annotations)

stuartwdouglas:
For the JVM build service we need to be able to
- Create a deployment with a specific name (this may be handled by an external service, although if we can have create/edit permission just for a single object that might give us a lot more flexibility and allow our operator to handle it rather than an external service)
- Some named config maps and secrets (ATM jvm-build-service-filter and jvm-build-secrets)

We also have some 'singleton' objects, where we expect only a single one to be present with a well known name (e.g. UserConfig), so it would be nice to be able to enforce this via permissions.

astefanutti:
Our controller needs to create and update different kind of resources, like Ingress, Route, KnativeService, KEDA's ScaleObject etc., but should only be able to update / patch / delete these resources if it "owns" them.

These resources have one of our controller API resource in their owning references ancestors, so this would be a way to express the permission claim subset.

Two corollaries:
* This generalises to "grant update / patch / delete permissions to the resource creator", but that may be more difficult to express or implement (creator identity is not currently persisted);
* This could be approximated with label selectors.

[...]

PR #2222, opened 2022-10-19
:seedling: pkg/../types_apiexport: add name, namespace

s-urbaniak:
## Summary
This adds name,namespace fields to permission claims so we can work independently on reverse permission claim authz and narrowly scoped permissionclaims by @nrb.

## Related issue(s)

Relates to #2089
Relates to #1937


pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -188,6 +188,17 @@ const (
 type PermissionClaim struct {
 	GroupResource `json:","`
 
+	// name of an object within the group/resource.
+	// It matches the metadata.name field of the underlying object.
+	// If unset, all objects having the group/resource are being claimed.
+	// +optional
+	Name string `json:"name,omitempty"`
```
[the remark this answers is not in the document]
s-urbaniak: Indeed, we didn't 🤔 Are double entries allowed with the same group/resources/identityhash?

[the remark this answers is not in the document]
s-urbaniak: @nrb i guess this case is missing in https://docs.google.com/document/d/1IHai7FlVQXPCD_MZObT0CzAkdOFSZuWGokHa_kn72HQ/edit as well.

pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -196,6 +201,19 @@ type PermissionClaim struct {
 	IdentityHash string `json:"identityHash,omitempty"`
 }
 
+type ClaimedResource struct {
+	// name of an object within a claimed group/resource.
+	// It matches the metadata.name field of the underlying object.
+	// If unset, all objects having the group/resource are being claimed.
+	// +optional
+	Name string `json:"name,omitempty"`
```
stevekuznetsov, line 209: Can we (trivially) add validation for these fields to match what an object's name would have to be?

pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -196,6 +201,19 @@ type PermissionClaim struct {
...
+
+	// namespace containing the named object. Matches metadata.namespace field.
+	// If namespace is set but name is not set this field is ignored.
```
sttts, line 212: this is strange. How can I claim a complete namespace?

pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -188,6 +188,11 @@ const (
 type PermissionClaim struct {
 	GroupResource `json:","`
 
+	// claimed is a list of claimed resources.
+	// If omitted all resources of the given group/resource will be claimed.
+	// +optional
+	Claimed []ClaimedResource `json:"claimed,omitempty"`
```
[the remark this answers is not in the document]
sttts: resourceSelector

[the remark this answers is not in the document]
sttts: later:
```
resourceSelector:
  name:
  namespace:
  labels:
  jsonPath:
```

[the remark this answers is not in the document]
sttts: What is default? What do we want to encourage?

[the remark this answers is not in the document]
sttts:
```
resourceSelector:
- name:
  namespace:
  labels:
  jsonPath:
  everything: true // this is requires all others are unset, and only length 1 is allowed for resourceSelector
```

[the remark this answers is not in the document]
sttts: + required for resourceSelector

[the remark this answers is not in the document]
sttts: alternative:
```
resourceSelector: []
everything: true
```

pkg/apis/apis/v1alpha1/types_apiexport.go
s-urbaniak: sorry i am not following 🙈 can you elaborate? do we have restrictions on names?

pkg/apis/apis/v1alpha1/types_apiexport.go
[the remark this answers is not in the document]
nrb: I like `resourceSelector`. I also think that `name/namespace`, `labels`, and `jsonPath` should all be mutually exclusive.

Given the security implications, requiring that `everything` is explicitly defined makes sense to me, too.

pkg/apis/apis/v1alpha1/types_apiexport.go
nrb: As I understand it, it's for scoping named resources to a namespace.

I suppose one could try to claim all instances of the RG in a namespace, but so far I haven't thought of or seen a use case for that.

pkg/apis/apis/v1alpha1/types_apiexport.go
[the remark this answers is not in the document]
nrb: You're right, it is missing.

I could see it being possible in the case of ConfigMaps - a developer may want to only see 2 or 3 ConfigMaps, or perhaps 2-3 secrets.

pkg/apis/apis/v1alpha1/types_apiexport.go
nrb: I think we'd have to ensure we match one or more of these restrictions: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

tl;dr - if we want to put the names in labels or paths, we'll need to make sure that the name provided follows those constraints.

pkg/apis/apis/v1alpha1/types_apiexport.go
s-urbaniak: @sttts it was a suggestion from your side when we did a quick chat. We can remove it if it doesn't make sense though.

pkg/apis/apis/v1alpha1/types_apiexport.go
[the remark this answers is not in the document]
s-urbaniak: @nrb I called the "everything" field `all` for the sake of shortness, wdyt?

pkg/apis/apis/v1alpha1/types_apiexport.go
s-urbaniak: hrm, i believe this would indeed need to be checked during admission. let me see if this can be done declaratively.

stevekuznetsov: If we can't trivially do it via some kube-builder annotation or something for validation, maybe not worth it.

pkg/apis/apis/v1alpha1/types_apiexport.go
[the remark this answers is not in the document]
nrb: Makes sense to me

pkg/apis/apis/v1alpha1/types_apiexport.go
nrb: The closest I can see with the kube-builder annotations (defined [here](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#schemaObject) would be a combo on maxLength and a regex pattern.

[API Machinery has some validation helpers](https://github.com/kubernetes/kubernetes/blob/c18c2447cf602ee1a76113200b3f5b6200a24a34/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go) that appear to be used within Kube itself.

s-urbaniak:
> CI / e2e-sharded - Failing

I cannot reproduce this locally 🤔

stevekuznetsov:
Does it fail 100% in CI?

s-urbaniak:
@stevekuznetsov so far twice but only on github 🤔 A local `ARTIFACT_DIR=$PWD/artifacts UNSAFE_E2E_HACK_DISABLE_ETCD_FSYNC=true E2E_PARALLELISM=8 make test-e2e-sharded WHAT=./test/e2e/virtual/apiexport TEST_ARGS="-run TestAPIExportAPIBindingsAccess -v"` yielded a success.
```

```

pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -197,6 +210,20 @@ type PermissionClaim struct {
...
+	// name of an object within a claimed group/resource.
+	// It matches the metadata.name field of the underlying object.
+	//
+	// +optional
+	Name string `json:"name,omitempty"`
+
+	// namespace containing the named object. Matches metadata.namespace field.
+	// If "name" is unset, all objects from the namespace are being claimed.
+	//
+	// +required
```
nrb, line 223: Is this required if we use a label selector? Or are you intending for `resourceSelector` and label selection to be mutually exclusive?

s-urbaniak: This won't be required if we use a label selector. For now, the constrained declared here is ok but once we add label selection into this struct, I guess we'll need CEL to express mutual exclusivity.

pkg/apis/apis/v1alpha1/types_apiexport.go
s-urbaniak: updated with the referenced regex but fyi: I just found a second ago a precedence for this in our code 😛 https://github.com/kcp-dev/kcp/blob/main/pkg/apis/apis/v1alpha1/types_apiexport.go#L229

pkg/apis/apis/v1alpha1/types_apiexport.go
nrb: Kube has (or had, not up-to-date on the migration) a solution for this on the `VolumeSource` of a PV. [See this comment](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L48).

I'll dig around some more if I can see the exact mechanism, I don't think we need CEL for this unless it's the new way of doing things.

stevekuznetsov:
CI is unhappy, #2278 , below

```
 1 files have incorrect boilerplate headers:
pkg/apis/test/pattern.go 
```

```
 go module files are out of date 
```

stevekuznetsov:
Flake on #2281 
/retest
/lgtm
/approve

pkg/apis/apis/v1alpha1/types_apiexport.go
s-urbaniak: I have precedence of mutual exlusitivity using CEL here: https://github.com/kcp-dev/kcp/pull/2222/files#diff-228a80bf14ca2df5c0d6053a9a4ed5f5ddb5d74a13c2618584741e2723afbacbR190
```
// +kubebuilder:validation:XValidation:rule="self.all != has(self.resourceSelector)",message="either \"all\" or \"resourceSelector\" must be set"
type PermissionClaim struct {
```

nrb: Awesome. I didn't see the `XValidation` marker when looking at kubebuilder markers, but that's perfect.

pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -186,9 +186,22 @@ const (
...
+//
+// +kubebuilder:validation:XValidation:rule="self.all != has(self.resourceSelector)",message="either \"all\" or \"resourceSelector\" must be set"
 type PermissionClaim struct {
 	GroupResource `json:","`
 
+	// all claims all resources for the given group/resource.
+	// This is mutually exclusive with resourceSelector.
+	// +required
+	// +kubebuilder:default=false
+	All bool `json:"all"`
```
sttts, line 198: why not omitempty? IMO having all all the time is ugly.

pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -186,9 +186,22 @@ const (
 // PermissionClaim identifies an object by GR and identity hash.
 // Its purpose is to determine the added permissions that a service provider may
 // request and that a consumer may accept and allow the service provider access to.
+//
+// +kubebuilder:validation:XValidation:rule="self.all != has(self.resourceSelector)",message="either \"all\" or \"resourceSelector\" must be set"
 type PermissionClaim struct {
 	GroupResource `json:","`
 
+	// all claims all resources for the given group/resource.
+	// This is mutually exclusive with resourceSelector.
```
sttts, line 195: where is this checked?

pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -197,6 +210,22 @@ type PermissionClaim struct {
...
+	// +optional
+	// +kubebuilder:validation:Pattern="^([a-z0-9][-a-z0-9_.]*)?[a-z0-9]$"
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name,omitempty"`
+
+	// namespace containing the named object. Matches metadata.namespace field.
+	// If "name" is unset, all objects from the namespace are being claimed.
+	//
+	// +required
+	Namespace string `json:"namespace"`
```
sttts, line 226: why not omit empty?

sttts: and why not optional?

sttts: being required does not extend to more selector types.

pkg/apis/apis/v1alpha1/types_apiexport.go
s-urbaniak: fixed, set `+optional` and `omitemtpy`

s-urbaniak: note: `omitempty` implies `+optional`. I debugged half of a day in controller-tools to find out.

pkg/apis/apis/v1alpha1/types_apiexport.go
s-urbaniak: fixed, however how do we want to validate this, using CEL?

PR #2233, opened 2022-10-19
:bug: Remove leftover comment

fgiloux:
## Summary

Remove a comment referencing a field `status.resourceSchemasInUse`, which does not exist. I also verified that there is no finalizer based mechanism in place that would match the behavior described in the comment.


ncdc (approved):

ncdc:
/retitle 📖 Remove reference to APIExport's status.resourceSchemasInUse (field was removed)

PR #2104, opened 2022-09-29
✨ Switch to cluster-aware k8s clients, listers & informers

stevekuznetsov:

PR #2240, opened 2022-10-20
🌱 cache: add replicateAPIResourceSchema test scenario 

p0lyn0mial:
## Summary
The new test checks if an APIResourceSchema is propagated to the cache server.
The test exercises a creation, modification and removal of the APIExport object.

## Related issue(s)

### DIFF
--- a/test/e2e/reconciler/cache/replication_test.go
+++ b/test/e2e/reconciler/cache/replication_test.go
@@ -62,6 +62,7 @@ type testScenario struct {
 var scenarios = []testScenario{
 	{"TestReplicateAPIExport", replicateAPIExportScenario},
 	{"TestReplicateAPIResourceSchema", replicateAPIResourceSchemaScenario},
+	{"TestReplicateAPIExportNegative", replicateAPIExportNegativeScenario},
 }
 
 // baseScenario an auxiliary struct that is used by replicateResourceScenario
@@ -197,6 +198,50 @@ func replicateResourceScenario(t *testing.T, scenario baseScenario) {
 	}, wait.ForeverTestTimeout, 400*time.Millisecond)
 }
 
+// replicateAPIExportNegativeScenario checks if modified or even deleted cached APIExport will be reconciled to match the original object
+func replicateAPIExportNegativeScenario(ctx context.Context, t *testing.T, server framework.RunningServer, kcpShardClusterClient clientset.ClusterInterface, cacheKcpClusterClient clientset.ClusterInterface) {
+	org := framework.NewOrganizationFixture(t, server)
+	cluster := framework.NewWorkspaceFixture(t, server, org, framework.WithShardConstraints(tenancyv1alpha1.ShardConstraints{Name: "root"}))
+	export := &apisv1alpha1.APIExport{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "mangodb",
+		},
+	}
+	scenario := baseScenario{
+		server:       server,
+		resourceName: export.Name,
+		resourceKind: "APIExport",
+		getSourceResource: func(cluster logicalcluster.Name) (interface{}, error) {
+			return kcpShardClusterClient.Cluster(cluster).ApisV1alpha1().APIExports().Get(ctx, export.Name, metav1.GetOptions{})
+		},
+		getCachedResource: func(cluster logicalcluster.Name) (interface{}, error) {
+			return cacheKcpClusterClient.Cluster(cluster).ApisV1alpha1().APIExports().Get(cacheclient.WithShardInContext(ctx, shard.New("root")), export.Name, metav1.GetOptions{})
+		},
+	}
+
+	t.Logf("Creating APIExport %s|%s", cluster, export.Name)
+	_, err := kcpShardClusterClient.Cluster(cluster).ApisV1alpha1().APIExports().Create(ctx, export, metav1.CreateOptions{})
+	require.NoError(t, err, "error creating APIExport %s|%s", cluster, export.Name)
+
+	t.Logf("Verify that APIExport %s|%s is propagated to the cached object", cluster, export.Name)
+	verifyResourceExistence(t, scenario, cluster)
+
+	t.Logf("Delete APIExport %s|%s", cluster, export.Name)
+	err = cacheKcpClusterClient.Cluster(cluster).ApisV1alpha1().APIExports().Delete(cacheclient.WithShardInContext(ctx, shard.New("root")), export.Name, metav1.DeleteOptions{})
+	require.NoError(t, err)
+	t.Logf("Verify that APIExport %s|%s is propagated to the cached object after deletion", cluster, export.Name)
+	verifyResourceExistence(t, scenario, cluster)
+
+	t.Logf("Update APIExport %s|%s so that it differs from the original object", cluster, scenario.resourceName)
+	cachedExport, err := cacheKcpClusterClient.Cluster(cluster).ApisV1alpha1().APIExports().Get(cacheclient.WithShardInContext(ctx, shard.New("root")), export.Name, metav1.GetOptions{})
+	require.NoError(t, err)
+	cachedExport.Spec.LatestResourceSchemas = append(cachedExport.Spec.LatestResourceSchemas, "foo")
+	_, err = cacheKcpClusterClient.Cluster(cluster).ApisV1alpha1().APIExports().Update(cacheclient.WithShardInContext(ctx, shard.New("root")), cachedExport, metav1.UpdateOptions{})
+	require.NoError(t, err)
+	t.Logf("Verify that APIExport %s|%s is propaged to the cached object after an update", cluster, export.Name)
+	verifyResourceExistence(t, scenario, cluster)
+}
+
 func verifyResourceExistence(t *testing.T, scenario baseScenario, cluster logicalcluster.Name) {
 	t.Logf("Get %s/%s %s from the root shard and the cache server for comparison", cluster, scenario.resourceName, scenario.resourceKind)
 	framework.Eventually(t, func() (bool, string) {
@@ -207,7 +252,7 @@ func verifyResourceExistence(t *testing.T, scenario baseScenario, cluster logica
 		cachedResource, err := scenario.getCachedResource(cluster)
 		if err != nil {
 			if !errors.IsNotFound(err) {
-				return false, err.Error()
+				return true, err.Error()
 			}
 			return false, err.Error()
 		}
--- a/test/e2e/reconciler/cache/replication_test.go
+++ b/test/e2e/reconciler/cache/replication_test.go
@@ -82,63 +82,39 @@ type baseScenario struct {
 // replicateAPIExportScenario tests if an APIExport is propagated to the cache server.
 // The test exercises creation, modification and removal of the APIExport object.
 func replicateAPIExportScenario(ctx context.Context, t *testing.T, server framework.RunningServer, kcpShardClusterClient clientset.ClusterInterface, cacheKcpClusterClient clientset.ClusterInterface) {
-	org := framework.NewOrganizationFixture(t, server)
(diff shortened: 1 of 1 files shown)


test/e2e/reconciler/cache/replication_test.go
```diff
@@ -60,102 +61,205 @@ type testScenario struct {
...
+		},
+		updateSourceResource: func(cluster logicalcluster.Name, res interface{}) (interface{}, error) {
+			apiResSchema, ok := res.(*apisv1alpha1.APIResourceSchema)
+			if !ok {
+				return nil, fmt.Errorf("%T is not *APIResourceSchema", res)
+			}
+			return kcpShardClusterClient.Cluster(cluster).ApisV1alpha1().APIResourceSchemas().Update(ctx, apiResSchema, metav1.UpdateOptions{})
+		},
+		updateSpecForSourceResource: func(res interface{}) error {
+			// no-op
```
p0lyn0mial, line 105: seems like all fields are immutable, checking . . .

p0lyn0mial: oh, there is an admission enforcing immutability.

stevekuznetsov (commented):
Seems reasonable, but would love to have it land after #2104

p0lyn0mial:
/hold 

will merge after https://github.com/kcp-dev/kcp/pull/2104 lands

nrb:
/unhold
2104 is in

test/e2e/reconciler/cache/replication_test.go
```diff
@@ -60,109 +61,264 @@ type testScenario struct {
...
 		}
 		return true, ""
 	}, wait.ForeverTestTimeout, 400*time.Millisecond)
 }
 
-// TestAllAgainstInProcessCacheServer runs all test scenarios against a cache server that runs with a kcp server
-func TestAllAgainstInProcessCacheServer(t *testing.T) {
+// TestAllReplicationScenariosAgainstInProcessCacheServer runs all test scenarios against a cache server that runs with a kcp server
+func TestAllReplicationScenariosAgainstInProcessCacheServer(t *testing.T) {
 	t.Parallel()
```
stevekuznetsov, line 322: This makes the suite parallel with the rest of the suites, but does not make each test case parallel - is that intentional? Do we want a `t.Parallel()` call in the `t.Run()`?

test/e2e/reconciler/cache/replication_test.go
```diff
@@ -60,109 +61,264 @@ type testScenario struct {
...
+			return false, fmt.Sprintf("replicated %s root/%s/%s is different that the original", scenario.resourceKind, cluster, cachedResourceMeta.GetName())
 		}
 		return true, ""
 	}, wait.ForeverTestTimeout, 400*time.Millisecond)
 }
 
-// TestAllAgainstInProcessCacheServer runs all test scenarios against a cache server that runs with a kcp server
-func TestAllAgainstInProcessCacheServer(t *testing.T) {
+// TestAllReplicationScenariosAgainstInProcessCacheServer runs all test scenarios against a cache server that runs with a kcp server
+func TestAllReplicationScenariosAgainstInProcessCacheServer(t *testing.T) {
```
stevekuznetsov, line 321: can we make these e.g. `TestCacheServerInProcess` and `TestCacheServerStandalone`? there's not real use to explaining that it's all scenarios, and there is nothing except for replication scenarios here

test/e2e/reconciler/cache/replication_test.go
```diff
@@ -60,109 +61,264 @@ type testScenario struct {
 // scenarios all test scenarios that will be run against in-process and standalone cache server
 var scenarios = []testScenario{
 	{"TestReplicateAPIExport", replicateAPIExportScenario},
+	{"TestReplicateAPIResourceSchema", replicateAPIResourceSchemaScenario},
```
stevekuznetsov, line 64: you don't need to format this as a Go test name, much more common to have e.g. `"replicate APIResourceSchema"` which will then turn into `TestCacheServerInProcess/replicate_APIResourceSchema`

test/e2e/reconciler/cache/replication_test.go
```diff
@@ -60,109 +61,264 @@ type testScenario struct {
...
+
+// baseScenario an auxiliary struct that is used by replicateResourceScenario
+type baseScenario struct {
+	server framework.RunningServer
+
+	resourceName string
+	resourceKind string
+
+	createSourceResource        func(logicalcluster.Name) error
+	getSourceResource           func(logicalcluster.Name) (interface{}, error)
```
stevekuznetsov, line 76: I think we talked about this before, but there's nothing more specific that this can be, like `runtime.Object`? and we don't want to use generics, because?

test/e2e/reconciler/cache/replication_test.go
```diff
@@ -60,109 +61,264 @@ type testScenario struct {
...
+			apifixtures.CreateSheriffsSchemaAndExport(ctx, t, cluster, kcpShardClusterClient.Cluster(cluster), "wild.wild.west", "testing replication to the cache server")
+			return nil
+		},
+		getSourceResource: func(cluster logicalcluster.Name) (interface{}, error) {
+			return kcpShardClusterClient.Cluster(cluster).ApisV1alpha1().APIResourceSchemas().Get(ctx, "today.sheriffs.wild.wild.west", metav1.GetOptions{})
+		},
+		updateSourceResource: func(cluster logicalcluster.Name, res interface{}) (interface{}, error) {
+			apiResSchema, ok := res.(*apisv1alpha1.APIResourceSchema)
+			if !ok {
+				return nil, fmt.Errorf("%T is not *APIResourceSchema", res)
```
stevekuznetsov, line 101: With generics you don't need to write this code

test/e2e/reconciler/cache/replication_test.go
```diff
@@ -60,109 +61,264 @@ type testScenario struct {
...
+		}
+		if err != nil {
+			return false, err.Error()
+		}
+		return false, fmt.Sprintf("replicated %s %s/%s wasn't removed", scenario.resourceKind, cluster, scenario.resourceName)
+	}, wait.ForeverTestTimeout, 400*time.Millisecond)
+}
+
+// replicateAPIExportNegativeScenario checks if modified or even deleted cached APIExport will be reconciled to match the original object
+func replicateAPIExportNegativeScenario(ctx context.Context, t *testing.T, server framework.RunningServer, kcpShardClusterClient clientset.ClusterInterface, cacheKcpClusterClient clientset.ClusterInterface) {
```
stevekuznetsov, line 202: Why is this APIExport-specific?

test/e2e/reconciler/cache/replication_test.go
```diff
@@ -60,109 +61,264 @@ type testScenario struct {
...
+			return false, err.Error()
+		}
+		if _, found := cachedResourceMeta.GetAnnotations()[genericapirequest.AnnotationKey]; !found {
+			t.Fatalf("replicated %s root|%s/%s, doesn't have %s annotation", scenario.resourceKind, cluster, cachedResourceMeta.GetName(), genericapirequest.AnnotationKey)
 		}
-		delete(cachedWildAPIExport.Annotations, genericapirequest.AnnotationKey)
-		if diff := cmp.Diff(cachedWildAPIExport, wildAPIExport, cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ResourceVersion")); len(diff) > 0 {
-			return false, fmt.Sprintf("replicated APIExport root/%s/%s is different that the original", cluster, wildAPIExport.Name)
+		delete(cachedResourceMeta.GetAnnotations(), genericapirequest.AnnotationKey)
+		if diff := cmp.Diff(cachedResource, originalResource, cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ResourceVersion")); len(diff) > 0 {
```
stevekuznetsov, line 268: Love to see cmpopts - nice! I think you could likely get the annotation key in there too if you cared

test/e2e/reconciler/cache/replication_test.go
```diff
@@ -60,109 +61,264 @@ type testScenario struct {
...
+			return false, fmt.Sprintf("replicated %s root|%s/%s is different that the original", scenario.resourceKind, cluster, cachedResourceMeta.GetName())
 		}
 		return true, ""
 	}, wait.ForeverTestTimeout, 400*time.Millisecond)
```
stevekuznetsov, line 272: How did we arrive at this poll period?

test/e2e/reconciler/cache/replication_test.go
p0lyn0mial: sure, np

test/e2e/reconciler/cache/replication_test.go
p0lyn0mial: in this particular case, we don't need strong types, we operate on ` meta.Accessor`

test/e2e/reconciler/cache/replication_test.go
p0lyn0mial: yeah, this is the only case that would benefit from using generics, I can look into that in a follow-up PR.

test/e2e/reconciler/cache/replication_test.go
p0lyn0mial: I tried but couldn't make it to work on the annotation key

test/e2e/reconciler/cache/replication_test.go
p0lyn0mial: just a random number

test/e2e/reconciler/cache/replication_test.go
p0lyn0mial: added the case for `APIResourceSchema`

test/e2e/reconciler/cache/replication_test.go
p0lyn0mial: I don't think so, too much concurrency can negatively impact a CPU. 
There is no difference on my local machine: `53`s vs `52`s

test/e2e/reconciler/cache/replication_test.go
stevekuznetsov: Can we use `meta.Accessor` then?

test/e2e/reconciler/cache/replication_test.go
stevekuznetsov: Let's use 100ms like the rest of the tests unless we really need to change it - for folks with fast workstations, 300ms extra poll time might double or triple test runtime :)

p0lyn0mial: > for folks with fast workstations, 300ms extra poll time might double or triple test runtime :)

that's why you want to use `t.Parallel` everywhere :)

test/e2e/reconciler/cache/replication_test.go
p0lyn0mial: sure, will change it.

p0lyn0mial: changed to `runtime.Object` because I've realized we need full object for comparison.

test/e2e/reconciler/cache/replication_test.go
stevekuznetsov: Hey, what takes 65 minutes in GitHub Actions takes 54seconds on my machine ... ;)

test/e2e/reconciler/cache/replication_test.go
```diff
@@ -60,109 +64,372 @@ type testScenario struct {
 // scenarios all test scenarios that will be run against in-process and standalone cache server
 var scenarios = []testScenario{
 	{"TestReplicateAPIExport", replicateAPIExportScenario},
+	{"TestReplicateAPIExportNegative", replicateAPIExportNegativeScenario},
+	{"TestReplicateAPIResourceSchema", replicateAPIResourceSchemaScenario},
+	{"TestReplicateAPIResourceSchemaNegative", replicateAPIResourceSchemaNegativeScenario},
+}
+
+// baseScenario an auxiliary struct that is used by replicateResourceScenario
+type baseScenario struct {
```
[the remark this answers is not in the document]
p0lyn0mial: updated to use the builder pattern, ptal

PR #2255, opened 2022-10-24
📖 Updated README.md links

eranco74:
## Summary
The links to `additional documentation` (Concepts, Workspaces, etc) are broken, this PR fixes the links path

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,13 +153,13 @@ Thanks for checking out our quickstart!
 If you're interested in learning more about all the features kcp has to offer, please check out our additional
 documentation:
 
-- [Concepts](docs/concepts.md) - a high level overview of kcp concepts
-- [Workspaces](docs/workspaces.md) - a more thorough introduction on kcp's workspaces
-- [Locations & scheduling](docs/locations-and-scheduling.md) - details on kcp's primitives that abstract over clusters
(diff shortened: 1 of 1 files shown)


avinal:
Ah indeed missed 😅. Now that our documentation site is up, I was thinking about putting links to site pages instead of markdowns. WDYT @hardys? 

In case you haven't checked the site yet, please do https://docs kcp.io/kcp/main

hardys:
> Ah indeed missed sweat_smile. Now that our documentation site is up, I was thinking about putting links to site pages instead of markdowns. WDYT @hardys?

My only concern with that is we seem to only publish docs for `main` to the site, so links in release branches will point to `main` docs, not the (potentially different) docs in each release branch.  

Leaving the links to the markdown seems like the simplest solution unless there is a plan to publish per-release docs on the site?

hardys:
@eranco74 the PR verifier is failing on [this check](https://github.com/kubernetes-sigs/kubebuilder-release-tools/blob/master/verify/main.go#L38) - I think you'll have to make the PR description a bit more verbose

avinal:
I have added the versioned docs mechanism in GitHub Actions. Since this is our first ever docs site so we have only the `main` ref. But with the next release we will have more versions.

PR #2256, opened 2022-10-24
🌱 cache: add e2e scenarios for testing behaviour of the cache server

p0lyn0mial:
## Summary
adds the following e2e scenarios for testing behaviour of the cache server:

1. `TestSchemaIsNotEnforced` - for making sure an obj with any schema can be created
2. `TestShardClusterNamesAssigned` - for making sure that a shard name and a cluster name are passed in the annotations
3. `TestUIDGenerationCreationTimeOverwrite` - for making sure that `UID, Genaration, CreationTime` can be overwritten 
4. `TestUIDGenerationCreationTimeOverwriteNegative`  - for making sure that `UID, Genaration, CreationTime` are assigned by the server if the shard annotation is not provided
5. `TestGenerationOnSpecChanges` - for making sure that the Generation is not increased after updating the spec
6. `TestDeletionWithFinalizers` - for making sure that an object with some `Finalizers` can be deleted immediately
7. `TestUpdatingSpecStatusSimultaneously` - for making sure that both spec and status can be updated at the same time (single request)

## Related issue(s)

### DIFF
--- a/test/e2e/cache/cache_server_test.go
+++ b/test/e2e/cache/cache_server_test.go
@@ -19,10 +19,7 @@ package cache
 import (
 	"context"
 	"encoding/json"
-	"fmt"
-	"sync"
 	"testing"
-	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/kcp-dev/logicalcluster/v2"
@@ -33,10 +30,8 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/util/wait"
 	genericrequest "k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/client-go/dynamic"
-	kubernetesscheme "k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 
@@ -399,52 +394,10 @@ func TestCacheServerAllScenarios(t *testing.T) {
 	cacheClientRestConfig, err := cacheClientConfig.ClientConfig()
 	require.NoError(t, err)
 
-	// TODO: p0lyn0mial: move to StartStandaloneCacheServer
-	t.Logf("waiting for the cache server at %v to become ready", cacheClientRestConfig.Host)
-	waitUntilCacheServerIsReady(ctx, t, cacheClientRestConfig)
-
 	cacheClientRT := cache2e.CacheClientRoundTrippersFor(cacheClientRestConfig)
 	for _, scenario := range scenarios {
 		t.Run(scenario.name, func(tt *testing.T) {
 			scenario.work(ctx, tt, cacheClientRT, logicalcluster.New("acme"), schema.GroupVersionResource{Group: "apis.kcp.dev", Version: "v1alpha1", Resource: "apiexports"})
 		})
-	}
-}
-
-func waitUntilCacheServerIsReady(ctx context.Context, t *testing.T, cacheClientRT *rest.Config) {
-	cacheClientRT = rest.CopyConfig(cacheClientRT)
-	if cacheClientRT.NegotiatedSerializer == nil {
-		cacheClientRT.NegotiatedSerializer = kubernetesscheme.Codecs.WithoutConversion()
-	}
-	client, err := rest.UnversionedRESTClientFor(cacheClientRT)
-	if err != nil {
-		t.Fatalf("failed to create unversioned client: %v", err)
-	}
-
-	wg := sync.WaitGroup{}
-	wg.Add(2)
-	for _, endpoint := range []string{"/livez", "/readyz"} {
-		go func(endpoint string) {
-			defer wg.Done()
-			waitForEndpoint(ctx, t, client, endpoint)
-		}(endpoint)
-	}
-	wg.Wait()
-}
-
-func waitForEndpoint(ctx context.Context, t *testing.T, client *rest.RESTClient, endpoint string) {
-	var lastError error
-	if err := wait.PollImmediateWithContext(ctx, 100*time.Millisecond, time.Minute, func(ctx context.Context) (bool, error) {
-		req := rest.NewRequest(client).RequestURI(endpoint)
-		_, err := req.Do(ctx).Raw()
-		if err != nil {
-			lastError = fmt.Errorf("error contacting %s: failed components: %w", req.URL(), err)
-			return false, nil
-		}
-
-		t.Logf("success contacting %s", req.URL())
-		return true, nil
-	}); err != nil && lastError != nil {
-		t.Error(lastError)
 	}
 }
--- a/test/e2e/cache/cache_server_test.go
+++ b/test/e2e/cache/cache_server_test.go
@@ -30,6 +30,7 @@ import (
 	"github.com/kcp-dev/logicalcluster/v2"
 	"github.com/stretchr/testify/require"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -62,8 +63,7 @@ var scenarios = []testScenario{
 	{"TestUIDGenerationCreationTimeOverwrite", testUIDGenerationCreationTime},
 	{"TestUIDGenerationCreationTimeNegativeOverwriteNegative", testUIDGenerationCreationTimeNegative},
 	{"TestGenerationOnSpecChanges", testGenerationOnSpecChanges},
-	// TODO: changing spec doesn't increase the Generation of a replicated object
-	// TODO: deleting an object with finalizers immediately removes the obj
+	{"TestDeletionWithFinalizers", testDeletionWithFinalizers},
 	// TODO: spec and status can be updated at the same time
 }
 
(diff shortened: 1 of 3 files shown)


test/e2e/cache/cache_server_test.go
```diff
@@ -0,0 +1,207 @@
...
+	cacheDynamicClient, err := dynamic.NewClusterForConfig(cacheClientRT)
+	require.NoError(t, err)
+	mangoDB := newFakeAPIExport("mangodb")
+	mangoDB.UID = "3"
+	mangoDB.Generation = 1
+	mangoDB.Annotations = map[string]string{genericrequest.AnnotationKey: "amber", "kcp.dev/cluster": cluster.String()}
+	mangoDBRaw, err := toUnstructured(mangoDB)
+	require.NoError(t, err)
+
+	t.Logf("Create abmer/%s/mangodb on the cache server", cluster)
```
stevekuznetsov, line 93: nit: we usually do cluster|namespace/name

test/e2e/cache/cache_server_test.go
```diff
@@ -0,0 +1,207 @@
...
+	require.NoError(t, err)
+	cachedMangoDB := &fakeAPIExport{}
+	require.NoError(t, json.Unmarshal(cachedMangoDBJson, cachedMangoDB))
+	cachedMangoDB.ResourceVersion = ""
+	if !reflect.DeepEqual(cachedMangoDB, mangoDB) {
+		t.Errorf("received object from the cache server differs from the expected one :\n%s", cmp.Diff(cachedMangoDB, mangoDB))
+	}
+}
+
+func newFakeAPIExport(name string) *fakeAPIExport {
```
stevekuznetsov, line 107: literal test data should IMHO be scoped to the test using it - otherwise it gets hard to reason about :)

test/e2e/cache/cache_server_test.go
```diff
@@ -0,0 +1,207 @@
...
+	t.Logf("Create abmer/%s/mangodb on the cache server", cluster)
+	cachedMangoDBRaw, err := cacheDynamicClient.Cluster(cluster).Resource(gvr).Create(ctx, mangoDBRaw, metav1.CreateOptions{})
+	require.NoError(t, err)
+
+	cachedMangoDBJson, err := cachedMangoDBRaw.MarshalJSON()
+	require.NoError(t, err)
+	cachedMangoDB := &fakeAPIExport{}
+	require.NoError(t, json.Unmarshal(cachedMangoDBJson, cachedMangoDB))
+	cachedMangoDB.ResourceVersion = ""
+	if !reflect.DeepEqual(cachedMangoDB, mangoDB) {
```
stevekuznetsov, line 102: Do you need to use semantic equality here?

test/e2e/cache/cache_server_test.go
```diff
@@ -0,0 +1,207 @@
...
+		t.Errorf("received object from the cache server differs from the expected one :\n%s", cmp.Diff(cachedMangoDB, mangoDB))
```
stevekuznetsov, line 103: So this test is saying that a shard, when POSTing to the cache server, can POST an object with UID/generation, and they get cleared when some other client requests the object?

test/e2e/cache/cache_server_test.go
p0lyn0mial: for the cache server we need to take into account a shard name

test/e2e/cache/cache_server_test.go
p0lyn0mial: will be resued across test cases

test/e2e/cache/cache_server_test.go
p0lyn0mial: yes, I need to make sure an obj wasn't altered in any other way.

test/e2e/cache/cache_server_test.go
p0lyn0mial: yes, I'm going to add a negative case as well ( the one that test if UID/generation is set when the shard annotation wasn't set)

test/e2e/cache/cache_server_test.go
stevekuznetsov: Build what we need now, not the future :P

test/e2e/cache/cache_server_test.go
stevekuznetsov: Can we update the test names to provide some of this information?

test/e2e/cache/cache_server_test.go
p0lyn0mial: added more scenarios to this PR.

pkg/cache/server/handler.go
```diff
@@ -58,8 +58,17 @@ func init() {
 // /shards/sapphire/clusters/system:sapphire/apis/apis.kcp.dev/v1alpha1/apiexports
 //
 // /shards/amber/clusters/system:amber/apis/apis.kcp.dev/v1alpha1/apiexports
+//
+// Note:
+// not all paths require to have a valid shard name,
+// as of today the following paths pass through: "/livez", "/readyz", "/healthz"
```
stevekuznetsov, line 64: Can we use `NonResourceRequest` to filter here?

test/e2e/cache/cache_server_test.go
```diff
@@ -0,0 +1,504 @@
...
+)
+
+type testScenario struct {
+	name string
+	work func(ctx context.Context, t *testing.T, cacheClientRT *rest.Config, cluster logicalcluster.Name, gvr schema.GroupVersionResource)
+}
+
+// scenarios holds all test scenarios
+var scenarios = []testScenario{
+	{"TestSchemaIsNotEnforced", testSchemaIsNotEnforced},
```
stevekuznetsov, line 61: my comment on the names is outstanding - the godoc you have on the methods would be good to have in the test name, when I break this functionality I want to understand what I'm looking at :)

test/e2e/cache/cache_server_test.go
```diff
@@ -0,0 +1,504 @@
...
+	raw, err := runtime.DefaultUnstructuredConverter.ToUnstructured(obj)
+	if err != nil {
+		return nil, err
+	}
+	unstructured.Object = raw
+	return unstructured, nil
+}
+
+// TODO: refactor with TestAllScenariosAgainstStandaloneCacheServer
+func TestCacheServerAllScenarios(t *testing.T) {
```
stevekuznetsov, line 392: You need a `t.Parallel()` and `framework.Suite()` here

test/e2e/cache/cache_server_test.go
```diff
@@ -0,0 +1,504 @@
...
+	cacheClientConfig := clientcmd.NewNonInteractiveClientConfig(cacheServerKubeConfig, "cache", nil, nil)
+	cacheClientRestConfig, err := cacheClientConfig.ClientConfig()
+	require.NoError(t, err)
+	t.Logf("waiting for the cache server at %v to become ready", cacheClientRestConfig.Host)
+	waitUntilCacheServerIsReady(ctx, t, cacheClientRestConfig)
+	cacheClientRT := cacheClientRoundTrippersFor(cacheClientRestConfig)
+
+	for _, scenario := range scenarios {
+		t.Run(scenario.name, func(tt *testing.T) {
+			scenario.work(ctx, tt, cacheClientRT, logicalcluster.New("acme"), schema.GroupVersionResource{Group: "apis.kcp.dev", Version: "v1alpha1", Resource: "apiexports"})
```
stevekuznetsov, line 455: You need a `tt.Parallel()` here

stevekuznetsov: Also, comment - if you use the variable `t` for the inner `*testing.T`, it is not possible to accidentally call the upper `*testing.T`, since it is shadowed. In the past these tests have called the wrong `*testing.T`. Let's please just use `t` since it cannot be incorrect.

test/e2e/cache/cache_server_test.go
p0lyn0mial: done

test/e2e/cache/cache_server_test.go
p0lyn0mial: yeah, that is true, I like to use `tt` just to highlight it is not `t`

also, I'm not convinced we need `tt.Parallel()` here, these tests are quick (a lot of them are under `100`ms), they just send a request to a local server.

test/e2e/cache/cache_server_test.go
sttts: I think Steve is saying that we want real test funcs like `TestSchemaIsNotEnforced` instead of some strings here with that name. I.e. put the tests at the top, and the mechanics to run them into helpers.

p0lyn0mial: we could do that in the future when we have a function (in the framework) that knows how to create a shared instance of the cache server. Before that the tests have to be called via `TestCacheServerAllScenarios` method. 

I'll move the `scenarios` struct closer to the the `TestCacheServerAllScenarios`.

test/e2e/reconciler/cache/helpers.go
```diff
@@ -75,15 +81,27 @@ func StartStandaloneCacheServer(ctx context.Context, t *testing.T, dataDir strin
...
 	t.Logf("Starting the cache server")
 	go func() {
-		// TODO (p0lyn0mial): check readiness of the cache server
 		require.NoError(t, preparedCachedServer.Run(ctx))
 	}()
+
+	cacheServerCertificatePath := path.Join(dataDir, "cache", "apiserver.crt")
+	t.Logf("Waiting for the cache server's certificate file at %s", cacheServerCertificatePath)
+	for {
+		time.Sleep(time.Second)
```
stevekuznetsov, line 90: 30s timeout, 100ms poll ? :)

p0lyn0mial: yes, very good idea, thx!

p0lyn0mial: pushed

Issue #2041, opened 2022-09-21
bug: Syncer deleting downstream namespaces unexpectedly on syncer startup


mikenairn:
### Describe the bug

Downstream namespaces (the ones on the physical cluster) are being deleted when they shouldn't be. This appears to be caused by the syncer thinking that the upstream workspace (the one in KCP) no longer exists when it checks for it's existence on startup . We have also observed the syncer going into a loop of constantly deleting and re-creating namespaces, although I've been unable to re-produce that locally.

syncer logs:

```
I0921 09:55:48.721618       1 namespace_process.go:77] "deleting downstream namespace because the upstream namespace doesn't exist" reconciler="kcp-workload-syncer-namespace" key="kcp-2d1x3k5i2rlu" namespace.workspace="" namespace.namespace="" namespace.name="kcp-2d1x3k5i2rlu" namespace.apiVersion="v1" upstreamWorkspace="root:kuadrant" upstreamNamespace="default"
...
I0921 09:55:54.447445       1 spec_process.go:223] Created downstream namespace kcp-2d1x3k5i2rlu for upstream namespace root:kuadrant|default
```

### Steps To Reproduce

1. Setup kcp with a single sync target and bind to it in a workspace
3. Scale down the syncer `kubectl --kubeconfig ~/.kube/config --context kind-kcp-cluster-1 scale deployments/kcp-syncer-kcp-cluster-1-4yiou870 --replicas=0 -n kcp-syncer-kcp-cluster-1-4yiou870`
4. Wait for sync targets "Ready" status to be false `kubectl get synctargets -o wide -w`
5. Scale up the syncer `kubectl --kubeconfig ~/.kube/config --context kind-kcp-cluster-1 scale deployments/kcp-syncer-kcp-cluster-1-4yiou870 --replicas=1 -n kcp-syncer-kcp-cluster-1-4yiou870`
6. Observe the kcp namespaces being deleted and re-created `kubectl --kubeconfig ~/.kube/config --context kind-kcp-cluster-1 get ns -w | grep kcp-`

### Expected Behaviour

Syncer should not delete namespaces from the physical cluster when they exist in KCP.

### Additional Context

Observed with KCP 0.8.x, syncer image ghcr.io/kcp-dev/kcp/syncer:v0.8.2

jmprusi:
I think that's the expected behavior if it goes like this:

1. SyncTarget Ready == true (as syncer is running)
[...]

PR #2264, opened 2022-10-26
:bug: Avoid syncers deleting namespace from other synctargets. 

jmprusi:
## Summary

This PR adds an additional check (synctarget UID) when deleting downstream namespaces to avoid a reconciliation loop between different syncers from different kcp instances. (or for example, an old syncer from a previous KCP deployment)

## Related issue(s)

related to: https://github.com/kcp-dev/kcp/issues/2041 (doesn't fix it as there are multiple root causes for that issue)

### DIFF
--- a/pkg/syncer/namespace/namespace_downstream_process.go
+++ b/pkg/syncer/namespace/namespace_downstream_process.go
@@ -72,6 +72,14 @@ func (c *DownstreamController) process(ctx context.Context, key string) error {
 		logger.Error(err, "failed to unmarshal namespace locator", "namespaceLocator", namespaceLocatorJSON)
 		return nil
 	}
+
+	// Check if the nsLocator SyncTarget UID is the same as ours. If not, we should ignore this namespace as it could be
+	// managed by different syncer.
+	if nsLocator.SyncTarget.UID != c.syncTargetUID {
+		logger.V(4).Info("downstream namespace is not handled by this sync target, ignoring")
+		return nil
+	}
+
 	logger = logger.WithValues(logging.WorkspaceKey, nsLocator.Workspace, logging.NamespaceKey, nsLocator.Namespace)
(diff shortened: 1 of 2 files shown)


pkg/syncer/namespace/namespace_downstream_process.go
```diff
@@ -72,6 +72,14 @@ func (c *DownstreamController) process(ctx context.Context, key string) error {
 		logger.Error(err, "failed to unmarshal namespace locator", "namespaceLocator", namespaceLocatorJSON)
 		return nil
 	}
+
+	// Check if the nsLocator SyncTarget UID is the same as ours. If not, we should ignore this namespace as it could be
+	// managed by another syncer or even a different cluster.
```
davidfestal: > or even a different cluster
 
wdym ?

stevekuznetsov (approved):

pkg/syncer/namespace/namespace_downstream_process.go
```diff
@@ -72,6 +72,14 @@ func (c *DownstreamController) process(ctx context.Context, key string) error {
 		logger.Error(err, "failed to unmarshal namespace locator", "namespaceLocator", namespaceLocatorJSON)
 		return nil
 	}
+
+	// Check if the nsLocator SyncTarget UID is the same as ours. If not, we should ignore this namespace as it could be
+	// managed by different syncer.
+	if nsLocator.SyncTarget.UID != c.syncTargetUID {
+		logger.V(4).Info("downstream namespace is not handled this sync target, ignoring")
```
davidfestal, line 79:
```suggestion
		logger.V(4).Info("downstream namespace is not handled by this sync target, ignoring")
```

jmprusi: fixed, thanks

jmprusi:
@davidfestal I've force pushed to trigger the jobs that didn't start. can you add lgtm again? thanks!

PR #2265, opened 2022-10-26
:seedling: make: add non-kind shared and sharded e2e

stevekuznetsov:
test/e2e: make sure tests actually run in parallel

---

test/e2e: be consistent about timeouts and poll intervals

---

test/e2e/workspacetype: poll, don't blindly wait for 5s

---

test/e2e/syncer: send one request, not many

You can't prove a negative, and eventually consistent systems don't like
to conform to your idea of "long enough" before synchronizing. We can't
assume that "nothing will happen" within 5s here. If we wanted to make
sure that the controller had seen & acknowledged the object, we should
have it post some observed generation.

Since these polls were not effective at best, and flaky at worst, we can
save 10s by just making one call.

---

make: add non-kind variants of shared & sharded e2e

---

make: be more intelligent with parallelism, count
    
First, we don't want to pass `-count 1` to `go test` unless someone
explicitly opts into this behavior, as this explicitly invalidates the
test cache. When re-running tests locally, being able to run the full
suite every time and use the cache is incredibly useful.
    
Second, we should only bother with limiting parallelism when we're
starting up a full `kcp` and `etcd` server per test case.
    
---

/cc @jmprusi @davidfestal 
/assign @ncdc @sttts

### DIFF
--- a/test/e2e/syncer/syncer_test.go
+++ b/test/e2e/syncer/syncer_test.go
@@ -329,25 +329,15 @@ func TestSyncerLifecycle(t *testing.T) {
 		return false, toYaml(deployment)
 	}, wait.ForeverTestTimeout, time.Millisecond*100, "upstream Deployment %s/%s didn't get the per-location deletion annotation set or there was an error", upstreamNamespace.Name, upstreamDeployment.Name)
 
-	t.Logf("Checking if upstream deployment %s/%s is getting deleted, shouldn't as the syncer will not remove its finalizer due to the virtual finalizer", upstreamNamespace.Name, upstreamDeployment.Name)
-	require.Never(t, func() bool {
-		_, err := upstreamKubeClusterClient.Cluster(wsClusterName).AppsV1().Deployments(upstreamNamespace.Name).Get(ctx, upstreamDeployment.Name, metav1.GetOptions{})
-		if apierrors.IsNotFound(err) {
-			return true
-		}
-		require.NoError(t, err)
-		return false
-	}, 5*time.Second, time.Second, "upstream Deployment %s/%s got deleted or there was an error", upstreamNamespace.Name, upstreamDeployment.Name)
+	t.Logf("Checking if upstream deployment %s/%s is deleted, shouldn't as the syncer will not remove its finalizer due to the virtual finalizer", upstreamNamespace.Name, upstreamDeployment.Name)
+	_, err = upstreamKubeClusterClient.Cluster(wsClusterName).AppsV1().Deployments(upstreamNamespace.Name).Get(ctx, upstreamDeployment.Name, metav1.GetOptions{})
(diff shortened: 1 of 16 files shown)


nrb:
Running `make test-e2e-shared-minimal` a second time on my machine (macOS 16.1 ARM) results in this error; deleting the whole `.kcp` directory fixes it.

```shell
KCP Error: cannot create the 'admin.kubeconfig` file with an empty token for the shard-admin user
error: kcp shard kcp terminated with exit code 1
```

stevekuznetsov:
@nrb that's a pre-existing boo-boo for these targets

test/e2e/syncer/syncer_test.go
```diff
@@ -329,25 +329,15 @@ func TestSyncerLifecycle(t *testing.T) {
...
 
-	t.Logf("Checking if upstream deployment %s/%s is getting deleted, shouldn't as the syncer will not remove its finalizer due to the virtual finalizer", upstreamNamespace.Name, upstreamDeployment.Name)
-	require.Never(t, func() bool {
-		_, err := upstreamKubeClusterClient.Cluster(wsClusterName).AppsV1().Deployments(upstreamNamespace.Name).Get(ctx, upstreamDeployment.Name, metav1.GetOptions{})
-		if apierrors.IsNotFound(err) {
-			return true
-		}
-		require.NoError(t, err)
-		return false
-	}, 5*time.Second, time.Second, "upstream Deployment %s/%s got deleted or there was an error", upstreamNamespace.Name, upstreamDeployment.Name)
```
ncdc, line 340: I think the intent here was to wait a ~short amount of time to make sure there weren't any unintentional/unexpected deletions that trickled in after a period of time.

test/e2e/workspacetype/controller_test.go
```diff
@@ -287,10 +288,16 @@ func TestClusterWorkspaceTypes(t *testing.T) {
 				})
 
 				t.Logf("Expect workspace to be stuck in initializing phase")
-				time.Sleep(5 * time.Second)
```
ncdc, line 290: This is another case where there's a desire to wait a ~short amount of time and make sure an event does not happen. Not sure how we want to approach those.

test/e2e/syncer/syncer_test.go
stevekuznetsov: I wrote in the commit why that is not a valid way to test this behavior.

test/e2e/workspacetype/controller_test.go
stevekuznetsov: It is not - we are waiting to see it be in initializing, no?

stevekuznetsov: My comment about proving negatives and assuming time-scales for eventually consistent systems applies to this case as well.

PR #2268, opened 2022-10-26
:sparkles: Add labels to APIExports for shard allocation

fgiloux:
## Summary

This PR adds a labels field to the VirtualWorkspace struct in the APIExport status next to the shard URL. This field gets populated with the labels set on the matching ClusterWorkspaceShard.

This is for controller managers to be able to select shard URLs based on the topology defined by the ClusterWorkspaceShard labels.

A default label: tenancy.kcp.dev/shard is added to all the ClusterWorkspaceShards and is populated with the name of the resource. The idea behind it is that many controllers may only watch resources in their own shard.

More information is available in [this document](https://docs.google.com/document/d/1zI9KOpLp6H_tAcKpGbLpQu_iUK9VbpgUSo0OFEd7Mig/edit?usp=sharing)


stevekuznetsov:
Doesn't this introduce a bunch of write amplification to `ClusterWorkspaceShard` labels, and weird downstream behavior when these fields are out of sync? Are we certain this is the best pattern?

pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -257,6 +257,12 @@ type VirtualWorkspace struct {
 	// +kubebuilder:format:URL
 	// +required
 	URL string `json:"url"`
+
+	// Map of string keys and values that can be used to organize and categorize
+	// (scope and select) objects. May match selectors of controllers and services.
+	//
+	// +optional
+	Labels map[string]string `json:"labels,omitempty"`
```
p0lyn0mial, line 265: how is this field going to be used? could it be defined as a `string`?

pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -257,6 +257,12 @@ type VirtualWorkspace struct {
 	// +kubebuilder:format:URL
 	// +required
 	URL string `json:"url"`
+
+	// Map of string keys and values that can be used to organize and categorize
+	// (scope and select) objects. May match selectors of controllers and services.
```
p0lyn0mial, line 262: s/objects/shards ?

pkg/reconciler/apis/apiexport/apiexport_reconcile.go
```diff
@@ -211,14 +211,9 @@ func (c *controller) updateVirtualWorkspaceURLs(ctx context.Context, apiExport *
...
-		desiredURLs.Insert(u.String())
-	}
-
-	apiExport.Status.VirtualWorkspaces = nil
-
-	for _, u := range desiredURLs.List() {
 		apiExport.Status.VirtualWorkspaces = append(apiExport.Status.VirtualWorkspaces, apisv1alpha1.VirtualWorkspace{
-			URL: u,
+			URL:    u.String(),
+			Labels: clusterWorkspaceShard.Labels,
```
p0lyn0mial, line 216: also, please make sure that the output (`apiExport.Status.VirtualWorkspaces)` will be stable (ordered), otherwise we will end up modifying the export on each sync

pkg/reconciler/tenancy/clusterworkspaceshard/clusterworkspaceshard_controller.go
```diff
@@ -128,33 +128,36 @@ func (c *Controller) processNextWorkItem(ctx context.Context) bool {
 	// other workers.
 	defer c.queue.Done(key)
 
-	if err := c.process(ctx, key); err != nil {
+	if requeue, err := c.process(ctx, key); err != nil {
 		runtime.HandleError(fmt.Errorf("%q controller failed to sync %q, err: %w", ControllerName, key, err))
 		c.queue.AddRateLimited(key)
 		return true
+	} else if requeue {
+		// only requeue if we didn't error, but we still want to requeue
```
p0lyn0mial, line 136: could you describe applicable cases?

pkg/apis/apis/v1alpha1/types_apiexport.go
p0lyn0mial: or maybe even a `struct`?

fgiloux:
> Doesn't this introduce a bunch of write amplification to ClusterWorkspaceShard labels

Yes it does introduce write amplification in the sense that  the shard labels are now copied onto each apiexport. This is already the case for the VirtualWorkspace URL in the APIExport status. The alternative would be not to have this information in the APIExport at all and have the controllers retrieving it directly from the ClusterWorkspaceShards in the root workspace. This supposes that these resources are globally available for read. If we want to avoid cross-region reads it probably means that they will need to be replicated onto the cache server.

> weird downstream behavior when these fields are out of sync?

Eventual consistency is business as usual. There may be delay in processing some resources and they may be API calls to a cluster that does not exist anymore until the reconciliation happens. Do you have something else in mind?

> Are we certain this is the best pattern?

I am happy with discussing alternatives

pkg/reconciler/apis/apiexport/apiexport_reconcile.go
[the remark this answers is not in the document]
fgiloux: What you are saying is aligned with Steve's comment on write amplification.
My questions would be
- do we have other use cases for ClusterWorkspaceShard labels than defining a topology whose primary consumers are controllers?
We can define prefixes and filter on them. I would argue that the location may not be the only criteria. Separating environments (prod, staging, dev) or tenants may be other ones.

pkg/apis/apis/v1alpha1/types_apiexport.go
fgiloux: I think that labels / selector is a common, powerful and well understood pattern. It can all be aggregated in a string but I am not sure of the benefit.

pkg/reconciler/tenancy/clusterworkspaceshard/clusterworkspaceshard_controller.go
fgiloux: this is a bit controversial. There was some logic for dealing with status modifications except that the reconcile loop was empty. I still kept the mechanism and the opportunity to flesh it out. I took the approach of doing 1 modification at once, meaning that it requeues each time a modification is done to give a chance to other chunks to do their processing independently. If no more modification is done no requeue.

pkg/reconciler/apis/apiexport/apiexport_reconcile.go
fgiloux: That's a good point. I need to check that

fgiloux:
Closing in favor of [Partition and PartitionSet](https://hackmd.io/EuPmah62QHSEs-po04Kn1Q?view)

PR #2275, opened 2022-10-27
:seedling: .github: restrict kind tests to those that require it

stevekuznetsov:

Depends on https://github.com/openshift/release/pull/33497


ncdc:
LG. Do you want to keep the job names as-is, or adjust to indicate they use kind in some way?

PR #2277, opened 2022-10-27
:bug: .github: cache Go build output

stevekuznetsov:

/assign @nrb

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,6 +35,7 @@ jobs:
       - uses: actions/setup-go@v3
         with:
           go-version: v1.18
+          cache: true
       - run: make build
 
       # Install kind with a local registry
@@ -64,6 +65,7 @@ jobs:
     - uses: actions/setup-go@v3
       with:
         go-version: v1.18
+        cache: true
     - run: make build
 
     # Install kind with a local registry


.github/workflows/ci.yaml
```diff
@@ -64,6 +65,7 @@ jobs:
     - uses: actions/setup-go@v3
       with:
         go-version: v1.18
+        cache: true
```
nrb, line 68: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows says

> Use cache-dependency-path input for cases when multiple dependency files are used, or they are located in different subdirectories.

I don't know if we can (ab)use that to get `ko`'s installation dir or not.

.github/workflows/ci.yaml
```diff
@@ -64,6 +65,7 @@ jobs:
...
     - run: make build
```
nrb, line 69: Can we skip this `make build` step in each job? Both `test-e2e` tagets rely on `build-all`, and if they're not using the output of `make build`, it seems like a waste.

nrb (approved):
/lgtm

A couple small comments, but I think we could investigate them as follow up if this doesn't cut down times enough.

.github/workflows/ci.yaml
stevekuznetsov: We need to have it here to have the binaries available for the image builds.

PR #2286, opened 2022-10-31
🌱 types_apiexport: make all and resourceSelector/namespace optional

s-urbaniak:
## Summary
This sets the fields `all` and `resourceSelector/namespace` optional.

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/pull/2222#discussion_r1008530913
Fixes https://github.com/kcp-dev/kcp/pull/2222#discussion_r1008532082

### DIFF
--- a/config/crds/apis.kcp.dev_apibindings.yaml
+++ b/config/crds/apis.kcp.dev_apibindings.yaml
@@ -53,7 +53,6 @@ spec:
                     records if the user accepts or rejects it.
                   properties:
                     all:
-                      default: false
                       description: all claims all resources for the given group/resource.
                         This is mutually exclusive with resourceSelector.
                       type: boolean
@@ -83,18 +82,22 @@ spec:
                           name:
                             description: name of an object within a claimed group/resource.
                               It matches the metadata.name field of the underlying
-                              object.
+                              object. If namespace is unset, all objects matching
+                              that name will be claimed.
                             maxLength: 253
+                            minLength: 1
                             pattern: ^([a-z0-9][-a-z0-9_.]*)?[a-z0-9]$
                             type: string
                           namespace:
                             description: namespace containing the named object. Matches
                               metadata.namespace field. If "name" is unset, all objects
                               from the namespace are being claimed.
+                            minLength: 1
                             type: string
-                        required:
-                        - namespace
                         type: object
+                        x-kubernetes-validations:
+                        - message: at least one field must be set
+                          rule: has(self.__namespace__) || has(self.name)
                       type: array
                     state:
                       enum:
@@ -102,13 +105,13 @@ spec:
                       - Rejected
                       type: string
                   required:
-                  - all
                   - resource
                   - state
                   type: object
                   x-kubernetes-validations:
                   - message: either "all" or "resourceSelector" must be set
-                    rule: self.all != has(self.resourceSelector)
+                    rule: (has(self.all) && self.all) != (has(self.resourceSelector)
+                      && size(self.resourceSelector) > 0)
                 type: array
               reference:
                 description: reference uniquely identifies an API to bind to.
@@ -158,7 +161,6 @@ spec:
                     allow the service provider access to.
                   properties:
                     all:
-                      default: false
                       description: all claims all resources for the given group/resource.
                         This is mutually exclusive with resourceSelector.
                       type: boolean
@@ -188,26 +190,30 @@ spec:
                           name:
                             description: name of an object within a claimed group/resource.
                               It matches the metadata.name field of the underlying
-                              object.
+                              object. If namespace is unset, all objects matching
+                              that name will be claimed.
                             maxLength: 253
+                            minLength: 1
                             pattern: ^([a-z0-9][-a-z0-9_.]*)?[a-z0-9]$
                             type: string
                           namespace:
                             description: namespace containing the named object. Matches
                               metadata.namespace field. If "name" is unset, all objects
                               from the namespace are being claimed.
+                            minLength: 1
                             type: string
-                        required:
-                        - namespace
                         type: object
+                        x-kubernetes-validations:
+                        - message: at least one field must be set
+                          rule: has(self.__namespace__) || has(self.name)
                       type: array
                   required:
-                  - all
                   - resource
                   type: object
                   x-kubernetes-validations:
                   - message: either "all" or "resourceSelector" must be set
-                    rule: self.all != has(self.resourceSelector)
+                    rule: (has(self.all) && self.all) != (has(self.resourceSelector)
+                      && size(self.resourceSelector) > 0)
                 type: array
               boundResources:
                 description: boundResources records the state of bound APIs.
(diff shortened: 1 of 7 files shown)


pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -193,9 +193,9 @@ type PermissionClaim struct {
 
 	// all claims all resources for the given group/resource.
 	// This is mutually exclusive with resourceSelector.
-	// +required
+	// +optional
 	// +kubebuilder:default=false
-	All bool `json:"all"`
+	All bool `json:"all,omitempty"`
```
sttts, line 198: was originally expecting this as part fo the ResourceSelector.

Related question: what does an empty, but set `resourceSelector` mean?

s-urbaniak: see my question here: https://github.com/kcp-dev/kcp/pull/2222#discussion_r1009281781

s-urbaniak: If we remove the required tag, this leaves space for new field but does not check invariants. The only thing i can think of is adding a CEL validation here?

s-urbaniak: added CEL validation, does this make things look better?

pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -210,6 +210,8 @@ type PermissionClaim struct {
 	IdentityHash string `json:"identityHash,omitempty"`
 }
 
+// TODO(sur): change the expression to "has(self.name) != has(self.__namespace__)" once https://github.com/google/cel-go/pull/571 is availabe in kcp/kubernetes (fixed in k8s 1.25).
+// +kubebuilder:validation:XValidation:rule="has(self.name) && !has(self.__namespace__) || !has(self.name) && has(self.__namespace__)",message="either \"name\" or \"namespace\" must be set"
```
s-urbaniak, line 214: note to reviewers: currently there is an unfortunate bug in 1.24's CEL.

pkg/apis/apis/v1alpha1/types_apiexport.go
s-urbaniak: > was originally expecting this as part fo the ResourceSelector.

clarified out of band: this is fine, as resource selector is a slice.

pkg/apis/apis/v1alpha1/types_apiexport.go
s-urbaniak: resolved, we don't need xor here

s-urbaniak:
found a quirk in API semantics, holding

nrb:
@s-urbaniak Ideally we'd make `All` and `ResourceSelector` mutually exclusive and one of them required, but Go type semantics make this a little frustrating.

I like keeping `All` explicit, because it's giving wide-ranging access. Accidentally omitting a resource selector could mean exposure of all `Secrets`, as an example. While that's not necessarily different from the current state of affairs in Kubernetes, I still think it's not a state we want to be in.

Could we use a bool pointer for `All` and validate that either `All` or `ResourceSelector` are non-nil (but not both)?

s-urbaniak:
unsetting WIP as the state of this PR indeed reflects what we discussed :-)

config/crds/apis.kcp.dev_apibindings.yaml
```diff
@@ -92,17 +92,18 @@ spec:
                               metadata.namespace field. If "name" is unset, all objects
                               from the namespace are being claimed.
                             type: string
-                        required:
-                        - namespace
                         type: object
+                        x-kubernetes-validations:
+                        - message: either "namespace" or "name" and "namespace" must
+                            be set
+                          rule: has(self.__namespace__) || has(self.name) && has(self.__namespace__)
```
sttts, line 99: isn't it the same as "namespace must be set" ? `A or (A and B)` is the same as `A`.

pkg/apis/apis/v1alpha1/types_apiexport.go
sttts: is there a check that either `All` is true or `len(selectors)>0`?

If I am not missing anything, an empty resource selector and `All==false` is valid now.

s-urbaniak: The rule `self.all != has(self.resourceSelector)` covers that, unit test: https://github.com/kcp-dev/kcp/blob/831cc1a5395cde8de761c8c0745977fef6f7adbf/pkg/apis/apis/v1alpha1/types_apiexport_test.go#L60-L68

s-urbaniak: fixed

pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -210,9 +210,11 @@ type PermissionClaim struct {
 	IdentityHash string `json:"identityHash,omitempty"`
 }
 
+// +kubebuilder:validation:XValidation:rule="has(self.__namespace__) || has(self.name)",message="at least one field must be set"
```
s-urbaniak, line 213: @sttts `self != {}` won't work, I get CEl compilation errors.

config/crds/apis.kcp.dev_apibindings.yaml
s-urbaniak: fixed, using a different expression to allow both `name` and `namespace`

pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -210,20 +209,28 @@ type PermissionClaim struct {
...
 type ResourceSelector struct {
 	// name of an object within a claimed group/resource.
 	// It matches the metadata.name field of the underlying object.
+	// If namespace is unset, all objects matching that name will be claimed.
```
nrb, line 216: If ns is unset, wouldn't it be a cluster-scoped object? That's what I would expect from the rest of Kube.

s-urbaniak: That's a very good question. Basically, yes, however in the authorizer we don't have any capability to check whether the request attributes reference a cluster-scoped object, hence my current implementation of the reverse permission authorizer doesn't make a difference and i don't see a similar one in RBAC authorizers from kube.

This is, however, somehow implied by the in-flight API Group/resource. But the invariant is not explicitely checked in the authorizer.

Conversely, one could set `name=foo` for a resource selector on a namespaced API Group/resource and the authorizer would reject any resource in any namespace matching that name.

This is how kube works as well. One can define a cluster role like:
```
kind: ClusterRole
metadata:
  name: cm-reader
rules:
- apiGroups: [""]
  resources: ["configmaps"]
  resourceNames: ["foo"]
  verbs: ["get", "watch", "list"]
```

and bind that to users. Then, a user will only be able to read configmap `foo` in any namespace but will be denied to modify any of them.

sttts:
/hold
/lgtm
/approve

@nrb are you fine to proceed? Seeing your comment. Still doubts?

nrb:
Yeah, let's proceed.

/unhold

PR #2288, opened 2022-10-31
🐛 Fix incorrect cleaning of deletion annotations

### LINKED ISSUE #2287: bug: deletion annotation not removed when scheduler scheduled this resource again
### Describe the bug

When a resource has been scheduled to Synctarget A, and should then be scheduled to SyncTarget B, a deletion annotation is first added to the resource for Synctarget A. If, before the resource is completely removed from SyncTarget A, it is scheduled back to SyncTarget A, the deletion annotation for SyncTarget A is not removed.
This prevents further syncing to happen correctly.

### Steps To Reproduce

...

### Expected Behaviour

Deletion annotations should be updated correctly as it is the case for RequestState labels, from based on the namespace-based scheduling result.

### Additional Context

_No response_

davidfestal:
## Summary

This PR fixes an incorrect cleaning of SyncTarget-related deletion annotations.

## Related issue(s)

Fixes #2287

### DIFF
--- a/pkg/reconciler/workload/resource/resource_reconcile_test.go
+++ b/pkg/reconciler/workload/resource/resource_reconcile_test.go
@@ -188,6 +188,9 @@ func TestComputePlacement(t *testing.T) {
 			}, map[string]string{
 				"state.workload.kcp.dev/cluster-3": "Sync",
 			}, nil, nil, "ns"),
+			wantAnnotationPatch: map[string]interface{}{
+				"deletion.internal.workload.kcp.dev/cluster-3": nil,
+			},
 		},
 		{name: "multiple locations, added and removed on namespace and object",
 			ns: namespace(map[string]string{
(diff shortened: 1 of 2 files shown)


qiujian16 (commented):
/approve
/lgtm

pkg/reconciler/workload/resource/resource_reconcile_test.go
```diff
@@ -188,6 +188,9 @@ func TestComputePlacement(t *testing.T) {
 			}, map[string]string{
 				"state.workload.kcp.dev/cluster-3": "Sync",
 			}, nil, nil, "ns"),
+			wantAnnotationPatch: map[string]interface{}{
```
davidfestal, line 191: @sttts do you agree that the test was previously wrong, or was it on purpose that the deletion annotation would be kept even when rescheduling to the SyncTarget ?

sttts: not sure. Leaving that to @qiujian16 and @jmprusi

PR #2294, opened 2022-11-02
🐛 Fix: watch a certain synctarget only

qiujian16:

## Summary

syncer should only watch a certain synctarget, but not all synctarget in the workspace.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/cliplugins/workload/plugin/sync.go
+++ b/pkg/cliplugins/workload/plugin/sync.go
@@ -479,9 +479,10 @@ func (o *SyncOptions) enableSyncerForWorkspace(ctx context.Context, config *rest
 			Resources:     []string{"synctargets"},
 		},
 		{
-			Verbs:     []string{"get", "list", "watch"},
-			APIGroups: []string{workloadv1alpha1.SchemeGroupVersion.Group},
-			Resources: []string{"synctargets"},
+			Verbs:         []string{"get", "list", "watch"},
+			APIGroups:     []string{workloadv1alpha1.SchemeGroupVersion.Group},
+			Resources:     []string{"synctargets"},
+			ResourceNames: []string{syncTargetName},
 		},
 		{
 			Verbs:         []string{"update", "patch"},
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -30,6 +30,7 @@ import (
 	"github.com/kcp-dev/logicalcluster/v2"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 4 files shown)


jmprusi:
@qiujian16 can we restrict the synctarget with the UID also?

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -479,9 +479,10 @@ func (o *SyncOptions) enableSyncerForWorkspace(ctx context.Context, config *rest
 			Resources:     []string{"synctargets"},
 		},
 		{
-			Verbs:     []string{"get", "list", "watch"},
-			APIGroups: []string{workloadv1alpha1.SchemeGroupVersion.Group},
-			Resources: []string{"synctargets"},
+			Verbs:         []string{"get", "list", "watch"},
+			APIGroups:     []string{workloadv1alpha1.SchemeGroupVersion.Group},
+			Resources:     []string{"synctargets"},
+			ResourceNames: []string{syncTargetName},
```
sttts, line 485: just curious, maybe not necessary to implement:

If we needed this in a production system, i.e. reduce permissions and make people update the syncer. How would we do that?

pkg/syncer/syncer.go
```diff
@@ -130,7 +131,12 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 		return err
 	}
 
-	kcpInformerFactory := kcpinformers.NewSharedInformerFactoryWithOptions(kcpClient, resyncPeriod)
+	// kcpInformerFactory to watch a certain syncerTarget
+	kcpInformerFactory := kcpinformers.NewSharedInformerFactoryWithOptions(kcpClient, resyncPeriod, kcpinformers.WithTweakListOptions(
```
stevekuznetsov, line 135: One final nit - can we make sure not to use `kcpClient` for anything - right now I think the resource sync controller uses it (as opposed to going through the VW?)

pkg/cliplugins/workload/plugin/sync.go
qiujian16: seems to me this permission should be controlled by controller instead of users?

pkg/syncer/syncer.go
qiujian16: this kcpClient is scoped in synctarget workspace https://github.com/kcp-dev/kcp/blob/main/pkg/syncer/syncer.go#L89. resource sync controller needs to update condition of synctarget. But we should not use kcpClusterClient there. Updated.

PR #2298, opened 2022-11-03
:bug: Clean up shadow CRDs after API bindings are deleted

### LINKED ISSUE #2300: bug: KCP consumes lots of memory over time as bound APIs are created and deleted
### Describe the bug

When the first `APIBinding` for a given `APIExport` is created, a CRD in the "shadow" workspace (i.e. `root:bound-crds`) is created.  This CRD is never deleted, even after all `APIBindings` for that `APIExport` are deleted.  Over time, this build up of CRDs can lead to excessive memory consumption.

### Steps To Reproduce

Given any `APIResourceSchema`, `APIExport`, and at lest one corresponding `APIBinding`, run this a few hundred times:

```
kubectl ws create test --enter
kubectl create -f secret.yaml # identity hash
kubectl create -f apis.kcp.dev_apiresourceschema.yaml
kubectl create -f apis.kcp.dev_apiexport.yaml
kubectl ws ..
kubectl ws create test2 --enter
kubectl create -f apis.kcp.dev_apibinding.yaml
```

### Expected Behaviour

Users should be able to create and delete sets of `APIResourceSchema` + `APIExport` + `APIBinding` over and over without it affecting the overall memory footprint of KCP.

### Additional Context

_No response_

kylape:
## Summary

When all API bindings that use a particular APIResourceSchema are deleted, the CRDs that back those bindings need to be deleted.

Fixes #2300

### DIFF
--- a/pkg/indexers/apibinding.go
+++ b/pkg/indexers/apibinding.go
@@ -53,6 +53,22 @@ func IndexAPIBindingByClusterAndAcceptedClaimedGroupResources(obj interface{}) (
 	return ret, nil
 }
 
+const APIBindingByBoundResourceUID = "byBoundResourceUID"
+
+func IndexAPIBindingByBoundResourceUID(obj interface{}) ([]string, error) {
+	apiBinding, ok := obj.(*apisv1alpha1.APIBinding)
+	if !ok {
+		return []string{}, fmt.Errorf("obj %T is not an APIBinding", obj)
+	}
+
+	var ret []string
+	for _, r := range apiBinding.Status.BoundResources {
+		ret = append(ret, r.Schema.UID)
+	}
+
+	return ret, nil
+}
+
 const APIBindingByBoundResources = "byBoundResources"
 
 func IndexAPIBindingByBoundResources(obj interface{}) ([]string, error) {
--- a/pkg/reconciler/apis/apibindingdeletion/apibinding_deletion_controller.go
+++ b/pkg/reconciler/apis/apibindingdeletion/apibinding_deletion_controller.go
@@ -260,7 +260,7 @@ func (c *Controller) process(ctx context.Context, key string) error {
 		if apibindingCopy.Finalizers[i] == APIBindingFinalizer {
 			continue
 		}
-		filtered = append(filtered, APIBindingFinalizer)
+		filtered = append(filtered, apibindingCopy.Finalizers[i])
 	}
 	if len(apibindingCopy.Finalizers) == len(filtered) {
 		return nil
--- a/pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
+++ b/pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
@@ -1,0 +1,267 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package crdcleanup
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	kcpapiextensionsclientset "github.com/kcp-dev/apiextensions-apiserver/pkg/client/clientset/versioned"
+	kcpapiextensionsv1informers "github.com/kcp-dev/apiextensions-apiserver/pkg/client/informers/externalversions/apiextensions/v1"
+	kcpcache "github.com/kcp-dev/apimachinery/pkg/cache"
+	"github.com/kcp-dev/logicalcluster/v2"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+
+	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
+	apisinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/apis/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/indexers"
+	"github.com/kcp-dev/kcp/pkg/logging"
+	"github.com/kcp-dev/kcp/pkg/reconciler/apis/apibinding"
+)
+
+const (
+	ControllerName = "kcp-crdcleanup"
+
+	DefaultIdentitySecretNamespace = "kcp-system"
+)
+
+// NewController returns a new controller for CRD cleanup
+func NewController(
+	crdInformer kcpapiextensionsv1informers.CustomResourceDefinitionClusterInformer,
+	crdClusterClient kcpapiextensionsclientset.ClusterInterface,
+	apiBindingInformer apisinformers.APIBindingInformer,
+) (*controller, error) {
+	queue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), ControllerName)
+
+	c := &controller{
+		queue: queue,
+		getCRD: func(clusterName logicalcluster.Name, name string) (*apiextensionsv1.CustomResourceDefinition, error) {
+			return crdInformer.Lister().Cluster(clusterName).Get(name)
+		},
+		getAPIBindingsByBoundResourceUID: func(name string) ([]*apisv1alpha1.APIBinding, error) {
+			return indexers.ByIndex[*apisv1alpha1.APIBinding](apiBindingInformer.Informer().GetIndexer(), indexers.APIBindingByBoundResourceUID, name)
+		},
+		deleteCRD: func(ctx context.Context, name string) error {
+			return crdClusterClient.ApiextensionsV1().CustomResourceDefinitions().Cluster(apibinding.ShadowWorkspaceName).Delete(ctx, name, metav1.DeleteOptions{})
+		},
+	}
+
+	indexers.AddIfNotPresentOrDie(
+		apiBindingInformer.Informer().GetIndexer(),
+		cache.Indexers{
+			indexers.APIBindingByBoundResourceUID: indexers.IndexAPIBindingByBoundResourceUID,
+		},
+	)
+
+	crdInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: func(obj interface{}) bool {
+			crd, ok := obj.(*apiextensionsv1.CustomResourceDefinition)
+			if !ok {
+				return false
+			}
+
+			return logicalcluster.From(crd) == apibinding.ShadowWorkspaceName
+		},
+		Handler: cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj interface{}) {
+				c.enqueueCRD(obj)
+			},
+			UpdateFunc: func(_, newObj interface{}) {
+				c.enqueueCRD(newObj)
+			},
+			DeleteFunc: func(obj interface{}) {
+				c.enqueueCRD(obj)
+			},
+		},
+	})
+
+	apiBindingInformer.Informer().AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				c.enqueueFromAPIBinding(oldObj, newObj)
+			},
+			DeleteFunc: func(obj interface{}) {
+				c.enqueueFromAPIBinding(nil, obj)
+			},
+		},
+	)
+
+	return c, nil
+}
+
+// controller deletes bound CRDs when they are no longer in use by any APIBindings
+type controller struct {
+	queue workqueue.RateLimitingInterface
+
+	getCRD                           func(clusterName logicalcluster.Name, name string) (*apiextensionsv1.CustomResourceDefinition, error)
+	getAPIBindingsByBoundResourceUID func(name string) ([]*apisv1alpha1.APIBinding, error)
+	deleteCRD                        func(ctx context.Context, name string) error
+}
+
+// enqueueCRD enqueues a CRD.
+func (c *controller) enqueueCRD(obj interface{}) {
+	key, err := kcpcache.DeletionHandlingMetaClusterNamespaceKeyFunc(obj)
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+
+	logger := logging.WithQueueKey(logging.WithReconciler(klog.Background(), ControllerName), key)
+	logger.V(4).Info("queueing CRD")
+	c.queue.Add(key)
+}
+
+func (c *controller) enqueueFromAPIBinding(oldObj, newObj interface{}) {
+	newBinding, ok := newObj.(*apisv1alpha1.APIBinding)
+	if !ok {
+		return
+	}
+
+	logger := logging.WithObject(logging.WithReconciler(klog.Background(), ControllerName), newBinding)
+
+	// Looking at old and new versions in case a schema gets removed from an APIExport.
+	// In that case, the last APIBinding to have the schema removed will trigger the CRD delete,
+	// but only the old version will have the reference to the schema.
+
+	uidSet := sets.String{}
+
+	if oldObj != nil {
+		oldBinding, ok := oldObj.(*apisv1alpha1.APIBinding)
+		if !ok {
+			return
+		}
+
+		for _, boundResource := range oldBinding.Status.BoundResources {
+			uidSet.Insert(boundResource.Schema.UID)
+		}
+	}
+
+	for _, boundResource := range newBinding.Status.BoundResources {
+		uidSet.Insert(boundResource.Schema.UID)
+	}
+
+	for uid := range uidSet {
+		key := kcpcache.ToClusterAwareKey(apibinding.ShadowWorkspaceName.String(), "", uid)
+		logging.WithQueueKey(logger, key).V(2).Info("queueing CRD via APIBinding")
+		c.queue.Add(key)
+	}
+}
+
+// Start starts the controller, which stops when ctx.Done() is closed.
+func (c *controller) Start(ctx context.Context, numThreads int) {
+	defer runtime.HandleCrash()
+	defer c.queue.ShutDown()
+
+	logger := logging.WithReconciler(klog.FromContext(ctx), ControllerName)
+	ctx = klog.NewContext(ctx, logger)
+	logger.Info("Starting controller")
+	defer logger.Info("Shutting down controller")
+
+	for i := 0; i < numThreads; i++ {
+		go wait.UntilWithContext(ctx, c.startWorker, time.Second)
+	}
+
+	<-ctx.Done()
+}
+
+func (c *controller) startWorker(ctx context.Context) {
+	for c.processNextWorkItem(ctx) {
+	}
+}
+
+func (c *controller) processNextWorkItem(ctx context.Context) bool {
+	// Wait until there is a new item in the working queue
+	k, quit := c.queue.Get()
+	if quit {
+		return false
+	}
+	key := k.(string)
+
+	logger := logging.WithQueueKey(klog.FromContext(ctx), key)
+	ctx = klog.NewContext(ctx, logger)
+	logger.V(4).Info("processing key")
+
+	// No matter what, tell the queue we're done with this key, to unblock
+	// other workers.
+	defer c.queue.Done(key)
+
+	if err := c.process(ctx, key); err != nil {
+		runtime.HandleError(fmt.Errorf("%q controller failed to sync %q, err: %w", ControllerName, key, err))
+		c.queue.AddRateLimited(key)
+		return true
+	}
+	c.queue.Forget(key)
+	return true
+}
+
+func (c *controller) process(ctx context.Context, key string) error {
+	clusterName, _, name, err := kcpcache.SplitMetaClusterNamespaceKey(key)
+	if err != nil {
+		return err
+	}
+
+	obj, err := c.getCRD(clusterName, name)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil // object deleted before we handled it
+		}
+		return err
+	}
+
+	logger := logging.WithObject(klog.FromContext(ctx), obj)
+	ctx = klog.NewContext(ctx, logger)
+
+	result, err := c.getAPIBindingsByBoundResourceUID(obj.Name)
+	if err != nil {
+		return err
+	}
+
+	if len(result) > 0 {
+		// An APIBinding that uses this bound CRD was found. Thus don't delete.
+		return nil
+	}
+
+	age := time.Since(obj.CreationTimestamp.Time)
+
+	ageThreshold := time.Second * 10
+	if age < ageThreshold {
+		duration := ageThreshold - age
+		logger.V(4).Info("Requeueing until CRD is older to give some time for the bindings to complete initialization", "duration", duration)
+		c.queue.AddAfter(key, duration)
+		return nil
+	}
+
+	logger.V(1).Info("Deleting CRD")
+	if err := c.deleteCRD(ctx, obj.Name); err != nil {
+		if errors.IsNotFound(err) {
+			return nil // object deleted before we handled it
+		}
+		return err
+	}
+
+	return nil
+}
--- a/pkg/reconciler/apis/crdcleanup/crdcleanup_controller_test.go
+++ b/pkg/reconciler/apis/crdcleanup/crdcleanup_controller_test.go
@@ -1,0 +1,127 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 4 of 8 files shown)


pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,265 @@
...
+
+	bindingList, err := c.apiBindingLister.List(labels.Everything())
+
+	if err != nil {
+		return err
+	}
+
+	for _, binding := range bindingList {
+		if len(binding.Status.BoundResources) == 0 {
+			logger.V(1).Info("APIBinding has no bound resources.  Need to wait until all bindings have bound resources")
```
ncdc, line 243: I worry a bit that an "invalid" binding (e.g. referenced APIExport doesn't exist, whatever...) could permanently interrupt the cleanup process

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,265 @@
...
+			c.queue.AddAfter(key, time.Second)
+			return nil
+		}
+		for _, br := range binding.Status.BoundResources {
+			if br.Schema.UID == obj.Name {
```
ncdc, line 248: I'm not entirely clear how this is different from the index check?

pkg/indexers/apibinding.go
```diff
@@ -53,6 +55,25 @@ func IndexAPIBindingByClusterAndAcceptedClaimedGroupResources(obj interface{}) (
...
+	if !ok {
+		return []string{}, fmt.Errorf("obj %T is not an APIBinding", obj)
+	}
+
+	logger := logging.WithObject(klog.Background(), apiBinding)
+
+	var ret []string
+	for _, r := range apiBinding.Status.BoundResources {
+		ret = append(ret, r.Schema.UID)
+		logger.V(3).Info("Adding to UID index %s", r.Schema.UID)
```
sttts, line 71: this is too heavy. Good for debugging totally. Just remove it before merge :)

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,265 @@
...
+	queue workqueue.RateLimitingInterface
+
+	apiExtensionsClient kcpapiextensionsclientset.ClusterInterface
+	kcpClusterClient    kcpclient.Interface
+
+	crdLister  kcpapiextensionsv1listers.CustomResourceDefinitionClusterLister
+	crdIndexer cache.Indexer
+
+	apiBindingLister  apislisters.APIBindingLister
+	apiBindingIndexer cache.Indexer
```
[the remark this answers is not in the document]
sttts: this controller is special. It's not a normal reconcile controller. It only deleted the main object.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,265 @@
...
+	bindingList, err := c.apiBindingLister.List(labels.Everything())
+
+	if err != nil {
+		return err
+	}
+
+	for _, binding := range bindingList {
+		if len(binding.Status.BoundResources) == 0 {
+			logger.V(1).Info("APIBinding has no bound resources.  Need to wait until all bindings have bound resources")
+			c.queue.AddAfter(key, time.Second)
```
sttts, line 244: why requeue? This happens automatically. You are watch bindings.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
sttts: how can that happen?

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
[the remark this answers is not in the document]
s-urbaniak: does anything speak against wrapping `c.apiExtensionsClient.ApiextensionsV1().CustomResourceDefinitions().Cluster(apibinding.ShadowWorkspaceName)` in a function pointer? This way `process` can be unit tested.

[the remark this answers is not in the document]
stevekuznetsov: Also the list-by-index :)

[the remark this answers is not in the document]
sttts: seeing where you are going. :+1:, it's better to have these two wrapper funcs and a trival unit test.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
kylape: I don't see how this controller can determine if any CRD can be deleted if it cannot see the set of schemas that every API binding points to.  Perhaps there is a way to identify an "invalid" binding such that this controller can safely ignore it?

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
kylape: You're right, I don't think it is.  This section should be only to determine if there are API bindings that are not yet ready to be scanned for schema UIDs.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
kylape: Ah, true.

kylape: Turns out we do in fact need this explicit requeue.  If the `APIBinding` that was discovered to have no bound resources is not actually related to this CRD (we have no reasonable way to know), then we lost perhaps our last opportunity to delete this particular CRD (in the event that it was the deletion event for the last `APIBinding` for this CRD).

kylape:
https://github.com/kcp-dev/kubernetes/pull/110 is required for this PR in order to stop KCP from panicking upon CRD deletion.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
sttts: Don't this is true. There is some case when BoundResources gets empty (could be in theory). The solution with requeue is not correct. It causes a potential DoS.

What we could do:
1. depend on resync to eventually relist the CRD
2. do something that normally a big nogo: enqueue both old and new object on update.

kylape: The `APIBinding` controller will create the bound CRD, wait for it to become established, and _then_ it will add the `BoundResource` to the API binding's status.  I've seen the potential race condition cause a problem when running e2e tests with `-count 20`, but not `-count 5`.  

> depend on resync to eventually relist the CRD

This seems reasonable for now at least, but how can we run the e2e test if we rely on the resync to catch edge cases?  Do we need to modify or remove our new check?  Can we change the resync period for the test?

> It causes a potential DoS

Could we call `queue.AddRateLimited(key)` instead if we're only concerned about DoS?

sttts: > The APIBinding controller will create the bound CRD, wait for it to become established, and then it will add the BoundResource to the API binding's status. I've seen the potential race condition cause a problem when running e2e tests with -count 20, but not -count 5.

Got it. Yes, this is true. Very good insight!

I see how your requeue helps. I am worried that it endlessly requeues. Thinking. Not sure how `AddRateLimited` really works internally. It's won't keep a giant per-key map with requeues, will it?

sttts: Just thinking out loud: could we check for APIResourceSchema existence, and everything else before the actual CRD creation, and only do the latter AFTER setting the BoundResources ?

Wondering whether we run into any other race that way.

sttts: Also the `creationTime` idea would help. Let a CRD live up to a minute or so and only delete then.

pkg/reconciler/apis/apibindingdeletion/apibinding_deletion_controller.go
```diff
@@ -260,7 +260,7 @@ func (c *Controller) process(ctx context.Context, key string) error {
 		if apibindingCopy.Finalizers[i] == APIBindingFinalizer {
 			continue
 		}
-		filtered = append(filtered, APIBindingFinalizer)
+		filtered = append(filtered, apibindingCopy.Finalizers[i])
```
stevekuznetsov, line 263: Nice catch!

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,260 @@
...
+			DeleteFunc: func(obj interface{}) {
+				c.enqueueFromAPIBinding(obj)
+			},
+		},
+	)
+
+	return c, nil
+}
+
+// controller reconciles APIExports. It ensures an export's identity secret exists and is valid.
```
stevekuznetsov, line 119: godoc needs updating

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,260 @@
...
+
+func (c *controller) enqueueFromAPIBinding(obj interface{}) {
+	binding, ok := obj.(*apisv1alpha1.APIBinding)
+	if !ok {
+		return
+	}
+
+	logger := logging.WithObject(logging.WithReconciler(klog.Background(), ControllerName), binding)
+
+	for _, boundResource := range binding.Status.BoundResources {
```
stevekuznetsov, line 150: Normally, we'd have an index on the CRD informer to track CRDs by their binding, and queue all matching CRDs. I *think* using the list here in status is still correct, but it would be good to ACK that we've thought through the edge cases here to ensure that we can't miss a CRD if it's still related to the binding but is not listed in the bound resources array.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,260 @@
...
+		if errors.IsNotFound(err) {
+			return nil // object deleted before we handled it
+		}
+		return err
+	}
+
+	logger := logging.WithObject(klog.FromContext(ctx), obj)
+	ctx = klog.NewContext(ctx, logger)
+
+	if !apihelpers.IsCRDConditionTrue(obj, apiextensionsv1.Established) {
```
stevekuznetsov, line 221: We exit early here since the API binding would not have the CRD listed in status yet?

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,260 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Index is used to quickly return
+	if len(result) > 0 {
+		return nil
+	}
+
+	// The index can't solely be relied on to be certain this CRD can be deleted.
```
stevekuznetsov, line 236: I don't think this is doing what you think this is doing. Right now the `c.listAPIBindings()` call is using the lister - because of the way the machinery works, whether you access the list of bindings from the lister & filter manually using Go code, or if you are accessing the filtered list via the indexer, you are fundamentally operating against the exact same set of objects in the in-memory cache, and you will never (under normal circumstances) get a different result. I think you might want to use the client to do a live call here, instead? It may be useful then also to add labels or something to the bound CRDs for efficient server-side selection, as CRDs are large objects and having to do a full LIST here will be a heavy call.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
kylape: Yes, correct.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
kylape: My comment here is misleading.  I'm not listing here due to a concern about resources in in the index being out of date due to an unprocessed event, etc.  I'm instead seeing if any `APIBindings` are in an unready state to avoid deleting the CRD immediately after it was created.  The `APIBinding` controller will wait for the bound CRD to become established before it adds the corresponding `BoundResource` to its status, so it's expected that the case where an `APIBinding` has no `BoundResources` will be common.  [This thread](https://github.com/kcp-dev/kcp/pull/2298#discussion_r1014475991) is discussing this as well. 

Note that if _any_ `APIBinding` is not ready, we cannot confidently delete any CRD.  

If that all sounds reasonable to you, then I'll update the comment in the code to be less confusing.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
stevekuznetsov: Yes, the index will be updated first.

Yes, removing APIRS from an APIExport is possible.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
stevekuznetsov: Can we add this to a comment, as it's subtle and not entirely clear from the code.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,260 @@
...
+	// All bindings need to be up-to-date to ensure we don't erroneously delete the CRD.
+	bindingList, err := c.listAPIBindings()
+	if err != nil {
+		return err
+	}
+
+	for _, binding := range bindingList {
+		if binding.Status.Phase != apisv1alpha1.APIBindingPhaseBound || !conditions.IsTrue(binding, apisv1alpha1.BindingUpToDate) {
+			logger.V(3).Info("APIBinding has not completed initialization.  Need to wait until all bindings have bound resources")
+			c.queue.Add(key)
```
stevekuznetsov, line 246: Returning an error here will re-queue with backoff, let's do that instead of manually re-inserting & returning `nil`, which is generally used to mean that processing is done and the key should be forgotten.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
kylape: So I started writing the index, and then I realized that there was another edge case: the `APIExport` reference in an `APIBinding` is invalid.  This makes the index much less clean than "by phase", some thing like "APIExportIsValidAndPhaseIsNotBound".  So I kept the list in place but added a big comment trying to explain why we are doing the list after the index lookup.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
kylape: > It's won't keep a giant per-key map with requeues, will it?
Yes, I'm afraid if some `APIBinding` somehow never made it to the `Bound` phase yet also had a valid `APIExport`, then eventually every CRD key would end up in the requeue map.  This should be considered a bug, and we could even alert if the requeue count increased too quickly for the CRD cleanup controller.

> could we check for APIResourceSchema existence, and everything else before the actual CRD creation, and only do the latter AFTER setting the BoundResources ?
Possibly, but this seems unintuitive to me.  The resource isn't actually bound until the CRD is created!

> creationTime idea would help.

If an `APIBinding` (that has a valid `APIExport`) never makes it to the `Bound` phase, then the `creationTime` idea would only delay the inevitable addition of the CRD to the requeue map.

Barring that edge case (which I believe doesn't exist today and if it did would be considered a bug), I can't think of any other scenarios that aren't covered by the code as-is, but I wouldn't mind adding a `creationTime` check as a "belts and suspenders" kind of thing.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
kylape: I'm not sure it's clear why this new edge case is related to the index.  

> Note that if any APIBinding is not ready, we cannot confidently delete any (bound) CRD.

The edge case make this statement untrue since `APIBindings` with an invalid `APIExport` will never become ready, and we must exclude them from consideration to prevent every bound CRD from being requeued forever.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
sttts: > then the creationTime idea would only delay the inevitable addition of the CRD to the requeue map.

No, I mean to delete the CRD after a minute, so give a minute of time to get over the race.

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -413,6 +413,9 @@ func generateCRD(schema *apisv1alpha1.APIResourceSchema) (*apiextensionsv1.Custo
 				apisv1alpha1.AnnotationBoundCRDKey:      "",
 				apisv1alpha1.AnnotationSchemaClusterKey: logicalcluster.From(schema).String(),
 				apisv1alpha1.AnnotationSchemaNameKey:    schema.Name,
+				// HACK: Need to actually persist the identity annotation with just a placeholder value to avoid a panic upon deletion
+				// https://github.com/kcp-dev/kcp/issues/2304
+				apisv1alpha1.AnnotationAPIIdentityKey: "placeholder",
```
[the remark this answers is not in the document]
sttts: or why does this cause panic?

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,345 @@
...
+		return
+	}
+
+	logger := logging.WithQueueKey(logging.WithReconciler(klog.Background(), ControllerName), key)
+	logger.V(4).Info("queueing CRD")
+	c.queue.Add(key)
+}
+
+func (c *controller) enqueueFromAPIBinding(oldObj, newObj interface{}) {
+	oldBinding, ok := oldObj.(*apisv1alpha1.APIBinding)
```
sttts, line 181: this looks wrong. On add this cast will fail.

sttts: The behavior is without consequence, but must be very explicit here with a nil check to be able to reason about the code. Now it looks like an conincidence.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,345 @@
...
+			return nil // object deleted before we handled it
+		}
+		return err
+	}
+
+	logger := logging.WithObject(klog.FromContext(ctx), obj)
+	ctx = klog.NewContext(ctx, logger)
+
+	if !apihelpers.IsCRDConditionTrue(obj, apiextensionsv1.Established) {
+		// Exit early because binding controller waits on CRDs to be established before adding corresponding BoundResource
```
sttts, line 279: with the new lookup logic through export and schema this shouldn't needed. It's a strange non-controller'esque logic that links two controllers. Not good.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,345 @@
...
+	//
+	// In both scenarios, the CRD is created _before_ the APIBinding's status is updated.
+	//
+	// To catch "being updated" APIBindings in scenario 1, simply check if the phase is not "Bound".
+	// To catch "being updated" APIBindings in scenario 2, check the BindingUpToDate condition is False (since the phase will never change after initial bind).
+	//
+	// While the BindingUpToDate condition is actually set to False _after_ the CRD is created,
+	// the CRD established check earlier in this function should ensure that the condition has a chance to get updated before trying to (prematurely) delete this CRD.
+
+	bindingList, err := c.listAPIBindings()
```
sttts, line 309: we cannot do this. There might be 5000 bindings. You need an index.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,345 @@
...
+			return apiBindingInformer.Lister().List(labels.Everything())
+		},
+		getResourceUIDs: func(binding *apisv1alpha1.APIBinding) ([]types.UID, error) {
+			exportCluster := logicalcluster.New(binding.Spec.Reference.Workspace.Path)
+			export, err := apiExportInformer.Lister().Get(client.ToClusterAwareKey(exportCluster, binding.Spec.Reference.Workspace.ExportName))
+
+			if errors.IsNotFound(err) {
+				export, err = temporaryRemoteShardAPIExportInformer.Lister().Get(client.ToClusterAwareKey(exportCluster, binding.Spec.Reference.Workspace.ExportName))
+			}
+
```
sttts, line 86: no new-lines before error handling

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,345 @@
...
+			for _, uid := range uids {
+				if string(uid) == obj.Name {
+					// Found an APIBinding that is still in the process of being bound that will eventually point to this CRD
+					return nil
+				}
+			}
+		}
+	}
```
sttts, line 334: am still not a fan of this super complex lookup, even when optimized with an index. @ncdc any better idea?

sttts: I tend to opt for a less correct, but good enough solution of just delaying deletion to give the controller time to converge.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,345 @@
...
+		},
+	})
+
+	apiBindingInformer.Informer().AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				c.enqueueFromAPIBinding(oldObj, newObj)
+			},
+			DeleteFunc: func(obj interface{}) {
+				c.enqueueFromAPIBinding(obj, obj)
```
sttts, line 148: this is strange. The method says old and new, but now you call it was obj and obj. Would expect `enqueueFromAPIBinding(nil, obj)`.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,345 @@
...
+
+	uidSet := map[string]bool{}
+
+	for _, boundResource := range oldBinding.Status.BoundResources {
+		uidSet[boundResource.Schema.UID] = false
+	}
+
+	for _, boundResource := range newBinding.Status.BoundResources {
+		uidSet[boundResource.Schema.UID] = false
+	}
```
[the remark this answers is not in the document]
kylape: From the comment: It looks at old and new versions in case a schema gets removed from an APIExport. In that case, the last APIBinding to have the schema removed will trigger the CRD delete, but only the old version will have the reference to the schema.

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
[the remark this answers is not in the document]
kylape: Updated comment with a longer explanation.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
kylape: `oldObj` can be `nil` now, and use of it in this function is wrapped in a nil check.

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
kylape: removed

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,264 @@
...
+) (*controller, error) {
+	queue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), ControllerName)
+
+	c := &controller{
+		queue: queue,
+		getCRD: func(clusterName logicalcluster.Name, name string) (*apiextensionsv1.CustomResourceDefinition, error) {
+			return crdInformer.Lister().Cluster(clusterName).Get(name)
+		},
+		getAPIBindingsByBoundResourceUID: func(name string) ([]interface{}, error) {
+			return apiBindingInformer.Informer().GetIndexer().ByIndex(indexers.APIBindingByBoundResourceUID, name)
```
ncdc, line 65: Please use `indexers.ByIndex[T]` so you can return a strongly typed slice

kylape: done :+1:

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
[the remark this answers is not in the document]
ncdc: What Kyle wrote makes sense to me

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -0,0 +1,264 @@
...
+		return
+	}
+
+	logger := logging.WithObject(logging.WithReconciler(klog.Background(), ControllerName), newBinding)
+
+	// Looking at old and new versions in case a schema gets removed from an APIExport.
+	// In that case, the last APIBinding to have the schema removed will trigger the CRD delete,
+	// but only the old version will have the reference to the schema.
+
+	uidSet := map[string]bool{}
```
ncdc, line 149: If you wanted to, you could use `sets.String`

pkg/reconciler/apis/crdcleanup/crdcleanup_controller_test.go
```diff
@@ -0,0 +1,127 @@
...
+				getAPIBindingsByBoundResourceUID: func(name string) ([]*apisv1alpha1.APIBinding, error) {
+					return tt.apiBindingsInIndex, nil
+				},
+				deleteCRD: func(ctx context.Context, name string) error {
+					deleteHappened = true
+					return nil
+				},
+			}
+
+			err := controller.process(context.TODO(), schemaUID)
```
ncdc, line 117:
```suggestion
			err := controller.process(context.Background(), schemaUID)
```

ncdc:
LGTM - please squash!
/approve

kylape:
Fixes https://github.com/kcp-dev/kcp/issues/2304

PR #2302, opened 2022-11-03
:bug: Add request, latency, and workqueue metrics to front proxy

### LINKED ISSUE #2301: bug: the front proxy doesn't expose metrics
### Describe the bug

The front proxy doesn't expose metrics, which means there are no SLIs to measure for SLOs across shards or when a shard is down.

### Steps To Reproduce

n/a

### Expected Behaviour

The front proxy should expose basic metrics like request count and latency.

### Additional Context

_No response_

csams:

## Summary
Add front proxy specific metrics to enable accurate SLO tracking across shards even in the case of a shard being down.

Fixes #2301

### DIFF
--- a/pkg/proxy/mapping.go
+++ b/pkg/proxy/mapping.go
@@ -19,11 +19,12 @@ package proxy
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"os"
 
+	"k8s.io/component-base/metrics/legacyregistry"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/yaml"
 
@@ -46,7 +47,7 @@ type PathMapping struct {
 }
 
 func NewHandler(ctx context.Context, o *proxyoptions.Options, index index.Index) (http.Handler, error) {
-	mappingData, err := ioutil.ReadFile(o.MappingFile)
+	mappingData, err := os.ReadFile(o.MappingFile)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read mapping file %q: %w", o.MappingFile, err)
 	}
@@ -70,6 +71,8 @@ func NewHandler(ctx context.Context, o *proxyoptions.Options, index index.Index)
 		w.WriteHeader(http.StatusOK)
 	}))
 
+	mux.Handle("/metrics", legacyregistry.Handler())
+
 	logger := klog.FromContext(ctx)
 	for _, m := range mapping {
 		logger.WithValues("mapping", m).V(2).Info("adding mapping")
@@ -84,7 +87,7 @@ func NewHandler(ctx context.Context, o *proxyoptions.Options, index index.Index)
 			return nil, fmt.Errorf("failed to create path mapping for path %q: %w", m.Path, err)
 		}
 
-		var handler http.HandlerFunc
+		var handler http.Handler
 		if m.Path == "/clusters/" {
 			clusterProxy := newShardReverseProxy()
 			clusterProxy.Transport = transport
@@ -93,7 +96,7 @@ func NewHandler(ctx context.Context, o *proxyoptions.Options, index index.Index)
 			// TODO: handle virtual workspace apiservers per shard
 			proxy := httputil.NewSingleHostReverseProxy(u)
 			proxy.Transport = transport
-			handler = proxy.ServeHTTP
+			handler = proxy
 		}
 
 		userHeader := "X-Remote-User"
--- a/pkg/proxy/metrics/metrics.go
+++ b/pkg/proxy/metrics/metrics.go
@@ -1,0 +1,60 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import (
+	"net/http"
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+
+	compbasemetrics "k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+// WithLatencyTracking tracks the number of seconds it took the wrapped handler
+// to complete.
+func WithLatencyTracking(delegate http.Handler) http.Handler {
+	return promhttp.InstrumentHandlerDuration(requestLatencies.HistogramVec, delegate)
+}
+
+// TODO(csams): enhance metrics to include shard url
+var (
+	requestLatencies = compbasemetrics.NewHistogramVec(
+		&compbasemetrics.HistogramOpts{
+			Name: "proxy_request_duration_seconds",
+			Help: "Response latency distribution in seconds for each verb and HTTP response code.",
+			Buckets: []float64{0.05, 0.1, 0.2, 0.4, 0.6, 0.8, 1.0, 1.25, 1.5, 2, 3,
+				4, 5, 6, 8, 10, 15, 20, 30, 45, 60},
+			StabilityLevel: compbasemetrics.ALPHA,
+		},
+		[]string{"method", "code"},
+	)
+)
+
+var registerMetrics sync.Once
+
+// Register metrics.
+func Register() {
+	registerMetrics.Do(func() {
+		legacyregistry.MustRegister(requestLatencies)
+	})
+}
+
+func init() {
+	Register()
+}
--- a/pkg/proxy/proxy.go
+++ b/pkg/proxy/proxy.go
@@ -21,10 +21,10 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"os"
 
 	"k8s.io/apimachinery/pkg/util/runtime"
 	userinfo "k8s.io/apiserver/pkg/authentication/user"
@@ -32,7 +32,7 @@ import (
 )
 
 func newTransport(clientCert, clientKeyFile, caFile string) (*http.Transport, error) {
-	caCert, err := ioutil.ReadFile(caFile)
+	caCert, err := os.ReadFile(caFile)
 	if err != nil {
 		return nil, fmt.Errorf("failed to read CA file %q: %w", caFile, err)
 	}
@@ -56,7 +56,7 @@ func newTransport(clientCert, clientKeyFile, caFile string) (*http.Transport, er
 
 // WithProxyAuthHeaders does client cert termination by extracting the user and groups and
 // passing them through access headers to the shard.
-func WithProxyAuthHeaders(delegate http.HandlerFunc, userHeader, groupHeader string, extraHeaderPrefix string) http.HandlerFunc {
+func WithProxyAuthHeaders(delegate http.Handler, userHeader, groupHeader string, extraHeaderPrefix string) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		if u, ok := request.UserFrom(r.Context()); ok {
 			appendClientCertAuthHeaders(r.Header, u, userHeader, groupHeader, extraHeaderPrefix)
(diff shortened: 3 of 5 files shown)


pkg/proxy/metrics/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+	)
+
+	requestLatencies = compbasemetrics.NewHistogramVec(
+		&compbasemetrics.HistogramOpts{
+			Name: "proxy_request_duration_seconds",
+			Help: "Response latency distribution in seconds for each verb and HTTP response code.",
+			// This metric is used for verifying api call latencies SLO,
+			// as well as tracking regressions in this aspects.
+			// Thus we customize buckets significantly, to empower both usecases.
+			Buckets: []float64{0.05, 0.1, 0.2, 0.4, 0.6, 0.8, 1.0, 1.25, 1.5, 2, 3,
```
stevekuznetsov, line 93: Kube provides helpers for these, let's use those if we can. See the authorization cache sync duration metrics we have.

pkg/proxy/metrics/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+// to complete.
+func WithLatencyTracking(delegate http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		sc := &statusCatcher{
+			ResponseWriter: w,
+		}
+
+		start := time.Now()
+		delegate.ServeHTTP(sc, r)
+		elapsedSeconds := time.Since(start).Seconds()
```
stevekuznetsov, line 67: Depending on how these get layered, you're going to count the time taken to observe metrics at the lower level in this. Might want to think of a factoring that only ends up adding one handler into the chain to avoid that, and any duplicate processing we're going to do (e.g. extracting cluster name, shards, etc from the request)

stevekuznetsov, line 67: Depending on how these get layered, you're going to count the time taken to observe metrics at the lower level in this. Might want to think of a factoring that only ends up adding one handler into the chain to avoid that, and any duplicate processing we're going to do (e.g. extracting cluster name, shards, etc from the request)

pkg/proxy/metrics/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+		verb := strings.ToUpper(r.Method)
+		requestLatencies.WithContext(r.Context()).WithLabelValues(verb, code).Observe(elapsedSeconds)
+	})
+}
+
+// TODO(csams): enhance these metrics to include shard url
+var (
+	requestCounter = compbasemetrics.NewCounterVec(
+		&compbasemetrics.CounterOpts{
+			Name:           "proxy_request_total",
```
stevekuznetsov, line 79: nit: naming convention would suggest `proxy_request_count` - but we actually get the counts for free from the histogram below, so we don't need this metric.

pkg/proxy/proxy.go
```diff
@@ -21,18 +21,18 @@ import (
...
+	"os"
 
 	"k8s.io/apimachinery/pkg/util/runtime"
 	userinfo "k8s.io/apiserver/pkg/authentication/user"
 	"k8s.io/apiserver/pkg/endpoints/request"
 )
 
 func newTransport(clientCert, clientKeyFile, caFile string) (*http.Transport, error) {
-	caCert, err := ioutil.ReadFile(caFile)
+	caCert, err := os.ReadFile(caFile)
```
stevekuznetsov, line 35: Is this a Go 1.19 thing? Is it valid under 1.18?

pkg/proxy/metrics/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	compbasemetrics "k8s.io/component-base/metrics"
+	"k8s.io/component-base/metrics/legacyregistry"
+)
+
+// used in the middleware below to capture http status code.
+type statusCatcher struct {
```
stevekuznetsov, line 31: Doing this to the round-tripper on the outgoing requests would not require us to do injection like this, right?

stevekuznetsov: Although that might not measure the same latency?

stevekuznetsov: Although that might not measure the same latency?

pkg/proxy/metrics/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+
+		code := strconv.Itoa(sc.statusCode)
+		verb := strings.ToUpper(r.Method)
+		requestCounter.WithContext(r.Context()).WithLabelValues(verb, code).Inc()
+	})
+}
+
+// WithLatencyTracking tracks the number of seconds it took the wrapped handler
+// to complete.
+func WithLatencyTracking(delegate http.Handler) http.Handler {
```
s-urbaniak, line 59: you can use https://github.com/prometheus/client_golang/blob/main/prometheus/promhttp/instrument_server.go#L84

pkg/proxy/metrics/metrics.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func (s *statusCatcher) WriteHeader(statusCode int) {
+	s.statusCode = statusCode
+	s.ResponseWriter.WriteHeader(statusCode)
+}
+
+// WithRequestCounter counts the number of requests that have passed through
+// the handler it wraps.
+func WithRequestCounter(delegate http.Handler) http.Handler {
```
s-urbaniak, line 43: you can use https://github.com/brancz/prometheus-example-app/blob/1ac20239640c6280cbdafc6e9c26b62814c89c97/vendor/github.com/prometheus/client_golang/prometheus/promhttp/instrument_server.go#L94

s-urbaniak, line 43: https://github.com/prometheus/client_golang/blob/main/prometheus/promhttp/instrument_server.go#L135

pkg/proxy/proxy.go
s-urbaniak: https://pkg.go.dev/io/ioutil

> Deprecated: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details.

should be valid up until 1.16

stevekuznetsov: ACK

pkg/proxy/metrics/metrics.go
csams: Sure, I can change it if you like. I just slightly modified https://github.com/kubernetes/apiserver/blob/master/pkg/endpoints/metrics/metrics.go#L75-L83 and kept the convention there.

pkg/proxy/metrics/metrics.go
csams: Ack - should I use the granularity used in the auth cache or keep what I lifted from https://github.com/kubernetes/apiserver/blob/master/pkg/endpoints/metrics/metrics.go#L101-L102?

pkg/proxy/proxy.go
csams: I'm using 1.19, but it should be valid under 1.18. ioutil has been deprecated since 1.16 according to the docs https://pkg.go.dev/io/ioutil

pkg/proxy/metrics/metrics.go
csams: I lifted the wrong ones anyway... :grimacing: I'll change it to use the helper.

pkg/proxy/metrics/metrics.go
stevekuznetsov: With the histogram it's 100% superfluous.

pkg/proxy/metrics/metrics.go
csams: > Depending on how these get layered, you're going to count the time taken to observe metrics at the lower level in this

I think that's okay unless you feel strongly otherwise. "One wrapper per metric" seems to be a pattern [prometheus uses](https://github.com/prometheus/client_golang/blob/main/prometheus/promhttp/instrument_server.go).

> and any duplicate processing we're going to do (e.g. extracting cluster name, shards, etc from the request)

I plan to open a followup PR that will extract the shard info near the top of the chain and attach it to the context so metrics and the shard reverse proxies can use it.

pkg/proxy/metrics/metrics.go
csams: > // The ObserverVec must have valid metric and label names and must have zero,
// one, or two non-const non-curried labels. For those, the only allowed label
// names are "code" and "method". The function panics otherwise.

I want to come back later and add a shard label.

pkg/proxy/metrics/metrics.go
csams: > // The ObserverVec must have valid metric and label names and must have zero,
// one, or two non-const non-curried labels. For those, the only allowed label
// names are "code" and "method". The function panics otherwise.

I want to come back later and add a shard label.

pkg/proxy/metrics/metrics.go
s-urbaniak: sure, that's still possible with

```
promhttp.InstrumentHandlerDuration(
		durationMetric.MustCurryWith(prometheus.Labels{"shard": "foo"}),
		promhttp.InstrumentHandlerCounter(counterMetric, handler),
	)
```

pkg/proxy/metrics/metrics.go
csams: That's fair. We can just use the +Inf bucket. I'll update the PR.

pkg/proxy/metrics/metrics.go
csams: ahhh thanks! I'll look into using it.

pkg/proxy/metrics/metrics.go
stevekuznetsov: You could just determine the shard at runtime, curry that into the metric, create the delegating handler at runtime and use it?  I'd hate to see us reimplement something that must have been solved as a library with battle-testing behind it :|

stevekuznetsov: Oops, didn't notice the comment here https://github.com/kcp-dev/kcp/pull/2302#discussion_r1013368578 which is the same as mine. Sorry

pkg/proxy/index/index_controller.go
```diff
@@ -30,6 +30,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
+	_ "k8s.io/component-base/metrics/prometheus/clientgo/leaderelection"
```
ncdc, line 33: I'd probably put these in maybe server.go?

pkg/proxy/mapping.go
```diff
@@ -19,11 +19,12 @@ package proxy
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"os"
 
+	"k8s.io/component-base/metrics/legacyregistry"
```
ncdc, line 27: Doesn't have to be solved in this PR if this is good enough to get things going forward, but should we be using metrics/registry instead of metrics/legacyregistry?

pkg/proxy/mapping.go
```diff
@@ -58,18 +59,18 @@ func NewHandler(ctx context.Context, o *proxyoptions.Options, index index.Index)
 
 	mux := http.NewServeMux()
 
-	// TODO: implement proper readyz handler
```
ncdc, line 61: I would leave this in

pkg/proxy/mapping.go
```diff
@@ -58,18 +59,18 @@ func NewHandler(ctx context.Context, o *proxyoptions.Options, index index.Index)
...
 	mux.Handle("/readyz", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.Write([]byte("OK")) //nolint:errcheck
 		w.WriteHeader(http.StatusOK)
 	}))
 
-	// TODO: implement proper livez handler
```
ncdc, line 67: I would leave this in

pkg/proxy/mapping.go
ncdc: Let's file an issue to figure out what to do. Being consistent for now is 👍

hardys:
@csams this lgtm, and tested locally - one question, do you think we should also add some `proxy_request_total` metrics similar to `apiserver_request_total` ?  

I imagine such metrics will be needed to support e.g an SLO that the service returns above a certain percentage of non-error responses (and will be useful for load-testing), I guess we can add as a follow up :) 

Also do you want to squash the fixup commits here or should we have tide squash on merge?

csams:
@hardys I added a request total originally to be consistent with the API server metrics, but @stevekuznetsov preferred to not duplicate the *_count or +Inf metrics that we get already with the histogram, so I removed it.

I'll squash the commits.

hardys:
> @hardys I added a request total originally to be consistent with the API server metrics, but @stevekuznetsov preferred to not duplicate the *_count or +Inf metrics that we get already with the histogram, so I removed it.

Ah thanks, I missed that resolved discussion, I'll use the `+Inf` bucket instead :+1:

PR #2303, opened 2022-11-03
:seedling: test-server: split New/Start/Ready phases

hardys:
## Summary
Decouple the steps to start the test server so we can define a new test-server, start it then call WaitForReady to ensure shard readiness

This is another change decoupled from the [shard-standalone-vw branch](https://github.com/hardys/kcp/commit/bf7cbeec5276aec5b4e84e9d54631307cb6b0c7c) from @ncdc

## Related issue(s)
Follow-up to #2297 and another incremental step towards integrating the the [shard-standalone-vw branch](https://github.com/hardys/kcp/commit/bf7cbeec5276aec5b4e84e9d54631307cb6b0c7c) from @ncdc 
Fixes #

### DIFF
--- a/cmd/sharded-test-server/frontproxy.go
+++ b/cmd/sharded-test-server/frontproxy.go
@@ -193,7 +193,7 @@ func startFrontProxy(
 
 		// intentionally load again every iteration because it can change
 		configLoader := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(&clientcmd.ClientConfigLoadingRules{ExplicitPath: filepath.Join(workDirPath, ".kcp/admin.kubeconfig")},
-			&clientcmd.ConfigOverrides{CurrentContext: "system:admin"},
+			&clientcmd.ConfigOverrides{CurrentContext: "base"},
 		)
 		config, err := configLoader.ClientConfig()
 		if err != nil {
--- a/cmd/test-server/kcp/shard.go
+++ b/cmd/test-server/kcp/shard.go
@@ -21,6 +21,7 @@ import (
 	"embed"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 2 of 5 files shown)


cmd/sharded-test-server/frontproxy.go
```diff
@@ -193,7 +193,7 @@ func startFrontProxy(
 
 		// intentionally load again every iteration because it can change
 		configLoader := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(&clientcmd.ClientConfigLoadingRules{ExplicitPath: filepath.Join(workDirPath, ".kcp/admin.kubeconfig")},
-			&clientcmd.ConfigOverrides{CurrentContext: "system:admin"},
+			&clientcmd.ConfigOverrides{CurrentContext: "base"},
```
hardys, line 196: @ncdc I can see this looks correct looking at the kubeconfig, but I don't fully understand why this is now required (e.g why did it work before), was this always broken and we just didn't see it because we didn't wait for the shard readyz?

Anyway, I do see that without this change the shard never becomes ready so clearly it's needed now :)

hardys:
/cc @ncdc @p0lyn0mial next incremental change derived from the shard-standalone-vw branch

Just to clarify - is it correct that when we switch to the standalone VW server, we can derive the availability of the VW server via the shard readyz?  

Just wanted to ensure I correctly understood the rationale for this change as I'm not super familiar with the shard/VW bootstrapping atm.

stevekuznetsov:
> we can derive the availability of the VW server via the shard readyz?

That doesn't seem right - they would be slightly interdependent at startup with e.g. apibinding initializer, but I wouldn't expect the shard -> vw flow.

cmd/sharded-test-server/frontproxy.go
ncdc: Yes

cmd/test-server/kcp/shard.go
```diff
@@ -114,26 +133,28 @@ func Start(ctx context.Context, name, runtimeDir, logFilePath string, args []str
 		}
 	}()
 
+	// Start a goroutine that will notify when the process has exited
 	terminatedCh := make(chan error, 1)
+	s.terminatedCh = terminatedCh
```
p0lyn0mial, line 138: nit: I think we could do `s.terminatedCh = make(chan error, 1)`

cmd/test-server/kcp/shard.go
```diff
@@ -114,26 +133,28 @@ func Start(ctx context.Context, name, runtimeDir, logFilePath string, args []str
...
 	go func() {
 		terminatedCh <- cmd.Wait()
```
p0lyn0mial, line 140: then we could use `s.terminatedCh` here

cmd/sharded-test-server/frontproxy.go
p0lyn0mial: ah, looks like `system:admin` is an invalid context

p0lyn0mial:
> Just to clarify - is it correct that when we switch to the standalone VW server, we can derive the availability of the VW server via the shard readyz?

The standalone VW server is a distinct binary, so we need to check readyz separately. When the VW server is run in-process it registers a startup hook so checking a shard readyz endpoint is enough.

cmd/test-server/kcp/shard.go
```diff
@@ -143,26 +164,32 @@ func Start(ctx context.Context, name, runtimeDir, logFilePath string, args []str
...
 			}
-			return nil, fmt.Errorf("kcp shard %s terminated with unknown error: %w", name, err)
+			return nil, fmt.Errorf("kcp Shard %s terminated with unknown error: %w", s.name, err)
 		default:
 		}
 
 		// intentionally load again every iteration because it can change
+		kubeconfigPath := filepath.Join(s.runtimeDir, "admin.kubeconfig")
 		configLoader := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(&clientcmd.ClientConfigLoadingRules{ExplicitPath: kubeconfigPath},
 			&clientcmd.ConfigOverrides{CurrentContext: "system:admin"},
```
p0lyn0mial, line 194: okay, I think it works because the rest client get an absolute path (we don't append `clusters/root`)

hardys:
> > Just to clarify - is it correct that when we switch to the standalone VW server, we can derive the availability of the VW server via the shard readyz?
> 
> The standalone VW server is a distinct binary, so we need to check readyz separately. When the VW server is run in-process it registers a startup hook so checking a shard readyz endpoint is enough.

Ok thanks - I guess this decoupling was added by @ncdc as a step towards that, but I don't see any check of the standalone VW server readyz in the branch, we can add that as a follow-up in the PR which enables standalone VW

cmd/test-server/kcp/shard.go
hardys: I think it works because this is a valid context for the shard admin.kubeconfig (it's the shard-admin user), but it's not for the front-proxy one?

```
$ grep system:admin .kcp/admin.kubeconfig 
$ grep -B3 system:admin .kcp-0/admin.kubeconfig 
- context:
    cluster: base
    user: shard-admin
  name: system:admin
```

cmd/test-server/kcp/shard.go
hardys: I don't think this will work, because `s.terminatedCh` is a receive-only channel, so we need to use the unconstrained version here

PR #2307, opened 2022-11-04
🐛 Fix root phase 0 bootstrapping sometimes failing

### LINKED ISSUE #2148: bug: root phase 0 bootstrapping can fail and return early from the kcp-start-informers poststart hook
### Describe the bug

root phase 0 bootstrapping can fail and return early from the kcp-start-informers poststart hook.

**/readyz** returns not ready forever

### Steps To Reproduce

1. Start kcp, let it fully bootstrap and become ready
2. Shut it down
3. Start it up again
4. You might see `failed to bootstrap root workspace phase 0` in the log with an error that an APIBinding could not be updated due to a conflict

### Expected Behaviour

Retry the update, don't return nil from the post start hook

### Additional Context

https://github.com/kcp-dev/kcp/blob/047c62dc797452d218260d342842f059290bac68/pkg/server/server.go#L180-L188

says `don't klog.Fatal. This only happens when context is cancelled` but that's not 100% true in this case. `configrootphase0.Bootstrap` calls 

https://github.com/kcp-dev/kcp/blob/047c62dc797452d218260d342842f059290bac68/config/root-phase0/bootstrap.go#L39

which does not block until success:

https://github.com/kcp-dev/kcp/blob/047c62dc797452d218260d342842f059290bac68/config/helpers/bootstrap.go#L275-L277

ncdc:
## Summary
Make sure that when binding root APIs, we keep retrying updates until they succeed.

## Related issue(s)

Fixes #2148

### DIFF
--- a/config/helpers/bootstrap.go
+++ b/config/helpers/bootstrap.go
@@ -30,8 +30,6 @@ import (
 	"text/template"
 	"time"
 
-	"github.com/kcp-dev/logicalcluster/v2"
-
 	extensionsapiserver "k8s.io/apiextensions-apiserver/pkg/apiserver"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -51,6 +49,7 @@ import (
 	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
 	tenancyhelper "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1/helper"
 	kcpclient "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
+	"github.com/kcp-dev/kcp/pkg/logging"
 )
 
 // TransformFileFunc transforms a resource file before being applied to the cluster.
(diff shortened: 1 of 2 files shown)


config/helpers/bootstrap.go
```diff
@@ -261,22 +262,39 @@ func BindRootAPIs(ctx context.Context, kcpClient kcpclient.Interface, exportName
...
 		if !apierrors.IsAlreadyExists(err) {
 			return err
 		}
 
-		existing, err := kcpClient.ApisV1alpha1().APIBindings().Get(ctx, exportName, metav1.GetOptions{})
-		if err != nil {
+		if err := wait.PollImmediateInfiniteWithContext(ctx, time.Second, func(ctx context.Context) (bool, error) {
+			existing, err := kcpClient.ApisV1alpha1().APIBindings().Get(ctx, exportName, metav1.GetOptions{})
+			if err != nil {
+				return false, err
```
stevekuznetsov, line 276: Why is it ok for this to return `err` and fail?

ncdc: You're right, this hides the same issue I am trying to fix. If we get an already-exists error trying to create, and then the get request fails, we're probably in some weird perma-fail state at that point. Could log + return `false, nil`, could panic...?

ncdc: Updated so we never return an error and always keep retrying

PR #2309, opened 2022-11-04
:sparkles: Add interactive cmd prompt for permission claims 

varshaprasad96:
## Summary
This PR introduces an interactive command prompt to edit the status of open claims for apibindings. The user can provide - Yes/No/Skip to accept/reject/skip any action on a permission claim respectively.

Use cases:
1. Edit permission claims related to a specific APIBinding:
```
➜  kcp git:(claims/edit) ./bin/kubectl-kcp claims edit apibinding cert-manager
Running interactive prompt. Enter (Yes/No/Skip)
Accept permission claim for Resource: secrets (APIBinding: cert-manager) > yes
```

2. Edit permission claims related to all apibindings in a workspace:
```
➜  kcp git:(claims/edit) ✗ ./bin/kubectl-kcp claims edit apibinding
Running interactive prompt. Enter (Yes/No/Skip)
No open claims found for apibinding: "apiresource.kcp.dev-1wked"
Accept permission claim for Resource: secrets (APIBinding: cert-manager) > Yes
No open claims found for apibinding: "tenancy.kcp.dev-59phz"
```

3. Use the edit command, but there are no open permission claims:
```
➜  kcp git:(claims/edit) ✗ ./bin/kubectl-kcp claims edit apibinding
Running interactive prompt. Enter (Yes/No/Skip)
No open claims found for apibinding: "apiresource.kcp.dev-1wked"
No open claims found for apibinding: "cert-manager"
```

What happens after editing?
- The APIBinding's `spec.AcceptablePermssionClaim` gets updated based on the user input. For example:

```
➜  kcp git:(claims/edit) ✗ k get apibinding cert-manager -o yaml
kind: APIBinding
metadata:
  ...
spec:
  permissionClaims:
  - all: true
    resource: secrets
    state: Accepted
  reference:
    workspace:
      exportName: cert-manager-stable
      path: root:cert-manager-service
status:
  appliedPermissionClaims:
  - all: true
    resource: secrets
  boundResources:
  - group: cert-manager.io
    resource: certificaterequests
    schema:
...
```

PS: With the updates added to the permission claims spec (like sub-resources, verbs), the prompt's UI will need to change accordingly.

Design doc: https://docs.google.com/document/d/1J31wXY1-2aCyyGFjUlusKoHDzV-zktAmRdIMjMf4OR0/edit

## Related issue(s)

Fixes #


pkg/cliplugins/claims/cmd/cmd.go
```diff
@@ -35,6 +35,12 @@ var (
 
 	# List permission claims and their respective status for all APIBindings in current workspace.
 	%[1]s claims get apibinding
+
+	# Edit the permission claims' status related to a specific APIBinding with an interactive prompt.
+	%[1]s claims get apibinding cert-manager
```
stevekuznetsov, line 40:
```suggestion
	%[1]s claims edit apibinding cert-manager
```

pkg/cliplugins/claims/cmd/cmd.go
```diff
@@ -35,6 +35,12 @@ var (
...
+
+	# Edit the permission claims' status for all APIBindings in current workspace with an interactive prompt.
+	%[1]s claims get apibinding
```
stevekuznetsov, line 43:
```suggestion
	%[1]s claims edit apibinding
```

stevekuznetsov: This reads weird - maybe require `--all` or something?

pkg/cliplugins/claims/cmd/cmd.go
```diff
@@ -85,5 +91,35 @@ func New(streams genericclioptions.IOStreams) *cobra.Command {
...
+		TraverseChildren: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return cmd.Help()
+		},
+	}
+
+	apibindingEditOpts := plugin.NewEditAPIBindingOptions(streams)
+	apibindingEditCmd := &cobra.Command{
+		Use:          "apibinding <apibinding_name>",
+		Short:        "Edit claims related to apibinding",
```
stevekuznetsov, line 108: Add doc to say this does all of them, and when

pkg/cliplugins/claims/plugin/edit.go
```diff
@@ -0,0 +1,193 @@
...
+	e.Options.BindFlags(cmd)
+}
+
+func (e *EditAPIBindingOptions) Run(ctx context.Context) error {
+	cfg, err := e.ClientConfig.ClientConfig()
+	if err != nil {
+		return err
+	}
+
+	_, currentClusterName, err := pluginhelpers.ParseClusterURL(cfg.Host)
```
stevekuznetsov, line 83: It looks like we're copy-pasting this parse + kubeconfig -> client code in every plugin. Why? Can we make a library and re-use it?

pkg/cliplugins/claims/plugin/edit.go
```diff
@@ -0,0 +1,193 @@
...
+			if err != nil {
+				allErrors = append(allErrors, err)
+			}
+		}
+	}
+	return utilerrors.NewAggregate(allErrors)
+}
+
+// getOpenClaims gets the claims which are in a set of `exportClaims` but not in `spec.AcceptableClaims` of the apibinding.
+func getOpenClaims(exportClaims []apiv1alpha1.PermissionClaim, aceptableClaims []apiv1alpha1.AcceptablePermissionClaim) []apiv1alpha1.PermissionClaim {
```
stevekuznetsov, line 154:
```suggestion
func getOpenClaims(exportClaims []apiv1alpha1.PermissionClaim, acceptableClaims []apiv1alpha1.AcceptablePermissionClaim) []apiv1alpha1.PermissionClaim {
```

pkg/cliplugins/claims/plugin/edit.go
```diff
@@ -0,0 +1,193 @@
...
+	openPermissionClaims := []apiv1alpha1.PermissionClaim{}
+
+	// Find the subset which is in exportClaims but not in AcceptableClaims.
+	exportClaimsMap := map[apiv1alpha1.GroupResource]apiv1alpha1.PermissionClaim{}
+	for _, e := range exportClaims {
+		exportClaimsMap[e.GroupResource] = e
+	}
+
+	acceptableClaimsMap := map[apiv1alpha1.GroupResource]apiv1alpha1.PermissionClaim{}
```
stevekuznetsov, line 163: These are not "acceptable" - they are already accepted, right?

pkg/cliplugins/claims/plugin/edit.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		return fmt.Errorf("current URL %q does not point to cluster workspace", cfg.Host)
+	}
+
+	kcpClusterClient, err := newKCPClusterClient(e.ClientConfig)
+	if err != nil {
+		return fmt.Errorf("error while creating kcp client %w", err)
+	}
+
+	// The status of an apibinding has a set of `exportClaims`. The subset which is present in exportClaims but not in spec.AcceptableClaims, are the open claims.
```
stevekuznetsov, line 93:
```suggestion
	// The status of an APIBinding has a set of PermissionClaims declared by the bound APIExport. The spec of an APIBinding has the subset of those PermissionClaims that the user has already accepted. The remaining PermissionClaims are open to be accepted or rejected.
```

pkg/cliplugins/claims/plugin/edit.go
```diff
@@ -0,0 +1,193 @@
...
+	// The command will list those open claims and update `spec.AcceptableClaims` with the status based on the input of the user.
```
stevekuznetsov, line 94:
```suggestion
	// The command will list those open claims and update `spec.PermissionClaims` with the input of the user.
```

pkg/cliplugins/claims/plugin/edit.go
```diff
@@ -0,0 +1,193 @@
...
+
+	apibindings := []apiv1alpha1.APIBinding{}
+	// List permission claims for all bindings in current workspace.
+	if e.allBindings {
+		bindings, err := kcpClusterClient.Cluster(currentClusterName).ApisV1alpha1().APIBindings().List(ctx, metav1.ListOptions{})
+		if err != nil {
+			return fmt.Errorf("error listing apibindings in %q workspace: %w", currentClusterName, err)
+		}
+		apibindings = append(apibindings, bindings.Items...)
```
stevekuznetsov, line 103: You don't need to re-allocate all of these - can just do

```go
apiBindings = bindings.Items
```

pkg/cliplugins/claims/plugin/edit.go
```diff
@@ -0,0 +1,193 @@
...
+	_, err = fmt.Fprintf(e.Out, "Running interactive prompt. Enter (%s/%s/%s)\n", "Yes", "No", "Skip")
+	if err != nil {
+		return err
+	}
+
+	allErrors := []error{}
+	for _, apibinding := range apibindings {
+		openClaims := getOpenClaims(apibinding.Status.ExportPermissionClaims, apibinding.Spec.PermissionClaims)
+		if len(openClaims) == 0 {
+			_, err = fmt.Fprintf(e.Out, "No open claims found for apibinding: %q\n", apibinding.Name)
```
stevekuznetsov, line 129: In the case that the user has not specified a particular binding, I could see this being spammy when they just want to see the things they need to react to

pkg/cliplugins/claims/plugin/edit.go
```diff
@@ -0,0 +1,193 @@
...
+	var state apiv1alpha1.AcceptablePermissionClaimState
+	if action == AcceptClaim {
+		state = apiv1alpha1.ClaimAccepted
+	} else if action == RejectClaim {
+		state = apiv1alpha1.ClaimRejected
+	}
+
+	apibinding.Spec.PermissionClaims = append(apibinding.Spec.PermissionClaims, apiv1alpha1.AcceptablePermissionClaim{PermissionClaim: permissionClaim, State: state})
+
+	_, err := kcpclusterclient.Cluster(clusterName).ApisV1alpha1().APIBindings().Update(ctx, apibinding, metav1.UpdateOptions{})
```
stevekuznetsov, line 191: You likely want to do the update once, aggregating all changes at once.

pkg/cliplugins/claims/plugin/util.go
```diff
@@ -0,0 +1,100 @@
...
+// Print the message for the user in the prompt. The resource name in a claim is a required field,
+// but the group can be empty.
+func printMessage(bindingName, claimGroup, claimResource string) string {
+	if claimGroup != "" {
+		return fmt.Sprintf("Accept permission claim for Group: %s, Resource: %s (APIBinding: %s) > ", claimGroup, claimResource, bindingName)
+	}
+	return fmt.Sprintf("Accept permission claim for Resource: %s (APIBinding: %s) > ", claimResource, bindingName)
+}
+
+func getRequiredInput(rd io.Reader, wr io.Writer, bindingName, claimGroup, claimResource string) (ClaimAction, error) {
```
stevekuznetsov, line 50: We will gain new fields in PermissionClaims soon. You should just pass the full claim into here so that the refactor is trivial when new fields exist.

pkg/cliplugins/claims/plugin/util.go
```diff
@@ -0,0 +1,100 @@
...
+		_, err := fmt.Fprint(wr, printMessage(bindingName, claimGroup, claimResource))
+		if err != nil {
+			return -1, err
+		}
+
+		value := readInput(reader)
+		if value != "" {
+			return inferText(value, wr)
+		}
+		_, err = fmt.Fprintf(wr, "Input is required. Enter `skip/s` instead. ")
```
stevekuznetsov, line 63: I don't understand what this means - what is `skip/s`?

pkg/cliplugins/claims/plugin/util.go
```diff
@@ -0,0 +1,100 @@
...
+		if err != nil {
+			return -1, err
+		}
+	}
+}
+
+func readInput(reader *bufio.Reader) string {
```
stevekuznetsov, line 70: How is this different from `readLine()`?

pkg/cliplugins/claims/plugin/util.go
```diff
@@ -0,0 +1,100 @@
...
+// readstdin reads a line from stdin and returns the value.
+func readLine(reader *bufio.Reader) string {
+	text, err := reader.ReadString('\n')
+	if err != nil {
+		log.Fatalf("Error when reading input: %v", err)
+	}
+	return strings.ToLower(strings.Trim(strings.TrimSpace(text), "`'\""))
+}
+
+func inferText(input string, wr io.Writer) (ClaimAction, error) {
```
stevekuznetsov, line 86: Is there any library to do this for us? I imagine there are other permutations as well.

pkg/cliplugins/claims/plugin/util.go
nrb: I read it to mean `skip` or `s` for shorthand, but it's not immediately clear.

I've normally seen the full argument and the shorthand combined in this format: `(s)kip`.

nrb:
A question I don't see addressed in the design doc - will there be a way for users to programmatically accept permission claims? I know there's danger there and we might not want to make it too easy, but I'm anticipating someone asking.

pkg/cliplugins/claims/cmd/cmd.go
varshaprasad96: The idea was to just to list all the open claims for all bindings in case a specific binding wasn't given. It follows a similar format to the `get` command. If we add `--all` flag then we should probably do so for `get` too.

pkg/cliplugins/claims/cmd/cmd.go
varshaprasad96: Have modified this to be `Edit open permission claims`, since listing all of them is just a one off case which is explained previously in the command help.

pkg/cliplugins/claims/plugin/edit.go
varshaprasad96: not necessarily accepted. They have a `AcceptableClaimState` (https://pkg.go.dev/github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1#AcceptablePermissionClaimState), where it can be in a rejected state too.

pkg/cliplugins/claims/plugin/edit.go
varshaprasad96: modified with a minor tweak

pkg/cliplugins/claims/plugin/edit.go
varshaprasad96: The idea for this was, we would not like the command to output nothing when there are no open claims. Say, for
`kubectl-kcp claims edit apibinding cert-manager`, just no output looks weird. Similarly, for `kubectl-kcp claims edit apibinding` just a single line ("no open permission claims") or no output would just be confusing to the user in terms of whether the command is actually checking all the  permission claims or not (IMO). We could make it more succinct, but imo having the name of apibinding is useful.

pkg/cliplugins/claims/plugin/edit.go
varshaprasad96: A single APIBinding is updated once with all the changes in its permission claims. Do you mean updating all the apibindings in a single go? I'm not sure if that would make a difference since each apibinding will require a separate http call.

pkg/cliplugins/claims/plugin/util.go
varshaprasad96: Yeah, `skip` or `s` was what I was trying to mean. Modified as @nrb suggested.

pkg/cliplugins/claims/plugin/util.go
varshaprasad96: We could make both the methods into one. The reason for having it different was in future we can extend this to parse an array or process string with a different delimiter. Say, 
```go
func readArray(reader *bufio.Reader) []string {
	arr := make([]string, 0)
	text := readLine(reader)

	for _, words := range strings.Split(text, ",") {
		arr = append(arr, strings.TrimSpace(words))
	}
	return arr
}
```
This is just to make post processing of string separate for future use cases.

pkg/cliplugins/claims/plugin/util.go
varshaprasad96: I tried digging into some available options like [go-prompt](https://github.com/c-bata/go-prompt) and [prompt-ui](https://github.com/manifoldco/promptui). These are shell libraries which have more advanced features like auto-complete/suggestions or selections . And for creating simple prompts as we have done, they follow a similar approach as [here](https://github.com/manifoldco/promptui/blob/master/_examples/prompt/main.go). And then also looked into some upstream repos, like kubectl (https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/util/editor/editoptions.go#L889) and kubebuilder (https://pkg.go.dev/sigs.k8s.io/kubebuilder/v3/pkg/plugin/util#YesNo), which seemed to follow a similar approach. For `(Y)es|(N)o|(S)kip` case, the available possibilities are less (precisely 6 - given that we are converting all inputs into a single case for inferring it).

pkg/cliplugins/claims/plugin/edit.go
varshaprasad96: My bad, sorry, please ignore my comment! You were right, it was getting updated for each permission claim, fixed it!

nrb:
> Agreed, for now my understanding was anyone who can access the workspace where APIBinding was created can change the state of the claim. Or maybe the right wording is, any user who can edit the apibinding can accept/reject respective claim.

That's probably sufficient; if someone wants to automate things, they can use Go or wrappers around `kubectl` to do it. This command can be purely interactive. 👍

ncdc:
@varshaprasad96 so sorry for losing sight of this PR! Are you still available to work on it, or would you be interested in trying to find someone to take it over for you?

varshaprasad96:
@ncdc I can work on this sometime this week and rebase with the latest changes.

sttts:
I still think we want this / something like this.

/remove-lifecycle stale

PR #2310, opened 2022-11-07
:sparkles: Front proxy user gating

csams:
## Summary
I'm exploring an authorization check in the front proxy that can prevent requests from reaching shards even for authenticated users.

It works through a rule set of allow and deny lists. Each rule is a regular expression. The set is defined in `root:kcp-system` as yaml embedded in a secret. A controller in the proxy monitors the secret for changes. A proxy admin can configure a default rule set file on the command line as a fallback.

Denies are checked first. Any matching denial rule results in an immediate unauthorized even if the user matches an allow rule.

Allows are checked next. Any success results in an authorized request.

All requests that match neither deny nor allow rules are unauthorized.

### DIFF
--- a/pkg/proxy/options/usergating.go
+++ b/pkg/proxy/options/usergating.go
@@ -17,8 +17,6 @@ limitations under the License.
 package options
 
 import (
-	"errors"
-
 	"github.com/spf13/pflag"
 )
 
@@ -40,8 +38,5 @@ func (o *UserGating) AddFlags(fs *pflag.FlagSet) {
 }
 
 func (c *UserGating) Validate() []error {
-	if c.Enabled && c.DefaultRulesFile == "" {
-		return []error{errors.New("user gating requires a default rules file")}
-	}
 	return nil
 }
--- a/pkg/proxy/options/usergating.go
+++ b/pkg/proxy/options/usergating.go
@@ -1,0 +1,47 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package options
+
+import (
+	"errors"
+
+	"github.com/spf13/pflag"
+)
+
+type UserGating struct {
+	Enabled          bool
+	DefaultRulesFile string
+}
+
+func NewUserGating() *UserGating {
+	return &UserGating{
+		Enabled:          false,
+		DefaultRulesFile: "",
+	}
+}
+
+func (o *UserGating) AddFlags(fs *pflag.FlagSet) {
+	fs.BoolVar(&o.Enabled, "enable-user-gating", false, "Enabled user gating.")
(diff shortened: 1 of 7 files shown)


pkg/proxy/options/usergating.go
```diff
@@ -0,0 +1,42 @@
...
+func NewUserGating() *UserGating {
+	return &UserGating{
+		Enabled:          false,
+		DefaultRulesFile: "",
+	}
+}
+
+func (o *UserGating) AddFlags(fs *pflag.FlagSet) {
+	fs.BoolVar(&o.Enabled, "enable-user-gating", false, "Enabled user gating.")
+	fs.StringVar(&o.DefaultRulesFile, "user-gating-rules", "", "Provide a YAML file containing user gating rules.")
```
sttts, line 37: why do we need two flags? Wouldn't the existence of a rules file be enough?

pkg/proxy/usergating/controller.go
```diff
@@ -0,0 +1,195 @@
...
+// usergate.kcp.dev in root:kcp-system. The secret contains yaml under the key
+// "usergate.yaml" that describes which users are allowed or denied access to
+// the system.
+func NewController(ctx context.Context, versionedInformers informers.SharedInformerFactory, defaultRules *ruleset.RuleSet) *Controller {
+	queue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), controllerName)
+	secretInformer := versionedInformers.Core().V1().Secrets().Cluster(tenancyv1alpha1.RootCluster)
+
+	c := &Controller{
+		queue:        queue,
+		GetSecret:    secretInformer.Lister().Secrets(UserGateNamespace).Get,
```
sttts, line 74: a proxy allowed to read root secrets make me nervous.

pkg/proxy/server.go
```diff
@@ -72,12 +80,30 @@ func NewServer(ctx context.Context, c CompletedConfig) (*Server, error) {
 		},
 	)
 
-	s.Handler, err = NewHandler(ctx, s.CompletedConfig.Options, s.IndexController)
-	if err != nil {
-		return s, err
+	if c.Options.UserGating.Enabled {
```
s-urbaniak, line 83: `if c.Options.UserGating.DefaultRulesFile != ""` to be able to get rid of the bool

pkg/proxy/usergating/controller.go
s-urbaniak: agreed and afais this is not necessary here.

s-urbaniak:
@csams suggested changes: Instead of having a "default rules file" option and a secret containing updated rules, why not 
1. hava `-rules-file` option only.
2. Mount a "rules" secret as part of the deployment.
3. Let the controller watch watch or (better) reload the rules file every, let's say 30 seconds.
4. Whenever the secret changes, it's mounted file and data are updated as well as per https://kubernetes.io/docs/concepts/configuration/secret/#mounted-secrets-are-updated-automatically

This way we get rid of complex watching and controller logic here 👍

pkg/proxy/usergating/controller.go
csams: We could (should?) have some other workspace or at least namespace for it.

pkg/proxy/options/usergating.go
csams: I didn't make the static file based rules required since the primary ruleset lives in some workspace where other systems can dynamically update it.

pkg/proxy/server.go
csams: I didn't make the static file based rules required since the primary ruleset lives in some workspace where other systems can dynamically update it.

csams:
Thanks for the feedback @s-urbaniak ! I should have included in the summary that the proxy is deployed separately from the shards and doesn't include an apiserver or backing etcd. The purpose of having it watch the secret on the root shard is so that other systems can dynamically update the rules through it. The rules file is just a fallback.

PR #2313, opened 2022-11-07
🌱 cache: WithShardScope doesn't require a shard name for well-know paths

p0lyn0mial:
## Summary
It looks like that not all requests require a valid shard name. In order to make checking the health status of the cache server, this PR changes the `WithShardScope` to not require a shard name for `/livez, /readyz, /healthz` paths.

## Related issue(s)

### DIFF
--- a/pkg/cache/server/handler.go
+++ b/pkg/cache/server/handler.go
@@ -47,7 +47,7 @@ func init() {
 
 // WithShardScope reads a shard name from the URL path and puts it into the context.
 // It also trims "/shards/" prefix from the URL.
-// If the path doesn't contain the shard name then a default "system:cache:server" name is assigned.
+// If the path doesn't contain the shard name then a 404 error is returned.
 //
 // For example:
 //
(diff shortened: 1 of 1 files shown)


pkg/cache/server/handler.go
```diff
@@ -58,8 +58,16 @@ func init() {
 // /shards/sapphire/clusters/system:sapphire/apis/apis.kcp.dev/v1alpha1/apiexports
 //
 // /shards/amber/clusters/system:amber/apis/apis.kcp.dev/v1alpha1/apiexports
+//
+// Note:
+// not all paths require to have a valid shard name,
+// as of today the following paths pass through: "/livez", "/readyz", "/healthz"
 func WithShardScope(handler http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		if path := req.URL.Path; path == "/livez" || path == "/readyz" || path == "/healthz" {
```
p0lyn0mial, line 67: maybe in the future it could be extended but for now these are the paths i know about

PR #2319, opened 2022-11-07
:seedling: proxy: optionally enable OIDC auth

hardys:
## Summary
This adds OIDC to the enabled optional auth methods

## Related issue(s)
Follow-up to #2178

### DIFF
--- a/pkg/proxy/options/authentication.go
+++ b/pkg/proxy/options/authentication.go
@@ -51,6 +51,7 @@ func NewAuthentication() *Authentication {
 	auth := &Authentication{
 		BuiltInOptions: kubeoptions.NewBuiltInAuthenticationOptions().
 			WithClientCert().
+			WithOIDC().
 			WithServiceAccounts().
 			WithTokenFile(),
 		// when adding new auth methods, also update AdditionalAuthEnabled below
@@ -62,7 +63,11 @@ func NewAuthentication() *Authentication {
 
 // When configured to enable auth other than ClientCert, this returns true
 func (c *Authentication) AdditionalAuthEnabled() bool {
-	return c.tokenAuthEnabled() || c.serviceAccountAuthEnabled()
+	return c.tokenAuthEnabled() || c.serviceAccountAuthEnabled() || c.oidcAuthEnabled()
+}
+
+func (c *Authentication) oidcAuthEnabled() bool {
+	return c.BuiltInOptions.OIDC != nil && c.BuiltInOptions.OIDC.IssuerURL != ""
 }
 
 func (c *Authentication) tokenAuthEnabled() bool {


hardys:
Some test notes - I tested this locally using [dex standalone](https://dexidp.io/docs/getting-started/)

Since the kube API requires SSL, I generated certs using the [gencert.sh script](https://github.com/dexidp/dex/blob/master/examples/k8s/gencert.sh), modified so the SAN contained localhost e.g:

```
 [alt_names]
DNS.1 = localhost
IP.1 = 127.0.0.1
```

Then I ran dex with a config like:

```
issuer: https://127.0.0.1:5554/dex

storage:
  type: sqlite3
  config:
    file: examples/dex.db

web:
  https: 127.0.0.1:5554
  tlsCert: /home/shardy/go/src/github.com/dexidp/dex/examples/k8s/ssl/cert.pem
  tlsKey: /home/shardy/go/src/github.com/dexidp/dex/examples/k8s/ssl/key.pem

staticClients:
- id: kcp
  name: kcp
  redirectURIs:
  - https://localhost:8000
  - https://localhost:18000
  secret: kcp

connectors:
- type: mockCallback
  id: mock
  name: Example
```

[kubelogin](https://github.com/int128/kubelogin/blob/master/docs/setup.md) can then be used to test the login e.g:

```
kubectl oidc-login setup --oidc-issuer-url=https://127.0.0.1:5554/dex --oidc-client-id=kcp --oidc-client-secret=kcp --certificate-authority /home/shardy/go/src/github.com/dexidp/dex/examples/k8s/ssl/ca.pem --local-server-cert=/home/shardy/go/src/github.com/dexidp/dex/examples/k8s/ssl/cert.pem --local-server-key=/home/shardy/go/src/github.com/dexidp/dex/examples/k8s/ssl/key.pem
```

Then create a clusterrolebinding as directed in the output.

Then we can configure the proxy when running sharded-test-server e.g:

```
go run ./cmd/sharded-test-server --proxy-token-auth-file=$PWD/test/e2e/framework/auth-tokens.csv --proxy-service-account-key-file=.kcp/service-account.crt --proxy-oidc-issuer-url=https://127.0.0.1:5554/dex --proxy-oidc-client-id=kcp --proxy-oidc-ca-file=/home/shardy/go/src/github.com/dexidp/dex/examples/k8s/ssl/ca.pem 
```

Then we can access KCP using OIDC auth with a kubeconfig containing a user like:

```
...
users:
- name: oidc
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - oidc-login
      - get-token
      - --oidc-issuer-url=https://127.0.0.1:5554/dex
      - --oidc-client-id=kcp
      - --oidc-client-secret=kcp
      - --certificate-authority=/home/shardy/go/src/github.com/dexidp/dex/examples/k8s/ssl/ca.pem
      - --local-server-cert=/home/shardy/go/src/github.com/dexidp/dex/examples/k8s/ssl/cert.pem
      - --local-server-key=/home/shardy/go/src/github.com/dexidp/dex/examples/k8s/ssl/key.pem
      command: kubectl
      env: null
      interactiveMode: IfAvailable
      provideClusterInfo: false
```

This is enough to prove the authn is working - but note when using the mock connector there is a limitation in that we [can't configure groups for users](https://github.com/dexidp/dex/issues/1080) - so if we want to test that, we'll need to run an IdP and use one of the "real" connectors instead.

hardys:
I did notice one weird thing: 

```
kubectl workspace
Current workspace is "root:users:zh:bg:https---127-0-0-1-5554-dex--g0w---4--0y---4--0w-g-tb2-r".
```

I don't think that's caused by this PR, but will investigate if it's a real issue with the username, or something caused by my test setup

hardys:
> I did notice one weird thing:
> 
> ```
> kubectl workspace
> Current workspace is "root:users:zh:bg:https---127-0-0-1-5554-dex--g0w---4--0y---4--0w-g-tb2-r".
> ```
> 
> I don't think that's caused by this PR, but will investigate if it's a real issue with the username, or something caused by my test setup

Ok so the token looks like e.g

```
  "iss": "https://127.0.0.1:5554/dex",
  "sub": "Cg0wLTM4NS0yODA4OS0wEgRtb2Nr",
  "aud": "kcp",
```

Then the `User` for the clusterolebinding is a combination of the `iss` and `sub` fields:

`kubectl create clusterrolebinding oidc-cluster-admin --clusterrole=cluster-admin --user='https://127.0.0.1:5554/dex#Cg0wLTM4NS0yODA4OS0wEgRtb2Nr'`

So the username ends up as `https://127.0.0.1:5554/dex#Cg0wLTM4NS0yODA4OS0wEgRtb2Nr` in the `user.DefaultInfo`

KCP then [replaces disallowed characters](https://github.com/kcp-dev/kcp/blob/main/pkg/server/home_workspaces.go#L498) with `-` so this output is expected and unrelated to this PR (although I wonder how this is going to work in practice with real SSO/OIDC, @csams any thoughts on that?)

hardys:
After discussing with @csams I now understand that the issue above is because the dex dev-setup isn't setting name/preferred_username in the claim - I've not yet figured out how to resolve that for test/dev but it sounds like in practice this should work fine.

Issue #1898, opened 2022-09-06
epic: Multiple SyncTarget support and basic coordination controllers 


davidfestal:
### Demo Objective

- [ ] Always consistently maintain syncer-specific views of KCP synced resources, when a resource is synced to several SyncTargets (either inside a location in specific cases, such a move of specific resources, or through several placement to several locations)
- [ ] Provide a way for coordination controllers to get the syncer-specific views of KCP syncer resources, especially status reported by the Syncer.
- [ ] Provide a way for coordination controllers to instruct transformations that should be applied by the syncer when syncing an resource downstream (at first through JsonPatch annotations, later on through more advanced transformation engines such as CEL definitions, driven by dedicated APIs).
- [ ] Provide a way for coordination controllers to instruct summarizing rules that should be applied by the syncer when syncing back a downstream resource downstream (at first just a Json annotation indicating if the syncer-specific status should be promoted to the main resource status. Later on more complete rules could be defined by dedicated APIs)
- [ ] Ensure the exclusiveness of coordination controllers on resources. only one coordination controller should be allowed to access a given coordinated resource in a workspace.

### Action Items

- [X] Scope of the current demo is necessary to fit in the prototype boundaries
- [ ] Contribute to the final demo script and recording

### Stories

- [ ] coordination controller support and demo
- [ ] Finish and merge Syncer virtual workspace transformations (and replace Syncer advanced scheduling feature by transformations)
- [ ] Reenable the splitter use-cases

PR #2321, opened 2022-11-07
🌱 Syncer: Use the features now provided by the Syncer VW, based on transformations

davidfestal:
## Summary

Syncer: Fix TODOs related to transformations: let's use the features now provided by the Syncer VW, based on transformations, instead of in-syncer implementation.

## Related issue(s)

https://github.com/kcp-dev/kcp/issues/1898


hack/logcheck.out
```diff
@@ -33,8 +33,8 @@
 /pkg/admission/webhook/generic_webhook.go:142:5: function "Errorf" should not be used, convert to contextual logging
 /pkg/admission/webhook/generic_webhook.go:172:4: function "Errorf" should not be used, convert to contextual logging
 /pkg/authorization/delegated/authorizer.go:45:3: function "Errorf" should not be used, convert to contextual logging
-/pkg/cliplugins/workload/plugin/sync.go:591:4: function "Infof" should not be used, convert to contextual logging
-/pkg/cliplugins/workload/plugin/sync.go:591:4: function "V" should not be used, convert to contextual logging
+/pkg/cliplugins/workload/plugin/sync.go:592:4: function "Infof" should not be used, convert to contextual logging
```
stevekuznetsov, line 36: Why not just use contextual logging?

pkg/syncer/shared/finalizer.go
```diff
@@ -73,31 +72,26 @@ func EnsureUpstreamFinalizerRemoved(ctx context.Context, gvr schema.GroupVersion
 	}
 	upstreamObj.SetFinalizers(desiredFinalizers)
 
-	//  TODO(jmprusi): This code block will be handled by the syncer virtual workspace, so we can remove it once
-	//                 the virtual workspace syncer is integrated
+	// TODO(davidfestal): The following code block should be removed as soon as the Advanced Scheduling feature is
```
stevekuznetsov, line 75: nit: maybe you all have no problem remembering all of your TODOs but I would suggest making this code panic or not compile or something to make it easier to find what you need to change when scheduling is removed :)

ncdc:
@davidfestal do you plan to proceed with this?

xrstf:
With Syncer/TMC having been removed, I think we can close this PR.

Issue #1779, opened 2022-08-17
Schedule synctarget based on compatible API


qiujian16:
followup of https://github.com/kcp-dev/kcp/issues/1777

in the user workspace, when user creates a placement to schedule workloads to a synctarget in a location. The scheduler should select the synctarget that has the compatible API with workload APIBinding in this user workspace.

This is part of the epic https://github.com/kcp-dev/kcp/issues/1895 to support compute type management.

davidfestal:
Linked with the wrong PR.
Let's reopen.

PR #2329, opened 2022-11-09
✨ Only schedule to synctarget with compatible APIs

qiujian16:
add a filter to filter out synctarget whose supported resouces do not include what in the apibindings.

## Summary
filter the synctarget based on what is the supported resource in synctarget and what is the served apibindings relating to compute in this workspace.

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/1779


pkg/reconciler/workload/placement/placement_reconcile_scheduling_test.go
```diff
@@ -93,6 +95,22 @@ func TestSchedulingReconcile(t *testing.T) {
 				workloadv1alpha1.InternalSyncTargetPlacementAnnotationKey: "aPkhvUbGK0xoZIjMnM2pA0AuV1g7i4tBwxu5m4",
 			},
 		},
+		{
```
davidfestal, line 98: Do we want to add another test when the only SyncTarget in the Location is incompatible (or simply does not have the API) ? This should fail, right ?

pkg/cliplugins/bind/plugin/bind_compute.go
```diff
@@ -248,7 +249,8 @@ func (o *BindComputeOptions) applyAPIBinding(ctx context.Context, client kcpclie
 		clusterName, name := logicalcluster.New(export).Split()
 		apiBinding := &apisv1alpha1.APIBinding{
 			ObjectMeta: metav1.ObjectMeta{
-				Name: apiBindingName(clusterName, name),
+				Name:   apiBindingName(clusterName, name),
+				Labels: map[string]string{apisv1alpha1.LabelAPIBindingKindKey: workloadv1alpha1.LabelKindAPIBindingValue},
```
davidfestal, line 253: I assume that we should now take in account the case when the `APIBinding` exists but does **not** have the `workload` kind (because having been bound previously with the regular `kcp bind apiexport` command). In such a case we should patch it with the new label.

pkg/apis/workload/v1alpha1/types.go
```diff
@@ -155,4 +155,7 @@ const (
 	// InternalSyncTargetKeyLabel is an internal label set on a SyncTarget resource that contains the full hash of the SyncTargetKey, generated with the ToSyncTargetKey(..)
 	// helper func, this label is used for reverse lookups of a syncTargetKey to SyncTarget.
 	InternalSyncTargetKeyLabel = "internal.workload.kcp.dev/key"
+
+	// LabelKindAPIBindingValue is the value of label key "apis.kcp.dev/kind" for compute kind apibinding
+	LabelKindAPIBindingValue = "compute"
```
davidfestal, line 160:
```suggestion
	LabelKindAPIBindingCompute = "compute"
```

davidfestal: would make it easier to understand what it is from other places in the code IMO.

davidfestal: Thanks !

test/e2e/reconciler/scheduling/multi_placements_test.go
```diff
@@ -170,18 +170,6 @@ func TestMultiPlacement(t *testing.T) {
 			return false, fmt.Sprintf("Failed to get service: %v", err)
 		}
 
-		bindings, _ := kcpClusterClient.ApisV1alpha1().APIBindings().List(logicalcluster.WithCluster(ctx, userClusterName), metav1.ListOptions{
```
davidfestal: Could you explain what is the link between the removal of these lines and this PR ?

pkg/cliplugins/bind/plugin/bind_compute.go
```diff
@@ -235,22 +236,43 @@ func (o *BindComputeOptions) applyAPIBinding(ctx context.Context, client kcpclie
...
+	var bindingsToPatch []*apisv1alpha1.APIBinding
 	for _, binding := range apiBindings.Items {
 		if binding.Spec.Reference.Workspace == nil {
 			continue
 		}
-		existingAPIExports.Insert(fmt.Sprintf("%s:%s", binding.Spec.Reference.Workspace.Path, binding.Spec.Reference.Workspace.ExportName))
+		apiExportKey := fmt.Sprintf("%s:%s", binding.Spec.Reference.Workspace.Path, binding.Spec.Reference.Workspace.ExportName)
+		existingAPIExports.Insert(apiExportKey)
+		if desiredAPIExports.Has(apiExportKey) {
+			if len(binding.Labels) == 0 || len(binding.Labels[apisv1alpha1.LabelAPIBindingKindKey]) == 0 {
```
davidfestal, line 247: Should we check when the label is already there with another value, and fail (to be prepared for the time other values my come) ?

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -308,6 +308,7 @@ func (o *SyncOptions) applySyncTarget(ctx context.Context, kcpClient kcpclient.I
 		if err != nil && !apierrors.IsAlreadyExists(err) {
 			return nil, fmt.Errorf("failed to create synctarget %q: %w", syncTargetName, err)
 		}
+		return syncTarget, nil
```
davidfestal, line 311: What is this change here expected to fix ?

pkg/cliplugins/bind/plugin/bind_compute.go
davidfestal: Thanks !

pkg/cliplugins/workload/plugin/sync.go
qiujian16: it should return early after create, otherwise will go to the the unnecessary following patch logic, and it may fail because of version conflict.

davidfestal: But it also returns in case the creation failed because of an `AlreadyExists` error, isn't it ? Should it ?

davidfestal (approved):

pkg/apis/apis/v1alpha1/types_apibinding.go
```diff
@@ -237,6 +237,9 @@ const (
 	AnnotationAPIIdentityKey = "apis.kcp.dev/identity"
 )
 
+// LabelAPIBindingKindKey is a label to identify the kind of APIBinding.
+const LabelAPIBindingKindKey = "apis.kcp.dev/kind"
```
sttts, line 241: what is it used for?

pkg/cliplugins/bind/plugin/bind_compute.go
```diff
@@ -234,21 +236,43 @@ func (o *BindComputeOptions) applyAPIBinding(ctx context.Context, client kcpclie
...
+		_, err = client.ApisV1alpha1().APIBindings().Patch(ctx, bindingName, types.MergePatchType, []byte(patchBytes), metav1.PatchOptions{})
+		if err != nil {
+			errs = append(errs, err)
+		}
+
+		_, err = fmt.Fprintf(o.Out, "apibinding %s patched to set label %s=%s.\n", bindingName, apisv1alpha1.LabelAPIBindingKindKey, workloadv1alpha1.LabelKindAPIBindingCompute)
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
```
sttts, line 266: this confuses me. Either the kind label is API and then the controller should copy it from export to binding. Or it is something for the bind command. I would be worried about the latter. Not every bindings comes from the bind command.

davidfestal: It expresses the intent to use this binding as an `APIBinding` for workloads (== to be synced).

You could `bind apiexport` a workspace to the `root:compute` `kubernetes` `APIExport`, without any intent to sync related resources to a `SyncTarget`, possibly having some controller reconcile related resources and use them in any other way.

But when you `bind compute`, you clearly express the intent to sync to a `SyncTarget`, and so the `APIBinding`s bound to involved `APIExport`s (those supported by the `SyncTarget`s) would be marked with this workload kind.

pkg/apis/apis/v1alpha1/types_apibinding.go
davidfestal: This is used to express some intent related to this binding. See my answer [below](https://github.com/kcp-dev/kcp/pull/2329#discussion_r1022599599)

pkg/cliplugins/bind/plugin/bind_compute.go
sttts: > Not every bindings comes from the bind command.

I don't think we can demand from the user that only the bind command copies this. It *must* come through a controller.

qiujian16: changed to use controller to sync an annotation from apiexport. please review if the annotation name makes sense

pkg/reconciler/workload/apibindinglabel/apibindinglabel_indexes.go
```diff
@@ -0,0 +1,49 @@
...
+// spec.reference.workspace.
+func indexAPIBindingsByWorkspaceExportFunc(obj interface{}) ([]string, error) {
+	apiBinding, ok := obj.(*apisv1alpha1.APIBinding)
+	if !ok {
+		return []string{}, fmt.Errorf("obj is supposed to be an APIBinding, but is %T", obj)
+	}
+
+	if apiBinding.Spec.Reference.Workspace != nil {
+		apiExportClusterName := logicalcluster.New(apiBinding.Spec.Reference.Workspace.Path)
+		if !ok {
```
davidfestal, line 40: I don't understand how we will reach this block

pkg/reconciler/workload/apibindinglabel/apibindinglabel_indexes.go
```diff
@@ -0,0 +1,49 @@
...
+
+	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/client"
+)
+
+const indexAPIBindingsByWorkspaceExport = ControllerName + "apiBindingsByWorkspaceExport"
+
+// indexAPIBindingsByWorkspaceExportFunc is an index function that maps an APIBinding to the key for its
+// spec.reference.workspace.
+func indexAPIBindingsByWorkspaceExportFunc(obj interface{}) ([]string, error) {
```
davidfestal, line 32: Isn't it already available in global indexers here: https://github.com/qiujian16/kcp/blob/c8d456da43271170975228b4e55af03173be2639/pkg/indexers/apibinding.go#L97 ?

pkg/apis/workload/v1alpha1/types.go
```diff
@@ -155,4 +155,7 @@ const (
 	// InternalSyncTargetKeyLabel is an internal label set on a SyncTarget resource that contains the full hash of the SyncTargetKey, generated with the ToSyncTargetKey(..)
 	// helper func, this label is used for reverse lookups of a syncTargetKey to SyncTarget.
 	InternalSyncTargetKeyLabel = "internal.workload.kcp.dev/key"
+
+	// IntergalComputeAPIBindingAnnotationKey is a annotation key to identify the APIBinding is for compute.
```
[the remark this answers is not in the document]
davidfestal: @sttts any thought or preference here ?

test/e2e/framework/kcp.go
```diff
@@ -850,6 +850,8 @@ func NewFakeWorkloadServer(t *testing.T, server RunningServer, org logicalcluste
 	require.NoError(t, err)
 	kubefixtures.Create(t, crdClient.ApiextensionsV1().CustomResourceDefinitions(),
 		metav1.GroupResource{Group: "apps.k8s.io", Resource: "deployments"},
+		metav1.GroupResource{Group: "core.k8s.io", Resource: "services"},
```
davidfestal, line 853: This should probably be cleaned up and managed in a distinct PR.

qiujian16: we have to create it otherwise e2e on fake downstream cluster will fail.

davidfestal: Yeah, I'm currently looking to take care of this in a distinct PR. Let's keep it here if your PR is merged before.

pkg/reconciler/workload/apibindinglabel/apibindinglabel_controller.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	apiExportInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: func(obj interface{}) bool {
+			switch t := obj.(type) {
+			case *apisv1alpha1.APIExport:
+				if len(t.Annotations) == 0 {
+					return false
+				}
+				_, ok := t.Annotations[workloadv1alpha1.InternalComputeAPIBindingAnnotationKey]
```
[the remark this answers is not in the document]
qiujian16: I actually do not see why this annotation should be removed from APIExport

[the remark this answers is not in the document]
qiujian16: updated

pkg/apis/workload/v1alpha1/types.go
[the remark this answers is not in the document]
qiujian16: change to label and remove internal

pkg/apis/workload/v1alpha1/types.go
```diff
@@ -155,4 +155,7 @@ const (
 	// InternalSyncTargetKeyLabel is an internal label set on a SyncTarget resource that contains the full hash of the SyncTargetKey, generated with the ToSyncTargetKey(..)
 	// helper func, this label is used for reverse lookups of a syncTargetKey to SyncTarget.
 	InternalSyncTargetKeyLabel = "internal.workload.kcp.dev/key"
+
+	// ComputeAPIBindingLabelKey is a annotation key to identify the APIBinding is for compute.
```
davidfestal, line 159:
```suggestion
	// ComputeAPIBindingLabel is a label set on an APIExport to identify that it is for compute.
	// It is automatically copied from an APIExport to the APIBindings bound to it.
```

pkg/apis/workload/v1alpha1/types.go
```diff
@@ -155,4 +155,7 @@ const (
...
+	ComputeAPIBindingLabelKey = "internal.workload.kcp.dev/compute"
```
davidfestal, line 160:
```suggestion
	ComputeAPIBindingLabel = "internal.workload.kcp.dev/compute"
```
In order to follow the naming pattern of `InternalSyncTargetKeyLabel`

pkg/apis/workload/v1alpha1/types.go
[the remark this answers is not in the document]
davidfestal: further commented

pkg/reconciler/workload/apibindinglabel/apibindinglabel_controller.go
```diff
@@ -0,0 +1,247 @@
...
+	if errors.IsNotFound(err) {
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+
+	labels := map[string]interface{}{} // nil means to remove the key
+	var apiExportHasLabel, apiBindingHasLabel bool
+	if len(apiExport.Labels) > 0 {
```
davidfestal, line 214: Is this test necessary ? retrieval from a nil map just works (as if the map doesn't contain anything), isn't it ?

pkg/reconciler/workload/placement/placement_reconcile.go
```diff
@@ -81,3 +82,8 @@ func (c *controller) patchPlacement(ctx context.Context, clusterName logicalclus
 	logger.WithValues("patch", string(data)).V(2).Info("patching Placement")
 	return c.kcpClusterClient.Cluster(clusterName).SchedulingV1alpha1().Placements().Patch(ctx, name, pt, data, opts, subresources...)
 }
+
+// listWorkloadAPIBindings list all apibindings in bound phase and the kind is compute
+func (c *controller) listWorkloadAPIBindings(clusterName logicalcluster.Name) ([]*apisv1alpha1.APIBinding, error) {
+	return c.apiBindingLister.Cluster(clusterName).List(labels.Set{workloadv1alpha1.ComputeAPIBindingLabelKey: ""}.AsSelector())
```
davidfestal, line 88: Can't we simply test on the existence of the label, instead of precisely checking that the value == empty string ?

pkg/reconciler/workload/placement/placement_reconcile_scheduling.go
```diff
@@ -116,12 +118,66 @@ func (r *placementSchedulingReconciler) getAllValidSyncTargetsForPlacement(clust
 		return locationWorkspace, nil, err
 	}
 
+	locationClusters, err = r.filterAPICompatible(ctx, clusterName, locationClusters)
```
davidfestal, line 121:
```suggestion
	locationSyncTargets, err = r.filterAPICompatible(ctx, clusterName, locationSyncTargets)
```
`locationClusters` is quite a old name and confusing (just besides `clusterName`)

pkg/reconciler/workload/placement/placement_reconcile_scheduling.go
```diff
@@ -116,12 +118,66 @@ func (r *placementSchedulingReconciler) getAllValidSyncTargetsForPlacement(clust
...
+		return locationWorkspace, nil, err
+	}
+
 	// find all the valid sync targets.
 	validClusters := locationreconciler.FilterNonEvicting(locationreconciler.FilterReady(locationClusters))
 
 	return locationWorkspace, validClusters, nil
 }
 
+func (r *placementSchedulingReconciler) filterAPICompatible(ctx context.Context, clusterName logicalcluster.Name, syncTargets []*workloadv1alpha1.SyncTarget) ([]*workloadv1alpha1.SyncTarget, error) {
```
davidfestal, line 132: Do you think we could have a way to surface to the Placement conditions the fact that no SyncTarget for the given location provides the expected APIs (with the list of missing APIs) ?
That would make debugging for the end-user much easier ?

test/e2e/framework/kcp.go
davidfestal: Done in PR https://github.com/kcp-dev/kcp/pull/2379/files#diff-6119c8a3bef16c15807ae827fde6e0f6b9e6d893969064b4cc24063755a90fffR848
Only `ingresses` are not added, but I don't see why we would need `ingresses` in all cases.

test/e2e/reconciler/scheduling/api_compatibility_test.go
```diff
@@ -0,0 +1,116 @@
...
+	locationClusterName := framework.NewWorkspaceFixture(t, source, orgClusterName)
+	userClusterName := framework.NewWorkspaceFixture(t, source, orgClusterName)
+
+	kcpClusterClient, err := kcpclientset.NewForConfig(source.BaseConfig(t))
+	require.NoError(t, err)
+
+	firstSyncTargetName := fmt.Sprintf("firstsynctarget-%d", +rand.Intn(1000000))
+	t.Logf("Creating a SyncTarget with no support APIExports and syncer in %s", locationClusterName)
+	_ = framework.NewSyncerFixture(t, source, locationClusterName,
+		framework.WithExtraResources("roles.rbac.authorization.k8s.io", "rolebindings.rbac.authorization.k8s.io"),
```
davidfestal, line 57:
```suggestion
```
the `roles` and `rolebindings` here should not be necessary anymore.
Fixed at the right place in PR https://github.com/kcp-dev/kcp/pull/2379/files#diff-46cd655522637a7621955fb1996a38373f7856662819a02009b111e03848fb13R374

pkg/reconciler/workload/placement/placement_reconcile_scheduling.go
```diff
@@ -111,17 +113,71 @@ func (r *placementSchedulingReconciler) getAllValidSyncTargetsForPlacement(clust
...
+	}
+
+	locationSyncTargets, err = r.filterAPICompatible(ctx, clusterName, locationSyncTargets)
 	if err != nil {
 		return locationWorkspace, nil, err
 	}
 
 	// find all the valid sync targets.
-	validClusters := locationreconciler.FilterNonEvicting(locationreconciler.FilterReady(locationClusters))
+	validClusters := locationreconciler.FilterNonEvicting(locationreconciler.FilterReady(locationSyncTargets))
```
davidfestal, line 127:
```suggestion
	validSyncTargets := locationreconciler.FilterNonEvicting(locationreconciler.FilterReady(locationSyncTargets))
```

pkg/reconciler/workload/placement/placement_reconcile_scheduling.go
```diff
@@ -99,27 +133,103 @@ func (r *placementSchedulingReconciler) getAllValidSyncTargetsForPlacement(clust
...
+				supported = false
+				break
+			}
+		}
+
+		if supported {
+			filteredSyncTarget = append(filteredSyncTarget, syncTargert)
+		} else {
+			logger.V(4).Info("Does not support APIBindings", "workspace", clusterName, "syncTarget", syncTargert.Name)
+			result.reason = append(result.reason, fmt.Sprintf("SyncTarget %s:%s does not support APIBindings", clusterName.String(), syncTargert.Name))
```
davidfestal, line 226: Could we have the list of the non-supported APIBindings in the reason ?

pkg/reconciler/workload/placement/placement_reconcile_scheduling.go
```diff
@@ -38,9 +43,17 @@ import (
...
-	getLocation    func(clusterName logicalcluster.Name, name string) (*schedulingv1alpha1.Location, error)
-	patchPlacement func(ctx context.Context, clusterName logicalcluster.Name, name string, pt types.PatchType, data []byte, opts metav1.PatchOptions, subresources ...string) (*schedulingv1alpha1.Placement, error)
+	listSyncTarget          func(clusterName logicalcluster.Name) ([]*workloadv1alpha1.SyncTarget, error)
+	listWorkloadAPIBindings func(clusterName logicalcluster.Name) ([]*apisv1alpha1.APIBinding, error)
+	getLocation             func(clusterName logicalcluster.Name, name string) (*schedulingv1alpha1.Location, error)
+	patchPlacement          func(ctx context.Context, clusterName logicalcluster.Name, name string, pt types.PatchType, data []byte, opts metav1.PatchOptions, subresources ...string) (*schedulingv1alpha1.Placement, error)
+}
+
+type schedulingResult struct {
+	clusterName      logicalcluster.Name
```
davidfestal, line 53:
```suggestion
	locationWorkspace      logicalcluster.Name
```

pkg/reconciler/workload/placement/placement_reconcile_scheduling_test.go
```diff
@@ -92,6 +101,38 @@ func TestSchedulingReconcile(t *testing.T) {
...
+			location:  newLocation("test-location"),
+			syncTargets: []*workloadv1alpha1.SyncTarget{
+				newSyncTarget("c1", true, workloadv1alpha1.ResourceToSync{GroupResource: apisv1alpha1.GroupResource{Resource: "services"}, State: workloadv1alpha1.ResourceSchemaIncompatibleState}),
+				newSyncTarget("c2", true),
+			},
+			apiBindings: []*apisv1alpha1.APIBinding{
+				newAPIBinding("kubernetes", apisv1alpha1.BoundAPIResource{Resource: "services"}),
+			},
+			wantPatch:  false,
+			wantStatus: corev1.ConditionFalse,
```
davidfestal, line 135: Can we test the value of the reason as well, to distinguish between reasons ?

test/e2e/framework/kcp.go
davidfestal: Is there a problem when the ingress is not there ?

pkg/apis/apis/v1alpha1/types_apibinding.go
```diff
@@ -235,6 +235,9 @@ const (
 	// for the request. This data is synthetic; it is not stored in etcd and instead is only applied when retrieving
 	// CRs for the CRD.
 	AnnotationAPIIdentityKey = "apis.kcp.dev/identity"
+	// AnnotationAPIExportExtraKeyPrefix is the prefix of an annotation set on an APIExport to provide extra info. The
+	// annotation with the prefix will be synced to the all the APIBindings bound to this APIExport.
+	AnnotationAPIExportExtraKeyPrefix = "extra.apis.kcp.dev/"
```
davidfestal, line 240: If by default this annotation is expected to be added to `APIExport`s and then replicated automatically to bound `APIBinding`s, why defining it in the go file related to  `APIBinding`s ? I would have expected it would be defined in `types_apiexport.go`, but maybe there's a reason I'm missing ?

@sttts I assume such an annotation was what you had initially in mind. WDYT about the syntax used here:
```
extra.apis.kcp.dev/<annotation name>
```
In our [preliminary design session](https://docs.google.com/document/d/10Tloa3IZ36PV5JJodxKeA3p65ivVqgpwmqqsOwUWewg/edit#bookmark=id.erq3116xk1sh) , the proposed syntax was `apis.kcp.dev/extra-<annotationName>`, but I tend to prefer adding `extra` in the annotation key prefix segment, and let the key name segment be available for the annotation name defined by third parties. And since the name segment of an annotation key supports dots:

> The name segment is required and must be 63 characters or less, beginning and ending with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between

that allows scoping the annotation name with the package of the third party:
```
extra.apis.kcp.dev/compute.workload.kcp.dev
```

pkg/reconciler/workload/placement/placement_reconcile_scheduling.go
```diff
@@ -51,21 +64,44 @@ func (r *placementSchedulingReconciler) reconcile(ctx context.Context, placement
...
+				return reconcileStatusContinue, placement, err
+			}
+		}
+
+		conditions.MarkFalse(placement,
+			schedulingv1alpha1.PlacementScheduled,
+			schedulingv1alpha1.ScheduleNoValidTargetReason,
+			conditionsv1alpha1.ConditionSeverityError,
+			"No valid target with reason: %s",
+			strings.Join(result.reason, ","),
```
davidfestal, line 95:
```suggestion
			strings.Join(result.reason, ", "),
```

pkg/apis/apis/v1alpha1/types_apibinding.go
sttts: Which annotation ends up on the binding? Without the extra prefix?

pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -69,6 +69,12 @@ type APIExport struct {
 	Status APIExportStatus `json:"status,omitempty"`
 }
 
+const (
+	// AnnotationAPIExportExtraKeyPrefix is the prefix of an annotation set on an APIExport to provide extra info. The
+	// annotation with the prefix will be synced to all the APIBindings bound to this APIExport.
+	AnnotationAPIExportExtraKeyPrefix = "extra.apis.kcp.dev/"
```
sttts, line 75: copied once or continously synced? If the later, can I remove?

davidfestal: continuously synced

davidfestal: And yes, you can remove it.

sttts: Please clarify in the description.

test/e2e/framework/kcp.go
davidfestal: I see, sure. Thanks for the answer.

pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -69,6 +69,12 @@ type APIExport struct {
 	Status APIExportStatus `json:"status,omitempty"`
 }
 
+const (
+	// AnnotationAPIExportExtraKeyPrefix is the prefix of an annotation set on an APIExport to provide extra info. The
+	// annotation with the prefix will be synced to all the APIBindings bound to this APIExport.
```
davidfestal, line 74:
```suggestion
	// AnnotationAPIExportExtraKeyPrefix is the prefix of an annotation set on an APIExport to provide extra info that will be made available to all APIBindings bound to this APIExport.
	// Any annotation with this prefix will be continuously synced to all the APIBindings bound to this APIExport. If the annotation is removed from the APIExport, it will also be removed from all APIBindings bound to this APIExport.
```

config/rootcompute/kube-1.24/apiexport-kubernetes.k8s.io.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   annotations:
     bootstrap.kcp.dev/create-only: "true"
     bootstrap.kcp.dev/battery: root-compute-workspace
+    extra.apis.kcp.dev/compute.workload.kcp.dev: ""
```
sttts, line 7: empty string is strange.

pkg/apis/scheduling/v1alpha1/types_placement.go
```diff
@@ -150,6 +150,18 @@ const (
 	// LocationNotMatchReason is a reason for PlacementReady condition that no matched location for
 	// this placement can be found.
 	LocationNotMatchReason = "LocationNoMatch"
+
+	// PlacementScheduled is a condition type for placement representing that a scheduling decision is
+	// made. The placement is NOT Scheduled when no valid schedule decision is available or an error
+	// occurs.
+	PlacementScheduled = "Scheduled"
```
sttts, line 157: missing type

pkg/apis/workload/v1alpha1/types.go
```diff
@@ -155,4 +155,10 @@ const (
 	// InternalSyncTargetKeyLabel is an internal label set on a SyncTarget resource that contains the full hash of the SyncTargetKey, generated with the ToSyncTargetKey(..)
 	// helper func, this label is used for reverse lookups of a syncTargetKey to SyncTarget.
 	InternalSyncTargetKeyLabel = "internal.workload.kcp.dev/key"
+
+	// ComputeAPIExportAnnotation is an annotation set on an APIExport when it will be used for compute,
+	// and its APIs are expected to be synced to a SyncTarget by the Syncer. The annotation will be continuously
+	// synced from the APIExport to all the APIBindings bound to this APIExport. The workload scheduler will
+	// check all the APIBindings with this annotation for scheduling purpose.
+	ComputeAPIExportAnnotation = "extra.apis.kcp.dev/compute.workload.kcp.dev"
```
sttts, line 163: *AnntationKey

test/e2e/reconciler/scheduling/api_compatibility_test.go
```diff
@@ -0,0 +1,98 @@
...
+	scheduledSyncTargetKey := workloadv1alpha1.ToSyncTargetKey(locationClusterName, secondSyncTargetName)
+	t.Logf("check placement should be scheduled to synctarget with supported API")
+	framework.Eventually(t, func() (bool, string) {
+		placement, err := kcpClusterClient.Cluster(userClusterName).SchedulingV1alpha1().Placements().Get(ctx, placementName, metav1.GetOptions{})
+		if err != nil {
+			return false, fmt.Sprintf("Failed to get placement: %v", err)
+		}
+
+		if len(placement.Annotations) == 0 || placement.Annotations[workloadv1alpha1.InternalSyncTargetPlacementAnnotationKey] != scheduledSyncTargetKey {
+			return false, fmt.Sprintf("Internal synctarget annotation for placement is not correct: %v", placement.Annotations)
```
sttts, line 86: say how/why it isn't correct

pkg/reconciler/workload/placement/placement_reconcile_scheduling.go
```diff
@@ -78,19 +114,17 @@ func (r *placementSchedulingReconciler) reconcile(ctx context.Context, placement
...
-
-	return reconcileStatusContinue, placement, nil
+	scheduledSyncTarget := result.validSyncTargets[rand.Intn(len(result.validSyncTargets))]
+	expectedAnnotations[workloadv1alpha1.InternalSyncTargetPlacementAnnotationKey] = workloadv1alpha1.ToSyncTargetKey(result.locationWorkspace, scheduledSyncTarget.Name)
+	updated, err := r.patchPlacementAnnotation(ctx, clusterName, placement, expectedAnnotations)
+	return reconcileStatusContinue, updated, err
 }
 
-func (r *placementSchedulingReconciler) getAllValidSyncTargetsForPlacement(clusterName logicalcluster.Name, placement *schedulingv1alpha1.Placement) (logicalcluster.Name, []*workloadv1alpha1.SyncTarget, error) {
+func (r *placementSchedulingReconciler) getAllValidSyncTargetsForPlacement(ctx context.Context, clusterName logicalcluster.Name, placement *schedulingv1alpha1.Placement) *schedulingResult {
```
sttts, line 123: embedding an error is very unusual in Golang. Why is this done?

sttts: schedulingResult is a very strange name for what it is: `getAllValidSyncTargetsForPlacement`. Either all targets, or a result of scheduling. Cannot be both.

pkg/reconciler/workload/placement/placement_reconcile_scheduling.go
```diff
@@ -51,21 +64,44 @@ func (r *placementSchedulingReconciler) reconcile(ctx context.Context, placement
...
+				return reconcileStatusContinue, placement, err
+			}
+		}
+
+		conditions.MarkFalse(placement,
+			schedulingv1alpha1.PlacementScheduled,
+			schedulingv1alpha1.ScheduleNoValidTargetReason,
+			conditionsv1alpha1.ConditionSeverityError,
+			"No valid target with reason: %s",
+			strings.Join(result.reason, ", "),
```
sttts, line 95: the reason field should be CamlCase, not comma separated, afaik

pkg/reconciler/workload/placement/placement_reconcile_scheduling.go
```diff
@@ -51,21 +64,44 @@ func (r *placementSchedulingReconciler) reconcile(ctx context.Context, placement
...
+		return reconcileStatusStop, placement, result.err
 	}
 
 	// no valid synctarget, clean the annotation.
-	if foundScheduled && len(syncTargets) == 0 {
-		expectedAnnotations[workloadv1alpha1.InternalSyncTargetPlacementAnnotationKey] = nil
-		updated, err := r.patchPlacementAnnotation(ctx, clusterName, placement, expectedAnnotations)
-		return reconcileStatusContinue, updated, err
+	if len(result.validSyncTargets) == 0 {
+		var err error
```
sttts, line 81: what is this err for?

config/rootcompute/kube-1.24/apiexport-kubernetes.k8s.io.yaml
davidfestal: yeah, at some it was a label, so it was less strange I assume.
We can decide to put `"true"` or `"false"`, but I don't see what it brings.
BTW there is a precedent it seems: see https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/workload/namespace/namespace_reconcile_placementbind.go#L59. In the code it nevers contains anything else than "" or nil (== the annotation is not there)

pkg/reconciler/workload/placement/placement_reconcile_scheduling.go
```diff
@@ -51,21 +64,44 @@ func (r *placementSchedulingReconciler) reconcile(ctx context.Context, placement
...
+		if foundScheduled {
+			expectedAnnotations[workloadv1alpha1.InternalSyncTargetPlacementAnnotationKey] = nil
+			placement, err = r.patchPlacementAnnotation(ctx, clusterName, placement, expectedAnnotations)
```
sttts, line 84: this will override other reconciler's changes, won't it? Intentional?

sttts: and worse: if the patch is a noop (can it be?) it won't trigger a requeue.

pkg/reconciler/workload/placement/placement_controller.go
```diff
@@ -262,12 +313,18 @@ func (c *controller) process(ctx context.Context, key string) error {
...
 
 	logger := logging.WithObject(klog.FromContext(ctx), obj)
 	ctx = klog.NewContext(ctx, logger)
 
 	reconcileErr := c.reconcile(ctx, obj)
+	newResource := &Resource{ObjectMeta: obj.ObjectMeta, Spec: &obj.Spec, Status: &obj.Status}
+
+	if err := c.commit(ctx, oldResource, newResource); err != nil {
+		return err
+	}
```
sttts, line 327: other controllers use this pattern, much easier to read IMO:

```golang
	old := workspace
	workspace = workspace.DeepCopy()

	logger := logging.WithObject(klog.FromContext(ctx), workspace)
	ctx = klog.NewContext(ctx, logger)

	var errs []error
	requeue, err := c.reconcile(ctx, workspace)
	if err != nil {
		errs = append(errs, err)
	}

	oldResource := &workspaceResource{ObjectMeta: old.ObjectMeta, Spec: &old.Spec, Status: &old.Status}
	newResource := &workspaceResource{ObjectMeta: workspace.ObjectMeta, Spec: &workspace.Spec, Status: &workspace.Status}
	if err := c.commit(ctx, oldResource, newResource); err != nil {
		errs = append(errs, err)
	}

	return requeue, utilerrors.NewAggregate(errs)
```

pkg/reconciler/apis/extraannotationsync/apibindinglabel_controller.go
```diff
@@ -0,0 +1,252 @@
...
+	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
+	kcpclientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned/cluster"
+	apisinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/apis/v1alpha1"
+	apislisters "github.com/kcp-dev/kcp/pkg/client/listers/apis/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/indexers"
+	"github.com/kcp-dev/kcp/pkg/logging"
+)
+
+const (
+	ControllerName = "kcp-api-export-extra-annotation-sync"
```
sttts, line 49: this looks like controller inflation. Why not add another reconciler to the existing apibinding controller.

qiujian16: existing apibinding controller only update status using committer. committer cannot update label and status once, and cannot remove the label key. If we add another reconciler in the apibinding controller, I think it will make it a bit messy, since we need to patch labels separately in this reconciler then return the updated placement to let other reconciler update status.

config/rootcompute/kube-1.24/apiexport-kubernetes.k8s.io.yaml
sttts: "true" is common

pkg/reconciler/apis/extraannotationsync/apibindinglabel_controller.go
sttts: we have precedence for that, see the clusterworkspaces controller. The only rules to follow:
1. each reconciler either modifies status or spec, not both
2. all status reconcilers come first, and reconciliation it stopped (+ requeued) when spec is changed.

pkg/reconciler/apis/extraannotationsync/apibindingannotation_controller.go
```diff
@@ -0,0 +1,252 @@
...
+
+	apiBindingInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj interface{}) { c.enqueueAPIBinding(obj, logger, "") },
+		UpdateFunc: func(_, obj interface{}) { c.enqueueAPIBinding(obj, logger, "") },
+	})
+
+	return c, nil
+}
+
+// controller reconciles sync workload.kcp.dev/compute label from APIExports to APIBindings
```
sttts, line 89: wrong go doc

sttts: and a little short

pkg/reconciler/workload/placement/placement_reconcile_scheduling.go
qiujian16: I am a bit confused, why do we need to explicit requeue after patch? if it is a noop patch, I think we do not expect requeue?

sttts: The patch func returns the updated object, which is a fresh copy of placement from etcd. Other reconcilers here in the controller might have other changes "queued" up, and those will get lost by `placement = updated` through this line.

config/rootcompute/kube-1.24/apiexport-kubernetes.k8s.io.yaml
qiujian16: done

pkg/reconciler/apis/extraannotationsync/apibindingannotation_controller.go
qiujian16: added more description

pkg/reconciler/workload/placement/placement_reconcile_scheduling.go
```diff
@@ -78,18 +89,20 @@ func (r *placementSchedulingReconciler) reconcile(ctx context.Context, placement
...
+	scheduledSyncTarget := validSyncTargets[rand.Intn(len(validSyncTargets))]
+	expectedAnnotations[workloadv1alpha1.InternalSyncTargetPlacementAnnotationKey] = workloadv1alpha1.ToSyncTargetKey(locationWorkspace, scheduledSyncTarget.Name)
+	updated, err := r.patchPlacementAnnotation(ctx, clusterName, placement, expectedAnnotations)
+	return reconcileStatusStopAndRequeue, updated, err
 }
 
-func (r *placementSchedulingReconciler) getAllValidSyncTargetsForPlacement(clusterName logicalcluster.Name, placement *schedulingv1alpha1.Placement) (logicalcluster.Name, []*workloadv1alpha1.SyncTarget, error) {
+func (r *placementSchedulingReconciler) getAllValidSyncTargetsForPlacement(ctx context.Context, clusterName logicalcluster.Name, placement *schedulingv1alpha1.Placement) (logicalcluster.Name, []*workloadv1alpha1.SyncTarget, error) {
 	if placement.Status.Phase == schedulingv1alpha1.PlacementPending || placement.Status.SelectedLocation == nil {
+		conditions.MarkFalse(placement,
```
sttts, line 100: this is a big 🔴 . Never ever mutate in a getter.

pkg/reconciler/workload/placement/placement_reconcile_scheduling.go
```diff
@@ -111,15 +130,92 @@ func (r *placementSchedulingReconciler) getAllValidSyncTargetsForPlacement(clust
...
+			}
+		}
+
+		if supported {
+			filteredSyncTargets = append(filteredSyncTargets, syncTargert)
+		}
+	}
+
+	if len(filteredSyncTargets) == 0 {
+		conditions.MarkFalse(placement,
```
sttts, line 210: another one 🔴 . Don't mutate anything when it is not clearly expected by the function name.

Issue #1981, opened 2022-09-14
Storage controller (syncer)


davidfestal:

davidfestal:
Brain-dump of the low-level implementation bits that come to my mind:
- Bootstrap the StorageController, co-located in the Syncer, dedicated to `PVC`s and `PV`s 
- Give the StorageController access to the Syncer virtual workspace upstream informer, to potentially get upstream resources (PVCs in our case)
- Setup a downstrea informer for the StorageController for PVs and PVCs that will;
    - select PVCs whose value of label `workloadv1alpha1.InternalDownstreamClusterLabel` is the current SyncTarget key
    - select all PVs.
- Have the StorageController react to downstream PVC changes:
  - when a PVC is created (initially synced from KCP), then set a `internal.workload.kcp.dev/delaysyncing` annotation to `true`, which will instruct delaying the update of the PVC status back to KCP
  - when a PVC is bound to a PV, then:
    - Get the upstream PVC corresponding to this downstream PVC (from the upstream informer)
    - Copy the NamespaceLocator annotation (found on the downstream PVC namespace) to the downstream PV (after removing the namespace since PV are cluster-wide-resources) 
    - Create a Json patch for the PV resource whose content would update the PVC reference to match the upstream PVC reference (namespace, uid, resource version, etc ...)
    - Add this JsonPatch as an annotation on the downstream PV (`internal.workload.kcp.dev/upsyncdiff`) 
    - Update the PV with the `Upsync` requestState label for the current SyncTarget to trigger the Upsyncing of the PV to the upstream workspace
- Have the StorageController react to upstream PV changes:
  - when a PV is created upstream and has the `Upsync` label for the current SyncTarget, then find the corresponding downstream PVC, and remove the `internal.workload.kcp.dev/delaysyncing` annotation
- Modify the StatusSyncer to also react to changes of the `internal.workload.kcp.dev/delaysyncing` annotation, and not react to status changes when this annotation is `true`

PR #2338, opened 2022-11-10
✨ Implement PV and PVC controllers

leseb:
## Summary

This implements two new syncers as laid out in
https://github.com/kcp-dev/kcp/issues/1981.

## Related issue(s)

Fixes: https://github.com/kcp-dev/kcp/issues/1981
Signed-off-by: Sébastien Han <seb@redhat.com>

### DIFF
--- a/pkg/syncer/shared/namespace.go
+++ b/pkg/syncer/shared/namespace.go
@@ -38,7 +38,7 @@ const (
 type NamespaceLocator struct {
 	SyncTarget  SyncTargetLocator   `json:"syncTarget"`
 	ClusterName logicalcluster.Name `json:"cluster,omitempty"`
-	Namespace   string              `json:"namespace"`
+	Namespace   string              `json:"namespace,omitempty"`
 }
 
 type SyncTargetLocator struct {
--- a/pkg/syncer/status/status_controller.go
+++ b/pkg/syncer/status/status_controller.go
@@ -137,7 +137,7 @@ func NewStatusSyncer(syncerLogger logr.Logger, syncTargetClusterName logicalclus
 				if newUnstrob.GetLabels()[workloadv1alpha1.ClusterResourceStateLabelPrefix+syncTargetKey] == string(workloadv1alpha1.ResourceStateUpsync) {
 					return
 				}
-				if !deepEqualFinalizersAndStatus(oldUnstrob, newUnstrob) {
+				if !notDeepEqualFinalizersAndStatusOrDelayStatusSyncingAnnotationChanged(oldUnstrob, newUnstrob) {
 					c.AddToQueue(gvr, newUnstrob, logger)
 				}
 			},
--- a/pkg/syncer/status/status_process.go
+++ b/pkg/syncer/status/status_process.go
@@ -38,8 +38,12 @@ import (
 	workloadcliplugin "github.com/kcp-dev/kcp/pkg/cliplugins/workload/plugin"
 	"github.com/kcp-dev/kcp/pkg/logging"
 	"github.com/kcp-dev/kcp/pkg/syncer/shared"
+	"github.com/kcp-dev/kcp/pkg/syncer/storage"
 	. "github.com/kcp-dev/kcp/tmc/pkg/logging"
 )
+
+var deepEqualFinalizersAndStatusFunc = deepEqualFinalizersAndStatus
+var hasDelayStatusSyncingAnnotationChangedFunc = hasDelayStatusSyncingAnnotationChanged
 
 func deepEqualFinalizersAndStatus(oldUnstrob, newUnstrob *unstructured.Unstructured) bool {
 	newFinalizers := newUnstrob.GetFinalizers()
@@ -49,6 +53,31 @@ func deepEqualFinalizersAndStatus(oldUnstrob, newUnstrob *unstructured.Unstructu
 	oldStatus := oldUnstrob.UnstructuredContent()["status"]
 
 	return equality.Semantic.DeepEqual(oldFinalizers, newFinalizers) && equality.Semantic.DeepEqual(oldStatus, newStatus)
+}
+
+func hasDelayStatusSyncingAnnotationChanged(oldUnstrob, newUnstrob *unstructured.Unstructured) bool {
+	// If the annotation was added
+	if !hasDelayStatusSyncingAnnotation(oldUnstrob) && hasDelayStatusSyncingAnnotation(newUnstrob) {
+		return true
+	}
+
+	// Annotation was removed
+	if hasDelayStatusSyncingAnnotation(oldUnstrob) && !hasDelayStatusSyncingAnnotation(newUnstrob) {
+		return false
+	}
+
+	// Both object have the annotation, nothing changed
+	return hasDelayStatusSyncingAnnotation(oldUnstrob) && hasDelayStatusSyncingAnnotation(newUnstrob)
+}
+
+func notDeepEqualFinalizersAndStatusOrDelayStatusSyncingAnnotationChanged(oldUnstrob, newUnstrob *unstructured.Unstructured) bool {
+	return !deepEqualFinalizersAndStatusFunc(oldUnstrob, newUnstrob) || hasDelayStatusSyncingAnnotationChangedFunc(oldUnstrob, newUnstrob)
+}
+
+func hasDelayStatusSyncingAnnotation(unstrob *unstructured.Unstructured) bool {
+	annotations := unstrob.GetAnnotations()
+	delayStatusSyncing, ok := annotations[storage.DelayStatusSyncing]
+	return ok && delayStatusSyncing == "true"
 }
 
 func (c *Controller) process(ctx context.Context, gvr schema.GroupVersionResource, key string) error {
--- a/pkg/syncer/status/status_process_test.go
+++ b/pkg/syncer/status/status_process_test.go
@@ -48,6 +48,7 @@ import (
 	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
 	ddsif "github.com/kcp-dev/kcp/pkg/informer"
 	"github.com/kcp-dev/kcp/pkg/syncer/indexers"
+	"github.com/kcp-dev/kcp/pkg/syncer/storage"
 )
 
 var scheme *runtime.Scheme
@@ -811,4 +812,87 @@ func updateDeploymentAction(namespace string, object runtime.Object, subresource
 		ActionImpl: deploymentAction("update", namespace, subresources...),
 		Object:     object,
 	}
+}
+
+func Test_hasDelayStatusSyncingAnnotation(t *testing.T) {
+	type args struct {
+		unstrob *unstructured.Unstructured
+	}
+	tests := []struct {
+		name string
+		args args
+		want bool
+	}{
+		{"no annotation", args{unstrob: &unstructured.Unstructured{}}, false},
+		{"annotation but incorrect content", args{unstrob: &unstructured.Unstructured{Object: map[string]interface{}{"metadata": map[string]interface{}{"annotations": map[string]interface{}{storage.DelayStatusSyncing: "foo"}}}}}, false},
+		{"annotation", args{unstrob: &unstructured.Unstructured{Object: map[string]interface{}{"metadata": map[string]interface{}{"annotations": map[string]interface{}{storage.DelayStatusSyncing: "true"}}}}}, true},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := hasDelayStatusSyncingAnnotation(tt.args.unstrob); got != tt.want {
+				t.Errorf("hasDelayStatusSyncingAnnotation() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func Test_hasDelayStatusSyncingAnnotationChanged(t *testing.T) {
+	type args struct {
+		oldUnstrob *unstructured.Unstructured
+		newUnstrob *unstructured.Unstructured
+	}
+	tests := []struct {
+		name string
+		args args
+		want bool
+	}{
+		{"no annotation", args{oldUnstrob: &unstructured.Unstructured{}, newUnstrob: &unstructured.Unstructured{}}, false},
+		{"annotation but incorrect content", args{oldUnstrob: &unstructured.Unstructured{Object: map[string]interface{}{"metadata": map[string]interface{}{"annotations": map[string]interface{}{storage.DelayStatusSyncing: "foo"}}}}, newUnstrob: &unstructured.Unstructured{Object: map[string]interface{}{"metadata": map[string]interface{}{"annotations": map[string]interface{}{storage.DelayStatusSyncing: "foo"}}}}}, false},
+		{"annotation is still there and hasn't changed", args{oldUnstrob: &unstructured.Unstructured{Object: map[string]interface{}{"metadata": map[string]interface{}{"annotations": map[string]interface{}{storage.DelayStatusSyncing: "true"}}}}, newUnstrob: &unstructured.Unstructured{Object: map[string]interface{}{"metadata": map[string]interface{}{"annotations": map[string]interface{}{storage.DelayStatusSyncing: "true"}}}}}, true},
+		{"annotation was removed", args{oldUnstrob: &unstructured.Unstructured{Object: map[string]interface{}{"metadata": map[string]interface{}{"annotations": map[string]interface{}{storage.DelayStatusSyncing: "true"}}}}, newUnstrob: &unstructured.Unstructured{Object: map[string]interface{}{"metadata": map[string]interface{}{"annotations": map[string]interface{}{"foo": "true"}}}}}, false},
+		{"annotation was added", args{oldUnstrob: &unstructured.Unstructured{Object: map[string]interface{}{"metadata": map[string]interface{}{"annotations": map[string]interface{}{"foo": "true"}}}}, newUnstrob: &unstructured.Unstructured{Object: map[string]interface{}{"metadata": map[string]interface{}{"annotations": map[string]interface{}{storage.DelayStatusSyncing: "true"}}}}}, true},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := hasDelayStatusSyncingAnnotationChanged(tt.args.oldUnstrob, tt.args.newUnstrob); got != tt.want {
+				t.Errorf("hasDelayStatusSyncingAnnotationChanged() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+// Both called functions have good unit tests already so this is only testing the OR case.
+func Test_notDeepEqualFinalizersAndStatusOrDelayStatusSyncingAnnotationChanged(t *testing.T) {
+	t.Run("deepEqualFinalizersAndStatusFunc is true / hasDelayStatusSyncingAnnotationChangedFunc istrue", func(t *testing.T) {
+		deepEqualFinalizersAndStatusFunc = func(oldObj, newObj *unstructured.Unstructured) bool { return true }
+		hasDelayStatusSyncingAnnotationChangedFunc = func(oldObj, newObj *unstructured.Unstructured) bool { return true }
+		want := true
+		if got := notDeepEqualFinalizersAndStatusOrDelayStatusSyncingAnnotationChanged(nil, nil); got != want {
+			t.Errorf("notDeepEqualFinalizersAndStatusOrDelayStatusSyncingAnnotationChanged() = %v, want %v", got, want)
+		}
+	})
+	t.Run("deepEqualFinalizersAndStatusFunc is false / hasDelayStatusSyncingAnnotationChangedFunc is false", func(t *testing.T) {
+		deepEqualFinalizersAndStatusFunc = func(oldObj, newObj *unstructured.Unstructured) bool { return true }
+		hasDelayStatusSyncingAnnotationChangedFunc = func(oldObj, newObj *unstructured.Unstructured) bool { return false }
+		want := false
+		if got := notDeepEqualFinalizersAndStatusOrDelayStatusSyncingAnnotationChanged(nil, nil); got != want {
+			t.Errorf("notDeepEqualFinalizersAndStatusOrDelayStatusSyncingAnnotationChanged() = %v, want %v", got, want)
+		}
+	})
+	t.Run("deepEqualFinalizersAndStatusFunc is false / hasDelayStatusSyncingAnnotationChangedFunc is true", func(t *testing.T) {
+		deepEqualFinalizersAndStatusFunc = func(oldObj, newObj *unstructured.Unstructured) bool { return false }
+		hasDelayStatusSyncingAnnotationChangedFunc = func(oldObj, newObj *unstructured.Unstructured) bool { return true }
+		want := true
+		if got := notDeepEqualFinalizersAndStatusOrDelayStatusSyncingAnnotationChanged(nil, nil); got != want {
+			t.Errorf("notDeepEqualFinalizersAndStatusOrDelayStatusSyncingAnnotationChanged() = %v, want %v", got, want)
+		}
+	})
+	t.Run("deepEqualFinalizersAndStatusFunc is true / hasDelayStatusSyncingAnnotationChangedFunc is true", func(t *testing.T) {
+		deepEqualFinalizersAndStatusFunc = func(oldObj, newObj *unstructured.Unstructured) bool { return true }
+		hasDelayStatusSyncingAnnotationChangedFunc = func(oldObj, newObj *unstructured.Unstructured) bool { return true }
+		want := true
+		if got := notDeepEqualFinalizersAndStatusOrDelayStatusSyncingAnnotationChanged(nil, nil); got != want {
+			t.Errorf("notDeepEqualFinalizersAndStatusOrDelayStatusSyncingAnnotationChanged() = %v, want %v", got, want)
+		}
+	})
 }
--- a/pkg/syncer/storage/storage_pv_controller.go
+++ b/pkg/syncer/storage/storage_pv_controller.go
@@ -1,0 +1,219 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package storage
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/go-logr/logr"
+	kcpcache "github.com/kcp-dev/apimachinery/v2/pkg/cache"
+	"github.com/kcp-dev/logicalcluster/v3"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+
+	ddsif "github.com/kcp-dev/kcp/pkg/informer"
+	"github.com/kcp-dev/kcp/pkg/logging"
+)
+
+const (
+	PersistentVolumeControllerName = "kcp-workload-syncer-storage-pv"
+)
+
+type PersistentVolumeController struct {
+	queue                                 workqueue.RateLimitingInterface
+	ddsifForUpstreamUpsyncer              *ddsif.DiscoveringDynamicSharedInformerFactory
+	ddsifForDownstream                    *ddsif.GenericDiscoveringDynamicSharedInformerFactory[cache.SharedIndexInformer, cache.GenericLister, informers.GenericInformer]
+	syncTarget                            syncTargetSpec
+	updateDownstreamPersistentVolumeClaim func(ctx context.Context, persistentVolumeClaim *corev1.PersistentVolumeClaim) (*corev1.PersistentVolumeClaim, error)
+	getDownstreamPersistentVolume         func(ctx context.Context, name string) (runtime.Object, error)
+	getDownstreamPersistentVolumeClaim    func(persistentVolumeClaimName, persistentVolumeClaimNamespace string) (runtime.Object, error)
+	getUpstreamPersistentVolume           func(clusterName logicalcluster.Name, persistentVolumeName string) (runtime.Object, error)
+}
+
+// NewPersistentVolumeController returns a new storage persistent volume syncer controller.
+func NewPersistentVolumeController(
+	syncerLogger logr.Logger,
+	syncTargetWorkspace logicalcluster.Name,
+	syncTargetName, syncTargetKey string,
+	downstreamKubeClient *kubernetes.Clientset,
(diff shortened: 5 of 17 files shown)


leseb:
I still haven't had the chance to test this code but it's ready for initial reviews. Thanks!

pkg/syncer/status/status_controller.go
```diff
@@ -85,12 +86,24 @@ func NewStatusSyncer(syncerLogger logr.Logger, syncTargetWorkspace logicalcluste
...
+						return
+					}
+
 					c.AddToQueue(gvr, obj, logger)
 				},
 				UpdateFunc: func(oldObj, newObj interface{}) {
 					oldUnstrob := oldObj.(*unstructured.Unstructured)
 					newUnstrob := newObj.(*unstructured.Unstructured)
 
+					// If the object has a delay syncing annotation we do nothing
```
davidfestal, line 101: In order not to leave an event, we should modify ` deepEqualFinalizersAndStatus` in order to also compare the value of the value of the delaying annotation, so that the removal of the annotation would enqueue of the resource again.

pkg/syncer/status/status_process.go
```diff
@@ -49,6 +50,12 @@ func deepEqualFinalizersAndStatus(oldUnstrob, newUnstrob *unstructured.Unstructu
 	return equality.Semantic.DeepEqual(oldFinalizers, newFinalizers) && equality.Semantic.DeepEqual(oldStatus, newStatus)
 }
 
+func hasDelaySyncAnnotation(unstrob *unstructured.Unstructured) bool {
```
davidfestal, line 53:
```suggestion
func hasDelayStatusSyncingAnnotation(unstrob *unstructured.Unstructured) bool {
```

pkg/syncer/storage/storage_pvc_controller.go
```diff
@@ -0,0 +1,209 @@
...
+		upstreamInformers: upstreamInformers,
+		syncTarget: syncTargetSpec{
+			name:      syncTargetName,
+			workspace: syncTargetWorkspace,
+			uid:       syncTargetUID,
+			key:       syncTargetKey,
+		},
+		updateDownstreamPersistentVolumeClaim: func(ctx context.Context, persistentVolumeClaim *corev1.PersistentVolumeClaim) (*corev1.PersistentVolumeClaim, error) {
+			var err error
+			err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
davidfestal, line 97: I wouldn't do any retry here.
When there is an error, the main process function return the error and the resource is queued again.
cf https://github.com/kcp-dev/kcp/pull/2338/files#diff-59babfb03aea1b519e11dc21eaa7df1f4e810988bb0ccf4e96a6a0cd15bb26e5R177

pkg/syncer/storage/storage_pv_controller.go
```diff
@@ -0,0 +1,184 @@
...
+		upstreamClient:    upstreamClient,
+		syncTarget: syncTargetSpec{
+			name:      syncTargetName,
+			workspace: syncTargetWorkspace,
+			uid:       syncTargetUID,
+			key:       syncTargetKey,
+		},
+		updateDownstreamPersistentVolumeClaim: func(ctx context.Context, persistentVolumeClaim *corev1.PersistentVolumeClaim) (*corev1.PersistentVolumeClaim, error) {
+			var err error
+			err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
davidfestal, line 86: I wouldn't do any retry here.
When there is an error, the main process function return the error and the resource is queued again.
cf https://github.com/kcp-dev/kcp/pull/2338/files#diff-59babfb03aea1b519e11dc21eaa7df1f4e810988bb0ccf4e96a6a0cd15bb26e5R177

pkg/syncer/storage/storage_pv_controller.go
```diff
@@ -0,0 +1,184 @@
...
+				persistentVolumeClaim, err = downstreamKubeClient.CoreV1().PersistentVolumeClaims(persistentVolumeClaim.Namespace).Update(ctx, persistentVolumeClaim, metav1.UpdateOptions{})
+				return err
+			})
+			if err != nil {
```
davidfestal, line 90: You'd rather let the caller define his own message.

pkg/syncer/storage/storage_pv_controller.go
```diff
@@ -0,0 +1,184 @@
...
+			// We don't need the namespace since the PV is cluster scoped.
+			return upstreamInformers.ForResource(pvSchemeGroupVersion).Lister().ByCluster(clusterName).Get(persistentVolumeName)
+		},
+	}
+
+	logger := logging.WithReconciler(syncerLogger, persistentVolumeControllerName)
+
+	// Add the PV informers to the controller to react to upstream PV events.
+	logger.V(2).Info("Setting up informer", pvSchemeGroupVersion.String())
+	upstreamInformers.ForResource(pvSchemeGroupVersion).Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
davidfestal, line 109: We should add a TODO here, so that we in fact point to the Upsyncer virtual workspace informer here (once @bipuladh  PR is available)

pkg/syncer/storage/storage_pv_process.go
```diff
@@ -0,0 +1,109 @@
...
+	upstreamPV, ok := upstreamPVObject.(*corev1.PersistentVolume)
+	if !ok {
+		return fmt.Errorf("failed to assert object to a PersistentVolume: %T", upstreamPVObject)
+	}
+
+	resourceState, ok := upstreamPV.GetLabels()[workloadv1alpha1.ClusterResourceStateLabelPrefix+c.syncTarget.key]
+	if ok && resourceState == string(workloadv1alpha1.ResourceStateUpsync) {
+		// Fetch the namespace locator annotation from the downstream PVC, this will be helpful to
+		// get the upstream PVC
+		nsLocatorJSON, ok := upstreamPV.GetAnnotations()[shared.NamespaceLocatorAnnotation]
```
davidfestal, line 58: The namespace locator annotation is on downstream resources, to indicate the corresponding upstream information.

davidfestal: So here you would just have to create a namespace locator containing the same information as what you downstream PV would contain. You should have all the information (upstream workspace, namespace, and synctarget info)

davidfestal: Like this (found in SpecProcess):
```
	desiredNSLocator := shared.NewNamespaceLocator(clusterName, c.syncTargetWorkspace, c.syncTargetUID, c.syncTargetName, upstreamNamespace)
	jsonNSLocator, err := json.Marshal(desiredNSLocator)
	if err != nil {
		return err
	}

	var downstreamNamespace string
       downstreamNamespaces, err := c.downstreamNSInformer.Informer().GetIndexer().ByIndex(byNamespaceLocatorIndexName, string(jsonNSLocator))
```

davidfestal: The PVC upstream namespace you would find in the upstream PV claimref.

pkg/syncer/storage/storage_pvc_controller.go
```diff
@@ -0,0 +1,209 @@
...
+			})
+			if err != nil {
+				return nil, fmt.Errorf("failed to update PersistentVolumeClaim %s/%s: %w", persistentVolumeClaim.Namespace, persistentVolumeClaim.Name, err)
+			}
+
+			return persistentVolumeClaim, nil
+		},
+		updateDownstreamPersistentVolume: func(ctx context.Context, persistentVolume *corev1.PersistentVolume) (*corev1.PersistentVolume, error) {
+			var err error
+			err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
davidfestal, line 109: Same as above

pkg/syncer/storage/storage_pvc_process.go
```diff
@@ -0,0 +1,238 @@
...
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		return err
+	}
+	logger = logger.WithValues(logging.NamespaceKey, namespace, logging.NameKey, name)
+
+	// Get the downstream PersistentVolumeClaim object for inspection
+	downstreamPVCObject, err := c.getDownstreamPersistentVolumeClaim(name, namespace)
+	if err != nil {
+		return fmt.Errorf("failed to get downstream PersistentVolumeClaim: %w", err)
```
davidfestal, line 54: We have to take in account the NoFound case here, and according to the logic, possibly ignore the reconciliation without failing.

pkg/syncer/storage/storage_pvc_process.go
```diff
@@ -0,0 +1,238 @@
...
+	downstreamPVC, ok := downstreamPVCObject.(*corev1.PersistentVolumeClaim)
+	if !ok {
+		return fmt.Errorf("failed to assert object to a PersistentVolumeClaim: %T", downstreamPVCObject)
+	}
+
+	logger.V(1).Info("processing downstream PersistentVolumeClaim")
+
+	_, ok = downstreamPVC.GetAnnotations()[DelayStatusSyncing]
+	// Is generation set to 0 when the object is created in KCP?
+	if !ok && downstreamPVC.Generation == 0 {
```
davidfestal, line 66: We cannot probably depend on generation here.
We have to think a bit more about how we st it here and unset it in the PV controller, so that both controllers wouldn't fight.

davidfestal: probably setting this whenever the PVC is not bound.

pkg/syncer/storage/storage_pvc_process.go
leseb: So this would handle the case where the object was removed during the reconcile loop?

davidfestal: yes, exactky.

pkg/syncer/shared/namespace.go
```diff
@@ -105,3 +106,22 @@ func PhysicalClusterNamespaceName(l NamespaceLocator) (string, error) {
 	// keep the namespaces short enough.
 	return fmt.Sprintf("kcp-%s", base36hash[:12]), nil
 }
+
+// IndexByNamespaceLocator is a cache.IndexFunc that indexes namespaces by the namespaceLocator annotation.
+func IndexByNamespaceLocator(obj interface{}) ([]string, error) {
```
stevekuznetsov, line 111: If this is exposed for other users, I'd expect a constant for the name of the index as well as a function to format the key as well

pkg/syncer/status/status_controller.go
```diff
@@ -84,13 +85,19 @@ func NewStatusSyncer(syncerLogger logr.Logger, syncTargetWorkspace logicalcluste
...
+					}
+
 					c.AddToQueue(gvr, obj, logger)
 				},
 				UpdateFunc: func(oldObj, newObj interface{}) {
 					oldUnstrob := oldObj.(*unstructured.Unstructured)
 					newUnstrob := newObj.(*unstructured.Unstructured)
 
-					if !deepEqualFinalizersAndStatus(oldUnstrob, newUnstrob) {
+					if notDeepEqualFinalizersAndStatusOrDelayStatusSyncingAnnotationChanged(oldUnstrob, newUnstrob) {
```
stevekuznetsov, line 100: Why does one function need to handle both cases?

pkg/syncer/status/status_controller.go
```diff
@@ -84,13 +85,19 @@ func NewStatusSyncer(syncerLogger logr.Logger, syncTargetWorkspace logicalcluste
 			logger.V(2).Info("Set up downstream informer", "gvr", gvr.String())
 			return cache.ResourceEventHandlerFuncs{
 				AddFunc: func(obj interface{}) {
+					unstrob := obj.(*unstructured.Unstructured)
+					if hasDelayStatusSyncingAnnotation(unstrob) {
```
stevekuznetsov, line 89: This kind of looks like a band-aid - why isn't it?

pkg/syncer/status/status_process_test.go
```diff
@@ -750,3 +751,86 @@ func (f *fakeSyncerInformers) SyncableGVRs() (map[schema.GroupVersionResource]*r
...
+			if got := hasDelayStatusSyncingAnnotationChanged(tt.args.oldUnstrob, tt.args.newUnstrob); got != tt.want {
+				t.Errorf("hasDelayStatusSyncingAnnotationChanged() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+// Both called functions have good unit tests already so this is only testing the OR case
+func Test_notDeepEqualFinalizersAndStatusOrDelayStatusSyncingAnnotationChanged(t *testing.T) {
+	t.Run("true / true", func(t *testing.T) {
```
stevekuznetsov, line 804: nit: test names and logging output is best written for folks that accidentally break these but have no idea how the system works - names like `"true / true"` and the `t.Errorf()` output in this test are opaque and fairly arcane, let's consider how we might make them more clear.

pkg/syncer/storage/storage_pv_controller.go
```diff
@@ -0,0 +1,210 @@
...
+		getUpstreamPersistentVolume: func(clusterName logicalcluster.Name, persistentVolumeName string) (runtime.Object, error) {
+			// We don't need the namespace since the PV is cluster scoped.
+			return upstreamInformers.ForResource(persistentVolumeSchemeGroupVersion).Lister().ByCluster(clusterName).Get(persistentVolumeName)
+		},
+		getDownstreamPersistentVolumeFromNamespaceLocator: func(namespaceLocator shared.NamespaceLocator) (runtime.Object, error) {
+			namespaceLocatorJSONBytes, err := json.Marshal(namespaceLocator)
+			if err != nil {
+				return nil, err
+			}
+			persistentVolumes, err := downstreamInformers.ForResource(persistentVolumeSchemeGroupVersion).Informer().GetIndexer().ByIndex(byNamespaceLocatorIndexName, string(namespaceLocatorJSONBytes))
```
stevekuznetsov, line 112: We should use the indices helpers here.

pkg/syncer/storage/storage_pv_controller.go
```diff
@@ -0,0 +1,210 @@
...
+			if err != nil {
+				return nil, err
+			}
+			if len(persistentVolumes) == 0 {
+				return nil, nil
+			}
+			// There should be only one persistent volume with the same namespace locator, return it.
```
stevekuznetsov, line 119: Let's handle the cases where there are no volumes, or many - better to emit some admin-visible signal here.

pkg/syncer/storage/storage_pv_process.go
```diff
@@ -0,0 +1,111 @@
...
+
+		// Remove the internal.workload.kcp.dev/delaystatussyncing annotation
+		removeAnnotationFromPersistentVolumeClaim(downstreamPVC)
+
+		_, err = c.updateDownstreamPersistentVolumeClaim(ctx, downstreamPVC)
+		if err != nil {
+			return fmt.Errorf("failed to update downstream PersistentVolumeClaim: %w", err)
+		}
+
+		logger.V(1).Info("Removed the", DelayStatusSyncing, "annotation from downstream PersistentVolumeClaim", downstreamPVC.Name)
```
stevekuznetsov, line 96: nit: logging key-value keys should not contain spaces

pkg/syncer/storage/storage_pvc_controller.go
```diff
@@ -0,0 +1,200 @@
...
+		upstreamInformers: upstreamInformers,
+		syncTarget: syncTargetSpec{
+			name:      syncTargetName,
+			workspace: syncTargetWorkspace,
+			uid:       syncTargetUID,
+			key:       syncTargetKey,
+		},
+		updateDownstreamPersistentVolumeClaim: func(ctx context.Context, persistentVolumeClaim *corev1.PersistentVolumeClaim) (*corev1.PersistentVolumeClaim, error) {
+			persistentVolumeClaim, err := downstreamKubeClient.CoreV1().PersistentVolumeClaims(persistentVolumeClaim.Namespace).Update(ctx, persistentVolumeClaim, metav1.UpdateOptions{})
+			if err != nil {
```
stevekuznetsov, line 96: nit: you can just return the values from the client call here, the `if` logic does not change anything

pkg/syncer/storage/storage_pv_controller.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return nil, err
+	}
+	c.downstreamInformer = downstreamInformers.ForResource(persistentVolumeSchemeGroupVersion)
+
+	return c, nil
+}
+
+func (c *PersistentVolumeController) AddToQueue(obj interface{}, logger logr.Logger) {
+	key, err := keyfunctions.DeletionHandlingMetaNamespaceKeyFunc(obj)
```
stevekuznetsov, line 151: If you are queueing objects from the upstream (kcp), you need the cluster-aware keyfunctions & splitters

pkg/syncer/storage/storage_pvc_process.go
```diff
@@ -0,0 +1,240 @@
...
+		upstreamClaimRef := &corev1.ObjectReference{
+			Kind:            upstreamPVC.Kind,
+			Namespace:       upstreamPVC.Namespace,
+			Name:            upstreamPVC.Name,
+			UID:             upstreamPVC.UID,
+			APIVersion:      upstreamPVC.APIVersion,
+			ResourceVersion: upstreamPVC.ResourceVersion,
+		}
+
+		// If the object being reconciled changed as a result, update it.
```
stevekuznetsov, line 124: We have a patcher utility, let's use that

pkg/syncer/status/status_controller.go
leseb: A band-aid?

pkg/syncer/status/status_process_test.go
leseb: Well it's a "simple" OR case so I didn't think too much 😅 , I'll try to think of something better.

hack/logcheck.out
```diff
@@ -1,177 +1,5 @@
...
-/pkg/virtual/workspaces/builder/org_listener.go:156:3: function "Errorf" should not be used, convert to contextual logging
-/pkg/virtual/workspaces/builder/org_listener.go:172:3: function "Errorf" should not be used, convert to contextual logging
-/pkg/virtual/workspaces/builder/org_listener.go:187:2: function "Infof" should not be used, convert to contextual logging
-/pkg/virtual/workspaces/cache/cache.go:71:4: function "Infof" should not be used, convert to contextual logging
-/pkg/virtual/workspaces/cache/cache.go:71:4: function "V" should not be used, convert to contextual logging
-/pkg/virtual/workspaces/cache/cache.go:85:2: function "Infof" should not be used, convert to contextual logging
-/pkg/virtual/workspaces/cache/cache.go:85:2: function "V" should not be used, convert to contextual logging
-/test/e2e/framework/config.go:42:2: function "InitFlags" should not be used, convert to contextual logging
+	go get github.com/kcp-dev/kcp/third_party/keyfunctions
+/pkg/syncer/storage/storage_pv_controller.go:47:2: could not import github.com/kcp-dev/kcp/third_party/keyfunctions (invalid package name: "")
```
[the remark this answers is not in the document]
leseb: I think I missed something during my last rebase, thanks for pointing it out.

pkg/syncer/status/status_controller.go
leseb: This way it's easier to return true if the annotation got removed and the finalizer/status did not change. So testing everything at once felt clearer.

pkg/syncer/storage/storage_pvc_process.go
leseb: Can you point me to it?

stevekuznetsov: Sure thing! Here: https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/committer/committer.go

pkg/syncer/status/status_controller.go
stevekuznetsov: Why are we delaying the sync? Why is it for some period of time, as opposed to until some event occurs?

pkg/syncer/storage/storage_pv_controller.go
stevekuznetsov: Under `pkg/indexers` - here: https://github.com/kcp-dev/kcp/blob/5916c3cd79ca30b3ed6e8a8983d10034f1a926e2/pkg/indexers/indexers.go#L74-L87

pkg/syncer/storage/storage_pv_process.go
leseb: I see numerous instances in the code with loggers using spaces, I'm a bit confused by this rule. Or perhaps I misunderstood what you meant?

stevekuznetsov: All of them should stop - it's bad UX for selecting things. Consider the difficulty of using `jq` to select on the key `"annotation from downstream"`

pkg/syncer/storage/storage_pvc_process.go
leseb: Confirmed offline that we need a committer that is not cluster-aware. So we can revisit this a followup PR.

pkg/syncer/storage/storage_pv_process.go
leseb: It's hard to build meaningful human-readable logs then, but anyways, thanks for clarifying.

pkg/syncer/storage/storage_pvc_process.go
leseb: Actually, I went ahead and implemented something. I'll split into individual commits at the end of the review.

pkg/syncer/storage/storage_pv_controller.go
leseb: I assume you meant `kcpcache.DeletionHandlingMetaClusterNamespaceKeyFunc()` so I went ahead and used it. Let me know if I misunderstood. Thanks!

pkg/syncer/storage/storage_pv_process.go
stevekuznetsov: Structured log _keys_ are not for humans. Expect those to be used to generate e.g. columns in tables, or for selection, etc. _Values_ can be whatever you'd like :)

pkg/syncer/storage/storage_pv_process.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+
+		logger.V(1).Info("Removed", DelayStatusSyncing, "PersistentVolumeClaim", downstreamPVC.Name)
+		return nil
+	}
+
+	return nil
+}
+
+func removeAnnotationFromPersistentVolumeClaim(downstreamPVC *corev1.PersistentVolumeClaim) {
```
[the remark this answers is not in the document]
leseb: I decided to split because the whole logic started to be quite long, also slitting up as smaller functions allow for better/smaller unit tests.

pkg/syncer/storage/storage_pvc_process.go
```diff
@@ -0,0 +1,271 @@
...
+	default:
+		return fmt.Errorf("unknown PVC phase %q", downstreamPVC.Status.Phase)
+	}
+
+	logger.V(1).Info("done processing downstream PersistentVolumeClaim")
+	return nil
+
+}
+
+func rmKeyFromJSON(jsonBlob, key string) (string, error) {
```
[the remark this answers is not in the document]
leseb: Ditto

pkg/syncer/storage/storage_pvc_process.go
```diff
@@ -0,0 +1,271 @@
...
+
+	blobToByte, err = json.Marshal(p)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal blob %q: %w", jsonBlob, err)
+	}
+
+	return string(blobToByte), nil
+}
+
+func addNsAllocatorToDownstreamPV(nsLocatorJSON string, downstreamPV *corev1.PersistentVolume) (*corev1.PersistentVolume, error) {
```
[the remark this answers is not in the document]
leseb: ditto

pkg/syncer/storage/storage_pv_process.go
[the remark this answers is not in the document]
davidfestal: I don't really see the value of unit tests on such a small amount of code honestly.
I prefer high-coverage unit tests of the controller `process` method.

pkg/reconciler/committer/committer.go
```diff
@@ -49,79 +49,112 @@ type Patcher[R runtime.Object] interface {
...
-		oldForPatch := forPatch(old)
-		// to ensure they appear in the patch as preconditions
-		oldForPatch.UID = ""
-		oldForPatch.ResourceVersion = ""
+}
 
-		oldData, err := json.Marshal(oldForPatch)
+// NewCommitterNonCluster returns a function that can patch instances of R based on meta, spec or
+// status changes using a non-cluster-aware patcher.
+func NewCommitterNonCluster[R runtime.Object, P Patcher[R], Sp any, St any](patcher Patcher[R]) func(context.Context, *Resource[Sp, St], *Resource[Sp, St]) error {
```
stevekuznetsov, line 78: We use the term `Scoped` to talk about clients inside of one cluster

pkg/syncer/shared/namespace.go
```diff
@@ -25,11 +25,14 @@ import (
...
 	"github.com/martinlindhe/base36"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
 const (
 	NamespaceLocatorAnnotation = "kcp.dev/namespace-locator"
+	// NamespaceLocatorIndexName is the name of the index that allows you to filter by namespace locator
+	NamespaceLocatorIndexName = "ns-locator"
```
stevekuznetsov, line 35: We have pkg/indexers for these helpers, let's co-locate

pkg/syncer/shared/namespace.go
```diff
@@ -105,3 +108,28 @@ func PhysicalClusterNamespaceName(l NamespaceLocator) (string, error) {
...
+		nsLocByte, err := json.Marshal(loc)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal locator %#v: %w", loc, err)
+		}
+		return []string{NamespaceLocatorIndexKey(nsLocByte)}, nil
+	}
+}
+
+// NamespaceLocatorIndexKey formats the index key for a namespace locator
+func NamespaceLocatorIndexKey(namespaceLocator []byte) string {
```
stevekuznetsov, line 133: Hm, this is not very useful. I would expect this to take `*NamespaceLocator` or similar to make it useful to callers.

pkg/reconciler/committer/committer.go
```diff
@@ -49,79 +49,112 @@ type Patcher[R runtime.Object] interface {
 	Patch(ctx context.Context, name string, pt types.PatchType, data []byte, opts metav1.PatchOptions, subresources ...string) (R, error)
 }
 
-// NewCommitter returns a function that can patch instances of R based on spec or status changes.
```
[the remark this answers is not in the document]
leseb: Is a separate commit accessible too? That's my goal once the reviews are done and the PR has a few approvals. Thanks.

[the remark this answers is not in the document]
leseb: https://github.com/kcp-dev/kcp/pull/2458

test/e2e/virtual/syncer/storage_pv_test.go
```diff
@@ -0,0 +1,281 @@
...
+						},
+					},
+				}
+
+				downstreamPVC, err = syncer.DownstreamKubeClient.CoreV1().PersistentVolumeClaims(downstreamNamespaceName).Create(ctx, downstreamPVC, metav1.CreateOptions{})
+				require.NoError(t, err)
+
+				// Create DownstreamPV linked to the downstream PVC
+
+				pvNamespaceLocator := shared.NewNamespaceLocator(logicalcluster.Name(workspaceName.RequestPath()), syncer.SyncTargetClusterName, types.UID(syncer.SyncerConfig.SyncTargetUID), syncer.SyncerConfig.SyncTargetName, "")
```
davidfestal, line 117: Using `.RequestPath()` is wrong here. It's going to put a URL where there should be a logical cluster `Name` (the unique ID).
I think `workspaceName` should be a logicalCluster.Name, and you could also pass the `workspacePath` as an argument to test cases as well. But these are 2 distinct things.

test/e2e/virtual/syncer/storage_pv_test.go
```diff
@@ -0,0 +1,281 @@
...
+		t.Run(testCase.name, func(t *testing.T) {
+			t.Parallel()
+			framework.Suite(t, "transparent-multi-cluster")
+
+			ctx, cancelFunc := context.WithCancel(context.Background())
+			t.Cleanup(cancelFunc)
+
+			orgPath, _ := framework.NewOrganizationFixture(t, server)
+
+			controllerPath, _ := framework.NewWorkspaceFixture(t, server, orgPath, framework.WithName("pv-controller"))
```
davidfestal, line 226: Yuu will need both the workspacePath, and the workspaceName, as you did in the PVC e2e test: 
https://github.com/kcp-dev/kcp/pull/2338/files#diff-c0313ded30e0be3ab94b6a461c4e05751f9e6b192b40afd57d93017130588219R254-R255

test/e2e/virtual/syncer/storage_pv_test.go
```diff
@@ -0,0 +1,291 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package syncer
+
```
davidfestal, line 18: Could we move those 2 tests in a new `/test/e2e/syncer/storage` package ?

They don't properly test the syncer **virtual workspace** feature, but new controllers added inside the Syncer itself.

davidfestal:
I created PR https://github.com/leseb/kcp/pull/2 against your branch (since it seems I cannot push directly to your PR forked branch). This fixes the pv e2e test.

davidfestal (changes requested):
Before merging this PR, we should add a feature flag to enable these actions: `KCPStorageMove`.

- See here for defining the new feature flag: https://github.com/leseb/kcp/blob/58ee6d48c9b5c3f4a69d076a954344785ac42505/pkg/features/kcp_features.go#L50

- See here for testing that feature flag presence in the syncer:  https://github.com/leseb/kcp/blob/58ee6d48c9b5c3f4a69d076a954344785ac42505/pkg/syncer/syncer.go#L424

- see here for setting the feature flag before a specific e2e test (which you will need): https://github.com/leseb/kcp/blob/58ee6d48c9b5c3f4a69d076a954344785ac42505/test/e2e/syncer/tunnels_test.go#L56

test/e2e/virtual/syncer/storage_pv_test.go
```diff
@@ -0,0 +1,291 @@
...
+				// Create DownstreamPV linked to the downstream PVC
+				upstreamKcpClient, err := kcpclientset.NewForConfig(syncer.SyncerConfig.UpstreamConfig)
+				require.NoError(t, err)
+
+				syncTarget, err := upstreamKcpClient.Cluster(wsPath).WorkloadV1alpha1().SyncTargets().Get(ctx,
+					syncer.SyncerConfig.SyncTargetName,
+					metav1.GetOptions{},
+				)
+				require.NoError(t, err)
+				pvNamespaceLocator := shared.NewNamespaceLocator(wsClusterName, logicalcluster.From(syncTarget),
```
davidfestal, line 127: strange that you need to take the logical cluster and UID from the syncTarget, though you can get them from the SyncerConfig, as you did above in line [79](https://github.com/kcp-dev/kcp/pull/2338/files#diff-aeaf51b0a6e9e0efee0df9bfca87018758b76a8eae5eca1da7695c7ce2949c9dR79).

Would be better if the method wold be consistent.

pkg/syncer/syncer.go
```diff
@@ -352,10 +369,47 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 				return informers, notSynced
 			},
 		},
-		map[string]controllermanager.ManagedController{},
+		upstreamUpSyncerManagedControllers,
 	)
-	go upstreamUpsyncerControllerManager.Start(ctx)
+	if kcpfeatures.DefaultFeatureGate.Enabled(kcpfeatures.SyncerStorage) {
```
davidfestal, line 374: let's start the upstreamUpsyncerControllerManager in any case, and not condition that to the feature gate.

davidfestal:
Do you think you could rebase the PR on top of main (should not hurt I assume) ?
A number of flakes on e2d tests have been fixed, so that it would be easier to understand if the failing tests in this PR come from this PR changes, or from old flakes.

PR #2341, opened 2022-11-10
✨  relax workspace object name validation

mjudeikis:
## Summary

Relaxes workspce object name validation accross the board to be able to start with digit. This allows use of GUID libaries to generate names. In addition this brings us closer to k8s upstream and RFC 1123. But now (not until workspaces are flatened) we are not abel fully re-use k8s validation 

## Related issue(s)

Fixes #


mjudeikis:
ping @ncdc

pkg/server/filters/filters.go
```diff
@@ -49,7 +49,11 @@ const (
 )
 
 var (
-	reClusterName = regexp.MustCompile(`^([a-z]([a-z0-9-]{0,61}[a-z0-9])?:)*[a-z]([a-z0-9-]{0,61}[a-z0-9])?$`)
+	// reClusterName is a regular expression for cluster names. It is based on
+	// modified RFC 1123. It allows for 63 characters for single name and includes
+	// KCP specific ':' separator for workspace nesting. We are not re-using k8s
+	// validation regex because it purpose is for single name validation
```
ncdc, line 55:
```suggestion
	// validation regex because its purpose is for single name validation
```

ncdc:
Just 1 typo in a comment

pkg/server/filters/filters.go
mjudeikis: :see_no_evil: :keyboard: :bomb:

ncdc (approved):

mjudeikis:
bye bye:
```
for {
	user.Name = uuid.New().String()
	// HACK: we can't start with digit for now. This will change in the future
	if !unicode.IsDigit(rune(user.Name[0])) {
	    break
	}
}
```

PR #2345, opened 2022-11-10
🐛 Fix unauthorized read of Custom Resources

ncdc:
## Summary
Port of the fix for https://github.com/kubernetes/kubernetes/issues/113756 (https://github.com/kubernetes/kubernetes/pull/113685)

## Related issue(s)

Fixes #


ncdc:
Note, this won't be needed if we get the 1.26 rebase in first.

ncdc:
Working on the rebase so we won't need this any more

PR #2355, opened 2022-11-14
✨ Version check between the compiled syncer and the deployment yaml

### LINKED ISSUE #2351: `kcp workload sync` command should put a recommended syncer version in the deployment, and the `syncer` binary should verify it matches.

davidfestal:

## Summary

The `kcp workload sync` command now adds an `--expectedKCPVersion` flag to the `syncer` areguments in the deployment, and the `syncer` binary verifies it matches its compiled version.

## Related issue(s)

Fixes kcp-dev/contrib-tmc#82

### DIFF
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -39,6 +39,8 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 
+	conditionsapi "github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/apis/conditions/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/util/conditions"
 	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
 	kcpclusterclientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned/cluster"
 	kcpinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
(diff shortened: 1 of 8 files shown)


pkg/syncer/syncer.go
```diff
@@ -83,13 +86,23 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 	logger.V(2).Info("starting syncer")
 
 	kcpVersion := version.Get().GitVersion
```
stevekuznetsov, line 88: This is not the `kcp` version, but the compiled version of the `syncer`, right? Are you attempting to add a check in this PR that the compiled version of the `syncer` matches the compiled version of the `kubectl kcp workload sync` command? The comments & code read like you're checking the `kcp` server version, which I don't think is what this is doing?

davidfestal: > Are you attempting to add a check in this PR that the compiled version of the syncer matches the compiled version of the kubectl kcp workload sync command
That's precisely what I'm doing yes, afaict.

> The comments & code read like you're checking the kcp server version

Can you point me to the comments you find confusing in this regard ?

stevekuznetsov: The variable in the code line that this comment is on is called `kcpVersion` - it's not the version of `kcp`. It's the version of the `syncer` :)

davidfestal: This variable was named this way previously, since the Syncer used to be compiled with KCP I assume. But I can rename it to make it clearer in this context.

davidfestal: Done

mjudeikis:
/close 
TMC

PR #2370, opened 2022-11-16
🐛 syncer: Fix wrong logic in dns-related process

davidfestal:
## Summary

We should:
- trigger the creation of the CoreDNS deployment (if not already created) by the Syncer on any resource syncing, at the condition that deployments are part of the synced GVRs for this synced target.
- wait for the CoreDNS deployment to be reachable through a service before syncing a resource, but only when syncing a deployment. In this case, the deployment resource should be requeued after a small delay.

## Related issue(s)

Fixes flaky e2e tests

### DIFF
--- a/pkg/syncer/spec/dns/dns_process.go
+++ b/pkg/syncer/spec/dns/dns_process.go
@@ -18,7 +18,6 @@ package dns
 
 import (
 	"context"
-	"errors"
 
 	"github.com/kcp-dev/logicalcluster/v2"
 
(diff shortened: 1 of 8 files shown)


pkg/syncer/spec/spec_process.go
```diff
@@ -171,43 +172,63 @@ func (c *Controller) process(ctx context.Context, gvr schema.GroupVersionResourc
...
+		Version:  "v1",
+		Resource: "deployments",
+	}
+	deploymentsSynced := false
+	if gvr == deploymentGVR {
+		deploymentsSynced = true
+	}
+
+	if _, exists := c.syncerInformers.InformerForResource(deploymentGVR); exists {
+		deploymentsSynced = true
```
stevekuznetsov, line 209: What's the meaning of checking for the informer existing, but not being synced?

davidfestal: if deployments are in the GVRs mentioned in the `SyncedResources` status field of the `SyncedTarget`, there will be an informer started for deployments in `syncerInformers`.
In such a case, let's not wait for the first deployment instance to come, and start the DNS Deployment as soon as possible, so that it will be already started when the first deployment comes, and we don't need to wait before syncing the deployment.

stevekuznetsov: Sure - my point is that if you start doing this before the informer is synced you may have weird behavior

davidfestal: Not sure understand why. I don't need the related informer here. In an informer has been started for deployments the means that deployments will very probably be synced by this Syncer in the current workspace, and will need an up and running CoreDNS POD dedicated to this workspace. So let's create it now.

stevekuznetsov: Hm - can we write this logic explicitly? Do we not have access to the configuration? Can we pass it down? Using an implicit check for a side-effect is hard to reason about.

davidfestal: > Do we not have access to the configuration

What configuration are you speaking of ? The GVRs to be synced are dynamic and come from the current SyncTarget status.

davidfestal: Let me add a comment.

pkg/syncer/spec/dns/dns_process.go
```diff
@@ -79,60 +78,66 @@ func NewDNSProcessor(
...
+			return false, err
+		}
+		if err := d.processService(ctx, dnsID); err != nil {
+			return false, err
+		}
+		ready = false
 	}
 
-	return nil
+	if endpoints != nil && !hasAtLeastOneReadyAddress(endpoints) {
```
lionelvillard, line 122: what about moving this just after getting the endpoint? I would make the code easier to read, at least for me :-)

pkg/syncer/spec/spec_process.go
davidfestal: I removed the check BTW since having a `SyncTarget` that do not expect to sync `deployment`s is going to be very rare, and probably not worth a premature optimization.

pkg/syncer/spec/dns/dns_process.go
davidfestal: Done

pkg/syncer/spec/dns/dns_process.go
```diff
@@ -147,15 +152,13 @@ func (d *DNSProcessor) processRole(ctx context.Context, name string) error {
 
 	expected := MakeRole(name, d.dnsNamespace)
```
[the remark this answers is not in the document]
davidfestal: Sure. while we're at it we should probably do that, especially if the update of the those resources is not done here.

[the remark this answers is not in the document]
davidfestal: done

lionelvillard:
I did some local testing, all work as expected. LGTM

davidfestal:
Seems this is not sufficient to fix the errors in the e2e tests. Let's postpone the merge here.

PR #2374, opened 2022-11-16
🐛Syncer namespace cleanup refactor

jmprusi:
## Summary

Refactor how the syncer handles the deletion of downstream namespaces.

* Removes the upstream namespace controller from syncer.
* Fixes an incorrect use of indexes. 
* Surface the syncer `--downstream-namespace-clean-delay` flag and set it to 2s for tests.
* Fixes an incorrect annotation removal on a test.


jmprusi (commented):
-

pkg/syncer/resourcesync/controller.go
```diff
@@ -466,3 +463,11 @@ func getAllGVRs(synctarget *workloadv1alpha1.SyncTarget) map[schema.GroupVersion
 
 	return gvrs
 }
+
+func IndexByNamespace(obj interface{}) ([]string, error) {
```
stevekuznetsov, line 467: Can we please use the upstream index.

jmprusi: thanks! done!

davidfestal (approved):

Issue #2332, opened 2022-11-09
feature: APIExportEndpointSlice reconciliation


fgiloux:
### Feature Description

This feature request follows #2331 

The aim is the implementation of the APIExportEndpointSlice reconciliation. As a first step its status should get populated with the endpoints for all the Shards of the referenced APIExport

### Proposed Solution

[API design](https://hackmd.io/EuPmah62QHSEs-po04Kn1Q?view)

### Alternative Solutions

Alternatives have been discussed in this [Google doc](https://docs.google.com/document/d/1zI9KOpLp6H_tAcKpGbLpQu_iUK9VbpgUSo0OFEd7Mig/edit#heading=h.7cfr9xt3gzbw)

### Want to contribute?

- [X] I would like to work on this issue.

### Additional Tasks

Follow-ups of #2432

- Permissions: The export reference must not expose existence of an APIExport or even the target workspace. Plus, we have to validate that you own the export (content subresource).
- Add end-to-end tests
- Switch all existing uses of the endpoints in APIExport to APIExportEndpointSlice.

fgiloux:
Follow-ups of #2432

- Permissions: The export reference must not expose existence of an APIExport or even the target workspace. Plus, we have to validate that you own the export (content subresource).
- Add end-to-end tests
- Switch all existing uses of the endpoints in APIExport to APIExportEndpointSlice.

fgiloux:
Closing as
- the feature functionalities have been implemented
- an admission webhook has been created for checking permissions
- end-to-end tests have been added
- there is no reference on the server side to the endpoints in APIExports.

PR #2381, opened 2022-11-18
:sparkles: Adding ClusterWorkspaceShard to the resources stored in the cache server

fgiloux:
## Summary

ClusterWorkspaceShards are created and stored in the root workspace. They contain the URLs to access the shards. This information is necessary to construct the URLs for accessing APIExport services that will get stored into the status of APIExportEndpointSlices.

APIExports and APIExportEndpointSlices can be created in any workspace in any shard. This is a reason for making  
ClusterWorkspaceShards globally available. 

The number of ClusterWorkspaceShards is also not high and changes should be rare, hence the number of writes and the cost should be low.

## Related issue(s)

#2332


pkg/cache/server/bootstrap/bootstrap.go
```diff
@@ -43,9 +43,13 @@ const SystemCacheServerShard = "system:cache:server"
 
 func Bootstrap(ctx context.Context, apiExtensionsClusterClient kcpapiextensionsclientset.ClusterInterface) error {
 	crds := []*apiextensionsv1.CustomResourceDefinition{}
-	for _, resource := range []string{"apiresourceschemas", "apiexports"} {
+	for _, resource := range []struct{ g, n string }{
```
p0lyn0mial, line 46: nit: `for _, gr := range []struct{group, resource string}{` ?

test/e2e/reconciler/cache/replication_test.go
```diff
@@ -251,6 +253,101 @@ func replicateAPIExportNegativeScenario(ctx context.Context, t *testing.T, serve
...
+	scenario := &replicateResourceScenario{resourceName: resourceName, resourceKind: "ClusterWorkspaceShard", server: server, kcpShardClusterClient: kcpShardClusterClient, cacheKcpClusterClient: cacheKcpClusterClient}
+
+	t.Logf("Create source ClusterWorkspaceShard %s/%s on the root shard for replication", cluster, resourceName)
+	scenario.CreateSourceResource(t, func() error {
+		cws := &tenancyv1alpha1.ClusterWorkspaceShard{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: resourceName,
+			},
+			Spec: tenancyv1alpha1.ClusterWorkspaceShardSpec{
+				BaseURL: "https://base.kcp.test.dev",
```
p0lyn0mial, line 271: I guess it is okay for now, it might break when this test is run on a kcp instance that is shared with other tests.

p0lyn0mial (commented):
looks very good.
I'd like to get approval from @sttts before we merge it.

test/e2e/reconciler/cache/replication_test.go
ncdc: This is something worth avoiding up front. What about using GenerateName instead of Name?

p0lyn0mial: I think we need a real address or a way to disable this shard from scheduling. We might try to put something on that shard in a multi shard env.

Maybe for this particular test scenario we should actually use `framework.PrivateKcpServer` ?

fgiloux: My gut feeling is that we should build the tests and the business logic in a way that is not sensible to that. If a node is not ready in Kubernetes no new pod gets scheduled onto it. I am wondering whether we would need a similar mechanism in kcp for the shards at some point.
Nevertheless we should forge the multi-shard tests so that the placement is constrained to a shard, which is know to be functional (or not for negative tests). It is good to have as few explicit or implicit dependencies between the test cases.
The alternative is to start an additional kcp server to have a "real address", which may make the tests slower. I would favor the other approach for this reason. What do you think?

stevekuznetsov: NACK to private KCP in every case. Why don't we have shard scheduling control?

p0lyn0mial: Would it be okay to add an annotation to the shard created by this test that would tell the scheduler to not consider the shard?

I think we would have to update the `isValidShard`(https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/tenancy/clusterworkspace/clusterworkspace_reconcile_scheduling.go#L229) function to look up for the special annotation.

stevekuznetsov: That might be a good approach in the near-term, but do we have any idea if we're going to add some sort of scheduling attributes to the shards for the long-term?

pkg/reconciler/cache/replication/replication_reconcile_test.go
```diff
@@ -48,49 +48,49 @@ func init() {
 func TestReconcileAPIExports(t *testing.T) {
 	scenarios := []struct {
 		name                                     string
-		initialLocalApiExports                   []runtime.Object
```
stevekuznetsov, line 51: nit: in the future, breaking spelling changes or other unrelated work into a separate commit aids in review

test/e2e/reconciler/cache/replication_test.go
```diff
@@ -445,8 +548,10 @@ func (b *replicateResourceScenario) getCachedResourceHelper(ctx context.Context,
 		return b.cacheKcpClusterClient.Cluster(cluster).ApisV1alpha1().APIExports().Get(cacheclient.WithShardInContext(ctx, shard.New("root")), b.resourceName, metav1.GetOptions{})
 	case "APIResourceSchema":
 		return b.cacheKcpClusterClient.Cluster(cluster).ApisV1alpha1().APIResourceSchemas().Get(cacheclient.WithShardInContext(ctx, shard.New("root")), b.resourceName, metav1.GetOptions{})
+	case "ClusterWorkspaceShard":
+		return b.cacheKcpClusterClient.Cluster(cluster).TenancyV1alpha1().ClusterWorkspaceShards().Get(cacheclient.WithShardInContext(ctx, shard.New("root")), b.resourceName, metav1.GetOptions{})
```
stevekuznetsov, line 552: broader question: why are we using strongly typed (as opposed to dynamic) clients and informers in here?

stevekuznetsov (commented):
Broadly LGTM

test/e2e/reconciler/cache/replication_test.go
p0lyn0mial: leads to more readable code in general and we also make sure that strongly typed clients/informers can be used with the cache server

test/e2e/reconciler/cache/replication_test.go
p0lyn0mial: I think that by default we will schedule randomly and we will allow for specifying a selector for those that have opinions

test/e2e/reconciler/cache/replication_test.go
stevekuznetsov: Are we doing special logic for each? Or the same? Convince me that we're not going to gain a case in this switch statement for every type ... ?

p0lyn0mial: We won't because caching is expensive :) 

My main goal was to test the kcp client with the cache.
Plus individual tests read well (i.e. https://github.com/kcp-dev/kcp/blob/93e69afd3c82f090282e150ac76eba5a7ffc8dc5/test/e2e/reconciler/cache/replication_test.go#L165). 

If we were to use the dynamic client we would have to use `unstructured`

p0lyn0mial: > Also was concerned we're re-writing tons and tons of boilerplate for each type when the logic to handle it is identical

Usually creation and a spec update are different. Initially I captured it with a `baseScenario` idea but Stefan didn't like it (https://github.com/kcp-dev/kcp/pull/2240#discussion_r1012810357).  

Maybe we could have a separate test for testing only the kcp client and then change the replication tests to be more dynamic. I think I could do that, wdyt?

pkg/reconciler/cache/replication/replication_reconcile.go
```diff
@@ -47,21 +48,32 @@ func (c *controller) reconcile(ctx context.Context, gvrKey string) error {
 			apisv1alpha1.SchemeGroupVersion.WithResource("apiexports"),
 			apisv1alpha1.SchemeGroupVersion.WithKind("APIExport"),
 			func(gvr schema.GroupVersionResource, cluster logicalcluster.Name, namespace, name string) (interface{}, error) {
-				return retrieveCacheObject(&gvr, c.cacheApiExportsIndexer, c.shardName, cluster, namespace, name)
+				return retrieveCacheObject(&gvr, c.cacheAPIExportsIndexer, c.shardName, cluster, namespace, name)
 			},
 			func(cluster logicalcluster.Name, _, name string) (interface{}, error) {
-				return c.localApiExportLister.Cluster(cluster).Get(name)
+				return c.localAPIExportLister.Cluster(cluster).Get(name)
```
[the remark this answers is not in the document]
fgiloux: too many caches. If we start renaming I would prefer global or crossShards vs cached.

pkg/reconciler/cache/replication/replication_reconcile_test.go
fgiloux: ack

pkg/reconciler/cache/replication/replication_controller.go
```diff
@@ -80,10 +82,19 @@ func NewController(
 		return nil, err
 	}
 
+	if err := cacheKcpInformers.Tenancy().V1alpha1().ClusterWorkspaceShards().Informer().AddIndexers(cache.Indexers{
```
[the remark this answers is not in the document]
p0lyn0mial: I didn't know we have `indexers.AddIfNotPresentOrDie`. I don't like its side effects:

- it removes entries from `toAdd` map
- it panics on error, why it cannot simply return an error?

test/e2e/reconciler/cache/replication_test.go
stevekuznetsov: Whichever you think is better.

p0lyn0mial: okay, i can prepare something and show it you, we can always trash it if we don't like it.

p0lyn0mial: this can be done in a separate PR, let's not block this pr on it.

stevekuznetsov:
Nope, this looks good to me - @p0lyn0mial PTAL

/lgtm
/approve
/hold

Issue #2274, opened 2022-10-27
bug: e2e-sharded failed to list *v1alpha1.ClusterWorkspace: Unauthorized


hardys:
### Describe the bug

In `e2e-sharded` runs we're seeing an error like this (this doesn't actually stop the job from passing so perhaps we need to look at additional test coverage)

```
PROXY  W1025 18:20:39.100256   31146 reflector.go:324] k8s.io/client-go@v0.0.0-20221025160842-38c73163e766/tools/cache/reflector.go:167: failed to list *v1alpha1.ClusterWorkspace: Unauthorized
PROXY  E1025 18:20:39.101457   31146 reflector.go:138] k8s.io/client-go@v0.0.0-20221025160842-38c73163e766/tools/cache/reflector.go:167: Failed to watch *v1alpha1.ClusterWorkspace: failed to list *v1alpha1.ClusterWorkspace: Unauthorized
```

After debugging locally, it appears to be because we're using the root shard-admin token for all shards [when creating the clients](https://github.com/kcp-dev/kcp/blob/main/pkg/proxy/server.go#L65) used by the clusterworkspace informer

We can see here in some local debug log output  that the shard-admin token hitting `kcp-1` is actually the token from `kcp-0`:

```
$ grep "invalid bearer token" kcp-1.log | head -n1
[...]

PR #2382, opened 2022-11-18
🐛 give the front proxy a distinct config for direct(internal) shard communication.

p0lyn0mial:
## Summary
The front proxy needs a direct and privileged way of authNZ with any shard. At the same time it needs to be able to validate an incoming certificate from a shard.

This PR allows the above by introducing a kubeconfig that holds a client certificate and a CA certificate for validating incoming certs.
The new kubeconfig can be passed via `shards-kubeconfig` flag.

Previously an incorrect configuration was used for communication with shards. As a result the proxy didn't work in a multi shard environment.

## Related issue(s)

required by https://github.com/kcp-dev/kcp/issues/342

fixes https://github.com/kcp-dev/kcp/issues/2274

### DIFF
--- a/pkg/proxy/config.go
+++ b/pkg/proxy/config.go
@@ -46,6 +46,7 @@ type ExtraConfig struct {
 	// the clients can wildcard-list/watch most kcp resources.
 	ResolveIdentities func(ctx context.Context) error
 	RootShardConfig   *rest.Config
+	ShardsConfig      *rest.Config
 
 	AuthenticationInfo    genericapiserver.AuthenticationInfo
 	ServingInfo           *genericapiserver.SecureServingInfo
@@ -85,6 +86,11 @@ func NewConfig(opts *proxyoptions.Options) (*Config, error) {
 	}
 	if err := c.Options.Authentication.ApplyTo(&c.AuthenticationInfo, c.ServingInfo, c.RootShardConfig); err != nil {
 		return nil, err
+	}
+
+	c.ShardsConfig, err = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(&clientcmd.ClientConfigLoadingRules{ExplicitPath: c.Options.ShardsKubeconfig}, nil).ClientConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to load shard kubeconfig: %w", err)
 	}
 
 	c.AdditionalAuthEnabled = c.Options.Authentication.AdditionalAuthEnabled()
--- a/pkg/proxy/config.go
+++ b/pkg/proxy/config.go
@@ -19,10 +19,12 @@ package proxy
 import (
 	"context"
 	"fmt"
+	"net/http"
 
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 
 	proxyoptions "github.com/kcp-dev/kcp/pkg/proxy/options"
 	bootstrap "github.com/kcp-dev/kcp/pkg/server/bootstrap"
@@ -78,7 +80,10 @@ func NewConfig(opts *proxyoptions.Options) (*Config, error) {
 		return nil, fmt.Errorf("failed to load root kubeconfig: %w", err)
 	}
 
-	c.RootShardConfig, c.ResolveIdentities = bootstrap.NewConfigWithWildcardIdentities(nonIdentityRootConfig, bootstrap.KcpRootGroupExportNames, bootstrap.KcpRootGroupResourceExportNames, nil)
+	var kcpShardIdentityRoundTripper func(rt http.RoundTripper) http.RoundTripper
+	kcpShardIdentityRoundTripper, c.ResolveIdentities = bootstrap.NewWildcardIdentitiesWrappingRoundTripper(bootstrap.KcpRootGroupExportNames, bootstrap.KcpRootGroupResourceExportNames, nonIdentityRootConfig, nil)
+	c.RootShardConfig = rest.CopyConfig(nonIdentityRootConfig)
+	c.RootShardConfig.Wrap(kcpShardIdentityRoundTripper)
 
 	var loopbackClientConfig *rest.Config
 	if err := c.Options.SecureServing.ApplyTo(&c.ServingInfo, &loopbackClientConfig); err != nil {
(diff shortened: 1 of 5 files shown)


pkg/proxy/config.go
```diff
@@ -87,6 +93,16 @@ func NewConfig(opts *proxyoptions.Options) (*Config, error) {
 		return nil, err
 	}
 
+	c.ShardsConfig, err = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: c.Options.ShardsKubeconfig},
+		// We override the Server here so that the user doesn't have to specify unused server value
+		// The Server must have HTTPS scheme otherwise CA won't be loaded (see IsConfigTransportTLS method)
```
p0lyn0mial, line 99: alternatively we could provide certificates via separate flags and construct the config manually

pkg/proxy/options/options.go
```diff
@@ -88,6 +90,9 @@ func (o *Options) Validate() []error {
 	if o.MappingFile == "" {
 		errs = append(errs, fmt.Errorf("--mapping-file is required"))
 	}
+	if len(o.ShardsKubeconfig) == 0 {
+		errs = append(errs, fmt.Errorf("--shards-kubeconfig is required"))
```
p0lyn0mial, line 94: the flag is now required

pkg/proxy/config.go
ncdc: That would probably be my preference, but I'm ok merging as-is and adjusting later if we decide to go that route, unless @sttts @stevekuznetsov prefer making the change now.

pkg/proxy/options/options.go
```diff
@@ -56,6 +57,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.MappingFile, "mapping-file", o.MappingFile, "Config file mapping paths to backends")
 	fs.StringVar(&o.RootDirectory, "root-directory", o.RootDirectory, "Root directory.")
 	fs.StringVar(&o.RootKubeconfig, "root-kubeconfig", o.RootKubeconfig, "The path to the kubeconfig of the root shard.")
+	fs.StringVar(&o.ShardsKubeconfig, "shards-kubeconfig", o.ShardsKubeconfig, "The path to the kubeconfig used for communication with other shards.")
```
hardys, line 60: nit - AFAICS these credentials are used to create ClusterWorkspace informers for all shards (including root), so I'd s/other/all here 

It's still a little confusing given we also require root-kubeconfig, but perhaps that can be deprecated/removed in a future iteration

pkg/proxy/config.go
```diff
@@ -88,10 +93,15 @@ func NewConfig(opts *proxyoptions.Options) (*Config, error) {
 		return nil, err
 	}
 
-	c.ShardsConfig, err = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(&clientcmd.ClientConfigLoadingRules{ExplicitPath: c.Options.ShardsKubeconfig}, nil).ClientConfig()
+	c.ShardsConfig, err = clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: c.Options.ShardsKubeconfig},
+		// We override the Server here so that the user doesn't have to specify unused server value
+		// The Server must have HTTPS scheme otherwise CA won't be loaded (see IsConfigTransportTLS method)
+		&clientcmd.ConfigOverrides{ClusterInfo: clientcmdapi.Cluster{Server: "https://fakeserver.io"}}).ClientConfig()
```
hardys, line 100: The fakeserver overrided here looks strange, is that needed, or can we just use whatever is specified in the shard kubeconfig?

hardys: Oh I see now that the server is empty in the file, if it can't remain empty here would it perhaps be better to default it to the root-shard URL?

pkg/proxy/config.go
p0lyn0mial: It would require at least 2 flags, one for the client cert pair and the other one for the public CA's certificate.

On the other hand kubeconfig has an advantage of specifying other means of authNZ, although I'm not sure we would ever use i.e. tokens.

pkg/proxy/config.go
p0lyn0mial: yes, please see https://github.com/kcp-dev/kcp/pull/2382#discussion_r1026411277 for alternatives.

pkg/proxy/options/options.go
p0lyn0mial: yeah, i'm not sure. In general we need a privileged credentials for making wildcard requests against all shards. In addition to that we also need access to the root cluster for identities.

pkg/proxy/config.go
hardys: But even if we stick with the kubeconfig approach instead of the separate flags (which personally I think is fine, and probably easier if folks already have a cert based root-kubeconfig that allows access to all shards), we could switch the fakeserver.io to the root shard URL, right?

p0lyn0mial: sure, we could do that. I think we could do that just to read the new config and then erase the host value.

pkg/proxy/options/options.go
hardys: Not blocking on it, but as previously mentioned I'd adjust the description slightly e.g:

`The path to the kubeconfig used for communication with all shards.`

p0lyn0mial: sorry, fixed :)

pkg/proxy/config.go
p0lyn0mial: I'd like to hear one more opinion on it. I'm not convinced we should change it to use the root shard URL.

p0lyn0mial: I hope that's okay with you :)

pkg/proxy/config.go
sttts: +1 for merging as is.

pkg/proxy/options/options.go
```diff
@@ -56,6 +57,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.MappingFile, "mapping-file", o.MappingFile, "Config file mapping paths to backends")
 	fs.StringVar(&o.RootDirectory, "root-directory", o.RootDirectory, "Root directory.")
 	fs.StringVar(&o.RootKubeconfig, "root-kubeconfig", o.RootKubeconfig, "The path to the kubeconfig of the root shard.")
+	fs.StringVar(&o.ShardsKubeconfig, "shards-kubeconfig", o.ShardsKubeconfig, "The path to the kubeconfig used for communication with all shards")
```
sttts, line 60: add that the server name is replaced later on with the shard hostname.

PR #2380, opened 2022-11-18
🌱 Reuse global index in placement controllers

qiujian16:
Signed-off-by: Jian Qiu <jqiu@redhat.com>

## Summary

avoid creating indexers in each placement/scheduling controllers.

## Related issue(s)

Fixes #

PR #2398, opened 2022-11-22
:seedling: Revert "Revert "Merge pull request #2380 from qiujian16/reuse-index""

stevekuznetsov:
This reverts commit c497e42e687de51fff8c5c982bc6bd68fab4903f.

Brings back #2380 
/assign @qiujian16


pkg/indexers/workload.go
```diff
@@ -21,11 +21,14 @@ import (
...
 
+	schedulingv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/scheduling/v1alpha1"
 	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
 )
 
 const (
-	SyncTargetsBySyncTargetKey = "SyncTargetsBySyncTargetKey"
+	SyncTargetsBySyncTargetKey            = "SyncTargetsBySyncTargetKey"
+	PlacementBySpecifiedLocationWorkspace = "PlacementBySpecifiedLocationWorkspace"
+	PlacementBySelectedLocationWorkspace  = "PlacementBySelectedLocationWorkspace"
```
stevekuznetsov, line 31: @qiujian16 I saw that the code had two indices for placements by location workspace - one based on spec, one on status - do we need both? When should a controller author use one or the other? We might want to add GoDoc on this.

Similarly, we have this duplicity for sync target keys as well, using spec https://github.com/kcp-dev/kcp/blob/5916c3cd79ca30b3ed6e8a8983d10034f1a926e2/pkg/indexers/workload.go#L31-L38 :

And status https://github.com/kcp-dev/kcp/blob/5916c3cd79ca30b3ed6e8a8983d10034f1a926e2/pkg/reconciler/workload/resource/resource_controller.go#L547-L558:

qiujian16: location is a bit different. The when using spec is for controller to select a location based on spec. The one for status is for controller to select a synctarget in the selected location. The latter controller can not use index based on spec, because the value in spec and status are not necessarily to be consistent.

InternalSyncTargetKeyLabel is a hash of synctarget key which is used to label the workload resources for scheduling. This index as an easy way to find the synctarget based on this hash value.

stevekuznetsov: >  The one for status is for controller to select a synctarget in the selected location.

Not sure I follow - both of them are indexing `Placement` objects. Could you go into more detail about what you mean?

> InternalSyncTargetKeyLabel is a hash of synctarget key which is used to label the workload resources for scheduling. This index as an easy way to find the synctarget based on this hash value.

OK - in what cases should the index compute this value, versus looking it up from labels?

qiujian16: > Not sure I follow - both of them are indexing Placement objects. Could you go into more detail about what you mean?

this https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/scheduling/placement/placement_controller.go uses index by spec, since this controller will finally set the locationRef (plus workspace) in status of location.

when status is set, https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/workload/placement/placement_controller.go will use the index by status to the do the scheduling to synctarget. This controller cannot use the index by spec, because the value in spec/status could be different.

davidfestal: Is is as efficient as using an index ?

stevekuznetsov: > The InternalSyncTargetKeyLabel is added on a workload resource (e.g. deployment).

As linked, this index is looking at the label on the SyncTarget itself:

https://github.com/kcp-dev/kcp/blob/5916c3cd79ca30b3ed6e8a8983d10034f1a926e2/pkg/reconciler/workload/resource/resource_controller.go#L547-L558

stevekuznetsov: @davidfestal index will make the scan O(1) at  the cost of some more memory whereas label selection will be O(object count) - both are in-memory, though, and I'm not sure it's something anyone would notice

davidfestal: Thanks for the answer @stevekuznetsov . I'm not against using the select then in order to avoid using the index here to retrieve the SyncTarget itself.

stevekuznetsov:
Good question for @qiujian16

qiujian16:
Since many code is changed after workspace refactor, how about we close this one. I will submit another PR for placement index.

Issue #2389, opened 2022-11-18
Flake: TestUpsyncerVirtualWorkspace: multiple failures


ncdc:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kcp-dev_kcp/2356/pull-ci-kcp-dev-kcp-main-e2e-sharded/1593629106879074304

Issue #2371, opened 2022-11-16
bug: flake: TestSyncerVirtualWorkspace/access_kcp_resources_through_syncer_virtual_workspace,_from_a_other_workspace_to_the_wildwest_resources_through_an_APIBinding


stevekuznetsov:
### Describe the bug

Seen this a couple times now, xref: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kcp-dev_kcp/2353/pull-ci-kcp-dev-kcp-main-e2e-sharded/1592881165528731648

### Steps To Reproduce

run tests

### Expected Behaviour

no fail

stevekuznetsov:
cc @qiujian16 @davidfestal

PR #2403, opened 2022-11-23
🐛 Fix Syncer-related e2e flake due to the use of environment variable

davidfestal:
## Summary

This fixes potential Syncer-related e2e flakes due to the use of an environment variable to pass to the Syncer the Namespace in which it should watch for DNS-related resources. 

## Related issue(s)

Possibly related flakes are https://github.com/kcp-dev/kcp/issues/2389 and https://github.com/kcp-dev/kcp/issues/2371

### DIFF
--- a/cmd/syncer/cmd/syncer.go
+++ b/cmd/syncer/cmd/syncer.go
@@ -18,6 +18,8 @@ package cmd
 
 import (
 	"context"
+	"errors"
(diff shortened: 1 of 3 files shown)


ncdc:
@davidfestal do you want to close 2389 and 2371, or ... ?

PR #2405, opened 2022-11-24
🐛 Fix parsing of requests to root path of a workload cluster

erwinvaneyk:
## Summary

kcp is rejecting (valid) calls to the root path of a cluster. With a default setup, the following error is returned:

```
# with KUBECONFIG pointing to some workload cluster within kcp
$ kubectl get --raw='/'
[{
  "metadata": {},
  "status": "Failure",
  "message": "unable to parse cluster: no `/` found in path root",
  "reason": "BadRequest",
  "code": 400
}]
```

This PR resolves this issue by fixing the parsing and removing the error. With this fix:
```
# with KUBECONFIG pointing to some workload cluster within kcp
$ kubectl get --raw='/'
{
  "paths": [
    "/api",
    "/api/v1",
    "/apis",
    "/apis/",
    "/apis/admissionregistration.k8s.io",
    "/apis/admissionregistration.k8s.io/v1",
    "/apis/apiextensions.k8s.io",
    ...
  ]
}
```

### DIFF
--- a/pkg/server/filters/filters.go
+++ b/pkg/server/filters/filters.go
@@ -89,11 +89,8 @@ func WithClusterScope(apiHandler http.Handler) http.HandlerFunc {
 
 			i := strings.Index(path, "/")
 			if i == -1 {
-				responsewriters.ErrorNegotiated(
-					apierrors.NewBadRequest(fmt.Sprintf("unable to parse cluster: no `/` found in path %s", path)),
-					errorCodecs, schema.GroupVersion{},
-					w, req)
-				return
+				path = path + "/"
+				i = len(path) - 1
 			}
 			clusterName, path = logicalcluster.New(path[:i]), path[i:]
 			req.URL.Path = path


ncdc:
I think it may make sense to support `/`, `/healthz`, and `/readyz`, but we shouldn't expose `/api` and `/apis` at the root.

erwinvaneyk:
@ncdc Maybe I could have worded the title a bit better, but this PR solely fixes requests to the root path (`/`) of a workload cluster. So an endpoint like this: `$HOST/apis/clusters/my-cluster/`—it doesn't change anything at the root path of kcp itself

ncdc:
Ahh ok, thanks for clarifying. I got confused from the description, because `kubectl get --raw='/'` ignores any path in the kubeconfig's server and literally requests `/`, which we don't allow. I thought you were changing that.

I do now see what your PR is doing, and it's 👍:

Before:
```
k --user shard-admin get --raw /clusters/root
Error from server (BadRequest): unable to parse cluster: no `/` found in path root
```

After:
```
k --user shard-admin get --raw /clusters/root
{
  "paths": [
    "/api",
    "/api/v1",
    "/apis",
    "/apis/",
    "/apis/admissionregistration.k8s.io",
...
```

ncdc:
Some additional testing:
```
# get /clusters
Error from server (NotFound): the server could not find the requested resource

# get /clusters/
Error from server (InternalError): an error on the server ("cluster name is empty in the request context - RequestInfo: &request.RequestInfo{IsResourceRequest:false, Path:\"/\", Verb:\"get\", APIPrefix:\"\", APIGroup:\"\", APIVersion:\"\", Namespace:\"\", Resource:\"\", Subresource:\"\", Name:\"\", Parts:[]string(nil)}") has prevented the request from succeeding
```

I think the error for /clusters is good. Should we fix /clusters/ to also return a NotFound?

@stevekuznetsov do you want an e2e for this/these to make sure we don't regress?

ncdc:
@erwinvaneyk would you be able to fix `/clusters/` too and also add a small e2e?

vincepri:
/lgtm

I'll follow-up with some fixes

ncdc (approved):

PR #2407, opened 2022-11-24
:bug: add support for standalone virtualworkspace server

### LINKED ISSUE #2056: Update test code to support standalone virtual workspace servers
Having a standalone virtual workspace topology requires an extremely detailed, specific configuration to ensure that requests originating from loopback clients to the virtual workspace server are proxied correctly from the shard to the virtual workspace server (a redirect won't work, because the virtual workspace server won't be able to verify the loopback client cert).

### LINKED ISSUE #2254: bug: sharded-test-server does not work with `run-virtual-workspaces=false`
### Describe the bug

If you run `./sharded-test-server --shard-run-virtual-workspaces=false` , it never starts
```
KCP-0 [-]poststarthook/kcp-start-kcp-apibinder-initializer failed: not finished
KCP-0 I1022 16:45:28.583005 3997122 httplog.go:131] "HTTP" verb="GET" URI="/readyz" latency="3.807285ms" userAgent="sharded-test-server/v0.0.0 (linux/amd64) kubernetes/$Format" audit-ID="2ca3425a-80e8-4cdf-914d-b27807feaba7" srcIP="192.168.1.138:54444" apf_pl="exempt" apf_fs="exempt" apf_execution_time="3.546496ms" resp=500
KCP-0 I1022 16:45:29.589317 3997122 shared_informer.go:332] caches populated
KCP-0 I1022 16:45:29.589576 3997122 healthz.go:257] poststarthook/kcp-start-kcp-apibinder-initializer check failed: readyz
```

### Steps To Reproduce

```
./sharded-test-server --shard-run-virtual-workspaces=false
```

### Expected Behaviour

Should work. 

### Additional Context

I observed similar when trying to deploy KCP outside single binary where KCP and VirtualWorkspaces container never becomes ready if ran separatly

hardys:
## Summary
This is another change decoupled from the [shard-standalone-vw branch](https://github.com/hardys/kcp/commit/bf7cbeec5276aec5b4e84e9d54631307cb6b0c7c) from @ncdc

## Related issue(s)
Follow up to ##2297 and #2303 which also merged changes from the WIP branch

Fixes #2254
Fixes #2056

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -316,7 +316,7 @@ test-e2e-sharded: require-kind build-all build-kind-images
 	kind get kubeconfig > "$(WORK_DIR)/.kcp/kind.kubeconfig"
 	rm -f "$(WORK_DIR)/.kcp/admin.kubeconfig"
 	UNSAFE_E2E_HACK_DISABLE_ETCD_FSYNC=true NO_GORUN=1 \
-		./bin/sharded-test-server --quiet --v=2 --log-dir-path="$(LOG_DIR)" --work-dir-path="$(WORK_DIR)" $(TEST_SERVER_ARGS) --number-of-shards=2 2>&1 & PID=$$!; echo "PID $$PID" && \
+		./bin/sharded-test-server --quiet --v=2 --log-dir-path="$(LOG_DIR)" --work-dir-path="$(WORK_DIR)" --shard-run-virtual-workspaces=false $(TEST_SERVER_ARGS) --number-of-shards=2 2>&1 & PID=$$!; echo "PID $$PID" && \
 	trap 'kill -TERM $$PID' TERM INT EXIT && \
 	while [ ! -f "$(WORK_DIR)/.kcp/admin.kubeconfig" ]; do sleep 1; done && \
 	NO_GORUN=1 GOOS=$(OS) GOARCH=$(ARCH) \
@@ -340,7 +340,7 @@ endif
 test-e2e-sharded-minimal: build-all
 	mkdir -p "$(LOG_DIR)" "$(WORK_DIR)/.kcp"
 	rm -f "$(WORK_DIR)/.kcp/admin.kubeconfig"
-	UNSAFE_E2E_HACK_DISABLE_ETCD_FSYNC=true NO_GORUN=1 ./bin/sharded-test-server --quiet --v=2 --log-dir-path="$(LOG_DIR)" --work-dir-path="$(WORK_DIR)" $(TEST_SERVER_ARGS) --number-of-shards=2 2>&1 & PID=$$!; echo "PID $$PID" && \
+	UNSAFE_E2E_HACK_DISABLE_ETCD_FSYNC=true NO_GORUN=1 ./bin/sharded-test-server --quiet --v=2 --log-dir-path="$(LOG_DIR)" --work-dir-path="$(WORK_DIR)" --shard-run-virtual-workspaces=false $(TEST_SERVER_ARGS) --number-of-shards=2 2>&1 & PID=$$!; echo "PID $$PID" && \
 	trap 'kill -TERM $$PID' TERM INT EXIT && \
 	while [ ! -f "$(WORK_DIR)/.kcp/admin.kubeconfig" ]; do sleep 1; done && \
 	NO_GORUN=1 GOOS=$(OS) GOARCH=$(ARCH) $(GO_TEST) -race $(COUNT_ARG) $(PARALLELISM_ARG) $(WHAT) $(TEST_ARGS) \
--- a/cmd/sharded-test-server/main.go
+++ b/cmd/sharded-test-server/main.go
@@ -171,9 +171,6 @@ func start(proxyFlags, shardFlags []string, logDirPath, workDirPath string, numb
 	}
 
 	standaloneVW := sets.NewString(shardFlags...).Has("--run-virtual-workspaces=false")
-	if standaloneVW {
-		shardFlags = append(shardFlags, fmt.Sprintf("--shard-virtual-workspace-url=https://%s:7444", hostIP))
-	}
 
 	cacheServerErrCh := make(chan indexErrTuple)
 	cacheServerConfigPath := ""
@@ -194,7 +191,7 @@ func start(proxyFlags, shardFlags []string, logDirPath, workDirPath string, numb
 	// start shards
 	var shards []*shard.Shard
 	for i := 0; i < numberOfShards; i++ {
-		shard, err := newShard(ctx, i, shardFlags, servingCA, hostIP.String(), logDirPath, workDirPath, cacheServerConfigPath)
+		shard, err := newShard(ctx, i, shardFlags, standaloneVW, servingCA, hostIP.String(), logDirPath, workDirPath, cacheServerConfigPath, clientCA)
 		if err != nil {
 			return err
 		}
@@ -211,7 +208,7 @@ func start(proxyFlags, shardFlags []string, logDirPath, workDirPath string, numb
 		vwPort = "7444"
 
 		for i := 0; i < numberOfShards; i++ {
-			virtualWorkspaceErrCh, err := startVirtual(ctx, i, logDirPath, workDirPath)
+			virtualWorkspaceErrCh, err := startVirtual(ctx, i, servingCA, hostIP.String(), logDirPath, workDirPath, clientCA)
 			if err != nil {
 				return fmt.Errorf("error starting virtual workspaces server %d: %w", i, err)
 			}
--- a/cmd/sharded-test-server/main.go
+++ b/cmd/sharded-test-server/main.go
@@ -201,21 +201,22 @@ func start(proxyFlags, shardFlags []string, logDirPath, workDirPath string, numb
 		shards = append(shards, shard)
 	}
 
+	// Start virtual-workspace servers
 	vwPort := "6444"
-	virtualWorkspacesErrCh := make(chan indexErrTuple)
+	var virtualWorkspaces []*VirtualWorkspace
 	if standaloneVW {
 		// TODO: support multiple virtual workspace servers (i.e. multiple ports)
 		vwPort = "7444"
 
 		for i := 0; i < numberOfShards; i++ {
-			virtualWorkspaceErrCh, err := startVirtual(ctx, i, servingCA, hostIP.String(), logDirPath, workDirPath, clientCA)
+			vw, err := newVirtualWorkspace(ctx, i, servingCA, hostIP.String(), logDirPath, workDirPath, clientCA)
 			if err != nil {
-				return fmt.Errorf("error starting virtual workspaces server %d: %w", i, err)
+				return err
 			}
-			go func(vwIndex int, vwErrCh <-chan error) {
-				err := <-virtualWorkspaceErrCh
-				virtualWorkspacesErrCh <- indexErrTuple{vwIndex, err}
-			}(i, virtualWorkspaceErrCh)
+			if err := vw.start(ctx); err != nil {
+				return err
+			}
+			virtualWorkspaces = append(virtualWorkspaces, vw)
 		}
 	}
 
@@ -230,6 +231,21 @@ func start(proxyFlags, shardFlags []string, logDirPath, workDirPath string, numb
 			err := <-terminatedCh
 			shardsErrCh <- indexErrTuple{i, err}
 		}(i, terminatedCh)
+	}
+
+	// Wait for virtual workspaces to be ready
+	virtualWorkspacesErrCh := make(chan indexErrTuple)
+	if standaloneVW {
+		for i, vw := range virtualWorkspaces {
+			terminatedCh, err := vw.waitForReady(ctx)
+			if err != nil {
+				return err
+			}
+			go func(i int, terminatedCh <-chan error) {
+				err := <-terminatedCh
+				virtualWorkspacesErrCh <- indexErrTuple{i, err}
+			}(i, terminatedCh)
+		}
 	}
 
 	// write kcp-admin kubeconfig talking to the front-proxy with a client-cert
--- a/cmd/sharded-test-server/shard.go
+++ b/cmd/sharded-test-server/shard.go
@@ -19,8 +19,10 @@ package main
 import (
 	"context"
 	"fmt"
+	"net"
 	"os"
 	"path/filepath"
+	"strconv"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apiserver/pkg/authentication/user"
@@ -105,7 +107,8 @@ func newShard(ctx context.Context, n int, args []string, standaloneVW bool, serv
 	}
 
 	if standaloneVW {
-		args = append(args, fmt.Sprintf("--shard-virtual-workspace-url=https://%s:%d", hostIP, 7444+n))
+		args = append(args, fmt.Sprintf("--shard-virtual-workspace-url=https://%s",
+			net.JoinHostPort(hostIP, virtualWorkspacePort(n))))
 	}
 
 	return shard.NewShard(
@@ -114,6 +117,10 @@ func newShard(ctx context.Context, n int, args []string, standaloneVW bool, serv
 		logFilePath,
 		args,
 	), nil
+}
+
+func virtualWorkspacePort(n int) string {
+	return strconv.Itoa(7444 + n)
 }
 
 func embeddedEtcdClientPort(n int) int {
--- a/cmd/sharded-test-server/shard.go
+++ b/cmd/sharded-test-server/shard.go
@@ -19,16 +19,18 @@ package main
 import (
 	"context"
 	"fmt"
+	"os"
 	"path/filepath"
 
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apiserver/pkg/authentication/user"
 	"k8s.io/klog/v2"
 
 	"github.com/kcp-dev/kcp/cmd/sharded-test-server/third_party/library-go/crypto"
 	shard "github.com/kcp-dev/kcp/cmd/test-server/kcp"
 )
 
-func newShard(ctx context.Context, n int, args []string, servingCA *crypto.CA, hostIP string, logDirPath, workDirPath, cacheServerConfigPath string) (*shard.Shard, error) {
+func newShard(ctx context.Context, n int, args []string, standaloneVW bool, servingCA *crypto.CA, hostIP string, logDirPath, workDirPath, cacheServerConfigPath string, clientCA *crypto.CA) (*shard.Shard, error) {
 	logger := klog.FromContext(ctx).WithValues("shard", n)
 
 	// create serving cert
@@ -49,6 +51,15 @@ func newShard(ctx context.Context, n int, args []string, servingCA *crypto.CA, h
 	}
 	if err := cert.WriteCertConfigFile(filepath.Join(workDirPath, ".kcp/apiserver.crt"), filepath.Join(workDirPath, ".kcp/apiserver.key")); err != nil {
 		return nil, fmt.Errorf("failed to write server cert: %w", err)
+	}
+
+	shardClientCert := filepath.Join(workDirPath, fmt.Sprintf(".kcp-%d/shard-client-cert.crt", n))
+	shardClientCertKey := filepath.Join(workDirPath, fmt.Sprintf(".kcp-%d/shard-client-cert.key", n))
+	shardUser := &user.DefaultInfo{Name: fmt.Sprintf("kcp-shard-%d", n), Groups: []string{"system:masters"}}
+	_, err = clientCA.MakeClientCertificate(shardClientCert, shardClientCertKey, shardUser, 365)
+	if err != nil {
+		fmt.Printf("failed to create shard client cert: %v\n", err)
+		os.Exit(1)
 	}
 
 	if err != nil {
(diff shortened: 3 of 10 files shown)


hardys:
Some of the e2e tests are failing hence marking this as WIP until resolved, any help with that or initial review feedback is welcome :)

pkg/server/virtual.go
```diff
@@ -115,6 +115,8 @@ func (s *Server) installVirtualWorkspaces(
 		return err
 	}
 
+	preparedRootAPIServer.AuditBackend = server.AuditBackend
```
ncdc, line 118: I think I added this back before we had audit working correctly, but IIRC @s-urbaniak got it wired in correctly, so we can probably remove this (?). It looks like line 92 has it covered?

test/e2e/virtual/workspaces/virtual_workspace_test.go
```diff
@@ -37,12 +40,15 @@ import (
...
 	"k8s.io/apimachinery/pkg/util/wait"
 	kuser "k8s.io/apiserver/pkg/authentication/user"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 
+	"github.com/kcp-dev/kcp/cmd/sharded-test-server/third_party/library-go/crypto"
 	virtualcommand "github.com/kcp-dev/kcp/cmd/virtual-workspaces/command"
 	virtualoptions "github.com/kcp-dev/kcp/cmd/virtual-workspaces/options"
 	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
```
ncdc, line 54: Is it now appropriate to remove the `t.Skip` on L170?

hardys: Ack thanks, I just noticed that while testing locally - will remove, are any further changes needed to close out https://github.com/kcp-dev/kcp/issues/2056 ?

hardys:
Thanks for the review feedback @ncdc - I pushed an update which fixes most (but not yet all) of the e2e failures, will leave this as WIP until all tests are green and the review feedback is addressed :+1:

cmd/sharded-test-server/virtual.go
```diff
@@ -27,30 +27,106 @@ import (
...
+	vwClientCertKeyPath, err := filepath.Abs(vwClientCertKey)
+	if err != nil {
+		fmt.Printf("error getting absolute path for %q: %v\n", vwClientCertKey, err)
+		os.Exit(1)
+	}
+
+	virtualWorkspaceKubeConfig := clientcmdapi.Config{
+		Clusters: map[string]*clientcmdapi.Cluster{
+			"shard": {
+				Server:               fmt.Sprintf("https://localhost:%d/clusters/system:admin", 6444+index),
```
hardys, line 93: @ncdc should this be `7444+index` so the `virtualworkspace.kubeconfig` points directly at the VW endpoint?

hardys: context - I realize this will redirect to the standalone VW server, but I was planning to add a readiness check for the VW readyz, and I realized I can't use this kubeconfig like we do e.g for [the cache server](https://github.com/kcp-dev/kcp/blob/main/cmd/sharded-test-server/cache.go#L92)

test/e2e/virtual/workspaces/virtual_workspace_test.go
ncdc: Nothing further needed. Please mark this PR as also fixing 2056. Thanks

cmd/sharded-test-server/virtual.go
```diff
@@ -27,30 +27,106 @@ import (
...
+	}
+
 	authenticationKubeconfigPath := filepath.Join(workDirPath, fmt.Sprintf(".kcp-%d", index), "admin.kubeconfig")
 	clientCAFilePath := filepath.Join(workDirPath, ".kcp", "client-ca.crt")
 
 	commandLine := framework.DirectOrGoRunCommand("virtual-workspaces")
 	commandLine = append(
 		commandLine,
-		fmt.Sprintf("--kubeconfig=%s", authenticationKubeconfigPath),
+		fmt.Sprintf("--kubeconfig=%s", filepath.Join(workDirPath, fmt.Sprintf(".kcp-virtual-workspaces-%d/virtualworkspace.kubeconfig", index))),
```
ncdc, line 124: I think we can reuse `kubeconfigPath` here?

pkg/server/handler.go
```diff
@@ -46,6 +47,7 @@ import (
 	apiserverdiscovery "k8s.io/apiserver/pkg/endpoints/discovery"
 	"k8s.io/apiserver/pkg/endpoints/handlers/responsewriters"
 	"k8s.io/apiserver/pkg/endpoints/request"
+	transport2 "k8s.io/client-go/transport"
```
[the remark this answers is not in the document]
hardys: done

pkg/server/virtual.go
hardys: Ah yes thanks, removed

test/e2e/virtual/workspaces/virtual_workspace_test.go
hardys: Ok removed the skip but one of the test cases `a user sees his own workspaces at the root, but no other workspaces` is failing, looking into why that is

cmd/sharded-test-server/virtual.go
hardys: done

test/e2e/virtual/workspaces/virtual_workspace_test.go
```diff
@@ -210,15 +214,15 @@ var testCases = []struct {
 
 			t.Logf("Create Workspace workspace1 in the virtual workspace")
 			var workspace1 *tenancyv1beta1.Workspace
-			require.Eventually(t, func() bool {
+			framework.Eventually(t, func() (bool, string) {
 				// RBAC authz uses informers and needs a moment to understand the new roles. Hence, try until successful.
 				var err error
 				workspace1, err = vwUser1Client.Cluster(server.orgClusterName).TenancyV1beta1().Workspaces().Create(ctx, testData.workspace1.DeepCopy(), metav1.CreateOptions{})
 				if err != nil {
 					t.Logf("Failed to create workspace1: %v", err)
```
stevekuznetsov, line 222: Remove the log, `framework.Eventually` will log your returned string

test/e2e/virtual/workspaces/virtual_workspace_test.go
```diff
@@ -210,15 +214,15 @@ var testCases = []struct {
...
-					return false
+					return false, fmt.Sprintf("%v", err)
```
stevekuznetsov, line 223: nit: err.Error()

test/e2e/virtual/workspaces/virtual_workspace_test.go
```diff
@@ -599,27 +584,94 @@ func testWorkspacesVirtualWorkspaces(t *testing.T, standalone bool) {
...
+			fmt.Printf("failed to create shard client cert: %v\n", err)
+			os.Exit(1)
+		}
+
 		portStr, err := framework.GetFreePort(t)
 		require.NoError(t, err)
 
-		tokenAuthFile := framework.WriteTokenAuthFile(t)
 		server = framework.PrivateKcpServer(t,
+			framework.PrivateKcpServerSkipReadyCheck(true),
```
stevekuznetsov, line 644: Why?

stevekuznetsov (commented):
Seems mostly reasonable, but boy is that an enormous test setup

hardys:
@ncdc quick update on this as I'm still struggling somewhat to get the e2e tests all working in standalone-vw mode

AFAICS some tests (for example `TestRootWorkspaces/a_user_sees_his_own_workspaces_at_the_root,_but_no_other_workspaces` are failing because they hit the `ForeverTestTimeout` of 30s.

Locally testing (with sharded-test-server configured with `--shard-run-virtual-workspaces=false`) e.g

```
GOOS=linux GOARCH=amd64 go test -v -failfast -parallel=1 -count=1 ./test/e2e/virtual/workspaces/... -run TestRootWorkspaces/a_user_sees_his_own_workspaces_at_the_root,_but_no_other_workspaces -args --use-default-kcp-server --root-shard-kubeconfig=$PWD/.kcp-0/admin.kubeconfig
```

We see that the test sometimes passes, and if I increase the test timeout it can pass reliably - in some test runs locally I'm seeing a delay of around 50 seconds waiting for the `user1-workspace` to appear (this is on baremetal so not resource constrained, although I do see high CPU usage).

In the shard logs we can see the proxied requests, and there are some bootstrapping authz errors which I'm thinking may be related?

```
I1213 17:56:41.503077 2707520 handler.go:138] "proxying virtual workspace request" inURL="/services/workspaces/root/apis/tenancy.kcp.dev/v1beta1/workspaces" outURL="https://10.19.133.66:7444/services/workspaces/root/apis/tenancy.kcp.dev/v1beta1/workspaces"
I1213 17:56:41.506245 2707520 httplog.go:131] "HTTP" verb="POST" URI="/clusters/root/apis/authorization.k8s.io/v1/subjectaccessreviews" latency="1.454168ms" userAgent="virtual-workspaces/v0.0.0 (linux/amd64) kubernetes/$Format/workspaces-virtual-workspace" audit-ID="5cd9b73b-6526-4798-ae35-3d85c4231534" srcIP="[::1]:35636" resp=201
I1213 17:56:41.507868 2707520 decorator.go:75] "bootstrap policy reason: "
I1213 17:56:41.507930 2707520 decorator.go:75] "local policy for cluster \"root\" reason: "
I1213 17:56:41.507940 2707520 decorator.go:75] "delegated bootstrap.authorization.kcp.dev authorizer reason: access denied\ndelegated local.authorization.kcp.dev authorizer reason: access denied"
I1213 17:56:41.507947 2707520 decorator.go:75] "delegated maxpermissionpolicy.authorization.kcp.dev authorizer reason: access denied"
I1213 17:56:41.507952 2707520 decorator.go:75] "delegated systemcrd.authorization.kcp.dev authorizer reason: access denied"
I1213 17:56:41.507958 2707520 decorator.go:75] "delegated content.authorization.kcp.dev authorizer reason: access denied"
```

If you have any suggestions where to look next that would be welcome, currently I'm not very sure why there is such a long delay in the workspace creation.

ncdc:
@hardys this particular test will go away as part of the workspace refactoring. Do you have a list of all the tests you're having problems with, and we can see if any of them still need addressing?

hardys:
> @hardys this particular test will go away as part of the workspace refactoring. Do you have a list of all the tests you're having problems with, and we can see if any of them still need addressing?

@ncdc I've mostly been testing e2e-sharded-minimal locally so I don't yet have an exhaustive list, but I'm seeing `TestInProcessWorkspacesVirtualWorkspaces/In-process_virtual_workspace_apiserver/create_a_sub-workspace_and_verify_that_only_users_with_right_permissions_can_access_workspaces_inside_it` also fail (similar to above, sometimes it passes)

I'm interested to understand why the long delay though, with the in-process VW the workspace list succeeds pretty much immediately, any idea why is it taking nearly a minute in standalone VW mode?

ncdc:
Not offhand. Maybe something to do with the workspace authz cache (which will be going away as part of the workspace refactor)

hardys:
> Not offhand. Maybe something to do with the workspace authz cache (which will be going away as part of the workspace refactor)

Ack thanks, sounds like I should probably pause work on this and rebase after the workspace refactor lands :)

s-urbaniak:
@hardys let me know if you need help after the rebase, from the logs it looks like the requesting user for the SAR passes all authorizers but the local and bootstrap policy one. Unfortanly kubernetes RBAC does not give any reason in case of denials, so the local and bootstrap policy authorizer return empty string.

hardys:
/test e2e-sharded

Looks like #2501 caused the failure on the last run so retrying

hardys:
@ncdc this is rebased, passing CI (finally! \o/) and ready for another review pass when you have time, thanks!

ncdc (approved):

hardys:
/hold 

The test-e2e failure looks like a real problem with `TestStandaloneWorkspacesVirtualWorkspaces/Standalone_virtual_workspace_apiserver` - investigating

hardys:
Turns out I lost a bunch of the new changes in `test/e2e/virtual/clusterworkspaces/virtual_workspace_test.go` due to a rebase mistake, now re-added those, but `TestStandaloneWorkspacesVirtualWorkspaces/Standalone_virtual_workspace_apiserver` is not passing for me locally so leaving the hold until I figure that out

hardys:
> @hardys let me know if you need help after the rebase, from the logs it looks like the requesting user for the SAR passes all authorizers but the local and bootstrap policy one for the `root` workspace. Unfortunately kubernetes RBAC does not give any reason in case of denials, so the local and bootstrap policy authorizer return empty string.

@s-urbaniak now rebased, and if you do have time to take another look, help would be welcome :)

AFAICS the standalone VW server is getting denied access to KCP e.g:

```
I0109 12:03:25.762947 3555316 httplog.go:131] "HTTP" verb="LIST" URI="/services/initializingworkspaces/system:apibindings/clusters/%2A/apis/core.kcp.io/v1alpha1/logicalclusters?limit=500&resourceVersion=0" latency="2.906727ms" userAgent="kcp/v0.0.0 (linux/amd64) kubernetes/$Format/kcp-apibinder-initializer" audit-ID="bde34168-5d78-4c6c-b82d-636834d6bb04" srcIP="[::1]:36338" resp=403
I0109 12:03:33.696294 3555316 impersonation.go:119] "Forbidden" URI="/services/initializingworkspaces/system:apibindings/clusters/%2A/apis/core.kcp.io/v1alpha1/logicalclusters?limit=500&resourceVersion=0" Reason="access denied" Error=<nil>
```

Seems authz related, but I've not yet figured out exactly what is causing the requests to be denied, as you say `<nil>` isn't the most informative, and I don't see many clues in the logs atm...

Makefile
```diff
@@ -310,7 +310,7 @@ test-e2e-sharded: require-kind build-all build-kind-images
 	mkdir -p "$(LOG_DIR)" "$(WORK_DIR)/.kcp"
 	kind get kubeconfig > "$(WORK_DIR)/.kcp/kind.kubeconfig"
 	rm -f "$(WORK_DIR)/.kcp/admin.kubeconfig"
-	UNSAFE_E2E_HACK_DISABLE_ETCD_FSYNC=true NO_GORUN=1 ./bin/sharded-test-server --quiet --v=2 --log-dir-path="$(LOG_DIR)" --work-dir-path="$(WORK_DIR)" $(TEST_SERVER_ARGS) --number-of-shards=2 2>&1 & PID=$$!; echo "PID $$PID" && \
+	UNSAFE_E2E_HACK_DISABLE_ETCD_FSYNC=true NO_GORUN=1 ./bin/sharded-test-server --quiet --v=2 --log-dir-path="$(LOG_DIR)" --work-dir-path="$(WORK_DIR)" --shard-run-virtual-workspaces=false $(TEST_SERVER_ARGS) --number-of-shards=2 2>&1 & PID=$$!; echo "PID $$PID" && \
```
stevekuznetsov, line 313: Why do we not run with a standalone VW here?

cmd/sharded-test-server/main.go
```diff
@@ -236,6 +234,21 @@ func start(proxyFlags, shardFlags []string, logDirPath, workDirPath string, numb
 		return err
 	}
 
+	// Wait for virtual workspaces to be ready
```
stevekuznetsov, line 237: Nit: neither this block nor the one below it is waiting for anything, got a bit misled by the comment

cmd/sharded-test-server/shard.go
```diff
@@ -83,11 +94,18 @@ func newShard(ctx context.Context, n int, args []string, servingCA *crypto.CA, h
...
+		fmt.Sprintf("--shard-client-cert-file=%s", shardClientCert),
+		fmt.Sprintf("--shard-client-key-file=%s", shardClientCertKey),
+		fmt.Sprintf("--shard-virtual-workspace-ca-file=%s", filepath.Join(workDirPath, ".kcp", "serving-ca.crt")),
 	)
 	if len(cacheServerConfigPath) > 0 {
 		args = append(args, fmt.Sprintf("--cache-server-kubeconfig-file=%s", cacheServerConfigPath))
 	}
 
+	if standaloneVW {
+		args = append(args, fmt.Sprintf("--shard-virtual-workspace-url=https://%s:%d", hostIP, 7444+n))
```
stevekuznetsov, line 106: nit: might be useful to close over these port additions to label them - `func virtualWorkspacePort(n int)`

cmd/sharded-test-server/virtual.go
```diff
@@ -19,70 +19,239 @@ package main
...
-	return terminatedCh, nil
+	return nil
+}
+
+func (v *VirtualWorkspace) waitForReady(ctx context.Context) (<-chan error, error) {
+	// wait for readiness
+	logger := klog.FromContext(ctx)
+	logger.WithValues("virtual-workspaces", v.index).Info("Waiting for virtual-workspaces /readyz to succeed")
+	for {
+		time.Sleep(100 * time.Millisecond)
```
stevekuznetsov, line 205: nit: prefer a `time.Ticker()` or any of the `wait.*` library functions for convenience.

stevekuznetsov: They do things like handle cancellation contexts for you etc

cmd/sharded-test-server/virtual.go
```diff
@@ -19,70 +19,239 @@ package main
...
+			continue
+		}
+
+		vwClient, err := kcpclientset.NewForConfig(vwConfig)
+		if err != nil {
+			logger.Error(err, "failed to create vw client")
+			continue
+		}
+
+		res := vwClient.RESTClient().Get().AbsPath("/readyz").Do(ctx)
```
stevekuznetsov, line 233: What does the output of this call look like? We have readiness checks for API-server-like binaries already, would be good to  reuse if this has e.g. post-start-hooks

pkg/server/handler.go
```diff
@@ -102,6 +104,54 @@ func WithAuditAnnotation(handler http.Handler) http.HandlerFunc {
...
+
+		if shardVirtualWorkspaceURL == nil || transport == nil {
+			// This handler func is only installed when these are both set. If this happens, it means we've regressed
+			// in the installation of this handler func, and a panic is appropriate.
+			panic("both shardVirtualWorkspaceURL and transport are required")
+		}
+
+		user, ok := request.UserFrom(req.Context())
+		if !ok {
+			http.Error(w, "no user", http.StatusBadRequest)
```
stevekuznetsov, line 128: Let's use the k8s error writers so clients get a `*metav1.Status`

pkg/server/handler.go
```diff
@@ -102,6 +104,54 @@ func WithAuditAnnotation(handler http.Handler) http.HandlerFunc {
...
+			return
+		}
+
+		proxy := &httputil.ReverseProxy{
```
stevekuznetsov, line 132: Only the transport here is request-specific - is it expensive to create the rest of this per-request instead of having some shared proxy?

test/e2e/framework/kcp.go
```diff
@@ -94,27 +94,42 @@ type kcpFixture struct {
...
+	for _, opt := range opts {
+		opt(cfg)
 	}
+	cfg.Args = append(cfg.Args, TestServerWithAuditPolicyFile(WriteEmbedFile(t, "audit-policy.yaml"))...,)
 
 	f := newKcpFixture(t, *cfg)
 	return f.Servers[serverName]
 }
 
+type PrivateKcpServerOption func(c *kcpConfig)
```
stevekuznetsov, line 111: nit: maybe look at the type-union approach I used in `test/e2e/framework/workspaces.go` to discriminate between option types a little harder?

Makefile
hardys: See the `sharded-test-server: support standalone vw ` commit - this means that `--shard-run-virtual-workspaces=false` now works and does run a standalone VW server, so the sharded jobs get switched over to use that mode

test/e2e/framework/kcp.go
hardys: To be honest in this case I'd rather go the opposite direction - unlike in your workspaces code (where there is a logical set of Privileged and Unprivileged options) here we have `kcpConfig` options which can potentially work regardless of whether the server instance is Private or Shared (the options are the same, it's just the usage of the resulting server that is different).

As a specific example, look at `PrivateWithScratchDirectories` vs the existing `WithScratchDirectories` - we don't really need both IMO, thus further discriminating between Private vs Shared probably doesn't make much sense?

This is carried over from work @ncdc started so I left it as-is for now, but I can either look at reworking this, or take an action to do it as a follow-up.

cmd/sharded-test-server/main.go
hardys: I'm confused - this block is doing `vw.waitForReady` and the one below is doing `shard.WaitForReady` ?

test/e2e/framework/kcp.go
ncdc: I'm fine with either fixing it now or in a follow up

hardys: Since it's not strictly related to the rest of this PR (and provides a way to clean up `TestAuditLogs`) I broke this part into https://github.com/kcp-dev/kcp/pull/2610

That removes `PrivateKcpServerOption` and just makes the existing `WithScratchDirectories` work with the PrivateKcpServer fixture.

cmd/sharded-test-server/main.go
hardys: I don't think that's correct - the `waitForReady` does block running in the main goroutine, but then a goroutine is spawned to collect any output from `terminatedCh` and push it to `virtualWorkspacesErrCh` (for later consumption in the `select` at the end) - this same pattern used in #2303

cmd/sharded-test-server/virtual.go
hardys: AFAICS this is consistent with the readyz checks elsewhere in the *test-server code, so perhaps we could consider going with this for now, then reworking all later so we don't end up with an inconsistent mix of approaches?

```
$ grep -R "RESTClient" ./*
./sharded-test-server/virtual.go:		res := vwClient.RESTClient().Get().AbsPath("/readyz").Do(ctx)
./sharded-test-server/frontproxy.go:		res := kcpClient.RESTClient().Get().AbsPath("/readyz").Do(ctx)
./sharded-test-server/cache.go:		res := cacheClient.RESTClient().Get().AbsPath("/readyz").Do(ctx)
./test-server/kcp/shard.go:		res := kcpClient.RESTClient().Get().AbsPath("/readyz").Do(ctx)
```

cmd/sharded-test-server/virtual.go
hardys: Done

pkg/server/handler.go
hardys: Done

test/e2e/framework/kcp.go
hardys: `PrivateKcpServerOption` now removed

cmd/sharded-test-server/shard.go
hardys: Done

pkg/server/handler.go
hardys: Good point, I moved the proxy creation up into `config.go`

hardys:
/hold cancel

@ncdc @stevekuznetsov this is ready for another review when you get some time, thanks!

ncdc (approved):

PR #2429, opened 2022-11-29
test/e2e/framework/users: fix AdmitWorkspaceAccess

p0lyn0mial:
## Related issue(s)

Fixes #


p0lyn0mial:
maybe we actually don't need it
/hold

PR #2336, opened 2022-11-09
✨ New Deployment Splitter as a coordination controller

davidfestal:
## Summary

This PR implements the DeploymentSplitter as a coordination controller.
 
## How to test:

```bash

# Start KCP:

./bin/kcp start

# Create a `mylocations` workspace in a dedicated org:

kubectl ws create org --enter
kubectl ws create synctargets --enter


# Prepare for the Syncer installation:

kubectl kcp workload sync west --replicas=0 --syncer-image=syncer --output-file=west-syncer.yaml
kubectl kcp workload sync east --replicas=0 --syncer-image=syncer --output-file=east-syncer.yaml


# k get Apply the syncer manifest to the west and east physical clusters:

kubectl --kubeconfig contrib/demos-unmaintained/demo/clusters/kind/us-west1.kubeconfig apply -f west-syncer.yaml
kubectl --kubeconfig contrib/demos-unmaintained/demo/clusters/kind/us-east1.kubeconfig apply -f east-syncer.yaml

# Start the syncers locally:

WEST_NAMESPACE=$(grep -e 'namespace: .*$' west-syncer.yaml | head -1 | sed -e 's/^  namespace: \(.*\)$/\1/')
WEST_UID=$(grep -e '- --sync-target-uid=.*$' west-syncer.yaml | head -1 | sed -e 's/^ *- --sync-target-uid=\(.*\)$/\1/')
NAMESPACE=$WEST_NAMESPACE ./bin/syncer --dns-image=ghcr.io/kcp-dev/kcp/syncer:main  --sync-target-name west --sync-target-uid "${WEST_UID}" --from-kubeconfig ${KUBECONFIG} --from-context system:admin --from-cluster root:org:synctargets --to-kubeconfig contrib/demos-unmaintained/demo/clusters/kind/us-west1.kubeconfig

EAST_NAMESPACE=$(grep -e 'namespace: .*$' east-syncer.yaml | head -1 | sed -e 's/^  namespace: \(.*\)$/\1/')
EAST_UID=$(grep -e '- --sync-target-uid=.*$' east-syncer.yaml | head -1 | sed -e 's/^ *- --sync-target-uid=\(.*\)$/\1/')
[...]

PR #2431, opened 2022-11-30
:sparkles:  Install Calico in kind for CI testing

lionelvillard:
## Summary

Default kind configuration does not support network policies. This PR modifies the kind configuration disabling kindnet and install Calico.

## Related issue(s)

/cc @davidfestal

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -87,7 +87,7 @@ jobs:
     - uses: container-tools/kind-action@v2
       name: Kubernetes KinD Cluster w/local registry
       with:
-        version: v0.14.0
+        version: v0.17.0
         config: test/e2e/kind/config-calico.yaml
 
     # Install Calico
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -47,9 +47,20 @@ jobs:
         name: Kubernetes KinD Cluster w/local registry
         with:
           version: v0.17.0
+          config: test/e2e/kind/config-calico.yaml
 
+      # Install Calico
+      - run: |-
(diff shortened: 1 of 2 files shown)


davidfestal:
Do you think you could upgrade the `kind` version to `0.17.0` at the same time ?
It is done in [another PR](https://github.com/kcp-dev/kcp/pull/2336/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR45), but it seems to me that this kind version upgrade fits better in this PR.

davidfestal (commented):
Commented about adding an upgrade of the kind version in this PR.

lionelvillard:
TestDeploymentCoordinator "flake"

/test-required

davidfestal:
> TestDeploymentCoordinator "flake"
> 
> /test-required

TBF it seems more than a flake here, since it never succeeds. Wondering if something in the change breaks them.

davidfestal:
@stevekuznetsov So do you confirm it is OK for you if we still merge this PR and @lionelvillard would create a follow-up PR for the changes you requested in this slack thread: https://kubernetes.slack.com/archives/C021U8WSAFK/p1672913451512589 ?

davidfestal:
> @stevekuznetsov So do you confirm it is OK for you if we still merge this PR and @lionelvillard would create a follow-up PR for the changes you requested in this slack thread: https://kubernetes.slack.com/archives/C021U8WSAFK/p1672913451512589 ?

I assume yes, as you mentioned here https://kubernetes.slack.com/archives/C021U8WSAFK/p1672931932256269?thread_ts=1672913451.512589&cid=C021U8WSAFK

Let me approve then.

PR #2432, opened 2022-11-30
:sparkles: APIExportEndpointSlice reconciliation

### LINKED ISSUE #2332: feature: APIExportEndpointSlice reconciliation

fgiloux:
## Summary

This  PR introduces some reconciliation logic for the population of an endpoint per shard in the status of the apiexportendpointslices. This will replace what was directly populated in the status of apiexports.

[Background](https://hackmd.io/EuPmah62QHSEs-po04Kn1Q?view)

Conditions are also added in this PR to the APIExportEndpointSlice status.

Filtering will be introduced in a subsequent PR.

## Related issue(s)

~~Fixes~~ Contributes to #2332


pkg/apis/apis/v1alpha1/types_apiexport.go
```diff
@@ -281,8 +281,9 @@ type APIExportStatus struct {
 	// +optional
 	Conditions conditionsv1alpha1.Conditions `json:"conditions,omitempty"`
 
-	// virtualWorkspaces contains all APIExport virtual workspace URLs.
-	// This field will get deprecated in favor of APIExportEndpointSlice.
+	// virtualWorkspaces (deprecated) contains all APIExport virtual workspace URLs.
+	// This field is deprecated in favor of endpoints listed in the status of
+	// APIExportEndpointSlice.
```
ncdc, line 286:
```suggestion
	// virtualWorkspaces contains all APIExport virtual workspace URLs.
	//
	// Deprecated: use APIExportEndpointSlice.status.endpoints instead.
```

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,247 @@
...
+type Resource = committer.Resource[*APIExportEndpointSliceSpec, *APIExportEndpointSliceStatus]
+type CommitFunc = func(context.Context, *Resource, *Resource) error
+
+// controller reconciles APIExportEndpointSlices. It ensures that the shard endpoints are populated
+//  in the status of every APIExportEndpointSlices.
+type controller struct {
+	queue workqueue.RateLimitingInterface
+
+	apiExportEndpointSliceLister  apislisters.APIExportEndpointSliceClusterLister
+	apiExportEndpointSliceIndexer cache.Indexer
```
ncdc, line 119: Please store function pointers instead of listers/indexers, e.g. `listAPIExportEndpointSlices func(...)`

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// kcpClusterClient is cluster aware and used to communicate with kcp API servers
+	kcpClusterClient kcpclientset.ClusterInterface
```
ncdc, line 122: Please store function pointers as needed

pkg/server/server.go
```diff
@@ -456,6 +456,12 @@ func (s *Server) Run(ctx context.Context) error {
 		}
 	}
 
+	if s.Options.Cache.Enabled && (s.Options.Controllers.EnableAll || enabled.Has("apiexportendpointslice")) {
```
ncdc, line 459: So this doesn't work if the cache server is disabled? What is the fallback option for a service provider wanting to know which URLs to use?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Don't keep the endpoints if the APIExport has been deleted
+			apiExportEndpointSlice.Status.APIExportEndpoints = nil
+			conditions.MarkFalse(
+				apiExportEndpointSlice,
+				apisv1alpha1.APIExportValid,
+				apisv1alpha1.APIExportNotFoundReason,
+				conditionsv1alpha1.ConditionSeverityError,
+				"APIExport %s|%s not found",
```
stevekuznetsov, line 82: Does this leak information? By creating APIExportSlices, I get to figure out which exports exist and where?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
```diff
@@ -0,0 +1,174 @@
...
+				workspaceRef.ExportName,
+			)
+			return nil
+		} else {
+			conditions.MarkFalse(
+				apiExportEndpointSlice,
+				apisv1alpha1.APIExportValid,
+				apisv1alpha1.InternalErrorReason,
+				conditionsv1alpha1.ConditionSeverityError,
+				"Error getting APIExport %s|%s: %v",
```
stevekuznetsov, line 93: This error is something _other_ than `NotFound` from a lister - unclear what it might even be, but does it make sense to update status here? Can we retry the reconciliation? What might a user do with the information here?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	conditions.MarkTrue(apiExportEndpointSlice, apisv1alpha1.APIExportValid)
+
+	if err = c.updateEndpoints(ctx, apiExportEndpointSlice, apiExport); err != nil {
+		conditions.MarkFalse(
+			apiExportEndpointSlice,
+			apisv1alpha1.APIExportEndpointSliceURLsReady,
+			apisv1alpha1.ErrorGeneratingURLsReason,
+			conditionsv1alpha1.ConditionSeverityError,
+			err.Error(),
```
stevekuznetsov, line 109: Similarly here, what is a user to do with an error to List?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
```diff
@@ -0,0 +1,174 @@
...
+		)
+		return err
+	}
+	conditions.MarkTrue(apiExportEndpointSlice, apisv1alpha1.APIExportEndpointSliceURLsReady)
```
stevekuznetsov, line 113: Don't we do this in `updateEndpoints()`?

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -202,7 +202,7 @@ type controller struct {
 	commit                     CommitFunc
 }
 
-// enqueueAPIBinding enqueues an APIExport .
+// enqueueAPIExport enqueues an APIExport.
```
stevekuznetsov, line 205: nit: not including unrelated stuff helps to review :)

pkg/reconciler/cache/replication/replication_reconcile.go
```diff
@@ -82,22 +82,22 @@ func (c *controller) reconcile(ctx context.Context, gvrKey string) error {
 
 // reconcileObject makes sure that the object under the given key from the local shard is replicated to the cache server.
 // the replication function handles the following cases:
-//  1. creation of the object in the cache server when the cached object is not found by retriveLocalObject
```
stevekuznetsov, line 85: nit: not including unrelated stuff helps to review :)

pkg/server/controllers.go
```diff
@@ -849,6 +850,39 @@ func (s *Server) installAPIExportController(ctx context.Context, config *rest.Co
...
+	config = rest.AddUserAgent(config, apiexportendpointslice.ControllerName)
+
+	kcpClusterClient, err := kcpclientset.NewForConfig(config)
+	if err != nil {
+		return err
+	}
+
+	c, err := apiexportendpointslice.NewController(
+		s.KcpSharedInformerFactory.Apis().V1alpha1().APIExportEndpointSlices(),
+		// ClusterWorkspaceShards and APIExports get retrieved from cache server
```
stevekuznetsov, line 864: Does a shard have some local storage for its' own definition? Surely also there will be APIExports on the shard, that do not need to be requested from the cache - we can progress in processing that data even when the cache is down, right?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
ncdc: I would think an APIExportEndpointSlice may only point at APIExports in the same Workspace. Is that accurate? Maybe we need to adjust APIExportEndpointSliceSpec slightly so it can't point at an APIExport in a different Workspace?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
ncdc: It's either impossible or essentially impossible to get other kinds of errors, given the current implementation of a lister. But in any case, I do think updating a condition is appropriate - it provides feedback to the user that something went wrong. They may not have any recourse, but it's better than giving them no information and having the key stuck in an endless exponential backoff loop.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
ncdc: Same as above, even if there is no user recourse, they need to be aware of any issues. Some information is better than no information.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
fgiloux: amended. Mind however that listers and Indexers are stored in both apiexport and apibinding controllers.

pkg/server/server.go
fgiloux: Yes. I should have raised that. Here are my views:
- sharded environment means cache server
- the construct (APIExportEndpointSlice) is of little value for non-sharded environment
- it is not a good UX for service providers to have a different API for sharded and non-sharded environments. Most probably they want to develop their service once and have it working for both

I would propose that in a non-sharded / no cache server environment informers for the local instance are passed to the controller instead of the ones for the cache server, e.g.:
~~~
c, err := apiexportendpointslice.NewController(
		s.KcpSharedInformerFactory.Apis().V1alpha1().APIExportEndpointSlices(),
		// ClusterWorkspaceShards and APIExports get retrieved from cache server
		s.CacheKcpSharedInformerFactory.Tenancy().V1alpha1().ClusterWorkspaceShards(),
		s.CacheKcpSharedInformerFactory.Apis().V1alpha1().APIExports(),
		kcpClusterClient,
	)
~~~
vs
~~~
c, err := apiexportendpointslice.NewController(
		s.KcpSharedInformerFactory.Apis().V1alpha1().APIExportEndpointSlices(),
		// ClusterWorkspaceShards and APIExports get retrieved from cache server
		s.KcpSharedInformerFactory.Tenancy().V1alpha1().ClusterWorkspaceShards(),
		s.KcpSharedInformerFactory.Apis().V1alpha1().APIExports(),
		kcpClusterClient,
	)
~~~
 As the cache server is K8 API compliant the same code should work with both. As there is a single shard all the relevant information should be provided by it.
 
 What do you think?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,247 @@
...
+
+	listClusterWorkspaceShards func() ([]*tenancyv1alpha1.ClusterWorkspaceShard, error)
```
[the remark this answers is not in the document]
fgiloux: Today all the ClusterWorkspaceShards are in the root workspace of the root shard. I think that the API or its use could be modified so that some information can be surfaced by the shard itself, e.g.:
- resource availability for scheduling
- bound apiexports for filtering out endpoints in APIExportEndpointSlice for shards where no APIBinding has been created for the APIExport
This would  require that ClusterWorkspaceShard or an alternative resource gets created in its own shard, not the root one.

Even if we allow the above I don't think we will filter by workspace in the near future. We would first implement the filtering by ClusterWorkspaceShard labels offered by Partition and PartitionSet.
 
Long story short I think that for now listing across workspaces should be ok. Do you see it differently?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
fgiloux: > I would think an APIExportEndpointSlice may only point at APIExports in the same Workspace. Is that accurate? 

APIExportEndpointSlice is similar to an APIBinding and can point to an APIExport in another workspace. It is desired so. Service provider wanting to cover shards in multiple regions, will want more often than not an APIExportEndpointSlice in each of them with the relevant subset of endpoints. Partitions will allow that.

> By creating APIExportSlices, I get to figure out which exports exist and where?

We would probably need some mechanism similar to what is available for APIBinding where you need privileges to bind an APIExport. I don't know the implementation details of it. It could be something we create an issue for and follow up with a subsequent PR. What do you think?

pkg/reconciler/apis/apiexport/apiexport_controller.go
fgiloux: it was in a separate commit this time :) , not good enough?

pkg/reconciler/cache/replication/replication_reconcile.go
fgiloux: same as above

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
fgiloux: yes one too many. Keeping it here and removing the one from updateEndpoints() to have them all in one place.

pkg/server/controllers.go
fgiloux: > Does a shard have some local storage for its' own definition? 

Currently not, it is all in the root workspace of the root shard. As per the above I think we would need it.

> here will be APIExports on the shard, that do not need to be requested from the cache - we can progress in processing that data even when the cache is down, right?

That's true but I am not sure it is worth it. We need the shard information for getting more than a single endpoint. This can only be retrieved from the cache server. I would rather keep things "frozen" when the cache server cannot be reached: existing services are not impacted by the unavailability of the cache server as long as the shards don't change. Looking only at the information on the local shard, which is currently only available for the root shard if the APIExport has been created there would only give a partial picture that may be worse than the stale snapshot captured in the APIExportEndpointSlices before the unavailability of the cache server.

stevekuznetsov: I think that is against our principle of single-shard functionality without external data. We need to support "working as much as possible" when the larger system has network partitions.

pkg/server/server.go
ncdc: We usually pass in both and have the controller try the local ones first, then fall back to the cache server ones. Right @stevekuznetsov @p0lyn0mial @sttts ?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
stevekuznetsov: Hm, what do we expose on APIBinding failing to find an export? I would be surprised if it's the raw error. Let's not add information leaks :)

fgiloux: Guess where I copied the code from :)
https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/apis/apibinding/apibinding_reconcile.go#L105-L131

Seriously, I see your point and will remove the raw error. I could do the same for APIBinding in a subsequent PR.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
ncdc: Thanks. Those controllers were written before we started using the function pointer approach. We'll come back and update them.

pkg/server/server.go
sttts: Let's not diverge between local and sharded mode. Let's run the cache server even in single-shard mode.

fgiloux: As discussed I will modify the flags so that the cache server is started in embedded mode by default and as a separate process if additional information is provided.

fgiloux:
@ncdc @sttts I have completed the changes, the tests pass except for ppc64le but this is an infrastructure issue.
Note, I have silenced the lint staticcheck warnings due to the deprecation of the VirtualWorkspaces field in the status of APIExport. It still needs to be supported and removed at a later stage. 
Let me know if there are concerns left with this PR.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller_test.go
```diff
@@ -0,0 +1,209 @@
...
+					conditions.TrueCondition(apisv1alpha1.APIExportValid),
+				)
+			}
+		})
+	}
+}
+
+// requireConditionMatches looks for a condition matching c in g. Only fields that are set in c are compared (Type is
+// required, though). If c.Message is set, the test performed is contains rather than an exact match.
+func requireConditionMatches(t *testing.T, g conditions.Getter, c *conditionsv1alpha1.Condition) {
```
stevekuznetsov, line 189: You can implement a type-specific comparator for `cmp.Diff()` to be able to use that library for the entire check over conditions, which gets you very nice output, etc.

pkg/cache/server/config.go
```diff
@@ -112,9 +112,11 @@ func NewConfig(opts *cacheserveroptions.CompletedOptions, optionalLocalShardRest
 
 	// disable SSA since the cache server should not change objects in any way
 	// note, that this will break when (if) the feature is locked, until then we should be fine
-	if err := utilfeature.DefaultMutableFeatureGate.Set(fmt.Sprintf("%s=false", features.ServerSideApply)); err != nil {
+	// TODO (fgiloux) best path forward to be confirmed.
```
ncdc, line 115: Does this need to be resolved before we merge?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller_test.go
fgiloux: Thanks for pointing that. I have made the change. Note that the previous approach is used for the tests of apiexport and apibinding reconciliation. The previous approach makes a "contain" comparison for "message", whereas I am now ignoring it, which is fine here. It would be possible to have a "contain" comparison with cmp by using a comparer, but was not needed.

fgiloux:
lint is complaining about:
~~~
 test/e2e/virtual/apiexport/authorizer_test.go:494:10: SA1019: export.Status.VirtualWorkspaces is deprecated: use APIExportEndpointSlice.status.endpoints instead (staticcheck)
		if len(export.Status.VirtualWorkspaces) > 0 {
		       ^
test/e2e/virtual/apiexport/authorizer_test.go:502:31: SA1019: export.Status.VirtualWorkspaces is deprecated: use APIExportEndpointSlice.status.endpoints instead (staticcheck)
	serviceApiExportVWCfg.Host = export.Status.VirtualWorkspaces[0].URL 
~~~
but these lines don't exist in the [file](https://github.com/kcp-dev/kcp/blob/3f9673c6cc7c37ed6d51edd8db7f7e915ec18da4/test/e2e/virtual/apiexport/authorizer_test.go#L287-L288): apiExport vs export

stevekuznetsov:
Usually the linter breaks like that if there's a compilation error somewhere else

fgiloux:
> Usually the linter breaks like that if there's a compilation error somewhere else

Here is what happened: Prow merges the PR with the current state of the main branch before running the lint check. New calls to the deprecated API were recently introduced that were breaking the checks. I rebased my PR, annotated the new calls with nolint and fingers crossed it should be fine now.

pkg/cache/server/config.go
p0lyn0mial: are you sure that SSA won't alter the cached objects in anyway? If it does then the replication controller will hot loop.

p0lyn0mial: thanks!

fgiloux: I have added the possibility to deactivate server side apply through parameters in the kubernetes fork. The last commit makes use of it.
If/When the PR for the kubernetes fork merges I can change the go.mod to point back to it instead of my branch.

fgiloux:
@ncdc I have set  k8s.io/apiextensions-apiserver in go.mod to point again to kcp-dev/kubernetes (the ServerSideApply deactivation commit). I left other references to the Kubernetes fork unchanged. Do you want to have it otherwise? Is there anything left before this PR can be merged?

pkg/server/controllers.go
ncdc: @p0lyn0mial would appreciate your thoughts on this

pkg/server/server.go
```diff
@@ -327,25 +327,23 @@ func (s *Server) Run(ctx context.Context) error {
...
 	}
 
-	if s.Options.Cache.Enabled {
-		if err := s.AddPostStartHook("kcp-start-optional-informers", func(hookContext genericapiserver.PostStartHookContext) error {
-			// TODO(p0lyn0mial): failing the optional hook should not render the main server unhealthy
-			logger := logger.WithValues("postStartHook", "kcp-start-optional-informers")
-			s.CacheKcpSharedInformerFactory.Start(hookContext.StopCh)
-			s.CacheKcpSharedInformerFactory.WaitForCacheSync(hookContext.StopCh)
+	if err := s.AddPostStartHook("kcp-start-optional-informers", func(hookContext genericapiserver.PostStartHookContext) error {
+		// TODO(p0lyn0mial): failing the optional hook should not render the main server unhealthy
```
ncdc, line 331: We can remove this comment, can't we?

ncdc: Or does this depend on `s.Options.Cache.KubeconfigFile` being set?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,249 @@
...
+type Resource = committer.Resource[*APIExportEndpointSliceSpec, *APIExportEndpointSliceStatus]
+type CommitFunc = func(context.Context, *Resource, *Resource) error
+
+// controller reconciles APIExportEndpointSlices. It ensures that the shard endpoints are populated
+//  in the status of every APIExportEndpointSlices.
+type controller struct {
+	queue workqueue.RateLimitingInterface
+
+	// kcpClusterClient is cluster aware and used to communicate with kcp API servers
+	kcpClusterClient kcpclientset.ClusterInterface
```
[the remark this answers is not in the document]
fgiloux: yes good point.

test/e2e/virtual/apiexport/virtualworkspace_test.go
```diff
@@ -684,6 +690,7 @@ func TestAPIExportPermissionClaims(t *testing.T) {
 	}, wait.ForeverTestTimeout, 100*time.Millisecond, "waiting on virtual workspace to be ready")
 
 	apiExportVWCfg := rest.CopyConfig(cfg)
+	//nolint:staticcheck // SA1019 VirtualWorkspaces is deprecated but not removed yet
```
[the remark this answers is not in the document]
fgiloux: This is debatable. On one side it makes changes easier on the other side developers don't get the warning anymore, which is the reason for the check in first place, isn't it?
If you feel strongly about it I am fine with making the change.

fgiloux:
rebased on workspace refactoring changes

pkg/server/controllers.go
p0lyn0mial: > Does a shard have some local storage for its' own definition? Surely also there will be APIExports on the shard, that do not need to be requested from the cache - we can progress in processing that data even when the cache is down, right?

yes, so, as of today our approach is to use two informers (maybe in the future we could unify and provide a single instance/interface).

The approeach so far was to first check the local informer and falling back to the second backed by the cache server. For example:
```
getAPIExport: func(clusterName logicalcluster.Name, name string) (*apisv1alpha1.APIExport, error) {
  apiExport, err := localAPIExportInformer.Lister().Cluster(clusterName).Get(name)
  if errors.IsNotFound(err) {
    return cacheApiExportInformer.Lister().Cluster(clusterName).Get(name)
  }
  return apiExport, err
}
```

I think we could apply this strategy to all places that require working with the caching layer.

When it comes to working with `Shard` resources in an event of a failure, we could at least register the shard we are currently running on since all we need is `Spec.VirtualWorkspaceURL`.

Does it make sense?

fgiloux: > I think we could apply this strategy to all places that require working with the caching layer.

I think that we need to discriminate between two types of requests:
- I am looking for APIExport A, or another global resource type, and the existing approach: check locally and then the cache looks ok to me.
- I want to build a complete view and I am looking for **all** shards. This is the scenario here. In that case I don't think that local requests bring some added value. 

> When it comes to working with Shard resources in an event of a failure, we could at least register the shard we are currently running on since all we need is Spec.VirtualWorkspaceURL

If the APIExportEndpointSlice was just created, nothing is broken, just not yet working. If the cache server is "temporarily" not accessible it is debatable but my gut feeling is that it is simpler, clearer, cleaner to wait for it to get reachable again and the reconciliation to happen.
If the APIExportEndpointSlice already existed, and had already been reconciled once, looking at the local shard only, may (not today) return a single shard. Reconciliation against that may mean to remove hundred of valid existing URLs and possibly one invalid (if a shard change was the trigger for the reconciliation) and replace them by a single URL (the local shard). I would argue that it is far worse than doing nothing, keeping  the stale state and waiting for the cache server to be reachable again to have the state finally up-to-date.

to come back to:

> Does a shard have some local storage for its' own definition?

As per my previsous answer: No. I am proposing to change that, to stop having the root shard being special and the cache server becoming **the** preferred communication mechanism between shards. I feel however that it is orthogonal to this pull request.

fgiloux: Added a commit following our [slack discussion](https://kubernetes.slack.com/archives/C021U8WSAFK/p1672668119924479). My preference would be without it.

pkg/apis/apis/v1alpha1/types_apiexportendpointslice.go
```diff
@@ -85,6 +92,20 @@ type APIExportEndpoint struct {
...
+func (in *APIExportEndpointSlice) GetConditions() conditionsv1alpha1.Conditions {
+	return in.Status.Conditions
+}
+
+func (in *APIExportEndpointSlice) SetConditions(conditions conditionsv1alpha1.Conditions) {
+	in.Status.Conditions = conditions
+}
+
+// These are valid conditions of APIExportEndpointSlice in addition to
+// APIExportValid and related reasons defined with the APIBinding type.
```
p0lyn0mial, line 104: is this comment correct?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,265 @@
...
+				}
+			}
+			return shards, err
+
+		},
+		getAPIExportEndpointSlice: func(clusterName logicalcluster.Name, name string) (*apisv1alpha1.APIExportEndpointSlice, error) {
+			return apiExportEndpointSliceClusterInformer.Lister().Cluster(clusterName).Get(name)
+		},
+		getAPIExport: func(clusterName logicalcluster.Name, path logicalcluster.Path, name string) (*apisv1alpha1.APIExport, error) {
+			apiExport, err := localAPIExportClusterInformer.Lister().Cluster(clusterName).Get(name)
```
p0lyn0mial, line 86: I don't think this is correct. I think you should use an indexer as well to get an APIExport by path, no?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
```diff
@@ -0,0 +1,146 @@
...
+
+	virtualworkspacesoptions "github.com/kcp-dev/kcp/cmd/virtual-workspaces/options"
+	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
+	conditionsv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/apis/conditions/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/util/conditions"
+	"github.com/kcp-dev/kcp/pkg/logging"
+	apiexportbuilder "github.com/kcp-dev/kcp/pkg/virtual/apiexport/builder"
+)
+
+func (c *controller) reconcile(ctx context.Context, apiExportEndpointSlice *apisv1alpha1.APIExportEndpointSlice) error {
```
p0lyn0mial, line 39: how does an `apiExportEndpointSlice`get created ? manually by the service provider?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,265 @@
...
+	c := &controller{
+		queue: queue,
+		listAPIExportEndpointSlices: func() ([]*apisv1alpha1.APIExportEndpointSlice, error) {
+			return apiExportEndpointSliceClusterInformer.Lister().List(labels.Everything())
+		},
+		listShards: func() ([]*corev1alpha1.Shard, error) {
+			shards, err := globalShardClusterInformer.Lister().List(labels.Everything())
+			if err != nil {
+				var localErr error
+				shards, localErr = localShardClusterInformer.Lister().List(labels.Everything())
```
p0lyn0mial, line 74: mhm, I think that `localShardClusterInformer` on the `root` shard will only hold some items.

Unless we want to put the list of shards from the cache into a local configmap :)

I was thinking about plugging and using  `c.ShardVirtualWorkspaceURL()`. I was also thinking about using it in the reconciler loop only for new `APIExportEndpointSlices` (as a fallback) not in the `listShards` func.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,265 @@
...
+			}
+			return shards, err
+
+		},
+		getAPIExportEndpointSlice: func(clusterName logicalcluster.Name, name string) (*apisv1alpha1.APIExportEndpointSlice, error) {
+			return apiExportEndpointSliceClusterInformer.Lister().Cluster(clusterName).Get(name)
+		},
+		getAPIExport: func(clusterName logicalcluster.Name, path logicalcluster.Path, name string) (*apisv1alpha1.APIExport, error) {
+			apiExport, err := localAPIExportClusterInformer.Lister().Cluster(clusterName).Get(name)
+			if errors.IsNotFound(err) {
```
p0lyn0mial, line 87: also, just in case, check for any other error.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
```diff
@@ -0,0 +1,146 @@
...
+			virtualworkspacesoptions.DefaultRootPathPrefix,
+			apiexportbuilder.VirtualWorkspaceName,
+			logicalcluster.From(apiExport).String(),
+			apiExport.Name,
+		)
+
+		desiredURLs.Insert(u.String())
+	}
+
+	apiExportEndpointSlice.Status.APIExportEndpoints = nil
```
p0lyn0mial, line 138: If the list of shards is empty (the cache server is unavailable) this will wipe previously set endpoints.

I think in such a case we should fallback on `c.ShardVirtualWorkspaceURL()` but only for new `apiExportEndpointSlice` - so that we won't erase previous work.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
```diff
@@ -0,0 +1,146 @@
...
+		// but provide a minimal, local list if none was populated
+		if apiExportEndpointSlice.Status.APIExportEndpoints != nil || len(shards) == 0 {
+			return fmt.Errorf("error listing Shards: %w", err)
+		}
+	}
+
+	desiredURLs := sets.NewString()
+	for _, shard := range shards {
+		logger = logging.WithObject(logger, shard)
+		if shard.Spec.VirtualWorkspaceURL == "" {
```
p0lyn0mial, line 112: is this normal ? should we report it somehow ?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
```diff
@@ -0,0 +1,146 @@
...
+			continue
+		}
+
+		u, err := url.Parse(shard.Spec.VirtualWorkspaceURL)
+		if err != nil {
+			// Should never happen
+			logger.Error(
```
p0lyn0mial, line 119: will this error contain the name of the `shard` ?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
```diff
@@ -0,0 +1,146 @@
...
+
+	return nil
+}
+
+func (c *controller) updateEndpoints(ctx context.Context,
+	apiExportEndpointSlice *apisv1alpha1.APIExportEndpointSlice,
+	apiExport *apisv1alpha1.APIExport) error {
+	logger := klog.FromContext(ctx)
+	shards, err := c.listShards()
+	if err != nil {
```
p0lyn0mial, line 101: listing from not synced lister won't give you an error but an empty list :(

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,265 @@
...
+	indexers.AddIfNotPresentOrDie(globalAPIExportClusterInformer.Informer().GetIndexer(), cache.Indexers{
+		indexers.ByLogicalClusterPathAndName: indexers.IndexByLogicalClusterPathAndName,
+	})
+
+	apiExportEndpointSliceClusterInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			c.enqueueAPIExportEndpointSlice(obj)
+		},
+		UpdateFunc: func(_, newObj interface{}) {
+			c.enqueueAPIExportEndpointSlice(newObj)
```
p0lyn0mial, line 104: do you know how frequent a `shard` object is updated ?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,265 @@
...
+		},
+	})
+
+	globalShardClusterInformer.Informer().AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj interface{}) {
+				c.enqueueAllAPIExportEndpointSlices(obj)
+			},
+			UpdateFunc: func(_, newObj interface{}) {
+				c.enqueueAllAPIExportEndpointSlices(newObj)
```
p0lyn0mial, line 117: do you know how frequent a `shard` object is updated ?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
fgiloux: > mhm, I think that localShardClusterInformer on the root shard will only hold some items.

That's what I have been saying. Today:
- only the root shard holds Shards objects
- it is IMHO sub-optimal (fragile) to have a "special" shard
 
> Unless we want to put the list of shards from the cache into a local configmap :)

we already have a local cache of the remote cache... 

> I was thinking about plugging and using c.ShardVirtualWorkspaceURL()

mind that some filtering through Partition is added in a later PR. Filtering based on bound APIs should also come later. This will be easier with a uniform approach.

> I was also thinking about using it in the reconciler loop only for new APIExportEndpointSlices (as a fallback) not in the listShards func.

I try to keep the pattern that requests are done through lister/getter functions of the controller. Amongst others it makes easier to test the reconciliation logic.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
fgiloux: It could happen if the Shard has not been reconciled yet, which should eventually happen

pkg/server/server.go
fgiloux: >  the main server should not wait for the cache server - there is plenty of work it could do without the cache

the logic has not changed with this PR. As per other comments the controller for APIExportEndpointSices reconciliation would need the informers related to the cache server to have synced. It may not be the case of other controllers.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
p0lyn0mial: nope, you should waiting for `waitForOptionalSync` when you start your controller.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
p0lyn0mial: I think this is incorrect, we could improve by waiting for `waitForOptionalSync` before starting your controller.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
p0lyn0mial: ok, I see your point, how about:

- waiting for `waitForOptionalSync` before running the controller
- removing `localShardClusterInformer` 
- keeping `localAPIExportClusterInformer`

that would allow for running the controller when we have seen some data from the cache and would allow to reconcile local APIExports even if the connection with the cache has been lost - in that case we would read what we already have in `globalShardClusterInformer`

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// Get APIExport
+	apiExportPath := logicalcluster.NewPath(apiExportEndpointSlice.Spec.APIExport.Path)
+	if apiExportPath.Empty() {
+		apiExportPath = logicalcluster.From(apiExportEndpointSlice).Path()
+	}
+	apiExport, err := c.getAPIExport(apiExportPath, apiExportEndpointSlice.Spec.APIExport.Name)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Don't keep the endpoints if the APIExport has been deleted
```
p0lyn0mial, line 54: what triggers the reconciliation when an APIExport was deleted?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
p0lyn0mial: since we only care about the URL address we could simply compare the old and the new objs and see if it makes sense to trigger the reconciliation

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
fgiloux: True. I am adding event handlers for creation and deletion of APIExports

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
ncdc: Putting aside @fgiloux's suggestion to avoid having the root shard be "special" ... @p0lyn0mial why would the `localShardClusterInformer` on `root` only hold a subset of Shards?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
fgiloux: In the next PR we care about the labels as well

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
fgiloux: added that. Thanks for pointing it out. It will however only work once your PR has merged and waitForOptionalSync has been fixed.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
fgiloux: done

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
fgiloux: done

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
p0lyn0mial: it seems that this change doesn't require a lot. I think it it a 3 liner, no?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
p0lyn0mial: oh, sorry, my mistake, the `root` shard holds the entire set.

p0lyn0mial: @fgiloux since we use `waitForOptionalSync` I think we can remove `localShardClusterInformer`

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,309 @@
...
+			}
+			return shards, err
+
+		},
+		getAPIExportEndpointSlice: func(clusterName logicalcluster.Name, name string) (*apisv1alpha1.APIExportEndpointSlice, error) {
+			return apiExportEndpointSliceClusterInformer.Lister().Cluster(clusterName).Get(name)
+		},
+		getAPIExport: func(path logicalcluster.Path, name string) (*apisv1alpha1.APIExport, error) {
+			apiExport, err := indexers.ByPathAndName[*apisv1alpha1.APIExport](apisv1alpha1.Resource("apiexports"), localAPIExportClusterInformer.Informer().GetIndexer(), path, name)
+			if err != nil {
```
p0lyn0mial, line 85: we should only read from the `globalAPIExportClusterInformer` when the item wasn't found in the `localAPIExportClusterInformer`

I think we could change it to:
```
if err == nil {
  return export, nil
}
if !apierrors.IsNotFound(err) {
  return nil, err
}
return indexers.ByPathAndName[*apisv1alpha1.APIExport](apisv1alpha1.Resource("apiexports"), globalAPIExportClusterInformer.Informer().GetIndexer(), path, name)
```

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,309 @@
...
+}
+
+// enqueueAPIExportEndpointSlicesForAPIExport enqueues APIExportEndpointSlices referencing a specific APIExport.
+func (c *controller) enqueueAPIExportEndpointSlicesForAPIExport(obj interface{}) {
+	key, err := kcpcache.DeletionHandlingMetaClusterNamespaceKeyFunc(obj)
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+	objectsForAPIExport, err := c.getAPIExportEndpointSlicesByAPIExport(key)
```
p0lyn0mial, line 187: I think this looks incorrect given the implementation of `indexAPIExportEndpointSliceByAPIExport`.

the `key` won't have a path rather `a logical cluster name`

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	logger := logging.WithQueueKey(logging.WithReconciler(klog.Background(), ControllerName), key)
+	logger.V(4).Info("queueing APIExportEndpointSlice")
+	c.queue.Add(key)
+}
+
+// enqueueAPIExportEndpointSlicesForAPIExport enqueues APIExportEndpointSlices referencing a specific APIExport.
+func (c *controller) enqueueAPIExportEndpointSlicesForAPIExport(obj interface{}) {
+	key, err := kcpcache.DeletionHandlingMetaClusterNamespaceKeyFunc(obj)
```
p0lyn0mial, line 182: change to:
- read the `core.LogicalClusterPathAnnotationKey` annotation for the given APIExport it will give you a path.
- combine the path with the name to get matching keys of `APIExportEndpointSlices` (`path.Join(export.Name)`)
- use the keys to get `APIExportEndpointSlices` objects use `GetByKey` method on the indexer

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,309 @@
...
+		UpdateFunc: func(_, newObj interface{}) {
+			c.enqueueAPIExportEndpointSlice(newObj)
+		},
+		DeleteFunc: func(obj interface{}) {
+			c.enqueueAPIExportEndpointSlice(obj)
+		},
+	})
+
+	globalAPIExportClusterInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
```
p0lyn0mial, line 121: why do we care when exports are added ?

pkg/apis/apis/v1alpha1/types_apiexportendpointslice.go
```diff
@@ -85,6 +92,20 @@ type APIExportEndpoint struct {
...
+const (
+	APIExportEndpointSliceURLsReady conditionsv1alpha1.ConditionType = "EndpointURLsReady"
+)
```
sttts, line 107: am not convinced of adding conditions to the slice object. It's a purely data output objects, similarly to Endpoints(Slices) in kube.

What should this condition even mean? IMO there is such a think as ready for it, in the sense that ready = in-sync.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,309 @@
...
+func (c *controller) enqueueAllAPIExportEndpointSlices(shard interface{}) {
+	list, err := c.listAPIExportEndpointSlices()
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+
+	logger := logging.WithObject(logging.WithReconciler(klog.Background(), ControllerName), shard.(*corev1alpha1.Shard))
+	for i := range list {
+		key, err := kcpcache.MetaClusterNamespaceKeyFunc(list[i])
```
p0lyn0mial, line 215: not `DeletionHandlingMetaClusterNamespaceKeyFunc` ?

pkg/apis/apis/v1alpha1/types_apiexportendpointslice.go
sttts: ic, this is for `APIExport %s|%s not found` because we have an export reference in the slices. So nvmd.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,309 @@
...
+	obj, err := c.getAPIExportEndpointSlice(clusterName, name)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil // object deleted before we handled it
+		}
+		return err
+	}
+
+	old := obj
+	obj = obj.DeepCopy()
```
p0lyn0mial, line 288: do we need a copy here? A patch request should not modify an object, right?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
sttts: why do we have to look into local shards? Always go through global.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,309 @@
...
+// controller reconciles APIExportEndpointSlices. It ensures that the shard endpoints are populated
+// in the status of every APIExportEndpointSlices.
+type controller struct {
+	queue workqueue.RateLimitingInterface
+
+	listShards                            func() ([]*corev1alpha1.Shard, error)
+	listAPIExportEndpointSlices           func() ([]*apisv1alpha1.APIExportEndpointSlice, error)
+	getAPIExportEndpointSlice             func(clusterName logicalcluster.Name, name string) (*apisv1alpha1.APIExportEndpointSlice, error)
+	getAPIExport                          func(path logicalcluster.Path, name string) (*apisv1alpha1.APIExport, error)
+	getAPIExportEndpointSlicesByAPIExport func(key string) ([]interface{}, error)
```
sttts, line 162: please properly strictly type this method

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
sttts: even this does not work. Refefence can be by the canonical path or by logical cluster name. I.e., an export is known under both and `getAPIExportEndpointSlicesByAPIExport` should better take an `*APIExport` as parameter and hide this logic.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
sttts: because an export reference might become valid

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
sttts: would rather like to see a filtered handler that only handles specific updates.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
p0lyn0mial: seems like we change both `old and obj` in https://github.com/kcp-dev/kcp/blob/8d95d652dd230a8b4a61598a9c7b77b32bf7972d/pkg/reconciler/committer/committer.go#L139

worth making sure we don't change object stored in the cache.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller_test.go
```diff
@@ -0,0 +1,189 @@
...
+		},
+	}
+
+	for name, tc := range tests {
+		tc := tc // to avoid t.Parallel() races
+
+		t.Run(name, func(t *testing.T) {
+
+			c := &controller{
+
```
sttts, line 75: nit: no empty line

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller_test.go
```diff
@@ -0,0 +1,189 @@
...
+			wantAPIExportEndpointSliceURLsReady: true,
+			wantAPIExportValid:                  true,
+		},
+	}
+
+	for name, tc := range tests {
+		tc := tc // to avoid t.Parallel() races
+
+		t.Run(name, func(t *testing.T) {
+
```
sttts, line 73: nit: no empty line

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
```diff
@@ -0,0 +1,146 @@
...
+func (c *controller) updateEndpoints(ctx context.Context,
+	apiExportEndpointSlice *apisv1alpha1.APIExportEndpointSlice,
+	apiExport *apisv1alpha1.APIExport) error {
+	logger := klog.FromContext(ctx)
+	shards, err := c.listShards()
+	if err != nil {
+		// Don't update the endpoint list if the cache server could not be reached
+		// but provide a minimal, local list if none was populated
+		if apiExportEndpointSlice.Status.APIExportEndpoints != nil || len(shards) == 0 {
+			return fmt.Errorf("error listing Shards: %w", err)
```
sttts, line 105: See Slack discussion: https://kubernetes.slack.com/archives/C021U8WSAFK/p1672830942189929

sttts: This logic here is strange: `err != nil` above should imply `len(shards)==0`.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
p0lyn0mial: how can we find a corresponding `APIExportEndpointSlices` for an `APIExport` then?

sttts: Find those referencing by canonical path (annotation), and then those referencing by cluster name. Return the union.

p0lyn0mial: oh, ic, so we miss one more point
... 
- get matching keys of `APIExportEndpointSlices` for `logicalcluster.From(export).Path().Join(export.Name)`

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
```diff
@@ -0,0 +1,142 @@
...
+		logger = logging.WithObject(logger, shard)
+		if shard.Spec.VirtualWorkspaceURL == "" {
+			continue
+		}
+
+		u, err := url.Parse(shard.Spec.VirtualWorkspaceURL)
+		if err != nil {
+			// Should never happen
+			logger.Error(
+				err, "error parsing Shard.Spec.VirtualWorkspaceURL",
```
sttts, line 116: nit: lower-case

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+
+	logger := logging.WithQueueKey(logging.WithReconciler(klog.Background(), ControllerName), key)
+	logger.V(4).Info("queueing APIExportEndpointSlice")
+	c.queue.Add(key)
+}
+
+// enqueueAPIExportEndpointSlicesForAPIExport enqueues APIExportEndpointSlices referencing a specific APIExport.
+func (c *controller) enqueueAPIExportEndpointSlicesForAPIExport(obj interface{}) {
+	export, ok := obj.(*apisv1alpha1.APIExport)
```
sttts, line 171: you must extract the object from a potential tomb stone here.
```suggestion
        if d, ok := obj.(cache.DeletedFinalStateUnknown); ok {
		obj = d.Obj
	}

	export, ok := obj.(*apisv1alpha1.APIExport)
```

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,317 @@
...
+		getAPIExportEndpointSlice: func(clusterName logicalcluster.Name, name string) (*apisv1alpha1.APIExportEndpointSlice, error) {
+			return apiExportEndpointSliceClusterInformer.Lister().Cluster(clusterName).Get(name)
+		},
+		getAPIExport: func(path logicalcluster.Path, name string) (*apisv1alpha1.APIExport, error) {
+			return indexers.ByPathAndName[*apisv1alpha1.APIExport](apisv1alpha1.Resource("apiexports"), apiExportClusterInformer.Informer().GetIndexer(), path, name)
+		},
+		getAPIExportKeys: func(index, keyString string) ([]string, error) {
+			return apiExportEndpointSliceClusterInformer.Informer().GetIndexer().IndexKeys(index, keyString)
+		},
+		getSliceByAPIExportKey: func(key string) (item *apisv1alpha1.APIExportEndpointSlice, exists bool, err error) {
```
sttts, line 80: these are strange abstractions, or actually no abstractions. Instead do:
```
getSlicesForAPIExports(export *APIExport)
```
with two lookups in the `indexAPIExportEndpointSliceByAPIExport` index.

fgiloux: replaced by using the informer directly in the enqueue logic and having a second struct for the reconcile as per your recommendation

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
fgiloux: done

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller_test.go
fgiloux: done

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller_test.go
fgiloux: done

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
fgiloux: removed

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
fgiloux: thanks. done.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
fgiloux: two lookup now. done

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_indexes.go
```diff
@@ -0,0 +1,43 @@
...
+	apiExportEndpointSlice, ok := obj.(*apisv1alpha1.APIExportEndpointSlice)
+	if !ok {
+		return []string{}, fmt.Errorf("obj %T is not an APIExportEndpointSlice", obj)
+	}
+
+	path := logicalcluster.NewPath(apiExportEndpointSlice.Spec.APIExport.Path)
+	if path.Empty() {
+		path = logicalcluster.From(apiExportEndpointSlice).Path()
+	}
+	key := client.ToClusterAwareKey(path, apiExportEndpointSlice.Spec.APIExport.Name)
```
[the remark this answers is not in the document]
fgiloux: done

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
fgiloux: done

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,330 @@
...
+
+	clusterKeys, err := c.apiExportEndpointSliceClusterInformer.Informer().GetIndexer().IndexKeys(indexAPIExportEndpointSliceByAPIExport, logicalcluster.From(export).Path().Join(export.Name).String())
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+	keys.Insert(clusterKeys...)
+
+	for _, key := range keys.List() {
+		slice, exists, err := c.apiExportEndpointSliceClusterInformer.Informer().GetIndexer().GetByKey(key)
```
sttts, line 194: why only one? There can be many, can't there?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,330 @@
...
+		if err != nil {
+			runtime.HandleError(err)
+			continue
+		} else if !exists {
+			runtime.HandleError(fmt.Errorf("APIBinding %q does not exist", key))
```
sttts, line 199: this is not an error. The index can change at any time, especially since the `IndexKeys` call above.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,330 @@
...
+	oldResource := &Resource{ObjectMeta: old.ObjectMeta, Spec: &old.Spec, Status: &old.Status}
+	newResource := &Resource{ObjectMeta: obj.ObjectMeta, Spec: &obj.Spec, Status: &obj.Status}
+	if err := c.commit(ctx, oldResource, newResource); err != nil {
+		errs = append(errs, err)
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+func filterShardEvent(oldObj, newObj interface{}) bool {
```
sttts, line 314: nit: in the filtered event handler for informers the logic is inverse, i.e. false for drop, true for keep. Maybe worth to follow that logic here too?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
fgiloux: ack

PR #2433, opened 2022-11-30
test/e2e/authorizer/authorizer_test: fix TestAuthorizer

p0lyn0mial:
## Related issue(s)

Fixes #


test/e2e/authorizer/authorizer_test.go
```diff
@@ -80,7 +80,11 @@ func TestAuthorizer(t *testing.T) {
 	createResources(t, ctx, dynamicClusterClient, kubeDiscoveryClient, org1.Join("workspace1"), "workspace1-resources.yaml")
 	createResources(t, ctx, dynamicClusterClient, kubeDiscoveryClient, org2.Join("workspace1"), "workspace1-resources.yaml")
 
-	framework.AdmitWorkspaceAccess(t, ctx, kubeClusterClient, org1, []string{"user-1", "user-2", "user-3"}, nil, false)
+	framework.AdmitWorkspaceAccess(t, ctx, kubeClusterClient, org1.Join("workspace1"), []string{"user-1"}, nil, true)
+	framework.AdmitWorkspaceAccess(t, ctx, kubeClusterClient, org1.Join("workspace1"), []string{"user-2"}, nil, false)
+	framework.AdmitWorkspaceAccess(t, ctx, kubeClusterClient, org1.Join("workspace2"), []string{"user-2"}, nil, true)
+	framework.AdmitWorkspaceAccess(t, ctx, kubeClusterClient, org1.Join("workspace2"), []string{"user-3"}, nil, false) //
+	framework.AdmitWorkspaceAccess(t, ctx, kubeClusterClient, org2.Join("workspace1"), []string{"user-2", "user-3"}, nil, false)
```
sttts, line 87: why are these necessary?

p0lyn0mial: required by the test cases defined in `TestAuthorizer` test - https://github.com/kcp-dev/kcp/blob/e782724616bee6291c7cac00276a6e513d3ab7fd/test/e2e/authorizer/authorizer_test.go#L104

p0lyn0mial: I think there isn't a simple answer, we would have to consider each test case separately.

sttts: just wondering why we need them now and not before. We didn't change the behaviour IMO.

s-urbaniak: previously giving access to the org was enough:
```
	framework.AdmitWorkspaceAccess(t, ctx, kubeClusterClient, org1, []string{"user-1", "user-2", "user-3"}, nil, false)
```

`AdmitWorkspaceAccess` gives access to the `content` subresource of the org workspace. Hence, those users didn't need additional permissions on the workspaces.

Now these permissions have to be set on a per workspace basis.

PR #2437, opened 2022-11-30
:seedling: ddsif: don't construct informers manually

stevekuznetsov:
Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

/cc @davidfestal 
/assign @ncdc

PR #2440, opened 2022-11-30
🌱 DDSIF: Make it more generally useful

davidfestal:
## Summary

Make `DynamicDiscoverySharedInformerFactory` more generally useful by:
- allowing it to work without CRDs
- providing 2 versions for cluster-aware and cluster-unaware clients (thanks to generics)

Context: it will make it a perfect fit to use in the Syncer (more precisely in any type of component syncing a dynamic list of GVR from KCP to normal Kube clusters).


davidfestal:
> FYI [#2437 (comment)](https://github.com/kcp-dev/kcp/pull/2437#issuecomment-1332492209)

Yeah, I saw this one.
What about merging this PR now (as long as you're OK with the content ;-) ) even before doing the change tried out in PR #2437 ?
The changes done in this PR wouldn't prevent using global informer factories as a next step, especially when Kube 1.25 allows removing informers from the dynamic factory.

I still need a small lint error though. Apart from this it should be ready to review.

pkg/informer/informer.go
```diff
@@ -37,17 +37,20 @@ import (
...
 	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/restmapper"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 
-	"github.com/kcp-dev/kcp/pkg/indexers"
+	indexerspkg "github.com/kcp-dev/kcp/pkg/indexers"
```
stevekuznetsov, line 53: why?

pkg/informer/informer.go
```diff
@@ -58,14 +61,28 @@ const (
 	byGroupVersionResourceIndex = "byGroupVersionResource"
 )
 
-// DynamicDiscoverySharedInformerFactory is a SharedInformerFactory that
+// GVRSource is a source fo information about available GVRs,
```
stevekuznetsov, line 64:
```suggestion
// GVRSource is a source of information about available GVRs,
```

pkg/informer/informer.go
```diff
@@ -58,14 +61,28 @@ const (
...
+// and provides a way to register for changes in the available GVRs (addition of removal)
```
stevekuznetsov, line 65:
```suggestion
// and provides a way to register for changes in the available GVRs (addition or removal)
```

pkg/informer/informer.go
```diff
@@ -58,14 +61,28 @@ const (
...
+// so that the list of informers can be updated.
+type GVRSource interface {
+	ListGVRs() []schema.GroupVersionResource
+	GetGVRPartialMetadata(gvr schema.GroupVersionResource) (*GVRPartialMetadata, error)
+	Synced() bool
+	Register(notifyUpdateNeeded func())
```
stevekuznetsov, line 71: Can we make this `Changes() <-chan whatever` - callback nesting is usually unpleasant to deal with :)

pkg/informer/informer.go
```diff
@@ -58,14 +61,28 @@ const (
...
+}
+
+// genericDynamicDiscoverySharedInformerFactory is a SharedInformerFactory that
```
stevekuznetsov, line 74: while we're re-naming this, it's not discovery based anymore, and "dynamic" shared informer factories already have a meaning upstream, so we should find a good name to expose that this manages the set of types itself over time

pkg/informer/informer.go
```diff
@@ -787,6 +890,96 @@ func newRESTMapper(fn func() (meta.RESTMapper, error)) restMapper {
 	}
 }
 
+type genericInformerBase[InformerType sharedIndexInformerBase, ListerType genericListerBase] interface {
```
stevekuznetsov, line 893: please put these types at the top, where they are used

pkg/informer/informer.go
```diff
@@ -58,14 +61,28 @@ const (
...
 // dynamically discovers new types and begins informing on them.
-type DynamicDiscoverySharedInformerFactory struct {
-	dynamicClient     kcpdynamic.ClusterInterface
-	filterFunc        func(interface{}) bool
-	indexers          cache.Indexers
-	crdIndexer        cache.Indexer
-	crdInformerSynced cache.InformerSynced
+// It is a generic implementation for both logical-cluster-aware and
+// logical-cluster-unaware (== single cluster, standard kube) informers.
+type genericDynamicDiscoverySharedInformerFactory[InformerType sharedIndexInformerBase, ListerType genericListerBase, GenericInformerType genericInformerBase[InformerType, ListerType]] struct {
```
stevekuznetsov, line 78: nit: `Type` is superfluous - use e.g.

```go
type genericDynamicDiscoverySharedInformerFactory[Informer sharedIndexInformerBase, Lister genericListerBase, GenericInformer genericInformerBase[Informer, Lister]] struct {
```

pkg/informer/informer.go
```diff
@@ -87,14 +104,40 @@ type DynamicDiscoverySharedInformerFactory struct {
...
+// It retrieves GVR-related information from CRD informers.
 func NewDynamicDiscoverySharedInformerFactory(
 	cfg *rest.Config,
 	filterFunc func(obj interface{}) bool,
 	crdInformer kcpapiextensionsv1informers.CustomResourceDefinitionClusterInformer,
 	indexers cache.Indexers,
+) (*DynamicDiscoverySharedInformerFactory, error) {
+	crdIndexer := crdInformer.Informer().GetIndexer()
+
+	// Add an index function that indexes a CRD by its group/version/resource.
```
stevekuznetsov, line 119: How is this unique to the cluster-aware case?

pkg/informer/informer.go
```diff
@@ -87,14 +104,40 @@ type DynamicDiscoverySharedInformerFactory struct {
...
+		return nil, err
+	}
+
+	return NewDelegatingDynamicDiscoverySharedInformerFactory(cfg, filterFunc, indexers, &crdGVRSource{
+		crdInformer: crdInformer,
+		crdIndexer:  crdIndexer,
+	})
+}
+
+// NewDelegatingDynamicDiscoverySharedInformerFactory returns a factory for cluster-aware
```
stevekuznetsov, line 132: I don't think `Delegating` gives much information to the user here - the shared informer factory doesn't delegate it's operation

pkg/informer/informer.go
```diff
@@ -104,20 +147,34 @@ func NewDynamicDiscoverySharedInformerFactory(
...
 	f := &DynamicDiscoverySharedInformerFactory{
-		dynamicClient:     metadataClusterClient,
-		filterFunc:        filterFunc,
-		indexers:          indexers,
-		crdIndexer:        crdInformer.Informer().GetIndexer(),
-		crdInformerSynced: crdInformer.Informer().HasSynced,
+		genericDynamicDiscoverySharedInformerFactory: genericDynamicDiscoverySharedInformerFactory[kcpcache.ScopeableSharedIndexInformer, kcpcache.GenericClusterLister, kcpinformers.GenericClusterInformer]{
+			filterFunc:            filterFunc,
+			indexers:              indexers,
+			listGVRs:              gvrSource.ListGVRs,
```
stevekuznetsov, line 153: why have function pointers to individual work, as opposed to passing the source?

pkg/informer/informer.go
```diff
@@ -143,34 +193,7 @@ func NewDynamicDiscoverySharedInformerFactory(
...
-			newEstablished := crdIsEstablished(newObj)
-			if newEstablished || oldEstablished != newEstablished {
-				notifyUpdateNeeded()
-			}
-		},
-		DeleteFunc: func(obj interface{}) {
-			notifyUpdateNeeded()
-		},
-	})
+	gvrSource.Register(notifyUpdateNeeded)
```
stevekuznetsov, line 196: yeah, just exposing the updatech is enough here

pkg/informer/informer.go
```diff
@@ -197,6 +220,13 @@ func byGroupVersionResourceIndexFunc(obj interface{}) ([]string, error) {
 	return ret, nil
 }
 
+// DynamicDiscoverySharedInformerFactory is a factory for cluster-aware
+// shared informers that discovers new types and informs on updates to resources of
+// those types.
+type DynamicDiscoverySharedInformerFactory struct {
```
stevekuznetsov, line 226: what's the point of this type?

pkg/informer/informer.go
```diff
@@ -219,15 +249,115 @@ func (d *scopedDynamicDiscoverySharedInformerFactory) ForResource(gvr schema.Gro
 	return clusterInformer.Cluster(d.cluster), nil
 }
 
+// NewSingleClusterDynamicDiscoverySharedInformerFactory returns a factory for
```
stevekuznetsov, line 252:
```suggestion
// NewScopedDynamicDiscoverySharedInformerFactory returns a factory for
```

this is the wording used today

pkg/informer/informer.go
```diff
@@ -219,15 +249,115 @@ func (d *scopedDynamicDiscoverySharedInformerFactory) ForResource(gvr schema.Gro
...
+// cluster-unaware (standard Kube) shared informers that discovers new types
+// and informs on updates to resources of those types.
+// It retrieves GVR-related information from CRD informers.
+func NewSingleClusterDynamicDiscoverySharedInformerFactory(
+	cfg *rest.Config,
+	filterFunc func(obj interface{}) bool,
+	crdInformer kcpapiextensionsv1informers.CustomResourceDefinitionClusterInformer,
```
stevekuznetsov, line 259: if this is cluster-unaware, why are we taking a cluster-aware informer?

pkg/informer/informer.go
```diff
@@ -219,15 +249,115 @@ func (d *scopedDynamicDiscoverySharedInformerFactory) ForResource(gvr schema.Gro
...
+		crdInformer: crdInformer,
+		crdIndexer:  crdIndexer,
+	})
+}
+
+// NewSingleClusterDynamicDiscoverySharedInformerFactory returns a factory for
+// cluster-unaware (standard Kube) shared informers that discovers new types
+// and informs on updates to resources of those types.
+// It receives the GVR-related information by delegating to a GVRSource.
+func NewSingleClusterDelegatingDynamicDiscoverySharedInformerFactory(
```
stevekuznetsov, line 281: Why don't we just have one parameterized constructor? What's the diff here?

pkg/informer/informer.go
```diff
@@ -787,6 +890,96 @@ func newRESTMapper(fn func() (meta.RESTMapper, error)) restMapper {
 	}
 }
 
+type genericInformerBase[InformerType sharedIndexInformerBase, ListerType genericListerBase] interface {
+	Informer() InformerType
+	Lister() ListerType
+}
+
+type sharedIndexInformerBase interface {
```
stevekuznetsov, line 898: Isn't there already an interface upstream for this?

pkg/informer/informer.go
```diff
@@ -425,14 +545,14 @@ func withGVRPartialMetadata(scope apiextensionsv1.ResourceScope, kind, singular
 	}
 }
 
-type gvrPartialMetadata struct {
+type GVRPartialMetadata struct {
```
stevekuznetsov, line 548: Why does this have to leak upwards?

pkg/informer/informer.go
```diff
@@ -453,21 +573,15 @@ func builtInInformableTypes() map[schema.GroupVersionResource]gvrPartialMetadata
...
-	crdGVRs := d.crdIndexer.ListIndexFuncValues(byGroupVersionResourceIndex)
-	for _, s := range crdGVRs {
-		parts := strings.Split(s, "/")
-		group := parts[0]
-		version := parts[1]
-		resource := parts[2]
-		gvr := gvrFor(group, version, resource)
-
+	crdGVRs := d.listGVRs()
+	for _, gvr := range crdGVRs {
```
stevekuznetsov, line 584: Having the interface know how to divulge GVRs and also then know how to turn them into partial metadata seems weird - just have one list call that knows how to return some super-type with both bits of information in there.

pkg/informer/informer.go
stevekuznetsov: Seems like this should be closed over by the crd source - that's where the index is used and must be set

pkg/informer/informer.go
stevekuznetsov: Maybe `SlefDiscoveringSharedInformerFactory` or something, idk

pkg/informer/informer.go
davidfestal: I don't see how it would allow to implement what is [done in the `crdGVRSource`](https://github.com/kcp-dev/kcp/pull/2440/files#diff-e35b562e46d045590564268b506fe50d11eb737d43679891bac1238dd5c48762R964).
But maybe I'm simply not used enough to working with channels.

pkg/informer/informer.go
davidfestal: In my understanding, the fact that it was called Dynamic**Discovery**SharedInformerFactory is more because it implements the `discovery.ServerResourcesInterface` interface, but in a dynamic way (you can subscribe), so that you don't need to do some regular polling of the remote discovery client.
In this sense it is still a `DynamicDiscoverySharedInformerFactory`, whatever be the source of the GVR list it is based on (CRD informer or anything else)

ncdc: The origin of "discovery" in the name is because it used to use discovery to determine the available APIs. That was inefficient (scaled linearly with the # of workspaces) and we replaced it with the combination of known built-in types plus all the CRDs in the system.

pkg/informer/informer.go
davidfestal: I'd even return a `map[schema.GroupVersionresource]GVRPartialMetadata` then.

pkg/informer/informer.go
davidfestal: Well, it wold be used by external implementations of `GVRSource`

pkg/informer/informer.go
davidfestal: Possibly, but I don't know it. Any pointer ?

davidfestal: Seems to be part of `client-go` `tools/cache/SharedInformer` interface, but the `SharedInterface` contains more than that.
I haven't found any interface that only provides one or several of the methods we need here.

davidfestal: I finally used `tools/cache/SharedIndexInformer`

pkg/informer/informer.go
stevekuznetsov: Do you want me to provide you with an example, or are you going to poke around with an implementation on your own

davidfestal: This should be good finally. Thank you.

davidfestal: I'm preparing to push fixes to this PR, but first checking it in the context of my Proof PR (with the refactored Syncer using it)

pkg/informer/informer.go
davidfestal: Because there was a naming conflict at some point. But it's fixed now.

pkg/informer/informer.go
davidfestal: I implemented it this way: https://github.com/kcp-dev/kcp/blob/2ea4239c59b359f7262da33eb05dc0d064e6c230/pkg/informer/informer.go#L162-L172
And the `Changes()` channel, in the case of CRDs is fed this way:   https://github.com/kcp-dev/kcp/blob/2ea4239c59b359f7262da33eb05dc0d064e6c230/pkg/informer/informer.go#L922-L945

pkg/informer/informer.go
davidfestal: So I renamed the types this way:
- the cluster-aware one: `DiscoveringDynamicSharedInformerFactory` (it is a dedicated type since it has the additional `Cluster` method) - same as the old one.
- the generic one `GenericDiscoveringDynamicSharedInformerFactory`, serving for the cluster-unaware case, and as a basis for the cluster-aware case (above).

And the `NewXXX` methods are named this way:
- `NewGenericDiscoveringDynamicSharedInformerFactory()` creates a **generic** ddsif from a `GVRSource` and a `newInformer()` function (`newInformer` will be different according to the case: cluster-aware / cluster-unaware)
- `NewScopedDiscoveringDynamicSharedInformerFactory()` creates a **cluster-unaware** ddsif from a `gvrSource` and a dynamic client (`dynamic.Interface`)
- `NewDiscoveringDynamicSharedInformerFactory()` creates a **cluster-aware** ddsif (`DiscoveringDynamicSharedInformerFactory`) from a `GVRSource` and a dynamic cluster client (`kcpdynamic.ClusterInterface`)
- `NewSelfDiscoveringDynamicSharedInformerFactory() ` creates a **cluster-aware** ddsif (`DiscoveringDynamicSharedInformerFactory`) from a CRD informer and a REST config. It is the same as the old one.

pkg/informer/informer.go
davidfestal: Done

pkg/informer/informer.go
davidfestal: Done

pkg/informer/informer.go
davidfestal: Fixed

pkg/informer/informer.go
davidfestal: fixed

pkg/informer/informer.go
davidfestal: true

pkg/informer/informer.go
davidfestal: Done

pkg/informer/informer.go
davidfestal: Done

pkg/informer/informer.go
davidfestal: fixed

pkg/informer/informer.go
davidfestal: I added a parameterized constructor with the common code.

pkg/informer/informer.go
```diff
@@ -126,108 +158,150 @@ func NewDynamicDiscoverySharedInformerFactory(
...
 	return f, nil
 }
 
-func byGroupVersionResourceKeyFunc(group, version, resource string) string {
-	return fmt.Sprintf("%s/%s/%s", group, version, resource)
+// NewScopedDiscoveringDynamicSharedInformerFactory returns a factory for
+// cluster-unaware (standard Kube) shared informers that discovers new types
+// and informs on updates to resources of those types.
+// It receives the GVR-related information by delegating to a GVRSource.
+func NewScopedDiscoveringDynamicSharedInformerFactory(
```
stevekuznetsov, line 180: We don't need this constructor - what keeps callers from just using the version with generics and the appropriate GVRSource?

pkg/informer/informer.go
```diff
@@ -126,108 +158,150 @@ func NewDynamicDiscoverySharedInformerFactory(
...
+		GenericDiscoveringDynamicSharedInformerFactory: f,
+	}, err
+}
+
+// DiscoveringDynamicSharedInformerFactory is a factory for cluster-aware
+// shared informers that discovers new types and informs on updates to resources of
+// those types.
+// It offers an additional `Cluster()` method that returns a new shared informer factory
+// based on the main informer factory, but scoped for a given logical cluster.
+type DiscoveringDynamicSharedInformerFactory struct {
```
stevekuznetsov, line 270: We also don't need this type - just the one that it embeds.

pkg/informer/informer.go
davidfestal: This constructor includes the implementation of the `newInformer()` function. It is obviously specific to the cluster-unaware case, but still part of the ddsif logic. So I don't see how it would make sense to have every caller define it.
IMO it makes sense to be encapsulated in this package.

pkg/informer/informer.go
```diff
@@ -126,108 +158,150 @@ func NewDynamicDiscoverySharedInformerFactory(
...
 
-func byGroupVersionResourceIndexFunc(obj interface{}) ([]string, error) {
-	crd, ok := obj.(*apiextensionsv1.CustomResourceDefinition)
-	if !ok {
-		return nil, fmt.Errorf("%T is not a CustomResourceDefinition", obj)
+// NewSelfDiscoveringDynamicSharedInformerFactory returns a factory for cluster-aware
+// shared informers that discovers new types and informs on updates to resources of
+// those types.
+// It retrieves GVR-related information from CRD informers.
+func NewSelfDiscoveringDynamicSharedInformerFactory(
```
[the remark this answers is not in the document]
davidfestal: I agree we should be able to remove this constructor, and directly pass the CRD-based GVR source (whose constructor we'll have to make public then). But then we'll also have to create the `metadataClusterClient` from the rest config at call site. If you also agree with that I can proceed.

pkg/informer/informer.go
davidfestal: Where will you define the additional `Cluster()` method then ? It makes no sense for the cluster-unaware case not for the generic case ?

pkg/informer/informer.go
```diff
@@ -126,108 +158,150 @@ func NewDynamicDiscoverySharedInformerFactory(
...
+func NewDiscoveringDynamicSharedInformerFactory(
+	dynamicClusterClient kcpdynamic.ClusterInterface,
+	filterFunc func(obj interface{}) bool,
+	tweakListOptions dynamicinformer.TweakListOptionsFunc,
+	gvrSource GVRSource,
+	indexers cache.Indexers,
+) (*DiscoveringDynamicSharedInformerFactory, error) {
+	f, err := NewGenericDiscoveringDynamicSharedInformerFactory[kcpcache.ScopeableSharedIndexInformer, kcpcache.GenericClusterLister](
+		func(gvr schema.GroupVersionResource, resyncPeriod time.Duration, indexers cache.Indexers) kcpinformers.GenericClusterInformer {
+			indexers[kcpcache.ClusterIndexName] = kcpcache.ClusterIndexFunc
```
stevekuznetsov, line 245: Does NewFilteredDynamicInformer not do this for you for free?

pkg/informer/informer.go
```diff
@@ -126,108 +158,150 @@ func NewDynamicDiscoverySharedInformerFactory(
...
+			indexers[kcpcache.ClusterAndNamespaceIndexName] = kcpcache.ClusterAndNamespaceIndexFunc
+			return kcpdynamicinformer.NewFilteredDynamicInformer(
+				dynamicClusterClient,
+				gvr,
+				resyncPeriod,
+				indexers,
+				nil,
```
stevekuznetsov, line 252: What is this `nil`? Why?

pkg/informer/informer.go
stevekuznetsov: Can we find a factoring that doesn't require these extra constructors?

pkg/informer/informer.go
stevekuznetsov: Ah sorry, I was expecting the cluster-unaware type to be called Scopable and the aware type to be Scoped, and got confused by this type name.

pkg/informer/informer.go
davidfestal: I tried, but I might have missed something.

stevekuznetsov: OK, let's keep it this way then!

pkg/informer/informer.go
```diff
@@ -58,14 +62,44 @@ const (
...
+// so that the list of informers can be updated.
+type GVRSource interface {
+	GVRs() map[schema.GroupVersionResource]GVRPartialMetadata
+	Synced() bool
+	Changes() <-chan struct{}
+	FilterBuiltinGVRs(gvr schema.GroupVersionResource) bool
+}
+
+type genericInformerBase[Informer cache.SharedIndexInformer, Lister genericListerBase] interface {
+	Informer() Informer
```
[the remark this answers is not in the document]
davidfestal: What do you mean ? They are the type parameters (initially they were named `InformerType` and `ListerType`).

pkg/informer/informer.go
davidfestal: Should be `tweakListOptions`. I'll fix it.

pkg/informer/informer.go
davidfestal: It isn't. It is done in the `dynamicSharedInformerFactory.ForResource()` implementation, but here we create the informer directly (for now).

davidfestal: BTW I just moved code that was already adding those indexes.

pkg/informer/informer.go
davidfestal: Fixed

pkg/informer/informer.go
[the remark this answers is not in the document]
davidfestal: I did it in a distinct commit https://github.com/kcp-dev/kcp/pull/2440/commits/601bccb0e887d15b25c1d2915b656ab972494176

Do you prefer it like that ? I not I can revert.

pkg/informer/informer.go
```diff
@@ -58,14 +62,44 @@ const (
...
+type GVRPartialMetadata struct {
+	Names apiextensionsv1.CustomResourceDefinitionNames
+	Scope apiextensionsv1.ResourceScope
+}
+
+// GVRSource is a source of information about available GVRs,
+// and provides a way to register for changes in the available GVRs (addition or removal)
+// so that the list of informers can be updated.
+type GVRSource interface {
+	GVRs() map[schema.GroupVersionResource]GVRPartialMetadata
```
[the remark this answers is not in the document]
davidfestal: Done

ncdc:
/hold

pending workspace refactor

davidfestal:
@stevekuznetsov I answered all your comments afaict. Is it OK for you ?
I also rebased on top of main, so that it should not be held anymlore by the workspace refactoring. cc @ncdc

pkg/metadata/dynamic.go
```diff
@@ -41,6 +42,19 @@ func NewDynamicMetadataClusterClientForConfig(config *rest.Config) (kcpdynamic.C
 	return kcpdynamic.NewForConfig(&metadataConfig)
 }
 
+// NewDynamicMetadataClientForConfig returns a dynamic cluster client that only
```
ncdc, line 45:
```suggestion
// NewDynamicMetadataClientForConfig returns a dynamic client that only
```

pkg/informer/informer.go
```diff
@@ -787,6 +847,114 @@ func newRESTMapper(fn func() (meta.RESTMapper, error)) restMapper {
...
+		result[gvr] = withGVRPartialMetadata(crd.Spec.Scope, crd.Spec.Names.Kind, crd.Spec.Names.Singular)
+	}
+	return result
+}
+
+func (s *crdGVRSource) Synced() bool {
+	return s.crdInformer.HasSynced()
+}
+
+func (s *crdGVRSource) crdIsEstablished(obj interface{}) bool {
```
ncdc, line 919: I don't think this needs any receiver?

pkg/informer/informer.go
```diff
@@ -58,14 +59,64 @@ const (
...
+type GVRSource interface {
+	// GVRs returns the required metadata about all GVRs known by the GVRSource
+	GVRs() map[schema.GroupVersionResource]GVRPartialMetadata
+
+	// Synced returns true if informers used by this GVRSource are synced
+	Synced() bool
+
+	// Changes returns a channel to which the GVRSource writes whenever
+	// its list of known GVRs has changed
+	Changes() <-chan struct{}
```
ncdc, line 81: I would probably call this something like `Subscribe()` given that, at least for the CRD GVR Source implementation, each time you call the function, it returns a new channel.

pkg/informer/informer.go
```diff
@@ -787,6 +847,114 @@ func newRESTMapper(fn func() (meta.RESTMapper, error)) restMapper {
...
+	if !ok {
+		utilruntime.HandleError(fmt.Errorf("unable to determine if CRD is established - got type %T instead", obj))
+		return false
+	}
+
+	return apihelpers.IsCRDConditionTrue(crd, apiextensionsv1.Established)
+}
+
+func (s *crdGVRSource) Changes() <-chan struct{} {
+	changes := make(chan struct{})
```
ncdc, line 930: We might need to make this buffered (1) and then use a `select` statement to try to send to the channel, or drop if already full. Otherwise, the event handler functions below block until a receiver can pop from the channel.

ncdc:
No major concerns. If you're willing, maybe you could remove the "Discovery"/"Discovering" terms from here - it's just a dynamic factory at this point, no discovery anywhere.

pkg/informer/informer.go
```diff
@@ -58,14 +59,64 @@ const (
...
+
+	// Changes returns a channel to which the GVRSource writes whenever
+	// its list of known GVRs has changed
+	Changes() <-chan struct{}
+
+	// FilterBuiltinGVRs is a function that allows filtering out
+	// builtin GVRs that should not be considered.
+	// The function should return false to ignore a builtin GVR,
+	// and true to keep it.
+	FilterBuiltinGVRs(gvr schema.GroupVersionResource) bool
```
sttts, line 87: strange method. If something is not to be considered by the consumer of this interface, don't list it in `GVRs()`.

pkg/informer/informer.go
```diff
@@ -58,14 +59,64 @@ const (
...
+	Names apiextensionsv1.CustomResourceDefinitionNames
+	Scope apiextensionsv1.ResourceScope
+}
+
+// GVRSource is a source of information about available GVRs,
+// and provides a way to register for changes in the available GVRs (addition or removal)
+// so that the list of informers can be updated.
+type GVRSource interface {
+	// GVRs returns the required metadata about all GVRs known by the GVRSource
+	GVRs() map[schema.GroupVersionResource]GVRPartialMetadata
```
sttts, line 74: what is *required* metadata?

pkg/informer/informer.go
```diff
@@ -58,14 +59,64 @@ const (
...
+
+	// Synced returns true if informers used by this GVRSource are synced
+	Synced() bool
```
sttts, line 77: Unclear semantics. As GVRs is giving me potentially incomplete information anyway, what does this func give me?

pkg/informer/informer.go
```diff
@@ -58,14 +59,64 @@ const (
...
+// for which an informer should be started.
+type GVRPartialMetadata struct {
+	Names apiextensionsv1.CustomResourceDefinitionNames
+	Scope apiextensionsv1.ResourceScope
+}
+
+// GVRSource is a source of information about available GVRs,
+// and provides a way to register for changes in the available GVRs (addition or removal)
+// so that the list of informers can be updated.
+type GVRSource interface {
```
sttts, line 72: Is this more than 
```
type Resources struct {
  discovery.ServerResourcesInterface
  Changes() <- chan struct{}
}
```
?

Introducing yet another discovery concept feels wrong.

pkg/informer/informer.go
davidfestal: The channel returned by `Changes()` is itself "forwarded" to a channel which follows the rules you're mentioning: https://github.com/davidfestal/kcp/blob/b5eef4f55befb1492d2171c9b3b320a22c3cf016/pkg/informer/informer.go#L179
Isn't it sufficient ?

davidfestal:
> No major concerns. If you're willing, maybe you could remove the "Discovery"/"Discovering" terms from here - it's just a dynamic factory at this point, no discovery anywhere.

Well, it's still discovering and managing GVRs through a GVRSource.

pkg/informer/informer.go
```diff
@@ -126,108 +175,126 @@ func NewDynamicDiscoverySharedInformerFactory(
...
-func (d *DynamicDiscoverySharedInformerFactory) ForResource(gvr schema.GroupVersionResource) (kcpkubernetesinformers.GenericClusterInformer, error) {
+// by calling Start on the GenericDiscoveringDynamicSharedInformerFactory before the GenericInformer can be used.
+func (d *GenericDiscoveringDynamicSharedInformerFactory[Informer, Lister, GenericInformer]) ForResource(gvr schema.GroupVersionResource) (GenericInformer, error) {
 	// See if we already have it
 	d.informersLock.RLock()
 	inf := d.informers[gvr]
 	d.informersLock.RUnlock()
 
-	if inf != nil {
+	if (genericInformerBase[Informer, Lister])(inf) != nil {
```
sttts, line 297: surprising type cast

pkg/informer/informer.go
```diff
@@ -58,14 +59,64 @@ const (
...
 // dynamically discovers new types and begins informing on them.
-type DynamicDiscoverySharedInformerFactory struct {
-	dynamicClient     kcpdynamic.ClusterInterface
-	filterFunc        func(interface{}) bool
-	indexers          cache.Indexers
-	crdIndexer        cache.Indexer
-	crdInformerSynced cache.InformerSynced
+// It is a generic implementation for both logical-cluster-aware and
+// logical-cluster-unaware (== single cluster, standard kube) informers.
+type GenericDiscoveringDynamicSharedInformerFactory[Informer cache.SharedIndexInformer, Lister genericListerBase, GenericInformer genericInformerBase[Informer, Lister]] struct {
```
sttts, line 115: why does this need to be parametrized by informers and listers? This makes everything here much more complicated.

sttts:
> providing 2 versions for cluster-aware and cluster-unaware clients (thanks to generics)

Why do we need this?

pkg/informer/informer.go
```diff
@@ -87,33 +138,31 @@ type DynamicDiscoverySharedInformerFactory struct {
...
-// NewDynamicDiscoverySharedInformerFactory returns a factory for shared
-// informers that discovers new types and informs on updates to resources of
-// those types.
-func NewDynamicDiscoverySharedInformerFactory(
-	cfg *rest.Config,
+// NewGenericDiscoveringDynamicSharedInformerFactory returns is an informer factory that
+// dynamically discovers new types and begins informing on them.
+// It is a generic implementation for both logical-cluster-aware and
+// logical-cluster-unaware (== single cluster, standard kube) informers.
+func NewGenericDiscoveringDynamicSharedInformerFactory[Informer cache.SharedIndexInformer, Lister genericListerBase, GenericInformer genericInformerBase[Informer, Lister]](
```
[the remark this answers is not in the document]
sttts: Would just call it DynamicSharedInformerFactory

pkg/informer/informer.go
sttts: clarified via Slack. Still kind of ugly, but ok.

pkg/informer/informer.go
davidfestal: This is, in addition to the GVR, the minimal API-related metadata, so: scope, kind and singular name.

pkg/informer/informer.go
davidfestal: The `GVRSource` is possibly based on an informer (CRD informer, or SyncTarget informer in the case of the Syncer). This `Synced`  method returns true when the GVRSource is ready to return a complete list of supported GVRs (=> typically when the underlying informer is Synced).
Maybe this function should be called `Ready` ?

sttts:
Generally, I think it's not a good idea to expose a generic type and generic constructors of weak abstractions. Would prefer to expose exactly two variants and stop leaking the complexity elsewhere.

pkg/informer/informer.go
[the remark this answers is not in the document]
davidfestal: We know `DynamicSharedInformerFactory` is already something in Kube: https://pkg.go.dev/k8s.io/client-go@v0.24.4/dynamic/dynamicinformer#DynamicSharedInformerFactory.
And if I'm not mistaken it is `Dynamc` because is provides informers for dynamic resources (backed by unstructured objects). However it doesn't "discover" the available GVRs provided by the server, and assumes all the GVRs requested by `ForResource()` are provided by the server. If it is not the case the retrieved Informer will never be Synced.

The `DiscoveringDynamicSharedInformerFactory`does something more. It maintains informers for a discovered list of GVRs known to be exposed by the server. That's why we renamed it **Discovering**DynamicSharedInformerFactory.

I agree it's probably not the best name, but giving the same name as something well-defined in Kube seems a bit weird to me as well.

davidfestal:
> Generally, I think it's not a good idea to expose a generic type and generic constructors of weak abstractions. Would prefer to expose exactly two variants and stop leaking the complexity elsewhere.

Not exposing the generic type (which was private), and exposing publicly several non-generic types and constructors was the initial approach (see [first commit](https://github.com/kcp-dev/kcp/pull/2440/commits/06670f25e94b3102c151ca083bc37d16678027a0#diff-e35b562e46d045590564268b506fe50d11eb737d43679891bac1238dd5c48762)), which was changed after the PR comments to use reduce the number of constructors and use generic types / constructors as much as possible.

See for example the following comments: https://github.com/kcp-dev/kcp/pull/2440#discussion_r1047946273, https://github.com/kcp-dev/kcp/pull/2440#discussion_r1037194453

pkg/metadata/dynamic.go
davidfestal: Fixed

pkg/informer/informer.go
davidfestal: Fixed, and updated the comment to mention it typically returns a new channel

pkg/informer/informer.go
davidfestal: It mainly came from refactoring the existing code. I removed it and included the addition of builtin GVRs in the CRD-based `GVRSource`.

pkg/informer/informer.go
davidfestal: I renamed this `Ready` as proposed.

pkg/informer/informer.go
davidfestal: This is required because `nil` is untyped.

Doing:

```golang
if inf != nil {
```

Raises the following error: `invalid operation: cannot compare inf != nil (mismatched types GenericInformer and untyped nil)`.

stevekuznetsov:
> Generally, I think it's not a good idea to expose a generic type and generic constructors of weak abstractions. Would prefer to expose exactly two variants and stop leaking the complexity elsewhere.

Yeah, my preference was to just expose the generic and let downstream consumers use it directly as necessary, but if we have implicit interdependencies between types and can't close over the creation func, I agree the semi-generic semi-not is also a silly end-state. :|

davidfestal:
> creation

@sttts @stevekuznetsov Can we merge as is and plan a followup issue to implement the right balance between your 2 approaches when it is found ? If possible and if it's the last remaining comment on this PR, I'd rather avoid many additional review rounds before merging, knowing that I have 2 levels of pending PRs depending on this one.

ncdc:
I would prefer to untangle the generics and go with the 2 distinct types and 2 distinct constructors (cluster-scoped, single-cluster).

davidfestal:
> > Generally, I think it's not a good idea to expose a generic type and generic constructors of weak abstractions. Would prefer to expose exactly two variants and stop leaking the complexity elsewhere.
> 
> Yeah, my preference was to just expose the generic and let downstream consumers use it directly as necessary, but if we have implicit interdependencies between types and can't close over the creation func, I agree the semi-generic semi-not is also a silly end-state. :|

davidfestal:
> > > Generally, I think it's not a good idea to expose a generic type and generic constructors of weak abstractions. Would prefer to expose exactly two variants and stop leaking the complexity elsewhere.
> > 
> > 
> > Yeah, my preference was to just expose the generic and let downstream consumers use it directly as necessary, but if we have implicit interdependencies between types and can't close over the creation func, I agree the semi-generic semi-not is also a silly end-state. :|

> I would prefer to untangle the generics and go with the 2 distinct types and 2 distinct constructors (cluster-scoped, single-cluster).

:-(

pkg/informer/informer.go
davidfestal: Fixed in Commit d991a46252745e8c5dffe3c7df404c214559bcac

ncdc:
Chatted with @sttts on Slack. Ok to proceed as-is as long as we follow-up later and do some cleaning up. Thanks!

ncdc (approved):

ncdc:
e2e-shared failed because a workspace never transitioned from initializing to ready. It kept waiting on an apibinding to be ready, but I saw this in the log:

```
E0104 17:02:24.849541     658 apibinding_controller.go:410] "kcp-apibinding" controller failed to sync "156fwnzyw1wmuhcf|workload.kcp.io-8xcjg", err: failed to patch *v1alpha1.APIBinding 156fwnzyw1wmuhcf|workload.kcp.io-8xcjg: APIBinding.apis.kcp.io "workload.kcp.io-8xcjg" is invalid: spec.reference: Invalid value: "object": internal error: runtime error: index out of range [1] with length 0 evaluating rule: APIExport reference must not be changed
```

PR #2442, opened 2022-12-01
🌱 pkg/authorization: switch to audit logger, and anonymizer, improve audit log messages

s-urbaniak:
## Summary
This streamlines the current authorizers and removes a lot of boilerplate.

**Note:** There is a hack in the workspace content authorizer which manipulates in-flight attributes. The only alternative right now is to leave that authorizer as-is as it is going to be removed anyways.

## Related issue(s)

Fixes #


pkg/authorization/audit_logger.go
```diff
@@ -52,6 +53,10 @@ func NewAuditLogger(auditPrefix string, delegate authorizer.Authorizer) authoriz
 			auditPrefix+"/"+auditReason, auditReasonMsg,
 		)
 
+		// Note: this deviates from upstream which doesn't log audit reasons.
+		// We should rethink if this should stay.
+		klog.V(5).Info(auditReasonMsg)
```
sttts, line 58: our tests run V(4). Would suggest to use the same here.

pkg/authorization/maximal_permission_policy_authorizer.go
```diff
@@ -140,30 +116,15 @@ func (a *MaximalPermissionPolicyAuthorizer) Authorize(ctx context.Context, attr
 
 	// If we can't find the export default to close
 	if !found {
-		kaudit.AddAuditAnnotations(
-			ctx,
-			MaximalPermissionPolicyAuditDecision, DecisionNoOpinion,
-			MaximalPermissionPolicyAuditReason, fmt.Sprintf("API export %q not found, path: %q", exportName, path),
-		)
-		return authorizer.DecisionNoOpinion, MaximalPermissionPolicyAccessNotPermittedReason, err
+		return authorizer.DecisionNoOpinion, MaximalPermissionPolicyAccessNotPermittedReason, fmt.Errorf("API export %q not found, path: %q", exportName, path)
```
sttts, line 119: this is strange. Why not `authorizer.DecisionNoOpinion, fmt.Errorf("API export %q not found, path: %q", exportName, path), err` ?

pkg/authorization/delegating_authorizer.go
```diff
@@ -0,0 +1,73 @@
...
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+)
+
+// NewConditionalDelegatingAuthorizer returns an authorizer that takes an authorizer
+// and delegates if the authorization decision is Allowed,
+// then returning the delegate's decision, reason, and error.
+//
+// Else, the delegate is not invoked at all
+// and the authorizer's decision (NoDecision, Denied), reason, and error is returned.
+func NewConditionalDelegatingAuthorizer(primary, delegate authorizer.Authorizer) authorizer.Authorizer {
```
sttts, line 32: not sure I like this construct. It's hard to understand what's going on. Would rather drop the annotations (and logging) for the delegation case.

pkg/authorization/audit_logger.go
s-urbaniak: changed to V(4)

pkg/authorization/maximal_permission_policy_authorizer.go
s-urbaniak: fixed

s-urbaniak: side note: the err was a bug, there is no error present in this case, hence changed to `return authorizer.DecisionNoOpinion, fmt.Sprintf("API export %q not found, path: %q", exportName, path), nil`

pkg/authorization/delegating_authorizer.go
s-urbaniak: yes, it's ugly, let's remove it. I'll:
- remove the delegation authorizer
- reintroduce delegation to authorizers
- drop audit logs in case of Allow decisions

s-urbaniak:
@sttts ptal

pkg/authorization/audit_logger.go
```diff
@@ -40,6 +41,10 @@ func NewAuditLogger(auditPrefix string, delegate authorizer.Authorizer) authoriz
 
 	return authorizer.AuthorizerFunc(func(ctx context.Context, attr authorizer.Attributes) (authorizer.Decision, string, error) {
 		dec, reason, err := delegate.Authorize(ctx, attr)
+		// skip logging allow decisions to not flood the audit log
```
ncdc, line 44: Do you want to skip adding the audit annotations in this case too (from L54-58)?

s-urbaniak: that was @sttts's OOB proposal, but I can just skip the logs.

s-urbaniak:
Let me change one more thing here, I am currently not satisfied how decisions end up in the audit log. I want to improve here.

s-urbaniak:
@sttts @ncdc now with the newest commit I think the audit logs look much cleaner and are way easier to reason about:

happy case:
```
  "annotations": {
    "authorization.k8s.io/decision": "allow",
    "authorization.k8s.io/reason": "access granted",
    "bootstrap.authorization.kcp.dev/decision": "Allowed",
    "bootstrap.authorization.kcp.dev/reason": "bootstrap policy reason: RBAC: allowed by ClusterRoleBinding \"system:kcp:tenancy:workspace-bootstrapper\" of ClusterRole \"system:kcp:tenancy:workspace-bootstrapper\" to Group \"system:kcp:tenancy>
    "content.authorization.kcp.dev/decision": "Allowed",
    "content.authorization.kcp.dev/reason": "systemcrd.authorization.kcp.dev: access granted",
    "maxpermissionpolicy.authorization.kcp.dev/decision": "Allowed",
    "maxpermissionpolicy.authorization.kcp.dev/reason": "bootstrap.authorization.kcp.dev: access granted",
    "systemcrd.authorization.kcp.dev/decision": "Allowed",
    "systemcrd.authorization.kcp.dev/reason": "maxpermissionpolicy.authorization.kcp.dev: access granted",
    "tenancy.kcp.dev/workspace": "root",
    "toplevel.authorization.kcp.dev/decision": "Allowed",
    "toplevel.authorization.kcp.dev/reason": "content.authorization.kcp.dev: access granted"
  }
```

This tells you "access granted", because bootstrap policy allowed access due to a cluster role binding. All other authorizer derive their decision based on delegation.

For the example above:
- top level authorizer allowed because content authorizer allowed.
- content authorizer allowed, because system CRD authorizer allowed
- system CRD authorizer allowed because max permission policy authorizer allowed
- max permission policy authorizer allowed because bootstrap policy authorizer allowed
- bootstrap policy allowed because of cluster role binding

deny case:
```
  "annotations": {
    "authorization.k8s.io/decision": "forbid",
    "authorization.k8s.io/reason": "access denied",
    "bootstrap.authorization.kcp.dev/decision": "NoOpinion",
    "bootstrap.authorization.kcp.dev/reason": "bootstrap policy reason: ",
    "content.authorization.kcp.dev/decision": "NoOpinion",
    "content.authorization.kcp.dev/reason": "systemcrd.authorization.kcp.dev: access denied",
    "local.authorization.kcp.dev/decision": "NoOpinion",
    "local.authorization.kcp.dev/reason": "cluster \"root:e2e-org-j6b4v:workspace2\" or bootstrap policy reason: ",
    "maxpermissionpolicy.authorization.kcp.dev/decision": "NoOpinion",
    "maxpermissionpolicy.authorization.kcp.dev/reason": "bootstrap.authorization.kcp.dev: access denied\nlocal.authorization.kcp.dev: access denied",
    "systemcrd.authorization.kcp.dev/decision": "NoOpinion",
    "systemcrd.authorization.kcp.dev/reason": "maxpermissionpolicy.authorization.kcp.dev: access denied",
    "tenancy.kcp.dev/workspace": "root:e2e-org-j6b4v:workspace2",
    "toplevel.authorization.kcp.dev/decision": "NoOpinion",
    "toplevel.authorization.kcp.dev/reason": "content.authorization.kcp.dev: access denied"
  }
```

The reasons can be derived as in the happy case. The "ugly" `bootstrap policy reason: ` is because the k8s RBAC authorizers return empty string here.

pkg/authorization/audit_logger.go
s-urbaniak: done

pkg/authorization/decorator.go
```diff
@@ -0,0 +1,128 @@
...
+	d.target = authorizer.AuthorizerFunc(func(ctx context.Context, attr authorizer.Attributes) (authorizer.Decision, string, error) {
+		dec, reason, err := target.Authorize(ctx, attr)
+
+		switch dec {
+		case authorizer.DecisionAllow:
+			reason = "access granted"
+		case authorizer.DecisionDeny:
+			reason = "access denied"
+		case authorizer.DecisionNoOpinion:
+			reason = "access denied"
```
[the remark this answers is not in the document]
s-urbaniak: this is a user facing anonymized reason message. no opinion implies no access in upstream, i'd leave as is.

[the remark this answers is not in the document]
s-urbaniak: or, reversely "why does this authorizer have no opinion?", because "access is denied".

pkg/authorization/maximal_permission_policy_authorizer.go
```diff
@@ -79,55 +74,34 @@ func NewMaximalPermissionPolicyAuthorizer(kubeInformers kcpkubernetesinformers.S
...
-		)
 		return authorizer.DecisionNoOpinion, MaximalPermissionPolicyAccessNotPermittedReason, err
 	}
 
 	bindingLogicalCluster, bound, err := a.getAPIBindingReferenceForAttributes(attr, lcluster)
 	if err != nil {
-		kaudit.AddAuditAnnotations(
-			ctx,
-			MaximalPermissionPolicyAuditDecision, DecisionNoOpinion,
-			MaximalPermissionPolicyAuditReason, fmt.Sprintf("error getting API binding reference: %v", err),
```
ncdc, line 110: Not sure if this matters, but it looks like we're losing small bits of info (e.g.: error getting API binding reference) with these changes. I think it's ok though?

pkg/server/options/authorization.go
```diff
@@ -101,24 +101,25 @@ func (s *Authorization) ApplyTo(config *genericapiserver.Config, informer kcpkub
...
-	bootstrapAuth, bootstrapRules := authorization.NewBootstrapPolicyAuthorizer(informer)
-	localAuth, localResolver := authorization.NewLocalAuthorizer(informer)
-	apiBindingAuth, err := authorization.NewMaximalPermissionPolicyAuthorizer(informer, kcpinformer,
-		union.New(bootstrapAuth, localAuth),
-	)
-	if err != nil {
-		return err
-	}
+	bootstrapAuth, bootstrapRules := authz.NewBootstrapPolicyAuthorizer(informer)
+	bootstrapAuth = authz.NewDecorator(bootstrapAuth, "bootstrap.authorization.kcp.dev").AddAuditLogging().AddAnonymization().AddReasonAnnotation()
```
ncdc, line 105: TODO: refactor so this is impossible to forget to do, and easier (I see we're also doing it in the virtual apiexport builder)

pkg/authorization/decorator.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		kaudit.AddAuditAnnotations(
+			ctx,
+			d.key+"/"+auditDecision, decisionString(dec),
+			d.key+"/"+auditReason, auditReasonMsg,
+		)
+
+		if dec != authorizer.DecisionAllow {
+			// Note: this deviates from upstream which doesn't log audit reasons.
```
stevekuznetsov, line 72: All of it goes into the audit log, no reason to log here IMHO

pkg/server/options/authorization.go
stevekuznetsov: Simple approach here would be to make the types coming out of `*.New*Authorizer()` useless without decoration

stevekuznetsov: Probably worth it to do in this PR if it's not a lot of effort

pkg/authorization/decorator.go
s-urbaniak: that was requested by @sttts, let him chime in here.

pkg/authorization/maximal_permission_policy_authorizer.go
s-urbaniak: In case of error, this information will definitely land in the audit logs before anonymization, we're not losing it 👍 I could `fmt.Errorf("error getting API binding reference: %w", err)` below.

pkg/authorization/audit_logger.go
s-urbaniak: i did, V(4) is in https://github.com/kcp-dev/kcp/pull/2442/files#diff-bce2789a10daa6a0f8a8fadb01ddbffa4a2db46d9cc62061ba01632822b941c6R75

pkg/server/options/authorization.go
s-urbaniak: We can also leave this in the constructor method. I moved it here because this is more plumbing. I am fine with either decision.

@stevekuznetsov good idea as well 👍 The returned type must not implement `Authorizer` though, maybe just a method name change in each authorizer, i.e. `UnsafeAuthorize(ctx, attr)` ?.

pkg/authorization/maximal_permission_policy_authorizer.go
ncdc: Right, I am referring to some of the `fmt.Sprintf` content being lost in this PR

pkg/server/options/authorization.go
s-urbaniak: unsafe doesn't sound good though 😆

pkg/authorization/maximal_permission_policy_authorizer.go
s-urbaniak: let me actually do that before we merge.

pkg/server/options/authorization.go
ncdc: Would we ever want to construct a FooAuthorizer that has a different decoration prefix, doesn't have audit logging, doesn't have anon, doesn't have reason annotation?

pkg/authorization/maximal_permission_policy_authorizer.go
ncdc: Thanks (and for clarity, this isn't the only place in the PR where this happened)

pkg/server/options/authorization.go
s-urbaniak: @ncdc probably not 🤔 that would speak for putting these things back into the constructor indeed.

s-urbaniak: @ncdc well, yes, the very last authorizer in the delegate chain doesn't need reason annotation.

pkg/authorization/decorator.go
sttts: this is for V(4). We want to see these as much as we want to see http logs IMO. Debugging/Developing without this is painful.

pkg/authorization/delegating_authorizer.go
s-urbaniak: done

pkg/authorization/maximal_permission_policy_authorizer.go
s-urbaniak: done

pkg/authorization/decorator.go
stevekuznetsov: Upstream chooses not to for security reasons? No strong opinion here - if httplog spits out the 403 + audit annotation on V(4) I'm happy.

pkg/server/options/authorization.go
stevekuznetsov: @sttts you commit to buying us all a beer on every instance of a new authorizer not using these filters?

sttts: For the record: on next KubeCon where we both meet, @stevekuznetsov will get a beer per authorizer until then that does not use the filters.

PR #2446, opened 2022-12-02
units for workspace_reconcile_scheduling

p0lyn0mial:
## Related issue(s)

Fixes #

### DIFF
--- a/pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling_test.go
+++ b/pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling_test.go
@@ -18,218 +18,231 @@ package workspace
 
 import (
 	"context"
+	"crypto/sha256"
+	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/kcp-dev/logicalcluster/v2"
+	"github.com/martinlindhe/base36"
 
+	kcpcache "github.com/kcp-dev/apimachinery/pkg/cache"
+	kcpfakekubeclient "github.com/kcp-dev/client-go/kubernetes/fake"
+	kcpclientgotesting "github.com/kcp-dev/client-go/third_party/k8s.io/client-go/testing"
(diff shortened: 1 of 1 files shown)


p0lyn0mial:
I think I will add one more case for testing assignment of `WorkspaceAPIBindingsInitializer`

pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling_test.go
```diff
@@ -470,6 +470,7 @@ func wellKnownClusterWorkspaceTypes() []*tenancyv1alpha1.ClusterWorkspaceType {
 		Name: "universal",
 		Path: "root",
 	}
+	type1.Spec.Initializer = true
```
p0lyn0mial, line 473: @sttts  shouldn't this `Initializer` be resolved by `r.transitiveTypeResolver.Resolve` ?

sttts: The initializer name is derived from the CWT workspace and name afaik. It's not resolved.

p0lyn0mial: okay but it will be resolved only for the immediate CWT not for all CWTs in the hierarchy, is that okay?

this is how the resolver is wired atm
```
transitiveTypeResolver: clusterworkspacetypeexists.NewTransitiveTypeResolver(func(clusterName logicalcluster.Name, name string) (*tenancyv1alpha1.ClusterWorkspaceType, error) {
				return c.clusterWorkspaceTypeLister.Cluster(clusterName).Get(name)
			}),
```

p0lyn0mial: unless there is something wrong with the way the test prepares the data

p0lyn0mial: yes, it was a test issue, fixed, PTAL.

PR #2447, opened 2022-12-02
🐛 cliplugins/bind: fix incorrect error message

jmprusi:
## Summary

When binding a workspace to compute, if some of the apibindings is not ready, the `bind` cli will output an error like:

```
error: bind compute is not ready placement-23sk5wr9: not bound to apiexport 'root:compute:root:compute': Unable to bind APIs: naming conflict with a bound API kubernetes, spec.names.plural=pods is...
```

This error is confusing, as the workspace is duplicated, and doesn't specify the export name: 'root:compute:root:compute'

This PR changes it to include the name of the export, as in: 'root:compute:kubernetes'


jmprusi:
TestSyncerLifecycle: Null Spec in deployment

/retest

PR #2449, opened 2022-12-02
:seedling: Simplify the addition of new resources to the replication e2e test

fgiloux:
## Summary

The original idea was to look at using the dynamic client for code reuse between resources being added to the cache server. 
Experimenting with it I felt that creating and manipulating resources is more painful than with the typed client.
In this pull request I tried to address a few things:

- to have the needed changes localized in one place for supporting additional resources.
- to separate what is resource specific, mainly get, update, delete functions from the plumbing and tests.

The changes should also allow to more easily add tests once and apply them to all the resources. In my eyes it also makes the tests easier to follow but I reckon that others may feel differently.

## Related issue(s)

Addresses #https://github.com/kcp-dev/kcp/pull/2381#discussion_r1029443074

### DIFF
--- a/test/e2e/reconciler/cache/replication_test.go
+++ b/test/e2e/reconciler/cache/replication_test.go
@@ -18,7 +18,6 @@ package cache
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"testing"
 	"time"
@@ -28,7 +27,6 @@ import (
 	"github.com/kcp-dev/logicalcluster/v3"
 	"github.com/stretchr/testify/require"
 
-	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
/hold

How is this more straightforward than using the dynamic client? Am I missing something here

fgiloux:
> How is this more straightforward than using the dynamic client? Am I missing something here

when I looked at rewriting the [APIExport, APIResourceSchema generation](https://github.com/kcp-dev/kcp/blob/5916c3cd79ca30b3ed6e8a8983d10034f1a926e2/test/e2e/fixtures/apifixtures/sheriffs.go#L117-L154) with something like [here](https://github.com/kcp-dev/kcp/blob/5916c3cd79ca30b3ed6e8a8983d10034f1a926e2/test/e2e/fixtures/apifixtures/sheriffs.go#L192-L199) I missed the tab completion and checks being done in the IDE. 

A better alternative may be to write plain yaml and serialize it into unstructured objects. To be honest I am not expecting a big gain out of it.

stevekuznetsov:
Creating the normal types and just transforming them to unstructured to use one implementation using a dynamic client seems pretty straightforward, and doesn't require the switch/case/overhead in this PR.

fgiloux:
> doesn't require the switch/case/overhead in this PR.

Which switch/case?

stevekuznetsov:
Not sure what you mean - literally all of the code this PR touches is redundant, right? Whether we're doing a switch or having each type mock out its own get/put/delete methods ... if we just take a `runtime.Object`, transform to unstructured and use the dynamic client we don't need any of it, yes?

fgiloux:
> literally all of the code this PR touches is redundant, right? 

It would be right if the tests were only setting metadata but it is not the case and cannot be the case when fields are mandatory.

>  if we just take a runtime.Object, transform to unstructured and use the dynamic client we don't need any of it, yes?

The only things that would be impacted are the init functions. All the other functions are already resource type agnostic.
In regards to the the init functions "get" and "delete" would be common (they are currently one liner) but not the data for "create" and "update". The biggest change is that this data would be supplied through yaml strings instead of structures. I can do that but I still think that the gain is low.

fgiloux:
Closing in favor of #2578

PR #2452, opened 2022-12-05
✨ Refactor Syncer based on the enhanced ddsif, with controller manager and endpoints controller

davidfestal:
## Summary

PROOF PR of a refactoring of the Syncer to use common ddsif informer factories for all the Syncer + other controllers.
It also includes a controller manager, which allows starting / stopping GVR-specific controllers when new GVRs are supported in the list of Synced resources.

It is based on the work done on the ddsif in PR https://github.com/kcp-dev/kcp/pull/2440

## Related issue(s)

No related issue. But it is a pre-requisite to the merge of PRs like https://github.com/kcp-dev/kcp/pull/2338

### DIFF
--- a/pkg/informer/informer.go
+++ b/pkg/informer/informer.go
@@ -353,13 +353,13 @@ func (d *GenericDiscoveringDynamicSharedInformerFactory[Informer, Lister, Generi
 	return inf
 }
 
-// Listers returns a map of per-resource-type listers for all types that are
+// Informers returns a map of per-resource-type generic informers for all types that are
 // known by this informer factory, and that are synced.
 //
 // If any informers aren't synced, their GVRs are returned so that they can be
 // checked and processed later.
-func (d *GenericDiscoveringDynamicSharedInformerFactory[Informer, Lister, GenericInformer]) Listers() (listers map[schema.GroupVersionResource]Lister, notSynced []schema.GroupVersionResource) {
-	listers = map[schema.GroupVersionResource]Lister{}
+func (d *GenericDiscoveringDynamicSharedInformerFactory[Informer, Lister, GenericInformer]) Informers() (informers map[schema.GroupVersionResource]GenericInformer, notSynced []schema.GroupVersionResource) {
+	informers = map[schema.GroupVersionResource]GenericInformer{}
 
 	d.informersLock.RLock()
 	defer d.informersLock.RUnlock()
@@ -372,38 +372,10 @@ func (d *GenericDiscoveringDynamicSharedInformerFactory[Informer, Lister, Generi
 			continue
 		}
 
-		listers[gvr] = informer.Lister()
+		informers[gvr] = informer
 	}
 
-	return listers, notSynced
-}
-
-// Lister returns a lister for the given resource-type, whether it is
-// known by this informer factory, and whether it is synced.
-func (d *GenericDiscoveringDynamicSharedInformerFactory[Informer, Lister, GenericInformer]) Lister(gvr schema.GroupVersionResource) (lister Lister, known, synced bool) {
-	d.informersLock.RLock()
-	defer d.informersLock.RUnlock()
-
-	informer, ok := d.informers[gvr]
-	if !ok {
-		return lister, false, false
-	}
-
-	return informer.Lister(), true, informer.Informer().HasSynced()
-}
-
-// Informer returns an informer for the given resource-type if it exists, whether it is
-// known by this informer factory, and whether it is synced.
-func (d *GenericDiscoveringDynamicSharedInformerFactory[Informer, Lister, GenericInformer]) Informer(gvr schema.GroupVersionResource) (informer Informer, known, synced bool) {
-	d.informersLock.RLock()
-	defer d.informersLock.RUnlock()
-
-	genericInformer, ok := d.informers[gvr]
-	if !ok {
-		return informer, false, false
-	}
-
-	return genericInformer.Informer(), true, genericInformer.Informer().HasSynced()
+	return informers, notSynced
 }
 
 // GVREventHandler is an event handler that includes the GroupVersionResource
--- a/pkg/syncer/controllermanager/controllermanager.go
+++ b/pkg/syncer/controllermanager/controllermanager.go
@@ -1,0 +1,188 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllermanager
+
+import (
+	"context"
+	"time"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+
+	"github.com/kcp-dev/kcp/pkg/logging"
+)
+
+const (
+	ControllerNamePrefix = "syncer-controller-manager-"
+)
+
+type InformerSource struct {
+	Subscribe func(id string) <-chan struct{}
+	Informer  func(gvr schema.GroupVersionResource) (informer cache.SharedIndexInformer, known, synced bool)
+}
+
+type Controller interface {
+	Start(ctx context.Context, numThreads int)
+}
+
+type ControllerDefintion struct {
+	RequiredGVRs []schema.GroupVersionResource
+	NumThreads   int
+	Create       func(syncedInformers map[schema.GroupVersionResource]cache.SharedIndexInformer) (Controller, error)
+}
+
+func NewControllerManager(ctx context.Context, suffix string, informerSource InformerSource, controllers map[string]ControllerDefintion) *ControllerManager {
+	controllerManager := ControllerManager{
+		name:                  ControllerNamePrefix + suffix,
+		queue:                 workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), ControllerNamePrefix+suffix),
+		informerSource:        informerSource,
+		controllerDefinitions: controllers,
+		startedControllers:    map[string]context.CancelFunc{},
+	}
+
+	apisChanged := informerSource.Subscribe(controllerManager.name)
+
+	logger := klog.FromContext(ctx)
+
+	go func() {
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-apisChanged:
+				logger.V(4).Info("got API change notification")
+				controllerManager.queue.Add("resync") // this queue only ever has one key in it, as long as it's constant we are OK
+			}
+		}
+	}()
+
+	return &controllerManager
+}
+
+type ControllerManager struct {
+	name                  string
+	queue                 workqueue.RateLimitingInterface
+	informerSource        InformerSource
+	controllerDefinitions map[string]ControllerDefintion
+	startedControllers    map[string]context.CancelFunc
+}
+
+// Start starts the controller, which stops when ctx.Done() is closed.
+func (c *ControllerManager) Start(ctx context.Context) {
+	defer utilruntime.HandleCrash()
+	defer c.queue.ShutDown()
+
+	logger := logging.WithReconciler(klog.FromContext(ctx), c.name)
+	logger.Info("Starting controller manager")
+	defer logger.Info("Shutting down controller manager")
+
+	go wait.UntilWithContext(ctx, c.startWorker, time.Second)
+	<-ctx.Done()
+}
+
+func (c *ControllerManager) startWorker(ctx context.Context) {
+	for c.processNextWorkItem(ctx) {
+	}
+}
+
+func (c *ControllerManager) processNextWorkItem(ctx context.Context) bool {
+	key, quit := c.queue.Get()
+	if quit {
+		return false
+	}
+	defer c.queue.Done(key)
+
+	c.UpdateControllers(ctx)
+	c.queue.Forget(key)
+	return true
+}
+
+func (c *ControllerManager) UpdateControllers(ctx context.Context) {
+	logger := klog.FromContext(ctx)
+	controllersToStart := map[string]map[schema.GroupVersionResource]cache.SharedIndexInformer{}
+controllerLoop:
+	for controllerName, controllerDefinition := range c.controllerDefinitions {
+		requiredGVRs := controllerDefinition.RequiredGVRs
+		informers := make(map[schema.GroupVersionResource]cache.SharedIndexInformer, len(requiredGVRs))
+		for _, gvr := range requiredGVRs {
+			if informer, known, synced := c.informerSource.Informer(gvr); !known {
+				continue controllerLoop
+			} else if !synced {
+				logger.V(2).Info("waiting for the informer to be synced before starting controller", "gvr", gvr, "controller", controllerName)
+				c.queue.AddAfter("resync", time.Second)
+				continue controllerLoop
+			} else {
+				informers[gvr] = informer
+			}
+		}
+		controllersToStart[controllerName] = informers
+	}
+
+	// Create and start missing controllers that have their required GVRs synced
+	newlyStartedControllers := map[string]context.CancelFunc{}
+	for controllerName, informers := range controllersToStart {
+		if _, ok := c.startedControllers[controllerName]; ok {
+			// The controller is already started
+			continue
+		}
+		controllerDefinition, ok := c.controllerDefinitions[controllerName]
+		if !ok {
+			logger.V(2).Info("cannot find controller definition", "controller", controllerName)
+			continue
+		}
+
+		// Create the controller
+		controller, err := controllerDefinition.Create(informers)
+		if err != nil {
+			logger.Error(err, "error creating controller", "controller", controllerName)
+			continue
+		}
+
+		for _, informer := range informers {
+			if err := informer.GetStore().Resync(); err != nil {
+				logger.Error(err, "error resyncing informer controller", "controller", controllerName)
+				continue
+			}
+		}
+
+		// Start the controller
+		controllerContext, cancelFunc := context.WithCancel(ctx)
+		go controller.Start(controllerContext, controllerDefinition.NumThreads)
+		newlyStartedControllers[c.name] = cancelFunc
+	}
+
+	// Remove obsolete controllers that don't have their required GVRs anymore
+	for controllerName, cancelFunc := range c.startedControllers {
+		if _, ok := controllersToStart[controllerName]; ok {
+			// The controller is still expected => don't remove it
+			continue
+		}
+		// The controller should not be running
+		// Stop it and remove it from the list of started controllers
+		cancelFunc()
+		delete(c.startedControllers, controllerName)
+	}
+
+	// Add missing controllers that were created and started above
+	for controllerName, cancelFunc := range newlyStartedControllers {
+		c.startedControllers[controllerName] = cancelFunc
+	}
+}
--- a/pkg/syncer/controllermanager/controllermanager.go
+++ b/pkg/syncer/controllermanager/controllermanager.go
@@ -28,34 +28,50 @@ import (
 	"k8s.io/klog/v2"
 
 	"github.com/kcp-dev/kcp/pkg/logging"
+	"github.com/kcp-dev/kcp/pkg/syncer/shared"
 )
 
 const (
 	ControllerNamePrefix = "syncer-controller-manager-"
 )
 
+// InformerSource is a dynamic source of informers per GVR,
+// which notifies when informers are added or removed for some GVR.
+// It is implemented by the DynamicSharedInformerFactory (in fact by
+// both the scoped or cluster-aware variants).
 type InformerSource struct {
+	// Subscribe registers for informer change notifications, returning a channel to which change notifications are sent.
+	// The id argument is the identifier of the subscriber, since there might be several subscribers subscribing
+	// to receive events from this InformerSource.
 	Subscribe func(id string) <-chan struct{}
-	Informer  func(gvr schema.GroupVersionResource) (informer cache.SharedIndexInformer, known, synced bool)
+
+	// Informers returns a map of per-resource-type SharedIndexInformers for all types that are
+	// known by this informer source, and that are synced.
+	//
+	// It also returns the list of informers that are known by this informer source, but sill not synced.
+	Informers func() (informers map[schema.GroupVersionResource]cache.SharedIndexInformer, notSynced []schema.GroupVersionResource)
 }
 
-type Controller interface {
-	Start(ctx context.Context, numThreads int)
-}
-
-type ControllerDefintion struct {
+// ManagedController defines a controller that should be managed by a ControllerManager,
+// to be started when the required GVRs are supported, and stopped when the required GVRs
+// are not supported anymore.
+type ManagedController struct {
 	RequiredGVRs []schema.GroupVersionResource
-	NumThreads   int
-	Create       func(syncedInformers map[schema.GroupVersionResource]cache.SharedIndexInformer) (Controller, error)
+	Create       CreateControllerFunc
 }
 
-func NewControllerManager(ctx context.Context, suffix string, informerSource InformerSource, controllers map[string]ControllerDefintion) *ControllerManager {
+type StartControllerFunc func(ctx context.Context)
+type CreateControllerFunc func(ctx context.Context) (StartControllerFunc, error)
+
+// NewControllerManager creates a new ControllerManager which will manage (create/start/stop) GVR-specific controllers according to informers
+// available in the provided InformerSource.
+func NewControllerManager(ctx context.Context, suffix string, informerSource InformerSource, controllers map[string]ManagedController) *ControllerManager {
 	controllerManager := ControllerManager{
-		name:                  ControllerNamePrefix + suffix,
-		queue:                 workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), ControllerNamePrefix+suffix),
-		informerSource:        informerSource,
-		controllerDefinitions: controllers,
-		startedControllers:    map[string]context.CancelFunc{},
+		name:               ControllerNamePrefix + suffix,
+		queue:              workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), ControllerNamePrefix+suffix),
+		informerSource:     informerSource,
+		managedControllers: controllers,
+		startedControllers: map[string]context.CancelFunc{},
 	}
 
 	apisChanged := informerSource.Subscribe(controllerManager.name)
@@ -77,12 +93,19 @@ func NewControllerManager(ctx context.Context, suffix string, informerSource Inf
 	return &controllerManager
 }
 
+// ControllerManager is a component that manages (create/start/stop) GVR-specific controllers according to available GVRs.
+// It reacts to the changes of supported GVRs in a DiscoveringDynamicSharedInformerFactory
+// (the GVRs for which an informer has been automatically created, started and synced),
+// and starts / stops registered GVRs-specific controllers according to the GVRs they depend on.
+//
+// For example this allows starting PVC / PV controllers only when PVC / PV resources are exposed by the Syncer and UpSyncer
+// virtual workspaces, and Informers for them have been started and synced by the corresponding ddsif.
 type ControllerManager struct {
-	name                  string
-	queue                 workqueue.RateLimitingInterface
-	informerSource        InformerSource
-	controllerDefinitions map[string]ControllerDefintion
-	startedControllers    map[string]context.CancelFunc
+	name               string
+	queue              workqueue.RateLimitingInterface
+	informerSource     InformerSource
+	managedControllers map[string]ManagedController
+	startedControllers map[string]context.CancelFunc
 }
 
 // Start starts the controller, which stops when ctx.Done() is closed.
@@ -110,63 +133,35 @@ func (c *ControllerManager) processNextWorkItem(ctx context.Context) bool {
 	}
 	defer c.queue.Done(key)
 
-	c.UpdateControllers(ctx)
+	c.process(ctx)
 	c.queue.Forget(key)
 	return true
 }
 
-func (c *ControllerManager) UpdateControllers(ctx context.Context) {
+func (c *ControllerManager) process(ctx context.Context) {
 	logger := klog.FromContext(ctx)
-	controllersToStart := map[string]map[schema.GroupVersionResource]cache.SharedIndexInformer{}
+	controllersToStart := map[string]CreateControllerFunc{}
+	syncedInformers, notSynced := c.informerSource.Informers()
 controllerLoop:
-	for controllerName, controllerDefinition := range c.controllerDefinitions {
-		requiredGVRs := controllerDefinition.RequiredGVRs
-		informers := make(map[schema.GroupVersionResource]cache.SharedIndexInformer, len(requiredGVRs))
+	for controllerName, managedController := range c.managedControllers {
+		requiredGVRs := managedController.RequiredGVRs
 		for _, gvr := range requiredGVRs {
-			if informer, known, synced := c.informerSource.Informer(gvr); !known {
+			informer := syncedInformers[gvr]
+			if informer == nil {
+				if shared.ContainsGVR(notSynced, gvr) {
+					logger.V(2).Info("waiting for the informer to be synced before starting controller", "gvr", gvr, "controller", controllerName)
+					c.queue.AddAfter("resync", time.Second)
+					continue controllerLoop
+				}
+				// The informer doesn't even exist for this GVR.
+				// Let's ignore this controller for now: one of the required GVRs has no informer started
+				// (because it has not been found on the SyncTarget in the supported resources to sync).
+				// If this required GVR is supported later on, the updateControllers() method will be called
+				// again after an API change notification comes through the informerSource.
 				continue controllerLoop
-			} else if !synced {
-				logger.V(2).Info("waiting for the informer to be synced before starting controller", "gvr", gvr, "controller", controllerName)
-				c.queue.AddAfter("resync", time.Second)
-				continue controllerLoop
-			} else {
-				informers[gvr] = informer
 			}
 		}
-		controllersToStart[controllerName] = informers
-	}
-
-	// Create and start missing controllers that have their required GVRs synced
-	newlyStartedControllers := map[string]context.CancelFunc{}
-	for controllerName, informers := range controllersToStart {
-		if _, ok := c.startedControllers[controllerName]; ok {
-			// The controller is already started
-			continue
-		}
-		controllerDefinition, ok := c.controllerDefinitions[controllerName]
-		if !ok {
-			logger.V(2).Info("cannot find controller definition", "controller", controllerName)
-			continue
-		}
-
-		// Create the controller
-		controller, err := controllerDefinition.Create(informers)
-		if err != nil {
-			logger.Error(err, "error creating controller", "controller", controllerName)
-			continue
-		}
-
-		for _, informer := range informers {
-			if err := informer.GetStore().Resync(); err != nil {
-				logger.Error(err, "error resyncing informer controller", "controller", controllerName)
-				continue
-			}
-		}
-
-		// Start the controller
-		controllerContext, cancelFunc := context.WithCancel(ctx)
-		go controller.Start(controllerContext, controllerDefinition.NumThreads)
-		newlyStartedControllers[c.name] = cancelFunc
+		controllersToStart[controllerName] = managedController.Create
 	}
 
 	// Remove obsolete controllers that don't have their required GVRs anymore
@@ -181,8 +176,23 @@ controllerLoop:
 		delete(c.startedControllers, controllerName)
 	}
 
-	// Add missing controllers that were created and started above
-	for controllerName, cancelFunc := range newlyStartedControllers {
+	// Create and start missing controllers that have their required GVRs synced
+	for controllerName, create := range controllersToStart {
+		if _, ok := c.startedControllers[controllerName]; ok {
+			// The controller is already started
+			continue
+		}
+
+		// Create the controller
+		start, err := create(ctx)
+		if err != nil {
+			logger.Error(err, "failed creating controller", "controller", controllerName)
+			continue
+		}
+
+		// Start the controller
+		controllerContext, cancelFunc := context.WithCancel(ctx)
+		go start(controllerContext)
 		c.startedControllers[controllerName] = cancelFunc
 	}
 }
--- a/pkg/syncer/endpoints/endpoint_downstream_controller.go
+++ b/pkg/syncer/endpoints/endpoint_downstream_controller.go
@@ -1,0 +1,160 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package endpoints
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/go-logr/logr"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+
+	ddsif "github.com/kcp-dev/kcp/pkg/informer"
+	"github.com/kcp-dev/kcp/pkg/logging"
+)
+
+const (
+	ControllerName = "syncer-endpoint-controller"
+)
+
+func NewEndpointController(
+	syncerLogger logr.Logger,
+	downstreamClient dynamic.Interface,
+	ddsifForDownstream *ddsif.GenericDiscoveringDynamicSharedInformerFactory[cache.SharedIndexInformer, cache.GenericLister, informers.GenericInformer],
+	syncedInformers map[schema.GroupVersionResource]cache.SharedIndexInformer,
+) (*EndpointController, error) {
+	endpointsGVR := corev1.SchemeGroupVersion.WithResource("endpoints")
+
+	c := &EndpointController{
+		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), ControllerName),
+	}
+
+	logger := logging.WithReconciler(syncerLogger, ControllerName)
+
+	endpointsInformer, ok := syncedInformers[endpointsGVR]
+	if !ok {
+		return nil, errors.New("endpoints informer should be available")
+	}
+
+	endpointsInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			c.AddToQueue(obj, logger)
+		},
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			c.AddToQueue(newObj, logger)
+		},
+		DeleteFunc: func(obj interface{}) {
+			c.AddToQueue(obj, logger)
+		},
+	})
+
+	return c, nil
+}
+
+type EndpointController struct {
+	queue workqueue.RateLimitingInterface
+}
+
+func (c *EndpointController) AddToQueue(obj interface{}, logger logr.Logger) {
+	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+	if err != nil {
+		utilruntime.HandleError(err)
+		return
+	}
+
+	logging.WithQueueKey(logger, key).V(2).Info("queueing", "key", key)
+	c.queue.Add(key)
+}
+
+// Start starts N worker processes processing work items.
+func (c *EndpointController) Start(ctx context.Context, numThreads int) {
+	defer utilruntime.HandleCrash()
+	defer c.queue.ShutDown()
+
+	logger := logging.WithReconciler(klog.FromContext(ctx), ControllerName)
+	ctx = klog.NewContext(ctx, logger)
+	logger.Info("Starting syncer workers")
+	defer logger.Info("Stopping syncer workers")
+
+	for i := 0; i < numThreads; i++ {
+		go wait.UntilWithContext(ctx, c.startWorker, time.Second)
+	}
+
+	<-ctx.Done()
+}
+
+// startWorker processes work items until stopCh is closed.
+func (c *EndpointController) startWorker(ctx context.Context) {
+	for c.processNextWorkItem(ctx) {
+	}
+}
+
+func (c *EndpointController) processNextWorkItem(ctx context.Context) bool {
+	// Wait until there is a new item in the working queue
+	key, quit := c.queue.Get()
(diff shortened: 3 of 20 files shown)


ncdc:
/hold

pending workspace refactor

pkg/syncer/shared/helpers.go
```diff
@@ -71,3 +71,12 @@ func GetDNSID(clusterName logicalcluster.Name, syncTargetUID types.UID, syncTarg
 
 	return fmt.Sprintf("kcp-dns-%s-%s-%s", syncTargetName, uid36hash[:8], workspace36hash[:8])
 }
+
+func Contains[T comparable](elems []T, v T) bool {
```
stevekuznetsov, line 75: FWIW these kinds of methods are more than likely to make the code significantly less performant [xref](https://planetscale.com/blog/generics-can-make-your-go-code-slower) - the non-generic `for` loop might just be a sensible thing to keep around.

pkg/syncer/spec/spec_controller.go
```diff
@@ -39,39 +39,40 @@ import (
...
 	. "github.com/kcp-dev/kcp/tmc/pkg/logging"
 )
 
 const (
-	controllerName              = "kcp-workload-syncer-spec"
-	byNamespaceLocatorIndexName = "syncer-spec-ByNamespaceLocator"
+	controllerName = "kcp-workload-syncer-spec"
 )
 
+var namespaceGVR schema.GroupVersionResource = schema.GroupVersionResource{Group: "", Version: "v1", Resource: "namespaces"}
```
stevekuznetsov, line 61:
```suggestion
var namespaceGVR schema.GroupVersionResource = corev1.GroupVersion.WithResource("namespaces")
```

pkg/syncer/shared/helpers.go
davidfestal: Fixed. In fact this method was not used anymore. I dropped it.

davidfestal: Thanks for the xref BTW

pkg/syncer/spec/spec_controller.go
davidfestal: Fixed

pkg/syncer/controllermanager/controllermanager.go
```diff
@@ -0,0 +1,188 @@
...
+				logger.V(4).Info("got API change notification")
+				controllerManager.queue.Add("resync") // this queue only ever has one key in it, as long as it's constant we are OK
+			}
+		}
+	}()
+
+	return &controllerManager
+}
+
+type ControllerManager struct {
```
sttts, line 80: what is a controller manager?

pkg/syncer/endpoints/endpoint_downstream_controller.go
```diff
@@ -0,0 +1,160 @@
...
+		},
+		DeleteFunc: func(obj interface{}) {
+			c.AddToQueue(obj, logger)
+		},
+	})
+
+	return c, nil
+}
+
+type EndpointController struct {
```
sttts, line 79: What is an endpoint controller?

pkg/syncer/endpoints/endpoint_downstream_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	queue workqueue.RateLimitingInterface
+}
+
+func (c *EndpointController) AddToQueue(obj interface{}, logger logr.Logger) {
```
sttts, line 83: please follow kcp conventions for naming this

pkg/syncer/endpoints/endpoint_downstream_controller.go
```diff
@@ -0,0 +1,160 @@
...
+
+const (
+	ControllerName = "syncer-endpoint-controller"
+)
+
+func NewEndpointController(
+	syncerLogger logr.Logger,
+	downstreamClient dynamic.Interface,
+	ddsifForDownstream *ddsif.GenericDiscoveringDynamicSharedInformerFactory[cache.SharedIndexInformer, cache.GenericLister, informers.GenericInformer],
+	syncedInformers map[schema.GroupVersionResource]cache.SharedIndexInformer,
```
sttts, line 49: why pass in a map if you can pass in `syncedInformers[endpointsGVR]` directly? The map it outer plumbing. Irrelevant here.

pkg/syncer/endpoints/endpoint_downstream_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	ddsif "github.com/kcp-dev/kcp/pkg/informer"
+	"github.com/kcp-dev/kcp/pkg/logging"
+)
+
+const (
+	ControllerName = "syncer-endpoint-controller"
+)
+
+func NewEndpointController(
+	syncerLogger logr.Logger,
```
sttts, line 46: We don't follow this pattern elsewhere, but create the logger in `processNextWorkItem`.

pkg/syncer/namespace/namespace_downstream_controller.go
```diff
@@ -100,29 +99,45 @@ func NewDownstreamController(
 			return downstreamClient.Resource(namespaceGVR).Delete(ctx, namespace, metav1.DeleteOptions{})
 		},
 		upstreamNamespaceExists: func(clusterName logicalcluster.Name, upstreamNamespaceName string) (bool, error) {
-			_, err := upstreamInformers.ForResource(namespaceGVR).Lister().ByCluster(clusterName).Get(upstreamNamespaceName)
+			lister, known, synced := ddsifForUpstreamSyncer.Lister(namespaceGVR)
+			if !known || !synced {
+				return false, errors.New("informer should be up and synced for namespaces in the upstream syncer informer factory")
```
sttts, line 104: what does "should" mean here?

pkg/syncer/namespace/namespace_downstream_controller.go
```diff
@@ -100,29 +99,45 @@ func NewDownstreamController(
...
 		},
 		listDownstreamNamespaces: func() ([]runtime.Object, error) {
-			return downstreamInformers.ForResource(namespaceGVR).Lister().List(labels.Everything())
+			lister, known, synced := ddsifForDownstream.Lister(namespaceGVR)
+			if !known || !synced {
+				return nil, errors.New("informer should be up and synced for namespaces in the downstream informer factory")
+			}
+			return lister.List(labels.Everything())
 		},
 		isDowntreamNamespaceEmpty: func(ctx context.Context, namespace string) (bool, error) {
```
sttts, line 129: unrelated to this PR: this function sounds dangerous. What is it used for?

pkg/syncer/shared/namespace.go
```diff
@@ -105,3 +106,7 @@ func PhysicalClusterNamespaceName(l NamespaceLocator) (string, error) {
 	// keep the namespaces short enough.
 	return fmt.Sprintf("kcp-%s", base36hash[:12]), nil
 }
+
+func IsNamespace(gvr schema.GroupVersionResource) bool {
+	return gvr.Resource == "namespaces" && gvr.Group == "" && gvr.Version == "v1"
+}
```
sttts, line 112: what's the value of this function? just compare with `corev1.Resource("namespaces")`.

pkg/syncer/controllermanager/controllermanager.go
```diff
@@ -0,0 +1,188 @@
...
+
+	"github.com/kcp-dev/kcp/pkg/logging"
+)
+
+const (
+	ControllerNamePrefix = "syncer-controller-manager-"
+)
+
+type InformerSource struct {
+	Subscribe func(id string) <-chan struct{}
```
sttts, line 38: what is id?

pkg/syncer/namespace/namespace_downstream_controller.go
davidfestal: It is used as a condition to allow the deletion of a downstream namespace. Downstream namespace deletion should only occur (after a delay) when no more synced resources exist inside it.

pkg/syncer/resourcesync/controller.go
```diff
@@ -63,47 +63,35 @@ const (
...
+	cachedDiscovery discovery.CachedDiscoveryInterface
+
+	syncTargetUID               types.UID
+	syncTargetLister            workloadv1alpha1listers.SyncTargetLister
+	synctargetInformerCacheSync cache.InformerSynced
+	kcpClient                   clientset.Interface
+
+	gvrs map[schema.GroupVersionResource]informer.GVRPartialMetadata
+
+	mutex sync.RWMutex
```
sttts, line 85: mutex for what?

pkg/syncer/resourcesync/controller.go
```diff
@@ -365,79 +330,163 @@ func (c *Controller) checkSSAR(ctx context.Context, gvr schema.GroupVersionResou
 	return sar.Status.Allowed, nil
 }
 
-// stopUnusedSyncerInformers stop syncers for gvrs not in requiredGVRs.
-func (c *Controller) stopUnusedSyncerInformers(ctx context.Context, requiredGVRs map[schema.GroupVersionResource]bool) {
+// removeUnusedGVRs stop syncers for gvrs not in requiredGVRs.
+func (c *Controller) removeUnusedGVRs(ctx context.Context, requiredGVRs map[schema.GroupVersionResource]bool) error {
```
sttts, line 334: stopUnusedSyncers

pkg/syncer/spec/spec_process.go
```diff
@@ -265,7 +265,12 @@ func (c *Controller) ensureDownstreamNamespaceExists(ctx context.Context, downst
 	}
 
 	// Check if the namespace already exists, if not create it.
```
sttts, line 267: misplaced comment

pkg/syncer/status/status_controller.go
```diff
@@ -60,15 +65,37 @@ type Controller struct {
...
 	c := &Controller{
 		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), controllerName),
 
-		upstreamClient:            upstreamClient,
-		downstreamClient:          downstreamClient,
-		downstreamNamespaceLister: downstreamInformers.ForResource(schema.GroupVersionResource{Version: "v1", Resource: "namespaces"}).Lister(),
+		upstreamClient:   upstreamClient,
+		downstreamClient: downstreamClient,
+
+		getDownstreamLister: func(gvr schema.GroupVersionResource) (cache.GenericLister, error) {
```
sttts, line 78: why these wrappers? Just use the ddsifs?

pkg/syncer/status/status_process.go
```diff
@@ -137,20 +157,28 @@ func (c *Controller) process(ctx context.Context, gvr schema.GroupVersionResourc
...
 
+	upstreamLister, err := c.getUpstreamLister(gvr)
+	if err != nil {
+		return err
+	}
+
 	if !resourceExists {
 		logger.Info("Downstream object does not exist. Removing finalizer on upstream object")
-		return shared.EnsureUpstreamFinalizerRemoved(ctx, gvr, syncerInformer.UpstreamInformer, c.upstreamClient, upstreamNamespace, c.syncTargetKey, upstreamClusterName, shared.GetUpstreamResourceName(gvr, downstreamName))
+		if err != nil {
```
sttts, line 167: ???

pkg/syncer/syncer.go
```diff
@@ -211,6 +237,39 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 		return err
 	}
 
+	ddsifForUpstreamSyncer, err := ddsif.NewDiscoveringDynamicSharedInformerFactory(upstreamSyncerClusterClient, nil, nil, resourceController, cache.Indexers{})
+	if err != nil {
+		return err
+	}
+
+	ddsifForUpstreamUpsyncer, err := ddsif.NewDiscoveringDynamicSharedInformerFactory(upstreamUpsyncerClusterClient, nil, nil,
```
sttts, line 245: am confused. Multiple ddsifs for upstream? Why?

pkg/syncer/syncer.go
```diff
@@ -211,6 +237,39 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
...
+		&FilteredGVRSource{
+			resourceController,
+			func(gvr schema.GroupVersionResource) bool {
+				if gvr.Resource == "persistentvolumes" && gvr.Group == "" {
```
sttts, line 249: `return gvr != corev1.Resource("persistentvolumes")`

pkg/syncer/controllermanager/controllermanager.go
```diff
@@ -0,0 +1,188 @@
...
+				controllerManager.queue.Add("resync") // this queue only ever has one key in it, as long as it's constant we are OK
+			}
+		}
+	}()
+
+	return &controllerManager
+}
+
+type ControllerManager struct {
+	name                  string
```
sttts, line 81: why does this object know its name?

pkg/syncer/controllermanager/controllermanager.go
```diff
@@ -0,0 +1,188 @@
...
+		return false
+	}
+	defer c.queue.Done(key)
+
+	c.UpdateControllers(ctx)
+	c.queue.Forget(key)
+	return true
+}
+
+func (c *ControllerManager) UpdateControllers(ctx context.Context) {
```
sttts, line 118: why public?

pkg/syncer/controllermanager/controllermanager.go
davidfestal: Still need to add GoDoc.
This is a component that manages GVR controllers.
It reacts to the changes of supported GVRs in a ddsif (GVRs for which an informer is created, started and synced), and starts / stops GVRs-specific controllers according to the GVRs they depend on.

pkg/syncer/controllermanager/controllermanager.go
```diff
@@ -0,0 +1,188 @@
...
+	key, quit := c.queue.Get()
+	if quit {
+		return false
+	}
+	defer c.queue.Done(key)
+
+	c.UpdateControllers(ctx)
+	c.queue.Forget(key)
+	return true
+}
```
sttts, line 116: is this a controller? Please follow the usual patterns and naming.

pkg/syncer/controllermanager/controllermanager.go
davidfestal: For example this allows starting PVC/PV controllers only when PVC / PV resources are exposed by the Syncer and UpSyncer virtual workspaces, and Informers for them have been started and Syncer by the corresponding ddsif.

pkg/syncer/controllermanager/controllermanager.go
```diff
@@ -0,0 +1,188 @@
...
+	Start(ctx context.Context, numThreads int)
+}
+
+type ControllerDefintion struct {
+	RequiredGVRs []schema.GroupVersionResource
+	NumThreads   int
+	Create       func(syncedInformers map[schema.GroupVersionResource]cache.SharedIndexInformer) (Controller, error)
+}
+
+func NewControllerManager(ctx context.Context, suffix string, informerSource InformerSource, controllers map[string]ControllerDefintion) *ControllerManager {
```
sttts, line 52: maybe I miss something: what is the added value of this controller manager concept? The controller definition are obviously static. If it is just about waiting for informer sync, create a simple helper/wrapper around a controller start func.

pkg/syncer/controllermanager/controllermanager.go
sttts: This is not "subscribe". It is "Events()". Subscribe would return and keep you subscribed.

pkg/syncer/endpoints/endpoint_downstream_controller.go
davidfestal: Sorry, still need to add GoDoc. Will do it.
This is part of the Proof PR (and would be finally split in a next PR when Upsyncing is finished and merged).

It is a controller that would annotate some Endpoints related to Synced services, to be UpSynced, so that a KNative controller (running against KCP) could see the Endpoint and know that the Service is effective.

In fact it is just a minimal example of GVR-specific controller. A similar one would exist to annotate (for upsyncing) Pods related to a synced deployment.

pkg/syncer/controllermanager/controllermanager.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/kcp-dev/kcp/pkg/logging"
+)
+
+const (
+	ControllerNamePrefix = "syncer-controller-manager-"
+)
+
+type InformerSource struct {
+	Subscribe func(id string) <-chan struct{}
+	Informer  func(gvr schema.GroupVersionResource) (informer cache.SharedIndexInformer, known, synced bool)
```
sttts, line 39: why `synced`? `SharedIndexInformer` includes `HasSynced`.

pkg/syncer/controllermanager/controllermanager.go
sttts: you need clear go docs. I got the meaning after reading your comments and the 188 lines of code.

pkg/syncer/controllermanager/controllermanager.go
sttts: disregard. https://github.com/kcp-dev/kcp/pull/2452/files#r1063304421

pkg/syncer/resourcesync/controller.go
```diff
@@ -261,6 +220,8 @@ func (c *Controller) process(ctx context.Context, key string) error {
 
 	requiredGVRs := getAllGVRs(syncTarget)
 
+	c.cachedDiscovery.Invalidate()
```
sttts, line 223: surprising

pkg/syncer/resourcesync/controller.go
```diff
@@ -63,47 +63,35 @@ const (
...
+	downstreamKubeClient kubernetes.Interface
+
+	cachedDiscovery discovery.CachedDiscoveryInterface
+
+	syncTargetUID               types.UID
+	syncTargetLister            workloadv1alpha1listers.SyncTargetLister
+	synctargetInformerCacheSync cache.InformerSynced
+	kcpClient                   clientset.Interface
+
+	gvrs map[schema.GroupVersionResource]informer.GVRPartialMetadata
```
sttts, line 83: gvrs of what? Are they static?

pkg/syncer/resourcesync/controller.go
```diff
@@ -365,79 +330,163 @@ func (c *Controller) checkSSAR(ctx context.Context, gvr schema.GroupVersionResou
...
-	if _, ok := c.syncerInformerMap[gvr]; ok {
-		logger.V(2).Info("Informer is started already")
-		return
+	for index, ch := range c.subscribers {
+		logger.V(4).Info("Attempting to notify subscribers", "index", index)
+		select {
+		case ch <- struct{}{}:
+			logger.V(4).Info("Successfully notified subscriber", "index", index)
+		default:
+			logger.V(4).Info("Unable to notify subscriber - channel full", "index", index)
```
sttts, line 430: why does this matter?

pkg/syncer/resourcesync/controller.go
```diff
@@ -365,79 +330,163 @@ func (c *Controller) checkSSAR(ctx context.Context, gvr schema.GroupVersionResou
...
+		}
 	}
+}
 
-	syncTargetKey := workloadv1alpha1.ToSyncTargetKey(c.syncTargetClusterName, c.syncTargetName)
+func (c *Controller) addGVR(ctx context.Context, gvr schema.GroupVersionResource, syncTarget *workloadv1alpha1.SyncTarget) error {
```
sttts, line 435: syncTarget?

pkg/syncer/resourcesync/controller.go
```diff
@@ -365,79 +330,163 @@ func (c *Controller) checkSSAR(ctx context.Context, gvr schema.GroupVersionResou
...
+		}
+	}
+	return nil, fmt.Errorf("unable to retrieve discovery for GVR: %s", gvr)
+}
+
+func (c *Controller) Ready() bool {
+	return c.synctargetInformerCacheSync()
+}
+
+func (c *Controller) Subscribe() <-chan struct{} {
```
sttts, line 407: Subscribe to what?

pkg/syncer/controllermanager/controllermanager.go
sttts: Actually, this is "subscribe". With subscribe I would expect either an event handler or the channel passed in, not returned (and logic to remove it when closed).

pkg/syncer/syncer.go
```diff
@@ -59,6 +66,23 @@ const (
 	heartbeatInterval = 20 * time.Second
 )
 
+type FilteredGVRSource struct {
```
sttts, line 69: public? Why at the top of the most central file of the syncer?

pkg/syncer/syncer.go
davidfestal: There are 2 distinct upstream virtual workspaces: one for the Syncer, one for the UpSyncer.
Since the VW URLs are distinct, so are the kube clients and finally the ddsifs.

pkg/syncer/controllermanager/controllermanager.go
davidfestal: This is the same as the DDSIF Subscribe() method.
```
// Subscribe registers for informer/discovery change notifications, returning a channel to which change notifications
// are sent.
func (d *GenericDiscoveringDynamicSharedInformerFactory[Informer, Lister, GenericInformer]) Subscribe(id string) <-chan struct{} {
```
`id` is the identifier of the subscriber, since there might be several subscribers subscribing to receive events from the `InformerSource`.

pkg/syncer/controllermanager/controllermanager.go
davidfestal: Because contrary to typical controllers (which it is not), there can be several instances of controller managers.
Typically one per ddsif.
So the name is useful to distinguish controller manager instances in logs, as well as when subscribing to the informer source (passing the instance name as the id)

pkg/syncer/controllermanager/controllermanager.go
davidfestal: It is not a typical controller. More something along the lines of https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/garbagecollector/garbagecollector_controller.go#L270-L271.

pkg/syncer/status/status_controller.go
davidfestal: Because I had initially planned to possibly change the tests in order to directly mock these functions.
Finally I didn't do it (didn't want to push too much in this PR) , but it's no excluded in the future if we want to move towards the typical unit test pattern used in other KCP controllers, and finally remove the use of the kube fake client in syncer unit tests.

OTOH as it is now, it allows encapsulating the check of the presence / syncing status of the listers, which makes the code in `process` easier to read IMO.

pkg/syncer/resourcesync/controller.go
davidfestal: wdym ?

pkg/syncer/controllermanager/controllermanager.go
davidfestal: GoDoc added

pkg/syncer/endpoints/endpoint_downstream_controller.go
davidfestal: GoDoc added

pkg/syncer/endpoints/endpoint_downstream_controller.go
davidfestal: Fixed

pkg/syncer/endpoints/endpoint_downstream_controller.go
davidfestal: This map is not necessary anymore with the change done to the ControllerManager (following your comment here) : the ControllerManager now only manages the start/stop of the GVR-specific controllers, not their creation anymore.

pkg/syncer/endpoints/endpoint_downstream_controller.go
davidfestal: Done

pkg/syncer/shared/namespace.go
davidfestal: Fixed.

pkg/syncer/resourcesync/controller.go
davidfestal: renamed the mutex to show the field it is protecting.

pkg/syncer/resourcesync/controller.go
davidfestal: The name was but the comment was still obsolete. I finally changed the comment.

pkg/syncer/spec/spec_process.go
davidfestal: Fixed

pkg/syncer/status/status_process.go
```diff
@@ -66,11 +69,33 @@ func (c *Controller) process(ctx context.Context, gvr schema.GroupVersionResourc
 
 	logger = logger.WithValues(DownstreamNamespace, downstreamNamespace, DownstreamName, downstreamName)
 
+	// get the downstream object
```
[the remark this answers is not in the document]
davidfestal: Fixed

pkg/syncer/status/status_process.go
davidfestal: ;-)

pkg/syncer/syncer.go
davidfestal: In fact I also added `endpoints` and `pods` in preparation to what will happen next. So it seems it makes sense to test the group separately from the resource names.

pkg/syncer/controllermanager/controllermanager.go
davidfestal: No reason. I fixed it

pkg/syncer/controllermanager/controllermanager.go
davidfestal: I modified the ControllerManager to only start/stop the GVR-specific controllers, but not create them anymore. This makes the code simpler.
Thanks for pointing this out.

pkg/syncer/controllermanager/controllermanager.go
davidfestal: I fixed that, and also removed the third `sync` returned value from `informer/GenericDiscoveringDynamicSharedInformerfactory.Informer()` since it is not useful.

pkg/syncer/resourcesync/controller.go
```diff
@@ -63,47 +63,35 @@ const (
...
-	mutex             sync.RWMutex
+	queue                workqueue.RateLimitingInterface
+	downstreamKubeClient kubernetes.Interface
+
+	cachedDiscovery discovery.CachedDiscoveryInterface
+
+	syncTargetUID               types.UID
+	syncTargetLister            workloadv1alpha1listers.SyncTargetLister
+	synctargetInformerCacheSync cache.InformerSynced
+	kcpClient                   clientset.Interface
```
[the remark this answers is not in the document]
davidfestal: upstream (not through the VW) to retrieve the SyncTarget. I updated the name.

pkg/syncer/resourcesync/controller.go
davidfestal: I renamed the field. This is the list of the GVRs that should be watched by the ddsifs, and this list will feed the `GVRs()` method (implementation of the `informer.GVRSource` interface.

pkg/syncer/resourcesync/controller.go
```diff
@@ -63,47 +63,35 @@ const (
...
+	mutex sync.RWMutex
+
+	// Support subscribers that want to know when Synced GVRs have changed.
+	subscribersLock sync.Mutex
+	subscribers     []chan<- struct{}
 }
 
 func NewController(
 	syncerLogger logr.Logger,
+	discoveryClient discovery.DiscoveryInterface,
```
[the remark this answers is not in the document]
davidfestal: upstream. I renamed the arg.

pkg/syncer/resourcesync/controller.go
davidfestal: Added GoDoc that should make it clear. This is the implementation of the `informer.GVRSource.Subscribe()` method.

pkg/syncer/syncer.go
davidfestal: Right. renamed it and moved it to the bottom.

pkg/syncer/controllermanager/controllermanager.go
davidfestal: Well, that"s not true finally. I still need to create the controller through the ControllerManager, since the controller queue is typically shot down at the end of the Start method. And I don't want to force a diverging implementation of those controllers.
However I implemented the Create support a bit more cleanly I think.

pkg/syncer/endpoints/endpoint_downstream_controller.go
davidfestal: > the ControllerManager now only manages the start/stop of the GVR-specific controllers, not their creation anymore.

well, not really in fact. See [other comment](https://github.com/kcp-dev/kcp/pull/2452#discussion_r1063720309).
However I removed this map.

pkg/syncer/controllermanager/controllermanager.go
sttts: call it `process()`

pkg/syncer/resourcesync/controller.go
```diff
@@ -152,7 +118,75 @@ func NewController(
...
+
+// controller is a control loop that watches synctarget. It starts/stops spec syncer and status syncer
+// per gvr based on synctarget.Status.SyncedResources.
+// All the spec/status syncer share the same downstreamNSInformer and upstreamSecretInformer. Informers
+// for gvr is started separated for each syncer.
+type controller struct {
+	queue                workqueue.RateLimitingInterface
+	downstreamKubeClient kubernetes.Interface
+
+	cachedDiscovery discovery.CachedDiscoveryInterface
```
sttts, line 131: discovery of what?

pkg/syncer/resourcesync/controller.go
sttts: Rediscovering on every sync target change is very strange.

pkg/syncer/resourcesync/controller.go
```diff
@@ -365,79 +380,128 @@ func (c *Controller) checkSSAR(ctx context.Context, gvr schema.GroupVersionResou
...
 			logger.WithValues("gvr", gvr.String()).V(2).Info("Stop syncer for gvr")
-			informer.cancel()
-			delete(c.syncerInformerMap, gvr)
+			delete(c.gvrsToWatch, gvr)
+			updated = true
+		}
+	}
+
+	if updated {
+		c.notifySubscribers(ctx)
```
sttts, line 401: don't do this within the lock

pkg/syncer/resourcesync/controller.go
```diff
@@ -365,79 +380,128 @@ func (c *Controller) checkSSAR(ctx context.Context, gvr schema.GroupVersionResou
...
+		return err
 	}
 
-	logger.V(2).Info("Start informer for gvr")
-	syncerCtx, cancel := context.WithCancel(ctx)
+	c.gvrsToWatch[gvr] = *partialMetadata
 
-	go downstreamInformer.Informer().Run(syncerCtx.Done())
-	go upstreamInformer.Informer().Run(syncerCtx.Done())
+	c.notifySubscribers(ctx)
```
sttts, line 468: don't do this within the lock

pkg/syncer/spec/spec_controller.go
```diff
@@ -39,39 +40,40 @@ import (
...
-	upstreamClient            kcpdynamic.ClusterInterface
-	downstreamClient          dynamic.Interface
-	syncerInformers           resourcesync.SyncerInformerFactory
-	downstreamNSInformer      informers.GenericInformer
+	upstreamClient   kcpdynamic.ClusterInterface
+	downstreamClient dynamic.Interface
+
+	getUpstreamLister                 func(gvr schema.GroupVersionResource) (kcpcache.GenericClusterLister, error)
+	getDownstreamLister               func(gvr schema.GroupVersionResource) (cache.GenericLister, error)
+	listDownstreamNamespacesByLocator func(jsonLocator string) ([]interface{}, error)
```
sttts, line 75: why untyped?

pkg/syncer/status/status_process.go
```diff
@@ -66,11 +69,30 @@ func (c *Controller) process(ctx context.Context, gvr schema.GroupVersionResourc
...
-	// to upstream
+	downstreamLister, err := c.getDownstreamLister(gvr)
+	if err != nil {
+		return err
+	}
+
+	var resourceExists bool
+	var downstreamObj runtime.Object
 	var namespaceLocator *shared.NamespaceLocator
 	var locatorExists bool
```
sttts, line 80: so many shared vars are usually a sign for bad code factoring

pkg/syncer/status/status_process_test.go
```diff
@@ -539,32 +591,81 @@ func TestStatusSyncerProcess(t *testing.T) {
...
+				}
+				_, unsynced = ddsifForDownstream.Listers()
+				for _, gvr := range unsynced {
+					informer, _ := ddsifForDownstream.ForResource(gvr)
+					cache.WaitForCacheSync(ctx.Done(), informer.Informer().HasSynced)
+					t.Logf("%s: downstream ddsif informer synced for gvr %s", t.Name(), gvr.String())
+				}
+				t.Logf("%s: gvrs updated", t.Name())
+				gvrsUpdated <- struct{}{}
+			}()
```
sttts, line 653: is there no simpler way to test the controller? This complex synchronization code is a warning sign.

pkg/syncer/syncer.go
```diff
@@ -136,7 +143,16 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
...
-	upstreamDynamicClusterClient, err := kcpdynamic.NewForConfig(upstreamConfig)
+	upstreamSyncerClusterClient, err := kcpdynamic.NewForConfig(upstreamConfig)
+	if err != nil {
+		return err
+	}
+
+	upstreamUpsyncConfig := rest.CopyConfig(cfg.UpstreamConfig)
+	// Upsyncing to Virtual Workspace for upsyncing
+	upstreamUpsyncConfig.Host = strings.Replace(syncerVirtualWorkspaceURL, "/services/syncer", "/services/upsyncer", 1)
+	upstreamUpsyncConfig.UserAgent = "kcp#upsyncer/" + kcpVersion
```
sttts, line 154: would prefer this to be the binary "syncer", and then something to qualify it as upsyncer controller.

pkg/syncer/syncer.go
```diff
@@ -224,41 +265,115 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
...
 	go apiImporter.Start(klog.NewContext(ctx, logger.WithValues("resources", resources)), importPollInterval)
-	go syncerInformers.Start(ctx, 1)
+	go syncTargetGVRSource.Start(ctx, 1)
 	go specSyncer.Start(ctx, numSyncerThreads)
 	go statusSyncer.Start(ctx, numSyncerThreads)
+
 	go downstreamNamespaceController.Start(ctx, numSyncerThreads)
 
+	// Create and start GVR-specific controllers through controller managers
+	upstreamSyncerControllerManager := controllermanager.NewControllerManager(ctx,
```
sttts, line 326: this ordering is dangerous. Please always do this:
1. create informers
2. create controllers
3. start informers
4. wait for informers
5. start controllers

pkg/syncer/namespace/namespace_downstream_controller.go
```diff
@@ -100,29 +99,45 @@ func NewDownstreamController(
...
+			if !known || !synced {
+				return nil, errors.New("informer should have already been created and synced for namespaces in the downstream informer factory")
+			}
+			return lister.List(labels.Everything())
 		},
 		isDowntreamNamespaceEmpty: func(ctx context.Context, namespace string) (bool, error) {
-			gvrs, err := syncerInformers.SyncableGVRs()
-			if err != nil {
-				return false, err
+			listers, notSynced := ddsifForDownstream.Listers()
```
sttts, line 130: elsewhere `Lister(gvr)` tells us whether it is syncer. Here it tells us what is not synced.

pkg/syncer/controllermanager/controllermanager.go
```diff
@@ -0,0 +1,193 @@
...
+
+func (c *ControllerManager) updateControllers(ctx context.Context) {
+	logger := klog.FromContext(ctx)
+	controllersToStart := map[string]CreateControllerFunc{}
+controllerLoop:
+	for controllerName, managedController := range c.managedControllers {
+		requiredGVRs := managedController.RequiredGVRs
+		for _, gvr := range requiredGVRs {
+			if informer, known := c.informerSource.Informer(gvr); !known {
+				continue controllerLoop
```
sttts, line 146: I don't understand this logic. Why should we not wait for sync on known==false ?

pkg/informer/informer.go
```diff
@@ -394,16 +394,16 @@ func (d *GenericDiscoveringDynamicSharedInformerFactory[Informer, Lister, Generi
 
 // Informer returns an informer for the given resource-type if it exists, whether it is
 // known by this informer factory, and whether it is synced.
-func (d *GenericDiscoveringDynamicSharedInformerFactory[Informer, Lister, GenericInformer]) Informer(gvr schema.GroupVersionResource) (informer Informer, known, synced bool) {
+func (d *GenericDiscoveringDynamicSharedInformerFactory[Informer, Lister, GenericInformer]) Informer(gvr schema.GroupVersionResource) (informer Informer, known bool) {
```
sttts, line 397: known is actually a strange name and https://github.com/kcp-dev/kcp/pull/2452/files#r1064592734 is a strange logic. Too much about state transitions, which is fragile.

pkg/syncer/controllermanager/controllermanager.go
```diff
@@ -0,0 +1,193 @@
...
+		if _, ok := c.startedControllers[controllerName]; ok {
+			// The controller is already started
+			continue
+		}
+
+		// Create the controller
+
+		start, err := create(ctx)
+		if err != nil {
+			logger.Error(err, "failed creating controller", "controller", controllerName)
```
sttts, line 168: this panic further down

pkg/syncer/controllermanager/controllermanager.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+		// The controller should not be running
+		// Stop it and remove it from the list of started controllers
+		cancelFunc()
+		delete(c.startedControllers, controllerName)
+	}
+
+	// Add missing controllers that were created and started above
+	for controllerName, cancelFunc := range newlyStartedControllers {
+		c.startedControllers[controllerName] = cancelFunc
```
sttts, line 191: why not set it directly above?

pkg/syncer/controllermanager/controllermanager.go
davidfestal: because if the informer doesn't exist for this GVR, and would later be added in the informer source, then the InformerSource (ddsif) will notify the controller manager that supported GVRs changed, and the informer will be known this time.

pkg/informer/informer.go
sttts: nvmd. I thought this func would start an informer.

pkg/syncer/controllermanager/controllermanager.go
sttts: nvmd. Misunderstood known. `exists` would be a better word.

pkg/syncer/endpoints/endpoint_downstream_controller.go
```diff
@@ -0,0 +1,160 @@
...
+	c := &controller{
+		queue: workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), ControllerName),
+	}
+
+	endpointsInformer, ok := ddsifForDownstream.Informer(endpointsGVR)
+	if !ok {
+		return nil, errors.New("endpoints informer should be available")
+	}
+
+	endpointsInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
sttts, line 61: remind me what the logic is that this handler gets initial add events?

pkg/syncer/status/status_process_test.go
davidfestal: It is related to the fact that we still create the StatusSyncer through the `NewStatusSyncer()` method, instead of the directly creating the StatusSyncer struct. So we need to pass real ddsif instances (similar to how previously we were passing the real informer factories), because these tests are still using fake client-go clients with reactors.

As mentioned in [this comment answer](https://github.com/kcp-dev/kcp/pull/2452#discussion_r1063557352), I had thought about switching the a fake-client-go-free, reactor-free test implementation, with only fake lister implementations.
I'd prefer keeping that for a distinct PR though.

pkg/syncer/controllermanager/controllermanager.go
davidfestal: > `exists` would be a better word.

You're right. I'll change this.

pkg/syncer/controllermanager/controllermanager.go
davidfestal: Thanks for pointing out. There should be a `continue` statement

pkg/syncer/controllermanager/controllermanager.go
davidfestal: right.

pkg/syncer/spec/spec_controller.go
davidfestal: Didn't think at using the generic `indexers.ByIndex()`. I'll fix it.

pkg/syncer/resourcesync/controller.go
davidfestal: Renamed to `upstreamSyncerDiscoveryClient`

pkg/syncer/resourcesync/controller.go
davidfestal: Fixed in commit da89d306186c852a5f9f9ab2a3a549f896f5f7e0

pkg/syncer/resourcesync/controller.go
davidfestal: Fixed in commit da89d306186c852a5f9f9ab2a3a549f896f5f7e0

pkg/syncer/spec/spec_controller.go
davidfestal: Fixed in commit da89d306186c852a5f9f9ab2a3a549f896f5f7e0

pkg/syncer/syncer.go
davidfestal: Fixed in commit da89d306186c852a5f9f9ab2a3a549f896f5f7e0

pkg/syncer/namespace/namespace_downstream_controller.go
davidfestal: Fixed in commit da89d306186c852a5f9f9ab2a3a549f896f5f7e0 by the fact that I refactored the ddsif a bit to do some cleanup, and have an `Informers()` method (instead of `Listers()`) which returns a map of `GenericLister`s so that it allows directly getting either an `SharedIndexInformer` or  a `Lister` from this map.
This allowed removing the additional `Lister()` and `Informer()` calls without loosing any feature.

pkg/syncer/controllermanager/controllermanager.go
davidfestal: Fixed in commit da89d306186c852a5f9f9ab2a3a549f896f5f7e0

pkg/informer/informer.go
davidfestal: Still fixed as part of the refactoring mentioned here: https://github.com/kcp-dev/kcp/pull/2452#discussion_r1064848346

pkg/syncer/controllermanager/controllermanager.go
davidfestal: Fixed in commit da89d306186c852a5f9f9ab2a3a549f896f5f7e0

pkg/informer/informer.go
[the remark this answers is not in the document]
davidfestal: Function removed as part of the informer cleanup.

pkg/syncer/controllermanager/controllermanager.go
davidfestal: Fixed in commit da89d306186c852a5f9f9ab2a3a549f896f5f7e0

pkg/syncer/controllermanager/controllermanager.go
davidfestal: Renamed

pkg/syncer/controllermanager/controllermanager.go
davidfestal: Changed while doing the cleanup mentioned in comment https://github.com/kcp-dev/kcp/pull/2452#discussion_r1064848346

pkg/syncer/resourcesync/controller.go
davidfestal: In fact I assume we could invalidate only in the `UpdateFunc()` of the Event Handler, when the `synctarget.Status.SyncedResources` changed between `oldObj` and `newObj` ?

pkg/syncer/endpoints/endpoint_downstream_controller.go
davidfestal: When this handler is added, the informer is already started. However, the AddEventHandler method supports this, and explicitly notifies initial add events as shown by the following code snippet:

https://github.com/kubernetes/client-go/blob/v0.24.9/tools/cache/shared_informer.go#L547-L563

davidfestal:
/unhold

workspace refactor is merged

pkg/informer/informer.go
```diff
@@ -353,13 +353,13 @@ func (d *GenericDiscoveringDynamicSharedInformerFactory[Informer, Lister, Generi
...
 
-// Listers returns a map of per-resource-type listers for all types that are
+// Informers returns a map of per-resource-type generic informers for all types that are
 // known by this informer factory, and that are synced.
 //
 // If any informers aren't synced, their GVRs are returned so that they can be
 // checked and processed later.
-func (d *GenericDiscoveringDynamicSharedInformerFactory[Informer, Lister, GenericInformer]) Listers() (listers map[schema.GroupVersionResource]Lister, notSynced []schema.GroupVersionResource) {
-	listers = map[schema.GroupVersionResource]Lister{}
+func (d *GenericDiscoveringDynamicSharedInformerFactory[Informer, Lister, GenericInformer]) Informers() (informers map[schema.GroupVersionResource]GenericInformer, notSynced []schema.GroupVersionResource) {
```
sttts, line 361: :+1:, much better

pkg/syncer/controllermanager/controllermanager.go
```diff
@@ -0,0 +1,199 @@
...
+
+	// Create and start missing controllers that have their required GVRs synced
+	for controllerName, create := range controllersToStart {
+		if _, ok := c.startedControllers[controllerName]; ok {
+			// The controller is already started
+			continue
+		}
+
+		// Create the controller
+
```
sttts, line 187: nit: no empty line

pkg/syncer/syncer.go
```diff
@@ -224,41 +265,132 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
...
+				}
+				return informers, notSynced
+			},
+		},
+		map[string]controllermanager.ManagedController{},
+	)
+	go upstreamUpsyncerControllerManager.Start(ctx)
+
+	downstreamSyncerControllerManager := controllermanager.NewControllerManager(ctx,
+		"downstream",
```
sttts, line 361: nit: downstream-syncer to be consistent

pkg/syncer/syncer.go
```diff
@@ -134,9 +141,18 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 
 	upstreamConfig := rest.CopyConfig(cfg.UpstreamConfig)
 	upstreamConfig.Host = syncerVirtualWorkspaceURL
-	rest.AddUserAgent(upstreamConfig, "kcp#spec-syncer/"+kcpVersion)
+	rest.AddUserAgent(upstreamConfig, "kcp#syncing/"+kcpVersion)
```
sttts, line 144: don't think that's a valid user agent.

davidfestal: It seems that the way user agents are set in the Syncer should be reworked globally to bring back both inner consistency, and also consistency with other KCP components.

Dedicated issue created: https://github.com/kcp-dev/kcp/issues/2581

sttts:
Failing DNS tests twice. Coincidence?

davidfestal:
> Failing DNS tests twice. Coincidence?

Unfortunately I approved PR https://github.com/kcp-dev/kcp/pull/2542 yesterday (GH tests were successful), but in fact  it might be that it needed some rebase first. And now the related tests are failing. So it doesn't seem to be related to this PR, but I still have to check.

davidfestal:
> > Failing DNS tests twice. Coincidence?
> 
> Unfortunately I approved PR #2542 yesterday (GH tests were successful), but in fact it might be that it needed some rebase first. And now the related tests are failing. So it doesn't seem to be related to this PR, but I still have to check.

Well, in fact the test is perfect and worked exactly as expected. it was this PR having a bug. See commit https://github.com/kcp-dev/kcp/pull/2452/commits/3de7ac8b4463f1721f2ad7489ebfa400d9ca17e2

Great and very useful e2e test @lionelvillard !

PR #2455, opened 2022-12-06
WIP 🌱 Syncer: Cleanup of DNS-related resources

davidfestal:
## Summary

Cleanup of DNS-related resources when no more downstream namespaces exist for a given upstream workspace. 

Disclaimer:
- based on PRs kcp-dev/kcp#2675 and kcp-dev/kcp#2946. Should be merged as soon as those have been.
- only the last commit is really part of this PR

## Related issue(s)

Fixes kcp-dev/contrib-tmc#80


pkg/syncer/spec/dns/resources.go
```diff
@@ -89,17 +100,23 @@ func MakeDeployment(name, namespace, image string) *appsv1.Deployment {
 	deployment.Spec.Template.Spec.Containers[0].Image = image
 	deployment.Spec.Template.Spec.Containers[0].Args[3] = name
 	deployment.Spec.Template.Spec.ServiceAccountName = name
+	deploymentTemplate.Labels = map[string]string{
```
lionelvillard, line 103: this should be `deployment`

davidfestal: Thanks !

davidfestal:
> > Can we do this with actual garbage collection and owner refs?
> 
> I think yes. Namespaced dependents can have cluster-scoped owners (Namespaces in this case).

Not sure about how you would use garbage collection in the specific case.
Do you mean that every DNS-related resource for a given KCP workspace should have as many owner references as the number of downstream namespaces originating from the given KCP workspace ?

In order to do this, in the `EnsureDNSUpAndReady` function we would need to:
- check that every DNS-related resource has the namespace of the currently-synced resource in its owner refs,
- update every DNS-related resource if necessary to add the `namespace` owner ref.
This would be required during *every* resource sync. I'd rather do more work during cleanup, and keep the resource syncing main reconciliation loop as quick as possible.

Then I assume that the DNS-related resources deletion would be triggered by cascading deletion when ***all*** the namespaces related to the given KCP workspace would be deleted. However we wouldn't control ***when*** this cleaning would happen, which would make it hard to synchronize this cleaning with the on-the-fly creation of those DNS-related resources when they are necessary (in the `EnsureDNSUpAndReady` function). Such a synchronization is for now still missing from this PR, but might be necessary and should be thought of.

Last point: when we start implementing the network isolation through NetworkPolicies on the physical cluster, we will need this concept of KCP tenant on downstream (== a originating KCP workspace + related SyncTarget). That seemed quite natural to use it here as well.

So, unless I'm missing something, applying the standard Kubernetes garbage collection and owner refs mechanism here is not as simple as it may seem, and may bring more slowness and complexity.

Let me finally just point to [a previous PR comment](https://github.com/kcp-dev/kcp/pull/2299#discussion_r1015479565) answer to remind that we are not in the standard Kubernetes case, because everything downstream is finally driven by the individual resources synced from upstream.

ncdc:
@davidfestal are you still interested in moving forward with this?

davidfestal:
/hold

on hold until PR https://github.com/kcp-dev/kcp/pull/2675 is merged.

davidfestal:
/hold

Wait until PRs #2675 and #2946 are merged.

mjudeikis:
/close
TMC

Issue #2278, opened 2022-10-27
bug: flake: TestSyncerLifecycle


stevekuznetsov:
### Describe the bug

```
I1027 17:33:59.650858   30405 spec_controller.go:267] "processing key" syncTarget.workspace="root:e2e-org-zfx2r:e2e-workspace-549q2" syncTarget.name="psyncer-01" syncTarget.key="7qFQgWTy6tlERvRELtSVSH1TZp0wWhDAhtd9gW" reconciler="kcp-workload-syncer-spec" key="root:e2e-org-zfx2r:e2e-workspace-549q2|test-syncer/syncer-test" gvr="apps/v1, Resource=deployments"
E1027 17:33:59.668414   30405 spec_process.go:483] "Error upserting upstream resource to downstream" err="Deployment.apps \"syncer-test\" is invalid: [spec.selector: Invalid value: \"null\": spec.selector in body must be of type object: \"null\", spec.template.spec.containers: Invalid value: \"null\": spec.template.spec.containers in body must be of type array: \"null\"]" syncTarget.workspace="root:e2e-org-zfx2r:e2e-workspace-549q2" syncTarget.name="psyncer-01" syncTarget.key="7qFQgWTy6tlERvRELtSVSH1TZp0wWhDAhtd9gW" reconciler="kcp-workload-syncer-spec" key="root:e2e-org-zfx2r:e2e-workspace-549q2|test-syncer/syncer-test" gvr="apps/v1, Resource=deployments" downstream.name="syncer-test"
E1027 17:33:59.668532   30405 spec_controller.go:274] kcp-workload-syncer-spec failed to sync {{"apps" "v1" "deployments"} "root:e2e-org-zfx2r:e2e-workspace-549q2|test-syncer/syncer-test"}, err: Deployment.apps "syncer-test" is invalid: [spec.selector: Invalid value: "null": spec.selector in body must be of type object: "null", spec.template.spec.containers: Invalid value: "null": spec.template.spec.containers in body must be of type array: "null"]
[...]

PR #2467, opened 2022-12-08
🐛 Fix no-op patches in committer

### LINKED ISSUE #2466: bug: committer sends "" patches when no changes, instead of being a no-op
### Describe the bug

#2458 introduced a scoped committer, but in doing so, it accidentally changed the behavior when there are no changes. Instead of performing no action when there are no changes, it sends a `""` patch.

### Steps To Reproduce

1. Use a Committer
2. Make `old` and `obj` be the same / have no differences
3. See a `""` patch issued

### Expected Behaviour

No patch

### Additional Context

_No response_

ncdc:
## Summary

Fix a bug introduced in #2458 that caused no-op patches to actually issue a `""` patch.

## Related issue(s)

Fixes #2466


vincepri (approved):
/lgtm

ncdc:
Flake #2278 
/retest

Issue #2406, opened 2022-11-24
bug: turning off root-compute-workspace causes a pile up of error logs


erwinvaneyk:
### Describe the bug

When I try to turn off the `root-compute-workspace` option, kcp seems to be functional but outputs a continuous stream of error logs:

```
I1124 02:24:06.267997   19472 httplog.go:90] &{system:apiserver 2d633e39-554d-4ac6-9295-f13218d3ff05 [system:masters] map[]} is acting as &{system:kcp:bootstrapper  [system:kcp:tenancy:workspace-bootstrapper system:authenticated] map[]}
E1124 02:24:06.268088   19472 memcache.go:189] couldn't get current server API group list: <nil>
I1124 02:24:06.268996   19472 httplog.go:90] &{system:apiserver 2d633e39-554d-4ac6-9295-f13218d3ff05 [system:masters] map[]} is acting as &{system:kcp:bootstrapper  [system:kcp:tenancy:workspace-bootstrapper system:authenticated] map[]}
I1124 02:24:06.269209   19472 httplog.go:90] &{system:apiserver 2d633e39-554d-4ac6-9295-f13218d3ff05 [system:masters] map[]} is acting as &{system:kcp:bootstrapper  [system:kcp:tenancy:workspace-bootstrapper system:authenticated] map[]}
E1124 02:24:06.269294   19472 memcache.go:189] couldn't get current server API group list: <nil>
I1124 02:24:06.269988   19472 httplog.go:90] &{system:apiserver 2d633e39-554d-4ac6-9295-f13218d3ff05 [system:masters] map[]} is acting as &{system:kcp:bootstrapper  [system:kcp:tenancy:workspace-bootstrapper system:authenticated] map[]}
I1124 02:24:06.270176   19472 httplog.go:90] &{system:apiserver 2d633e39-554d-4ac6-9295-f13218d3ff05 [system:masters] map[]} is acting as &{system:kcp:bootstrapper  [system:kcp:tenancy:workspace-bootstrapper system:authenticated] map[]}
E1124 02:24:06.270255   19472 memcache.go:189] couldn't get current server API group list: <nil>
I1124 02:24:06.271059   19472 httplog.go:90] &{system:apiserver 2d633e39-554d-4ac6-9295-f13218d3ff05 [system:masters] map[]} is acting as &{system:kcp:bootstrapper  [system:kcp:tenancy:workspace-bootstrapper system:authenticated] map[]}
[...]

PR #2474, opened 2022-12-12
🐛 Add missing battery annotation for root-compute-workspace

qiujian16:

## Summary

the annotation should be added to all resources in root:compute workspace

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/2406

### DIFF
--- a/config/rootcompute/kube-1.24/clusterrole-kubernetes-maximal-permission-policy.yaml
+++ b/config/rootcompute/kube-1.24/clusterrole-kubernetes-maximal-permission-policy.yaml
@@ -1,6 +1,8 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
+  annotations:
+    bootstrap.kcp.dev/battery: root-compute-workspace
   name: compute:apiexport:kubernetes:maximal-permission-policy
 rules:
 - apiGroups: [""]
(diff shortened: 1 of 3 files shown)


config/rootcompute/kube-1.24/clusterrole-kubernetes-maximal-permission-policy.yaml
```diff
@@ -1,6 +1,9 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
+  annotations:
+    bootstrap.kcp.dev/create-only: "true"
```
davidfestal, line 5: Do we really want to use `createOnly` here ?

Don't we want the policy to be updated if necessary at every restart, in order to update the rules for every resource ? Never updating it seems strange to me, and will make future additions of GVRs in the `root:compute:kubernetes` APIExport quite hard.
By the way the same question stands for the `apiexport-kubernetes.k8s.io.yaml` file.
@sttts thoughts ?

qiujian16: updated

stevekuznetsov (approved):

davidfestal (approved):

ncdc:
#1994 
/retest

ncdc:
#2479 
/retest-required

ncdc:
Not sure why the GH actions jobs for e2e-sharded and e2e-shared-server are not running. I'm going to close and reopen the PR to see if that will trigger them.

PR #2477, opened 2022-12-12
🌱 Use caching authorizers per-workspace in initializingworkspaces/builder

vincepri:
This changeset is scoped to the virtual/initializingworkspace package.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/authorization/delegated/caching_authorizer.go
+++ b/pkg/authorization/delegated/caching_authorizer.go
@@ -1,0 +1,109 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package delegated
+
+import (
+	"context"
+	"time"
+
+	kcpkubernetesclientset "github.com/kcp-dev/client-go/kubernetes"
+	"github.com/kcp-dev/logicalcluster/v3"
+
+	"k8s.io/apimachinery/pkg/util/cache"
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+)
+
+// CachingOptions contains options to create a new Delegated Caching Authorizer.
+type CachingOptions struct {
+	Options
+
+	// TTL is the default time-to-live when a delegated authorizer
+	// is stored in the internal cache.
+	TTL time.Duration
+}
+
+func (c *CachingOptions) defaults() {
+	c.Options.defaults()
+	if c.TTL == 0 {
+		c.TTL = 12 * time.Hour
+	}
+}
+
+// CachingAuthorizerFunc looks similar to authorizer.AuthorizerFunc with the
+// additional cache parameter for delegated authorizers.
+type CachingAuthorizerFunc func(ctx context.Context, cache Cache, a authorizer.Attributes) (authorizer.Decision, string, error)
+
+// Cache contains methods that define a delegated caching authorizer.
+type Cache interface {
+	// Get returns the delegated authorizer for the given logical cluster.
+	Get(clusterName logicalcluster.Name) (authorizer.Authorizer, error)
+}
+
+// NewCachingAuthorizer creates a new Authorizer that holds an internal cache of
+// Delegated Authorizer(s).
+func NewCachingAuthorizer(client kcpkubernetesclientset.ClusterInterface, auth CachingAuthorizerFunc, opts CachingOptions) *cachingAuthorizer {
+	opts.defaults()
+	return &cachingAuthorizer{
+		opts:   &opts,
+		auth:   auth,
+		cache:  cache.NewExpiring(),
+		client: client,
+	}
+}
+
+// cachingAuthorizer is a wrapper around authorizer.Authorize that uses
+// an internal expiring cache.
+type cachingAuthorizer struct {
+	opts  *CachingOptions
+	cache *cache.Expiring
+
+	auth   CachingAuthorizerFunc
+	client kcpkubernetesclientset.ClusterInterface
+}
+
+// load loads the authorizer from the cache, if any.
+func (c *cachingAuthorizer) load(clusterName logicalcluster.Name) authorizer.Authorizer {
+	value, ok := c.cache.Get(clusterName)
+	if !ok || value == nil {
+		return nil
+	}
+	authz, ok := value.(authorizer.Authorizer)
+	if !ok {
+		return nil
+	}
+	return authz
+}
+
+func (c *cachingAuthorizer) Get(clusterName logicalcluster.Name) (authorizer.Authorizer, error) {
+	if authz := c.load(clusterName); authz != nil {
+		return authz, nil
+	}
+
+	// Create the delegated authorizer.
+	authz, err := NewDelegatedAuthorizer(clusterName, c.client, c.opts.Options)
+	if err != nil {
+		return nil, err
+	}
+
+	// Store the cache and return.
+	c.cache.Set(clusterName, authz, c.opts.TTL)
+	return authz, nil
+}
+
+func (c *cachingAuthorizer) Authorize(ctx context.Context, attr authorizer.Attributes) (authorized authorizer.Decision, reason string, err error) {
+	return c.auth(ctx, c, attr)
+}
--- a/pkg/virtual/initializingworkspaces/builder/build.go
+++ b/pkg/virtual/initializingworkspaces/builder/build.go
@@ -87,6 +87,8 @@ func BuildVirtualWorkspace(
 		v.Schema.Raw = bs // wipe schemas. We don't want validation here.
 	}
 
+	cachingAuthorizer := delegated.NewCachingAuthorizer(kubeClusterClient, authorizerWithCache, delegated.CachingOptions{})
+
 	wildcardLogicalClustersName := initializingworkspaces.VirtualWorkspaceName + "-wildcard-logicalclusters"
 	wildcardLogicalClusters := &virtualworkspacesdynamic.DynamicVirtualWorkspace{
 		RootPathResolver: framework.RootPathResolverFunc(func(urlPath string, requestContext context.Context) (accepted bool, prefixToStrip string, completedContext context.Context) {
@@ -104,7 +106,7 @@ func BuildVirtualWorkspace(
 			completedContext = dynamiccontext.WithAPIDomainKey(completedContext, apiDomain)
 			return true, prefixToStrip, completedContext
 		}),
-		Authorizer: newAuthorizer(kubeClusterClient),
+		Authorizer: cachingAuthorizer,
 		ReadyChecker: framework.ReadyFunc(func() error {
 			return nil
 		}),
@@ -141,7 +143,7 @@ func BuildVirtualWorkspace(
 			completedContext = dynamiccontext.WithAPIDomainKey(completedContext, apiDomain)
 			return true, prefixToStrip, completedContext
 		}),
-		Authorizer: newAuthorizer(kubeClusterClient),
+		Authorizer: cachingAuthorizer,
 		ReadyChecker: framework.ReadyFunc(func() error {
 			return nil
 		}),
@@ -183,7 +185,7 @@ func BuildVirtualWorkspace(
 			completedContext = dynamiccontext.WithAPIDomainKey(completedContext, apiDomain)
 			return true, prefixToStrip, completedContext
 		}),
-		Authorizer: newAuthorizer(kubeClusterClient),
+		Authorizer: cachingAuthorizer,
 		ReadyChecker: framework.ReadyFunc(func() error {
 			select {
 			case <-workspaceContentReadyCh:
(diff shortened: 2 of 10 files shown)


pkg/virtual/initializingworkspaces/builder/build.go
```diff
@@ -115,7 +113,7 @@ func BuildVirtualWorkspace(
 			completedContext = dynamiccontext.WithAPIDomainKey(completedContext, apiDomain)
 			return true, prefixToStrip, completedContext
 		}),
-		Authorizer: newAuthorizer(kubeClusterClient),
+		Authorizer: newCachingAuthorizer(kubeClusterClient),
```
vincepri, line 116: Can the cache be shared here across all the BuildVirtualWorkspace calls?

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
```diff
@@ -0,0 +1,71 @@
...
+type cachingAuthorizer struct {
+	client kcpkubernetesclientset.ClusterInterface
+
+	cacheMu sync.RWMutex
+	cache   map[string]authorizer.Authorizer
+}
+
+func (c *cachingAuthorizer) loadOrStore(workspace logicalcluster.Name) (authorizer.Authorizer, error) {
+	c.cacheMu.RLock()
+	if authz, ok := c.cache[workspace.String()]; ok && authz != nil {
```
vincepri, line 31: Is a key by workspace enough in this case?

pkg/virtual/initializingworkspaces/builder/build.go
ncdc: In this package, I'd say yes

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
```diff
@@ -0,0 +1,71 @@
...
+
+func newCachingAuthorizer(client kcpkubernetesclientset.ClusterInterface) authorizer.Authorizer {
+	return &cachingAuthorizer{client: client, cache: make(map[string]authorizer.Authorizer)}
+}
+
+type cachingAuthorizer struct {
+	client kcpkubernetesclientset.ClusterInterface
+
+	cacheMu sync.RWMutex
+	cache   map[string]authorizer.Authorizer
```
vincepri, line 26: Would we prefer an LRU cache here instead? Would that make the caching authorizer struct a bit more portable between here and other places?

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/kcp-dev/kcp/pkg/apis/tenancy/initialization"
+	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/authorization/delegated"
+	dynamiccontext "github.com/kcp-dev/kcp/pkg/virtual/framework/dynamic/context"
+	"github.com/kcp-dev/logicalcluster/v2"
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+	"k8s.io/klog/v2"
+)
+
+func newCachingAuthorizer(client kcpkubernetesclientset.ClusterInterface) authorizer.Authorizer {
```
ncdc, line 20: We're trying to return specific types whenever possible (xref https://github.com/golang/go/wiki/CodeReviewComments#interfaces) so I think you can return a `*cachingAuthorizer` here.

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
```diff
@@ -0,0 +1,71 @@
...
+	return &cachingAuthorizer{client: client, cache: make(map[string]authorizer.Authorizer)}
+}
+
+type cachingAuthorizer struct {
+	client kcpkubernetesclientset.ClusterInterface
+
+	cacheMu sync.RWMutex
```
ncdc, line 27: We usually call this `lock` or `cacheLock`

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
ncdc: With how you have it implemented, I'd say 👍

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
ncdc: This is probably an ok start? It doesn't remove entries when a workspace is deleted, but we could add that if we wanted.

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
```diff
@@ -0,0 +1,73 @@
...
+	cache   map[string]authorizer.Authorizer
+}
+
+func (c *cachingAuthorizer) loadOrStore(workspace logicalcluster.Name) (authorizer.Authorizer, error) {
```
ncdc, line 31:
```suggestion
func (c *cachingAuthorizer) loadOrStore(clusterName logicalcluster.Name) (authorizer.Authorizer, error) {
```

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
```diff
@@ -0,0 +1,73 @@
...
+	if authz, ok := c.cache[workspace.String()]; ok && authz != nil {
+		c.cacheMu.RUnlock()
+		return authz, nil
+	}
+	c.cacheMu.RUnlock()
+
+	// We haven't found the authorizer, lock the map, create and store it.
+	c.cacheMu.Lock()
+	defer c.cacheMu.Unlock()
+	authz, err := delegated.NewDelegatedAuthorizer(workspace, c.client)
```
ncdc, line 42: Need 1 more existence check to avoid duplicate creates

pkg/virtual/initializingworkspaces/builder/build.go
vincepri: Perfect, updated!

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
vincepri: How would I know within the builder that a workspace has been deleted?

ncdc: We'd do like we've done in other places (e.g. quota) - spin up a controller that reacts to informer events, in this case deletions.

vincepri: Got it! That sounds good and can definitely do that, is that preferable to a basic LRU cache?

ncdc: I'd probably start with the LRU as that's less code. WDYT?

vincepri: pushed, ptal

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
```diff
@@ -0,0 +1,102 @@
...
+
+// cachingAuthorizer is a wrapper around athorizer.Authorize that uses
+// an internal expiring cache.
+type cachingAuthorizer struct {
+	client kcpkubernetesclientset.ClusterInterface
+	cache  *cache.Expiring
+}
+
+// load loads the authorizer from the cache, if any.
+// calls should be made while holding the lock.
```
ncdc, line 49:
```suggestion
```

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
```diff
@@ -0,0 +1,101 @@
...
+	logger := klog.FromContext(ctx)
+	workspace, name, err := initialization.TypeFrom(tenancyv1alpha1.ClusterWorkspaceInitializer(dynamiccontext.APIDomainKeyFrom(ctx)))
+	if err != nil {
+		logger.V(2).Info(err.Error())
+		return authorizer.DecisionNoOpinion, "unable to determine initializer", fmt.Errorf("access not permitted")
+	}
+
+	authz, err := c.loadOrStore(workspace)
+	if err != nil {
+		return authorizer.DecisionNoOpinion, "error", err
```
ncdc, line 87: @s-urbaniak would appreciate guidance on what to use here for the reason

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
```diff
@@ -0,0 +1,101 @@
...
+	c.cache.Set(clusterName, authz, 12*time.Hour)
+	return authz, nil
+}
+
+func (c *cachingAuthorizer) Authorize(ctx context.Context, attr authorizer.Attributes) (authorized authorizer.Decision, reason string, err error) {
+	logger := klog.FromContext(ctx)
+	workspace, name, err := initialization.TypeFrom(tenancyv1alpha1.ClusterWorkspaceInitializer(dynamiccontext.APIDomainKeyFrom(ctx)))
+	if err != nil {
+		logger.V(2).Info(err.Error())
+		return authorizer.DecisionNoOpinion, "unable to determine initializer", fmt.Errorf("access not permitted")
```
ncdc, line 82: @s-urbaniak PTAL

ncdc (commented):
LGMT overall. Would like @s-urbaniak to review the authz/audit pieces

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
```diff
@@ -0,0 +1,102 @@
...
+	"k8s.io/klog/v2"
+
+	corev1alpha1 "github.com/kcp-dev/kcp/pkg/apis/core/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/apis/tenancy/initialization"
+	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/authorization/delegated"
+	dynamiccontext "github.com/kcp-dev/kcp/pkg/virtual/framework/dynamic/context"
+)
+
+func newCachingAuthorizer(client kcpkubernetesclientset.ClusterInterface) *cachingAuthorizer {
```
sttts, line 38: am surprised to see this a a private helper. Isn't this a very generic pattern we want to use in many places?

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
```diff
@@ -0,0 +1,102 @@
...
+	return &cachingAuthorizer{client: client, cache: cache.NewExpiring()}
+}
+
+// cachingAuthorizer is a wrapper around athorizer.Authorize that uses
```
s-urbaniak, line 42: s/athorizer.Authorize/authorizer.Authorize

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
s-urbaniak: I tend to agree, however in its current form the caching authorizer is not reusable. It is hardcoded to assert the `initialize` verb against `workspacetypes`. I rather suggest to either:

a) make the `Authorize` logic pluggable and make this constructor public or
b) leave this private and call it `newInitializingWorkspacesAuthorizer` and leave the caching logic as an internal implementation detail.

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
s-urbaniak: This is taken 1:1 from the current implementation, so if anything I suggest to take improvements as a separate PR. However, my comment for this (and thus the previous implementation):

- specifying a reason in case of errors is not necessary. If an error occurs, the reason is discarded as per https://github.com/kubernetes/kubernetes/blob/00aae4c10c204d258e62a96f30970e9b91b29149/staging/src/k8s.io/apiserver/pkg/endpoints/filters/authorization.go#L67-L71
- We have to be careful of anonymizing here, so using the authorization decorator https://github.com/kcp-dev/kcp/blob/ffc21ea3df5d5ce692d9a05665c690fd483ebe95/pkg/authorization/decorator.go#L48 to anonymize is the way to go.

However, it's ok to address this as a follow-up as this PR introduces caching of delegated authorizers only.

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
s-urbaniak: As commented above, `reason` can be left empty string, however the error should be anonmyized.

s-urbaniak:
As this is caching delegated authorizers and not authorize decisions per se I would be rather explicit about this in the godocs by saying that this caches authorizers per requested logical cluster and each cached authorizer additionally caches "allow" results (TTL being 5 minutes) and "deny" results (TTL being 30 seconds) per requested attributes. So it is rather a "cache of caches".

Additionally I suggest to add an additional constructor in the delegated authorizer and have the TTLs configurable so we know the exact timings here rather than on relying on central defaults (https://github.com/kcp-dev/kcp/blob/8d95d652dd230a8b4a61598a9c7b77b32bf7972d/pkg/authorization/delegated/authorizer.go#L38-L39) so we have local overall predictable cache behavior here in the constructor.

pkg/virtual/initializingworkspaces/builder/caching_authorizer.go
s-urbaniak: Personal preference is to go with b) so we can indeed reuse this in other places 👍

pkg/authorization/delegated/caching_authorizer.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return authorizer.DecisionNoOpinion, "error", err
+	}
+
+	SARAttributes := authorizer.AttributesRecord{
+		APIGroup:        tenancyv1alpha1.SchemeGroupVersion.Group,
+		APIVersion:      tenancyv1alpha1.SchemeGroupVersion.Version,
+		User:            attr.GetUser(),
+		Name:            name,
+		ResourceRequest: c.opts.Resource != "",
```
vincepri, line 128: Is this reasonable to do? @s-urbaniak

vincepri: In alternative, we could just expose this in the Options?

s-urbaniak: @vincepri thinking 🤔 how about, for reuse, we leave the cache simply being a cache and not let it implement the `Authorize` k8s interface. That way consumers of the cached authorizer can refactor to use the cache in a minimal way. Instead of instantiating using `NewDelegatedAuthorizer`, they simply invoke a `GetDelegatedAuthorizer(clusterName)` method on the cache and leave existing logic as-is. Here is a commit where I did the refactoring to reduce the amount of comments here: https://github.com/kcp-dev/kcp/commit/99fbf479a9f96ab023306c7602bef1b5f698cb52

Using this the diff in the initiliazingworkspaces builder is minimal and boils down to:
```

+       cachingAuthorizer := newAuthorizer(kubeClusterClient, delegated.NewDelegatedAuthorizerCache(kubeClusterClient, delegated.CachingOptions{}))

...

-func newAuthorizer(client kcpkubernetesclientset.ClusterInterface) authorizer.AuthorizerFunc {
+func newAuthorizer(client kcpkubernetesclientset.ClusterInterface, authorizerCache *delegated.DelegatedAuthorizerCache) authorizer.AuthorizerFunc {
        return func(ctx context.Context, attr authorizer.Attributes) (authorizer.Decision, string, error) {
                clusterName, name, err := initialization.TypeFrom(corev1alpha1.LogicalClusterInitializer(dynamiccontext.APIDomainKeyFrom(ctx)))
                if err != nil {
                        return authorizer.DecisionNoOpinion, "unable to determine initializer", fmt.Errorf("access not permitted")
                }

-               authz, err := delegated.NewDelegatedAuthorizer(clusterName, client)
+               authz, err := authorizerCache.GetDelegatedAuthorizer(clusterName)
                if err != nil {
                        return authorizer.DecisionNoOpinion, "error", err
                }
```

Raison d'etre:
- Authorizers derive the cluster name using different ways, i.e. some use a specific context (like here), some derive it from the request path
- The subject access review attributes are very different between authorizers

WDYT?

pkg/authorization/delegated/caching_authorizer.go
```diff
@@ -0,0 +1,104 @@
...
+	cache *cache.Expiring
+
+	auth   CachingAuthorizerFunc
+	client kcpkubernetesclientset.ClusterInterface
+}
+
+// load loads the authorizer from the cache, if any.
+func (c *cachingAuthorizer) load(clusterName logicalcluster.Name) authorizer.Authorizer {
+	value, ok := c.cache.Get(clusterName)
+	if !ok && value == nil {
```
ncdc, line 76: `||`?

pkg/authorization/delegated/caching_authorizer.go
```diff
@@ -0,0 +1,104 @@
...
+	"time"
+
+	kcpkubernetesclientset "github.com/kcp-dev/client-go/kubernetes"
+	"github.com/kcp-dev/logicalcluster/v3"
+
+	"k8s.io/apimachinery/pkg/util/cache"
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+)
+
+type CachingOptions struct {
```
ncdc, line 30: Please add godoc

ncdc (approved):

PR #2485, opened 2022-12-13
🌱 Add per-cluster static endpoint filter/handler

### LINKED ISSUE #2036: feature: make /readyz /live /healthz /version available in every workspace
### Feature Description

The mentioned endpoints should return the same status as the global one.

### Proposed Solution

Redirect those endpoints under a workspace URL to the top-level one. This could implemented via a function `WithPerWorkspaceStaticEndpoints` in https://github.com/kcp-dev/kcp/blob/main/pkg/server/handler.go, and wired into the handler chain.

### Alternative Solutions

_No response_

### Want to contribute?

### Additional Context

Note: /openapi and /metrics are different and need more thoughts.

vincepri:
## Summary

Adds support to handle well-known endpoints like
/healthz /livez /readyz and /version.

The requests is rewritten to point to the root path instead of a cluster
specific one.

## Related issue(s)

Fixes #2036


vincepri:
/hold

This might actually not be needed at all, the rewriting is already handled in `filters.WithScopeCluster` — The request seems to just not go through 🤔

PR #2512, opened 2022-12-19
✨ Use distroless/static:debug as base image

### LINKED ISSUE #798: Change base image to e.g. distroless
I'm not sure if things have changed recently, but at least in the past, alpine builds were not consistently reproducible. Additionally, we should keep our attack surface to a minimum by using a base image with the least amount of extra stuff in it, so something like distroless is a better choice (and consistent with Kubernetes).

vincepri:

## Summary

This PR also adds support for better caching when building layers

## Related issue(s)

Fixes #798

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,5 @@
+# syntax=docker/dockerfile:1.4
+
 # Copyright 2022 The KCP Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,3 +1,5 @@
+# syntax=docker/dockerfile:1.4
```
vincepri, line 1: This is for completeness and to make sure clients don't assume the version

Dockerfile
```diff
@@ -14,48 +16,49 @@
...
 # Build the binary
 FROM golang:1.19 AS builder
-
 WORKDIR /workspace
+
+# Install dependencies.
+RUN apt-get update && apt-get install -y jq && mkdir bin
+
+# Install kubectl.
+RUN wget "https://dl.k8s.io/release/$(wget -q https://dl.k8s.io/release/stable.txt -O-)/bin/linux/$(uname -m | sed 's/aarch.*/arm64/;s/armv8.*/arm64/;s/x86_64/amd64/')/kubectl" -O bin/kubectl && chmod +x bin/kubectl
```
stevekuznetsov, line 25: Should we try to match the version we have in `go.mod`?

vincepri: Like the one specified in `k8s.io/kubernetes v1.24.3`? We can, although not absolutely necessary given that clients _should_ work across a number of Kubernetes versions. We can probably pass a Dockerfile arg to specify the version?

Dockerfile
```diff
@@ -14,48 +16,50 @@
...
 # Build the binary
 FROM golang:1.19 AS builder
-
 WORKDIR /workspace
+
+# Install dependencies.
+RUN apt-get update && apt-get install -y jq && mkdir bin
+
+# Install kubectl.
+ARG kubernetes_version=1.24.2
```
stevekuznetsov, line 25: (as a follow-up) is it possible to run `go mod` to get this value dynamically?

stevekuznetsov (approved):

Dockerfile
vincepri: Yeah we can do that with grep and whatnot, haven't found a clean way to set it

stevekuznetsov:
```
$ go list -m -json k8s.io/kubernetes | jq .Version
"v1.24.3"
```

PR #2469, opened 2022-12-09
:sparkles: Partition reconciliation

fgiloux:
## Summary

This PR implements a reconciliation mechanism so that the endpoints listed in the status of an APIExportEndpointSlice match the filtering provided by the specified partition.

[Design document](https://hackmd.io/EuPmah62QHSEs-po04Kn1Q?view)

## Related issue(s)

Follows #2432
Fixes #2333

PR #2513, opened 2022-12-20
:sparkles: Partitionset reconciliation

### LINKED ISSUE #2334: feature: PartitionSet reconciliation
### Feature Description

This feature request follows https://github.com/kcp-dev/kcp/issues/2331

Its aim is the implementation of a reconciliation mechanism for PartitionSet. As part of this mechanism Partitions get created to match the dimensions and the target domain defined in the spec of PartitionSet.

A Partition is a subset of the kcp shards.
The shard topology is configured through the labels of the ClusterWorkspaceShards. 

### Proposed Solution

[API design](https://hackmd.io/EuPmah62QHSEs-po04Kn1Q?view)

### Alternative Solutions

Alternatives have been discussed in this [Google doc](https://docs.google.com/document/d/1zI9KOpLp6H_tAcKpGbLpQu_iUK9VbpgUSo0OFEd7Mig/edit#heading=h.7cfr9xt3gzbw)

### Want to contribute?

- [X] I would like to work on this issue.

### Additional Context

_No response_

fgiloux:
## Summary

This PR implements a reconciliation mechanism so that shard partitions can be automatically created based on specified dimensions and selector.

[Design document](https://hackmd.io/EuPmah62QHSEs-po04Kn1Q?view)

## Related issue(s)

Follows #2469
Fixes #2334


pkg/apis/topology/v1alpha1/types_partitionset.go
```diff
@@ -58,16 +60,47 @@ type PartitionSetSpec struct {
...
 	// +optional
 
-	// selector (optional) is a label selector that filters shard targets.
-	Selector *metav1.LabelSelector `json:"selector,omitempty"`
+	// MatchExpressions (optional) specifies filtering for shard targets.
+	// MatchExpressions are used here instead of Selector to avoid possible collisions with
+	// the MatchLabels generated for the Partition according to the dimensions in the PartitionSet.
+	// This does not limit the filtering as any MatchLabel can also be expressed as
+	// a MatchExpression.
+	MatchExpressions []metav1.LabelSelectorRequirement `json:"matchExpressions,omitempty"`
```
p0lyn0mial, line 68: Would it make sense to make `PartitionSetSpec.Dimensions` and/or `PartitionSetSpec.matchExpressions/PartitionSetSpec.selector` immutable ?

p0lyn0mial (commented):
I did a first pass, for now only over the API. Let me know what you think.

pkg/apis/topology/v1alpha1/types_partitionset.go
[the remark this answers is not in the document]
p0lyn0mial: Also could we rename it to `shardSelector/shardMatchExpression` ?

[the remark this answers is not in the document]
fgiloux: as per the comment " MatchExpressions are used here instead of Selector to avoid possible collisions with the MatchLabels generated for the Partition".
PartitionSets are used to generate Partitions. Partitions have a LabelSelector. LabelSelectors have two parts:
-  MatchLabels
- MatchExpressions

MatchLabels of Partitions are generated from the Dimension field of PartitionSet. MatchExpressions of Partitions are copied from MatchExpressions of PartitionSet. Keeping  this separation provides a clearer UX. It would be more difficult for users to understand the generated Partitions from a PartitionSet if MatchLabels could be a mix of  the Dimension generation and MatchLabels inherited from the PartitionSet. As per the comment there is no trade-off in terms of capabilities: any MatchLabels can be expressed by MatchExpressions.

fgiloux: As I see it this is not the likely flow.  This does not take in considerations that Partitions need to be colocated with the APIExportEndpointSlices that consume them. Here is how I see the flow:
- A service provider update a PartitionSet
- The set of Partitions bound to the PartitionSet is updated accordingly in the PartitionSet workspace
- Each of this Partition is copied into the workspace where the intended APIExportEndpointSlice is located. Each APIExportEndpointSlice is intended for an operator/controller manager instance. It makes URLs for a set of shards identified by the Partition available to it.
- The service provider can choose at any convenient time to patch the APIExportEndpointSlices with the new Partitions
There is no modification of the service and no disruption until the APIExportEndpointSlices are patched.

The steps above could be automated but it is another discussion.

pkg/reconciler/topology/partitioning_test.go
```diff
@@ -0,0 +1,125 @@
+/*
+Copyright 2022 The KCP Authors.
```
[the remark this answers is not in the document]
fgiloux: The PR and commit were created in 2022

pkg/apis/topology/v1alpha1/types_partitionset.go
p0lyn0mial: I'm not sure if I read the code correctly but it seems that we are deleting old partitions potentially being used by APIExportEndpointSlice which seems to be disruptive.

xref: https://github.com/kcp-dev/kcp/blob/c991948b1e6f7b14152b0a2714930b1b1c42b4b5/pkg/reconciler/topology/partitionset_reconcile.go#L123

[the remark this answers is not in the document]
p0lyn0mial: > as per the comment " MatchExpressions are used here instead of Selector to avoid possible collisions with the MatchLabels generated for the Partition".

if we have a collision, does it mean we can achieve the same by providing a selector on a `Partition` resource :) ?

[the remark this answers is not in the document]
fgiloux: If I understand your point correctly PartitionSet is just a convenient way to create Partitions. Service Providers are free to create Partitions the way it best suits them: using PartitionSet or directly creating the resource. If they do it manually they can set Selectors: MatchLabels and MatchExpressions in Partitions in a similar way to what is generated by the controller.

pkg/reconciler/topology/resources.go
```diff
@@ -0,0 +1,39 @@
...
+
+	topologyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/topology/v1alpha1"
+)
+
+// generatePartition generates the Partition specifications based on
+// the provided matchExpressions and matchLabels.
+func generatePartition(name string, matchExpressions []metav1.LabelSelectorRequirement, matchLabels map[string]string) *topologyv1alpha1.Partition {
+	return &topologyv1alpha1.Partition{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: name + "-",
```
p0lyn0mial, line 30: It appears to me that providers will have to manually introspect generated `Partitions` to make use of them because names appear random and there is no explanation of the generated selector, am I right?

Wondering if we could/should improve this?

fgiloux: Names are generated as Partitions are. We could create deterministic names by hashing the Partition spec but I am not sure it would be reasonable. An annotation may be better suited to store the hash. This would be stable as the PartitionSet reconciliation only does Partition create/delete, no update.
That said I am not sure it would be a significant improvement:
- Shard topology should be stable, e.g. you may add new shards in a region time to time but adding a new region is a rare event that would need action from Service Providers for them to cover it. It would also mean just a few additional Partitions to be taken care of. Deletion is not much different.
- PartitionSet change is driven by Service Provider and it means that they want to change the distribution of their service deployments. In that case they can check the deltas between the deployed Partitions and the newly generated one, even if a random name is used for Partitions. Some tooling on the client side may help with that.
There may be bits in the workflow that can be improved and I am open to proposals. My current thinking is that this could get better addressed when we develop some client side tooling that automates service providers deployments and topology changes, possibly taking a PartitionSet and an APIExport as parameter. I am a bit wary with having this automation on the server side because:
- that could drive automated changes that introduce big disruption and cost increases by service providers, so that it would need to be gated, hence may not bring much gain compared to a client-side approach.
- I am not sure that cross-shard communication will always be possible.

pkg/reconciler/topology/resources.go
```diff
@@ -22,12 +22,20 @@ import (
 	topologyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/topology/v1alpha1"
 )
 
+const dimensionsLabelKeyBase = "partition.kcp.io/"
```
sttts, line 25: `partition.topology.kcp.io/`

fgiloux:
@sttts @p0lyn0mial I have implemented the changes we discussed:
- using the dimension values for Partition names (still using name generator taking care of overflows)
- removed the labels with the dimensions (as the information is now made visible by the name)
- using a selector in PartitionSet rather than matchExpressions only (some collision check was needed)

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,278 @@
...
+	for _, dimension := range dimensions {
+		if _, ok := selector.MatchLabels[dimension]; ok {
+			return false, fmt.Errorf("MatchLabel with key %s conflicts with a dimension, please use matchExpressions instead", dimension)
+		}
+	}
+	return true, nil
+}
+
+func (c *controller) deletePartitions(ctx context.Context, partitionSet *topologyv1alpha1.PartitionSet) error {
+	partitionSet.Status.Count = 0
```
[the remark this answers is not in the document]
p0lyn0mial: oh, it tracks the number of associated `partitions`.
should we then decrease this number by the number of successful calls to `deletePartition` ?

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,278 @@
...
+			topologyv1alpha1.PartitionsReady,
+			topologyv1alpha1.ErrorGeneratingPartitionsReason,
+			conditionsv1alpha1.ConditionSeverityError,
+			"unable to retrieve associated Partitions",
+		)
+		return err
+	}
+
+	var matchLabelsMap map[string]map[string]string
+	if partitionSet.Spec.ShardSelector != nil {
```
p0lyn0mial, line 128: it looks like we could get rid of this `if` statement since the `partition` function can already handle nil `shardSelectorLabels`

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,278 @@
...
+			}
+		}
+	}
+	conditions.MarkTrue(partitionSet, topologyv1alpha1.PartitionsReady)
+	return nil
+}
+
+// partition populates shard label selectors according to dimensions.
+// It only keeps selectors that have at least one match.
+func partition(shards []*corev1alpha1.Shard, dimensions []string, shardSelectorLabels map[string]string) (matchLabelsMap map[string]map[string]string) {
```
p0lyn0mial, line 220: would it make sense to rename to `partitionByDimensionsAndLabels`?

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+	// Sorting for consistent comparison.
+	sort.Strings(labels)
+	for _, shard := range shards {
+		key := ""
+		selector := make(map[string]string)
+		for _, label := range labels {
+			labelValue, ok := shard.Labels[label]
+			if !ok {
+				break
```
p0lyn0mial, line 235: what does it mean?
what if we didn't find let's say the second label out of four? (`len(key) > 0` will apply)

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,278 @@
...
+		if !equality.Semantic.DeepEqual(oldMatchExpressions, newMatchExpressions) {
+			if err := c.deletePartition(ctx, logicalcluster.From(oldPartition).Path(), oldPartition.Name); err != nil && !apierrors.IsNotFound(err) {
+				conditions.MarkFalse(
+					partitionSet,
+					topologyv1alpha1.PartitionsReady,
+					topologyv1alpha1.ErrorGeneratingPartitionsReason,
+					conditionsv1alpha1.ConditionSeverityError,
+					"old partition could not get deleted",
+				)
+				return err
```
p0lyn0mial, line 155: again, since this is a network call, we could be more greedy and try all `oldPartitions`, wdyt?

pkg/server/server.go
```diff
@@ -453,6 +453,12 @@ func (s *Server) Run(ctx context.Context) error {
 		}
 	}
 
+	if s.Options.Controllers.EnableAll || enabled.Has("partition") {
```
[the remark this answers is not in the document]
fgiloux: I would be ok with changing to partitionset. I used partition as we don't have any other controller specific to Partitions. Partitions alone don't get reconciled. It is only in the context of an APIExportEndpointSlice that the controller takes the referenced Partition in consideration for populating  the endpoints. The same mechanism could also be leveraged in other places in the future.

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,278 @@
...
+	sort.Strings(labels)
+	for _, shard := range shards {
+		key := ""
+		selector := make(map[string]string)
+		for _, label := range labels {
+			labelValue, ok := shard.Labels[label]
+			if !ok {
+				break
+			}
+			key = key + "+" + label + "=" + labelValue
```
p0lyn0mial, line 237: i'm not sure if i understand why the `key` has to collect all labels with their values, couldn't we calculate some `hash` from the labels and values ?

pkg/reconciler/topology/partitionset_controller.go
```diff
@@ -0,0 +1,346 @@
...
+
+// enqueuePartition maps a Partition to a PartitionSet for enqueuing.
+func (c *controller) enqueuePartition(obj interface{}) {
+	key, err := kcpcache.DeletionHandlingMetaClusterNamespaceKeyFunc(obj)
+	if err != nil {
+		runtime.HandleError(err)
+		return
+	}
+
+	logger := logging.WithObject(logging.WithReconciler(klog.Background(), ControllerName), obj.(*topologyv1alpha1.Partition))
```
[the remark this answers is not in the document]
fgiloux: done

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,278 @@
...
+			_, err = c.createPartition(ctx, logicalcluster.From(partitionSet).Path(), partition)
+			if err != nil && !apierrors.IsAlreadyExists(err) {
+				conditions.MarkFalse(
+					partitionSet,
+					topologyv1alpha1.PartitionsReady,
+					topologyv1alpha1.ErrorGeneratingPartitionsReason,
+					conditionsv1alpha1.ConditionSeverityError,
+					"partition could not get created",
+				)
+				return err
```
p0lyn0mial, line 210: since this is a network call, we could be more greedy and try to create all `partitions`

pkg/reconciler/topology/partitionset_indexes.go
```diff
@@ -0,0 +1,50 @@
...
+func indexPartitionsByPartitionSetFunc(obj interface{}) ([]string, error) {
+	partition, ok := obj.(*topologyv1alpha1.Partition)
+	if !ok {
+		return []string{}, fmt.Errorf("obj is supposed to be a Partition, but is %T", obj)
+	}
+
+	for _, ownerRef := range partition.OwnerReferences {
+		if ownerRef.Kind == "PartitionSet" {
+			path := logicalcluster.From(partition).Path()
+			if !ok {
```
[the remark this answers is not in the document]
fgiloux: done

p0lyn0mial (commented):
I did a first pass, let's see if we can make move some parts into an admission plugin. I think that would simplify the controller. Let me know what you think, thanks!

pkg/reconciler/topology/partitionset_controller.go
```diff
@@ -0,0 +1,346 @@
...
+func filterShardEvent(oldObj, newObj interface{}) bool {
+	oldShard, ok := oldObj.(*corev1alpha1.Shard)
+	if !ok {
+		return false
+	}
+	newShard, ok := newObj.(*corev1alpha1.Shard)
+	if !ok {
+		return false
+	}
+	if !reflect.DeepEqual(oldShard.Labels, newShard.Labels) {
```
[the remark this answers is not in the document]
fgiloux: return !reflect.DeepEqual(oldShard.Labels, newShard.Labels)
done

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,278 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+
+	corev1alpha1 "github.com/kcp-dev/kcp/pkg/apis/core/v1alpha1"
+	conditionsv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/apis/conditions/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/util/conditions"
+	topologyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/topology/v1alpha1"
+)
+
+func (c *controller) reconcile(ctx context.Context, partitionSet *topologyv1alpha1.PartitionSet) error {
+	if ok, err := isSelectorWithoutConflict(partitionSet.Spec.ShardSelector, partitionSet.Spec.Dimensions); !ok {
```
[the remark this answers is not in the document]
fgiloux: We could. I found the logic to complicate for CEL and I tend to avoid admission plugins for non-security related checks but yes I could do that if you think that it improves the UX.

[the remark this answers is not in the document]
p0lyn0mial: we could move it either to CEL or an admission plugin, it seems like we could also move https://github.com/kcp-dev/kcp/pull/2513#discussion_r1091111068. 
I think it would simply this controller and would avoid deleting `partitions`

@sttts wdyt?

pkg/reconciler/topology/partitionset_reconcile.go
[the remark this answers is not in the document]
fgiloux: `deletePartitions` (mind the `s` at the end) is only call in case of an issue with the selector (lines 55 and 83) the count should then be set back to 0. In case of regular deletion because of a change in shards `deletePartition`without `s` is called (lines 147 and 182). In case of no error with the selector the number of endpoints is already captured in the size of `matchLabelsMap`, no need for additional counting.
I will move `partitionSet.Status.Count = 0`one level up as it is not clear from the name of the function that it does that as well.

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,278 @@
...
+	conditions.MarkTrue(partitionSet, topologyv1alpha1.PartitionSetValid)
+
+	shards, err := c.listShards(labelSelector)
+	if err != nil {
+		conditions.MarkFalse(
+			partitionSet,
+			topologyv1alpha1.PartitionsReady,
+			topologyv1alpha1.ErrorGeneratingPartitionsReason,
+			conditionsv1alpha1.ConditionSeverityError,
+			"error listing Shards",
```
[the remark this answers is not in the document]
fgiloux: The error is returned by the reconcile function, hence logged by the controller

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+		// No need to requeue if the selector conflicts
+		return nil
+	}
+	var labelSelector labels.Selector
+	if partitionSet.Spec.ShardSelector == nil {
+		labelSelector = labels.Everything()
+	} else {
+		var err error
+		labelSelector, err = metav1.LabelSelectorAsSelector(partitionSet.Spec.ShardSelector)
```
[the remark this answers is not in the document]
fgiloux: see above. We can discuss it on slack

pkg/reconciler/topology/partitionset_reconcile.go
fgiloux: I cannot reference `MatchLabels` on a nil pointer

pkg/reconciler/topology/partitionset_reconcile.go
fgiloux: I don't see the need for now. There is currently a single way of partitioning. `Labels`could also be ambiguous in this context. `Dimensions`are a way to slice and dice by `Labels`. `Labels`are also used for filtering, defining the boundaries of the piece to slice and dice. What I mean is that `Dimensions`and `Labels`are not of the same nature for the partitioning.

pkg/reconciler/topology/partitionset_reconcile.go
fgiloux: true. Adding a boolean to check the match

pkg/reconciler/topology/partitionset_reconcile.go
fgiloux: This is a kind of hash. We could use a hash function for it. I am not sure that the processing time would be shorter and it was nice to have something readable by debugging.

pkg/reconciler/topology/partitionset_controller.go
```diff
@@ -0,0 +1,346 @@
...
+		listShards: func(selector labels.Selector) ([]*corev1alpha1.Shard, error) {
+			return globalShardClusterInformer.Lister().List(selector)
+		},
+		listPartitionSets: func() ([]*topologyv1alpha1.PartitionSet, error) {
+			return partitionSetClusterInformer.Lister().List(labels.Everything())
+		},
+		getPartitionSet: func(clusterName logicalcluster.Name, name string) (*topologyv1alpha1.PartitionSet, error) {
+			return partitionSetClusterInformer.Lister().Cluster(clusterName).Get(name)
+		},
+		getPartitionsByPartitionSet: func(key string) ([]*topologyv1alpha1.Partition, error) {
```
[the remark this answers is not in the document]
fgiloux: yes I can change the signature of the method but still we will need to calculate the key in the method to use the indexer, won't we?

pkg/reconciler/topology/partitionset_reconcile.go
fgiloux: I am not sure how you would do that here. There is no single selector for the Partitions to delete.

pkg/reconciler/topology/partitionset_reconcile.go
fgiloux: I am not sure how you would do that

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,278 @@
...
+	if err != nil {
+		return err
+	}
+	existingPartitions, err := c.getPartitionsByPartitionSet(key)
+	if err != nil {
+		return err
+	}
+	for _, partition := range existingPartitions {
+		if err := c.deletePartition(ctx, logicalcluster.From(partition).Path(), partition.Name); err != nil && !apierrors.IsNotFound(err) {
+			return err
```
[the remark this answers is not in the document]
p0lyn0mial: just loop over `existingPartitions`, try to `deletePartition`, collect any errors and return at the end :) ?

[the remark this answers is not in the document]
fgiloux: ok so it makes no difference in terms of network calls, does it?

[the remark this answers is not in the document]
p0lyn0mial: It does, we try all `existingPartitions` even if we failed to delete a `partition`.  Hopefully on the next iteration we will have less work do to. Does it make sense?

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,281 @@
...
+)
+
+func (c *controller) reconcile(ctx context.Context, partitionSet *topologyv1alpha1.PartitionSet) error {
+	if ok, err := isSelectorWithoutConflict(partitionSet.Spec.ShardSelector, partitionSet.Spec.Dimensions); !ok {
+		conditions.MarkFalse(
+			partitionSet,
+			topologyv1alpha1.PartitionSetValid,
+			topologyv1alpha1.PartitionSetInvalidSelectorReason,
+			conditionsv1alpha1.ConditionSeverityError,
+			err.Error(),
```
sttts, line 45: this needs context

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,281 @@
...
+		)
+		conditions.MarkFalse(
+			partitionSet,
+			topologyv1alpha1.PartitionsReady,
+			topologyv1alpha1.ErrorGeneratingPartitionsReason,
+			conditionsv1alpha1.ConditionSeverityError,
+			"",
```
sttts, line 52: this needs context

fgiloux:
@sttts @p0lyn0mial as discussed:
- I removed the check when matchLabels and dimensions overlap.
- I have added openapi/CEL validation for label selector.

ncdc:
#2665 
/retest

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,242 @@
...
+				return err
+			}
+		}
+	}
+	conditions.MarkTrue(partitionSet, topologyv1alpha1.PartitionsReady)
+	return nil
+}
+
+// partition populates shard label selectors according to dimensions.
+// It only keeps selectors that have at least one match.
```
p0lyn0mial, line 197: with https://github.com/kcp-dev/kcp/pull/2513/files#r1091643038 this comment is not true, correct?

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,242 @@
...
+func partition(shards []*corev1alpha1.Shard, dimensions []string, shardSelectorLabels map[string]string) (matchLabelsMap map[string]map[string]string) {
+	matchLabelsMap = make(map[string]map[string]string)
+	labels := make([]string, len(dimensions), len(dimensions)+len(shardSelectorLabels))
+	copy(labels, dimensions)
+	for label := range shardSelectorLabels {
+		labels = append(labels, label)
```
p0lyn0mial, line 203: shouldn't we also take values into account ?

pkg/reconciler/topology/partitionset_reconcile.go
p0lyn0mial: if we won't reduce the size of the key it might grow indefinitely, no?

pkg/reconciler/topology/resources.go
```diff
@@ -0,0 +1,49 @@
...
+
+// generatePartition generates the Partition specifications based on
+// the provided matchExpressions and matchLabels.
+func generatePartition(name string, matchExpressions []metav1.LabelSelectorRequirement, matchLabels map[string]string, dimensions []string) *topologyv1alpha1.Partition {
+	pname := name
+	labels := make([]string, len(dimensions))
+	copy(labels, dimensions)
+	sort.Strings(labels)
+	for _, label := range labels {
+		pname = pname + "-" + matchLabels[label]
```
p0lyn0mial, line 35: are there any name length limit ?
what if we hit one ?

fgiloux: there are name limits. GenerateName takes care of it

pkg/reconciler/topology/partitionset_reconcile.go
fgiloux: If I am following properly: The Shards passed to the `partition` function is a subset of all the Shards that comply with the Selector configured on PartitionSet.

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,209 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+
+	corev1alpha1 "github.com/kcp-dev/kcp/pkg/apis/core/v1alpha1"
+	conditionsv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/apis/conditions/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/util/conditions"
+	topologyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/topology/v1alpha1"
+)
+
+func (c *controller) reconcile(ctx context.Context, partitionSet *topologyv1alpha1.PartitionSet) error {
+	var labelSelector labels.Selector
```
p0lyn0mial, line 37: nit: we could get rid of the `else` clause by doing
```
labelSelector := labels.Everything()
	if partitionSet.Spec.ShardSelector != nil {...}
```

pkg/reconciler/topology/partitionset_reconcile.go
fgiloux: I will make the comment clearer, what I meant: "It only keeps selectors that have at least one Shard matching them so that Partition not referring to any Shards should not get created"

pkg/reconciler/topology/partitionset_reconcile.go
fgiloux: The keys are only used as a transient and quick way of comparing the desired state of Partitions with the existing ones. They are not stored.

pkg/reconciler/topology/partitionset_reconcile.go
```diff
@@ -0,0 +1,209 @@
...
+	if partitionSet.Spec.ShardSelector != nil {
+		newMatchExpressions = partitionSet.Spec.ShardSelector.MatchExpressions
+	}
+	for _, oldPartition := range oldPartitions {
+		// MatchExpressions need to be the same
+		oldMatchExpressions := []metav1.LabelSelectorRequirement{}
+		if oldPartition.Spec.Selector != nil {
+			oldMatchExpressions = oldPartition.Spec.Selector.MatchExpressions
+		}
+		if !equality.Semantic.DeepEqual(oldMatchExpressions, newMatchExpressions) {
```
p0lyn0mial, line 108: would it make sense to change this code so that there is only one path that deletes a partition ? i.e.
```
if  !matchExpressions || !partitionKey {
 delete...
}
```

pkg/reconciler/topology/partitionset_reconcile.go
fgiloux: changed

pkg/reconciler/topology/resources.go
p0lyn0mial: ok, thx.

pkg/reconciler/topology/partitionset_reconcile.go
p0lyn0mial: okay, so the values will be taken into okay just on different level, thx.

p0lyn0mial:
@fgiloux I'd also put the new controller into a new directory i.e. `topology/partitionset/` I think atm it is under `topology/`

pkg/reconciler/topology/partitionset_reconcile.go
fgiloux: `deletePartition` and `conditions.MarkFalse`are two lines. If I understand your point correctly, in term of readability I prefer to have the action next to the evaluation rather than differing it.

config/crds/topology.kcp.io_partitionsets.yaml
```diff
@@ -66,11 +65,26 @@ spec:
                         key:
                           description: key is the label key that the selector applies
                             to.
+                          maxLength: 317
                           type: string
+                          x-kubernetes-validations:
+                          - message: Label key prefixes are optional and limited to
```
p0lyn0mial, line 71: I have no way of verifying this validation, I think we will validate it in an e2e test.

fgiloux:
> @fgiloux I'd also put the new controller into a new directory i.e. `topology/partitionset/` I think atm it is under `topology/`

I can do that. It is currently the only controller in topology

config/crds/topology.kcp.io_partitionsets.yaml
fgiloux: yes I will add checks for the admission to the e2e PR

config/root-phase0/apiresourceschema-partitionsets.topology.kcp.io.yaml
```diff
@@ -47,8 +47,8 @@ spec:
               items:
                 type: string
               type: array
-            selector:
-              description: selector (optional) is a label selector that filters shard
+            shardSelector:
+              description: ShardSelector (optional) specifies filtering for shard
```
sttts, line 51: lower-case

pkg/apis/topology/v1alpha1/types_partitionset.go
```diff
@@ -58,16 +60,43 @@ type PartitionSetSpec struct {
 
 	// +optional
 
-	// selector (optional) is a label selector that filters shard targets.
-	Selector *metav1.LabelSelector `json:"selector,omitempty"`
+	// ShardSelector (optional) specifies filtering for shard targets.
```
sttts, line 63: lower-case

pkg/reconciler/topology/partitionset/partitionset_controller.go
```diff
@@ -0,0 +1,346 @@
...
+	topologyv1alpha1client "github.com/kcp-dev/kcp/pkg/client/clientset/versioned/typed/topology/v1alpha1"
+	coreinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/core/v1alpha1"
+	topologyinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions/topology/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/indexers"
+	"github.com/kcp-dev/kcp/pkg/logging"
+	"github.com/kcp-dev/kcp/pkg/reconciler/committer"
+)
+
+const (
+	ControllerName = "kcp-partitionset"
```
sttts, line 50: kcp-topology-partitionset

pkg/reconciler/topology/partitionset/partitionset_controller.go
```diff
@@ -0,0 +1,346 @@
...
+		},
+		getPartitionSet: func(clusterName logicalcluster.Name, name string) (*topologyv1alpha1.PartitionSet, error) {
+			return partitionSetClusterInformer.Lister().Cluster(clusterName).Get(name)
+		},
+		getPartitionsByPartitionSet: func(partitionSet *topologyv1alpha1.PartitionSet) ([]*topologyv1alpha1.Partition, error) {
+			key, err := kcpcache.MetaClusterNamespaceKeyFunc(partitionSet)
+			if err != nil {
+				return nil, err
+			}
+			list, err := partitionClusterInformer.Informer().GetIndexer().ByIndex(indexPartitionsByPartitionSet, key)
```
sttts, line 79: I think we have a helper somewhere for this code, using generics.

pkg/reconciler/topology/partitionset/partitionset_controller.go
```diff
@@ -0,0 +1,346 @@
...
+		commit: committer.NewCommitter[*PartitionSet, Patcher, *PartitionSetSpec, *PartitionSetStatus](kcpClusterClient.TopologyV1alpha1().PartitionSets()),
+	}
+
+	globalShardClusterInformer.Informer().AddEventHandler(
+		cache.ResourceEventHandlerFuncs{
+			AddFunc: func(obj interface{}) {
+				c.enqueueAllPartitionSets(obj)
+			},
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				if filterShardEvent(oldObj, newObj) {
```
sttts, line 109: explain in a comment what the filtering is and why it makes sense

pkg/reconciler/topology/partitionset/partitionset_controller.go
```diff
@@ -0,0 +1,346 @@
...
+			},
+			DeleteFunc: func(obj interface{}) {
+				c.enqueuePartition(obj)
+			},
+		},
+	)
+
+	indexers.AddIfNotPresentOrDie(partitionClusterInformer.Informer().GetIndexer(), cache.Indexers{
+		indexPartitionsByPartitionSet: indexPartitionsByPartitionSetFunc,
+	})
```
sttts, line 149: nit: indexes first, event handler after that.

pkg/reconciler/topology/partitionset/partitionset_controller.go
```diff
@@ -0,0 +1,346 @@
...
+type Patcher = topologyv1alpha1client.PartitionSetInterface
+type Resource = committer.Resource[*PartitionSetSpec, *PartitionSetStatus]
+type CommitFunc = func(context.Context, *Resource, *Resource) error
+
+// controller reconciles PartitionSets. It ensures that Partitions are aligned
+// with the PartitionSets specifications and theShards.
+type controller struct {
+	queue workqueue.RateLimitingInterface
+
+	// kcpClusterClient is cluster aware and used to communicate with kcp API servers
```
sttts, line 166: no need for the comment. This is convention.

pkg/reconciler/topology/partitionset/partitionset_controller.go
```diff
@@ -0,0 +1,346 @@
...
+) (*controller, error) {
+	queue := workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), ControllerName)
+
+	c := &controller{
+		queue:            queue,
+		kcpClusterClient: kcpClusterClient,
+		listShards: func(selector labels.Selector) ([]*corev1alpha1.Shard, error) {
+			return globalShardClusterInformer.Lister().List(selector)
+		},
+		listPartitionSets: func() ([]*topologyv1alpha1.PartitionSet, error) {
```
sttts, line 68: this looks suspicious. You should never need to list all partition sets.

sttts: Nvmd. I understand why you need it.

pkg/reconciler/topology/partitionset/partitionset_reconcile.go
```diff
@@ -0,0 +1,208 @@
...
+
+	corev1alpha1 "github.com/kcp-dev/kcp/pkg/apis/core/v1alpha1"
+	conditionsv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/apis/conditions/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/util/conditions"
+	topologyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/topology/v1alpha1"
+)
+
+func (c *controller) reconcile(ctx context.Context, partitionSet *topologyv1alpha1.PartitionSet) error {
+	labelSelector := labels.Everything()
+	if partitionSet.Spec.ShardSelector == nil {
```
sttts, line 38: != nil ?

pkg/reconciler/topology/partitionset/partitionset_reconcile.go
```diff
@@ -0,0 +1,208 @@
...
+
+	var matchLabelsMap map[string]map[string]string
+	if partitionSet.Spec.ShardSelector != nil {
+		matchLabelsMap = partition(shards, partitionSet.Spec.Dimensions, partitionSet.Spec.ShardSelector.MatchLabels)
+	} else {
+		matchLabelsMap = partition(shards, partitionSet.Spec.Dimensions, nil)
+	}
+	partitionSet.Status.Count = uint16(len(matchLabelsMap))
+	existingMatches := map[string]struct{}{}
+	var placeHolder struct{}
```
sttts, line 95: just use boolean type for the map

pkg/reconciler/topology/partitionset/partitionset_reconcile.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	var matchLabelsMap map[string]map[string]string
+	if partitionSet.Spec.ShardSelector != nil {
+		matchLabelsMap = partition(shards, partitionSet.Spec.Dimensions, partitionSet.Spec.ShardSelector.MatchLabels)
+	} else {
+		matchLabelsMap = partition(shards, partitionSet.Spec.Dimensions, nil)
+	}
+	partitionSet.Status.Count = uint16(len(matchLabelsMap))
+	existingMatches := map[string]struct{}{}
```
sttts, line 94: this can be made type-safe. Compare https://goplay.space/#IxSJ1_iWaY9.

pkg/reconciler/topology/partitionset/partitionset_reconcile.go
```diff
@@ -0,0 +1,208 @@
...
+			i++
+		}
+		sort.Strings(sortedOldKeys)
+		for _, key := range sortedOldKeys {
+			partitionKey = partitionKey + "+" + key + "=" + oldMatchLabels[key]
+		}
+		if _, ok := matchLabelsMap[partitionKey]; ok {
+			existingMatches[partitionKey] = placeHolder
+		} else {
+			if err := c.deletePartition(ctx, logicalcluster.From(oldPartition).Path(), oldPartition.Name); err != nil {
```
sttts, line 142: && !IsNotFound

pkg/reconciler/topology/partitionset/partitionset_reconcile.go
```diff
@@ -0,0 +1,208 @@
...
+					topologyv1alpha1.PartitionsReady,
+					topologyv1alpha1.ErrorGeneratingPartitionsReason,
+					conditionsv1alpha1.ConditionSeverityError,
+					"old partition could not get deleted",
+				)
+				return err
+			}
+		}
+	}
+	// Create partitions when no existing partition for the set has the same selector.
```
sttts, line 154: nit: empty line above

pkg/server/server.go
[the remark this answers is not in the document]
sttts: we should honestly remove all these ifs. We don't use them anywhere.

pkg/reconciler/topology/partitionset/partitionset_reconcile.go
sttts, line 142: We usually put `logger.V(2)` before every client call

pkg/reconciler/topology/partitionset/partitionset_reconcile.go
```diff
@@ -0,0 +1,208 @@
...
+	for key, matchLabels := range matchLabelsMap {
+		if _, ok := existingMatches[key]; !ok {
+			partition := generatePartition(partitionSet.Name, newMatchExpressions, matchLabels, partitionSet.Spec.Dimensions)
+			partition.OwnerReferences = []metav1.OwnerReference{
+				*metav1.NewControllerRef(partitionSet, topologyv1alpha1.SchemeGroupVersion.WithKind("PartitionSet")),
+			}
+			_, err = c.createPartition(ctx, logicalcluster.From(partitionSet).Path(), partition)
```
sttts, line 161: logger output

pkg/reconciler/topology/partitionset/partitionset_reconcile.go
```diff
@@ -0,0 +1,208 @@
...
+// so that Partitions not referring to any Shard would not get created.
+func partition(shards []*corev1alpha1.Shard, dimensions []string, shardSelectorLabels map[string]string) (matchLabelsMap map[string]map[string]string) {
+	matchLabelsMap = make(map[string]map[string]string)
+	labels := make([]string, len(dimensions), len(dimensions)+len(shardSelectorLabels))
+	copy(labels, dimensions)
+	for label := range shardSelectorLabels {
+		labels = append(labels, label)
+	}
+	// Sorting for consistent comparison.
+	sort.Strings(labels)
```
sttts, line 189:
```
=== empty line ===
// Sorting for consistent comparison.
sort.Strings(labels)
=== empty line ===
```
or 
```
sort.Strings(labels) // Sorting for consistent comparison.
```

pkg/reconciler/topology/partitionset/resources.go
```diff
@@ -0,0 +1,49 @@
+/*
+Copyright 2022 The KCP Authors.
```
sttts, line 2: 2023

sttts: and everywhere else

config/crds/topology.kcp.io_partitionsets.yaml-patch
```diff
@@ -0,0 +1,58 @@
...
+    - message: Label key prefixes are optional and limited to 253 characters. They should match "^[a-z0-9]([-a-z0-9]{0,251}[a-z0-9])?$". Names are limited to 63 characters and should match "^[A-Za-z0-9]([-A-Za-z0-9_.]{0,61}[A-Za-z0-9])?$". "/" is used as separator.
+      rule: 'self.contains("/") ? self.split("/", 2)[0].matches("^[a-z0-9]([-a-z0-9]{0,251}[a-z0-9])?$") && self.split("/", 2)[1].matches("^[A-Za-z0-9]([-A-Za-z0-9_.]{0,61}[A-Za-z0-9])?$") : self.matches("^[A-Za-z0-9]([-A-Za-z0-9_.]{0,61}[A-Za-z0-9])?$")'
+- op: add
+  path: /spec/versions/name=v1alpha1/schema/openAPIV3Schema/properties/spec/properties/shardSelector/properties/matchExpressions/items/properties/values/items/maxLength
+  value:
+    63
+- op: add
+  path: /spec/versions/name=v1alpha1/schema/openAPIV3Schema/properties/spec/properties/shardSelector/properties/matchExpressions/items/properties/values/items/pattern
+  value:
+    ^[A-Za-z0-9]([-A-Za-z0-9_.]{0,61}[A-Za-z0-9])?$
```
sttts, line 36: @vincepri label selector validation should come from controller-tools, shouldn't it?

pkg/reconciler/topology/partitionset/partitionset_reconcile.go
```diff
@@ -0,0 +1,208 @@
...
+		matchLabelsMap = partition(shards, partitionSet.Spec.Dimensions, nil)
+	}
+	partitionSet.Status.Count = uint16(len(matchLabelsMap))
+	existingMatches := map[string]struct{}{}
+	var placeHolder struct{}
+	newMatchExpressions := []metav1.LabelSelectorRequirement{}
+	if partitionSet.Spec.ShardSelector != nil {
+		newMatchExpressions = partitionSet.Spec.ShardSelector.MatchExpressions
+	}
+	for _, oldPartition := range oldPartitions {
```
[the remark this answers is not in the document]
fgiloux: added

pkg/reconciler/topology/partitionset/partitionset_reconcile.go
```diff
@@ -0,0 +1,208 @@
...
+		// MatchLabels need to be the same
+		partitionKey := ""
+		oldMatchLabels := map[string]string{}
+		if oldPartition.Spec.Selector != nil {
+			oldMatchLabels = oldPartition.Spec.Selector.MatchLabels
+		}
+
+		// Sorting the keys of the old partition for consistent comparison
+		sortedOldKeys := make([]string, len(oldMatchLabels))
+
```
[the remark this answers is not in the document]
fgiloux: removed

pkg/reconciler/topology/partitionset/partitionset_reconcile.go
fgiloux: added

fgiloux: added

pkg/reconciler/topology/partitionset/partitionset_reconcile.go
fgiloux: added

pkg/reconciler/topology/partitionset/partitionset_reconcile.go
fgiloux: added

pkg/reconciler/topology/partitionset/partitionset_reconcile.go
fgiloux: changed

pkg/reconciler/topology/partitionset/resources.go
fgiloux: I created the PR in 2022. Anyways I can change that if it matters

pkg/reconciler/topology/partitionset/partitionset_reconcile.go
sttts: I meant the key, i.e. that the key is a string but could be a struct. No need to think about representation.

p0lyn0mial:
/lgtm

thanks!

Issue #2501, opened 2022-12-15
flake: TestDeploymentCoordinator


vincepri:
### Describe the bug

```
2022-12-15T15:11:39.0852488Z --- FAIL: TestDeploymentCoordinator (70.09s)
2022-12-15T15:11:39.0855832Z     kcp.go:127: shared kcp server will target configuration "/home/runner/work/kcp/kcp/.kcp/admin.kubeconfig"
2022-12-15T15:11:39.0859071Z     workspaces.go:196: Created organization workspace root:e2e-org-znfgz
2022-12-15T15:11:39.0863338Z     workspaces.go:126: Created root:universal workspace root:e2e-org-znfgz:synctargets
2022-12-15T15:11:39.0866448Z     workspaces.go:126: Created root:universal workspace root:e2e-org-znfgz:workload-1
2022-12-15T15:11:39.0871331Z     workspaces.go:126: Created root:universal workspace root:e2e-org-znfgz:workload-2
2022-12-15T15:11:39.0873940Z     util.go:112: Saving test artifacts for test "TestDeploymentCoordinator" under "/tmp/e2e/TestDeploymentCoordinator/3181476435".
2022-12-15T15:11:39.0879096Z     syncer.go:141: Configuring workspace root:e2e-org-znfgz:synctargets for syncing
2022-12-15T15:11:39.0886386Z     kubectl.go:59: running: KUBECONFIG=/tmp/e2e/TestDeploymentCoordinator/3181476435/artifacts/root_e2e-org-znfgz_synctargets.kubeconfig kubectl kcp workload sync east --syncer-image=kind.local/syncer-c2e3073d5026a8f7f2c47a50c16bdbec:a4d77b220ce934e4e12fb443b04f413e518272ebd263465dca7285428a404ec4 --output-file=- --qps=-1 --feature-gates=CustomResourceValidationExpressions=true --api-import-poll-interval=5s --downstream-namespace-clean-delay=2s
2022-12-15T15:11:39.0887729Z     kubectl.go:89: running: kubectl --kubeconfig /home/runner/work/kcp/kcp/.kcp/kind.kubeconfig apply -f -
2022-12-15T15:11:39.0888663Z     syncer.go:589: checking secret kcp-syncer-east-2fu92b3q/kcp-syncer-east-2fu92b3q for annotation kubernetes.io/service-account.name=kcp-syncer-east-2fu92b3q
[...]

PR #2516, opened 2022-12-20
🌱 Add indexers.ByPathAndName helper

ncdc:
## Summary
Add indexers.ByPathAndName helper to reduce lots of copying 😄 

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/indexers/indexers.go
+++ b/pkg/indexers/indexers.go
@@ -22,8 +22,10 @@ import (
 
 	"github.com/kcp-dev/logicalcluster/v3"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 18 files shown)


pkg/indexers/indexers.go
```diff
@@ -124,3 +125,18 @@ func ByIndex[T runtime.Object](indexer cache.Indexer, indexName, indexValue stri
 
 	return ret, nil
 }
+
+// ByPathAndName returns the instance of T from the indexer with the matching path and name.
```
sttts, line 129: mention that this is partial and depends on the path annotation.

Also: it matches both a canonical path **and** the cluster name.

stevekuznetsov: Full path or any permutation?

ncdc:
Flake #2501

pkg/indexers/indexers.go
ncdc: Updated the comment; PTAL

stevekuznetsov (approved):
Seems fine - how do I know which types will have a working annotation?

PR #2519, opened 2022-12-20
🌱 Use deletion handling keyfunc where needed

ncdc:
## Summary
Switch places that potentially might get a cache.DeletedFinalStateUnknown to use the deletion handling key function variant.

## Related issue(s)

Fixes #


vincepri (approved):
/lgtm

PR #2542, opened 2023-01-04
🌱 Add e2e DNS tests

lionelvillard:
## Summary

Add e2e tests related to the kcp DNS service. Check names correctly resolves across namespaces in the same workspace and that they don't resolve across workspaces

## Related issue(s)

Fixes #

### DIFF
--- a/test/e2e/framework/workspaces.go
+++ b/test/e2e/framework/workspaces.go
@@ -18,18 +18,26 @@ package framework
 
 import (
 	"context"
+	"embed"
 	"fmt"
 	"strings"
 	"testing"
 	"time"
 
+	kcpdynamic "github.com/kcp-dev/client-go/dynamic"
 	"github.com/kcp-dev/logicalcluster/v3"
 	"github.com/stretchr/testify/require"
 
+	kcpapiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/kcp/clientset/versioned"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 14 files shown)


test/e2e/framework/workspaces.go
```diff
@@ -156,3 +164,21 @@ func NewOrganizationFixture(t *testing.T, server RunningServer, options ...Clust
 	org := NewOrganizationFixtureObject(t, server, options...)
 	return logicalcluster.Name(org.Status.Cluster)
 }
+
+// CreateResources creates all resources from a filesystem in the given workspace identified by upstreamConfig
+func CreateResources(t *testing.T, ctx context.Context, fs embed.FS, upstreamConfig *rest.Config, clusterName logicalcluster.Path, transformers ...helpers.TransformFileFunc) error {
```
sttts, line 169: this has not much to do with workspace, does it?

Please derive ctx from `t`.

davidfestal: `test/e2e/framework/util.go` ?

davidfestal: > Please derive ctx from t.

`ctx` is typically already derived from t at the test level: https://github.com/kcp-dev/kcp/blob/bbde21cffe43aee7b8f9e78019da6135ea86fb16/test/e2e/reconciler/deployment/deployment_coordinator_test.go#L51-L52

davidfestal: Maybe instead we should keep `ctx` and remove the `t` arg from here, and just return the error instead of using `require.NoError` inside this function.

lionelvillard: all done.

lionelvillard:
test lint

davidfestal (approved):

PR #2544, opened 2023-01-04
🐛 Docker build/push should not use a matrix

vincepri:
By using a matrix github actions spins up different VMs to build these images, upon pushing tags and the manifest, the different VMs have no information about the other builds, so each build context will build and push a manifest only for its own image.

## Related issue(s)

Fixes #

### DIFF
--- a/.github/workflows/kcp-image.yaml
+++ b/.github/workflows/kcp-image.yaml
@@ -16,24 +16,11 @@ jobs:
     if: github.repository_owner == 'kcp-dev'
     name: Build KCP Image
     runs-on: ubuntu-latest
-    outputs:
-      sha_short: ${{ env.sha_short }}
-    strategy:
-      matrix:
-        platforms:
-        - linux/amd64
-        - linux/arm64
-        - linux/ppc64le
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-go@v3
         with:
           go-version: v1.19
-
-      - name: Get the short sha
-        id: vars
-        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/kcp-image.yaml
```diff
@@ -60,14 +54,14 @@ jobs:
...
         with:
           context: .
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
-          platforms: ${{ matrix.platform }}
-          cache-from: type=gha,scope=$GITHUB_REF_NAME-{{ matrix.platform }}
-          cache-to: type=gha,scope=$GITHUB_REF_NAME-{{ matrix.platform }},mode=max
+          platforms: linux/amd64,linux/arm64,linux/ppc64le
+          cache-from: type=gha,scope=$GITHUB_REF_NAME-buildx
```
astefanutti, line 65: Should it be `${{ github.ref_name }}`?

.github/workflows/kcp-image.yaml
```diff
@@ -60,14 +54,14 @@ jobs:
             type=raw,value=${{ env.sha_short }}
```
astefanutti, line 54: Not in the direct scope, but maybe an opportunity to have that replaced with `type=sha,prefix=,suffix=,format=short` so the logic that computes / sets `sha_short` can be removed.

vincepri: done ptal

ncdc (approved):

PR #2545, opened 2023-01-04
🌱 Skip upsynced resources in resource scheduling

davidfestal:
## Summary

Resources upsynced from a downstream cluster, and thus having the `Upsync` label should not be taken in account by resource scheduling.

## Related issue(s)

No issue dedicated. Spotted while testing PR https://github.com/kcp-dev/kcp/pull/2338

### DIFF
--- a/pkg/reconciler/workload/resource/resource_reconcile.go
+++ b/pkg/reconciler/workload/resource/resource_reconcile.go
@@ -47,6 +47,11 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName logical
 	logger := logging.WithObject(logging.WithReconciler(klog.Background(), ControllerName), obj).WithValues("groupVersionResource", gvr.String(), "logicalCluster", lclusterName.String())
 	logger.V(4).Info("reconciling resource")
 
+	if isUpSynced(obj.GetLabels()) {
+		logger.V(4).Info("resource is in Upsync mode; ignoring")
+		return nil
+	}
+
 	// if the resource is a namespace, let's return early. nothing to do.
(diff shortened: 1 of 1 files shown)


pkg/reconciler/workload/resource/resource_reconcile.go
```diff
@@ -47,6 +47,11 @@ func (c *Controller) reconcileResource(ctx context.Context, lclusterName logical
 	logger := logging.WithObject(logging.WithReconciler(klog.Background(), ControllerName), obj).WithValues("groupVersionResource", gvr.String(), "logicalCluster", lclusterName.String())
 	logger.V(4).Info("reconciling resource")
 
+	if isUpSynced(obj.GetLabels()) {
+		logger.Info("resource is in Upsync mode; ignoring")
```
sttts, line 51: verbosity at least 3, probably more like 4.

pkg/reconciler/workload/resource/resource_controller.go
```diff
@@ -486,6 +486,17 @@ func (c *Controller) enqueueSyncTargetKey(syncTargetKey string) {
 	}
 }
 
+// isUpSynced returns true if the labels of the resource contain at least
+// one `ResourceState` label with the `Upsync` value.
+func isUpSynced(labels map[string]string) bool {
```
sttts, line 491: as a special helper only used in one stop, please move it into the other file.

davidfestal: fixed

pkg/reconciler/workload/resource/resource_reconcile.go
davidfestal: fixed

PR #2547, opened 2023-01-04
✨ Allow home workspaces to be accessed via `user:<username>`

vincepri:

## Related issue(s)

Fixes #


vincepri:
Might need some guidance on how to add an e2e test which makes sure the front proxy actually respects the annotation

vincepri:
/retitle ✨ Allow home workspaces to be accessed via `user:<username>`

PR #2549, opened 2023-01-04
🐛 cel-go: update to pull in data race fix

### LINKED ISSUE #2548: bug: CEL validation sometimes fails with weird index out of bounds errors
### Describe the bug

There is a data race in the CEL library that is sometimes causing "strange" errors where you wouldn't expect them

### Steps To Reproduce

1. Perform patch/update operations on types that have CEL validation
2. Sometimes they fail

Example when updating an APIBinding's status:

```
APIBinding.apis.kcp.io "workload.kcp.io-8xcjg" is invalid: spec.reference: Invalid value: "object": internal error: runtime error: index out of range [1] with length 0 evaluating rule: APIExport reference must not be changed
```

### Expected Behaviour

No error

### Additional Context

There is a data race when reslicing the decorators slice. https://github.com/google/cel-go/pull/620 is the upstream fix.

ncdc:
## Summary
cel-go: update to pull in data race fix

## Related issue(s)

Fixes #2548

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -188,6 +188,7 @@ require (
 )
 
 replace (
+	github.com/google/cel-go => github.com/kcp-dev/cel-go v0.0.0-20230104211646-409eafa0bc12
 	github.com/kcp-dev/kcp/pkg/apis => ./pkg/apis
 	k8s.io/api => github.com/kcp-dev/kubernetes/staging/src/k8s.io/api v0.0.0-20221221132345-4bcc0e78af51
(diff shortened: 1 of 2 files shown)


ncdc:
#2550 tracks going back to upstream

go.mod
```diff
@@ -188,6 +188,7 @@ require (
 )
 
 replace (
+	github.com/google/cel-go => github.com/kcp-dev/cel-go v0.0.0-20230104211646-409eafa0bc12
```
vincepri, line 191: Couldn't we just set the version to be the commit here, or do we need the fork?

ncdc: It hasn't been backported yet upstream

ncdc: We need to be on 0.12 though

PR #2556, opened 2023-01-05
:seedling: Switch cel-go to upstream v0.12.6

### LINKED ISSUE #2550: Switch back to upstream google/cel-go when a new 0.12.x release is out
#2549 fixes a data race bug in cel-go, but uses our fork which has a backport of the fix. Once there is a new upstream 0.12.x release, we should switch back to that.

vincepri:

## Related issue(s)

Fixes #2550

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -92,7 +92,7 @@ require (
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
 	github.com/golang/protobuf v1.5.2 // indirect
 	github.com/google/btree v1.0.1 // indirect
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -188,7 +188,7 @@ require (
 )
 
 replace (
-	github.com/google/cel-go => github.com/kcp-dev/cel-go v0.0.0-20230104211646-409eafa0bc12
+	github.com/google/cel-go => github.com/google/cel-go v0.12.6
```
ncdc, line 191: Can we remove this entirely?

ncdc: K, thanks for checking!

PR #2557, opened 2023-01-05
⚠️ Move Workspace.Status.{Cluster, URL} to Spec

vincepri:
## Summary

In order to preserve the URL field, given that Status may be wiped under a number of conditions and most times URL cannot be easily or at all reconstructured, this changeset moves the field under Spec.

In addition, now the admission code also checks that any modification to the URL is performed by a system user.

## Related issue(s)

Fixes #

### DIFF
--- a/config/crds/tenancy.kcp.io_workspaces.yaml
+++ b/config/crds/tenancy.kcp.io_workspaces.yaml
@@ -28,7 +28,7 @@ spec:
       name: Phase
       type: string
     - description: URL to access the workspace
-      jsonPath: .status.URL
+      jsonPath: .spec.URL
       name: URL
       type: string
     - jsonPath: .metadata.creationTimestamp
@@ -66,6 +66,19 @@ spec:
             default: {}
             description: WorkspaceSpec holds the desired state of the ClusterWorkspace.
             properties:
+              URL:
+                description: "URL is the address under which the Kubernetes-cluster-like
+                  endpoint can be found. This URL can be used to access the workspace
+                  with standard Kubernetes client libraries and command line tools.
+                  \n Set by the system."
+                type: string
+              cluster:
+                description: "cluster is the name of the logical cluster this workspace
+                  is stored under. \n Set by the system."
+                type: string
+                x-kubernetes-validations:
+                - message: cluster is immutable
+                  rule: self == oldSelf
               shard:
                 description: "location constraints where this workspace can be scheduled
                   to. \n If the no location is specified, an arbitrary location is
@@ -149,22 +162,15 @@ spec:
                 - message: path is immutable
                   rule: '!has(oldSelf.path) || !has(self.path) || self.path == oldSelf.path'
             type: object
+            x-kubernetes-validations:
+            - message: URL cannot be unset
+              rule: '!has(oldSelf.URL) || has(self.URL)'
+            - message: cluster cannot be unset
+              rule: '!has(oldSelf.cluster) || has(self.cluster)'
           status:
             default: {}
             description: WorkspaceStatus communicates the observed state of the Workspace.
             properties:
-              URL:
-                description: url is the address under which the Kubernetes-cluster-like
-                  endpoint can be found. This URL can be used to access the workspace
-                  with standard Kubernetes client libraries and command line tools.
-                type: string
-              cluster:
-                description: cluster is the name of the logical cluster this workspace
-                  is stored under.
-                type: string
-                x-kubernetes-validations:
-                - message: cluster is immutable
-                  rule: self == oldSelf
               conditions:
                 description: Current processing state of the ClusterWorkspace.
                 items:
@@ -228,11 +234,6 @@ spec:
                 - Ready
                 type: string
             type: object
-            x-kubernetes-validations:
-            - message: cluster is immutable
-              rule: '!has(oldSelf.cluster) || has(self.cluster)'
-            - message: URL cannot be unset
-              rule: '!has(oldSelf.URL) || has(self.URL)'
         required:
         - spec
         type: object
--- a/pkg/admission/workspace/admission.go
+++ b/pkg/admission/workspace/admission.go
@@ -152,6 +152,8 @@ func (o *workspace) Validate(ctx context.Context, a admission.Attributes, _ admi
 		return fmt.Errorf("failed to convert unstructured to ClusterWorkspace: %w", err)
 	}
 
+	isSystemPrivileged := sets.NewString(a.GetUserInfo().GetGroups()...).Has(kuser.SystemPrivilegedGroup)
+
 	switch a.GetOperation() {
 	case admission.Update:
 		u, ok = a.GetOldObject().(*unstructured.Unstructured)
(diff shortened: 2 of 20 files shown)


pkg/reconciler/tenancy/workspace/workspace_reconcile_shard_url.go
```diff
@@ -0,0 +1,80 @@
...
+	conditionsv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/apis/conditions/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/util/conditions"
+	"github.com/kcp-dev/kcp/pkg/logging"
+)
+
+type shardURLReconciler struct {
+	getShardByHash func(hash string) (*corev1alpha1.Shard, error)
+}
+
+func (r *shardURLReconciler) reconcile(ctx context.Context, workspace *tenancyv1beta1.Workspace) (reconcileStatus, error) {
```
sttts, line 41: shouldn't we move `cluster` at the same time? It will always be set in sync with the URL.

pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling_test.go
```diff
@@ -84,6 +85,7 @@ func TestReconcileScheduling(t *testing.T) {
 		},
 		{
 			name:                  "two-phase commit, part two: location is set",
+			reconcileSpecURL:      true,
```
sttts, line 88: this is hard to understand. Why should a test ever skip this?

pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling.go
```diff
@@ -150,19 +147,13 @@ func (r *schedulingReconciler) reconcile(ctx context.Context, workspace *tenancy
...
-			// shouldn't happen since we just checked in isValidShard
-			conditions.MarkFalse(workspace, tenancyv1alpha1.WorkspaceScheduled, tenancyv1alpha1.WorkspaceReasonReasonUnknown, conditionsv1alpha1.ConditionSeverityError, "Invalid connection information on target Shard: %v.", err)
-			return reconcileStatusStopAndRequeue, err // requeue
-		}
-
-		u.Path = path.Join(u.Path, clusterName.Path().RequestPath())
-		workspace.Status.URL = u.String()
+		// We've done the two phase commit, set the workspace as scheduled.
 		workspace.Status.Cluster = clusterName.String()
 		conditions.MarkTrue(workspace, tenancyv1alpha1.WorkspaceScheduled)
```
sttts, line 152: I think this PR is the wrong direction. We should rather set cluster and URL (in spec) here (that's what scheduling is), and then set the condition in another pass. And this other pass can be in this very same reconciler.

pkg/reconciler/tenancy/workspace/workspace_reconcile_shard_url.go
vincepri: It can't be set at the same time, given that Cluster is under Status and the URL is under Spec now. The committer code actively disallows setting both metadata/spec and status in the same pass (the tests prior to these changes were encountering a panic) and this is the main reason why a new reconciler was introduced

pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling.go
vincepri: Yeah that's what we had before, see https://github.com/kcp-dev/kcp/pull/2557#discussion_r1063510107 for reasoning on splitting it up

pkg/admission/workspace/admission.go
```diff
@@ -163,27 +165,40 @@ func (o *workspace) Validate(ctx context.Context, a admission.Attributes, _ admi
...
 			return admission.NewForbidden(a, errors.New("spec.type is immutable"))
 		}
 
-		if old.Status.Cluster != "" && cw.Status.Cluster == "" {
-			return admission.NewForbidden(a, errors.New("status.cluster cannot be unset"))
+		if old.Spec.Cluster != "" && cw.Spec.Cluster == "" {
+			return admission.NewForbidden(a, errors.New("spec.cluster cannot be unset"))
 		}
 
 		if cw.Status.Phase != corev1alpha1.LogicalClusterPhaseScheduling {
```
sttts, line 186: this is still a pretty hard condition on status. Not sure it leads to problems.

pkg/admission/workspace/admission_test.go
```diff
@@ -374,28 +380,28 @@ func TestValidate(t *testing.T) {
 						Annotations: map[string]string{"experimental.tenancy.kcp.io/owner": "{}"},
 					},
 					Spec: tenancyv1beta1.WorkspaceSpec{
+						Cluster: "somewhere",
```
sttts, line 383: missing a test that rejects changes for non-privileged users

pkg/admission/workspace/admission.go
vincepri: 👍 I'll keep an eye out for errors in the e2e, although this isn't ideal to check for the Phase. We could just remove these checks instead?

pkg/apis/tenancy/v1beta1/types.go
```diff
@@ -81,6 +83,23 @@ type WorkspaceSpec struct {
 	//
 	// +optional
 	Location *WorkspaceLocation `json:"shard,omitempty"`
+
+	// Cluster is the name of the logical cluster this workspace is stored under.
```
sttts, line 87: lower-case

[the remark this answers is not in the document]
vincepri: Why lowecase?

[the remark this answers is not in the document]
ncdc: For API types, we go against godoc conventions and match the casing of the json tag, so the OpenAPI docs are generated correctly.

config/crds/tenancy.kcp.io_workspaces.yaml
```diff
@@ -66,6 +66,19 @@ spec:
             default: {}
             description: WorkspaceSpec holds the desired state of the ClusterWorkspace.
             properties:
+              URL:
```
vincepri, line 69: Should this be just `url` instead of uppercase? @ncdc

ncdc: I can't remember offhand what the conventions suggest doing. I'd leave any changes to it to a separate PR, if needed. WDYT?

pkg/apis/tenancy/v1beta1/types.go
```diff
@@ -58,6 +58,8 @@ type Workspace struct {
 }
 
 // WorkspaceSpec holds the desired state of the ClusterWorkspace.
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.URL) || has(self.URL)",message="URL cannot be unset"
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.cluster) || has(self.cluster)",message="cluster is immutable"
```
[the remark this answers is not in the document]
ncdc: Yeah, it is immutable. This validation doesn't enforce that, though. This passes validation on update if:
1. old.cluster is empty, or
2. old.cluster is set and new.cluster is set

There is a separate validation on the Cluster field itself to ensure immutability:
```
// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="cluster is immutable"
```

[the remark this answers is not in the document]
vincepri: Ah ok, in this specific validation piece I can see how the message should be different, thanks!

config/crds/tenancy.kcp.io_workspaces.yaml
vincepri: Sure thing, was mostly wondering if that'd be something we'd want to do or not, given that we're introducing a new field under spec and removing the older one

vincepri:
Flake #2501

pkg/admission/workspace/admission.go
```diff
@@ -163,27 +165,40 @@ func (o *workspace) Validate(ctx context.Context, a admission.Attributes, _ admi
 			return fmt.Errorf("failed to convert unstructured to ClusterWorkspace: %w", err)
 		}
 
+		if old.Spec.Cluster != "" && cw.Spec.Cluster == "" {
+			return admission.NewForbidden(a, errors.New("spec.cluster cannot be unset"))
+		}
+		if old.Spec.Cluster != cw.Spec.Cluster && !isSystemPrivileged {
+			return admission.NewForbidden(a, errors.New("spec.Cluster can only be changed by system privileged users"))
```
ncdc, line 172:
```suggestion
			return admission.NewForbidden(a, errors.New("spec.cluster can only be changed by system privileged users"))
```

ncdc (approved):

PR #2559, opened 2023-01-06
:sparkles: server: wire cache informers instead of root informers

sttts:
- wire the cache informers into the relevant controllers that are needed to create and initialize a workspace.
- add workspace types to replication to the cache server
- fix workspace initialization VW on a non-root shard
- add `--location-selector` flag to `kubectl ws create` plugin

### DIFF
--- a/cmd/sharded-test-server/cache.go
+++ b/cmd/sharded-test-server/cache.go
@@ -40,11 +40,11 @@ import (
 )
 
 func startCacheServer(ctx context.Context, logDirPath, workingDir string) (<-chan error, string, error) {
-	red := color.New(color.BgHiRed, color.FgHiWhite).SprintFunc()
-	inverse := color.New(color.BgHiWhite, color.FgHiRed).SprintFunc()
+	cyan := color.New(color.BgHiCyan, color.FgHiWhite).SprintFunc()
+	inverse := color.New(color.BgHiWhite, color.FgHiCyan).SprintFunc()
 	out := lineprefix.New(
-		lineprefix.Prefix(red(strings.ToUpper("cache"))),
-		lineprefix.Color(color.New(color.FgHiRed)),
+		lineprefix.Prefix(cyan(strings.ToUpper("cache"))),
+		lineprefix.Color(color.New(color.FgHiCyan)),
 	)
 	loggerOut := lineprefix.New(
 		lineprefix.Prefix(inverse(strings.ToUpper("cache"))),
--- a/pkg/cliplugins/workspace/plugin/kubeconfig.go
+++ b/pkg/cliplugins/workspace/plugin/kubeconfig.go
@@ -449,6 +449,8 @@ type CreateWorkspaceOptions struct {
 	IgnoreExisting bool
 	// ReadyWaitTimeout is how long to wait for the workspace to be ready before returning control to the user.
 	ReadyWaitTimeout time.Duration
+	// LocationSelector is the location selector to use when creating the workspace to select a matching shard.
+	LocationSelector string
 
 	kcpClusterClient kcpclientset.ClusterInterface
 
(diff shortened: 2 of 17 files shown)


fgiloux:
Suggestion: do we want to use something like "global" instead of "cache" as a prefix for informers and like bound to the cache server? I am just concerned that "cache" is also used for the local (client-go) cache and it may not be obvious at first sight which cache is referred to.

pkg/server/controllers.go
```diff
@@ -952,10 +954,6 @@ func (s *Server) installAPIExportEndpointSliceController(ctx context.Context, co
 			logger.Error(err, "failed to finish post-start-hook")
 			return nil // don't klog.Fatal. This only happens when context is cancelled.
 		}
-		if err := s.waitForOptionalSync(hookContext.StopCh); err != nil {
-			logger.Error(err, "failed to finish post-start-hook")
-			return nil // don't klog.Fatal. This only happens when context is cancelled.
-		}
```
fgiloux, line 958: What guarantees that shard informers for the cache server have synced before APIExportEndpointSliceController is started?

sttts: does it matter?

fgiloux: ack

cmd/sharded-test-server/cache.go
```diff
@@ -40,11 +40,11 @@ import (
 )
 
 func startCacheServer(ctx context.Context, logDirPath, workingDir string) (<-chan error, string, error) {
-	red := color.New(color.BgHiRed, color.FgHiWhite).SprintFunc()
-	inverse := color.New(color.BgHiWhite, color.FgHiRed).SprintFunc()
+	cyan := color.New(color.BgHiCyan, color.FgHiWhite).SprintFunc()
```
p0lyn0mial, line 43: why `BgHiCyan` is better than `color.BgHiRed` :)

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -486,6 +488,10 @@ func (o *CreateWorkspaceOptions) Complete(args []string) error {
 
 // Validate validates the CreateWorkspaceOptions are complete and usable.
 func (o *CreateWorkspaceOptions) Validate() error {
+	if _, err := metav1.ParseToLabelSelector(o.LocationSelector); err != nil {
```
p0lyn0mial, line 491: is an empty `LocationSelector` a valid label ?

pkg/reconciler/cache/replication/replication_reconcile.go
```diff
@@ -75,6 +76,17 @@ func (c *controller) reconcile(ctx context.Context, gvrKey string) error {
 			func(cluster logicalcluster.Name, _, name string) (interface{}, error) {
 				return c.localShardLister.Cluster(cluster).Get(name)
 			})
+	case tenancyv1alpha1.SchemeGroupVersion.WithResource("workspacetypes").String():
+		return c.reconcileObject(ctx,
+			keyParts[1],
+			tenancyv1alpha1.SchemeGroupVersion.WithResource("workspacetypes"),
```
p0lyn0mial, line 82: i think we could also extend our e2e tests to cover the new type

pkg/server/controllers.go
```diff
@@ -683,8 +683,11 @@ func (s *Server) installAPIBindingController(ctx context.Context, config *rest.C
 		if err := wait.PollImmediateInfiniteWithContext(goContext(hookContext), time.Millisecond*100, func(ctx context.Context) (bool, error) {
 			crdsSynced := s.ApiExtensionsSharedInformerFactory.Apiextensions().V1().CustomResourceDefinitions().Informer().HasSynced()
 			exportsSynced := s.KcpSharedInformerFactory.Apis().V1alpha1().APIExports().Informer().HasSynced()
+			cacheExportsSynced := s.CacheKcpSharedInformerFactory.Apis().V1alpha1().APIExports().Informer().HasSynced()
+			schemasSynced := s.KcpSharedInformerFactory.Apis().V1alpha1().APIExports().Informer().HasSynced()
```
p0lyn0mial, line 687: this should be `APIResourceSchemas`

sttts: fixed in follow-up PR

pkg/reconciler/cache/replication/replication_reconcile.go
p0lyn0mial: can be done in a follow-up

PR #2573, opened 2023-01-09
🌱 crdcleanup: strongly type enqueue params

ncdc:
## Summary
Strongly type the params used in our enqueue functions to avoid accidentally enqueueing the wrong kind of object.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
+++ b/pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
@@ -80,22 +80,18 @@ func NewController(
 
 	crdInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
 		FilterFunc: func(obj interface{}) bool {
-			crd, ok := obj.(*apiextensionsv1.CustomResourceDefinition)
-			if !ok {
-				return false
-			}
-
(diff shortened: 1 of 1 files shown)


vincepri (approved):
/lgtm

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
```diff
@@ -80,33 +80,29 @@ func NewController(
 
 	crdInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
 		FilterFunc: func(obj interface{}) bool {
-			crd, ok := obj.(*apiextensionsv1.CustomResourceDefinition)
```
stevekuznetsov, line 83: Technically this is a regression in behavior but I think in practice it's the more correct thing to do anyway?

stevekuznetsov (approved):

pkg/reconciler/apis/crdcleanup/crdcleanup_controller.go
ncdc: Yes, but it's better to panic to indicate programmer error I'd say.

PR #2575, opened 2023-01-09
🌱 apiexport: strongly type enqueue params

ncdc:
## Summary
Strongly type the params used in our enqueue functions to avoid accidentally enqueueing the wrong kind of object.

Also switch to struct function pointers for lister access.

## Related issue(s)

Fixes #


vincepri (approved):
/lgtm

PR #2584, opened 2023-01-10
WIP 🌱 pkg/virtual/apiexport: add permission claims authorizer

s-urbaniak:
## Summary
This adds a permission claims authorizer which respects a permission claim's resource selector.

It partially backports functionality from https://github.com/kcp-dev/kcp/pull/2089 (reverse permission claims)

/cc @nrb 

TODOS:
- [ ] Add e2e tests

## Related issue(s)

https://github.com/kcp-dev/kcp/pull/2456

### DIFF
--- a/pkg/virtual/apiexport/authorizer/permission_claims_authorizer.go
+++ b/pkg/virtual/apiexport/authorizer/permission_claims_authorizer.go
@@ -1,0 +1,167 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package authorizer
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/kcp-dev/logicalcluster/v3"
+
+	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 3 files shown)


pkg/virtual/apiexport/authorizer/permission_claims_authorizer.go
```diff
@@ -0,0 +1,167 @@
...
+		for _, claim := range binding.Spec.PermissionClaims {
+			if claim.Group != attr.GetAPIGroup() || claim.Resource != attr.GetResource() {
+				continue
+			}
+			// If at least one claimed resource is not accepted by the user, access must be denied
+			if claim.State != apisv1alpha1.ClaimAccepted {
+				return authorizer.DecisionNoOpinion, fmt.Sprintf("claim %v of API export %q/%q is not accepted", claim, logicalcluster.From(apiExport), apiExport.Name), nil
+			}
+			requestingAllResources := attr.GetName() == "" && attr.GetNamespace() == ""
+			resourceSelectorMatches := claim.All || requestingAllResources
```
nrb, line 121: Do we want to allow this? I thought `claim.All == false && resourceSelector == nil` meant that we still denied it because the user did not make an explicit choice.

Does https://github.com/kcp-dev/kcp/blob/main/pkg/apis/apis/v1alpha1/types_apiexport.go#L199 not prevent this case?

This seems like it has the potential to leak things if a user forgets a field, even if it is better on UX.

nrb: See this comment when the field was introduced: https://github.com/kcp-dev/kcp/pull/2286#issuecomment-1300653317

nrb: I guess even I'm not following that rule either 🤦‍♂️  https://github.com/kcp-dev/kcp/pull/2456/files#diff-1ce4a4f7308c17b5ca27f9f57099ae355d2086266eea334dc6f9ae977000c2e6R150

s-urbaniak:
this has to be done using admission

PR #2585, opened 2023-01-10
🌱 Debug dns e2e test failures

davidfestal:
## Summary

Dummy PR to test / debug failures of the new Syncer DNS e2e test on main

## Related issue(s)

Related to the fact that these e2e tests fail on PR https://github.com/kcp-dev/kcp/pull/2452.
Need to know if it is also the case on top of main.


davidfestal:
Not needed anymore. The e2e test works correctly.

PR #2588, opened 2023-01-10
🐛 WorkspaceType admission should respect LimitAllow{Parents, Children}.None

### LINKED ISSUE #2582: bug: Root Type Workspace Under Non-Admin User Workspace
### Describe the bug

I am able to create a `Root` type workspace under my user home workspace (e.g., `root:users:wg:bl:rh-sso-rh-ee-zxiao`), which does not complies with the `Root` type definition.  Expect to only able to create `Root` only under the top level workspace and only by admin.

### Steps To Reproduce

1. Login to a non-admin user
2. Navigate to the user's home workspace
3. Create a `Root` type workspace with the following command
```
kubectl kcp ws create <ws_name> --type=Root
```

### Expected Behaviour

Not able to create because user is non-admin and attempt to create under user's home root

### Additional Context

_No response_

vincepri:
Signed-off-by: Vince Prignano <vincepri@redhat.com>

## Related issue(s)

Fixes #2582

### DIFF
--- a/pkg/admission/workspacetypeexists/admission.go
+++ b/pkg/admission/workspacetypeexists/admission.go
@@ -453,7 +453,14 @@ func (r *transitiveTypeResolver) resolve(wt *tenancyv1alpha1.WorkspaceType, seen
 func validateAllowedParents(parentAliases, childAliases []*tenancyv1alpha1.WorkspaceType, parentType, childType logicalcluster.Path) error {
 	var errs []error
 	for _, childAlias := range childAliases {
-		if childAlias.Spec.LimitAllowedParents == nil || len(childAlias.Spec.LimitAllowedParents.Types) == 0 {
+		if childAlias.Spec.LimitAllowedParents == nil {
+			continue
+		}
+		if childAlias.Spec.LimitAllowedParents.None {
+			errs = append(errs, fmt.Errorf("workspace type %s cannot have any parent", childType))
+			continue
+		}
+		if len(childAlias.Spec.LimitAllowedParents.Types) == 0 {
 			continue
 		}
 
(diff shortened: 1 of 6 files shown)


pkg/admission/workspacetypeexists/admission.go
```diff
@@ -487,11 +494,15 @@ func validateAllowedParents(parentAliases, childAliases []*tenancyv1alpha1.Works
 func validateAllowedChildren(parentAliases, childAliases []*tenancyv1alpha1.WorkspaceType, parentType, childType logicalcluster.Path) error {
 	var errs []error
 	for _, parentAlias := range parentAliases {
-		if parentAlias.Spec.LimitAllowedChildren == nil || len(parentAlias.Spec.LimitAllowedChildren.Types) == 0 {
+		if parentAlias.Spec.LimitAllowedChildren == nil {
```
vincepri, line 497: @sttts While working on the above change, I noticed here that we were skipping the `LimitAllowedChildren.None` check if `Types` was empty, although I would expect Types always be empty if None is set to true? Is that not the case?

If the above assumption is correct, is there any place I could add validation that when `.None` is true, `.Types` should be empty?

stevekuznetsov: Might be able to express that using the kube-builder annotations or yaml-patch on the CRD

sttts: yes, that seems to be a bug.

vincepri: ptal, added CEL validation for the types

stevekuznetsov (approved):

ncdc (approved):

PR #2590, opened 2023-01-10
:sparkles: DRY committer factory functions

csams:
## Summary
Extract common code from scoped and non-scoped committer functions.


pkg/reconciler/committer/committer.go
```diff
@@ -110,6 +110,7 @@ func generatePatchAndSubResources[Sp any, St any](old, obj *Resource[Sp, St]) ([
 
 	// Simultaneous updates of spec and status are never allowed.
 	if specOrObjectMetaChanged && statusChanged {
+		// TODO(panic): remove panic and handle the error
```
stevekuznetsov, line 113: This can't be handled at runtime - it needs to be handled at code-writing-time, which is what "programmer error" is intended to communicate - perhaps we need to word that better?

stevekuznetsov (commented):
DRY bits look good!

pkg/reconciler/committer/committer.go
csams: Ack - my concern was opening up a DOS vector in the case of a rarely used and untested controller code path, but `runtime.HandleCrash` when the controller starts should protect a prod system in that case.

PR #2592, opened 2023-01-11
🐛 cache: stop reporting an error from a CRD lister on incorrect cluster name

p0lyn0mial:
## Summary
since all available CRDs for the cache server are stored in `bootstrap.SystemCRDLogicalCluster`
and there is only **a single** registry/instance per a CRD
there is no need to filter by the given cluster

stops printing the following message
`"programmer error" err="cluster-unaware crd lister got asked for root cluster"`
when a request for a resource was made (i.e.: `shards/*/clusters/abc/apis/apis.kcp.io/v1alpha1/apiexports'`) (https://github.com/kcp-dev/kubernetes/blob/b0ffa0d98215b3ec8a1b00f2f001ad86784be8c6/staging/src/k8s.io/apiextensions-apiserver/pkg/apiserver/customresource_handler.go#L299)

## Related issue(s)

Fixes #


stevekuznetsov (approved):
/retest

PR #2594, opened 2023-01-11
:bug: reconciler/cache: remove reflection and fix replicate+NotFound logic

sttts:
Maybe fixing the npe flake.
Maybe fixing the npe flake.
Maybe fixing the npe flake.
Maybe fixing the npe flake.
Maybe fixing the npe flake.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -33,7 +33,6 @@ linters:
   - prealloc
   - revive
   - staticcheck
-  - thelper
   - unconvert
   - unused
   - usestdlibvars
--- a/pkg/reconciler/cache/replication/replication_reconcile.go
+++ b/pkg/reconciler/cache/replication/replication_reconcile.go
@@ -160,9 +160,6 @@ func (c *controller) reconcileObject(ctx context.Context,
 	if err != nil && !errors.IsNotFound(err) {
 		return err
 	}
-	if !replicate {
-		return nil
-	}
 	if errors.IsNotFound(err) {
 		// issue a live GET to make sure the localObject was removed
 		_, err = c.dynamicKcpLocalClient.Cluster(cluster.Path()).Resource(gvr).Namespace(namespace).Get(ctx, name, metav1.GetOptions{})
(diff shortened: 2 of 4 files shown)


pkg/reconciler/cache/replication/replication_reconcile.go
```diff
@@ -106,7 +117,7 @@ func (c *controller) reconcile(ctx context.Context, gvrKey string) error {
 				if err != nil {
 					return nil, true, err
```
p0lyn0mial, line 118: why is this `return nil, true, err` and not `return nil, false, err` ?

pkg/reconciler/cache/replication/replication_reconcile.go
```diff
@@ -121,7 +132,7 @@ func (c *controller) reconcile(ctx context.Context, gvrKey string) error {
...
```
p0lyn0mial, line 133: same question

.golangci.yaml
```diff
@@ -33,7 +33,6 @@ linters:
   - prealloc
   - revive
   - staticcheck
-  - thelper
```
vincepri, line 36: Why remove this linter?

pkg/reconciler/cache/replication/replication_reconcile.go
sttts: does not really matter. But yes, false would be more canonical

ncdc: Let's fix to be consistent?

pkg/reconciler/cache/replication/replication_reconcile.go
ncdc: Let's fix to be consistent?

.golangci.yaml
sttts: discussed in Slack

pkg/reconciler/cache/replication/replication_reconcile.go
sttts: fixed

pkg/reconciler/cache/replication/replication_reconcile.go
sttts: fixed

ncdc (approved):

PR #2602, opened 2023-01-11
:bug: committer: fix equality check in statusless committer

sttts:
The old bytes comparison was always false.


ncdc (approved):

ncdc:
#2603 
/retest

stevekuznetsov:
Any chance we can share more code between these two?

PR #2604, opened 2023-01-11
:bug: Role and ClusterRole merger should not modify cache objects

vincepri:
Signed-off-by: Vince Prignano <vincepri@redhat.com>

## Summary

This PR also reverts 0831b69e29298cb6d6561d9cfa0a200882155ef7

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/virtual/framework/wrappers/rbac/merging.go
+++ b/pkg/virtual/framework/wrappers/rbac/merging.go
@@ -20,6 +20,7 @@ import (
 	"github.com/kcp-dev/logicalcluster/v3"
 
 	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 	rbaclisters "k8s.io/client-go/listers/rbac/v1"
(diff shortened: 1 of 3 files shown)


pkg/virtual/framework/wrappers/rbac/merging.go
```diff
@@ -240,40 +241,53 @@ func (l mergedRoleBindingNamespaceLister) Get(name string) (*rbacv1.RoleBinding,
...
 			}
 		}
-		if foundIndex != -1 {
-			array[foundIndex].Rules = append(array[foundIndex].Rules, roleToBeMerged.Rules...)
-		} else {
-			array = append(array, roleToBeMerged)
-		}
+
+		// The Role is a new one, deepcopy it in the current array.
+		current = append(current, cachedRole.DeepCopy())
```
sttts, line 256: use DeepCopyInto. It saves allocations.

sttts: Why don't we keep track of deepcopied roles and only do it when the second instance is seen?

sttts:
/lgtm
/approve

I think we can improve that even more. But this stops the bleeding.

pkg/virtual/framework/wrappers/rbac/merging_test.go
```diff
@@ -0,0 +1,208 @@
+/*
+Copyright 2021 The KCP Authors.
```
AlainRoy, line 2: 2023?

vincepri: :) Copy paste

PR #2606, opened 2023-01-12
🐛 IndexByLogicalClusterPathAndName should return clusterpath and name

qiujian16:

## Summary

I think IndexByLogicalClusterPathAndName should return clusterPath+Name by default?

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/indexers/indexers.go
+++ b/pkg/indexers/indexers.go
@@ -109,7 +109,7 @@ func IndexByLogicalClusterPathAndName(obj interface{}) ([]string, error) {
 		}, nil
 	}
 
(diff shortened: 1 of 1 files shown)


davidfestal:
The existing code effectively seemed wrong. But are we aware of something broken by this erroneous code ?
If yes, shouldn't we add a test case (unit test or e2e) that corresponds to this fix ?

davidfestal (commented):
See my previous comment here: https://github.com/kcp-dev/kcp/pull/2606#issuecomment-1379987515

sttts:
Nice finding.

/lgtm
/approve

Issue #2568, opened 2023-01-09
maxPermissionPolicyAuth and systemCRDAuth should be part of deep SAR


sttts:
_Originally posted by @sttts in https://github.com/kcp-dev/kcp/pull/2543#discussion_r1064995122_

s-urbaniak:
See the comment in https://github.com/kcp-dev/kcp/pull/2543#discussion_r1067816521.

I verified and at least the system CRD authorizer is not skipped as part of deep SAR.

Issue #2384, opened 2022-11-18
CRUD operations via APIExport VW on resources claimed from root workspace with local maximum permission policy always return 403 


astefanutti:
### Describe the bug

CRUD operations via APIExport VW on resources claimed from root workspace with local maximum permission policy always return 403.

### Steps To Reproduce

1. Create an APIExport that claims permission on `scheduling.kcp.dev/placements` from the `scheduling.kcp.dev` APIExport in the `root` workspace
2. Bind that APIExport in a consumer workspace with the claim permission accepted
3. Create a placement resource via the APIExport VW API server
4. This returns 403, while any authenticated user is granted permission via the local maximum permission policy defined in the `root` workspace

### Expected Behaviour

The local maximum permission policy should correctly apply to requests made to the APIExport VW API server.

sttts:
The solution IMO is incorrect.

PR #2607, opened 2023-01-12
WIP pkg/authorization: remove deep SAR check from maximal permission policy

s-urbaniak:
## Summary
Currently, deep SAR is ommitted in maximal permission policy authorizer. That was done in order to fix https://github.com/kcp-dev/kcp/issues/2384. This fix is not right according to @sttts hence the removal here.

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/2568


stevekuznetsov (commented):
Can we wire up a test that would enforce the correct understanding of this feature?

s-urbaniak:
This needs deeper thought as we have an e2e test `TestRootAPIExportAuthorizers` verifying the intended behavior. Simply disabling it breaks e2e. However, we have to think deeper about a proper fix.

PR #2610, opened 2023-01-12
:seedling: e2e: TestAuditLogs remove duplicate audit-log-path

hardys:
## Summary

This decouples scratch directory creation from the `PrivateKcpServer` fixture, so that we can log to the regular artifactDir location (which means the log data will show up in saved CI job artifacts), and avoid the duplicate flag and associated test cleanup.

## Related issue(s)
Follows on from #2591 to ensure audit log output shows up in the saved CI artifacts for all tests.

### DIFF
--- a/test/e2e/audit/audit_log_test.go
+++ b/test/e2e/audit/audit_log_test.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


test/e2e/homeworkspaces/home_workspaces_test.go
```diff
@@ -91,7 +91,15 @@ func TestUserHomeWorkspaces(t *testing.T) {
 		t.Run(testCase.name, func(t *testing.T) {
 			t.Parallel()
 			tokenAuthFile := framework.WriteTokenAuthFile(t)
-			server := framework.PrivateKcpServer(t, framework.WithCustomArguments(append(framework.TestServerArgsWithTokenAuthFile(tokenAuthFile), testCase.kcpArgs...)...))
```
stevekuznetsov, line 94: We're not using the directories to determine the location of the auth file, is that on purpose? Why does this test need custom dirs?

hardys: Yeah you're right, it's only really needed by `TestAuditLogs` - updated and I'll retitle the PR, thanks!

stevekuznetsov (approved):

test/e2e/audit/audit_log_test.go
```diff
@@ -36,25 +37,28 @@ func TestAuditLogs(t *testing.T) {
 	t.Parallel()
 	framework.Suite(t, "control-plane")
 
-	server := framework.PrivateKcpServer(t, framework.WithCustomArguments([]string{"--audit-log-path", "./audit-log", "--audit-policy-file", "./policy.yaml"}...))
-
-	ctx, cancelFunc := context.WithCancel(context.Background())
+	artifactDir, dataDir, err := framework.ScratchDirs(t)
+	require.NoError(t, err)
 
-	t.Cleanup(func() {
```
stevekuznetsov, line 43: I'm assuming we lost this since we defer cleanup to the scratch-dir cleanup routines?

hardys: Yes since the audit-log ends up in the regular artifactDir which is [created via testing.TempDir](https://github.com/kcp-dev/kcp/blob/main/test/e2e/framework/util.go#L114) it'll be cleaned with the other test output unless ARTIFACT_DIR is specified.

PR #2616, opened 2023-01-12
:sparkles: Add Upsync controller (replaces PR #2214)

### LINKED ISSUE #1980: Storage - Upsync controller (syncer)
- Create an UpSync controller (similar to the Status controller), but which would create / update resource in upstream according to watched resource downstream
  - The downstream resources watched by the UpSync controller (to be upsynced) should:
    - have an `UpSync` label 
    - have a `NamespaceLocator` annotation if they are Cluster-wide resources (to mark the SyncTarget + uer workspace they would be upsynced to)
  - When Upsynced, the resources:
    - should have their NamespaceLocator annotation deleted
    - have the `Upsync` value in the `workloadv1alpha1.ClusterResourceStateLabelPrefix` label
 - Have the UpSync controller also watch the upstream resources that have the `Upsync` value in the `workloadv1alpha1.ClusterResourceStateLabelPrefix` label, so that if any of those was deleted or updated, it would be overriden again by the downstream resource.
- Have the UpSync controller point to a different URL (not the Syncer Virtual Workspace  URL, but an UpSyncer virtual workspace URL)

davidfestal:
## Summary

This PR implements the Upsyncer controller inside the Syncer.
It is a replacement of the [initial PR](https://github.com/kcp-dev/kcp/pull/2214) that had to be reworked and rebased.

## Related issue(s)

Fixes #1980

### DIFF
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -53,6 +53,7 @@ import (
 	"github.com/kcp-dev/kcp/pkg/syncer/resourcesync"
 	"github.com/kcp-dev/kcp/pkg/syncer/spec"
 	"github.com/kcp-dev/kcp/pkg/syncer/status"
+	"github.com/kcp-dev/kcp/pkg/syncer/upsync"
 	. "github.com/kcp-dev/kcp/tmc/pkg/logging"
 )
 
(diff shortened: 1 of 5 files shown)


davidfestal:
Still missing:
- add unit tests to check the cases when the status is changed
- add unit tests to check the case when the `UpSync` labels are not there
- e2e tests

davidfestal:
Finally we kept [initial PR](https://github.com/kcp-dev/kcp/pull/2214) for the continuation of the work.
Closing this one then.

Issue #2597, opened 2023-01-11
bug: cannot bind workspaces to location workspace


kasturinarra:
### Describe the bug

when trying to bind workspace to location workspace below error is seen
error: apibindings.apis.kcp.dev "kubernetes-1pre20xf" is forbidden: unable to create APIImport: no permission to bind to export "kubernetes"

### Steps To Reproduce

1. Go to user home workspace
2. create new kcp workspace called `qe-experiment`
3. kcp workload sync command to add sync target
4. wait for synctarget to be ready
5. create deployment in the same ws  `qe-experiment` , deployment in 0/1 state
6. Bind using the command `kubectl kcp bind compute "root:users:lw:ao:rh-sso-knarrakcp:qe-experiment"`

### Expected Behaviour

workspace should be binded to location workspace

stevekuznetsov:
@kasturinarra what version of `kcp` are you using?

kasturinarra:
> @kasturinarra what version of `kcp` are you using? And what version of `kubectl kcp` plugin?

@stevekuznetsov both are v0.10.0

[knarra@knarra ~]$ kubectl kcp --version
kcp version v1.24.3+kcp-v0.10.0

qiujian16:
is location workspace also qe-experiment? I think you run sync command against qe-experiment workspace?

kasturinarra:
> is location workspace also qe-experiment? I think you run sync command against qe-experiment workspace?

yes, location workspace also qe-experiment.

kasturinarra:
Thanks to @ncdc and @davidfestal we were finally able to root cause the issue. 

After performing steps below i see that i could successfully bind workspaces to location workspace.
1. create a clusterrole in root:compute workspace and yaml for the clusterrole looks something like below

[knarra@knarra kcp]$ cat /tmp/clusterrole.yaml 
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: system:kcp:apiexport:tenancy:bind
rules:
- apiGroups: ["apis.kcp.dev"]
  resources:
  - "apiexports"
  resourceNames:
  - "kubernetes"
  verbs: ["bind"]

2. create clusterrolebinding in root:compute workspace and yaml for the clusterrolebinding would look something like below

[knarra@knarra kcp]$ cat /tmp/clusterrolebinding.yaml 
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
[...]

PR #2618, opened 2023-01-13
🐛 Add bind permission for kubernetes apiexport

qiujian16:

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/2597


qiujian16:
maybe also need to backport to 0.10.0?

davidfestal:
> maybe also need to backport to 0.10.0?

I assume it would be good yes.

davidfestal (approved):

davidfestal:
> > maybe also need to backport to 0.10.0?
> 
> I assume it would be good yes.

@qiujian16 see my answer above. Do you want to do it ?

Issue #2579, opened 2023-01-10
bug: Placement is not updated when the SyncTarget Heartbeat condition changes


davidfestal:
### Describe the bug

When the hearbeat condition of a `SyncTarget` changes (syncer started / stopped), this should impact placements which point to the location workspace containing this location.
That's not the case anymore on `main`.
That means that id a Syncer is started after KCP is running, and the SyncTarget becomes Ready, it will still not be taken in account by the placement scheduler.

### Steps To Reproduce

1. Run KCP
2. Create a location workspace,
3. In the location workspace create a SyncTarget with the `workload sync ` CLI command
4. Create a workload workspace and bind it to the location workspace with the `bind compute` CLI command
5. The placement created during bind will **not** be scheduled, and will have the `No Valid SyncTargets` message in the related condition.
6. Now start the syncer related to the Synctarget, and wait for the SyncTarget (in the location workspace) to be Ready
7. The Placement in the workload workspace should be updated and scheduled to the SyncTarget but it is not.

### Expected Behaviour

When the hearbeat condition of a `SyncTarget` changes (syncer started / stopped), this should impact placements which point to the location workspace containing this location.

davidfestal:
I started observing this after the workspace refactoring.
@sttts any idea ?

davidfestal:
cc @qiujian16

PR #2624, opened 2023-01-16
🐛 Enqueue location upon synctarget update

qiujian16:
We cannot directly read the logicalcluster path from synctarget to get the related placement, since the value of that path is not unique. Instead, we need to get the location of the related synctarget at first, which always has the path annotation.

## Related issue(s)

Fixes  https://github.com/kcp-dev/kcp/issues/2579

### DIFF
--- a/pkg/reconciler/workload/placement/placement_controller.go
+++ b/pkg/reconciler/workload/placement/placement_controller.go
@@ -37,7 +37,6 @@ import (
 
 	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
 	"github.com/kcp-dev/kcp/pkg/apis/core"
-	corev1alpha1 "github.com/kcp-dev/kcp/pkg/apis/core/v1alpha1"
 	schedulingv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/scheduling/v1alpha1"
 	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
 	kcpclientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned/cluster"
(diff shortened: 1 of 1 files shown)


pkg/reconciler/workload/placement/placement_controller.go
```diff
@@ -269,31 +269,16 @@ func (c *controller) enqueueSyncTarget(obj interface{}) {
...
-		// placements referencing by path
-		placementsByPath, err := c.placementIndexer.ByIndex(bySelectedLocationPath, path)
-		if err != nil {
-			runtime.HandleError(err)
-			return
-		}
-		placements = append(placements, placementsByPath...)
-	}
 
 	logger := logging.WithObject(logging.WithReconciler(klog.Background(), ControllerName), obj.(*workloadv1alpha1.SyncTarget))
```
davidfestal, line 279: Seems to me that the `SyncTarget` value added here to the logger, will be, at least in some logger implementation, overwritten by the [same operation applied in `enqueueLocation()`](https://github.com/kcp-dev/kcp/pull/2624/files#diff-5d404ae02f7e27afad2fd77bb0230e44b70b580a6494ecb21926e1df21d7df01L218)

pkg/reconciler/workload/placement/placement_controller.go
```diff
@@ -269,31 +269,16 @@ func (c *controller) enqueueSyncTarget(obj interface{}) {
...
-	for _, placement := range placements {
-		c.enqueuePlacement(placement, logger, " because of SyncTarget")
+	for _, location := range locations {
+		c.enqueueLocation(location, logger, " because of SyncTarget")
```
davidfestal, line 281: Instead of managing the source (==reason) of the `enqueue` by a suffix, can't we simply add **distinct** values ?
Something like: `logger.WithValues(logging.FromPrefix("locationReason"))` and `logger.WithValues(logging.FromPrefix("syncTargetReason"))` ?

So you can accumulate the reason objects if needed, and would have always a fixed log message.

pkg/reconciler/workload/placement/placement_controller.go
qiujian16: I think they will have different key https://github.com/kcp-dev/kcp/blob/main/pkg/logging/constants.go#L84

pkg/reconciler/workload/placement/placement_controller.go
```diff
@@ -250,50 +250,35 @@ func (c *controller) enqueueAPIBinding(obj interface{}) {
...
+		c.enqueuePlacement(placement, logger)
 	}
 }
 
-func (c *controller) enqueueSyncTarget(obj interface{}) {
-	key, err := kcpcache.DeletionHandlingMetaClusterNamespaceKeyFunc(obj)
-	if err != nil {
-		runtime.HandleError(err)
-		return
+func (c *controller) enqueueSyncTarget(obj interface{}, logger logr.Logger) {
```
sttts, line 260: this is very noisy. We should rethink whether we really need to enqueue half of the world.

1. What is the exact reason we have to enqueue a sync target?
2. What can we do to get the placements more directly?

sttts: Goal: only sync the placement that actually schedules to that SyncTarget.

qiujian16: ideally we should only enqueue location here. The missing part is  we do not know what APIs are compatible in the location. So if a synctarget is updated with "api compaible/non-comptaible", we have to select the clusters here.

sttts: > So if a synctarget is updated with "api compaible/non-comptaible"

For that we could have a special case, couldn't we?

sttts: Also, can we queue only those locations that are affected, e.g. that include the SyncTarget? If you remove a label from a SyncTarget, we could also see that and queue those locations where this SyncTarget was removed from.

sttts: Nvmd:

On the other hand, the new change does not really cause more noise, but it will list all locations and from there get all placements that reference these locations. So there is hope that these are not all placements, vs. the old code that queued all placements.

sttts:
With https://github.com/kcp-dev/kcp/pull/2624/files#r1101774288

/lgtm
/approve

PR #2629, opened 2023-01-16
:seedling: Admission cleanup

fgiloux:
## Summary

Cleaning up fields that seem to be unused in APIBinding admission.


p0lyn0mial (commented):
what about removing `ValidateAPIBinding`metdhod ? Isn't it already covered by OpenAPIv3 validation?

fgiloux:
> what about removing `ValidateAPIBinding`metdhod ? Isn't it already covered by OpenAPIv3 validation?

Let's discuss it on slack and possibly have it in a separate PR as it seems that it was [intentional](https://github.com/kcp-dev/kcp/blob/33064b51ef3a3777a4beccf2f1df93d99e4b6435/pkg/admission/apibinding/validation.go#L58).

ncdc (approved):

PR #2640, opened 2023-01-18
🌱 test/e2e: add support for dynamic users using client certs

s-urbaniak:
## Summary
This adds support to generate users on the fly in e2e tests. Currently, only users predefined in a token fixture file can be used. This allows for more descriptive user names and also allows adding groups dynamically.

## Related issue(s)

N/A

### DIFF
--- a/cmd/test-server/main.go
+++ b/cmd/test-server/main.go
@@ -28,6 +28,7 @@ import (
 
 	genericapiserver "k8s.io/apiserver/pkg/server"
 
+	"github.com/kcp-dev/kcp/cmd/sharded-test-server/third_party/library-go/crypto"
 	shard "github.com/kcp-dev/kcp/cmd/test-server/kcp"
 )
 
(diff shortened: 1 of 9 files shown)


s-urbaniak:
oh we need to add some logic to handle the sharded use case which uses a dedicated client cert for the front proxy.

cmd/sharded-test-server/frontproxy.go
```diff
@@ -132,7 +132,7 @@ func startFrontProxy(
 		fmt.Sprintf("--root-directory=%s", filepath.Join(workDirPath, ".kcp-front-proxy")),
 		fmt.Sprintf("--root-kubeconfig=%s", filepath.Join(workDirPath, ".kcp/root.kubeconfig")),
 		fmt.Sprintf("--shards-kubeconfig=%s", filepath.Join(workDirPath, ".kcp-front-proxy/shards.kubeconfig")),
-		fmt.Sprintf("--client-ca-file=%s", filepath.Join(workDirPath, ".kcp/client-ca.crt")),
+		fmt.Sprintf("--client-ca-file=%s", filepath.Join(workDirPath, "test/e2e/framework/client-ca.crt")),
```
hardys, line 135: I guess this is temporary, and we'll generate an extra CA cert like is [already done elsewhere](https://github.com/kcp-dev/kcp/blob/main/cmd/sharded-test-server/main.go#L69..L100) e.g we can create a new one in the [shard setup](https://github.com/kcp-dev/kcp/blob/main/cmd/sharded-test-server/shard.go) and rename the resulting files?

s-urbaniak: yeah, this is temporary to see if we have other cascading e2e failures and it seems we do

PR #2596, opened 2023-01-11
:sparkles: enable shard scheduling and fix e2e tests

sttts:
Switching on random scheduling of workspaces to shards.

The current state of e2e tests:

✅ apibinding
:white_check_mark:  audit
✅ authorizer
:white_check_mark:  cache
:white_check_mark: test/e2e/conformance (@fgiloux https://github.com/sttts/kcp/pull/14 - the fix pulled into this PR)
:white_check_mark:  customresourcedefinition
:white_check_mark:  garbagecollector (@p0lyn0mial: Originally reported as `red` but after running it a few times on my local machine and examining [a CI run](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kcp-dev_kcp/2596/pull-ci-kcp-dev-kcp-main-e2e-shared/1615003616358174720) I moved to `green`)
:white_check_mark:  homeworkspaces
:white_check_mark: test/e2e/watchcache (flakes, @p0lyn0mial, fixed by: https://github.com/kcp-dev/kcp/pull/2632)
:white_check_mark:   workspacetype
:white_check_mark:  reconciler/apiexport
:white_check_mark:  reconciler/cache (@p0lyn0mial, fixed by: https://github.com/kcp-dev/kcp/pull/2620)
:white_check_mark: test/e2e/reconciler/workspace (@p0lyn0mial, fixed by: https://github.com/kcp-dev/kcp/pull/2639)
✅ quota
✅  reconcilder/*
✅  virtual/*

~~:red_circle:  quota (@ncdc)~~
~~:red_circle: apibinding~~
~~:red_circle:  authorizer, flakes (@sur)~~
⚫ skipped: test/e2e/reconciler/locationworkspace (@p0lyn0mial, temporal fix in https://github.com/kcp-dev/kcp/pull/2650)
⚫ skipped: test/e2e/reconciler/scheduling (@p0lyn0mial, temporal fix in https://github.com/kcp-dev/kcp/pull/2651)
⚫ skipped: test/e2e/reconciler/cluster (@p0lyn0mial, temporal fix in https://github.com/kcp-dev/kcp/pull/2657)
⚫ skipped: test/e2e/recocniler/namespace (@p0lyn0mial, temporal fix in https://github.com/kcp-dev/kcp/pull/2658)
⚫ skipped: test/e2e/syncer (@p0lyn0mial, temporal fix in https://github.com/kcp-dev/kcp/pull/2652)
[...]

PR #2641, opened 2023-01-18
🐛 reconciler/apiexport: use the gloabl informer to get Shards

p0lyn0mial:
## Summary
since `Shard` resources are only on the root shard the `apiexport` controller must use the global (cached) informer otherwise the list will always be empty on a non-root shard.

## Related issue(s)

xref: https://github.com/kcp-dev/kcp/pull/2596


fgiloux:
this looks good to me too

ncdc (approved):

PR #2219, opened 2022-10-18
:book: Add hugo documentation structure to docs

avinal:
## Summary

- support for versioned docs
- docs are contained in the repo itself
- docs are built and deployed to github pages on every commit to main and release
- breaking changes to docs structure

Signed-off-by: Avinal Kumar <avinal@redhat.com>

PR #2656, opened 2023-01-19
:book: Create "older" under docs Releases

### LINKED ISSUE #2655: bug: documentation is outdated but allegedly current
### Describe the bug

I think we agreed that https://docs.kcp.io/kcp should have the capability to hold docs for multiple releases, and the existence of the "Releases" menu looks like an attempt to realize that.  But:

1. there is only one choice available now, "main", and
2. the documentation there is for older releases.

See the attached screenshot.

Screen Shot 2023-01-19 at 3 22 03 PM

I understand that it will take some time to update the documentation for the current release.  As we work through that process, it would probably be less confusing if there were at least two choices under the "Releases" menu, one getting updated for the current truth and one preserving the existing documentation (seems a shame to simply discard what has been written).

### Steps To Reproduce

1. Try to follow the documented steps with the currrent release

### Expected Behaviour

See above.

### Additional Context

_No response_

MikeSpreitzer:
## Summary
This PR creates an "older" "Release" under docs, to preserve the existing documentation --- because it is accurate for older releases but not the main branch anymore.

## Related issue(s)

Fixes #2655


hardys:
I'm not familiar with the tools used to publish the docs website, but we already cut release branches, so surely we'd be better having some way to navigate the rendered docs on each branch vs a catch-all "older" category?

Also this appears to move a bunch of stuff that is still relevant to main so I'm not entirely sure this is the right approach

MikeSpreitzer:
I am open to specific suggestions for how to do better.

hardys:
> I am open to specific suggestions for how to do better.

I think we need to configure the website to show each branch in the "releases" drop-down menu, as I said I'm not familiar with the configuration but it appears we need to add the releases [here](https://github.com/kcp-dev/kcp/blob/main/docs/config.toml#L153..L155)?  

Searching I found [these docs](https://www.docsy.dev/docs/adding-content/versioning/#adding-a-version-drop-down-menu) which seem to indicate we just need to add `params.versions` for each stable branch?

Perhaps @avinal as the author of #2219 may have some additional input

avinal:
> think we need to configure the website to show each branch in the "releases" drop-down menu, as I said I'm not familiar with the configuration but it appears we need to add the releases [here](https://github.com/kcp-dev/kcp/blob/main/docs/config.toml#L153..L155)?

Yes, it just needs to be configured for the drop-down, I am not sure why this isn't showing up. I remember some conversation on this but I was on vacation and forgot about it, sorry about that. I will check and reply ASAP.

avinal:
@hardys I was able to find the problem, we were not able to test the logic when it was implemented. It seems on release the folder was not moved and hence the older versioned docs are not showing up. https://github.com/kcp-dev/kcp/actions/runs/3687694096/jobs/6241597908#step:12:134

I will open a fix for this. This PR is probably not needed and the other problem is, with every new release storing older docs will eat up space, so we prefer to store only their HTML render in a separate branch (gh-pages) and copy them over for every release.

avinal:
@MikeSpreitzer if there are docs improvements in this pull request, please put them into a new pull request.

ncdc:
Closing this as we've moved to a new docs framework and we have multiple versions now.

PR #2659, opened 2023-01-20
:bug: server: relax ShardVirtualWorkspace flag validation

hardys:
## Summary
In #2407 we introduced this validation, but it doesn't handle the case where the VW server is not standalone, but you still want to specify `--shard-virtual-workspace-url`

This scenario is likely when deploying a sharded environment, since you'll want the VW traffic to go via the front-proxy regardless of whether the VW server is standalone or not.

## Related issue(s)
Fixes an issue introduced via #2407


ncdc (approved):

hardys:
/retest

Appears to be stuck in merge pool, not sure if retest will help but giving it a try...

hardys:
/test e2e-shared

Raised #2665 to track the test failure, I can't reproduce locally so I'm assuming it's a flake

Issue #2589, opened 2023-01-10
bug: flake: FAIL: pkg/syncer/spec TestSpecSyncerProcess/SpecSyncer_sync_to_downstream,_syncer_finalizer_already_there


stevekuznetsov:
### Describe the bug

```
 === FAIL: pkg/syncer/spec TestSpecSyncerProcess/SpecSyncer_sync_to_downstream,_syncer_finalizer_already_there (0.15s)
I0110 17:19:39.459535   36355 shared_informer.go:255] Waiting for caches to sync for kcp-ddsif-gvr-source
I0110 17:19:39.459605   36355 shared_informer.go:262] Caches are synced for kcp-ddsif-gvr-source
I0110 17:19:39.459658   36355 shared_informer.go:255] Waiting for caches to sync for kcp-ddsif-gvr-source
I0110 17:19:39.459742   36355 shared_informer.go:262] Caches are synced for kcp-ddsif-gvr-source
    spec_process_test.go:1178: TestSpecSyncerProcess/SpecSyncer_sync_to_downstream,_syncer_finalizer_already_there: upstream ddsif synced
    spec_process_test.go:1180: TestSpecSyncerProcess/SpecSyncer_sync_to_downstream,_syncer_finalizer_already_there: downstream ddsif synced
    spec_process_test.go:1207: TestSpecSyncerProcess/SpecSyncer_sync_to_downstream,_syncer_finalizer_already_there: secrets and namespaces informers synced
    spec_process_test.go:1194: TestSpecSyncerProcess/SpecSyncer_sync_to_downstream,_syncer_finalizer_already_there: gvrs updated
I0110 17:19:39.462162   36355 spec_process.go:277] "Created downstream namespace for upstream namespace"
I0110 17:19:39.475732   36355 spec_process.go:519] "Upserted upstream resource to downstream" downstream.name="theDeployment"
    spec_process_test.go:1225: 
        	Error Trace:	spec_process_test.go:1225
        	Error:      	Should be empty, but was   []testing.Action{
[...]

PR #2667, opened 2023-01-23
:bug: server: revert non-standalone VW URL

hardys:
## Summary
Partially reverts logic introduced in #2407 so we don't unconditionally use the external URL - when non-standalone VW mode is used we still need the old behavior, or the loopback client can't access the non-standalone VW server.

This breaks shard bootstrapping (evidently not in our e2e scenarios, but in a real deployment where probes mean the proxy isn't ready until the shard bootstrap is complete)

I suspect further work will be required to make standalone VW mode work when deployed on a real cluster, since the ExternalAddress will go via the proxy, which depends on proxy bootstrapping (when the shard is deployed via a Service)

### DIFF
--- a/pkg/server/controllers.go
+++ b/pkg/server/controllers.go
@@ -811,9 +811,10 @@ func (s *Server) installAPIBinderController(ctx context.Context, config *rest.Co
 	// Client used to create APIBindings within the initializing workspace
 	config = rest.CopyConfig(config)
 	config = rest.AddUserAgent(config, initialization.ControllerName)
+	config.Host += initializingworkspacesbuilder.URLFor(tenancyv1alpha1.WorkspaceAPIBindingsInitializer)
 
-	vwURL := fmt.Sprintf("https://%s", s.GenericConfig.ExternalAddress)
 	if !s.Options.Virtual.Enabled && s.Options.Extra.ShardVirtualWorkspaceURL != "" {
+		vwURL := fmt.Sprintf("https://%s", s.GenericConfig.ExternalAddress)
 		if s.Options.Extra.ShardVirtualWorkspaceCAFile == "" {
 			// TODO move verification up
 			return fmt.Errorf("s.Options.Extra.ShardVirtualWorkspaceCAFile is required")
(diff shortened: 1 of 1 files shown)


hardys:
@ncdc FYI - I think this presents a problem for "real" deployments with standalone VW, since if we point `ShardVirtualWorkspaceURL` at the proxy, we have a chicken/egg issue wrt shard bootstrapping since the proxy isn't available until the shard `Service` endpoint is exposed (which depends on bootstrapping/readyz)

hardys:
/hold

e2e-sharded failure does look related e.g

```
W0124 10:26:12.676402     712 reflector.go:324] k8s.io/client-go@v0.0.0-20230113194502-8f8fb5fa4eea/tools/cache/reflector.go:167: failed to list *v1alpha1.LogicalCluster: Get "https://10.129.28.8:7445/services/initializingworkspaces/system:apibindings/clusters/%2A/apis/core.kcp.io/v1alpha1/logicalclusters?limit=500&resourceVersion=0": x509: certificate is valid for localhost, 10.129.28.8, not apiserver-loopback-client
```

#2407 passed so I need to figure out what changed and reproduce locally, currently I'm not clear why this previously worked given that AFAICS the controllers always use the loopback client...

hardys:
Ok the failure in the previous revision was my mistake, I used `ShardVirtualWorkspaceURL` which is for shard->vw traffic, we still need to use `ExternalAddress` in the standalone VW case, so the initializingworkspaces requests hit the proxy.

This should now be ready to land as a first step (enables the integrated-VW case to work again when the shard is deployed behind a `Service`)

However @ncdc FYI I think we'll have to do further work to enable the standalone-vw case on real clusters - currently the shard bootstrap dependency on `ExternalAddress` presents a chicken/egg problem given that the proxy isn't available until the shard is ready.

hardys:
> @hardys let's chat tomorrow?

Sure, but we could also consider landing this as an interim measure (to unblock the unstable test environment)

hardys:
/hold cancel

This could land now to unblock things, but I'll follow-up with @ncdc to discuss the plan for standalone vw - we could also look at reworking sharded-test-server so e2e better replicates the serialized bootstrapping behavior implied by using services and probes in a "real" deployment.

hardys:
> we could also look at reworking sharded-test-server so e2e better replicates the serialized bootstrapping behavior implied by using services and probes in a "real" deployment.

Turns out the proxy startup is already serialized (we wait until the shard is ready), so the main difference is `--external-hostname` isn't pointing to the proxy in CI, pushed https://github.com/kcp-dev/kcp/pull/2672 so we have a reproducer

ncdc (approved):

ncdc:
Flake #2589 
/retest

PR #2672, opened 2023-01-25
[WIP] :bug: sharded-test-server: Add external-hostname flag

hardys:
## Summary
We are seeing real cluster deployments of KCP fail because the external-hostname and shard-external-url flags typically point to the front-proxy endpoint, this aligns the e2e behavior so we can work through the issues with chicken/egg availability during shard bootstrapping and discuss the best solution.

### DIFF
--- a/cmd/sharded-test-server/shard.go
+++ b/cmd/sharded-test-server/shard.go
@@ -82,6 +82,7 @@ func newShard(ctx context.Context, n int, args []string, standaloneVW bool, serv
 			fmt.Sprintf("--embedded-etcd-peer-port=%d", embeddedEtcdPeerPort(n)),
 		)
 	}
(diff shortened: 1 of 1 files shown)


hardys:
@ncdc I think this reproduces the issues seen in "real" cluster deployments, because we're using the [ExternalAddress for the vwURL](https://github.com/kcp-dev/kcp/blob/main/pkg/server/controllers.go#L815) since https://github.com/kcp-dev/kcp/pull/2407 - I tried to work around that in https://github.com/kcp-dev/kcp/pull/2667

How is controller sharding handled?  I assume we can't just use `ShardBaseURL` for the controller traffic because it may potentially need to be routed to another shard via the proxy indexer?

ncdc:
@hardys I think I missed your question above somehow. Wanted to follow up and see if we still want this PR? If so, how can I help?

PR #2676, opened 2023-01-25
:seedling: pkg/client: publish as a separate module

stevekuznetsov:

Supersedes https://github.com/kcp-dev/kcp/pull/2192


pkg/client/go.mod
```diff
@@ -0,0 +1,59 @@
+module github.com/kcp-dev/kcp/pkg/client
+
+go 1.19
+
+require (
+	github.com/kcp-dev/apimachinery/v2 v2.0.0-alpha.0
+	github.com/kcp-dev/client-go v0.0.0-20230112161950-ed70a87bb0c8
+	github.com/kcp-dev/kcp/pkg/apis v0.0.0-00010101000000-000000000000
```
ncdc, line 8: I assume this needs to be a real version?

stevekuznetsov: Dang. You're totally right ... I don't know how we would handle that without going full-blown staging repo.

ncdc: Do something similar but without a "staging repo" - auto-generate go.mod in here?

astefanutti: Could Go workspaces help? It seems it's being considered upstream:
* https://groups.google.com/a/kubernetes.io/g/dev/c/zcGFMru0gPU
* https://www.youtube.com/watch?v=_zW49yF5wE0.

stevekuznetsov: Ha! Even my editor told me to use workspaces. Dang. Let me look into that, thanks for the tip @astefanutti

stevekuznetsov:
For #2794

ncdc:
Replaced by #2869

Issue #2635, opened 2023-01-17
:seedling: Update controllers to use committer package


csams:
Many controllers use copy+paste abstraction for common controller functions, but they aren't all consistent. The `committer` package encapsulates type safe spec and status patching for arbitrary resources, so controllers can delegate to it for consistency in that phase of reconciliation.

This issue captures the work to update existing controllers to use the `committer` package.

- [ ] pkg/reconciler/apis/extraannotationsync/apibindingannotation_controller.go
- [x] pkg/reconciler/apis/permissionclaimlabel/permissionclaimlabel_controller.go
- [ ] pkg/reconciler/apis/permissionclaimlabel/permissionclaimlabel_resource_controller.go
- [ ] pkg/reconciler/core/logicalclusterdeletion/logicalcluster_deletion_controller.go
- [ ] pkg/reconciler/core/shard/shard_controller.go #2873 
- [x] pkg/reconciler/scheduling/location/location_controller.go
- [x] pkg/reconciler/scheduling/placement/placement_controller.go
- [x] pkg/reconciler/tenancy/bootstrap/bootstrap_controller.go
- [ ] pkg/reconciler/tenancy/logicalcluster/logicalcluster_controller.go
- [ ] pkg/reconciler/tenancy/workspacetype/clusterworkspacetype_controller.go
- [ ] pkg/reconciler/workload/apiexportcreate/apiexportcreate_controller.go
- [ ] pkg/reconciler/workload/apiexport/workload_apiexport_controller.go
- [x] pkg/reconciler/workload/basecontroller/controller.go #2691 
- [ ] pkg/reconciler/workload/namespace/namespace_controller.go #2874 
- [ ] pkg/reconciler/workload/synctarget/synctarget_controller.go
- [ ] pkg/syncer/resourcesync/controller.go
- [ ] pkg/syncer/spec/spec_process.go
- [ ] pkg/syncer/status/status_process.go
- [ ] pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_controller.go
- [ ] pkg/virtual/syncer/controllers/apireconciler/syncer_apireconciler_controller.go

PR #2691, opened 2023-01-27
:seedling: Simplify heartbeat controller

csams:
## Summary
- Remove the base controller as it was only used by the heartbeat controller
- Collapse the base controller into the heartbeat controller
- Simplify the heartbeat controller code and structure
- Update the heartbeat controller to use the committer package.

## Related issue(s)
#2635


sttts:
This base controller has only one consumer. Why do we have it as separate package?

ncdc:
If I had to guess, we probably kept moving consumers away until we were only left with 1, but never collapsed. I'll do that.

stevekuznetsov (approved):

PR #2702, opened 2023-01-30
:book: Add example on how resource name should be provided for syncing

kasturinarra:
## Summary
Current documentation does not clearly state what should be the form of resourcename while syncing to the pyhiscal cluster. Adding an example and clear message about the same.
## Related issue(s)

Fixes #


davidfestal:
@kasturinarra Do you think you could take the opportunity of completing the related CLI help: https://github.com/kcp-dev/kcp/blob/main/pkg/cliplugins/workload/plugin/sync.go#L126 ?

davidfestal (approved):

PR #2703, opened 2023-01-30
:seedling:  test/e2e: DRY up waiting on conditions

stevekuznetsov:

/assign @ncdc


stevekuznetsov:
Probably can follow-up with a summary of all conditions in the case where we can't find the one that is being waited on - if we ensure a stable sort order for those, they will de-dupe correctly in the `framework.Eventually()` call.

ncdc (approved):

PR #2562, opened 2023-01-06
:sparkles: wire cross-shard authorization

sttts:
- replicate ClusterRoles+ClusterBindings that are labeled to be replicated
- added replication label controllers for ClusterRoles+Bindings relevant to APIExport authorization
- allow deep SAR again the cache server data

Issue #2685, opened 2023-01-26
bug: flake: quota


stevekuznetsov:
### Describe the bug

```
 === FAIL: test/e2e/quota TestKubeQuotaNormalCRDs (unknown)
    quota_test.go:252: shared kcp server will target configuration "/go/src/github.com/kcp-dev/kcp/.kcp/admin.kubeconfig"
    quota_test.go:268: Created root:organization workspace root:e2e-workspace-cx2ww as /clusters/1zcegg5d1igixo6t
    quota_test.go:275: Created root:universal workspace root:e2e-workspace-cx2ww:one as /clusters/30ln42srrbq4011n
    quota_test.go:276: Created root:universal workspace root:e2e-workspace-cx2ww:two as /clusters/224pc4kizldywhnr
    quota_test.go:278: Install a normal sheriffs CRD into workspace 1 "root:e2e-workspace-cx2ww:one"
I0126 19:16:42.920442   30739 bootstrap.go:129] "bootstrapping CRD" crd="sheriffs.aj8y3ikb.io"
    quota_test.go:281: Install another normal sheriffs CRD with a different schema into workspace 2 "root:e2e-workspace-cx2ww:two"
I0126 19:16:43.124610   30739 bootstrap.go:129] "bootstrapping CRD" crd="sheriffs.aj8y3ikb.io"
I0126 19:16:43.145999   30739 bootstrap.go:150] "bootstrapped CRD" crd="sheriffs.aj8y3ikb.io" customresourcedefinition.workspace="224pc4kizldywhnr" customresourcedefinition.namespace="" customresourcedefinition.name="sheriffs.aj8y3ikb.io" customresourcedefinition.apiVersion="" duration="21.344088ms"
I0126 19:16:43.146095   30739 bootstrap.go:169] "waiting for CRD to be established" crd="sheriffs.aj8y3ikb.io" customresourcedefinition.workspace="224pc4kizldywhnr" customresourcedefinition.namespace="" customresourcedefinition.name="sheriffs.aj8y3ikb.io" customresourcedefinition.apiVersion=""
    quota_test.go:303: Creating ws 1 quota
    quota_test.go:307: Waiting for ws 1 quota to show usage
    quota_test.go:316: Create 2 sheriffs to reach the quota limit
    quota_test.go:317: Creating aj8y3ikb.io/v1 sheriffs root:e2e-workspace-cx2ww:one|default/ws1-1
[...]

PR #2712, opened 2023-01-31
:bug: Fix Makefile clean target and add .kcp cleanup

hardys:
## Summary
The `clean` target was added in https://github.com/kcp-dev/kcp/pull/2562 but doesn't currently work due to the wildcard in the `GOBIN_DIR` test.

Also following some discussion on slack about cleanup for the `.kcp*` directories I added cleanup so we can more easily switch between e.g shared/sharded e2e tests locally e.g `make clean test-e2e-sharded-minimal`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -382,9 +382,14 @@ verify-modules: modules  ## Verify go modules are up to date
 	hack/verify-go-modules.sh
 
 .PHONY: clean
-clean:
+clean: clean-workdir
 	rm -fr $(TOOLS_DIR)
 	rm -f $(GOBIN_DIR)/*
+
+.PHONY: clean-workdir
+clean-workdir: WORK_DIR ?= .
+clean-workdir:
+	rm -fr $(WORK_DIR)/.kcp*
 
 .PHONY: help
 help: ## Show this help.
--- a/Makefile
+++ b/Makefile
@@ -383,8 +383,8 @@ verify-modules: modules  ## Verify go modules are up to date
 
 .PHONY: clean
 clean:
-	! test -e $(TOOLS_DIR) || rm -r $(TOOLS_DIR)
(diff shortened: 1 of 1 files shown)


ncdc:
I would recommend also adding a new target to just clean the `.kcp*` directories - you probably don't want to clean the tools/bin dirs every time you stop/restart the server.

hardys:
> I would recommend also adding a new target to just clean the `.kcp*` directories - you probably don't want to clean the tools/bin dirs every time you stop/restart the server.

Thanks - I added a `clean-workdir` target

Makefile
```diff
@@ -382,9 +382,14 @@ verify-modules: modules  ## Verify go modules are up to date
 	hack/verify-go-modules.sh
 
 .PHONY: clean
-clean:
-	! test -e $(TOOLS_DIR) || rm -r $(TOOLS_DIR)
-	! test -e $(GOBIN_DIR)/* || rm $(GOBIN_DIR)/*
+clean: clean-workdir
+	! test -d $(TOOLS_DIR) || rm -r $(TOOLS_DIR)
```
ncdc, line 386: Super nit-y but would it be simpler to do `rm -rf $(TOOLS_DIR)` instead? And same for GOBIN_DIR and WORK_DIR/.kcp*?

hardys: Yeah I thought the same, but @sttts did it this way in https://github.com/kcp-dev/kcp/pull/2562 so I decided to stick with the current approach - I agree that would be simpler and I can't see how the test/rm will be any safer?

ncdc: Cool, I'd vote to just do `rm -rf`

hardys:
/test e2e-sharded

#2685

ncdc (approved):

PR #2716, opened 2023-01-31
🌱 manifest: remove outdated manifests

hardys:
## Summary
In the [community call](https://github.com/kcp-dev/kcp/issues/2710) we agreed to remove these and go ahead with updating the [helm-charts repo](https://github.com/kcp-dev/helm-charts) instead (which contains similar manifests which are templated)

The helm-charts repo is also outdated, but I pushed [a PR](https://github.com/kcp-dev/helm-charts/pull/24) aiming to resolve that, and since we probably don't have capacity to maintain the manifests in both places it was agreed to remove these.

Also I did look into how we might revive these using kustomize, but it looks difficult due to the templating we need for externalHostname (and choice of ingress resource type) - AFAICS helm is a better fit for our requirements so we can support multiple versions and configurations.


hardys:
/test e2e-sharded

Looks like #2389 flake

hardys:
/retest

#2739

hardys:
/retest 

#2389

Issue #2713, opened 2023-01-31
bug/flake: TestAPIExportAuthorizers


ncdc:
From https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kcp-dev_kcp/2707/pull-ci-kcp-dev-kcp-main-e2e-sharded/1620426348143054848 for PR #2707
```
=== RUN   TestAPIExportAuthorizers
=== PAUSE TestAPIExportAuthorizers
=== CONT  TestAPIExportAuthorizers
    authorizer_test.go:69: shared kcp server will target configuration "/go/src/github.com/kcp-dev/kcp/.kcp/admin.kubeconfig"
=== CONT  TestAPIExportAuthorizers
    kcp.go:980: waiting for readiness for server at https://10.128.38.8:6444
=== CONT  TestAPIExportAuthorizers
    kcp.go:1023: success contacting https://10.128.38.8:6444/livez
=== CONT  TestAPIExportAuthorizers
    kcp.go:1023: success contacting https://10.128.38.8:6444/readyz
    kcp.go:1001: server at https://10.128.38.8:6444 is ready
=== CONT  TestAPIExportAuthorizers
    assertions.go:1691: Waiting for condition, but got: workspace phase is Initializing, not Ready
=== CONT  TestAPIExportAuthorizers
    authorizer_test.go:74: Created root:organization workspace root:e2e-workspace-ccdmr as /clusters/2j1clh7hrfj2r5l4 on shard "shard-1"
=== CONT  TestAPIExportAuthorizers
    authorizer_test.go:77: Created root:universal workspace root:e2e-workspace-ccdmr:service-provider-1 as /clusters/1huhzk2qxbm3y4si on shard "shard-1"
    authorizer_test.go:78: Created root:universal workspace root:e2e-workspace-ccdmr:service-provider-2 as /clusters/2hf0bx9k2m0adhb5 on shard "shard-1"
    authorizer_test.go:79: Created root:universal workspace root:e2e-workspace-ccdmr:tenant as /clusters/1nyczkrpz8rtgduv on shard "shard-1"
    authorizer_test.go:80: Created root:universal workspace root:e2e-workspace-ccdmr:tenant-shadowed-crd as /clusters/4x24bik73ykm94us on shard "shard-1"
=== CONT  TestAPIExportAuthorizers
[...]

PR #2717, opened 2023-01-31
🌱 Trying to fix doc generation

ncdc:


.github/workflows/docs-gen-and-push.yaml
```diff
@@ -125,7 +125,10 @@ jobs:
 
       # copy back older versions to the current docs
       - name: Copy everything from archive to docs
-        run: rsync -av docs/v* public
+        run: |
+          if ls docs/v* >/dev/null; then
+            rsync -av docs/v* public
+          fi
```
avinal, line 131: Should we add an else branch as well, printing "No such folder"? will be easier to track in logs. (probably an overkill)

avinal (approved):
Looks Good to me

ncdc:
Flake #2713, which should be resolved now. Hopefully another retest will get past that.
/retest

PR #2725, opened 2023-02-01
📖 document replicating a new resource

p0lyn0mial:
## Summary
Describes required steps for adding a new resource for replication.

## Related issue(s)

Fixes #


p0lyn0mial:
@s-urbaniak since you are about to add a new resource for replication please verify if the instruction is valid. Thanks!

PR #2533, opened 2022-12-23
:bug: workload/resource: handle upsynced resources

jmprusi:
## Summary
This PR fixes an issue with upsynced resources getting it's state changed into "Sync" when the initial synctarget is gone or unhealthy.

If the resource is upsynced, check if the synctarget still exists, if not, delete the resource, if the synctarget still exists, do nothing, exit early.

## Requires
- https://github.com/kcp-dev/kcp/pull/2730

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/2530

PR #2730, opened 2023-02-01
🌱 tmc e2e : Split SyncerFixture

davidfestal:
## Summary

Split the `SyncerFixture` start method to provide a way to only maintain Synctarget heartbeat and import apis, without
effectively syncing upstream resources.

## Related issue(s)

No related issue, but this is required in order to avoid conflicting with the e2e tests of the Upsyncer virtual workspace with the implementation of the Upsyncer controller (on the Syncer side) is merged.
This will also allow lighter e2e tests for all the tests that don't need effective syncing of resources to the downstream cluster.

### DIFF
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -402,12 +402,12 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 		go startSyncerTunnel(ctx, upstreamConfig, downstreamConfig, logicalcluster.From(syncTarget), cfg.SyncTargetName)
 	}
 
-	StartHeartbeatKeeper(ctx, kcpSyncTargetClient, cfg.SyncTargetName, cfg.SyncTargetUID)
+	StartHeartbeat(ctx, kcpSyncTargetClient, cfg.SyncTargetName, cfg.SyncTargetUID)
 
 	return nil
 }
 
-func StartHeartbeatKeeper(ctx context.Context, kcpSyncTargetClient kcpclientset.Interface, syncTargetName, syncTargetUID string) {
+func StartHeartbeat(ctx context.Context, kcpSyncTargetClient kcpclientset.Interface, syncTargetName, syncTargetUID string) {
 	logger := klog.FromContext(ctx)
 
 	// Attempt to heartbeat every interval
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -42,6 +42,7 @@ import (
 	"k8s.io/klog/v2"
 
 	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
+	kcpclientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned"
 	kcpclusterclientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned/cluster"
 	kcpinformers "github.com/kcp-dev/kcp/pkg/client/informers/externalversions"
 	kcpfeatures "github.com/kcp-dev/kcp/pkg/features"
@@ -401,6 +402,14 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 		go startSyncerTunnel(ctx, upstreamConfig, downstreamConfig, logicalcluster.From(syncTarget), cfg.SyncTargetName)
 	}
 
+	StartHeartbeatKeeper(ctx, kcpSyncTargetClient, cfg.SyncTargetName, cfg.SyncTargetUID)
+
+	return nil
+}
+
+func StartHeartbeatKeeper(ctx context.Context, kcpSyncTargetClient kcpclientset.Interface, syncTargetName, syncTargetUID string) {
+	logger := klog.FromContext(ctx)
+
 	// Attempt to heartbeat every interval
 	go wait.UntilWithContext(ctx, func(ctx context.Context) {
 		var heartbeatTime time.Time
@@ -409,11 +418,11 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 		// Attempt to heartbeat every second until successful. Errors are logged instead of being returned so the
 		// poll error can be safely ignored.
 		_ = wait.PollImmediateInfiniteWithContext(ctx, 1*time.Second, func(ctx context.Context) (bool, error) {
-			patchBytes := []byte(fmt.Sprintf(`[{"op":"test","path":"/metadata/uid","value":%q},{"op":"replace","path":"/status/lastSyncerHeartbeatTime","value":%q}]`, cfg.SyncTargetUID, time.Now().Format(time.RFC3339)))
-			syncTarget, err = kcpSyncTargetClient.WorkloadV1alpha1().SyncTargets().Patch(ctx, cfg.SyncTargetName, types.JSONPatchType, patchBytes, metav1.PatchOptions{}, "status")
+			patchBytes := []byte(fmt.Sprintf(`[{"op":"test","path":"/metadata/uid","value":%q},{"op":"replace","path":"/status/lastSyncerHeartbeatTime","value":%q}]`, syncTargetUID, time.Now().Format(time.RFC3339)))
+			syncTarget, err := kcpSyncTargetClient.WorkloadV1alpha1().SyncTargets().Patch(ctx, syncTargetName, types.JSONPatchType, patchBytes, metav1.PatchOptions{}, "status")
 			if err != nil {
 				logger.Error(err, "failed to set status.lastSyncerHeartbeatTime")
-				return false, nil //nolint:nilerr
+				return false, nil
 			}
 
 			heartbeatTime = syncTarget.Status.LastSyncerHeartbeatTime.Time
@@ -421,8 +430,6 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 		})
 		logger.V(5).Info("Heartbeat set", "heartbeatTime", heartbeatTime)
 	}, heartbeatInterval)
-
-	return nil
 }
 
 type filteredGVRSource struct {
--- a/test/e2e/framework/syncer.go
+++ b/test/e2e/framework/syncer.go
@@ -130,40 +130,25 @@ func WithDownstreamPreparation(prepare func(config *rest.Config, isFakePCluster 
 	}
 }
 
-// CreateAndStart creates a SyncTarget resource through the `workload sync` CLI command,
-// applies the syncer-related resources in the physical cluster,
-// and then starts a new syncer against the given upstream kcp workspace.
-// Whether the syncer runs in-process or deployed on a pcluster will depend
-// whether --pcluster-kubeconfig and --syncer-image are supplied to the test invocation.
-func (sf *syncerFixture) CreateAndStart(t *testing.T) *StartedSyncerFixture {
+// Create creates a SyncTarget resource through the `workload sync` CLI command,
+// applies the syncer-related resources in the physical cluster.
+// No resource will be effectively synced after calling this method.
+func (sf *syncerFixture) Create(t *testing.T) *appliedSyncerFixture {
 	t.Helper()
-
-	useDeployedSyncer := len(TestConfig.PClusterKubeconfig()) > 0
 
 	artifactDir, _, err := ScratchDirs(t)
 	if err != nil {
 		t.Errorf("failed to create temp dir for syncer artifacts: %v", err)
 	}
 
-	downstreamConfig, downstreamKubeconfigPath, downstreamKubeClient, syncerConfig, syncerID := sf.createSyncTargetAndApplySyncerResources(t, useDeployedSyncer, artifactDir)
+	useDeployedSyncer := len(TestConfig.PClusterKubeconfig()) > 0
 
-	ctx, cancelFunc := context.WithCancel(context.Background())
-	t.Cleanup(cancelFunc)
+	var (
+		downstreamConfig         *rest.Config
+		downstreamKubeconfigPath string
+		syncerID                 string
+	)
 
-	sf.startSyncer(ctx, t, useDeployedSyncer, downstreamKubeClient, syncerConfig, syncerID, artifactDir, downstreamKubeconfigPath)
-
-	startedSyncer := &StartedSyncerFixture{
-		sf.buildAppliedSyncerFixture(ctx, t, downstreamConfig, downstreamKubeClient, syncerConfig, syncerID),
-	}
-
-	// The sync target becoming ready indicates the syncer is healthy and has
-	// successfully sent a heartbeat to kcp.
-	startedSyncer.WaitForSyncTargetReady(ctx, t)
-
-	return startedSyncer
-}
-
-func (sf *syncerFixture) createSyncTargetAndApplySyncerResources(t *testing.T, useDeployedSyncer bool, artifactDir string) (downstreamConfig *rest.Config, downstreamKubeconfigPath string, downstreamKubeClient kubernetesclient.Interface, syncerConfig *syncer.SyncerConfig, syncerID string) {
 	// Write the upstream logical cluster config to disk for the workspace plugin
 	upstreamRawConfig, err := sf.upstreamServer.RawConfig()
 	require.NoError(t, err)
@@ -301,15 +286,32 @@ func (sf *syncerFixture) createSyncTargetAndApplySyncerResources(t *testing.T, u
 	}
 	require.NotEmpty(t, syncerID, "failed to extract syncer namespace from yaml produced by plugin:\n%s", string(syncerYAML))
 
-	syncerConfig = syncerConfigFromCluster(t, downstreamConfig, syncerID, syncerID)
+	syncerConfig := syncerConfigFromCluster(t, downstreamConfig, syncerID, syncerID)
 
-	downstreamKubeClient, err = kubernetesclient.NewForConfig(downstreamConfig)
+	downstreamKubeClient, err := kubernetesclient.NewForConfig(downstreamConfig)
 	require.NoError(t, err)
 
-	return
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	t.Cleanup(cancelFunc)
+
+	return sf.buildAppliedSyncerFixture(ctx, t, downstreamConfig, downstreamKubeClient, downstreamKubeconfigPath, syncerConfig, syncerID)
 }
 
-func (sf *syncerFixture) startSyncer(ctx context.Context, t *testing.T, useDeployedSyncer bool, downstreamKubeClient kubernetesclient.Interface, syncerConfig *syncer.SyncerConfig, syncerID, artifactDir, downstreamKubeconfigPath string) {
+// StartSyncer starts a new Syncer against the upstream kcp workspaces
+// Whether the syncer runs in-process or deployed on a pcluster will depend
+// whether --pcluster-kubeconfig and --syncer-image are supplied to the test invocation.
+func (sf *appliedSyncerFixture) StartSyncer(t *testing.T) *StartedSyncerFixture {
+	t.Helper()
+
+	useDeployedSyncer := len(TestConfig.PClusterKubeconfig()) > 0
+	artifactDir, _, err := ScratchDirs(t)
+	if err != nil {
+		t.Errorf("failed to create temp dir for syncer artifacts: %v", err)
+	}
+
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	t.Cleanup(cancelFunc)
+
 	if useDeployedSyncer {
 		t.Cleanup(func() {
 			ctx, cancelFn := context.WithDeadline(context.Background(), time.Now().Add(wait.ForeverTestTimeout))
@@ -318,10 +320,10 @@ func (sf *syncerFixture) startSyncer(ctx context.Context, t *testing.T, useDeplo
 			// collect syncer logs
 			t.Logf("Collecting syncer pod logs")
 			func() {
-				t.Logf("Listing downstream pods in namespace %s", syncerID)
-				pods, err := downstreamKubeClient.CoreV1().Pods(syncerID).List(ctx, metav1.ListOptions{})
+				t.Logf("Listing downstream pods in namespace %s", sf.SyncerID)
+				pods, err := sf.DownstreamKubeClient.CoreV1().Pods(sf.SyncerID).List(ctx, metav1.ListOptions{})
 				if err != nil {
-					t.Logf("failed to list pods in %s: %v", syncerID, err)
+					t.Logf("failed to list pods in %s: %v", sf.SyncerID, err)
 					return
 				}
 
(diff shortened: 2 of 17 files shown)


test/e2e/reconciler/locationworkspace/local_apiexport_test.go
```diff
@@ -70,7 +70,7 @@ func TestSyncTargetLocalExport(t *testing.T) {
 		framework.WithExtraResources("services"),
 		framework.WithSyncTargetName(syncTargetName),
 		framework.WithSyncedUserWorkspaces(computeWorkspace),
-	).Start(t)
+	).BootstrapSyncTarget(t)
```
sttts, line 73: Why not `StartSyncer`? Why this cryptic name?

davidfestal: It doesn't start the syncer anymore.
The `Start()` function still starts the Syncer as before.

The new, alternate `BootstrapSyncTarget()` method mimics all the actions the Syncer would have done to make the Synctarget usable:
- heartbeat
- API importing
- API authorization checking of APIs in the downstream cluster.

Since I'm quite bad at naming, so I'm very open to another name :-)

pkg/syncer/resourcesync/controller.go
```diff
@@ -312,6 +314,21 @@ func (c *controller) process(ctx context.Context, key string) error {
...
+	if err := c.PatchSyncTargetCondition(ctx, unauthorizedGVRs, syncTarget); err != nil {
+		errs = append(errs, err)
+	}
+
+	return errors.NewAggregate(errs)
+}
+
+type GVRAuthorizationChecker struct {
+	DownstreamKubeClient kubernetes.Interface
+	SyncTargetClient     workloadv1alpha1typed.SyncTargetInterface
```
sttts, line 326: this hints at bad factoring if these must be public. They shouldn't.

pkg/syncer/resourcesync/controller.go
```diff
@@ -312,6 +314,21 @@ func (c *controller) process(ctx context.Context, key string) error {
...
+}
+
+func (c *GVRAuthorizationChecker) PatchSyncTargetCondition(ctx context.Context, unauthorizedGVRs []string, syncTarget *workloadv1alpha1.SyncTarget) error {
```
sttts, line 329: why does a checker (whatever that is, but it is hopefully side-effect free) some patching?

test/e2e/framework/syncer.go
```diff
@@ -511,6 +515,256 @@ func (sf *syncerFixture) Start(t *testing.T) *StartedSyncerFixture {
 	return startedSyncer
 }
 
+// BootstrapSyncTarget simulate the start of a Syncer and performs what the Syncer does to bring and maitain
+// the SyncTarget to a Ready status (heartbeat, and API schema compatibility).
+// No resource will be without effectively synced after calling this method.
+func (sf *syncerFixture) BootstrapSyncTarget(t *testing.T) *StartedSyncerFixture {
```
sttts, line 521: strange name if this simulates only

test/e2e/framework/syncer.go
```diff
@@ -511,6 +515,256 @@ func (sf *syncerFixture) Start(t *testing.T) *StartedSyncerFixture {
...
+				}
+			}
+
+			return authChecker.PatchSyncTargetCondition(ctx, unauthorizedGVRs, syncTarget)
+		})
+		require.NoError(t, err)
+	}, 5*time.Second)
+
+	return startedSyncer
+}
```
sttts, line 766: to be honest, after our conversation to create a fake syncer fixture I expected a 20 liner, maybe 50 lines, not 250. Why do we need such giant logic to keep a sync target alive? My gut feeling: this is yet another maintenance disaster.

test/e2e/framework/syncer.go
```diff
@@ -511,6 +515,256 @@ func (sf *syncerFixture) Start(t *testing.T) *StartedSyncerFixture {
...
+
+	go wait.UntilWithContext(ctx, func(ctx context.Context) {
+		err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			syncTarget, err := kcpClusterClient.Cluster(syncerConfig.SyncTargetPath).WorkloadV1alpha1().SyncTargets().Get(ctx, syncerConfig.SyncTargetName, metav1.GetOptions{})
+			require.NoError(t, err)
+
+			authChecker := resourcesync.GVRAuthorizationChecker{
+				DownstreamKubeClient: downstreamKubeClient,
+				SyncTargetClient:     kcpSyncTargetClient.WorkloadV1alpha1().SyncTargets(),
+			}
```
sttts, line 742: why does this logic matter for a generic fake syncer fixture? Just set the right condition in the SyncTarget statically and be done with it. No need to simulate permission checks.

sttts: instead have a very dedicated e2e test that creates a syncer (or multiple) only testing this permission logic.

davidfestal: You're right.
I removed this from the PR.
The tests that depend on this still use the full Syncer, as most of themstill expect the resources to be effectively synced.

test/e2e/framework/syncer.go
davidfestal: Renamed it to `SimulateStartWithoutSyncing`

test/e2e/framework/syncer.go
davidfestal: I removed the whole `GVRAuthorizationChecker` trick, and also factorized common code found in the current `Start` method. So I assume this is in a better shape now.
The number of lines specific to this fake Syncer start is effectively now back to about 50 lines.

pkg/syncer/resourcesync/controller.go
davidfestal: The whole `GVRAuthorizationChecker` is removed.

pkg/syncer/resourcesync/controller.go
davidfestal: They were public in order to allow the e2e framework code to reuse them.
But you're right: that was ugly, and finally not as necessary as it initially seemed to be.

I completely removed `GVRAuthorizationChecker` as you suggested [here](https://github.com/kcp-dev/kcp/pull/2730#discussion_r1095689446)

pkg/syncer/resourcesync/controller.go
```diff
@@ -312,6 +314,21 @@ func (c *controller) process(ctx context.Context, key string) error {
...
 	}
 
+	if err := c.PatchSyncTargetCondition(ctx, unauthorizedGVRs, syncTarget); err != nil {
+		errs = append(errs, err)
+	}
+
+	return errors.NewAggregate(errs)
+}
+
+type GVRAuthorizationChecker struct {
```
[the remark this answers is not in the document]
davidfestal: Nothing anymore. I removed it :-)

test/e2e/framework/syncer.go
```diff
@@ -503,6 +523,64 @@ func (sf *syncerFixture) Start(t *testing.T) *StartedSyncerFixture {
...
+
+	kcpBootstrapClusterClient, err := kcpclientset.NewForConfig(syncerConfig.UpstreamConfig)
+	require.NoError(t, err)
+	kcpSyncTargetClient := kcpBootstrapClusterClient.Cluster(syncerConfig.SyncTargetPath)
+
+	// Artificially consider that all the resources to sync are authorized on the physical cluster side.
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		syncTarget, err := kcpSyncTargetClient.WorkloadV1alpha1().SyncTargets().Get(ctx, syncerConfig.SyncTargetName, metav1.GetOptions{})
+		require.NoError(t, err)
+		conditions.Set(syncTarget, conditions.TrueCondition(workloadv1alpha1.SyncerAuthorized))
```
sttts, line 552: am curious where this is needed? Do we depend on conditions in controller or VW logic?

test/e2e/framework/syncer.go
```diff
@@ -503,6 +523,64 @@ func (sf *syncerFixture) Start(t *testing.T) *StartedSyncerFixture {
...
+		kcpImporterInformerFactory.Apiresource().V1alpha1().APIResourceImports(),
+		resources,
+		syncerConfig.SyncTargetPath, syncerConfig.SyncTargetName, types.UID(syncerConfig.SyncTargetUID))
+	require.NoError(t, err)
+
+	kcpImporterInformerFactory.Start(ctx.Done())
+	kcpSyncTargetInformerFactory.Start(ctx.Done())
+	kcpSyncTargetInformerFactory.WaitForCacheSync(ctx.Done())
+
+	go apiImporter.Start(klog.NewContext(ctx, klog.FromContext(ctx).WithValues("resources", resources)), 5*time.Second)
```
sttts, line 578: why is the api importer needed in this fixture?

Could we make it optional in another `StartAPIImporter` method?

I.e. something like `syncerFxture.StartHeartBeat(...).StartAPIImport(...)`?

sttts: Rule of thumb: smaller fixtures

sttts: And a fixture without the need of a downstream cluster would be probably a good idea.

davidfestal: For the SyncTarget to be ready to use, we would expected, in most tests, that the API compatibility check has been run and that the `Accepted` (or `Imcompatible` in some cases) state would be set to every resource of the `synctarget.Status.SyncedResources` array.

There are very few e2e tests that don't even require this information in the SyncTarget, even maybe none.

davidfestal: I have separated the 2 functions as you proposed:

```
syncerFixture.CreateAndStartHeartBeat(t).StartAPIImporter(t)
```

>And a fixture without the need of a downstream cluster would be probably a good idea.

I'd rather keep that for later, since I don't see the need of this right now, and this would involve more coding / refactoring. Do you agree ?

test/e2e/framework/syncer.go
davidfestal: We don't depend on this condition you're right, only in some tests that can then explicitely use the whole Syncer Fixture.
I removed it.

davidfestal:
@jmprusi Is the new ability to call `StopHeartBeat()` and call the `StartHeartBeat()` again later working for your new e2e test case ?

pkg/syncer/syncer.go
```diff
@@ -394,6 +395,14 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
 		go startSyncerTunnel(ctx, upstreamConfig, downstreamConfig, logicalcluster.From(syncTarget), cfg.SyncTargetName)
 	}
 
+	StartHeartbeatKeeper(ctx, kcpSyncTargetClient, cfg.SyncTargetName, cfg.SyncTargetUID)
+
+	return nil
+}
+
+func StartHeartbeatKeeper(ctx context.Context, kcpSyncTargetClient kcpclientset.Interface, syncTargetName, syncTargetUID string) {
```
sttts, line 403: what's a keeper?
image

test/e2e/framework/syncer.go
```diff
@@ -525,9 +560,92 @@ type StartedSyncerFixture struct {
...
+	// Start the heartbeat keeper to have the SyncTarget always ready during the e2e test.
+	syncer.StartHeartbeatKeeper(ctx, kcpSyncTargetClient, sf.SyncerConfig.SyncTargetName, sf.SyncerConfig.SyncTargetUID)
+
+	startedSyncer := &StartedSyncerFixture{
+		sf,
+	}
+
+	// The sync target becoming ready indicates the syncer is healthy and has
+	// successfully sent a heartbeat to kcp.
+	startedSyncer.WaitForClusterReady(ctx, t)
```
sttts, line 590: cluster?

test/e2e/framework/syncer.go
```diff
@@ -525,9 +560,92 @@ type StartedSyncerFixture struct {
...
+	return sf
+}
+
+// StartedSyncerFixture contains the configuration used to start a syncer and interact with its
+// downstream cluster.
+type StartedSyncerFixture struct {
+	*appliedSyncerFixture
+}
+
+// StopHeartBeat stop maitining the heartbeat for this Syncer SyncTarget.
```
sttts, line 639: maintaining

test/e2e/framework/syncer.go
```diff
@@ -127,19 +130,44 @@ func WithDownstreamPreparation(prepare func(config *rest.Config, isFakePCluster
...
 
-// Start starts a new syncer against the given upstream kcp workspace. Whether the syncer run
-// in-process or deployed on a pcluster will depend whether --pcluster-kubeconfig and
-// --syncer-image are supplied to the test invocation.
-func (sf *syncerFixture) Start(t *testing.T) *StartedSyncerFixture {
+// CreateAndStart creates SyncTarget resource, applies it in the physical cluster,
+// and then starts a new syncer against the given upstream kcp workspace.
+// Whether the syncer runs in-process or deployed on a pcluster will depend
+// whether --pcluster-kubeconfig and --syncer-image are supplied to the test invocation.
+func (sf *syncerFixture) CreateAndStart(t *testing.T) *StartedSyncerFixture {
```
sttts, line 137: am confused. How is a SyncTarget applied to the physical cluster?

test/e2e/framework/syncer.go
```diff
@@ -525,9 +560,92 @@ type StartedSyncerFixture struct {
...
+		kcpImporterInformerFactory.Apiresource().V1alpha1().APIResourceImports(),
+		resources,
+		sf.SyncerConfig.SyncTargetPath, sf.SyncerConfig.SyncTargetName, types.UID(sf.SyncerConfig.SyncTargetUID))
+	require.NoError(t, err)
+
+	kcpImporterInformerFactory.Start(ctx.Done())
+	kcpSyncTargetInformerFactory.Start(ctx.Done())
+	kcpSyncTargetInformerFactory.WaitForCacheSync(ctx.Done())
+
+	go apiImporter.Start(klog.NewContext(ctx, klog.FromContext(ctx).WithValues("resources", resources)), 5*time.Second)
```
sttts, line 628: I think I asked before, maybe the comment/answer got lost: why do we have to run a full api importer, why not just set some conditions to fake the outcome?

test/e2e/reconciler/namespace/controller_test.go
```diff
@@ -90,12 +90,9 @@ func TestNamespaceScheduler(t *testing.T) {
...
 				t.Log("Deploy a syncer")
 				// Create and Start a syncer against a workload cluster so that there's a ready cluster to schedule to.
-				//
-				// TODO(marun) Extract the heartbeater out of the syncer for reuse in a test fixture. The namespace
-				// controller just needs ready clusters which can be accomplished without a syncer by having the
-				// heartbeater update the sync target so the heartbeat controller can set the cluster ready.
 				syncerFixture := framework.NewSyncerFixture(t, server, server.path,
-					framework.WithExtraResources("services")).Start(t)
+					framework.WithExtraResources("services"),
+				).Create(t).StartAPIImporter(t).StartHeartBeat(t)
```
sttts, line 95: so "Deploy a syncer" is now wrong?

test/e2e/framework/syncer.go
sttts: This looks a whole lot like `Create(t).StartSyncer(t)`, i.e. it overlaps with `Create`, no?

test/e2e/framework/syncer.go
davidfestal: This is the existing name of the method. I can change it if you prefer, of course, to `WaitForSyncTargetReady`

test/e2e/reconciler/scheduling/api_compatibility_test.go
```diff
@@ -58,14 +58,14 @@ func TestSchedulingOnSupportedAPI(t *testing.T) {
 		framework.WithSyncTargetName(firstSyncTargetName),
 		framework.WithSyncedUserWorkspaces(userWS),
 		framework.WithAPIExports(""),
-	).Start(t)
+	).Create(t).StartAPIImporter(t).StartHeartBeat(t)
```
sttts, line 61: fix Logf above

test/e2e/reconciler/scheduling/api_compatibility_test.go
```diff
@@ -58,14 +58,14 @@ func TestSchedulingOnSupportedAPI(t *testing.T) {
...
 
 	secondSyncTargetName := fmt.Sprintf("secondsynctarget-%d", +rand.Intn(1000000))
 	t.Logf("Creating a SyncTarget with global kubernetes APIExports and syncer in %s", locationPath)
 	_ = framework.NewSyncerFixture(t, source, locationPath,
 		framework.WithSyncTargetName(secondSyncTargetName),
 		framework.WithSyncedUserWorkspaces(userWS),
-	).Start(t)
+	).Create(t).StartAPIImporter(t).StartHeartBeat(t)
```
sttts, line 68: fix Logf above

pkg/syncer/syncer.go
davidfestal: It keeps the heartbeat up-to-date. But I agree it's probably a bad naming.
Changing it to simply `StartHeartBeat()`

test/e2e/framework/syncer.go
davidfestal: Afair you asked about replacing the the whole logic of checking authorization for synced resources in the downstream cluster, which drives the "Authorized" condition. This logic I effectively removed, since the Authorized condition is not checked by any KCP controller.

It seems to me that the API importer is something else: It doesn't drive a direct condition, but allows the compatibility check controller, which runs on the KCP side, to set the `Accepted` status on every resource of the `syncTarget.syncedResources` array.
If we try to set this status manually on the SyncTarget in the test fixture, the KCP controller will very probably reset the values to incompatible until it has the downstream schemas imported. 

See my previous comment: https://github.com/kcp-dev/kcp/pull/2730#discussion_r1097433097

test/e2e/reconciler/namespace/controller_test.go
davidfestal: Oh, just change the logs and comments, sure !

davidfestal: Something like

`Create the SyncTarget and start both the Syncer APIImporter and Syncer HeartBeat`

would make sense to you ?

test/e2e/framework/syncer.go
davidfestal: Yeah, bad wording. The SyncTarget is **not** what is applied. But the Syncer-related resources are applied in the physical cluster.

test/e2e/framework/syncer.go
davidfestal: Fixed in commit 13a4a065c80f0bf3815d69defc0ffcd0728fce2d

test/e2e/framework/syncer.go
davidfestal: Fixed in commit 13a4a065c80f0bf3815d69defc0ffcd0728fce2d

test/e2e/framework/syncer.go
davidfestal: Fixed in commit 13a4a065c80f0bf3815d69defc0ffcd0728fce2d

test/e2e/reconciler/namespace/controller_test.go
davidfestal: Fixed in commit 13a4a065c80f0bf3815d69defc0ffcd0728fce2d

test/e2e/reconciler/scheduling/api_compatibility_test.go
davidfestal: Fixed in commit 13a4a065c80f0bf3815d69defc0ffcd0728fce2d

test/e2e/reconciler/scheduling/api_compatibility_test.go
davidfestal: Fixed in commit 13a4a065c80f0bf3815d69defc0ffcd0728fce2d

pkg/syncer/syncer.go
davidfestal: Fixed in commit 13a4a065c80f0bf3815d69defc0ffcd0728fce2d

test/e2e/framework/syncer.go
jmprusi: Thanks! worked perfectly for my PR! being able to start/stop the heartbeat allows us to test some "complex" scheduling scenarios!

test/e2e/framework/syncer.go
```diff
@@ -503,17 +527,30 @@ func (sf *syncerFixture) Start(t *testing.T) *StartedSyncerFixture {
 		SyncerVirtualWorkspaceConfig:   syncerVWConfig,
 		UpsyncerVirtualWorkspaceConfig: upsyncerVWConfig,
 	}
+}
 
-	// The sync target becoming ready indicates the syncer is healthy and has
-	// successfully sent a heartbeat to kcp.
-	startedSyncer.WaitForClusterReady(ctx, t)
+// CreateAndStart creates a SyncTarget resource through the `workload sync` CLI command,
```
sttts, line 532: wrong comment

test/e2e/framework/syncer.go
sttts: > This looks a whole lot like Create(t).StartSyncer(t), i.e. it overlaps with Create, no?

jmprusi:
> @jmprusi Is the new ability to call `StopHeartBeat()` and call the `StartHeartBeat()` again later working for your new e2e test case ?

Yes, I'm using it for testing the resource scheduler interaction with upsynced resources (https://github.com/kcp-dev/kcp/pull/2533), and being able to stop the syncer and restart is required.

test/e2e/framework/syncer.go
davidfestal: Oops, wrong copy/paste

test/e2e/framework/syncer.go
davidfestal: I changed it to have either:

```
syncerFixture.Create(t).StartSyncer(t)
```
or
```
syncerFixture.Create(t).StartAPIImporter(t).StartHeartBeat(t)
```

test/e2e/framework/syncer.go
davidfestal: Fixed.

test/e2e/framework/syncer.go
```diff
@@ -127,19 +130,30 @@ func WithDownstreamPreparation(prepare func(config *rest.Config, isFakePCluster
...
+		t.Errorf("failed to create temp dir for syncer artifacts: %v", err)
+	}
+
+	useDeployedSyncer := len(TestConfig.PClusterKubeconfig()) > 0
+
+	var (
+		downstreamConfig         *rest.Config
+		downstreamKubeconfigPath string
+		syncerID                 string
+	)
```
sttts, line 150: vars at the top are an anti-pattern. Move them down where they are needed.

test/e2e/framework/syncer.go
```diff
@@ -127,19 +130,24 @@ func WithDownstreamPreparation(prepare func(config *rest.Config, isFakePCluster
...
 }
 
-// Start starts a new syncer against the given upstream kcp workspace. Whether the syncer run
-// in-process or deployed on a pcluster will depend whether --pcluster-kubeconfig and
-// --syncer-image are supplied to the test invocation.
-func (sf *syncerFixture) Start(t *testing.T) *StartedSyncerFixture {
+// Create creates a SyncTarget resource through the `workload sync` CLI command,
+// applies the syncer-related resources in the physical cluster.
+// No resource will be effectively synced after calling this method.
+func (sf *syncerFixture) Create(t *testing.T) *appliedSyncerFixture {
```
sttts, line 136: nit: can we be more descriptive? Is this `CreateSyncTargetAndApplyToDownstream`?

davidfestal: That's it yes, and if the name is not too long for you I'm glad to change it.

davidfestal: Renamed.

PR #2734, opened 2023-02-01
🌱 Add make target to download e2e logs from Prow

ncdc:


stevekuznetsov (commented):
Super nit but if input is unset can you print a usage block

ncdc:
Added

stevekuznetsov (approved):

PR #2735, opened 2023-02-01
🌱 More APIExport VW e2e deflaking

### LINKED ISSUE #2732: bug/flake: authorizer_test.go:437: verify that service-provider-2-admin cannot lists CRD shadowed sherriffs resources in the tenant workspace "root:e2e-workspace-nvqzj:tenant-shadowed-crd" via the virtual apiexport apiserver
From https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kcp-dev_kcp/2423/pull-ci-kcp-dev-kcp-main-e2e-sharded/1620860539901054976

ncdc:
Wrap creating an APIBinding in an Eventually block to give the system time to be eventually consistent.

Change the expected error when service provider 2 tries to list cowboys in a workspace where the APIBinding shadowed a CRD.

Fixes #2732

### DIFF
--- a/test/e2e/virtual/apiexport/authorizer_test.go
+++ b/test/e2e/virtual/apiexport/authorizer_test.go
@@ -298,38 +298,44 @@ func TestAPIExportAuthorizers(t *testing.T) {
 	}, wait.ForeverTestTimeout, time.Millisecond*100, "waiting for cowboys CRD to become established failed")
 
 	t.Logf("Create a cowboys APIBinding in consumer workspace %q that points to the today-cowboys export from %q but shadows a local cowboys CRD at the same time", tenantShadowCRDPath, serviceProvider2Path)
-	require.NoError(t, apply(t, ctx, tenantShadowCRDPath, tenantUser,
-		&apisv1alpha1.APIBinding{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: "cowboys",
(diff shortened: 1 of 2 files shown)


test/e2e/virtual/apiexport/authorizer_test.go
```diff
@@ -453,8 +459,8 @@ metadata:
 		if err == nil {
 			return false, "expected error, got none"
 		}
-		if !apierrors.IsNotFound(err) {
-			return false, fmt.Sprintf("expected a not-found error, but got %v", err)
+		if !apierrors.IsForbidden(err) {
```
ncdc, line 462: @s-urbaniak @sttts not sure if this is right, but when it fails in CI, it gets forbidden for wait.ForeverTestTimeout. So I'm guessing the NotFound happens initially (or never) and then transitions to forbidden (eventual consistency). Would like to understand the "why" though.

s-urbaniak: agreed, this seems to be a temporary state before final replication of logical clusters finalizes and needs investigation.

s-urbaniak: correction: if the "Forbidden" is a temporary state than it should converge against "IsNotFound" which is not the case.

PR #2744, opened 2023-02-02
🌱 Make syncer vw global

ncdc:


pkg/virtual/syncer/builder/template.go
```diff
@@ -55,7 +55,7 @@ type templateProvider struct {
 	kubeClusterClient    kcpkubernetesclientset.ClusterInterface
```
davidfestal, line 55: Should we do something with this as well ? It is used to call the SAR in the authorizer, and we need replicated service accounts here (cf [this discussion](https://kubernetes.slack.com/archives/C021U8WSAFK/p1675252445541419?thread_ts=1675251008.266109&cid=C021U8WSAFK) and followings).
But to work this will require the work from @s-urbaniak about replicating some serviceaccounts.

pkg/virtual/syncer/builder/template.go
```diff
@@ -55,7 +55,7 @@ type templateProvider struct {
...
 	dynamicClusterClient kcpdynamic.ClusterInterface
 	kcpClusterClient     kcpclientset.ClusterInterface
```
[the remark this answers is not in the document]
ncdc: Removed

pkg/virtual/syncer/builder/template.go
ncdc: We don't need to vary this client. It's always talking to its shard.

pkg/reconciler/cache/replication/replication_controller.go
```diff
@@ -114,12 +116,12 @@ func NewController(
 				local:  localKcpInformers.Tenancy().V1alpha1().WorkspaceTypes().Informer(),
 				global: globalKcpInformers.Tenancy().V1alpha1().WorkspaceTypes().Informer(),
 			},
-			tenancyv1alpha1.SchemeGroupVersion.WithResource("synctargets"): {
+			workloadv1alpha1.SchemeGroupVersion.WithResource("synctargets"): {
```
stevekuznetsov, line 119: Why did this not break anything in the past ... ?

stevekuznetsov (approved):

pkg/reconciler/cache/replication/replication_controller.go
ncdc: I assume nothing was using these from the cache server

pkg/reconciler/cache/replication/replication_controller.go
```diff
@@ -114,12 +116,12 @@ func NewController(
...
 				kind:   "SyncTarget",
 				local:  localKcpInformers.Workload().V1alpha1().SyncTargets().Informer(),
 				global: globalKcpInformers.Workload().V1alpha1().SyncTargets().Informer(),
 			},
-			tenancyv1alpha1.SchemeGroupVersion.WithResource("locations"): {
+			schedulingv1alpha1.SchemeGroupVersion.WithResource("locations"): {
```
sttts, line 124: didn't we say in slack to not replicate locations?

ncdc: I can remove it. Was mainly focused on fixing the typo.

davidfestal:
Apparently failing test is https://github.com/kcp-dev/kcp/issues/2762 which has been fixed in the meantime.
Let's retest.

ncdc:
Probably worth tackling the flakes...

davidfestal:
@ncdc I find it strange that this tests fails constantly here, while PR #2767 didn't have it even after running the tests several times.
Could it be related to some change in this PR ?

davidfestal:
this e2e error mainly means that the expected api-resources that should be exposed by the Syncer Virtual Workspace are not in some cases.

ncdc:
Entirely possible. I can try to look later

PR #2747, opened 2023-02-03
🌱 /pkg/proxy/server: handle /readyz and /livez outside of the auth chain

s-urbaniak:
## Summary
To simplify deployment we want the `/readyz` and `/livez` not be handled by the regular auth chain. This allows us to use simple httpGet kube probes.

## Related issue(s)

Related helm chart update: https://github.com/kcp-dev/helm-charts/pull/24

### DIFF
--- a/pkg/proxy/mapping.go
+++ b/pkg/proxy/mapping.go
@@ -59,18 +59,6 @@ func NewHandler(ctx context.Context, o *proxyoptions.Options, index index.Index)
 
 	mux := http.NewServeMux()
 
-	// TODO: implement proper readyz handler
-	mux.Handle("/readyz", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


s-urbaniak:
cc @sttts @hardys

hardys:
/retest

failed to clone a repo so presumably temporary network disruption

hardys:
/lgtm

Thanks for picking this up @s-urbaniak - for historical context, I didn't do this in https://github.com/kcp-dev/kcp/pull/2178 because I was thinking we don't want any unauthenticated access to the proxy at all.

However I didn't consider liveness probes at that point (or that `httpGet` doesn't support any kind of authentication), and after discussion it seems reasonable to expose these given there is a very low cost for hitting them (probably less or comparable to rejecting an unauthenticated request to any other path)

ncdc:
#2762
/retest

ncdc:
PV flake, which has been fixed
/retest

PR #2749, opened 2023-02-03
:book: Fix unknown flag --apiexport

kasturinarra:
## Summary
In one of the kcp bind compute --apiexports has been added as --apiexport and if user tries to copy paste the same into his command line it throws an error which reads "unknown flag --apiexport" so fixing the same in docs
## Related issue(s)

Fixes #


kasturinarra:
@davidfestal @ncdc could some one please help review and merge ? thanks !!

davidfestal (approved):

PR #2753, opened 2023-02-06
🌱 Small improvements to dev workflow

fabriziopandini:
## Summary

Proposing a small change to the Makefile so it is possible to do 

KIND_CLUSTER_NAME=foo build-kind-images 

Which is nice to have when testing scheduling on multiple clusters.

Also, proposing to have a tmp folder in .gitignore where each developer can store temporary files, and get rid of .DS_Store files on mac.

## Related issue(s)

Fixes #

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,8 +5,17 @@ contrib/demo/clusters/kind/*.kubeconfig
 coverage.*
 tools
 build
-.vscode
 
 # JetBrains IDE files
 .idea/
 *.iml
+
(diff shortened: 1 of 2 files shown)


ncdc:
/ok-to-test
/approve

Just a couple of minor nits

fabriziopandini:
@ncdc thanks for the quick feedback!
fixed

ncdc (approved):

Issue #2665, opened 2023-01-23
flake: TestGarbageCollectorTypesFromBinding


hardys:
### Describe the bug

Seeing [failures](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kcp-dev_kcp/2659/pull-ci-kcp-dev-kcp-main-e2e-shared/1617467644934361088) like: 

```
{Failed  === RUN   TestGarbageCollectorTypesFromBinding/tc1
=== PAUSE TestGarbageCollectorTypesFromBinding/tc1
=== CONT  TestGarbageCollectorTypesFromBinding/tc1
=== CONT  TestGarbageCollectorTypesFromBinding/tc1
    garbagecollector_test.go:153: Created root:universal workspace root:e2e-workspace-xm267:gc-api-binding-1 as /clusters/2hgmlhj8joun35pp
    garbagecollector_test.go:155: Create a binding in the user workspace
    garbagecollector_test.go:179: Wait for the binding to be ready
=== CONT  TestGarbageCollectorTypesFromBinding/tc1
    garbagecollector_test.go:196: Wait for being able to list cowboys in the user workspace
=== CONT  TestGarbageCollectorTypesFromBinding/tc1
    garbagecollector_test.go:206: Creating owner cowboy
=== CONT  TestGarbageCollectorTypesFromBinding/tc1
    garbagecollector_test.go:217: Creating owned configmap
=== CONT  TestGarbageCollectorTypesFromBinding/tc1
    garbagecollector_test.go:228: Creating owned cowboy
    garbagecollector_test.go:247: Deleting owner cowboy
=== CONT  TestGarbageCollectorTypesFromBinding/tc1
    garbagecollector_test.go:251: Waiting for the owned configmap to be garbage collected
=== CONT  TestGarbageCollectorTypesFromBinding/tc1
    assertions.go:1691: Waiting for condition, but got: configmap not garbage collected: owned
=== CONT  TestGarbageCollectorTypesFromBinding/tc1
    garbagecollector_test.go:252: 
        	Error Trace:	util.go:318
        	            				garbagecollector_test.go:252
        	Error:      	Condition never satisfied
        	Test:       	TestGarbageCollectorTypesFromBinding/tc1
        	Messages:   	error waiting for owned configmap to be garbage collected
    --- FAIL: TestGarbageCollectorTypesFromBinding/tc1 (39.68s)
```

`TestGarbageCollectorNormalCRDs` also fails in a similar way

[...]

Issue #1878, opened 2022-09-01
test flake: TestKubeQuotaCoreV1TypesFromBinding


ncdc:
From https://github.com/kcp-dev/kcp/runs/8137476634?check_suite_focus=true

Note that `Waiting for condition, but got: 0` is a misnomer: the zero value when reading from the map is `0`, but we check for key existence + value, so it's just the log message that's "wrong".

```
2022-09-01T14:39:37.5381433Z --- FAIL: TestKubeQuotaCoreV1TypesFromBinding (0.00s)
2022-09-01T14:39:37.5381978Z     --- FAIL: TestKubeQuotaCoreV1TypesFromBinding/tc4 (35.59s)
2022-09-01T14:39:37.5385011Z         kcp.go:106: shared kcp server will target configuration "/home/runner/work/kcp/kcp/.kcp/admin.kubeconfig"
2022-09-01T14:39:37.5385816Z         workspaces.go:191: Created organization workspace root:e2e-org-4gcvr
2022-09-01T14:39:37.5386752Z         workspaces.go:124: Created root:universal workspace root:e2e-org-4gcvr:e2e-workspace-5f8tr
2022-09-01T14:39:37.5387510Z         workspaces.go:124: Created root:universal workspace root:e2e-org-4gcvr:e2e-workspace-7jdpp
2022-09-01T14:39:37.5388290Z         quota_test.go:128: Check that there is no services resource in the user workspace
2022-09-01T14:39:37.5388747Z         quota_test.go:132: Getting services CRD
2022-09-01T14:39:37.5389246Z         quota_test.go:135: Converting services CRD to APIResourceSchema
2022-09-01T14:39:37.5389717Z         quota_test.go:139: Creating APIResourceSchema
2022-09-01T14:39:37.5390110Z         quota_test.go:143: Creating APIExport
2022-09-01T14:39:37.5390566Z         quota_test.go:158: Create a binding in the user workspace
2022-09-01T14:39:37.5391011Z         quota_test.go:176: Wait for binding to be ready
2022-09-01T14:39:37.5391524Z         quota_test.go:183: Wait for being able to list Services in the user workspace
2022-09-01T14:39:37.5392006Z         quota_test.go:192: Create quota in user workspace
2022-09-01T14:39:37.5392482Z         quota_test.go:207: Waiting for quota to show 0 used Services
2022-09-01T14:39:37.5392967Z         assertions.go:1691: Waiting for condition, but got: 0
[...]

PR #2764, opened 2023-02-08
🐛 Prevent gc/quota workqueue backlogs

ncdc:
## Summary
Run the initial monitor syncing in a goroutine to avoid any stalling here to block a worker.

## Related issue(s)

Maybe #1878 #2665

### DIFF
--- a/pkg/reconciler/garbagecollector/garbagecollector_controller.go
+++ b/pkg/reconciler/garbagecollector/garbagecollector_controller.go
@@ -290,10 +290,13 @@ func (c *Controller) startGarbageCollectorForLogicalCluster(ctx context.Context,
 		}
 	}()
 
-	// Make sure the GC monitors are synced at least once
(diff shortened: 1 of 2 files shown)


ncdc:
https://github.com/kcp-dev/kubernetes/blob/f48a2e575b3bf96221060778f57689972c3709b1/pkg/controller/resourcequota/resource_quota_controller_patch.go#L80 for quota (and similarly for gc) is blocking and never ends because the `ctx` we use with `resourceQuotaController.UpdateMonitors(ctx, c.dynamicDiscoverySharedInformerFactory.ServerPreferredResources)` has no timeout. We could consider giving it one?

ncdc:
@sttts interested in your thoughts

nrb:
/lgtm

Adding a timeout on the context sounds good to me; is there really a benefit to leaving it off?

ncdc:
Not really, upstream's is 30 seconds IIRC. I guess I thought we wouldn't have an issue since we don't have API aggregation...

PR #2769, opened 2023-02-08
🌱 Fix git in Prow

ncdc:
## Summary
Newer versions of git have a security measure in place to prevent git from parsing the .git/config file or running any hooks if the user running git is not the owner of the repository files. Prow runs as a random UID, which is not guaranteed to be the same as the UID that owns the cloned files, so this change disables the ownership check and considers all cloned repositories "safe" regardless of owner.

## Related issue(s)

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,14 @@
 # We need bash for some conditional logic below.
 SHELL := /usr/bin/env bash -e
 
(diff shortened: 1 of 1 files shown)


ncdc:
Looks like this fixed the issue. Self-merging to unblock other PRs.

Issue #2762, opened 2023-02-08
flake: TestUpsyncerVirtualWorkspace/update_a_persistentvolume_in_kcp_through_upsyncer_virtual_workspace,_try_to_remove_the_upsync_state_label,_expect_error


ncdc:
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kcp-dev_kcp/2747/pull-ci-kcp-dev-kcp-main-e2e-sharded/1623282406905614336

```
=== RUN   TestUpsyncerVirtualWorkspace/update_a_persistentvolume_in_kcp_through_upsyncer_virtual_workspace,_try_to_remove_the_upsync_state_label,_expect_error.
=== PAUSE TestUpsyncerVirtualWorkspace/update_a_persistentvolume_in_kcp_through_upsyncer_virtual_workspace,_try_to_remove_the_upsync_state_label,_expect_error.
=== CONT  TestUpsyncerVirtualWorkspace/update_a_persistentvolume_in_kcp_through_upsyncer_virtual_workspace,_try_to_remove_the_upsync_state_label,_expect_error.
=== CONT  TestUpsyncerVirtualWorkspace/update_a_persistentvolume_in_kcp_through_upsyncer_virtual_workspace,_try_to_remove_the_upsync_state_label,_expect_error.
    assertions.go:1691: Waiting for condition, but got: workspace phase is Initializing, not Ready
=== CONT  TestUpsyncerVirtualWorkspace/update_a_persistentvolume_in_kcp_through_upsyncer_virtual_workspace,_try_to_remove_the_upsync_state_label,_expect_error.
    virtualworkspace_test.go:1928: Created root:organization workspace root:e2e-workspace-rr7tp as /clusters/1frk2srbeaqtnxsu on shard "root"
=== CONT  TestUpsyncerVirtualWorkspace/update_a_persistentvolume_in_kcp_through_upsyncer_virtual_workspace,_try_to_remove_the_upsync_state_label,_expect_error.
    virtualworkspace_test.go:1930: Created root:universal workspace root:e2e-workspace-rr7tp:upsyncer as /clusters/2d8bqqfioo3e85eh on shard "root"
    virtualworkspace_test.go:1933: [11:53:56.179579] Deploying syncer into workspace root:e2e-workspace-rr7tp:upsyncer
[...]

PR #2780, opened 2023-02-10
🌱 syncer vw: add API definition logging

ncdc:
Trying to figure out #2762

### DIFF
--- a/pkg/virtual/syncer/controllers/apireconciler/syncer_apireconciler_indexes.go
+++ b/pkg/virtual/syncer/controllers/apireconciler/syncer_apireconciler_indexes.go
@@ -17,8 +17,6 @@ limitations under the License.
 package apireconciler
 
 import (
-	"fmt"
-
 	"github.com/kcp-dev/logicalcluster/v3"
 
 	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/virtual/syncer/controllers/apireconciler/syncer_apireconciler_indexes.go
```diff
@@ -43,23 +38,20 @@ func IndexAPIExportsByAPIResourceSchemas(obj interface{}) ([]string, error) {
...
-	clusterName := logicalcluster.From(synctarget)
-	if len(synctarget.Spec.SupportedAPIExports) == 0 {
+	clusterName := logicalcluster.From(syncTarget)
+	if len(syncTarget.Spec.SupportedAPIExports) == 0 {
 		return []string{client.ToClusterAwareKey(clusterName.Path(), reconcilerapiexport.TemporaryComputeServiceExportName)}, nil
 	}
 
-	keys := make([]string, 0, len(synctarget.Spec.SupportedAPIExports))
-	for _, export := range synctarget.Spec.SupportedAPIExports {
-		if len(export.Export) == 0 {
```
ncdc, line 58: FYI @davidfestal @sttts I'm pretty sure this is a bug, right?

sttts: looks like, yes

ncdc: There is a _chance_ this is the root cause of #2762

davidfestal: I agree it seems a bug that might be related to the mentioned issue.

PR #2782, opened 2023-02-13
🌱 workspace_reconcile_scheduling: allow for skipping a shard with a special annotation during scheduling

p0lyn0mial:
## Summary
introduces an annotation that can be used to indicate that a shard is unschedulable. The annotation is meant to be used by e2e tests that otherwise started a private instance of kcp server.

I decided to go with an annotation because it is easier to change/delete than an API field.

## Related issue(s)

Fixes #

### DIFF
--- a/pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling.go
+++ b/pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling.go
@@ -53,8 +53,13 @@ const (
 	// WorkspaceShardHashAnnotationKey keeps track on which shard LogicalCluster must be scheduled. The value
 	// is a base36(sha224) hash of the Shard name.
 	WorkspaceShardHashAnnotationKey = "internal.tenancy.kcp.io/shard"
+
 	// workspaceClusterAnnotationKey keeps track of the logical cluster on the shard.
 	workspaceClusterAnnotationKey = "internal.tenancy.kcp.io/cluster"
+
+	// unschedulableAnnotationKey is the annotation key used to indicate that a shard is unschedulable.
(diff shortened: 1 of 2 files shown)


pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling.go
```diff
@@ -53,8 +53,13 @@ const (
 	// WorkspaceShardHashAnnotationKey keeps track on which shard LogicalCluster must be scheduled. The value
 	// is a base36(sha224) hash of the Shard name.
 	WorkspaceShardHashAnnotationKey = "internal.tenancy.kcp.io/shard"
+
 	// workspaceClusterAnnotationKey keeps track of the logical cluster on the shard.
 	workspaceClusterAnnotationKey = "internal.tenancy.kcp.io/cluster"
+
+	// unschedulableAnnotationKey is the annotation key used to indicate that a shard is unschedulable.
+	// The annotation is meant to be used by e2e tests that otherwise started a private instance of kcp server.
+	unschedulableAnnotationKey = "internal.kcp.io/unschedulable"
```
sttts, line 62: `experimental.core.kcp.io/unschedulable`

p0lyn0mial: done

fgiloux:
An alternative would have been to amend the e2e framework so that the initial shards have a specific label. The fixture used to create workspaces in e2e tests could then use this label for scheduling when no different configuration is provided. With this approach there would be no API or pseudo-API change.

sttts:
> An alternative would have been to amend the e2e framework so that the initial shards have a specific label

We also have e.g. cli invocations. We cannot easily change those.

PR #2784, opened 2023-02-14
🐛 Fix location cannot be found by placement during scheduling

qiujian16:

## Summary

It happens when setting the alternative workspace name on placement, which needs to be considered during scheduling

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/2737https://github.com/kcp-dev/kcp/issues/2737

### DIFF
--- a/test/e2e/reconciler/scheduling/controller_test.go
+++ b/test/e2e/reconciler/scheduling/controller_test.go
@@ -34,6 +34,7 @@ import (
 
 	apisv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/apis/v1alpha1"
 	schedulingv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/scheduling/v1alpha1"
+	"github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/util/conditions"
 	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
 	kcpclientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned/cluster"
 	"github.com/kcp-dev/kcp/test/e2e/framework"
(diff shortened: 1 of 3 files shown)


davidfestal (changes requested):
Shouldn't we add either a unit test or an e2e test related to this fix and the previously-failing scenario ?

qiujian16:
> Shouldn't we add either a unit test or an e2e test related to this fix and the previously-failing scenario ?

yeah, this is reason I am holding this PR

qiujian16:
e2e added
/unhold

test/e2e/reconciler/scheduling/controller_test.go
```diff
@@ -257,3 +258,63 @@ func TestScheduling(t *testing.T) {
...
+		return placement.Status.Phase == schedulingv1alpha1.PlacementPending, ""
+	}, wait.ForeverTestTimeout, time.Millisecond*100)
+
+	syncTargetName := "synctarget"
+	t.Logf("Creating a SyncTarget and syncer in %s", locationPath)
+	_ = framework.NewSyncerFixture(t, source, locationPath,
+		framework.WithExtraResources("services"),
+		framework.WithSyncTargetName(syncTargetName),
+		framework.WithSyncedUserWorkspaces(userWorkspace),
+	).CreateSyncTargetAndApplyToDownstream(t).StartSyncer(t)
```
davidfestal, line 314: You don't need to use `StartSyncer()` here, which will start the whole syncer, and start effective syncing (rather heavy and invasive).
Instead you could just call the `StartAPIImporter().StartHearBeat()`.

Going 1 step further you could even call the `StartAPIImporter` first, and wait for the `APIImporterReady` condition , and test that the placement is not ready (because the Heartbeat is not set), and then call `StartHeartBeat`, after which the placement will finally be updated.

This also allows testing the case when the heartbeat stops (Syncer stopped or unreachable) and comes back, a how the placements follows.

davidfestal (approved):

PR #2787, opened 2023-02-15
🌱 Add the kcp playground plugin

fabriziopandini:
## Summary
As per discussion in the office hours meeting from Feb 14th, proposing a new kubectl plugin that will allow users/developers to spin un playgrounds with kcp, pclusters, and an initial setup defined in a config file similar to the example included in the PR.
 
The playground is implemented as a test fixture and internally leverages all the methods of the existing test framework (kudos to all the developers), so in the future, it will be possible to consider if to use the playgroundFixture as an alternative way to do the initial setup in E2E tests.
But this is an eventual long-term goal and I'm quoting now only for giving a little bit of context on the proposed code structure.

@ncdc if you want to s/playground/testbed let me know that I will do this before people start reviewing (no strong opinions from my side, so I'm ok with whatever the team prefers)

## Related issue(s)

Fixes #


ncdc:
@fabriziopandini ❤️ 

playground is fine for now 😄 

/ok-to-test

vincepri (commented):
<3 Thank you for this!

Would it be possible to have a "quick start" sort of document in our new book that shows how to use the Playground?

kasturinarra:
> <3 Thank you for this!
> 
> Would it be possible to have a "quick start" sort of document in our new book that shows how to use the Playground?

+1

fabriziopandini:
Add some more capabilities and examples covering all the tutorials in the book and more (cowboys API, east-west placement with or without coordination controller).

I will follow up adding a page in the book explaining how to use the playground.

fabriziopandini:
@ncdc @vincepri @kasturinarra I have added doc and cleaned up examples
I think the feature set has now a good critical mass, so I will stop development and wait for feedback on what is in

docs/content/en/developers/kcp-playground.md
```diff
@@ -0,0 +1,331 @@
+---
+linkTitle: "kcp playground"
+description: >
+  Alternative approach for testing KCP
+---
+
+# KCP playground
+
+The kcp playground is a kubectl plugin that allows users/developers to spin un playgrounds with kcp, pclusters,
```
sttts, line 9:
```suggestion
The kcp playground is a kubectl plugin that allows users/developers to spin up playgrounds with kcp, pclusters, 
```

sttts:
How are the playgrounds tested? I fear a similar situation as with the demos we had long time ago, i.e. they broke eventually and it was just too much to maintain them.

MikeSpreitzer:
Is there a story here for developers of the stuff that playground runs (e.g., the `kcp` process, the syncer)?

docs/content/en/developers/kcp-playground.md
```diff
@@ -0,0 +1,331 @@
...
+kubectl kcp playground use pcluster us-east1
+kubectl get cowboys -A
+```
+
+Take your time to look at this setup to understand how API exports and workload placement can be used in kcp.
+
+## Build your own playgrounds
+
+You can build your own playgrounds by starting from one of the above config files
+or defining a new config from scratch using the playground config format described below.
```
sttts, line 177: can't we reference a file with yaml comments and test that automatically? A giant yaml blob here will eventually break.

MikeSpreitzer:
For example, consider a developer of the `kcp` or syncer binaries that wants to debug it in their IDE, using breakpoints and examining call stacks and data structures.  How does that happen?

ncdc:
I use Goland today (and used VSCode in the past) to launch and debug kcp. You should be able to do something similar for the syncer. I don't believe this PR is trying to help in that area (?).

MikeSpreitzer:
My point is that while developers are familiar with how to launch and debug using their favorite IDE, the playground launches processes without debugging, independent of any IDE.

docs/content/en/developers/kcp-playground.md
fabriziopandini: Done!
The test loads the example config file and checks it is valid

fabriziopandini:
> For example, consider a developer of the kcp or syncer binaries that wants to debug it in their IDE, using breakpoints and examining call stacks and data structures. How does that happen?

The playground is not a replacement for an IDE, and it was never intended to be.

However, nothing prevents you to use the playground in combination with an IDE, as a way to quickly set up a scenario with a test of objects. This works with some caveats because you are using two tools that are not aware of one of the other, but some time is useful so I'm using it mostly in the following configurations:
- I start KCP in debug mode in my IDE, then use the playground to quickly create a set of objects on the KCP instance managed by the IDE
- I start a playground (KCP + a set of objects), then start a controller I'm playing with in my IDE pointing to the playground; I think the same can be done with the syncer too.

test/kubectl-kcp-playground/plugin/fixture.go
```diff
@@ -0,0 +1,688 @@
...
+	"github.com/kcp-dev/kcp/pkg/apis/core"
+	schedulingv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/scheduling/v1alpha1"
+	tenancyv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/tenancy/v1alpha1"
+	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
+	kcpclientset "github.com/kcp-dev/kcp/pkg/client/clientset/versioned/cluster"
+	"github.com/kcp-dev/kcp/test/e2e/framework"
+)
+
+const (
+	defaultRootDirectory = ".kcp-playground"
```
nikhita, line 59: The docs ask to use `export KUBECONFIG=.kcp/playground.kubeconfig`.

Either this line needs to be updated to `.kcp` or the docs need to use `.kcp-playground`.

fabriziopandini:
/close
Sorry but currently I don't have the bandwidth to follow up with this work, if someone is interested I will be happy to hand over it

PR #2803, opened 2023-02-16
[WIP] :seedling: Add per-service dockerfiles and update build workflow

hardys:
## Summary

Currently we build a single large kcp image, which contains multiple binaries, and also client binaries (kubectl and plugins) which aren't required for deployment on a cluster.

However for development the "all in one" Dockerfile is useful, so we can split it into stages, then use a common builder stage for multiple per-service images.

The GH workflow currently only publishes the kcp and syncer image (via two workflows), so that's adjusted here to align with the new Makefile target, this should enable smaller images for kcp cluster deployments, and add missing images e.g cache-server which are required for sharded deployments (which isn't currently supported e.g via https://github.com/kcp-dev/helm-charts/


hardys:
/cc @ncdc 

One open question - this moves the `kcp` image to the minimal kcp-only Dockerfile, are we OK with that, and should we perhaps still publish the "all in one" root Dockerfile image e.g as `kcp-dev:latest` or similar?

Another option is to rename the kcp-only image, since we'll have to modify the [helm chart images](https://github.com/kcp-dev/helm-charts/blob/main/charts/kcp/values.yaml#L32) anyway since the `kcp` image will no longer contain both the `kcp` and `kcp-front-proxy` binaries

hardys:
Also @stevekuznetsov we [previously discussed](https://github.com/kcp-dev/helm-charts/pull/24#discussion_r1089108007) the GH images vs the prow ones 

It seems to me that the GH images are more easily discoverable for upstream users, so maintaining them has some valie provided it's not much overhead to maintain the workflow/Dockerfiles?

My intention will be to reconfigure prow so it builds using these same Dockerfiles, then we'd have two image locations, but the content would be the same (which seems like a better situation that currently, where the prow image content and entrypoint is different, and also not derived from the in-tree Dockerfile)

hardys:
Hmm it looks like we can't do multiplatform image builds while also using `load` to decouple the build from generating/pushingeach image ref https://github.com/docker/roadmap/issues/371 :thinking: 

I guess that means we do need a workflow for each image, and probably a copy/paste Dockerfile builder stage to build in every one :disappointed:

images/Dockerfile.cache-server
```diff
@@ -0,0 +1,6 @@
+# Use distroless as minimal base image to package the manager binary
```
vincepri, line 1: Should we invert the name of these files so we keep syntax highlighting and such? e.g. `cache-server.Dockerfile?

images/Dockerfile.kcp-front-proxy
```diff
@@ -0,0 +1,6 @@
+# Use distroless as minimal base image to package the manager binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM gcr.io/distroless/static:debug
+WORKDIR /
+COPY --from=kcp-build workspace/bin/kcp-front-proxy /
+ENTRYPOINT ["/kcp-front-proxy"]
```
vincepri, line 6: These Dockerfile all look very similar, could it be standardized in a way that we pass the names of the binary to build and the path as an argument?

images/Dockerfile.cache-server
hardys: We could, or perhaps a subdirectory would be clearer e.g `images/cache-server/Dockerfile`?

images/Dockerfile.kcp-front-proxy
hardys: Yes I thought the same, we could but I wasn't completely sure if we'll end up with any binary-specific additions, such as those already added for the workdir etc to the kcp image - if not we could indeed use a single file with a per-binary build arg.

Issue #1975, opened 2022-09-14
implement pod access authorization in the Syncer itself.


davidfestal:
Currently the syncer trust blindly the proxied request sent via the tunneler reverse connection, this can be a security issue. 

There should be a validation at the syncer side that ensures that the proxied requests targets:

- A pod subresource
- On a valid namespace, owned by the syncer
- On a valid POD, that is owned by the syncer and in Upsynced state.

PR #2805, opened 2023-02-16
✨ Provide access to physical cluster pod definitions from upstream KCP, relying on upsyncing

### LINKED ISSUE #1977: Provide access to physical cluster pod definitions from upstream KCP, possibly relying on inverse syncing

davidfestal:
## Summary

This PR leverages the Upsyncer to bring POD definitions of synced deployments, at the KCP level.

## Related issue(s)

Fixes #1977


pkg/features/kcp_features.go
```diff
@@ -97,7 +97,7 @@ func (f *kcpFeatureGate) Type() string {
 // here. The features will be available throughout Kubernetes binaries.
 var defaultGenericControlPlaneFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
 	LocationAPI:  {Default: true, PreRelease: featuregate.Alpha},
-	SyncerTunnel: {Default: false, PreRelease: featuregate.Alpha},
+	SyncerTunnel: {Default: true, PreRelease: featuregate.Alpha},
```
sttts, line 100: all security is in-place?

pkg/syncer/spec/mutators/deployment.go
```diff
@@ -44,7 +46,9 @@ type DeploymentMutator struct {
 	syncTargetClusterName logicalcluster.Name
 	syncTargetUID         types.UID
 	syncTargetName        string
+	syncTargetKey         string
```
sttts, line 49: key in which sense?

davidfestal: The one used in labels, and annotations. I'll comment.

pkg/features/kcp_features.go
davidfestal: No, you're right, after discussing it, we still need to implement https://github.com/kcp-dev/kcp/issues/1975.
I'll remove the feature flag switch commit from this PR

pkg/syncer/spec/mutators/deployment.go
davidfestal: I finally simplified the code so that this field is not required anymore.

pkg/features/kcp_features.go
davidfestal: Done

PR #2815, opened 2023-02-19
🌱 Add --labels to kubectl kcp workload sync

fabriziopandini:
## Summary
This PR adds a --labels flag to be used for setting labels in the SyncTarget created by `kcp workload sync`, thus allowing to easier to test scenarios like the east/west workload placement described in [this PR](https://github.com/kcp-dev/kcp/pull/2336) (it is not required anymore to patch SyncTarget after they are created).

The same capability is added to the `syncerFixture` in the test framework

### DIFF
--- a/pkg/cliplugins/workload/plugin/sync.go
+++ b/pkg/cliplugins/workload/plugin/sync.go
@@ -96,6 +96,8 @@ type SyncOptions struct {
 	Burst int
 	// SyncTargetName is the name of the SyncTarget in the kcp workspace.
 	SyncTargetName string
+	// SyncTargetLabels are the labels to be applied to the SyncTarget in the kcp workspace.
+	SyncTargetLabels string
 	// APIImportPollInterval is the time interval to push apiimport.
 	APIImportPollInterval time.Duration
 	// FeatureGates is used to configure which feature gates are enabled.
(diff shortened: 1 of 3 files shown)


pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -140,6 +142,7 @@ func (o *SyncOptions) BindFlags(cmd *cobra.Command) {
 			"Options are:\n"+strings.Join(kcpfeatures.KnownFeatures(), "\n")) // hide kube-only gates
 	cmd.Flags().DurationVar(&o.APIImportPollInterval, "api-import-poll-interval", o.APIImportPollInterval, "Polling interval for API import.")
 	cmd.Flags().DurationVar(&o.DownstreamNamespaceCleanDelay, "downstream-namespace-clean-delay", o.DownstreamNamespaceCleanDelay, "Time to wait before deleting a downstream namespaces.")
+	cmd.Flags().StringVar(&o.SyncTargetLabels, "labels", o.SyncTargetLabels, "A set of key=value pairs to be used as a labels for the sync target in kcp")
```
davidfestal, line 145:
```suggestion
	cmd.Flags().StringVar(&o.SyncTargetLabels, "labels", o.SyncTargetLabels, "A set of key=value pairs to be used as labels on the SyncTarget created in kcp")
```

davidfestal: Should we also mention the fact that they would be comma-separated ?

davidfestal: Also why not use a `StringSliceVar` as for `apiexports` or `resources` arguments ?
This would take care of the comma-separated part of it, as the doc mentions

> 
> ```
> --ss="v1,v2" --ss="v3"
> will result in
> 
> []string{"v1", "v2", "v3"}
> ```

davidfestal (commented):
Thanks for this PR !
Just a few minor comments

pkg/cliplugins/workload/plugin/sync.go
fabriziopandini: good suggestion, done!
now you can do
```
kubectl kcp workload sync west --replicas=0 --syncer-image=syncer --output-file=west-syncer.yaml \
     --labels=region=east,foo=bar \
     --labels=bar=baz
```

fabriziopandini:
comment addressed, test errors seem a flake but I will check again after tests for this latest commit completes

pkg/cliplugins/workload/plugin/sync.go
davidfestal: Nice

davidfestal (approved):

PR #2816, opened 2023-02-20
:bug: cli: honor expanded `~` in kcp ws

stevekuznetsov:
Today, a user has to write `kubectl kcp ws '~'` to go home, as most shells will expand `~` to the home directory path before passing the argument to our CLI. This is awkward and we can do better.

We know that the literal content of the user's home directory cannot be a valid workspace name, since we don't allow path separators in workspace names. So, we can just handle the case where the argument we got *is* the home directory just like we handle it being `~`.

/assign @mjudeikis


mjudeikis:
/approve
/lgtm 
:+1:

PR #2821, opened 2023-02-21
🌱 e2e/apiexportendpointslice: use SharedKcpServer

p0lyn0mial:
## Summary
it is possible to use a shared server because the
scheduler skips shards annotated with `experimental.core.kcp.io/unschedulable`

## Related issue(s)

Fixes #

### DIFF
--- a/test/e2e/reconciler/apiexportendpointslice/apiexportendpointslice_test.go
+++ b/test/e2e/reconciler/apiexportendpointslice/apiexportendpointslice_test.go
@@ -19,6 +19,7 @@ package apiexportendpointslice
 import (
 	"context"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
@@ -159,11 +160,11 @@ func TestAPIExportEndpointSliceWithPartition(t *testing.T) {
 	require.True(t, conditions.IsTrue(slice, apisv1alpha1.APIExportEndpointSliceURLsReady), "expecting the URLs ready condition")
 }
 
-func TestAPIExportEndpointSliceWithPartitionPrivate(t *testing.T) {
+func TestAPIExportEndpointSliceWithNewShard(t *testing.T) {
 	t.Parallel()
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-	server := framework.PrivateKcpServer(t)
+	server := framework.SharedKcpServer(t)
 
 	// Create Organization and Workspaces
 	orgPath, _ := framework.NewOrganizationFixture(t, server)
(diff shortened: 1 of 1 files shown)


fgiloux:
This looks good to me. I just have a couple of points:

Can you please amend the comment here?
https://github.com/kcp-dev/kcp/blob/73cc8efef729522a1dc30f1e7318559982817460/test/e2e/reconciler/apiexportendpointslice/apiexportendpointslice_test.go#L268-L270

This could be a separate test function:
https://github.com/kcp-dev/kcp/blob/73cc8efef729522a1dc30f1e7318559982817460/test/e2e/reconciler/apiexportendpointslice/apiexportendpointslice_test.go#L342-L382
It was not the case previously due to the cost of setting up the private environment

p0lyn0mial:
Updated the comment. IMO there is no need for extracting logic to a helper function. I like when tests are self-contained and self-descriptive.

fgiloux:
> Updated the comment. IMO there is no need for extracting logic to a helper function. I like when tests are self-contained and self-descriptive.

I did not mean a helper function. I meant a completely separate test function as it is not directly related to the tests part of the same function. It is not critical, just thinking it may be nicer.

test/e2e/reconciler/apiexportendpointslice/apiexportendpointslice_test.go
```diff
@@ -352,12 +356,28 @@ func TestAPIExportEndpointSliceWithPartitionPrivate(t *testing.T) {
...
+					}
+				}
+				if !found {
+					return false, fmt.Sprintf("didn't find matching shard for %v endpoint, shards: %#v", apiExportEndpoint.URL, shards.Items)
+				}
+			}
 			return true, ""
 		}
-		return false, fmt.Sprintf("expected 1 endpoint, but got: %#v", sliceWithAll.Status.APIExportEndpoints)
+		return false, fmt.Sprintf("expected %d endpoint, but got: %#v", len(shards.Items), sliceWithAll.Status.APIExportEndpoints)
```
fgiloux, line 381: I am concerned that you may be introducing a flake here: What does guarantee no shard was created by a parallel test case between your call to list shards and the one to get the APIExportEndpointSlices?

p0lyn0mial: good point, assuming schedulable shards won't be removed we could change this logic to expect `>=` `initial shards+ shards added by this tests`.

p0lyn0mial:
> I did not mean a helper function. I meant a completely separate test function as it is not directly related to the tests part of the same function. It is not critical, just thinking it may be nicer.

ah, this could be a default (happy-path) scenario I always wanted, yeah I can do that :)

PR #2825, opened 2023-02-21
🌱 docs: actually add main.py module to git

ncdc:


docs/main.py
```diff
@@ -0,0 +1,67 @@
...
+                continue
+            if child.file.name == 'index':
+                # don't include index pages
+                continue
+
+            # Because some pages might not have been loaded yet, we have to do so now, to get title/metadata.
+            child.read_source(config)
+
+            # Copy so we don't modify the original
+            child = copy.copy(child)
```
nrb, line 59: I think you want `copy.deepcopy` here. `copy.copy` will copy a reference, not the values.

https://docs.python.org/3/library/copy.html for official docs and http://pymotw.com/2/copy/ for an example showing that objects are the same. afaik these functions didn't meaningfully change between Python 2 and 3.

ncdc: TY!

PR #2828, opened 2023-02-21
:book: Add initial APIs content

ncdc:


stevekuznetsov:
Any reason not to merge?

/lgtm
/approve

ncdc:
Would merge and iterate

nrb:
Agreed on merging and iterating, this is better than what we have.

/lgtm

docs/content/en/concepts/apis-in-kcp.md
```diff
@@ -0,0 +1,376 @@
...
+The identity's **hash** is similar to a public key; it is not private and there are times when other consumers of your
+exported APIs need to know and reference it.
+
+Given 2 Workspaces, each with its own `APIExport` that exports `widgets.example.kcp.dev`, kcp uses the identity hash (in
+a mostly transparent manner) to ensure the correct instances associated with the appropriate `APIResourceSchema` are
+served to clients. See [Run Your Controller](#Run-Your-Controller) for more information.
+
+#### Permission Claims
+
+An API provider may need to access APIs that come from other APIExports, or that are built in to kcp. This is done by
```
lionelvillard, line 244: this seems to contradict the documentation in the code:

```
// PermissionClaim identifies an object by GR and identity hash.
// Its purpose is to determine the added permissions that a service provider may
// request and that a consumer may accept and allow the service provider access to.
```

ncdc: What part looks contradictory?

lionelvillard: Confusing more than contradict. I would rephase it, maybe something like this: "API provider may need to access resources in consumer workspaces. These resources might come from other APIExports, ...."

Does that make sense?

ncdc: Yes, thanks for the clarification

ncdc: Pushed an update - WDYT?

lionelvillard: much better. Thanks!

ncdc:
Re-tagging and un-holding
/unhold

PR #2829, opened 2023-02-21
✨ Upsync Endpoints of synced services

### LINKED ISSUE #2817: feature: Upsync Endpoints
Upsync Endpoints related to a Service, so that the standard KNative Service can find the Endpoints related to a deployment, and know that an Application is effectively being served.

davidfestal:
## Summary

Upsync Endpoints of synced services, so that controllers that rely of their existence to check application readiness (like KNative does) can work.

## Related issue(s)

Fixes #2817


pkg/apis/workload/v1alpha1/types.go
```diff
@@ -161,4 +161,15 @@ const (
...
+	// UpsyncDerivedResourcesAnnotationKey is an annotation set on a SyncTarget resource that contains a
+	// command-separated list of stringified GroupResource (<resource>.<group>) for the derived resources
+	// that are expected to be upsynced.
+	// To allow upsyncing the endpoints resource related to a synced service, the service should be annotated with:
+	//
+	//   workload.kcp.io/upsync-derived-resources: endpoints
+	//
+	// For now, only endpoints can be upsynced on demand by the syncer with this mechanism,
+	// but the list of such resources will increase in the future.
+	UpsyncDerivedResourcesAnnotationKey = "workload.kcp.io/upsync-derived-resources"
```
sttts, line 174: is this experimental and going to move into the SyncTarget spec?

config/rootcompute/kube-1.24/apiexport-kubernetes.k8s.io.yaml
```diff
@@ -11,6 +11,7 @@ spec:
     - v124.services.core
     - v124.deployments.apps
     - v124.pods.core
+    - v124.endpoints.core
```
sttts, line 14: what about EndpointSlices? Does one need them? Does one need both or do endpoints always exist?

davidfestal: For now we didn't have any use-case requiring upsyncing the `EndpointSlice`s.
This PR tackles minimal requirements spotted by the KNative support scenario.
Additional resources will be added in the future as required.

It would still be possible to import `EndpointSlice`s from the physical cluster if needed, and sync them as any other resource if necessary.

pkg/apis/workload/v1alpha1/types.go
davidfestal: No it's not experimental.
And it's not expected to be added to the SyncTarget (obviously it is a copy/paste erorr in the comment).
The real comment should be:
```
	// UpsyncDerivedResourcesAnnotationKey is an annotation set on synced resources that contains a
	// command-separated list of stringified GroupResource (<resource>.<group>) for the derived resources
	// that are expected to be upsynced.
	// To allow upsyncing an Endpoints resource related to a synced service, the Service instance should be annotated with:
	//
	//   workload.kcp.io/upsync-derived-resources: endpoints
	//
	// For now, only endpoints can be upsynced on demand by the syncer with this mechanism,
	// but the list of such resources will increase in the future
```

davidfestal: I fixed the comment in commit 5ba3c356469a5a968009cdd20ac69fc0f7620251

davidfestal: > and is this annotation user-facing in the first place? Who should be able to change it?
Yes, it is user-facing. For now the use should be able to change it. Mid-term it could be done by a dedicated coordination controller.

>(now it is protected through admission).
Are *all* the `workload.kcp.io/xxx` annotation protected ?

sttts: > Are all the workload.kcp.io/xxx annotation protected ?

yes. You have to opt-out in the admission plugin for that.

> Yes, it is user-facing. For now the use should be able to change it. Mid-term it could be done by a dedicated coordination controller.

So it is experimental?

davidfestal: The presence of this annotation isn't sufficient to trigger the Upsync, but it will be used as an additional condition by the logic that exists downstream and that marks resources for Upsync (by setting the `RequestState` label to `Upsync`).
In the case of Endpoints, the controller in the Syncer implements the logic that identifies which Endpoints should be Upsynced, and part of this logic is to look at this annotation on the related Service.
 
In any case, there is also a protection on the Upsyncer virtual workspace side + Syncer about the resource types that can be upsynced: for now the list is fixed: only PVs, Pods and Endpoints can be upsynced.

davidfestal: Finally made the annotation experimental and added more comments on the annotation definition.

davidfestal:
@sttts Is it OK for you know with the annotation made `experimental` and the added comments on it ?

pkg/apis/workload/v1alpha1/types.go
```diff
@@ -161,4 +161,31 @@ const (
...
+	// In addition, Upsyncing is limited to a limited, well-defined list of resource types on the KCP side,
+	// so that simply adding this annotation on a synced resource will not be a security risk.
+	//
+	// This annotation is user-facing and would typically be set by the client creating the synced resource
+	// in the KCP workspace, be it the end-user or a third-party controller.
+	//
+	// It is experimental since the provided user-experience is unsatisfactory,
+	// and further work should be done to define such (up)syncing strategies at a more appropriate level
+	// (SyncTarget, KCP namespace, KCP workspace ?).
+	ExperimentalUpsyncDerivedResourcesAnnotationKey = "experimental.workload.kcp.io/upsync-derived-resources"
```
sttts, line 190: better :+1:

davidfestal: would you approve ? ;-)

Issue #2737, opened 2023-02-01
bug:  `kcp bind compute` does not work as expected for locations on a different workspace


pdettori:
### Describe the bug

I created a new workspace `inventory` under the home workspace (e.g. `kvdk2spgmbix`), and generate a syncer yaml from there and deploy the syncer on a pCluster.  At this point I check the path for the workspace where my location is:
```
$ kubectl ws .
Current workspace is "kvdk2spgmbix:inventory".
```
Then, I create a new workspace under the home workspace, for example `ws1` and enter there:
```
kubectl ws 
kubectl ws create ws1 --enter
```
Finally, I try to bind to the workspace containing the location:
```
kubectl kcp bind compute kvdk2spgmbix:inventory
```
But I get this error:

```
Binding APIExport "root:compute:kubernetes".
placement placement-1syj10fy created.
error: placement "placement-1syj10fy" not ready: placement is not scheduled: No selected location is scheduled
```
I found that in order to make this work, I have to use the cluster name, not the name in the form `kvdk2spgmbix:inventory`.
So I had to look up the cluster name in the yaml of the inventory workspace - e.g.
```
$ kubectl ws 
$ kubectl get workspace inventory -o yaml | grep " cluster:"
cluster: 36jjanpr01gl8d45
```
then I can bind with this cluster name and placement becomes healthy:

```
$ kubectl kcp bind compute 36jjanpr01gl8d45
Binding APIExport "root:compute:kubernetes".
placement placement-i152ylce created.
Placement "placement-i152ylce" is ready.
```

### Steps To Reproduce

1.  create fresh kind cluster
```
kind delete cluster --name control && kind create cluster --name control
```
2. fetch & merge latest main, run build
```
make build
export PATH=$(pwd)/bin:$PATH  
```
3. Build locally the syncer image
```
brew install ko
ko build --local --push=false -B ./cmd/syncer -t $(git rev-parse --short HEAD)
```
(add --platform linux/amd64,linux/arm64 if building on a MBP M1)
4. Load image on kind
```
kind load docker-image ko.local/syncer:$(git rev-parse --short HEAD) --name control
```
5. Cleanup previous kcp state (if any) &  start kcp
```
rm -rf .kcp
kcp start
```
6. Open another terminal, and enter:
```
export KUBECONFIG=$(pwd)/.kcp/admin.kubeconfig
kubectl ws 
kubectl ws create inventory --enter
[...]

PR #2830, opened 2023-02-22
🐛 Fix workspace output when enter into home workspace

qiujian16:
## Summary

the workspace output when run `kubectl ws "~"` is not user friendly. 

I was 

```
$ k ws "~"
Current workspace is "kvdk2spgmbix".
```

with the fix, it is changed to

$ kubectl ws "~"
Current workspace is "user:kcp-admin". 

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/2737

### DIFF
--- a/pkg/cliplugins/workspace/plugin/kubeconfig.go
+++ b/pkg/cliplugins/workspace/plugin/kubeconfig.go
@@ -215,8 +215,21 @@ func (o *UseWorkspaceOptions) Run(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
-		newServerHost = homeWorkspace.Spec.URL
-
+		u, homeCluster, err := pluginhelpers.ParseClusterURL(homeWorkspace.Spec.URL)
+		if err != nil {
+			return fmt.Errorf("home URL %q does not point to a workspace: %w", homeWorkspace.Spec.URL, err)
+		}
+		homeLogicalCluster, err := o.kcpClusterClient.Cluster(homeCluster).CoreV1alpha1().LogicalClusters().Get(ctx, corev1alpha1.LogicalClusterName, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get logical cluster of home workspace: %w", err)
+		}
+		homePath, ok := homeLogicalCluster.Annotations[core.LogicalClusterPathAnnotationKey]
+		if !ok {
+			return fmt.Errorf("cannot find the path info from the home workspace")
+		}
+		cluster := logicalcluster.NewPath(homePath)
+		u.Path = path.Join(u.Path, cluster.RequestPath())
+		newServerHost = u.String()
 	case ".":
 		cfg, err := o.ClientConfig.ClientConfig()
 		if err != nil {
--- a/pkg/cliplugins/workspace/plugin/kubeconfig_test.go
+++ b/pkg/cliplugins/workspace/plugin/kubeconfig_test.go
@@ -1232,6 +1232,25 @@ func TestUse(t *testing.T) {
 				return false, nil, nil
 			})
 
+			client.PrependReactor("get", "logicalclusters", func(action kcptesting.Action) (handled bool, ret runtime.Object, err error) {
+				getAction := action.(kcptesting.GetAction)
+				if getAction.GetCluster() != homeWorkspaceLogicalCluster {
+					return false, nil, nil
+				}
+				if getAction.GetName() == corev1alpha1.LogicalClusterName {
+					logicalCluster := &corev1alpha1.LogicalCluster{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: corev1alpha1.LogicalClusterName,
+							Annotations: map[string]string{
+								core.LogicalClusterPathAnnotationKey: homeWorkspaceLogicalCluster.String(),
+							},
+						},
+					}
+					return true, logicalCluster, nil
+				}
+				return false, nil, nil
+			})
+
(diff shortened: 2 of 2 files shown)


pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -215,8 +215,21 @@ func (o *UseWorkspaceOptions) Run(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
-		newServerHost = homeWorkspace.Spec.URL
-
+		u, homeCluster, err := pluginhelpers.ParseClusterURL(homeWorkspace.Spec.URL)
+		if err != nil {
+			return fmt.Errorf("home URL %q does not point to a workspace", homeWorkspace.Spec.URL)
```
s-urbaniak, line 220: `fmt.Errorf("home URL %q does not point to a workspace: %w", homeWorkspace.Spec.URL, err)`

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -215,8 +215,21 @@ func (o *UseWorkspaceOptions) Run(ctx context.Context) error {
...
+		}
+		homeLogicalCluster, err := o.kcpClusterClient.Cluster(homeCluster).CoreV1alpha1().LogicalClusters().Get(ctx, corev1alpha1.LogicalClusterName, metav1.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("failed to get logical cluster of home workspace: %v", err)
```
s-urbaniak, line 224: use `%w` rather than `%v`

s-urbaniak:
some unit tests failed, maybe these are related?

davidfestal:
> some unit tests failed, maybe these are related?

In any case it would be great to ensure there is a unit test covering the fixed bug.

qiujian16:
test updated. actually the original test case should cover this, but th cluster annotation set in the test case is not the same as what it is in the real server.

pkg/cliplugins/workspace/plugin/kubeconfig_test.go
```diff
@@ -753,7 +753,7 @@ func TestUse(t *testing.T) {
 				AuthInfos: map[string]*clientcmdapi.AuthInfo{"test": {Token: "test"}},
 			},
 			existingObjects: map[logicalcluster.Name][]string{
-				core.RootCluster: {"~"},
+				core.RootCluster: {"~", homeWorkspaceLogicalCluster.String()},
```
davidfestal, line 756: Not sure I understand why this addition is necessary.
Afaict, it the failing case would be better tested if we distinguish between `homeWorkspaceLogicalCluster` *name* and `homeWorkspaceLogicalCluster` hierarchical path. Afaik for now we use the same value (path everywhere). But the failing test was precisely related to the fact that those 2 values are different.

qiujian16: this is not necessary, removed.

davidfestal: > Afaict, it the failing case would be better tested if we distinguish between homeWorkspaceLogicalCluster name and homeWorkspaceLogicalCluster hierarchical path. Afaik for now we use the same value (path everywhere). But the failing test was precisely related to the fact that those 2 values are different.

what about this part of my comment ?

pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -215,8 +215,21 @@ func (o *UseWorkspaceOptions) Run(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
-		newServerHost = homeWorkspace.Spec.URL
-
+		u, homeCluster, err := pluginhelpers.ParseClusterURL(homeWorkspace.Spec.URL)
```
sttts, line 218: `homeWorkspace.Spec.URL` has the path. We just have to extract it from the URL. We have a helper here somewhere.

sttts: Would rather not go through the `LogicalCluster` for that.

pdettori:
I tested this PR and it does resolve issue #2737

pkg/cliplugins/workspace/plugin/kubeconfig.go
qiujian16: I do not think currently the value lf homeWorkspace.Spec.URL is the same as path annotation in `LogicalCluster`. `homeWorkspace.Spec.URL` is `https://xxx/clusters/kvdk2spgmbix `  while path annotation is `user:kcp-admin`. Should `homeWorkspace.Spec.URL`  be `https://xxx/clusters/user:kcp-admin`?

davidfestal: > Would rather not go through the LogicalCluster for that.

@sttts what would be an alternate solution ?

sttts: > Should homeWorkspace.Spec.URL be https://xxx/clusters/user:kcp-admin?

Yes, I think it should.

davidfestal: Ah, so that means that the problem is at a different level.

qiujian16: @sttts would you check if https://github.com/kcp-dev/kcp/pull/2846 is the appropriate way.

ncdc:
Replaced by #2846

PR #2836, opened 2023-02-22
🌱  tmc: split apart virtual workspaces

sttts:
Allows us to run a tmc-enabled virtual workspace command.

This PR converts the plumbing of VWs to the common options/config/server pattern, and with that every layer can be tweaked in a derived command. Especially, this means we can add more virtual workspaces to the base command (which does not include tmc).

Note: this PR does not yet create a tmc-less command. But it could be now with these changes.


pkg/virtual/framework/rootapiserver/server.go
```diff
@@ -0,0 +1,128 @@
...
+// It does NOT expose oauth, related oauth endpoints, or any kube APIs.
+type RootAPIServer struct {
+	GenericAPIServer *genericapiserver.GenericAPIServer
+}
+
+func NewServer(c completedConfig, delegationTarget genericapiserver.DelegationTarget) (*RootAPIServer, error) {
+	delegateAPIServer := delegationTarget
+	for _, vw := range c.Extra.VirtualWorkspaces {
+		var err error
+		delegateAPIServer, err = vw.Register(vw.Name, c.Generic, delegateAPIServer)
```
mjudeikis, line 60: Do I get this right? Root API server now can take list of virtual workspaces servers and becomes routing layer to route to delegated child API servers?

pkg/reconciler/workload/synctarget/synctarget_reconcile.go
```diff
@@ -30,7 +30,7 @@ import (
 	virtualworkspacesoptions "github.com/kcp-dev/kcp/cmd/virtual-workspaces/options"
 	corev1alpha1 "github.com/kcp-dev/kcp/pkg/apis/core/v1alpha1"
 	workloadv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/workload/v1alpha1"
-	syncerbuilder "github.com/kcp-dev/kcp/pkg/virtual/syncer/builder"
+	syncerbuilder "github.com/kcp-dev/kcp/tmc/pkg/virtual/syncer/builder"
```
mjudeikis, line 33: Does this mean that syncer is now part of the TMC stack? Moves out of the core with this change?

mjudeikis (commented):
I lack a background in APIServer machinery to do a meaningful review of the underlying code. Still had a few questions. 
How do Virtual workspaces deal with different paths?
I can see rootAPIServer now claims `const DefaultRootPathPrefix string = "/services"`  and looks like TMC is not doing any at the moment, but in theory, it could? Or this is not something desired here?

pkg/reconciler/workload/synctarget/synctarget_reconcile.go
sttts: the virtual workspace for syncer, yes

pkg/virtual/framework/rootapiserver/server.go
sttts: It was like that already before. What changes is that the plumbing allows adding to that list after the options have been completed, i.e. in the config or server layer. Before authorization was wired in a way that this list could not be changed anymore.

tmc/pkg/virtual/syncer/builder/build.go
```diff
@@ -30,9 +30,9 @@ import (
 	"github.com/kcp-dev/kcp/pkg/indexers"
 	"github.com/kcp-dev/kcp/pkg/virtual/framework/forwardingregistry"
 	"github.com/kcp-dev/kcp/pkg/virtual/framework/rootapiserver"
-	"github.com/kcp-dev/kcp/pkg/virtual/syncer/controllers/apireconciler"
-	"github.com/kcp-dev/kcp/pkg/virtual/syncer/transformations"
-	"github.com/kcp-dev/kcp/pkg/virtual/syncer/upsyncer"
+	apireconciler2 "github.com/kcp-dev/kcp/tmc/pkg/virtual/syncer/controllers/apireconciler"
+	transformations2 "github.com/kcp-dev/kcp/tmc/pkg/virtual/syncer/transformations"
+	upsyncer2 "github.com/kcp-dev/kcp/tmc/pkg/virtual/syncer/upsyncer"
```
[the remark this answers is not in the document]
sttts: Goland magic. Let me fix.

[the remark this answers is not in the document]
sttts: done

PR #2846, opened 2023-02-27
🐛 Fix the spec.URL in workspace using canonicalPath

qiujian16:
The spec.URL in workspace should use canonicalPath

## Summary
The spec.URL of workspace should use canonicalPath instead of absolute path.

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/2737


qiujian16:
/hold for testing
this is still need to be confirmed whether it is the appropriate approach.

ncdc:
Code LGTM. @sttts does this violate any tenets for workspace URLs?

ncdc (approved):

PR #2869, opened 2023-03-06
:seedling: split `pkg/client` with go workspaces

### LINKED ISSUE #2794: bug: Unable to use KCP client set
### Describe the bug

Tried getting go module of KCP ..
 getting this logs
 

go: github.com/kcp-dev/kcp/pkg/apis@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
go: k8s.io/cloud-provider@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/component-helpers@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/controller-manager@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/kube-aggregator@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/kube-controller-manager@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/kubelet@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/mount-utils@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/pod-security-admission@v0.0.0: invalid version: unknown revision v0.0.0

### Steps To Reproduce

` go get github.com/kcp-dev/kcp@latest`
 

### Expected Behaviour

resolving deps should be done properly

### Additional Context

_No response_

mjudeikis:
## Summary

Split `pkg/client` to separate the module

Follow-up PR will be needed to replace with real version once we publish first version

## Related issue(s)

Fixes #2794
Supersedes [#2676](https://github.com/kcp-dev/kcp/pull/2676)
Supersedes https://github.com/kcp-dev/kcp/pull/2192


go.mod
```diff
@@ -107,6 +107,7 @@ require (
 	github.com/jonboulle/clockwork v0.2.2 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/json-iterator/go v1.1.12 // indirect
+	github.com/kcp-dev/kcp/pkg/client v0.0.0-00010101000000-000000000000 // indirect
```
mjudeikis, line 110: went bit onto a rabbit hole here. How does one update this with the real version before we do first release of the new module? 
https://github.com/kcp-dev/kcp/pull/2676#discussion_r1087148680

I suspect we need to do this first, release and do one more PR?

ncdc:
Will review more thoroughly later, but one thing I wanted to mention quickly is about go.work. I don't think we want to check that in to the repo? I think that's usually something developers could/would customize locally. I'd actually probably add it to .gitignore...

go.mod
ncdc: This can stay as 0.0.0

cmd/apigen/go.mod
```diff
@@ -5,21 +5,22 @@ go 1.19
 require (
 	github.com/bombsimon/logrusr/v3 v3.1.0
 	github.com/go-logr/logr v1.2.3
-	github.com/google/go-cmp v0.5.5
```
ncdc, line 8: Are the changes in this file required? Could they be reverted?

go.mod
```diff
@@ -10,7 +10,7 @@ require (
 	github.com/coredns/coredns v1.9.3
 	github.com/davecgh/go-spew v1.1.1
 	github.com/egymgmbh/go-prefix-writer v0.0.0-20180609083313-7326ea162eca
-	github.com/emicklei/go-restful v2.9.5+incompatible
+	github.com/emicklei/go-restful v2.16.0+incompatible
```
ncdc, line 13: Revert?

go.mod
```diff
@@ -19,7 +19,8 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/kcp-dev/apimachinery/v2 v2.0.0-alpha.0
 	github.com/kcp-dev/client-go v0.0.0-20230126185145-aeff170a288b
-	github.com/kcp-dev/kcp/pkg/apis v0.0.0-00010101000000-000000000000
+	github.com/kcp-dev/kcp/pkg/apis v0.11.0
```
ncdc, line 22: Revert to 0s

mjudeikis: So this is a bit interesting. You might know the right way to do this.
`pkg/client` now pulls in `github.com/kcp-dev/kcp/pkg/apis v0.11.0` as it is for APIs. And as it is a child module, I don't think we can do `replaces api ../apis` in it?

So as we have `pkg/client` in root go mod as `./pkg/client` go mod automatically places this version here (it still overrides bellow), but I'm not sure how to force 0s here now.

ncdc: Ah, so it's coming in automatically. That's fine.

go.mod
mjudeikis: Ok, it was coming from other sub-modules, forced all of them down, and this one went down too (including root one too)... :man_juggling:

cmd/apigen/go.mod
```diff
@@ -8,14 +8,15 @@ require (
 	github.com/google/go-cmp v0.5.5
 	github.com/kcp-dev/kcp/pkg/apis v0.10.1-0.20230123155139-7b3126cbf96c
 	github.com/sirupsen/logrus v1.8.1
-	github.com/spf13/pflag v1.0.5
+	github.com/spf13/pflag v1.0.6-0.20210604193023-d5e0c0615ace
```
ncdc, line 11: What's causing this bump, as well as the other bumps in this file?

mjudeikis: go mod tidy to force update sum file ;/ and same for bellow. 
how worries you are for these. I can try remove them, but it playing with `go get && go mod tidy` x repeat until you match version :/

mjudeikis: I tried to reduce the scope for more core packages, but these still gets bumped by tooling

ncdc: I was thinking that if the versions matched across the go.mod files, we wouldn't see the bumps. I can try to align them if you'd like.

mjudeikis: let me try once more tomorrow :)

mjudeikis: Ah, I think this is what you said, root go mod has this version, to it was aligned as you stated. https://github.com/kcp-dev/kcp/blob/main/go.mod#L30

cmd/apigen/go.mod
```diff
@@ -8,14 +8,15 @@ require (
...
 	k8s.io/api v0.24.3
 	k8s.io/apiextensions-apiserver v0.24.3
 	k8s.io/apimachinery v0.24.3
-	sigs.k8s.io/yaml v1.2.0
+	sigs.k8s.io/yaml v1.3.0
```
mjudeikis, line 15: same for this. Root one is 1.3

mjudeikis:
k, cleaned the root go.mod to be minimal change. All the other child go.mod are just being aligned with root one now.

ncdc (approved):

PR #2874, opened 2023-03-07
🌱 Switch workload/namespace controller to use committer

ncdc:
## Summary
1. Switch committer to allow changes to both spec/meta and status in the same reconcile loop. If both were changed, only the spec/meta patch is sent. If nothing else changes, the presumption is the status patch will go through on the next reconcile iteration.
2. Switch workload/namespace controller to use committer. This involved removing the individual places where it was sending namespace patches directly and replacing them with calculating the desired state of the namespace in the reconcile logic. Then, after reconcile() is done, the committer will send a patch if needed.

## Related issue(s)

Part of #2635

### DIFF
--- a/pkg/reconciler/workload/namespace/namespace_reconcile.go
+++ b/pkg/reconciler/workload/namespace/namespace_reconcile.go
@@ -20,60 +20,37 @@ import (
 	"context"
 	"time"
 
-	"github.com/kcp-dev/logicalcluster/v3"
-
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/labels"
-	utilserrors "k8s.io/apimachinery/pkg/util/errors"
-
-	schedulingv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/scheduling/v1alpha1"
 )
 
-type reconcileStatus int
-
-const (
-	reconcileStatusStop reconcileStatus = iota
-	reconcileStatusContinue
-)
-
-type reconciler interface {
-	reconcile(ctx context.Context, ns *corev1.Namespace) (reconcileStatus, *corev1.Namespace, error)
+type reconcileResult struct {
+	stop         bool
+	requeueAfter time.Duration
 }
 
-func (c *controller) reconcile(ctx context.Context, ns *corev1.Namespace) error {
-	reconcilers := []reconciler{
-		&bindNamespaceReconciler{
-			listPlacement:  c.listPlacement,
-			patchNamespace: c.patchNamespace,
-		},
-		&placementSchedulingReconciler{
-			listPlacement:  c.listPlacement,
-			enqueueAfter:   c.enqueueAfter,
-			patchNamespace: c.patchNamespace,
-			now:            time.Now,
-		},
-		&statusConditionReconciler{
-			patchNamespace: c.patchNamespace,
-		},
-	}
+type reconcileFunc func(ctx context.Context, key string, ns *corev1.Namespace) (reconcileResult, error)
 
-	var errs []error
+func (c *controller) reconcile(ctx context.Context, key string, ns *corev1.Namespace) error {
+	reconcilers := []reconcileFunc{
+		c.reconcilePlacementBind,
+		c.reconcileScheduling,
+		c.reconcileStatus,
+	}
 
 	for _, r := range reconcilers {
-		var err error
-		var status reconcileStatus
-		status, ns, err = r.reconcile(ctx, ns)
+		result, err := r(ctx, key, ns)
 		if err != nil {
-			errs = append(errs, err)
+			return err
 		}
-		if status == reconcileStatusStop {
+
+		if result.stop {
 			break
+		}
+
+		if result.requeueAfter > 0 {
+			c.queue.AddAfter(key, result.requeueAfter)
 		}
 	}
 
-	return utilserrors.NewAggregate(errs)
-}
-
-func (c *controller) listPlacement(clusterName logicalcluster.Name) ([]*schedulingv1alpha1.Placement, error) {
-	return c.placementLister.Cluster(clusterName).List(labels.Everything())
+	return nil
 }
--- a/pkg/reconciler/workload/namespace/namespace_reconcile_placementbind.go
+++ b/pkg/reconciler/workload/namespace/namespace_reconcile_placementbind.go
@@ -18,73 +18,54 @@ package namespace
 
 import (
 	"context"
-	"encoding/json"
 
 	"github.com/kcp-dev/logicalcluster/v3"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/klog/v2"
 
 	schedulingv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/scheduling/v1alpha1"
 	"github.com/kcp-dev/kcp/sdk/apis/third_party/conditions/util/conditions"
 )
 
-// bindNamespaceReconciler updates the existing annotation and creates an empty one if
-// at least one placement matches and there is no annotation. It delete the annotation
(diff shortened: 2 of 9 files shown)


ncdc:
@stevekuznetsov curious for your thoughts on this

pkg/apis/scheduling/v1alpha1/types_location.go
```diff
@@ -25,7 +25,8 @@ const (
 	// representation of the location labels in order to use them in a table column in the CLI.
 	LocationLabelsStringAnnotationKey = "scheduling.kcp.io/labels"
 
-	// PlacementAnnotationKey is the label key for the label holding a PlacementAnnotation struct.
+	// PlacementAnnotationKey is the label key for a namespace that indicates the namespace's labels match the selector
+	// in at least one ready Placement.
```
sttts, line 29: this drops the specification of the data format. Has this become a string and not a struct?

pkg/reconciler/workload/namespace/namespace_reconcile.go
```diff
@@ -18,62 +18,23 @@ package namespace
...
-			break
-		}
+	if err := c.reconcilePlacementBind(ctx, ns); err != nil {
+		errs = append(errs, err)
+	}
+	if err := c.reconcileScheduling(ctx, key, ns); err != nil {
+		errs = append(errs, err)
+	}
+	if err := c.reconcileStatus(ctx, ns); err != nil {
+		errs = append(errs, err)
```
sttts, line 36: how does this block now cope with changes to spec and status?

ncdc: See the comment in committer - if both change, the status changes are not sent

ncdc:
@sttts I've reverted my committer changes and restored the reconcile status continue/stop logic, but have still converted this to use committer instead of issuing patches directly. PTAL!

pkg/reconciler/workload/namespace/namespace_reconcile_placementbind.go
```diff
@@ -18,73 +18,53 @@ package namespace
...
-type bindNamespaceReconciler struct {
-	listPlacement func(clusterName logicalcluster.Name) ([]*schedulingv1alpha1.Placement, error)
-
-	patchNamespace func(ctx context.Context, clusterName logicalcluster.Path, name string, pt types.PatchType, data []byte, opts metav1.PatchOptions, subresources ...string) (*corev1.Namespace, error)
-}
-
-func (r *bindNamespaceReconciler) reconcile(ctx context.Context, ns *corev1.Namespace) (reconcileStatus, *corev1.Namespace, error) {
-	logger := klog.FromContext(ctx)
+func (c *controller) reconcilePlacementBind(
+	_ context.Context,
```
stevekuznetsov, line 37: Are we fulfilling some interface with these?

pkg/reconciler/workload/namespace/namespace_reconcile_placementbind.go
```diff
@@ -18,73 +18,53 @@ package namespace
...
-	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/klog/v2"
 
 	schedulingv1alpha1 "github.com/kcp-dev/kcp/pkg/apis/scheduling/v1alpha1"
 	"github.com/kcp-dev/kcp/pkg/apis/third_party/conditions/util/conditions"
 )
 
-// bindNamespaceReconciler updates the existing annotation and creates an empty one if
-// at least one placement matches and there is no annotation. It delete the annotation
+// reconcilePlacementBind updates the existing annotation and creates an empty one if
```
stevekuznetsov, line 32: super-nit: which annotation?

pkg/reconciler/workload/namespace/namespace_reconcile_scheduling.go
```diff
@@ -37,30 +33,24 @@ import (
 
 const removingGracePeriod = 5 * time.Second
 
-// placementSchedulingReconciler reconciles the state.workload.kcp.io/<syncTarget> labels according the
+// reconcileScheduling reconciles the state.workload.kcp.io/<syncTarget> labels according the
```
stevekuznetsov, line 36: super-nit: this comment (and a bunch of the others) are mechanistic, but would be more useful if they were semantic - what does it _mean_ to move an annotation from a placement to a label on a synctarget? What's the point of what we're doing here?

pkg/reconciler/workload/namespace/namespace_reconcile_scheduling_test.go
```diff
@@ -322,19 +320,17 @@ func TestMultiplePlacements(t *testing.T) {
 				return testCase.placements, nil
 			}
 
-			var patched bool
-			reconciler := &placementSchedulingReconciler{
-				listPlacement:  listPlacement,
-				patchNamespace: patchNamespaceFunc(&patched, ns),
-				enqueueAfter:   func(*corev1.Namespace, time.Duration) {},
+			c := &controller{
+				queue:          workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter()),
```
stevekuznetsov, line 324: nit: since we're returning some constants like `reconcileStatusStop` etc - can we also return some sentinel to instruct the controller how to requeue with a timeout? Seems like we let the upper-level logic handle the queue in most but not all cases, and if we made such a change all queue management logic would be in one spot?

ncdc: Yeah I was considering that. Will change to a struct.

pkg/reconciler/workload/namespace/namespace_reconcile_scheduling.go
ncdc: "I'm just refactoring existing code" 😛 but let me see if I can clarify

pkg/reconciler/workload/namespace/namespace_reconcile_placementbind.go
ncdc: Yes, the one in namespace_reconcile.go

pkg/reconciler/workload/namespace/namespace_reconcile_scheduling.go
stevekuznetsov: I know you are - hence, super-nit :)

ncdc: Filed as #2896

pkg/reconciler/workload/namespace/namespace_reconcile_placementbind.go
ncdc: Updated

pkg/reconciler/workload/namespace/namespace_reconcile_scheduling_test.go
ncdc: Updated

pkg/apis/scheduling/v1alpha1/types_location.go
sttts: @davidfestal @qiujian16 @jmprusi can one of you clarify?

pkg/reconciler/workload/namespace/namespace_reconcile_placementbind.go
```diff
@@ -18,73 +18,57 @@ package namespace
...
-}
-
-func (r *bindNamespaceReconciler) reconcile(ctx context.Context, ns *corev1.Namespace) (reconcileStatus, *corev1.Namespace, error) {
-	logger := klog.FromContext(ctx)
+func (c *controller) reconcilePlacementBind(
+	_ context.Context,
+	_ string,
+	ns *corev1.Namespace,
+) (reconcileResult, error) {
+	result := reconcileResult{}
```
sttts, line 42: any deeper reason for this state? Just return the object where necessary. Easier to read and less state is better.

ncdc: I mainly did it this way so I didn't have to do `return reconcileResult{}` all over the place, but I can switch to that if you think it's better?

pkg/apis/scheduling/v1alpha1/types_location.go
ncdc: I looked everywhere and this is the only place "PlacementAnnotation" is mentioned. Also the code only ever sets the empty string and checks for presence.

PR #2876, opened 2023-03-07
:sparkles: WIP: add network policy blocking cross-tenant pod traffic

### LINKED ISSUE #1531: PCluster pods can port scan service and pod network and look for open servers
Without NetworkPolicies in place, a tenant workload can connect to any IP in the location. This means it can port scan and maliciously interact with other tenants' workloads and services.

> Question was whether we want to sync NetworkPolicies. My take:
> 1. we should use network policies to stop any ingress+egress traffic with namespaces that are not part of the workspace, but allow all ingress+egress within synced namespaces of the same workspace. As far as I see, namespace label selectors should be able to do that.
> 2. if the user wants to specify extra policies between namespaces of one workspace, I don't see why this shouldn't work and shouldn't be offered. I.e. during syncing we have to transform the policy a little to include the policy in (1). I hope that works. Somebody has to work that out in detail.

lionelvillard:
## Summary

Create a network policy per namespace to prevent traffic between pods belonging to different tenant.

See https://github.com/kcp-dev/kcp/blob/3a5f114b6de2a033dcafe8d7d282d456ce5a70ea/pkg/syncer/spec/networkpolicy_tenant.yaml for more details. 

## Related issue(s)

Fixes kcp-dev/contrib-tmc#113

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -19,6 +19,7 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/kcp-dev/apimachinery/v2 v2.0.0-alpha.0
 	github.com/kcp-dev/client-go v0.0.0-20230126185145-aeff170a288b
+	github.com/kcp-dev/kcp/pkg/apis v0.11.0
 	github.com/kcp-dev/kcp/sdk v0.0.0-00010101000000-000000000000
 	github.com/kcp-dev/logicalcluster/v3 v3.0.2
 	github.com/martinlindhe/base36 v1.1.1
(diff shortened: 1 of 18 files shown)


davidfestal:
@lionelvillard Should I wait for e2e tests to be fixed before reviewing ?

lionelvillard:
@davidfestal If you want you can start looking at the [network policy template](https://github.com/kcp-dev/kcp/blob/3a5f114b6de2a033dcafe8d7d282d456ce5a70ea/pkg/syncer/spec/networkpolicy_tenant.yaml) while I'm fixing the e2e tests.

Issue #2625, opened 2023-01-16
bug: workload_apiexport_controller should not hard code root:compute workspace 


qiujian16:
### Describe the bug

based on discussion on https://kubernetes.slack.com/archives/C021U8WSAFK/p1670319753630859?thread_ts=1669982903.359109&cid=C021U8WSAFK workload_apiexport_controller should not assume the existence of root:compute workspace. 

To resolve this

1. the `kcp compute bind` should create the local apiexport and empty resource schema in location workspace when "--resources" is set.
2. workload_apiexport_controller should fill the resource schema only but not create any resource schemas.

With this we do not need to check the overlap of resource schemas in global apiexport in root:compute workspace. 

### Steps To Reproduce

The code is shown in https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/workload/apiexport/workload_apiexport_reconcile.go#L50-L57

### Expected Behaviour

no root:compute definition in workload_apiexport_reconcile.go

PR #2877, opened 2023-03-08
🐛 Remove root:compute deps in controller

qiujian16:
1. create imported-api APIExport during sync cmd if resources flag is set
2. controller updated imported-api APIExport with updated APIResourceSchema

## Related issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/2625

### DIFF
--- a/pkg/cliplugins/workload/plugin/sync.go
+++ b/pkg/cliplugins/workload/plugin/sync.go
@@ -43,6 +43,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/validation"
@@ -56,7 +57,9 @@ import (
 	"github.com/kcp-dev/kcp/pkg/cliplugins/base"
 	"github.com/kcp-dev/kcp/pkg/cliplugins/helpers"
 	kcpfeatures "github.com/kcp-dev/kcp/pkg/features"
+	"github.com/kcp-dev/kcp/pkg/reconciler/workload/apiexport"
 	apiresourcev1alpha1 "github.com/kcp-dev/kcp/sdk/apis/apiresource/v1alpha1"
+	apisv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/apis/v1alpha1"
 	tenancyv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/tenancy/v1alpha1"
 	workloadv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/workload/v1alpha1"
 	kcpclient "github.com/kcp-dev/kcp/sdk/client/clientset/versioned"
@@ -196,6 +199,13 @@ func (o *SyncOptions) Validate() error {
 		}
 	}
 
+	for _, apiExport := range o.APIExports {
+		_, name := logicalcluster.NewPath(apiExport).Split()
+		if name == workloadv1alpha1.ImportedAPISExportName {
+			errs = append(errs, fmt.Errorf("%s is a reserved APIExport name and should not be set", workloadv1alpha1.ImportedAPISExportName))
+		}
+	}
+
 	return utilerrors.NewAggregate(errs)
 }
 
@@ -320,11 +330,45 @@ func (o *SyncOptions) applySyncTarget(ctx context.Context, kcpClient kcpclient.I
 		})
 	}
 
-	// if ResourcesToSync is not empty, add export in synctarget workspace.
-	if len(o.ResourcesToSync) > 0 && !sets.NewString(o.APIExports...).Has("kubernetes") {
-		supportedAPIExports = append(supportedAPIExports, tenancyv1alpha1.APIExportReference{
-			Export: "kubernetes",
-		})
+	// create local apiexport if resources flag is set
+	if len(o.ResourcesToSync) > 0 {
+		apiExport, err := kcpClient.ApisV1alpha1().APIExports().Get(ctx, workloadv1alpha1.ImportedAPISExportName, metav1.GetOptions{})
+		switch {
+		case apierrors.IsNotFound(err):
+			fmt.Fprintf(o.ErrOut, "Creating APIExport %q\n", workloadv1alpha1.ImportedAPISExportName)
+			apiExport = &apisv1alpha1.APIExport{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: workloadv1alpha1.ImportedAPISExportName,
+					Annotations: map[string]string{
+						workloadv1alpha1.ComputeAPIExportAnnotationKey: "true",
+					},
+				},
+				Spec: apisv1alpha1.APIExportSpec{
+					LatestResourceSchemas: []string{},
+				},
+			}
+			apiExport, _ = mergeLatestResourceSchema(apiExport, o.ResourcesToSync)
+			_, err = kcpClient.ApisV1alpha1().APIExports().Create(ctx, apiExport, metav1.CreateOptions{})
+			if err != nil && !apierrors.IsAlreadyExists(err) {
+				return nil, err
+			}
+		case err != nil:
+			return nil, err
+		default:
+			if apiExport, modified := mergeLatestResourceSchema(apiExport, o.ResourcesToSync); modified {
+				_, err = kcpClient.ApisV1alpha1().APIExports().Update(ctx, apiExport, metav1.UpdateOptions{})
+				if err != nil {
+					return nil, err
+				}
+			}
+		}
+
+		// if ResourcesToSync is not empty, add export in synctarget workspace.
+		if !sets.NewString(o.APIExports...).Has(workloadv1alpha1.ImportedAPISExportName) {
+			supportedAPIExports = append(supportedAPIExports, tenancyv1alpha1.APIExportReference{
+				Export: workloadv1alpha1.ImportedAPISExportName,
+			})
+		}
 	}
 
 	syncTarget, err := kcpClient.WorkloadV1alpha1().SyncTargets().Get(ctx, syncTargetName, metav1.GetOptions{})
@@ -418,13 +462,13 @@ func (o *SyncOptions) getResourcesForPermission(ctx context.Context, config *res
 			return false, nil //nolint:nilerr
 		}
 
-		// skip if there is only the local kubernetes APIExport in the synctarget workspace, since we may not get syncedResources yet.
-		clusterName := logicalcluster.From(syncTarget)
+		if len(syncTarget.Spec.SupportedAPIExports) == 0 {
+			return true, nil
+		}
 
+		// skip if there is only the local imported-apis APIExport in the synctarget workspace, since we may not get syncedResources yet.
 		if len(syncTarget.Spec.SupportedAPIExports) == 1 &&
-			syncTarget.Spec.SupportedAPIExports[0].Export == "kubernetes" &&
-			(len(syncTarget.Spec.SupportedAPIExports[0].Path) == 0 ||
-				syncTarget.Spec.SupportedAPIExports[0].Path == clusterName.String()) {
+			syncTarget.Spec.SupportedAPIExports[0].Export == workloadv1alpha1.ImportedAPISExportName {
 			return true, nil
 		}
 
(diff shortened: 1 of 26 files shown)


pkg/reconciler/workload/apiexport/workload_apiexport_reconcile.go
```diff
@@ -145,15 +125,17 @@ func (r *schemaReconciler) reconcile(ctx context.Context, export *apisv1alpha1.A
...
-			schemasByResourceGroup[resourceGroup] = existingSchema
 		}
 	}
 
 	// create missing or outdated schemas
 	outdatedOrMissing := expectedResourceGroups.Difference(upToDate)
 	for _, resourceGroup := range outdatedOrMissing.List() {
 		logger.WithValues("schema", resourceGroup).V(2).Info("missing or outdated schema on APIExport, adding")
-		resource := resourcesByResourceGroup[resourceGroup]
+		resource, ok := resourcesByResourceGroup[resourceGroup]
```
davidfestal, line 135: I'm not sure about the impact of this line (continuing if the missing resource is not found in the NegotiatedAPIresources).
Will the APIResourceSchema be removed from the APIExport finally because it was not found in the NegotiatedAPIResources ?
What if this resource later exists in the downstream cluster ? the schema would not be added to the APIExport, and we would have lost one of the resources initially added in the APIExport by the `workload sync` CLI command.

I might be missing something but the logic is not clear to me.

qiujian16: it won't, the APIResourceSchema reference in the  APIExport  is only replaced by the updated one. I agree the logic here is not quite clear. Need to do a refactor

pkg/reconciler/workload/apiexport/workload_apiexport_controller.go
```diff
@@ -48,7 +48,7 @@ const (
 	ControllerName = "kcp-workload-apiexport"
 
 	// TemporaryComputeServiceExportName is a temporary singleton name of compute service exports.
-	TemporaryComputeServiceExportName = "kubernetes"
+	TemporaryComputeServiceExportName = "imported-apis"
```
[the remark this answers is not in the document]
qiujian16: moved this to apis definition with updated name

pkg/apis/workload/v1alpha1/synctarget_types.go
```diff
@@ -175,6 +174,9 @@ type SyncTargetList struct {
 	Items []SyncTarget `json:"items"`
 }
 
+// ImportedAPISExportName is singleton name of compute service exports in location workspace.
```
stevekuznetsov, line 177: Is the fact that this is a singleton a new constraint on the topology of a `Location` workspace? If so, can we add this to the documentation?

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -672,6 +711,34 @@ func (o *SyncOptions) enableSyncerForWorkspace(ctx context.Context, config *rest
 	return string(saTokenBytes), syncerID, syncTarget, nil
 }
 
+func mergeLatestResourceSchema(apiExport *apisv1alpha1.APIExport, resourceToSync []string) (*apisv1alpha1.APIExport, bool) {
+	desiredResourceGroup := sets.NewString()
+	var modified bool
+	for _, schema := range apiExport.Spec.LatestResourceSchemas {
+		comps := strings.SplitN(schema, ".", 3)
```
stevekuznetsov, line 718: Why can we assume that these schema names have this format? Can we use some other (strongly-typed) data instead of strings here?

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -672,6 +711,34 @@ func (o *SyncOptions) enableSyncerForWorkspace(ctx context.Context, config *rest
...
+		if len(comps) != 3 {
+			continue
+		}
+		desiredResourceGroup.Insert(fmt.Sprintf("%s.%s", comps[1], comps[2]))
+	}
+	for _, resource := range resourceToSync {
+		nameParts := strings.SplitN(resource, ".", 2)
+		resourceName := nameParts[0]
+		apiGroup := "core"
```
stevekuznetsov, line 727: Is this a constant we can import?

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -672,6 +711,34 @@ func (o *SyncOptions) enableSyncerForWorkspace(ctx context.Context, config *rest
...
+		if len(nameParts) > 1 {
+			apiGroup = nameParts[1]
+		}
+		if !desiredResourceGroup.Has(fmt.Sprintf("%s.%s", resourceName, apiGroup)) {
```
stevekuznetsov, line 731: We should use `GroupResource` types and their `.String()` methods.

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -672,6 +711,34 @@ func (o *SyncOptions) enableSyncerForWorkspace(ctx context.Context, config *rest
...
+			// the rev-0 here is a placeholder and will be replaced by rv of negotiated APIResourceSchema finally.
+			schemaName := fmt.Sprintf("rev-0.%s.%s", resourceName, apiGroup)
```
stevekuznetsov, line 733: Similarly here - please use `GroupVersionResource`.

pkg/cliplugins/workload/plugin/sync.go
stevekuznetsov: Will this always handle multi-part group names correctly?

pkg/reconciler/workload/apiexport/workload_apiexport_reconcile.go
```diff
@@ -106,25 +87,26 @@ func (r *schemaReconciler) reconcile(ctx context.Context, export *apisv1alpha1.A
...
+	// we check all schemas reference in the apiexport. if it is missing, we should create the schema
+	// if it is outdated, we should create the schema and delete the outdated schema.
+	upToDateResourceGroup := sets.NewString()
+	expectedResourceGroups := sets.NewString()
+
+	// schemaNamesByResourceGroup records the upToData APIResourceSchemaName for the resourceGroup, and the
+	// APIExport will be updated accordingly
+	schemaNamesByResourceGroup := map[string]string{}
 	for _, schemaName := range export.Spec.LatestResourceSchemas {
 		_, resource, group, ok := split3(schemaName, ".")
```
stevekuznetsov, line 103: Why do we have an entirely different splitting mechanism here? Can we collapse these? And, can we use some more strongly-formatted data? Does validation on APIResourceSchema require that the name be `revision.resource.group`?

qiujian16: yes, it requires the schema name to be revision.resource.group by admission.

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -416,11 +451,15 @@ func (o *SyncOptions) getResourcesForPermission(ctx context.Context, config *res
...
+		}
+
 		// skip if there is only the local kubernetes APIExport in the synctarget workspace, since we may not get syncedResources yet.
 		clusterName := logicalcluster.From(syncTarget)
 
 		if len(syncTarget.Spec.SupportedAPIExports) == 1 &&
-			syncTarget.Spec.SupportedAPIExports[0].Export == "kubernetes" &&
+			syncTarget.Spec.SupportedAPIExports[0].Export == workloadv1alpha1.ImportedAPISExportName &&
 			(len(syncTarget.Spec.SupportedAPIExports[0].Path) == 0 ||
 				syncTarget.Spec.SupportedAPIExports[0].Path == clusterName.String()) {
```
davidfestal, line 464: This test mainly doesn't work anymore (after the workspace refactoring), especially if the local export has been added with the path, and not the cluster name.
Don't we want to ensure that `imported-apis` (or any other name) would be a reserved name, and then we wouldn't even allow adding it as an `apiexports` arg, especially if it has a path ?

pkg/reconciler/workload/apiexport/workload_apiexport_reconcile.go
```diff
@@ -106,25 +87,26 @@ func (r *schemaReconciler) reconcile(ctx context.Context, export *apisv1alpha1.A
...
-		expectedResourceGroups.Insert(schemaName)
 		resourcesByResourceGroup[schemaName] = r
 	}
 
 	// reconcile schemas in export
-	upToDate := sets.NewString()
-	schemasByResourceGroup := map[string]*apisv1alpha1.APIResourceSchema{}
+	// we check all schemas reference in the apiexport. if it is missing, we should create the schema
+	// if it is outdated, we should create the schema and delete the outdated schema.
+	upToDateResourceGroup := sets.NewString()
```
davidfestal, line 96:
```suggestion
	upToDateResourceGroups := sets.NewString()
```

pkg/reconciler/workload/apiexport/workload_apiexport_reconcile.go
```diff
@@ -106,25 +87,26 @@ func (r *schemaReconciler) reconcile(ctx context.Context, export *apisv1alpha1.A
...
+	expectedResourceGroups := sets.NewString()
+
+	// schemaNamesByResourceGroup records the upToData APIResourceSchemaName for the resourceGroup, and the
```
davidfestal, line 99: > records the upToData

*records the updated*, or *records the up-to-date* ?

pkg/reconciler/workload/apiexport/workload_apiexport_reconcile.go
```diff
@@ -198,38 +185,20 @@ func (r *schemaReconciler) reconcile(ctx context.Context, export *apisv1alpha1.A
...
-	if err != nil {
-		return reconcileStatusStop, err
-	}
-	for _, schema := range allSchemas {
-		if !referencedSchemaNames[schema.Name] && metav1.IsControlledBy(schema, export) {
-			logging.WithObject(logger, schema).V(2).Info("deleting schema of APIExport")
-			if err := r.deleteAPIResourceSchema(ctx, clusterName.Path(), schema.Name); err != nil && !apierrors.IsNotFound(err) {
-				return reconcileStatusStop, err
+	for schemaName := range outDatedSchemaNames {
+		if _, err := r.getAPIResourceSchema(ctx, clusterName, schemaName); err != nil {
```
davidfestal, line 189: Why do we need to do the get first ? We don't use the result.

pkg/reconciler/workload/defaultlocation/defaultlocation_controller.go
```diff
@@ -324,20 +300,5 @@ func (c *controller) process(ctx context.Context, key string) error {
 		return err
 	}
 
-	// patch the apiexport, so we do not create the location/apibinding again even if it is deleted.
```
davidfestal, line 327: Do we need to keep a similar logic to avoid creating the location/apibinding again even if it is deleted ?

pkg/cliplugins/workload/plugin/sync.go
qiujian16: the check here is to get the synced resources list from the sycnTarget, so we can correctly set the rbac for syncer on spoke. However, if there are only imported-apis APIExport, we can not get this list because the list is only updated when syncer is started.

davidfestal:
@qiujian16 there's still a test failing.

pkg/cliplugins/workload/plugin/sync.go
qiujian16: change to use a func that return revision and groupResource.

pkg/reconciler/workload/apiexport/workload_apiexport_reconcile.go
qiujian16: we do not need this. It was only to avoid an unnecessary delete. But we can always delete it.

pkg/reconciler/workload/apiexport/apiresourceschema.go
```diff
@@ -72,3 +75,12 @@ func toAPIResourceSchema(r *apiresourcev1alpha1.NegotiatedAPIResource, name stri
 
 	return schema
 }
+
+// ParseAPIResourceSchemaName parses name of APIResourceSchema to a gr and schema if it is valid.
+func ParseAPIResourceSchemaName(name string) (schema.GroupResource, string, bool) {
```
stevekuznetsov, line 80: There are no users of the second `string` this returns - can we just omit it until we need it?

pkg/reconciler/workload/defaultlocation/defaultlocation_controller.go
davidfestal: Can we avoid this annotation at all ?

we could also change a bit the logic and do the following:
- Create / delete the local APBinding systematically according to the presence / absence of the local APIExport
- Create the default location if no location already exists.

pkg/cliplugins/workload/plugin/sync.go
davidfestal: Yes, I know this.

I was speaking of this part of the test:

`syncTarget.Spec.SupportedAPIExports[0].Path == clusterName.String()`

pkg/apis/workload/v1alpha1/synctarget_types.go
davidfestal: It's mainly a rename of the previous local `kubernetes` APIExport.
But I agree that we should update the documentation.

pkg/cliplugins/workload/plugin/sync.go
davidfestal: Do you mean, should we use:
`corev1.GenericControlPlaneGroupName` ?

Even in the kubernetes fork, it doesn't seem to be used in such cases. Here is an example of those:

https://github.com/kcp-dev/kubernetes/blob/653a0f16cab7cd12271e7433fc8466686d524f84/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go#L75

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -416,13 +460,13 @@ func (o *SyncOptions) getResourcesForPermission(ctx context.Context, config *res
 			return false, nil //nolint:nilerr
 		}
 
-		// skip if there is only the local kubernetes APIExport in the synctarget workspace, since we may not get syncedResources yet.
-		clusterName := logicalcluster.From(syncTarget)
+		if len(syncTarget.Spec.SupportedAPIExports) == 0 {
+			return true, nil
+		}
 
+		// skip if there is only the local kubernetes APIExport in the synctarget workspace, since we may not get syncedResources yet.
```
davidfestal, line 467:
```suggestion
		// skip if there is only the local 'imported-apis' APIExport in the synctarget workspace, since we may not get syncedResources yet.
```

pkg/cliplugins/workload/plugin/sync.go
```diff
@@ -195,6 +198,13 @@ func (o *SyncOptions) Validate() error {
 		}
 	}
 
+	for _, apiExport := range o.APIExports {
+		_, name := logicalcluster.NewPath(apiExport).Split()
+		if name == workloadv1alpha1.ImportedAPISExportName {
+			errs = append(errs, fmt.Errorf("%s is a reversed APIExport name and should not be set", workloadv1alpha1.ImportedAPISExportName))
```
davidfestal, line 204:
```suggestion
			errs = append(errs, fmt.Errorf("%s is a reserved APIExport name and should not be set", workloadv1alpha1.ImportedAPISExportName))
```

PR #2881, opened 2023-03-08
✨ Enable the SyncerTunnel featuregate by default

jmprusi:
## Summary

This PR enables the SyncerTunel feature-gate by default.

## Related issue(s)

Fixes #


davidfestal:
Do we still need the `privateKCPServer` if we're not using the feature flag anymore ?

jmprusi:
> Do we still need the `privateKCPServer` if we're not using the feature flag anymore ?

Got rid of one in the second test, but on the first test it looks like I need to keep it to add the tokenFile, there's a way to get the same behaviour with Shared?

Issue #2865, opened 2023-03-03
Flake: TestPartitionSet


ncdc:
```
=== RUN   TestPartitionSet
=== PAUSE TestPartitionSet
=== CONT  TestPartitionSet
    partitionset_test.go:46: shared kcp server will target configuration "/go/src/github.com/kcp-dev/kcp/.kcp/admin.kubeconfig"
    kcp.go:1131: waiting for readiness for server at https://10.130.54.187:6443
=== CONT  TestPartitionSet
    kcp.go:1174: success contacting https://10.130.54.187:6443/livez
    kcp.go:1174: success contacting https://10.130.54.187:6443/readyz
    kcp.go:1152: server at https://10.130.54.187:6443 is ready
=== CONT  TestPartitionSet
    assertions.go:1691: Waiting for condition, but got: workspace phase is Initializing, not Ready
=== CONT  TestPartitionSet
    partitionset_test.go:50: Created root:organization workspace root:e2e-workspace-tx67x as /clusters/1bqz47jj1m6qkl2o on shard "root"
    assertions.go:1691: Waiting for condition, but got: workspace phase is Scheduling, not Ready
=== CONT  TestPartitionSet
    assertions.go:1691: Waiting for condition, but got: workspace phase is Initializing, not Ready
=== CONT  TestPartitionSet
    partitionset_test.go:51: Created root:universal workspace root:e2e-workspace-tx67x:partitionset as /clusters/240v2xs82vbfcn5c on shard "root"
    partitionset_test.go:60: Creating a partitionSet not matching any shard
    partitionset_test.go:92: Creating a shard matching the partitionSet
    partitionset_test.go:133: Creating a shard in a second region
=== CONT  TestPartitionSet
    assertions.go:1691: Waiting for condition, but got: expected 2 partitions, but got 3
    partitionset_test.go:163: 
        	Error Trace:	util.go:332
        	            				partitionset_test.go:163
        	Error:      	Condition never satisfied
        	Test:       	TestPartitionSet
        	Messages:   	expected 2 partitions
--- FAIL: TestPartitionSet (49.29s)
```

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kcp-dev_kcp/2864/pull-ci-kcp-dev-kcp-main-e2e-shared/1631663786190442496

PR #2675, opened 2023-01-25
✨ Syncer: support multiple shards

davidfestal:
## Summary

This PR refactors the Syncer in order to make it shard-aware (supporting multiple shards)

## Related issue(s)

Fixes #2832

PR #2882, opened 2023-03-08
:bug: Decouple internal and external logical-cluster-admin access

### LINKED ISSUE #2872: bug: logical cluster admin can't access both internal/external endpoints
### Describe the bug

Currently the logical cluster admin kubeconfig is used to access both the [external front-proxy endpoint](https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/tenancy/workspace/workspace_controller.go#L194..L195), and also the [shard directly via the baseURL](https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/tenancy/workspace/workspace_reconcile.go#L71..L72) - this won't work in the case where different CA certs are used for shard and front-proxy (such as in the [helm-charts](https://github.com/kcp-dev/helm-charts))

This only works in CI because we use the same serving-ca.crt for both shard and front-proxy, which is Ok for development but probably not a common choice for production deployments.

### Steps To Reproduce

1. Deploy KCP in a configuration where different CA certs are used for shard and front-proxy (such as in the [helm-charts](https://github.com/kcp-dev/helm-charts))
2. Create a workspace
3. Note that the workspace is stuck `Initializing`

The logs show that the single CA from the logical-cluster-admin kubeconfig can't access both endpoints for the two phases of scheduling (thanks to @sttts for the help figuring this out):

1. The creation of the logical cluster object. At this point the shard name hash is known (an annotation on the Workspace). We cannot go through the proxy because it does not know about the logical cluster yet.
2. the initialization phase. At this point the logical cluster exists and every access is like any other workspace access. I.e. it should go through the proxy.

### Expected Behaviour

The two phases of scheduling should complete successfully.

### Additional Context

[Discussion on slack](https://kubernetes.slack.com/archives/C021U8WSAFK/p1678103495741019)

tl;dr we agreed to add a `--external-logical-cluster-admin-kubeconfig` similar to the previous (reverted) change [adding an external admin kubeconfig](https://github.com/kcp-dev/kcp/commit/339a071a6b3fb2d4e9d3f6d72e900183a8885ca0), but limited to the permissions needed for workspace scheduling and deletion.

hardys:
## Summary
Currently the logical cluster admin kubeconfig is used to access both the [external front-proxy endpoint](https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/tenancy/workspace/workspace_controller.go#L194..L195), and also the [shard directly via the baseURL](https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/tenancy/workspace/workspace_reconcile.go#L71..L72) - this won't work in the case where different CA certs are used for shard and front-proxy (such as in the [helm-charts](https://github.com/kcp-dev/helm-charts))

This only works in CI because we use the same serving-ca.crt for both shard and front-proxy, (in a future PR I will propose a change which better reflects the likely scenario in production deployments).

Fixes #2872

### DIFF
--- a/cmd/sharded-test-server/frontproxy.go
+++ b/cmd/sharded-test-server/frontproxy.go
@@ -312,7 +312,10 @@ func writeShardKubeConfig(workDirPath string) error {
 }
 
 func writeLogicalClusterAdminKubeConfig(hostIP, workDirPath string) error {
-	baseHost := fmt.Sprintf("https://%s", net.JoinHostPort(hostIP, "6443"))
+	// The logical-cluster-admin-kubeconfig references the root shard, it's
+	// also used to access other shards directly; also see the external
+	// config below which is used for connections via the front-proxy
+	baseHost := fmt.Sprintf("https://%s", net.JoinHostPort(hostIP, "6444"))
 
 	var kubeConfig clientcmdapi.Config
 	kubeConfig.AuthInfos = map[string]*clientcmdapi.AuthInfo{
@@ -337,4 +340,33 @@ func writeLogicalClusterAdminKubeConfig(hostIP, workDirPath string) error {
 	}
 
 	return clientcmd.WriteToFile(kubeConfig, filepath.Join(workDirPath, ".kcp/logical-cluster-admin.kubeconfig"))
+}
+
+func writeExternalLogicalClusterAdminKubeConfig(hostIP, workDirPath string) error {
+	// The external config references the front-proxy endpoint
+	baseHost := fmt.Sprintf("https://%s", net.JoinHostPort(hostIP, "6443"))
+
+	var kubeConfig clientcmdapi.Config
+	kubeConfig.AuthInfos = map[string]*clientcmdapi.AuthInfo{
+		"external-logical-cluster-admin": {
+			ClientKey:         filepath.Join(workDirPath, ".kcp/external-logical-cluster-admin.key"),
+			ClientCertificate: filepath.Join(workDirPath, ".kcp/external-logical-cluster-admin.crt"),
+		},
+	}
+	kubeConfig.Clusters = map[string]*clientcmdapi.Cluster{
+		"base": {
+			Server:               baseHost,
+			CertificateAuthority: filepath.Join(workDirPath, ".kcp/serving-ca.crt"),
+		},
+	}
+	kubeConfig.Contexts = map[string]*clientcmdapi.Context{
+		"base": {Cluster: "base", AuthInfo: "external-logical-cluster-admin"},
+	}
+	kubeConfig.CurrentContext = "base"
+
+	if err := clientcmdapi.FlattenConfig(&kubeConfig); err != nil {
+		return err
+	}
+
+	return clientcmd.WriteToFile(kubeConfig, filepath.Join(workDirPath, ".kcp/external-logical-cluster-admin.kubeconfig"))
 }
--- a/pkg/admission/logicalcluster/admission.go
+++ b/pkg/admission/logicalcluster/admission.go
@@ -190,8 +190,7 @@ func (o *plugin) Validate(ctx context.Context, a admission.Attributes, _ admissi
 		if err != nil {
 			return fmt.Errorf("LogicalCluster cannot be deleted: %w", err)
 		}
-		groups := sets.NewString(a.GetUserInfo().GetGroups()...)
-		if !logicalCluster.Spec.DirectlyDeletable && !groups.Has(kuser.SystemPrivilegedGroup) && !groups.Has(bootstrap.SystemLogicalClusterAdmin) {
+		if !logicalCluster.Spec.DirectlyDeletable {
 			return admission.NewForbidden(a, fmt.Errorf("LogicalCluster cannot be deleted"))
 		}
 
--- a/pkg/admission/logicalcluster/admission.go
+++ b/pkg/admission/logicalcluster/admission.go
@@ -127,7 +127,7 @@ func (o *plugin) Validate(ctx context.Context, a admission.Attributes, _ admissi
 	}
 
 	groups := sets.NewString(a.GetUserInfo().GetGroups()...)
-	if groups.Has(kuser.SystemPrivilegedGroup) || groups.Has(bootstrap.SystemLogicalClusterAdmin) || groups.Has(bootstrap.SystemKcpWorkspaceBootstrapper) {
+	if groups.Has(kuser.SystemPrivilegedGroup) || groups.Has(bootstrap.SystemLogicalClusterAdmin) || groups.Has(bootstrap.SystemExternalLogicalClusterAdmin) || groups.Has(bootstrap.SystemKcpWorkspaceBootstrapper) {
 		return nil
 	}
 
--- a/pkg/authorization/bootstrap/policy.go
+++ b/pkg/authorization/bootstrap/policy.go
@@ -35,10 +35,13 @@ const (
 	// We need a separate group (not the privileged system group) for this because system-owned workspaces (e.g. root:users) need
 	// a workspace owner annotation set, and the owner annotation is skipped/not set for the privileged system group.
 	SystemKcpWorkspaceBootstrapper = "system:kcp:tenancy:workspace-bootstrapper"
-	// SystemLogicalClusterAdmin is a group used by the scheduler to create LogicalCluster resources.
+	// SystemLogicalClusterAdmin is a group used by the workspace scheduler to create LogicalCluster resources.
 	// This group allows it to skip the entire authorization stack except the bootstrap policy authorizer.
 	// Otherwise, the requests would be rejected because the LogicalCluster resource does not exist yet.
 	SystemLogicalClusterAdmin = "system:kcp:logical-cluster-admin"
+	// SystemExternalLogicalClusterAdmin is a group used by the workspace controllers to manage LogicalCluster
+	// resources after creation, using a subset of permissions allowed for the internal logical-cluster-admin.
+	SystemExternalLogicalClusterAdmin = "system:kcp:external-logical-cluster-admin"
 	// SystemKcpWorkspaceAccessGroup is a group that gives a user system:authenticated access to a workspace.
 	SystemKcpWorkspaceAccessGroup = "system:kcp:workspace:access"
 )
@@ -49,6 +52,7 @@ func clusterRoleBindings() []rbacv1.ClusterRoleBinding {
 		clusterRoleBindingCustomName(rbacv1helpers.NewClusterBinding("cluster-admin").Groups(SystemKcpAdminGroup).BindingOrDie(), "system:kcp:admin:cluster-admin"),
 		clusterRoleBindingCustomName(rbacv1helpers.NewClusterBinding(SystemKcpWorkspaceBootstrapper).Groups(SystemKcpWorkspaceBootstrapper, "apis.kcp.io:binding:"+SystemKcpWorkspaceBootstrapper).BindingOrDie(), SystemKcpWorkspaceBootstrapper),
 		clusterRoleBindingCustomName(rbacv1helpers.NewClusterBinding(SystemLogicalClusterAdmin).Groups(SystemLogicalClusterAdmin).BindingOrDie(), SystemLogicalClusterAdmin),
+		clusterRoleBindingCustomName(rbacv1helpers.NewClusterBinding(SystemExternalLogicalClusterAdmin).Groups(SystemExternalLogicalClusterAdmin).BindingOrDie(), SystemExternalLogicalClusterAdmin),
 	}
 }
 
@@ -74,6 +78,14 @@ func clusterRoles() []rbacv1.ClusterRole {
 			Rules: []rbacv1.PolicyRule{
 				rbacv1helpers.NewRule("*").Groups(core.GroupName).Resources("logicalclusters", "logicalclusters/status").RuleOrDie(),
 				rbacv1helpers.NewRule("delete", "update", "get").Groups(tenancy.GroupName).Resources("workspaces").RuleOrDie(),
+			},
+		},
+		{
+			ObjectMeta: metav1.ObjectMeta{Name: SystemExternalLogicalClusterAdmin},
+			Rules: []rbacv1.PolicyRule{
+				rbacv1helpers.NewRule("delete", "update", "get").Groups(core.GroupName).Resources("logicalclusters", "logicalclusters/status").RuleOrDie(),
+				rbacv1helpers.NewRule("delete", "update", "get").Groups(tenancy.GroupName).Resources("workspaces").RuleOrDie(),
+				rbacv1helpers.NewRule("access").URLs("/").RuleOrDie(),
 			},
 		},
 		{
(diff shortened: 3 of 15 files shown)


cmd/sharded-test-server/virtual.go
```diff
@@ -88,9 +88,9 @@ func newVirtualWorkspace(ctx context.Context, index int, servingCA *crypto.CA, h
 		return nil, fmt.Errorf("failed to create vw client cert: %w", err)
 	}
 
-	servingCAPath, err := filepath.Abs(filepath.Join(workDirPath, ".kcp/serving-ca.crt"))
+	servingCAPath, err := filepath.Abs(filepath.Join(workDirPath, ".kcp/shard-serving-ca.crt"))
 	if err != nil {
-		return nil, fmt.Errorf("error getting absolute path for %q: %w", filepath.Join(workDirPath, ".kcp/serving-ca.crt"), err)
+		return nil, fmt.Errorf("error getting absolute path for %q: %w", filepath.Join(workDirPath, ".kcp/shard-serving-ca.crt"), err)
```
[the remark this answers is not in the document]
hardys: I was unclear about that - indeed the e2e tests are failing because `admin.kubeconfig` can't connect to the VW servers directly when the CA certs are decoupled - do we expect users to connect directly to the VW server endpoints, or only via the proxy?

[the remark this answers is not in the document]
sttts: They connect directly. There could potentially be a proxy in between. But the paths used for different services must be non-overlapping.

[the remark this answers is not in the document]
hardys: I removed the splitting of CA certs for now, will revisit as a follow-up

hardys:
/retitle :bug: Decouple internal and external logical-cluster-admin access

cmd/sharded-test-server/frontproxy.go
```diff
@@ -312,7 +312,7 @@ func writeShardKubeConfig(workDirPath string) error {
 }
 
 func writeLogicalClusterAdminKubeConfig(hostIP, workDirPath string) error {
-	baseHost := fmt.Sprintf("https://%s", net.JoinHostPort(hostIP, "6443"))
+	baseHost := fmt.Sprintf("https://%s", net.JoinHostPort(hostIP, "6444"))
```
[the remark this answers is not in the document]
sttts: ack

pkg/reconciler/tenancy/workspace/workspace_controller.go
```diff
@@ -119,13 +118,12 @@ type workspaceResource = committer.Resource[*tenancyv1alpha1.WorkspaceSpec, *ten
 type Controller struct {
 	queue workqueue.RateLimitingInterface
 
-	shardName                 string
-	shardExternalURL          func() string
```
sttts, line 123: do you understand why this closure can be removed? I remember that this was about early plumbing but late computation of the value, maybe about shard object creation.

pkg/server/options/options.go
```diff
@@ -167,6 +167,7 @@ func (o *Options) AddFlags(fss *cliflag.NamedFlagSets) {
 	fs.StringVar(&o.Extra.ShardClientCertFile, "shard-client-cert-file", o.Extra.ShardClientCertFile, "Path to a client certificate file the shard uses to communicate with other system components.")
 	fs.StringVar(&o.Extra.ShardClientKeyFile, "shard-client-key-file", o.Extra.ShardClientKeyFile, "Path to a client certificate key file the shard uses to communicate with other system components.")
 	fs.StringVar(&o.Extra.LogicalClusterAdminKubeconfig, "logical-cluster-admin-kubeconfig", o.Extra.LogicalClusterAdminKubeconfig, "Kubeconfig holding admin(!) credentials to other shards. Defaults to the loopback client")
+	fs.StringVar(&o.Extra.ExternalLogicalClusterAdminKubeconfig, "external-logical-cluster-admin-kubeconfig", o.Extra.ExternalLogicalClusterAdminKubeconfig, "Kubeconfig holding admin(!) credentials talking to the external address (e.g. the front-proxy). Defaults to the loopback client")
```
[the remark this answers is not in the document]
hardys: Yes - it's a copy paste from the existing flag, shall I update both here?

pkg/reconciler/tenancy/workspace/workspace_controller.go
hardys: It can be removed because instead of explicitly setting shardExternalURL in the client, we rely on the server set in the `externalLogicalClusterAdminConfig` (see `workspace_reconcile.go` below) - AFAICS this was only required before so the endpoint from `logicalClusterAdminConfig` could overridden?

ncdc: Haven't reviewed this PR, but the closure was required before because at the time the controller is created, the shard external URL isn't necessarily set/correct.

pkg/server/options/options.go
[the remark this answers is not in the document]
sttts: yes

pkg/reconciler/tenancy/workspace/workspace_controller.go
hardys: Thanks for clarifying @ncdc - yes AFAICS it's so we can (default to genericspiserver.ExternalAddress)[https://github.com/kcp-dev/kcp/blob/main/pkg/server/config.go#L116] which presumably is not initialized at the point of creating the config.

However in this case we're no longer using the shardExternalURL at all (the server from the new external-logical-cluster-admin-kubeconfig is used instead), so I removed the argument which passes the function.

pkg/authorization/bootstrap/policy.go
```diff
@@ -76,6 +80,12 @@ func clusterRoles() []rbacv1.ClusterRole {
 				rbacv1helpers.NewRule("delete", "update", "get").Groups(tenancy.GroupName).Resources("workspaces").RuleOrDie(),
 			},
 		},
+		{
+			ObjectMeta: metav1.ObjectMeta{Name: SystemExternalLogicalClusterAdmin},
+			Rules: []rbacv1.PolicyRule{
+				rbacv1helpers.NewRule("delete", "update", "get").Groups(core.GroupName).Resources("logicalclusters", "logicalclusters/status").RuleOrDie(),
+			},
```
hardys, line 87: @sttts and/or @s-urbaniak I added a new group as discussed but this rule doesn't seem to be enough to get logical clusters at cluster scope:

```
I0316 17:47:58.446040  572728 httplog.go:131] "HTTP" verb="GET" URI="/clusters/1jys1xdusxcx96v2/apis/core.kcp.io/v1alpha1/logicalclusters/cluster" latency="538.409µs" userAgent="kcp/v1.24.3+kcp (linux/amd64) kubernetes/b5751ce/kcp-workspace+root" audit-ID="a0d98940-2691-4798-8f98-a5148703ba03" srcIP="10.19.133.66:38396" resp=403
E0316 17:47:58.446644  572728 workspace_controller.go:225] "kcp-workspace" controller failed to sync "root|e2e-workspace-hktgw", err: logicalclusters.core.kcp.io "cluster" is forbidden: User "external-logical-cluster-admin" cannot get resource "logicalclusters" in API group "core.kcp.io" at the cluster scope: access denied
```

Any pointers on how to make the authz work without the short-circuit logic in the Authorizer like the regular `logical-cluster-admin` ?

pkg/reconciler/tenancy/workspace/workspace_controller.go
davidfestal: I use it in the upcoming Sharded Syncer PR, in this commit: https://github.com/kcp-dev/kcp/pull/2675/commits/d057305bfcf7fcaa62c652dedc36ef823e2cf48a

Backgound is that the virtual workspaces (either in the embedded or standalone mode) need to know the Shard External URL, in order to use it in their logic, especially the Syncer URL (cf our discussion about exposing the shard URL to the syncer as an annotation @sttts).

pkg/authorization/bootstrap/policy.go
hardys: After debugging with @s-urbaniak we concluded that the new `external-logical-cluster-admin` group still needs to be special-cased in the admission and content authorizer code (like the existing `logical-cluster-admin`), however we can avoid the special-casing in the requiredgroups authorizer, and the bootstrap policy rules are significantly more restrictive in the external case.

cmd/sharded-test-server/frontproxy.go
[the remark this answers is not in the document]
s-urbaniak: can we make a small comment here so future readers can reason about it as well?

pkg/admission/logicalcluster/admission.go
```diff
@@ -191,7 +191,7 @@ func (o *plugin) Validate(ctx context.Context, a admission.Attributes, _ admissi
 			return fmt.Errorf("LogicalCluster cannot be deleted: %w", err)
 		}
 		groups := sets.NewString(a.GetUserInfo().GetGroups()...)
-		if !logicalCluster.Spec.DirectlyDeletable && !groups.Has(kuser.SystemPrivilegedGroup) && !groups.Has(bootstrap.SystemLogicalClusterAdmin) {
+		if !logicalCluster.Spec.DirectlyDeletable && !groups.Has(kuser.SystemPrivilegedGroup) && !groups.Has(bootstrap.SystemLogicalClusterAdmin) && !groups.Has(bootstrap.SystemExternalLogicalClusterAdmin) {
```
s-urbaniak, line 194: i may misread: but if the user is external logical cluster admin, it already short-circuits in line 130, no?

pkg/authorization/bootstrap/policy.go
```diff
@@ -39,6 +39,9 @@ const (
 	// This group allows it to skip the entire authorization stack except the bootstrap policy authorizer.
 	// Otherwise, the requests would be rejected because the LogicalCluster resource does not exist yet.
 	SystemLogicalClusterAdmin = "system:kcp:logical-cluster-admin"
+	// SystemExternalLogicalClusterAdmin is a group used by the scheduler to manage LogicalCluster resources.
```
s-urbaniak, line 42: can we be a bit more precise what scheduler we mean?

s-urbaniak: actually, we lack the same precision in the paragraph above as well :D

s-urbaniak: additionally it would be great if we could expand on the difference to `system:kcp:logical-cluster-admin`.

pkg/authorization/workspace_content_authorizer.go
```diff
@@ -116,6 +116,11 @@ func (a *workspaceContentAuthorizer) Authorize(ctx context.Context, attr authori
 		return DelegateAuthorization("logical cluster admin access", a.delegate).Authorize(ctx, attr)
 	}
 
+	// always let external-logical-cluster-admins through
+	if isUser && sets.NewString(attr.GetUser().GetGroups()...).Has(bootstrap.SystemExternalLogicalClusterAdmin) {
```
s-urbaniak, line 120: `sets.NewString(attr.GetUser().GetGroups()...` is now used twice. as this fragment is invoked for every request I suggest to store it in a local variable.

pkg/proxy/options/authentication.go
```diff
@@ -55,7 +56,7 @@ func NewAuthentication() *Authentication {
 			WithServiceAccounts().
 			WithTokenFile(),
 		// when adding new auth methods, also update AdditionalAuthEnabled below
-		DropGroups: []string{user.SystemPrivilegedGroup},
+		DropGroups: []string{user.SystemPrivilegedGroup, bootstrap.SystemLogicalClusterAdmin},
```
s-urbaniak, line 59: can we add a comment why we are doing this for `SystemLogicalClusterAdmin` ?

pkg/admission/logicalcluster/admission.go
hardys: Hmm yes, I just copied the existing `SystemLogicalClusterAdmin` but it seems the conditional should simply be `if !logicalCluster.Spec.DirectlyDeletable` as the other conditions are redundant - I'll add a commit to clean this up, thanks!

pkg/authorization/workspace_content_authorizer.go
hardys: Thanks, it's actually used three times, now moved to a `groups` variable :+1:

pkg/proxy/options/authentication.go
hardys: Done

pkg/admission/logicalcluster/admission.go
hardys: Done

cmd/sharded-test-server/frontproxy.go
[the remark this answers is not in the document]
hardys: Done

hardys:
/test e2e-sharded

Looks like #2865

pkg/reconciler/tenancy/workspace/workspace_reconcile.go
```diff
@@ -97,14 +97,21 @@ func (c *Controller) reconcile(ctx context.Context, ws *tenancyv1alpha1.Workspac
 		return indexers.ByPathAndName[*tenancyv1alpha1.WorkspaceType](tenancyv1alpha1.Resource("workspacetypes"), c.globalWorkspaceTypeIndexer, path, name)
 	}
 
+	// create external client that goes through the front-proxy
+	externalConfig := restclient.CopyConfig(c.externalLogicalClusterAdminConfig)
+	kcpExternalClient, err := kcpclientset.NewForConfig(externalConfig)
```
s-urbaniak, line 102: curiosity question: why does `kcpExternalClient` need to be instantiated at every reconciliation run? can't it be a cached member of `*Controller`?

hardys: Thanks, you're right - I'll move it back to `workspace_controller.go` :+1:

hardys:
@sttts this is ready for another review pass when you get time please - it's blocker for rebasing deployments based on the helm templates

pkg/authorization/workspace_content_authorizer.go
```diff
@@ -112,10 +113,15 @@ func (a *workspaceContentAuthorizer) Authorize(ctx context.Context, attr authori
...
-	if isUser && sets.NewString(attr.GetUser().GetGroups()...).Has(bootstrap.SystemLogicalClusterAdmin) {
+	if isUser && groups.Has(bootstrap.SystemLogicalClusterAdmin) {
 		return DelegateAuthorization("logical cluster admin access", a.delegate).Authorize(ctx, attr)
 	}
 
+	// always let external-logical-cluster-admins through
+	if isUser && groups.Has(bootstrap.SystemExternalLogicalClusterAdmin) {
+		return DelegateAuthorization("external logical cluster admin access", a.delegate).Authorize(ctx, attr)
+	}
+
```
sttts, line 124: this we don't want. Why is it necessary that the external logical cluster admin skips the existence test? It shouldn't.

hardys: I moved it into the switch with the other checks

PR #2907, opened 2023-03-20
:bug: fix(CLI): sync command should honor provided IO streams

astefanutti:
Removes the hard coded `os.Stdout` as output for the `kcp workload sync` command, and relies on the provided IO streams instead.

This makes it easier to test / leverage the syncer command from code.


ncdc (approved):

ncdc:
CI DNS issue, I think they're saying it's resolved now.

/retest

PR #2908, opened 2023-03-20
🌱 TMC E2E tests sharding support - step 1

davidfestal:
## Summary

First step towars sharding support for TMC e2e tests

This PR is a prefactored PR coming from PR https://github.com/kcp-dev/kcp/pull/2675

## Related issue(s)

#2834

### DIFF
--- a/test/e2e/framework/syncer.go
+++ b/test/e2e/framework/syncer.go
@@ -310,41 +310,26 @@ func (sf *syncerFixture) CreateSyncTargetAndApplyToDownstream(t *testing.T) *app
 	ctx, cancelFunc := context.WithCancel(context.Background())
 	t.Cleanup(cancelFunc)
 
-	rawConfig, err := sf.upstreamServer.RawConfig()
-	require.NoError(t, err)
-
 	kcpClusterClient, err := kcpclientset.NewForConfig(syncerConfig.UpstreamConfig)
 	require.NoError(t, err)
-	var virtualWorkspaceURL string
 	var syncTargetClusterName logicalcluster.Name
-	Eventually(t, func() (success bool, reason string) {
-		syncTarget, err := kcpClusterClient.Cluster(syncerConfig.SyncTargetPath).WorkloadV1alpha1().SyncTargets().Get(ctx, syncerConfig.SyncTargetName, metav1.GetOptions{})
-		require.NoError(t, err)
-		if len(syncTarget.Status.VirtualWorkspaces) != 1 {
-			return false, ""
-		}
-		virtualWorkspaceURL = syncTarget.Status.VirtualWorkspaces[0].SyncerURL
-		syncTargetClusterName = logicalcluster.From(syncTarget)
-		return true, "Virtual workspace URL is available"
-	}, wait.ForeverTestTimeout, 100*time.Millisecond, "Syncer Virtual Workspace URL not available")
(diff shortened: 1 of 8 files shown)


test/e2e/framework/syncer.go
```diff
@@ -310,42 +313,26 @@ func (sf *syncerFixture) CreateSyncTargetAndApplyToDownstream(t *testing.T) *app
...
-	var virtualWorkspaceURL string
+	var syncerVirtualWorkspaceURLs []string
+	var upsyncerVirtualWorkspaceURLs []string
 	var syncTargetClusterName logicalcluster.Name
 	Eventually(t, func() (success bool, reason string) {
 		syncTarget, err := kcpClusterClient.Cluster(syncerConfig.SyncTargetPath).WorkloadV1alpha1().SyncTargets().Get(ctx, syncerConfig.SyncTargetName, metav1.GetOptions{})
 		require.NoError(t, err)
-		if len(syncTarget.Status.VirtualWorkspaces) != 1 {
-			return false, ""
+		if len(syncTarget.Status.VirtualWorkspaces) != len(sf.upstreamServer.ShardNames()) {
```
stevekuznetsov, line 324: IMHO it is not sufficient for downstream users to compare the lengths of these two slices to determine if the sync target is "ready" - the consumer should _stream_ the set of virtual workspace URLs and handle them as they pop into and out of existence.

stevekuznetsov: Note that:

 - set equality is really what we want, not length
 - making partial progress when one URL is missing / slow to be posted is desired

davidfestal: > downstream users 

what do you mean here ?

> compare the lengths of these two slices to determine if the sync target is "ready" 

Having the VW URLs set is not what makes the SyncTarget in general. In the Syncer, we watch these URLs and react to any change (when a new shard is added).
But in the context of e2e tests, when the number of shards is a characteristic of the test case (how the shared server is being deployed by the related fixture), we wait until the SyncTarget has the expected number of virtual workspaces).
 
> set equality is really what we want, not length

equality with what ? chacking the context of the SyncTarget VW URLs, to ensure they match (though not being equal with) the VW URLS of the shards, would be the purpose of a dedicated test. I don't think it makes sense. 

> making partial progress when one URL is missing / slow to be posted is desired

I don't really understand the point here, in the e2e fixture.

stevekuznetsov: We should never write code that's end-to-end test specific. All that does is remove useful learning from real use-cases.

> equality with what ? 

Shards A, B, C -> virtual workspace URLs for those shards. Using the current logic we race: shards A B D, URLs A, B, C still is called "ready"

> I don't really understand the point here, in the e2e fixture.

A  static list of URLs is the wrong abstraction with which we should expose the URLs associated with this virtual view. If/when we start having dynamic shard lifecycle in the tests, this will lead to flakes.

sttts:
/approve

@stevekuznetsov had comments. Leaving lgtm to him.

test/e2e/framework/syncer.go
davidfestal: According to what we discussed, I changed the static URL lists in the SyncTarget fixture to be functions.
This we can easily follow the same pattern as what is done for APIExport and not depend anymore on the number of shards.

davidfestal: Fixed in the rebased commit: https://github.com/kcp-dev/kcp/pull/2908/commits/9403375bf7af954a59fe15e6ceea4ff71b10c0bb

PR #2910, opened 2023-03-20
✨ WIP: Also tunnel `service/proxy` requests to support service-based webhooks

davidfestal:
## Summary

Also tunnel `service/proxy` to the synced Service on the physical cluster, the same way as pod sub-resources are tunneled to the physical cluster.

Disclaimer: still in WIP because:
- it needs additional e2e tests for the `services/proxy` case.
- it still needs thought about : how do we manage the case when a service is scheduled to 2 SyncTargets (with 2 placements for example)

## Related issue(s)

First step of issue kcp-dev/contrib-tmc#37


stevekuznetsov:
>it still needs thought about : how do we manage the case when a service is scheduled to 2 SyncTargets (with 2 placements for example)

Could this also affect `Service`s that are backed by a `Deployment` that is spread between clusters? Does this merit a deeper design?

davidfestal:
> > it still needs thought about : how do we manage the case when a service is scheduled to 2 SyncTargets (with 2 placements for example)
> 
> Could this also affect `Service`s that are backed by a `Deployment` that is spread between clusters? Does this merit a deeper design?

Yes, that's what I meant by this. For the case when there are several SyncTargets assigned, this needs a deeper design.

This leads me to the following short-term plan:
- Error when more than 1 SyncTarget is assigned to the service (stll TBD)
- Submit the service proxying to a dedicated feature flag
- Create another issue to more deeply design / implement the multiple-SyncTarget case
- Merge this PR (when ready with additional unit and e2e tests)

Enabling service-based webhooks (which is the first use-case for tunneled service proxying) is quite critical to TMC real use with real existing controllers.

mjudeikis:
/close
TMC

PR #2669, opened 2023-01-24
✨ Make apigen a standalone go module

ncdc:
## Summary
Make apigen a standalone go module so it can be installed by itself.

## Related issue(s)

Fixes #2668

PR #2929, opened 2023-03-29
:sparkles:  Drop apigen module + docs

mjudeikis:
## Summary

~~Move apigen cmd to use new SDK sub-package~~
Remove apigen go module to avoid stale code. 
Update release docs

## Related issue(s)

Fixes #


cmd/apigen/go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/bombsimon/logrusr/v3 v3.1.0
 	github.com/go-logr/logr v1.2.3
 	github.com/google/go-cmp v0.5.9
-	github.com/kcp-dev/kcp/pkg/apis v0.11.0
+	github.com/kcp-dev/kcp/sdk v0.0.0-20230328195704-b5e3a0e1e69a
```
stevekuznetsov, line 9: We want the output of `apigen` to be correct as per the latest commits in the package, on every merge - does this continue to be true implicitly because of `go.work`?

mjudeikis: Good questions. I don't see how this changes the current behaviour as apigen points to apis module which is already released.
If we want to change behaviour where it generates from `main`, we need to use `go.work` in generations jobs, but I think this is is separate work for different PR. If this is the case, lets rase an Issue and track it there?

stevekuznetsov: Do you have an example of a PR that changed API definitions but did not change the generated config? That would be a regression at some point - the point of this tool was to have up-to-date configs ;|

mjudeikis: I suspect regression happened when apigen moved to the module of its own.
@ncdc do you remember why this was done https://github.com/kcp-dev/kcp/pull/2669 ?
We can have few options here:
1. use `go.work` in generate job
2. drop apigen module (my preference)
3. keep `cmd/apigen` as module for easy install but have one `hack/apigen` which would be without module and be using root https://github.com/kcp-dev/kcp/blob/main/go.mod#L193 

If the goal is to keep apigen as "installable" separate from main repo - we need either do 1 or 3.

mjudeikis: When I think more of this - I think we need to drop module in apigen all together and if somebody wants to run its separately - use binary. If binary is not an option - vendoring everything or maintaining your own module would be the path.

mjudeikis: changed to it just removed this go module as I think it causes more issues than delivering value. 
@ncdc your input would be great as you created the initial module for this.

ncdc: What if we moved apigen into the sdk module?

ncdc: We can put tools in the sdk if we'd like. I do like being able to `go install` or `go get` a tool because we have nice conventions for doing that in our Makefiles/scripts, vs. having to `curl` something, but I don't want to get in your way of making progress. We can drop the separate go mod for now, and maybe move it into sdk later if that makes sense. WDYT @stevekuznetsov ?

mjudeikis: @stevekuznetsov whats you take on this? Move to SDK or drop until somebody asks again?

MikeSpreitzer:
FYI, edge-mc uses apigen (see the Makefile).

PR #2930, opened 2023-03-29
:sparkles: add a new field that reports errors in case resource was incompatible

### LINKED ISSUE #2911: bug: Report detailed compatibility errors on incompatible `SyncTarget.Status.ResourcesToSync`
### Describe the bug

When a resource to sync on a SyncTarget is incompatible, there are not details about the compatibility errors between the KCP schema and the SyncTarget schema.
That makes it quite hard to understand what happens.

### Steps To Reproduce

1. Create a SyncTarget that binds to the `root:compute:kubernetes` APIExport
2. Install this SyncTarget in a Kubernetes physical cluster with an old release (1.21 for example)
3. See that some of the resources in the `SyncTarget.Status.ResourcesToSync` list have the `Incompatible` state, but don't provide any more detail.

### Expected Behaviour

We should have a compatibility detailed error message.
This should be possible, since the internal code used to do the compatibility check returns a list of errors (missing field, incompatible types, etc...)

### Additional Context

_No response_

rahulii:
## Summary
This PR add a new field that reports incompatibility errors!
## Related issue(s)

Fixes kcp-dev/contrib-tmc#33

### DIFF
--- a/pkg/reconciler/workload/synctargetexports/synctargetcompatible_reconcile.go
+++ b/pkg/reconciler/workload/synctargetexports/synctargetcompatible_reconcile.go
@@ -123,6 +123,7 @@ func (e *apiCompatibleReconciler) reconcile(ctx context.Context, syncTarget *wor
 				field.NewPath(gvr.String()), upstreamSchema, downStreamSchema, false)
 			if err != nil {
 				syncTarget.Status.SyncedResources[i].State = workloadv1alpha1.ResourceSchemaIncompatibleState
+				syncTarget.Status.SyncedResources[i].Message = err.Error()
 				continue
 			}
 
--- a/pkg/reconciler/workload/synctargetexports/synctargetcompatible_reconcile.go
+++ b/pkg/reconciler/workload/synctargetexports/synctargetcompatible_reconcile.go
@@ -129,9 +129,7 @@ func (e *apiCompatibleReconciler) reconcile(ctx context.Context, syncTarget *wor
 
 			// since version is ordered, so if the current version is comptaible, we can skip the check on other versions.
 			syncTarget.Status.SyncedResources[i].State = workloadv1alpha1.ResourceSchemaAcceptedState
-			if syncTarget.Status.SyncedResources[i].Message != "" {
-				syncTarget.Status.SyncedResources[i].Message = ""
-			}
+			syncTarget.Status.SyncedResources[i].Message = ""
 			break
 		}
 	}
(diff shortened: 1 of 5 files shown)


pkg/reconciler/workload/synctargetexports/synctargetcompatible_reconcile.go
```diff
@@ -123,6 +123,7 @@ func (e *apiCompatibleReconciler) reconcile(ctx context.Context, syncTarget *wor
 				field.NewPath(gvr.String()), upstreamSchema, downStreamSchema, false)
 			if err != nil {
 				syncTarget.Status.SyncedResources[i].State = workloadv1alpha1.ResourceSchemaIncompatibleState
+				syncTarget.Status.SyncedResources[i].Message = err.Error()
```
davidfestal, line 126: Do you reset this field to empty when the incompatibility if fixed ?

sdk/apis/workload/v1alpha1/synctarget_types.go
```diff
@@ -136,6 +136,11 @@ type ResourceToSync struct {
 	// +kubebuilder:default=Pending
 	// +optional
 	State ResourceCompatibleState `json:"state,omitempty"`
+	
+	// A human readable message indicating details about the compatibility errors
```
davidfestal, line 140:
```suggestion
	// A human readable message indicating details about the compatibility errors.
```

pkg/reconciler/workload/synctargetexports/synctargetcompatible_reconcile.go
rahulii: made changes!can you PTAL now?

pkg/reconciler/workload/synctargetexports/synctargetcompatible_reconcile.go
```diff
@@ -123,11 +123,15 @@ func (e *apiCompatibleReconciler) reconcile(ctx context.Context, syncTarget *wor
...
 				continue
 			}
 
 			// since version is ordered, so if the current version is comptaible, we can skip the check on other versions.
 			syncTarget.Status.SyncedResources[i].State = workloadv1alpha1.ResourceSchemaAcceptedState
+			if syncTarget.Status.SyncedResources[i].Message != "" {
```
stevekuznetsov, line 132: nit: no need for the `if` check here, can unconditionally clear it

stevekuznetsov (commented):
/ok-to-test

stevekuznetsov:
If an end-to-end test already exists for this case, can we check that message is set how we expect there?

rahulii:
> If an end-to-end test already exists for this case, can we check that message is set how we expect there?

@stevekuznetsov where are the e2e tests located?

davidfestal:
@rahulii You should be able to add the comparison of the message in the following e2e test:

https://github.com/kcp-dev/kcp/blob/24b8050b6540b5a2913b6db72b7db3878cd5df2f/test/e2e/reconciler/locationworkspace/synctarget_test.go#L118-L124

mjudeikis:
/close
TMC related

PR #2931, opened 2023-03-29
:bug: change state to pending state in case no corresponding APIResourceImp…

### LINKED ISSUE #2928: bug: `syncedResources` state should be Pending when APIResourceImport is missing
### Describe the bug

The SyncTarget reconciler sets the `syncedResources` state (for a given resource) to `Incompatible` when no corresponding APIResourceImport object is found. The `Incompatible` state indicates the resource schema is incompatible with the physical cluster schema, which I think in this case is wrong. The `Pending` state is better as it indicates the syncer was not able to report compatibility of the resource yet.

This line of code should be changed: https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/workload/synctargetexports/synctargetcompatible_reconcile.go#L118

### Steps To Reproduce

Create a SyncTarget.

### Expected Behaviour

Set state to Pending.

### Additional Context

_No response_

rahulii:
…ort object is found

## Summary
This PR changes the state of syncedResources to Pending (earllier it was Incompatible) in case no corresponding APIResourceImport object is found!

## Related issue(s)

Fixes kcp-dev/contrib-tmc#28

### DIFF
--- a/pkg/reconciler/workload/synctargetexports/synctargetcompatible_reconcile_test.go
+++ b/pkg/reconciler/workload/synctargetexports/synctargetcompatible_reconcile_test.go
@@ -84,7 +84,7 @@ func TestSyncTargetCompatibleReconcile(t *testing.T) {
 				}),
 			},
 			wantSyncedResources: []workloadv1alpha1.ResourceToSync{
-				{GroupResource: apisv1alpha1.GroupResource{Group: "apps", Resource: "deployments"}, Versions: []string{"v1"}, State: workloadv1alpha1.ResourceSchemaIncompatibleState},
+				{GroupResource: apisv1alpha1.GroupResource{Group: "apps", Resource: "deployments"}, Versions: []string{"v1"}, State: workloadv1alpha1.ResourceSchemaPendingState},
 			},
 		},
 		{
--- a/test/e2e/reconciler/locationworkspace/synctarget_test.go
+++ b/test/e2e/reconciler/locationworkspace/synctarget_test.go
@@ -116,7 +116,7 @@ func TestSyncTargetExport(t *testing.T) {
 		}
 
 		if syncTarget.Status.SyncedResources[0].Resource != "cowboys" ||
-			syncTarget.Status.SyncedResources[0].State != workloadv1alpha1.ResourceSchemaIncompatibleState {
+			syncTarget.Status.SyncedResources[0].State != workloadv1alpha1.ResourceSchemaPendingState {
 			return false
 		}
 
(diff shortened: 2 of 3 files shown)


rahulii:
cc: @lionelvillard

davidfestal:
> @lionelvillard I fixed the unit test.. but are these e2e test failing due to the changes made?

I think they might be yes.

Could it be that the, according to your changes, the stated expected in the following test should be now be `Pending` instead of `Incompatible` : https://github.com/kcp-dev/kcp/blob/replicate-synctarget-related-rbac/test/e2e/reconciler/locationworkspace/synctarget_test.go#L119 ?

Maybe we need to add some cases in these tests, to differentiate between these 2 states that previously were only one ?

rahulii:
@davidfestal @lionelvillard PTAL!

pkg/reconciler/workload/synctargetexports/synctargetcompatible_reconcile_test.go
```diff
@@ -84,7 +84,7 @@ func TestSyncTargetCompatibleReconcile(t *testing.T) {
 				}),
 			},
 			wantSyncedResources: []workloadv1alpha1.ResourceToSync{
-				{GroupResource: apisv1alpha1.GroupResource{Group: "apps", Resource: "deployments"}, Versions: []string{"v1"}, State: workloadv1alpha1.ResourceSchemaIncompatibleState},
+				{GroupResource: apisv1alpha1.GroupResource{Group: "apps", Resource: "deployments"}, Versions: []string{"v1"}, State: workloadv1alpha1.ResourceSchemaPendingState},
```
davidfestal, line 87: It seems to me that the test case title should be changed.
And probably another test case should be added for the real "incompatible" case

test/e2e/reconciler/locationworkspace/synctarget_test.go
```diff
@@ -116,7 +116,7 @@ func TestSyncTargetExport(t *testing.T) {
 		}
 
 		if syncTarget.Status.SyncedResources[0].Resource != "cowboys" ||
-			syncTarget.Status.SyncedResources[0].State != workloadv1alpha1.ResourceSchemaIncompatibleState {
+			syncTarget.Status.SyncedResources[0].State != workloadv1alpha1.ResourceSchemaPendingState {
```
davidfestal, line 119: Since we're not testing the incompatible state anymore, maybe an additional resource could be added that would be incompatible.
This could be done by:
- creating an additional APIExport with 1 APIResourceSchema in a workspace
- adding the same CRD in the physical cluster (by `WithDownstreamPreparation` when creating the SyncerFixture), but with a small field of the CRD schema changed (for example changing string to int)
- checking that this reosurce on the Synctarget is in the Incompatible state

rahulii: https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/workload/synctargetexports/synctargetcompatible_reconcile_test.go#L119

I think the test case is already there!
@davidfestal

rahulii: @davidfestal is this needs to be part of same PR?

mjudeikis:
/close
TMC related

PR #2944, opened 2023-04-04
:sparkles: pkg/indexers: add helers for cache server fallback

stevekuznetsov:
Every single consumer of these functions operating over replicated data needs to use the same exact pattern to access their data. There's no need to make them all duplicate the logic.

/cc @nrb @sttts 
/assign @p0lyn0mial

### DIFF
--- a/pkg/indexers/indexers.go
+++ b/pkg/indexers/indexers.go
@@ -127,6 +127,23 @@ func ByIndex[T runtime.Object](indexer cache.Indexer, indexName, indexValue stri
 	return ret, nil
 }
 
+// ByIndexWithFallback returns all instances of T that match indexValue in indexer, if any. If none match,
+// the same query is done of globalIndexer. Any errors short-circuit this logic.
+func ByIndexWithFallback[T runtime.Object](indexer, globalIndexer cache.Indexer, indexName, indexValue string) ([]T, error) {
+	// Try local informer first
+	results, err := ByIndex[T](indexer, indexName, indexValue)
+	if err != nil {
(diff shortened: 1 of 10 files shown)


stevekuznetsov:
Test failed on ... DNS?

pkg/indexers/indexers.go
```diff
@@ -142,3 +159,21 @@ func ByPathAndName[T runtime.Object](groupResource schema.GroupResource, indexer
 	}
 	return objs[0].(T), nil
 }
+
+// ByPathAndNameWithFallback returns the instance of T from the indexer with the matching path and name. Path may be a canonical path
+// or a cluster name. Note: this depends on the presence of the optional "kcp.io/path" annotation. If no instance is found, globalIndexer
+// is searched as well. Any errors short-circuit this logic.
+func ByPathAndNameWithFallback[T runtime.Object](groupResource schema.GroupResource, indexer, globalIndexer cache.Indexer, path logicalcluster.Path, name string) (ret T, err error) {
```
p0lyn0mial, line 166: Why do we need both `ByIndexWithFallback` and `ByPathAndNameWithFallback` methods? 
It seems that both methods eventually make use of `indexer.ByIndex` :)

The difference seems to be that `ByIndex` returns an array of object whereas `ByPathAndName` returns a single object.

Do we need this distinction or could we unify into one method?

stevekuznetsov: @p0lyn0mial please feel free to propose a simpler factoring! One was not obvious to me since the error checking is different.

PR #2946, opened 2023-04-04
:sparkles: Make the tunneler shard-aware

### LINKED ISSUE #2943: Create one tunneler per shard in the Syncer
With sharding, every shard that contains user workspaces synced to a given syncer should have a tunnel established by the Syncer.
Until now, only one tunneler was created by the Syncer, to the shard that contained the Synctarget location workspace.

Now the Syncer should connect a tunneler for each shard.

davidfestal:
## Summary

Make the tunneler shard-aware:
- use global sync target informer on the KCP side
- create 1 tunneler per shard on the Syncer side.

Based on PR #2675. Only the 3 last commits are part of this PR.

## Related issue(s)

Fixes #2943


cmd/virtual-workspaces/options/options.go
```diff
@@ -94,6 +96,8 @@ func (o *Options) AddFlags(flags *pflag.FlagSet) {
 	o.CoreVirtualWorkspaces.AddFlags(flags)
 	o.TmcVirtualWorkspaces.AddFlags(flags)
 
+	flags.StringVar(&o.ShardExternalURL, "shard-external-url", o.ShardExternalURL, "URL used by outside clients to talk to the kcp shard this virtual workspace is related to")
```
stevekuznetsov, line 99: Are we certain that now and forever shards will be 1:1 with VW servers? Why? It's not clear to me that these need to be related, the scaling dimensions are different and one does not have storage.

sdk/apis/workload/v1alpha1/types.go
```diff
@@ -188,4 +188,15 @@ const (
 	// and further work should be done to define such (up)syncing strategies at a more appropriate level
 	// (SyncTarget, KCP namespace, KCP workspace ?).
 	ExperimentalUpsyncDerivedResourcesAnnotationKey = "experimental.workload.kcp.io/upsync-derived-resources"
+
+	// InternalWorkspaceURLAnnotationKey is an annotation dynamically added on resources exposed
+	// by the Syncer Virtual Workspace to be synced by the Syncer.
+	// It contains the external URL of the workspace the resource is part of.
```
stevekuznetsov, line 194: Moving workspaces between shards should be O(1) or as close as we can get to it. Having to batch-mutate N objects is a no-go I think.

pkg/reconciler/workload/synctarget/synctarget_reconcile.go
```diff
@@ -48,6 +48,10 @@ func (c *Controller) reconcile(ctx context.Context, syncTarget *workloadv1alpha1
 
 	var rootShardKey string
 	for _, workspaceShard := range workspaceShards {
+		// Don't include shards whose sole purpose is to be used in tests, but which might contain unreachable URLs
+		if workspaceShard.Annotations["experimental.core.kcp.io/unschedulable"] == "true" {
```
stevekuznetsov, line 53: We 100000% should not write test-specific code here. Every single part of this system needs to handle a) a dynamic list of shards and b) shards being potentially unreachable for a short moment.

pkg/syncer/syncer.go
```diff
@@ -282,11 +282,25 @@ func StartSyncer(ctx context.Context, cfg *SyncerConfig, numSyncerThreads int, i
...
 			var cacheSyncsForAlwaysRequiredGVRs []cache.InformerSynced
+			waitForCacheSyncErrors := map[schema.GroupVersionResource]map[string]int{}
 			for _, alwaysRequiredGVR := range alwaysRequiredGVRs {
 				if informer, err := ddsifForUpstreamSyncer.ForResource(alwaysRequiredGVR); err != nil {
 					return nil, err
 				} else {
+					errors := make(map[string]int)
+					_ = informer.Informer().SetWatchErrorHandler(func(r *cache.Reflector, err error) {
+						errStr := err.Error()
+						errors[errStr]++
```
stevekuznetsov, line 298: I'm not 100% certain but I am pretty sure that the watch error handling can happen in a different goroutine, so we would need some mutexes around this.

pkg/syncer/synctarget/tunneler_reconciler.go
```diff
@@ -0,0 +1,80 @@
...
+		// Create the tunneler
+		c.startShardTunneler(shardTunnelerContext, shardURL)
+		c.tunnelers[shardURL] = tunneler{
+			cancelFunc,
+		}
+	}
+	return reconcileStatusContinue, utilserrors.NewAggregate(errs)
+}
+
+type tunneler struct {
```
stevekuznetsov, line 78: Why do we need to allocate a new struct just to hold onto the function pointer?

cmd/virtual-workspaces/options/options.go
davidfestal: That's not exactly 1:1, but a given virtual workspace is related to one and only one shard.

@sttts Do you want to provide additional context here ?

sdk/apis/workload/v1alpha1/types.go
davidfestal: Nothing is batch-mutated here.
This annotation is added "on-the-fly" by the Syncer virtual workspace when exposing resources to the Syncer.

pkg/reconciler/workload/synctarget/synctarget_reconcile.go
davidfestal: > Every single part of this system needs to handle a) a dynamic list of shards and b) shards being potentially unreachable for a short moment.

That's the case.

As concerning skipping shards with this annotation, the [declaration of the annotation itself](https://github.com/kcp-dev/kcp/blob/sharded-tunneler/pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling.go#L61) mentions that shards so annotated are dedicated to e2e tests and should not be taken in account. They are [not taken in account by the workspace scheduling](https://github.com/kcp-dev/kcp/blob/sharded-tunneler/pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling.go#L209).

So I don't see why they should be taken in account by the Syncer.

pkg/syncer/synctarget/tunneler_reconciler.go
davidfestal: Just a rest of copied code. I'll clean this up.

pkg/syncer/syncer.go
davidfestal: :+1:

pkg/reconciler/workload/synctarget/synctarget_reconcile.go
sttts: +1 that this looks wrong. Unschedulable only means that these shards should be ignored during scheduling. Every component in the system must be able to cope with this status, and even already scheduled workloads must keep functioning.

cmd/virtual-workspaces/options/options.go
stevekuznetsov: "is related to one and only one" ... is what I mean by `1:1` ??

pkg/syncer/synctarget/tunneler_reconciler.go
davidfestal: Fixed

pkg/syncer/syncer.go
davidfestal: This code block was removed.

pkg/reconciler/workload/synctarget/synctarget_reconcile.go
davidfestal: commit removed, since the syncer is now robust and supports unreachable Shard URLs

cmd/virtual-workspaces/options/options.go
davidfestal: You can have a single Shard external URL for several shards (for example behind a proxy).
But you would never have a single virtual workspace URL for several shards. There is (at least) 1 per shard.
So the shard external URL is always well-defined for a given virtual workspace, though a single shard external URL
could be shared by several virtual workspaces.

PR #2950, opened 2023-05-09
📖 Add rebase documentation

ncdc:
## Summary
Add documentation on the rebase process.

### DIFF
--- a/docs/content/developers/rebasing-kubernetes.md
+++ b/docs/content/developers/rebasing-kubernetes.md
@@ -1,0 +1,245 @@
+# Rebasing Kubernetes
+
+This describes the process of rebasing kcp onto a new Kubernetes version. For the examples below, we'll be rebasing 
+onto v1.26.3.
+
+# 1. Update kcp-dev/apimachinery
+
+1. Create a new branch for the update, such as `1.26-prep`.
+2. Update go.mod:
+   1. You may need to change the go version at the top of the file to match what's in go.mod in the root of the 
+      Kubernetes repo. This is not required, but probably a good idea.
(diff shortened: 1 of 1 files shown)


stevekuznetsov (approved):
This looks like all the highlights - nice write-up!

xrstf:
This is great and will be an amazing help in the future :-)

waltforme:
Maybe a dumb question: Why does kcp-dev/apimachinery `require`s k8s.io/api? Sounds like a 'reversed' dependency? Thanks!

stevekuznetsov:
@waltforme I don't know where the `kcp-dev/apimachinery` name came from in the beginning, to be honest. It's not related to `k8s.io/apimachinery` and untangling it might be a good use of time. The contents in the repo are related like this:

- [`kcp-dev/apimachinery/pkg/cache`](https://github.com/kcp-dev/apimachinery/tree/main/pkg/cache) is similar to `k8s.io/client-go/tools/cache` [here](https://github.com/kubernetes/client-go/tree/master/tools/cache)
- [`kcp-dev/apimachinery/pkg/client`](https://github.com/kcp-dev/apimachinery/tree/main/pkg/client) is the machinery behind the cross-cluster clients generated by [`kcp-dev/code-generator`](https://github.com/kcp-dev/code-generator) and committed to [`kcp-dev/client-go`](https://github.com/kcp-dev/client-go)

Perhaps deleting the repo and subsuming the content into the more appropriate places would be good.

waltforme:
I see. Now the dependency makes sense. Also it explains why the repository is so small.
Maybe I should take a closer look before asking this.
Thanks for the detailed explanations! @stevekuznetsov

Issue #2954, opened 2023-05-16
epic: Remove TMC from the KCP repo 


scheeles:
As discussed in the last KCP edge meeting, the focus in the KCP repos should be on the core.

This means that TMC should be removed from this repo. If later someone shows interest in maintaining TMC, the code can be move to a dedicated repo.

PR #2955, opened 2023-05-16
:warning: Remove deployment-coordinator [WIP]

xrstf:
## Summary
This PR is part of a series to work on #2954. As `tmc/pkg/` is used in for example `cmd/kcp/`, it's not a trivial `rm -f tmc/` to get rid of the code, but this PR is a much simpler first change.


davidfestal:
Wouldn't it make more sense to move the TMC things inside the `/tmc` repository, so that it can then be globally moved to another GH repo ? Instead of fully removing only part of the TMC code ?

davidfestal:
wdyt @sttts ?

xrstf:
I was already queueing up to also remove cmd/syncer and related code in another PR, as so far I am not aware of any maintainers for that part of the codebase.

If a standalone repository is a more appropriate resting place, I'm not quite sure how to extract the TMC stuff in a meaningful way, short of just forking `kcp-dev/kcp` into `kcp-dev/kcp-with-tmc` ? What would you suggest? :)

sttts:
> Wouldn't it make more sense to move the TMC things inside the /tmc repository, so that it can then be globally moved to another GH repo ? Instead of fully removing only part of the TMC code ?

@davidfestal moving things into a tmc directory and keep everything working is more work I guess. An alternative approach might be to label all PRs removing tmc code. Then one can reconstruct what we removed.

xrstf:
We could also just keep a branch name on the commit before the first cleanup PR is merged.

davidfestal:
> We could also just keep a branch name on the commit before the first cleanup PR is merged.

I like this idea.

sttts:
>> We could also just keep a branch name on the commit before the first cleanup PR is merged.
> I like this idea.

Let's do it.

sttts:
https://github.com/kcp-dev/kcp/tree/main-pre-tmc-removal

xrstf:
This has been superseded by #2963, which removes more code and has been rebased on the 1.26-rebase already.

PR #2945, opened 2023-04-04
✨ Update to Kubernetes 1.26.3

ncdc:
## Summary
Update to Kubernetes 1.26.3.

## Todos
- [x] Unit tests pass
- [x] Boilerplate
- [x] Lint
- [ ] Update go.mod to real Kube branch
- [x] Enable ValidatingAdmissionPolicy

## Related issue(s)

Fixes #2772

PR #2958, opened 2023-05-19
:bug: logicalcluster-deletion: skip bound resources

sttts:
Skip deleting bound resources in the logicalcluster deletor as this can
race with the apibinding deletion controller. If the apibinding deletion
controller deletes an APIBinding and all its instances and then the
logicalcluster deletor tries to delete the collection of the same
resources (because it saw the resource in discovery), it will get a 404
and go into rate limited backoff, which we don't want.

Taken from @ncdc's rebase PR https://github.com/kcp-dev/kcp/pull/2945.


xrstf:
Closing & re-opening this PR to migrate it to the new Prow. Please do not be alarmed.

mjudeikis:
@sttts is this still valid to merge? Looks green'ish

sttts:
Yes. I just extracted it at the time because we didn't know where some flakes came from.

Issue #2479, opened 2022-12-12
bug: flake: workspace(s) not initialized in time


ncdc:
### Describe the bug

We've lately been seeing random test failures that look like this:

```
=== FAIL: test/e2e/apibinding TestProtectedAPI (30.83s)
    kcp.go:127: shared kcp server will target configuration "/go/src/github.com/kcp-dev/kcp/.kcp/admin.kubeconfig"
    assertions.go:1691: Waiting for condition, but got: workspace phase is Initializing, not Ready
    workspaces.go:182: 
        	Error Trace:	util.go:297
        	            				workspaces.go:182
        	            				apibinding_protected_test.go:50
        	Error:      	Condition never satisfied
        	Test:       	TestProtectedAPI
        	Messages:   	failed to wait for organization workspace e2e-org-xgk49 to become ready
```

### Steps To Reproduce

1. Run e2e
2. Sometimes it fails like this

### Expected Behaviour

No failure

ncdc:
Example: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kcp-dev_kcp/2475/pull-ci-kcp-dev-kcp-main-e2e-sharded/1602363246315048960

stevekuznetsov:
We need to amend that `framework.Eventually` to spit out to conditions/reasons and/or add the ClusterWorkspace object to the artifacts, or this is not something that can be debugged post-hoc.

ncdc:
One specific (new?) reason this happens: https://github.com/kcp-dev/kcp/pull/2440#issuecomment-1371202704

ncdc:
Saw this happen again in https://github.com/kcp-dev/kcp/actions/runs/3968949742/jobs/6802786983. Did some digging into the logs, and it looks like the `kcp-workspacetypes-bootstrap-root:universal` and `kcp-apibinder-initializer` reconcilers fight with each other for a bit over patching the `LogicalCluster`. Eventually the bootstrapper bit completes (removing its initializer).

Then the apibinder initializer tries multiple times to remove its initializer. I see this over and over:

```
[...]

PR #2961, opened 2023-05-21
🐛 Fix KCPRequestInfoResolver for kind: Cluster

### LINKED ISSUE #2811: bug: Status subresource returns 404 for any CR with Kind: Cluster
### Describe the bug

While playing around with integration between Cluster API and KCP I hit an issue where KCP returns 404 whenever we try to patch the /Status subresource on objects with Kind: Cluster.

In order to double-check the issue I have re-created the problem using wild west example and changing the Kind in it (see below)

### Steps To Reproduce

Apply the following resources:

<details><summary>APIResourceSchema</summary>

```yaml
apiVersion: apis.kcp.io/v1alpha1
kind: APIResourceSchema
metadata:
  name: v230208-22d06c1f.clusters.wildwest.dev
spec:
  group: wildwest.dev
  names:
    kind: Cluster
    listKind: ClusterList
    plural: clusters
    singular: cluster
  scope: Namespaced
  versions:
    - name: v1alpha1
      schema:
        description: Cowboy is part of the wild west
        properties:
          apiVersion:
            description: 'APIVersion defines the versioned schema of this representation
            of an object. Servers should convert recognized schemas to the latest
            internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
            type: string
          kind:
            description: 'Kind is a string value representing the REST resource this
            object represents. Servers may infer this from the endpoint the client
            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
            type: string
          metadata:
            type: object
          spec:
            description: CowboySpec holds the desired state of the Cowboy.
            properties:
              intent:
                type: string
            type: object
          status:
            description: CowboyStatus communicates the observed state of the Cowboy.
            properties:
              result:
                type: string
            type: object
        type: object
      served: true
      storage: true
      subresources:
        status: {}
```
</details>

<details><summary>APIExport</summary>

```yaml
apiVersion: apis.kcp.io/v1alpha1
kind: APIExport
metadata:
  creationTimestamp: null
  name: wildwest-dev
spec:
  latestResourceSchemas:
    - v230208-22d06c1f.clusters.wildwest.dev
status: {}
```
</details>

<details><summary>APIBinding (if running in a workspace different than root, adjust the path accordingly)</summary>

```yaml
apiVersion: apis.kcp.io/v1alpha1
kind: APIBinding
metadata:
  name: wildwest-dev
spec:
  reference:
    export:
      name: wildwest-dev
      path: root
```
</details>

<details><summary>Sample Cluster</summary>

```yaml
apiVersion: wildwest.dev/v1alpha1
kind: Cluster
metadata:
  name: woody
spec: {}
```
</details>

then run 
```
curl -k -H "Authorization: Bearer <your token>"  https://192.168.1.77:6443/clusters/root/apis/wildwest.dev/v1alpha1/namespaces/default/clusters/woody/status
```

You will get "404 page not found"

### Expected Behaviour

To be able to use custom resources of Kind: Cluster

### Additional Context

_No response_

nikhita:
## Summary

The cache server preprocesses the path to remove the cluster prefix in https://github.com/kcp-dev/kcp/blob/88d6071b2aaf47bcd5e854f8209e4c3bd8035b35/pkg/cache/server/config.go#L172-L186

When a resource of `kind: Cluster` is created and the first cluster prefix is dropped, the path would be a k8s request info path (`/apis/...`) but will still have the `clusters` keyword in the path.

Due to this, the regex in the kcp request info resolver does not work well.

This commit updates the resolver to include the following changes:

1. If a path is a k8s request info path (begins with `/api` or `/apis`), then the path is not checked against the regex.

2. The regex is updated to preserve the forward slash in the strippedURL.

## Related issue(s)

Fixes #2811

### DIFF
--- a/pkg/server/requestinfo/kcp_request_info_resolver.go
+++ b/pkg/server/requestinfo/kcp_request_info_resolver.go
@@ -19,6 +19,7 @@ package requestinfo
 import (
 	"net/http"
 	"regexp"
+	"strings"
 
 	"k8s.io/apiserver/pkg/endpoints/request"
 )
@@ -33,16 +34,28 @@ func NewKCPRequestInfoResolver() *KCPRequestInfoResolver {
 	}
 }
 
-var clustersRE = regexp.MustCompile(`/clusters/[^/]+/(.*)$`)
+var clustersRE = regexp.MustCompile(`/clusters/[^/]+(/.*)$`)
 
 func (k *KCPRequestInfoResolver) NewRequestInfo(req *http.Request) (*request.RequestInfo, error) {
-	matches := clustersRE.FindStringSubmatch(req.URL.Path)
-	if len(matches) == 2 {
-		// matches[0] is the leftmost pattern that matches (which includes /clusters/...) - skip over that
-		strippedURL := matches[1]
-		t := req.Clone(req.Context())
-		t.URL.Path = strippedURL
-		return k.delegate.NewRequestInfo(t)
+	if !containsPrefix([]string{"/apis/", "/api/"}, req.URL.Path) {
+		matches := clustersRE.FindStringSubmatch(req.URL.Path)
+		if len(matches) == 2 {
+			// matches[0] is the leftmost pattern that matches (which includes /clusters/...) - skip over that
+			strippedURL := matches[1]
+			t := req.Clone(req.Context())
+			t.URL.Path = strippedURL
+			return k.delegate.NewRequestInfo(t)
+		}
 	}
 	return k.delegate.NewRequestInfo(req)
+}
+
+// containsPrefix returns true if the specified path has a prefix that contained in given prefix Set.
+func containsPrefix(prefixSet []string, path string) bool {
+	for _, prefix := range prefixSet {
+		if strings.HasPrefix(path, prefix) {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/server/requestinfo/kcp_request_info_resolver_test.go
+++ b/pkg/server/requestinfo/kcp_request_info_resolver_test.go
@@ -1,0 +1,114 @@
+/*
+Copyright 2023 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package requestinfo
+
+import (
+	"net/http"
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestRequestInfoResolver(t *testing.T) {
+	tests := map[string]struct {
+		path                string
+		expectedPath        string
+		expectedResource    string
+		expectedSubresource string
+	}{
+		"path with /api prefix": {
+			path:             "/api/v1/namespaces/default/pods",
+			expectedPath:     "/api/v1/namespaces/default/pods",
+			expectedResource: "pods",
+		},
+		"path with /apis prefix": {
+			path:             "/apis/wildwest.dev/v1alpha1/namespaces/default/cowboys/woody",
+			expectedPath:     "/apis/wildwest.dev/v1alpha1/namespaces/default/cowboys/woody",
+			expectedResource: "cowboys",
+		},
+		"path with /apis prefix -- status subresource": {
+			path:                "/apis/wildwest.dev/v1alpha1/namespaces/default/cowboys/woody/status",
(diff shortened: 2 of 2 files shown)


sttts:
/ok-to-test

Invited you to the org.

pkg/server/requestinfo/kcp_request_info_resolver_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+func TestRequestInfoResolver(t *testing.T) {
+	tests := map[string]struct {
+		path                string
+		expectedPath        string
+		expectedResource    string
+		expectedSubresource string
+	}{
+		"path with /api prefix": {
+			path:             "/api/v1/namespaces/default/pods",
```
stevekuznetsov, line 35: Why are these valid paths? You need to prefix by cluster or service, there's no default logical cluster to access without specifying.

sttts: at some point we redirected to `system:admin`, but disabled or broke that behaviour. Now the behaviour is a nasty error messages in the log. We should decide how to move forward and make it an hard error early in the stack, e.g. here.

sttts: Btw, these tests here might (or not, to be verifiy) actually be correct. In `WithInClusterServiceAccountRequestRewrite` we handle the case of a service account calling back without the `/cluster` prefix, but with the cluster set in the JWT.

pkg/server/requestinfo/kcp_request_info_resolver.go
```diff
@@ -33,16 +34,28 @@ func NewKCPRequestInfoResolver() *KCPRequestInfoResolver {
 	}
 }
 
-var clustersRE = regexp.MustCompile(`/clusters/[^/]+/(.*)$`)
+var clustersRE = regexp.MustCompile(`/clusters/[^/]+(/.*)$`)
```
sttts, line 37: am curious: why don't we start with `^` aka match the beginning of the string? Now we would also match `/foo/cluster/...`. Don't think that's intentional.

pkg/server/requestinfo/kcp_request_info_resolver_test.go
nikhita: We also strip the `/clusters/..` prefix here before passing to the RequestInfoResolver:

https://github.com/kcp-dev/kcp/blob/88d6071b2aaf47bcd5e854f8209e4c3bd8035b35/pkg/cache/server/config.go#L172-L186

pkg/server/requestinfo/kcp_request_info_resolver.go
nikhita: >  am curious: why don't we start with ^ aka match the beginning of the string? 

This is to handle all `/services/...` URLs. The request resolver was actually added to handle the `/services/...` URLs - https://github.com/kcp-dev/kcp/commit/963ffdeb5818f1c125b6e1dd8e28f05127062228

Some more additional context is in https://github.com/kcp-dev/kcp/issues/2479#issuecomment-1404177658

sttts: the service URLs rechnically have nothing to do with the main kcp server. Would rather expect them to strip the prefix first before apply this resolver.

pkg/server/requestinfo/kcp_request_info_resolver.go
```diff
@@ -33,16 +34,28 @@ func NewKCPRequestInfoResolver() *KCPRequestInfoResolver {
...
 
 func (k *KCPRequestInfoResolver) NewRequestInfo(req *http.Request) (*request.RequestInfo, error) {
-	matches := clustersRE.FindStringSubmatch(req.URL.Path)
-	if len(matches) == 2 {
-		// matches[0] is the leftmost pattern that matches (which includes /clusters/...) - skip over that
-		strippedURL := matches[1]
-		t := req.Clone(req.Context())
-		t.URL.Path = strippedURL
-		return k.delegate.NewRequestInfo(t)
+	if !containsPrefix([]string{"/apis/", "/api/"}, req.URL.Path) {
```
sttts, line 40: with my comment [above](https://github.com/kcp-dev/kcp/pull/2961/files#r1228275536), this line could go and we would be back at the old code, right?

pkg/server/requestinfo/kcp_request_info_resolver_test.go
sttts: if we strip the cluster prefix, why do we need this special requestinfo resolver in the first place?

pkg/server/requestinfo/kcp_request_info_resolver.go
mjudeikis: Im bit lost. Will this solve :ncdc: comment about 60s timeouts? As if I want to build service with streaming to VW backed services, this might be an issue.?

mjudeikis: So expectations is that WE ALWAYS trim first prefix for `/services`, and first `/clusters`?

PR #2974, opened 2023-06-06
:seedling:  Add release notes into PR template

mjudeikis:
## Related issue(s)

Fixes #

```release-note
NONE
```

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -19,3 +19,17 @@ Please copy the appropriate `:text:` or icon to the beginning of your PR title:
 ## Related issue(s)
 
 Fixes #
+
+## Release Notes
+
+<!--
+Please add a release note using the following format:
+
+```release-note
+<description of change>
+```
+-->
+
+```release-note
+NONE
+```


xrstf:
Before we merge this, maybe we should first revisit my hasty decision to enable the release-note plugin.

Currently we use goreleaser, which will not use the release-note block. We used it at Kubermatic with a custom changelog generator, inspired by Kubernetes upstream. Do we want this in kcp even, or do we want to continue using goreleaser?

Without goreleaser, I guess there is more work to be done in preparing the Github release info.

@sttts WDYT?

mjudeikis:
I like release notes plugin, makes things easier and PR author responsibility to write meaningful messages. Especially when we don't have release team

embik (approved):

embik:
I too like the release plugin.

sttts:
> @sttts WDYT?

@xrstf no strong opinions. We should use what has proven to work well and minimizes work. What is that special goreleaser you are talking about? A fork?

xrstf:
> What is that special goreleaser you are talking about? A fork?

https://github.com/kubermatic/gchl

AFAIK it did not start as a fork of any existing tooling. It's a simple tool to use the Github API to query for PRs, then extract the release-notes, reformat them a bit, sort them alphabetically and output markdown on stdout. Compared to kubernetes tooling (I think) it does not attempt any grouping ("API Changes", "New Features", "...") itself, but relies on a human to postprocess the Markdown.

sttts:
I like the release note block too. Looks like we all agree on that. 

@xrstf `https://github.com/kubermatic/gchl` IMO is worth a try. @mjudeikis wants to try doing a release soonish. Maybe the right time to switch over and see how it goes?

xrstf:
Cool, then let's merge this.

I can

/lgtm

someone else needs to /approve :grin:

embik:
I don't know why but the merge pool was empty although this PR looked good, so I hope the retest will move it over the finish line and let prow merge it.

PR #2975, opened 2023-06-06
:seedling: Add Prowjobs

xrstf:
## Summary
This moves the Prowjobs that were previously part of the openshift/release repo into a simple in-repo configuration (i.e. a `.prow.yaml`).

I combined the various verify tasks into a single `pull-kcp-verify` job, but had to pull the codegen task out into its own Prow job, as running it as part of other tasks seems to make it get stuck and I have not further investigated why. `client-gen` seems to just sit forever (I checked the pod, CPU usage was basically idling).

part of kcp-dev/infra#43

## Release Notes
```release-note
NONE
```

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,4 +1,0 @@
-build_root_image:
-  namespace: ci
-  name: kcp-dev-build-root
-  tag: "1.19"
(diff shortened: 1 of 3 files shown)


sttts:
How can we set the required jobs? Just manually or is there something in prow?

xrstf:
All jobs that were executed are required by default, unless a job has `optional: true` set in the `.prow.yaml`. On Github's end all we need to do is make `tide` a required status check.

Issue #2978, opened 2023-06-11
bug: go install apigen does not work


mjudeikis:
### Describe the bug

After `sdk` consolidation you can't install apigen via `go install` due to bellow error:
```
 go install -tags tools "github.com/kcp-dev/kcp/sdk/cmd/apigen@latest"
go: github.com/kcp-dev/kcp/sdk/cmd/apigen@latest (in github.com/kcp-dev/kcp/sdk@v0.0.0-20230609202750-bc275d9c1b5d):
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.
```

Relate to this: https://github.com/golang/go/issues/44840

We should remove `replace` statements from SDK. A consequence of this will be that it will use the "frozen in time" released version of APIs to generate content. Which might not be desired in the development flow. 

As the original intent to move apigen was so we can do `go install` and at the same time use the latest/main content, this didn't worked. 

### Steps To Reproduce

run `github.com/kcp-dev/kcp/sdk/cmd/apigen@latest` in fresh repo outside kcp

### Expected Behaviour

works

mjudeikis:
But before we do this, we need to cut new tag release

mjudeikis:
hacked it like this now:
```
# KO_DEFAULTBASEIMAGE=bitnami/git KO_DOCKER_REPO=quay.io/mangirdas/kcp-apigen ko publish --platform linux/amd64,linux/ppc64le,linux/arm64 --sbom=none -B cmd/apigen/*
    docker run --user $(id -u) --workdir /data -v $(pwd):/data:z quay.io/mangirdas/kcp-apigen/main.go:latest --input-dir /data/config/crds --output-dir /data/config/root-phase0 
```

mjudeikis:
fixed in 0.20

PR #2979, opened 2023-06-11
:bug: Remove replace stanzas

mjudeikis:
## Summary

I think adding those was a mistake as they prevent `go install` so removing

## Related issue(s)

Fixes #https://github.com/kcp-dev/kcp/issues/2978

```release-note
NONE
```


sttts:
Doesn't have this the side-effect that we depend on out-of-repo versions?

PR #2996, opened 2023-07-18
:sparkles: Add leader elections for controllers

### LINKED ISSUE #2216: feature: add leader election for controllers
### Feature Description

As a kcp platform operator, I would like to be able to run multiple copies of the kcp process for high availability.

As a kcp platform operator, I need to ensure that multiple copies of the controllers do not fight with each other.

### Proposed Solution

We can enable leader election (optionally) so only 1 copy of each controller is active at a time.

### Alternative Solutions

_No response_

### Want to contribute?

### Additional Context

_No response_

embik:
## Summary

This PR adds leader elections for controllers in kcp. I will add a better summary documenting the full implementation later, but this is a "early preview" so the branch has some visibility.

## Related issue(s)

Fixes #2216

## Release Notes

```release-note
Add leader elections to kcp that can enabled with `--enable-leader-election` flag
```


pkg/server/server.go
```diff
@@ -517,3 +540,63 @@ func goContext(parent genericapiserver.PostStartHookContext) context.Context {
...
+				logger.Info("started leading")
+
+				for controllerName, controller := range s.controllers {
+					logger := logger.WithValues("controller", controllerName)
+					go s.runController(ctx, controller, logger)
+				}
+			},
+			OnStoppedLeading: func() {
+				logger.Info("stopped leading")
+			},
```
MadhavJivrajani, line 587: I'm trying to work through the scenario which looks like:
```
starts leading -> stops leading -> starts leading
```
If this were to happen, we end up re-spawning all registered controllers.

pkg/server/server.go
```diff
@@ -517,3 +540,63 @@ func goContext(parent genericapiserver.PostStartHookContext) context.Context {
...
+				}
+
+				logger.WithValues("new-leader", current_id).Info("new leader is someone else")
+			},
+		},
+		WatchDog: leaderelection.NewLeaderHealthzAdaptor(time.Second * 5),
+		Name:     "kcp-controllers",
+	})
+
+	logger.Info("leader election has stopped")
```
MadhavJivrajani, line 601: `RunOrDie` exits either when some error occurs (through a `panic`), the `ctx` is cancelled or if the lease couldn't be renewed. If lease couldn't be renewed, we exit this function entirely and don't re-enter it again.

pkg/server/server.go
```diff
@@ -491,6 +498,22 @@ func (s *Server) Run(ctx context.Context) error {
...
+	// add post start hook that starts all registered controllers.
+	s.AddPostStartHook("kcp-start-controllers", func(hookContext genericapiserver.PostStartHookContext) error {
+		logger := klog.FromContext(ctx).WithValues("postStartHook", "kcp-start-controllers")
+
+		if s.ExtraConfig.LeaderElectionEnabled {
+			go s.leaderElectAndRun(goContext(hookContext))
+		} else {
+			for controllerName, controller := range s.controllers {
+				logger := logger.WithValues("controller", controllerName)
+				go s.runController(goContext(hookContext), controller, logger)
```
MadhavJivrajani, line 510: nit: this bit of logic to start registered controllers can probably be put into a function `startControllers` so that we can abstract away anything else needed like recovery handlers etc. that might be needed to add in later on.

MadhavJivrajani (commented):
Hey @embik 👋🏼 
Finally getting around to reviewing this, apologies for the delay.

From an initial pass, I just have a couple of questions around leader state transitions and what happens when those occur.

For the sake of completeness, here is the document that we were working on to compare how controllers were "managed" by kcm and kcp: https://hackmd.io/bZhJ7dtoT4qfqf8q9u9hng?view

The way kcm handles `OnStoppedLeading` answers my questions, but I'm not sure if we can do that since afaict, the kcp server is run in the foreground as a standalone binary, so we probably can't exit and rely on a daemon to bring it back.

pkg/server/server.go
embik: My understanding - and it might be flawed - is that all controllers should terminate because the context passed to them is canceled upon losing a leader election. So the idea for me was that re-spawning all controllers is exactly what we want when winning a leader election.

However, I haven't gotten around to actually test this yet. So I could be wrong here.

pkg/server/server.go
embik: This is a good point - any thoughts on what should happen in that case? We could let the kcp process crash so it gets restarted or put this in a loop, but I'm anxious to just let it keep crashing again and again if there is a bigger problem.

embik: I have chosen to implement a continuous loop around the leader election, so that if a kcp instance loses the leader role it will restart `RunOrDie` (until the overarching context is canceled).

pkg/server/server.go
embik: I moved this to `s.startControllers`.

embik:
/retest

I suspect it's a flake this time.

pkg/server/controllers.go
```diff
@@ -84,8 +83,54 @@ import (
...
+
+	// controllers can define their own custom wait functions in case
+	// they need to start early. If they do not define one, we will wait
+	// for everything to sync.
+	if controller.Wait != nil {
+		err = controller.Wait(ctx, s)
+	} else {
+		err = s.WaitForSync(ctx.Done())
+	}
+
```
sttts, line 116: no new line in front of `if err != nil`

pkg/server/controllers.go
```diff
@@ -84,8 +83,54 @@ import (
...
+}
+
+func (s *Server) startControllers(ctx context.Context) {
+	for _, controller := range s.controllers {
+		go s.runController(ctx, controller)
+	}
+}
+
+func (s *Server) runController(ctx context.Context, controller *controllerWrapper) {
+	var err error
```
sttts, line 102: move down where it is used

pkg/server/server.go
```diff
@@ -532,3 +541,79 @@ func goContext(parent genericapiserver.PostStartHookContext) context.Context {
 func (s *Server) WaitForPhase1Finished() {
 	<-s.rootPhase1FinishedCh
 }
+
+func (s *Server) leaderElectAndStartControllers(ctx context.Context) {
+	logger := klog.FromContext(ctx)
+
+	leaderElectionConfig := rest.CopyConfig(s.GenericConfig.LoopbackClientConfig)
+	leaderElectionConfig = rest.AddUserAgent(leaderElectionConfig, "kcp-leader-election")
+	// TODO(embik): is there a better way to access / generate the system:admin workspace url?
```
sttts, line 550: after rebase we have a `LocalAdminCluster` constant.

Today we have [pkg/genericcontrolplane/localadmin/cluster.go](https://github.com/kcp-dev/kubernetes/blob/c790f4246d25fa4d2787d5ac815c863cc56cdc56/pkg/genericcontrolplane/localadmin/cluster.go#L20). Witht that `localadmin.Cluster.Path().Path()` gives you `/clusters/system:admin`.

pkg/server/server.go
```diff
@@ -532,3 +541,79 @@ func goContext(parent genericapiserver.PostStartHookContext) context.Context {
...
+	leaderElectionConfig.Host = fmt.Sprintf("%s/clusters/system:admin", leaderElectionConfig.Host)
+
+	hostname, err := os.Hostname()
```
sttts, line 553: isn't that usually constant in containers? Do we need a flag?

pkg/server/server.go
```diff
@@ -532,3 +541,79 @@ func goContext(parent genericapiserver.PostStartHookContext) context.Context {
...
+	if err != nil {
+		logger.Error(err, "failed to determine hostname")
```
sttts, line 555: any way to make this fatal, or at least bubble it up in the stack?

pkg/server/server.go
embik: I looked at how kube-controller-manager does it and it also uses `os.Hostname`: https://github.com/kubernetes/kubernetes/blob/46835f8792dfb4a17345e592d1325bf63bc054e4/cmd/kube-controller-manager/app/controllermanager.go#L256

I guess it would be nice for this to work outside of container environments as well (e.g. on the local system).

sttts: if it is a non-issue, all is good :-) just looked suspicious.

pkg/server/controllers.go
embik: fixed.

pkg/server/controllers.go
embik: fixed.

pkg/server/server.go
embik: Went with `localadmin.Cluster.Path().RequestPath()`, can replace it if the rebase lands before this PR. I've also moved setting up the `leaderElectionConfig` outside of this function to ensure I can properly join URLs and fail the post-start-hook if it doesn't work.

pkg/server/server.go
embik: I've moved this to before the goroutine and pass it as parameter now. The post-start-hook will bubble up this error now.

PR #2998, opened 2023-08-08
Update workspaces.md - clarify HomeBucket Workspace definition

xtreme-sameer-vohra:
## Summary
:book:  Fix typo in HomeBucket definition


sttts:
Actually, this `HomeBucket` logic is gone (if I remember right that we removed it, to be verified). We just forgot to remove the doc or update them.

PR #2988, opened 2023-06-22
:book:  Update release process doc

mjudeikis:
## Summary

Nit updates to release docs

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

PR #3010, opened 2023-09-21
[release-0.20] 📖  update release process doc

embik:
## Summary

I would like to cherry-pick #2988 for two reasons:

1. The release process for 0.20 as documented on the release branch is incorrect since 0.20 is the first release under community governance and using Prow for automation. Now we likely will never release a 0.20.1 and instead move to 0.21.0.
2. However, we also never published 0.20 documentation. The reason for that I believe is the GitHub Actions triggers not running when the `release-*` branch is first created. A push event to the branch would be necessary, which we could achieve with a simple cherry-pick PR like this one being merged.

Therefore I would like to merge at least one commit to `release-0.20`, so we get documentation deployed to docs.kcp.io. it might not be strictly necessary and the docs are probably outdated, but I think it would be nice to have the version at least available on docs.kcp.io, now that people are perhaps curious what project is going to join the CNCF here.

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/content/developers/publishing-a-new-kcp-release.md
+++ b/docs/content/developers/publishing-a-new-kcp-release.md
@@ -35,19 +35,19 @@ kcp has 2 go modules, and a unique tag is needed for each module every time we c
     git tag --sign --message "$TAG" "$TAG" "$REF"
     ```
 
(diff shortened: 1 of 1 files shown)


xrstf (approved):
/approve

embik:
/retest

c'mon!

PR #3013, opened 2023-09-26
:sparkles: Add --cors-allowed-origins flag to kcp-front-proxy

xrstf:
## Summary
The APIserver config already has a flag to configure allowed CORS origins, but the front-proxy does not use the APIServer options and so did not get this flag. This PR adds it to the proxy's options to make the flag available.

## Release Notes

```release-note
Add `--cors-allowed-origins` flag to kcp-front-proxy
```

### DIFF
--- a/pkg/proxy/options/options.go
+++ b/pkg/proxy/options/options.go
@@ -27,13 +27,14 @@ import (
 )
 
 type Options struct {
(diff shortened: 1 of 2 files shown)


sttts:
/approve

Leaving lgtm to @mjudeikis or @embik

embik (approved):

PR #3018, opened 2023-10-09
:sparkles:  Add CRDs to built-in types

mjudeikis:
## Summary

Attempt to add CRDs to built-in types so they could be claimed

## Related issue(s)

Fixes #

## Release Notes

```release-note
CRD resources are now part of Built-in types. 
```

### DIFF
--- a/pkg/crdpuller/discovery.go
+++ b/pkg/crdpuller/discovery.go
@@ -495,7 +495,10 @@ func (sc *SchemaConverter) VisitKind(k *proto.Kind) {
 }
 
 func (sc *SchemaConverter) VisitReference(r proto.Reference) {
-	reference := r.Reference()
+	reference := r.Reference() // recursive CRDs are not supported
+	if reference == "io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.JSONSchemaProps" {
+		return
+	}
 	if sc.visited.Has(reference) {
 		*sc.errors = append(*sc.errors, fmt.Errorf("recursive schema are not supported: %s", reference))
 		return
--- a/pkg/informer/informer_test.go
+++ b/pkg/informer/informer_test.go
@@ -17,7 +17,6 @@ limitations under the License.
 package informer
 
 import (
-	"reflect"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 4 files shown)


pkg/informer/informer_test.go
```diff
@@ -103,11 +104,6 @@ func TestBuiltInInformableTypes(t *testing.T) {
 	}
 
 	allKnownTypes := kcpscheme.Scheme.AllKnownTypes()
-
-	// CRDs are not included in the genericcontrolplane scheme (because they're part of the apiextensions apiserver),
-	// so we have to manually add them
-	allKnownTypes[schema.GroupVersionKind{Group: "apiextensions.k8s.io", Version: "v1", Kind: "CustomResourceDefinition"}] = reflect.TypeOf(struct{}{})
```
sttts, line 109: isn't that still the case?

pkg/crdpuller/discovery.go
```diff
@@ -495,7 +495,10 @@ func (sc *SchemaConverter) VisitKind(k *proto.Kind) {
 }
 
 func (sc *SchemaConverter) VisitReference(r proto.Reference) {
-	reference := r.Reference()
+	reference := r.Reference() // recursive CRDs are not supported
+	if reference == "io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.JSONSchemaProps" {
+		return
+	}
```
sttts, line 501: what's the consequence of this? Who wants to crdpull CRDs?

mjudeikis: need to test crd pulling case :/ tbh this was very quick PR. Needs more testing. And this feedback kelps to identify the areas for testing :)

sttts: Note that pulling **the** CRD type, not CRDs (= instances), is very meta 😄

mjudeikis: I think this does not impact CRD pulling. We already converts APIS into CRDs 
Tested crd-puller and it just worked. Might be some edge cases but not sure how to identify those without knowing what I'm looking for ;)

pkg/informer/informer_test.go
mjudeikis: I don't think so. We add this to schema bellow? No?

PR #3023, opened 2023-10-12
:bug: Fix frontproxy

mjudeikis:
## Summary

Fix frontproxy deployment 

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/proxy/options/authentication.go
+++ b/pkg/proxy/options/authentication.go
@@ -117,6 +117,7 @@ func (c *Authentication) ApplyTo(authenticationInfo *genericapiserver.Authentica
 			versionedInformers.Core().V1().Secrets().Lister(),
 			versionedInformers.Core().V1().ServiceAccounts().Lister(),
 		)
+		authenticatorConfig.SecretsWriter = tokenGetterClient.CoreV1().Secrets()
 	}
 
 	// Sets up a union Authenticator for all enabled auth methods


pkg/proxy/options/authentication.go
```diff
@@ -117,6 +117,7 @@ func (c *Authentication) ApplyTo(authenticationInfo *genericapiserver.Authentica
 			versionedInformers.Core().V1().Secrets().Lister(),
 			versionedInformers.Core().V1().ServiceAccounts().Lister(),
 		)
+		authenticatorConfig.SecretsWriter = tokenGetterClient.CoreV1().Secrets()
```
sttts, line 120: what does a writer here?

mjudeikis: It is used in LegacyServiceAccount validator. It just fails to start if this is not set and APIServer fails . 
I think its for SA management of secrets

mjudeikis:
```
"ts":1697038810878.5505,"caller":"dynamiccertificates/dynamic_serving_content.go:113","msg":"Loaded a new cert/key pair","v":2,"name":"serving-cert::/etc/kcp-front-proxy/tls/tls.crt::/etc/kcp-front-proxy/tls/tls.key"}
{"ts":1697038811067.4224,"caller":"dynamiccertificates/dynamic_cafile_content.go:119","msg":"Loaded a new CA Bundle and Verifier","v":2,"name":"client-ca-bundle::/etc/kcp-front-proxy/client-ca/tls.crt"}
{"ts":1697038811067.9343,"caller":"transport/cert_rotation.go:137","msg":"Starting client certificate rotation controller\n","v":3}
{"ts":1697038811068.258,"caller":"cli/run.go:74","msg":"command failed","err":"while creating legacy validator, err: SecretsWriter must be provided"}
```
API server fails to start with this. We missed this in rebase

sttts: Though cert rotation also worries me. Why in the frontproxy? Or in general: why is this part of the authentication plugin? Seems odd.

mjudeikis: Might be how the front proxy was worried together. At this point it's not even starting. But you say authentication plugin should not even be there?

mjudeikis: If this is the case, this means we have wrong api-server flavor config in frontproxy case. 
Maybe worth merging this for now as it blocks usage of main branch all together with front proxy and looking how to get rid of it as followup?

sttts: The auth plugin is expected to be there. We use it in the proxy. I just wonder what else ppl have put into the auth plugins and whether we want to inherit it.

sttts:
/lgtm
/approve

I think this shows that we have to rethink service account handling in the front-proxy. The front-proxy will never know all serviceaccounts in an informer, and it changing last-used annotations is also questionable. In sharded, potentially multi-region mode this needs more thought.

PR #3024, opened 2023-10-12
:seedling:  Add metrics-viewer 'battery'

mjudeikis:
## Summary

While hunting down OOM/go-routine issue found that I needed to setup metrics scrapping. For this I needed service account to access them. Where it is easily done, I found this uslefull for few reasons;
1. I did it multiple times myself so it kinda got to the place - its need to be in the batteries
2. When using TILT, it does not have a clear phases approach. It usually - You deploy everything and wait. So I ended up maintaining a separate job just to deploy this into KCP with dependency created:
`KCP -> Job -> Prometheus`

Having this tilt case becomes - "While deploying Prometheus, read this secret and configure scraping and removing job out of the picture"

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/config/root-phase0/metrics-cluster-role-binding.yaml
+++ b/config/root-phase0/metrics-cluster-role-binding.yaml
@@ -1,0 +1,16 @@
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: metrics-viewer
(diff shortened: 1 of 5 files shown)


sttts:
Sgtm.

embik (approved):
/approve

PR #3029, opened 2023-10-16
:seedling: Increase resource requests on Prow jobs but only run if code changed

embik:
## Summary

This PR increases the memory and cpu requests on Prow jobs for e2e tests to ensure that they get their own nodes with minimal interference. At the same time, it adds `run_if_changed` to ensure that we no longer run e2e tests on non-code changes.

In addition, I'm reducing the number of runs in `pull-kcp-test-e2e-multiple-runs` from 3 to 2. That should be "good enough" while reducing the runtime of the job by one third.

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -96,8 +96,9 @@ presubmits:
               cpu: 2
 
   - name: pull-kcp-test-e2e
-    always_run: true
     decorate: true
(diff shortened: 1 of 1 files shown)


xrstf (approved):
/approve

PR #3033, opened 2023-10-20
⚠️ Add `--enable-auth-write` flag and store shard admin token hash as hex-encoded

embik:
## Summary

This PR is part of improvements to allow running kcp on Kubernetes. As of now, `kcp start` always tries to write `admin.kubeconfig` and `.admin-token-store` to disk, which means that in the Helm chart we [need to provide a PVC to the kcp server Deployment](https://github.com/kcp-dev/helm-charts/blob/e818463611c3204ebb24ba26ddf7135085d8e55d/charts/kcp/templates/server-deployment.yaml#L1). Eventually, with multiple replicas, we would need a RWX volume and kcp processes would overwrite each other on those files.

Looking at how/where `admin.kubeconfig` is used, it seems to be mainly for admins to access their kcp installation. At least I could not find it being read into kcp itself anywhere. Which means we don't have to provide this in the Kubernetes setup. To have a consistent shard admin token, this PR still requires `--authentication-admin-token-path` to be passed with a pre-configured token hash file (including a sha256 hash). kcp reads that file and accepts the token hash without trying to write the `admin.kubeconfig`. 

Because the token hash file is of greater importance now, I looked at it and saw that it stores sha256 checksums in raw bytes. Which means looking at the file you get some gibberish. It's not clear if that was intentional or not, but IMHO this file should be human-readable and because of that, this PR introduces the **breaking change that the token file is now hex-encoded**. This means you can now pass the output of piping your token to `sha256sum` into a file and pass that to kcp.

## Related issue(s)

Fixes #

## Release Notes

```release-note
Add `--enable-auth-write` flag with can disable `admin.kubeconfig` and `.admin-token-store` being written to disk. The shard admin token hash is now stored as sha256 hex-encoded string
```

### DIFF
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -92,6 +92,7 @@ type ExtraConfig struct {
 	IdentityConfig    *rest.Config
 
 	// authentication
+	writeAdminKubeconfig                      bool
 	kcpAdminToken, shardAdminToken, userToken string
 	shardAdminTokenHash                       []byte
 
@@ -317,7 +318,7 @@ func NewConfig(opts kcpserveroptions.CompletedOptions) (*Config, error) {
 		return nil, err
 	}
 	var userToken string
-	c.kcpAdminToken, c.shardAdminToken, userToken, c.shardAdminTokenHash, err = opts.AdminAuthentication.ApplyTo(c.GenericConfig)
+	c.kcpAdminToken, c.shardAdminToken, userToken, c.shardAdminTokenHash, c.writeAdminKubeconfig, err = opts.AdminAuthentication.ApplyTo(c.GenericConfig)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/server/options/authentication.go
+++ b/pkg/server/options/authentication.go
@@ -20,9 +20,11 @@ import (
 	"bytes"
 	"context"
 	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/google/uuid"
 	"github.com/spf13/pflag"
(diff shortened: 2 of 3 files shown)


sttts:
Not sure this is the right direction. Instead just use the authentication mechanisms available. There is a token file auth plugin, and of course client certs. Either would be able to provide a cross-shard admin. The admin created in-place on startup is really just for development. Don't put it on a PV at all. Just ignore it.

embik:
> Not sure this is the right direction. Instead just use the authentication mechanisms available. There is a token file auth plugin, and of course client certs. Either would be able to provide a cross-shard admin. The admin created in-place on startup is really just for development. Don't put it on a PV at all. Just ignore it.

That's a fair point, I forgot about the token auth plugin. Would you still be open to a flag enabling/disabling writing kubeconfig and admin token hash file (completely)? Ideally, at some point, I would like to make the root filesystem read-only, and if we can avoid adding tokens to the authenticator that we won't use I would also sleep slightly better at night.

sttts:
Sure, disabling admin token generation is fine.

sttts:
Is the admin user a battery?

sttts:
```
      --batteries-included strings
                A list of batteries included (= default objects that might be unwanted in production, but are very helpful in trying out kcp or for development). These are the possible values:
                user,workspace-types.

                - workspace-types:         creates "organization" and "team" WorkspaceTypes in the root workspace.
                - root-compute-workspace:  create a root:compute workspace, and kubernetes APIExport in it for deployments/services/ingresses
                - user:                    creates an additional non-admin user and context named "user" in the admin.kubeconfig
```
We have `user` already. That's very similar. Would call the battery `admin`. Wdyt?

mjudeikis:
sgtm `admin` battery

embik:
Sounds good to me too, I will work on implementing it as battery instead. Thanks for the input folks!

PR #3036, opened 2023-10-21
:seedling:  Add owner aliases to API approvers

mjudeikis:
## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


sttts:
Would keep this very group smaller, not default to the whole set of maintainers. It is ok if not every maintainer is listed here. It's a special role to carefully review every comma in API definitions.

PR #3041, opened 2023-11-03
:sparkles: Add `admin` battery to kcp server and enable it by default

embik:
## Summary

This is a follow-up to #3033 and the discussion it created. This new approach adds a new battery called `admin` and enables it by default. Without this battery, `admin.kubeconfig` will not be generated and no admin token will be created to be injected into a new authenticator.

I've also added a check to make sure that the `user` battery is not enabled when `admin` is not.

## Related issue(s)

Fixes #

## Release Notes

```release-note
Add new `admin` battery which is enabled by default
```


embik:
I guess this might not be a flake.

PR #3035, opened 2023-10-21
:seedling:  Make LogicalCluster claimable

mjudeikis:
## Summary

Attempt to make LogicalCluster native resource so it can be used in claims. 

## Related issue(s)

Fixes #

## Release Notes

```release-note
Make LogicalCluster claimable resource for deeper level integration ontop of KCP
```

PR #3042, opened 2023-11-13
:bug: Be aware of sharding in APIBinding tests for LogicalCluster

embik:
## Summary

#3035 added some new tests for binding `LogicalClusters`. The test works fine when running "normal" e2e tests, but `test-e2e-sharded-minimal` fails more or less all the time. The reason is that the test was accessing the VirtualWorkspace URL at index 0, but sharded testing makes the situation more complex (with two VW servers). The APIExport and APIBinding workspaces were always added to `shard-1`, and then another virtual workspace is presenting the data we are looking for.

This adjusts the tests to consider multiple VWs and gather objects across VWs for the two new tests.

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/cmd/sharded-test-server/cache.go
+++ b/cmd/sharded-test-server/cache.go
@@ -100,7 +100,7 @@ func startCacheServer(ctx context.Context, logDirPath, workingDir string, synthe
 		case err := <-terminatedCh:
 			var exitErr *exec.ExitError
 			if err == nil {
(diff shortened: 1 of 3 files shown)


embik:
/retest

Another flake (?), different test this time:

```
=== FAIL: test/e2e/reconciler/workspacedeletion TestWorkspaceDeletion/nested_worksapce_cleanup_when_an_org_workspace_is_deleted (54.23s)
```

PR #3045, opened 2023-11-20
:sparkles:  Add Workspace Mount API

mjudeikis:
## Summary

At some point we talked that it would be nice to have SymLinks , like bellow. Where anybody interacting with `root:w1:ww2` would be interacting with `root:w2:ww2` 
```
root 
    - w1
       - ww1
       - ww2 ->Symlinked to `ww2`
   - w2
      - ww1
      - ww2
```

This is raw API proposal for it. For now it would be in annotations as json struct until we agree this is ok to add into Workspace spec. 

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


sdk/apis/tenancy/v1alpha1/types_mounts.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Mount is a workspace mount that can be used to mount a workspace into another workspace or resource.
+// Mounting itself is done at front proxy level.
+type Mount struct {
+	// Type is the type of the mount.
+	Kind MountKind `json:"kind,omitempty"`
+	// Source is an ObjectReference to the object that is mounted.
+	Source *corev1.ObjectReference `json:"ref,omitempty"`
+	// Path is the path where the object is mounted in relation to the workspace root.
+	Path string `json:"path,omitempty"`
```
sttts, line 40: why is this necessary?

sdk/apis/tenancy/v1alpha1/types_mounts.go
```diff
@@ -0,0 +1,59 @@
...
+
+const (
+	SymLink MountKind = "symlink"
+)
+
+// Mount is a workspace mount that can be used to mount a workspace into another workspace or resource.
+// Mounting itself is done at front proxy level.
+type Mount struct {
+	// Type is the type of the mount.
+	Kind MountKind `json:"kind,omitempty"`
```
sttts, line 36: which other kinds do you expect?

sdk/apis/tenancy/v1alpha1/types_mounts.go
```diff
@@ -0,0 +1,59 @@
+/*
+Copyright 2021 The KCP Authors.
```
sttts, line 2: 2023

sdk/apis/tenancy/v1alpha1/types_mounts.go
mjudeikis: How would we plug this into frontProxy?  In the similar way you derive from WorkspaceType, maybe something similar to inheritance is needed here to try to avoid hardcoding multiple types into frontProxy. So thinking that either inheritance needed to be some basic first type in tenancy API? 

If we don't keep any implementation details in the `Mount type, how would one BYO type?

sttts: The path is defined by the current parent workspace plus `:<workspace-name>`.

sdk/apis/tenancy/v1alpha1/types_mounts.go
```diff
@@ -0,0 +1,72 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+
+	corev1alpha1 "github.com/kcp-dev/kcp/sdk/apis/core/v1alpha1"
+	conditionsv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/third_party/conditions/apis/conditions/v1alpha1"
+)
+
+// Mount is a workspace mount that can be used to mount a workspace into another workspace or resource.
+// Mounting itself is done at front proxy level.
+type Mount struct {
```
mjudeikis, line 30: @sttts  I think I was not clear on the intention of this. I think this iteration represents a better view:
0. Source is used to trigger reconcile for external components. 
1. Name is used to map to status field (unique per Workspace)
2. Status is a placeholder for status to be put by an external component.

sttts: you mean that one resource (e.g. KubeCluster) could provider multiple mounts in multiple workspace? Thinking about one `mycluster` workspace and one `mycluster-sudo` workspace.

sttts: and of course we could default that value to `"default"`.

sdk/apis/tenancy/v1alpha1/types_workspace.go
```diff
@@ -195,6 +200,11 @@ type WorkspaceStatus struct {
 	//
 	// +optional
 	Initializers []corev1alpha1.LogicalClusterInitializer `json:"initializers,omitempty"`
+
+	// Mounts is a list of mounts that are mounted into this workspace and their statuses.
+	//
+	// +optional
+	Mounts []MountStatus `json:"mounts,omitempty"`
```
sttts, line 207: this is already the next step, isn't it? If we have mounts here, we don't need the annotation anymore?

sdk/apis/tenancy/v1alpha1/types_mounts.go
mjudeikis: Its possible, hecne list of mounts and statuses
 Lets say "sudo" could be different mountpoint, mountpoints per namespaces" where we mount only 1 namespace could be mountpoint,

sdk/apis/tenancy/v1alpha1/types_mounts.go
```diff
@@ -0,0 +1,72 @@
...
+	// Name is the name of the mount.
+	Name string `json:"name,omitempty"`
+	// Source is an ObjectReference to the object that is mounted.
+	Source *corev1.ObjectReference `json:"ref,omitempty"`
+}
+
+// MountStatus is the status of a mount. It is used to indicate the status of a mount,
+// potentially managed outside of the core API.
+type MountStatus struct {
```
sttts, line 39: this is for the `Workspace`, not for the `KubeCluster` (or whatever provider is used)?

mjudeikis:
Will move to differnet media

PR #3046, opened 2023-11-22
🐛 Fix container image tags for tagged kcp releases

xrstf:
## Summary
For every commit to a release branch, we want to update the `release-X.Y` tag on GHCR. Likewise, when a `vX.Y.Z` is tagged, we also want to update `release-X.Y`.

However my original code assumed that we could always deduce the branch name from the tag. Since we only ran the script (before 0.21 was released) for untagged commits, this worked, because those have `PULL_BASE_REF=<branchname>` as their env variables. But for tagged releases, the `PULL_BASE_REF` is the tag and there is no actual branch in the Prow pod (technically, the tag becomes the branch). So it's simply not possible to get the correct branch. Additionally, a tag might be set on a commit that is included in multiple branches -- which would win?

To work around this, I adjusted the script to _construct_ the branch name, assuming that we will always follow the `release-MAJOR-MINOR` syntax for branches.

Fixes #3030

## Release Notes
```release-note
NONE
```

### DIFF
--- a/hack/build-image.sh
+++ b/hack/build-image.sh
@@ -42,14 +42,21 @@ architectures="amd64 arm64 ppc64le"
 
 # when building locally, just tag with the current HEAD hash
 version="$(git rev-parse --short HEAD)"
+branchName=""
 
 # deduce the tag from the Prow job metadata
 if [ -n "${PULL_BASE_REF:-}" ]; then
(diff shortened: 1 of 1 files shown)


embik (approved):
/approve

PR #3055, opened 2023-12-08
:memo: Enable OpenAPI validation for Virtual Workspace

LokeshvarKr:
## Enable OpenAPI V3 validation for Virtual Workspace to avoid using the --validate=false argument.


mjudeikis:
@LokeshvarKr Thanks for this.
Can you add right title, release notes as per PR template and sign DCO so CI is happy?

PR #3057, opened 2023-12-09
:sparkles: Experimental mounts API

mjudeikis:
## Summary

Adds experimental mounts API to be able to mount workspaces under the hood to different destinations. 

## Related issue(s)

Fixes #

## Release Notes

```release-note
Add experimental mounts API 
```


pkg/server/localproxy.go
```diff
@@ -133,27 +136,47 @@ func WithLocalProxy(
...
+		if r.URL != "" {
+			url, err := url.Parse(r.URL)
+			if err != nil {
+				logger.WithValues("cluster", cluster.Name, "url", r.URL).Error(err, "invalid url")
+				http.Error(w, "invalid url", http.StatusInternalServerError)
+				return
+			}
+			logger.WithValues("from", path, "to", r.URL).V(4).Info("mounting cluster")
+			proxy := httputil.NewSingleHostReverseProxy(url)
+			// TODO(mjudeikis): remove this once we have a real cert wired in dev mode
```
embik, line 157: Where would this cert come from / what is missing to add it?

embik (commented):
Some minor Qs.

pkg/index/index.go
```diff
@@ -54,6 +53,7 @@ func New(rewriters []PathRewriter) *State {
 		shardWorkspaceName:        map[string]map[logicalcluster.Name]string{},
 		shardClusterParentCluster: map[string]map[logicalcluster.Name]logicalcluster.Name{},
 		shardBaseURLs:             map[string]string{},
+		workspaceNameMount:        map[logicalcluster.Name]map[string]string{},
```
sttts, line 56: what are the two strings?

pkg/index/index.go
```diff
@@ -78,17 +80,25 @@ func (c *State) UpsertWorkspace(shard string, ws *tenancyv1alpha1.Workspace) {
 	clusterName := logicalcluster.From(ws)
 
 	c.lock.RLock()
-	got := c.shardWorkspaceNameCluster[shard][clusterName][ws.Name]
+	cluster := c.shardWorkspaceNameCluster[shard][clusterName][ws.Name]
+	mount := c.workspaceNameMount[clusterName][ws.Name] // experimental feature
```
sttts, line 84: what is a "mount"? A URL? A name?

pkg/server/localproxy.go
```diff
@@ -133,27 +136,47 @@ func WithLocalProxy(
...
+			logger.WithValues("from", path, "to", r.URL).V(4).Info("mounting cluster")
+			proxy := httputil.NewSingleHostReverseProxy(url)
+			//TODO(MVP): remove this once we have a real cert wired in dev mode
+			proxy.Transport = &http.Transport{
+				TLSClientConfig: &tls.Config{
+					InsecureSkipVerify: true,
+				},
+			}
+
+			proxy.ServeHTTP(w, req)
```
sttts, line 164: Does the front-proxy need the same?

pkg/index/index.go
mjudeikis: `Cluster, workspace, serialized-to-string-mount-point-object`

pkg/index/index.go
mjudeikis: At this point, I'm thinking maybe we should store just a URL, not a full object. Good question.

pkg/server/localproxy.go
mjudeikis: I think front proxy handles this by default as It return the url in `Result`. Will double check.

pkg/index/index.go
sttts: clusterWorkspaceMountAnnotation ?

pkg/index/index.go
sttts: Sounds fine. Maybe call it mountURL ?

pkg/server/localproxy.go
sttts: right, might be. It's proxying every request.

mjudeikis: yes, and it already can hook certificates for custom workspace paths.

pkg/index/index.go
mjudeikis: hmmm. It's a bit more complicated when I read it again. I stored full serialized objects so I could do "poor man DeepCopy". Meaning it would update this in the cache if anything changes in the mount objects. And idea behind this is that we will need to make a workspace not ready if a mountpoint becomes not ready. 

And due to this, I stored all serialized objects. As alternative is to store some aggregate of status.

Plus as we store in serialized to the string in the annotations, saves unmarshaling to objects here.

PR #3064, opened 2023-12-19
:sparkles: Add 'metrics-viewer' user to ClusterRoleBinding in `metrics-viewer` battery

embik:
## Summary

I've been thinking about how to get metrics out of kcp for the Helm chart. The `metrics-viewer` battery is a great starting point, but requires you to extract the service account token out of kcp, which seems hard to do.

This PR adds another entity to the `ClusterRoleBinding` subjects: A `metrics-viewer` user. In the Helm chart, we can create a certificate for that user entity and then pass that certificate as credentials to e.g. a `ServiceMonitor` for [prometheus-operator](https://prometheus-operator.dev/). That seems to be the smoothest course of action to me.

## Related issue(s)

Fixes #

## Release Notes

```release-note
Add a `metrics-viewer` user subject to the ClusterRoleBinding created by the battery `metrics-viewer`, for which credentials can be generated outside of kcp
```

### DIFF
--- a/config/root/metrics-3-cluster-role-binding.yaml
+++ b/config/root/metrics-3-cluster-role-binding.yaml
@@ -9,6 +9,9 @@ subjects:
   name: metrics
   namespace: default
   apiGroup: ""
+- kind: User
+  name: metrics-viewer
+  apiGroup: ""
 roleRef:
   kind: ClusterRole
   name: metrics-viewer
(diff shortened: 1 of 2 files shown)


mjudeikis:
/lgtm
/approve
nice one :) I was missing this part :D

PR #3070, opened 2024-01-03
🐛 Generate `admin.kubeconfig` with correct cluster for `system:admin` context

embik:
## Summary

A long while ago, `/` redirected to the `system:admin` workspace. That is no longer the case, and as such the `admin.kubeconfig` generated by kcp should have a separate cluster/server from the `base` one, so that the `system:admin` context in it is functional.

I'm also adding a `shard-base` context because the combination of "base" server and "shard-admin" credentials are required in a lot of testing and sharding.

I also found this outdated information in the docs, so replaced it with some more up-to-date information.

## Related issue(s)

Fixes #

## Release Notes

```release-note
Fix `system:admin` context and add `system:base` in generated `admin.kubeconfig`
```

### DIFF
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -235,14 +235,14 @@ func NewConfig(opts kcpserveroptions.CompletedOptions) (*Config, error) {
 	// The informers here are not used before the informers are actually started (i.e. no race).
 	if len(c.Options.Extra.RootShardKubeconfigFile) > 0 {
 		// TODO(p0lyn0mial): use kcp-admin instead of system:admin
-		nonIdentityRootKcpShardSystemAdminConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(&clientcmd.ClientConfigLoadingRules{ExplicitPath: c.Options.Extra.RootShardKubeconfigFile}, &clientcmd.ConfigOverrides{CurrentContext: "system:admin"}).ClientConfig()
+		nonIdentityRootKcpShardBaseConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(&clientcmd.ClientConfigLoadingRules{ExplicitPath: c.Options.Extra.RootShardKubeconfigFile}, &clientcmd.ConfigOverrides{CurrentContext: "shard-base"}).ClientConfig()
 		if err != nil {
 			return nil, fmt.Errorf("failed to load the kubeconfig from: %s, for the root shard, err: %w", c.Options.Extra.RootShardKubeconfigFile, err)
 		}
 
 		var kcpShardIdentityRoundTripper func(rt http.RoundTripper) http.RoundTripper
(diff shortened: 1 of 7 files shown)


embik:
I guess this is breaking e2e tests, I will need to check why.

sttts:
Lgtm. Any idea why e2e break?

embik:
> Lgtm. Any idea why e2e break?

Looking into it. I can replicate them on my local system, so I assume it's not a flake.

embik:
Should work now. The e2e framework needed the shard-admin credentials and was "utilising" the fact that the `system:admin` context was using `/` as server, so now I need to fetch the base URL from the `base` context and combine the two.

embik:
@sttts I had to rework the PR a bit. Turns out sharding is looking for the `system:admin` context in kubeconfigs and started failing because of that.

I've resolved this problem by creating a new `system:base` context that is not usable from `kubectl` (just as the `base` one), but that is now used by sharding and testing code. PTAL.

pkg/server/config.go
```diff
@@ -234,8 +234,8 @@ func NewConfig(opts kcpserveroptions.CompletedOptions) (*Config, error) {
 	// hence we postpone this to getOrCreateKcpIdentities() in the kcp-start-informers post-start hook.
 	// The informers here are not used before the informers are actually started (i.e. no race).
 	if len(c.Options.Extra.RootShardKubeconfigFile) > 0 {
-		// TODO(p0lyn0mial): use kcp-admin instead of system:admin
-		nonIdentityRootKcpShardSystemAdminConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(&clientcmd.ClientConfigLoadingRules{ExplicitPath: c.Options.Extra.RootShardKubeconfigFile}, &clientcmd.ConfigOverrides{CurrentContext: "system:admin"}).ClientConfig()
+		// TODO(p0lyn0mial): use kcp-admin instead of system:base
+		nonIdentityRootKcpShardSystemAdminConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(&clientcmd.ClientConfigLoadingRules{ExplicitPath: c.Options.Extra.RootShardKubeconfigFile}, &clientcmd.ConfigOverrides{CurrentContext: "system:base"}).ClientConfig()
```
sttts, line 238: this var name is strange now

pkg/server/options/authentication.go
```diff
@@ -198,11 +198,17 @@ func createKubeConfig(kcpAdminToken, shardAdminToken, userToken, baseHost, tlsSe
...
+			CertificateAuthorityData: caData,
+			TLSServerName:            tlsServerName,
+		},
 	}
 	kubeConfig.Contexts = map[string]*clientcmdapi.Context{
 		"root":         {Cluster: "root", AuthInfo: kcpAdminUserName},
 		"base":         {Cluster: "base", AuthInfo: kcpAdminUserName},
-		"system:admin": {Cluster: "base", AuthInfo: shardAdminUserName},
+		"system:admin": {Cluster: "system:admin", AuthInfo: shardAdminUserName},
+		"system:base":  {Cluster: "base", AuthInfo: shardAdminUserName},
```
sttts, line 211: the colon notation is confusing IMO. shard-base or shard-admin-base?

embik: I like `shard-base`. Will adjust.

pkg/server/config.go
embik: Updated to `nonIdentityRootKcpShardBaseConfig`.

Issue #3011, opened 2023-09-25
bug: `kubectl-ws '~'` changes hostname used in kubeconfig server URL


MikeSpreitzer:
### Describe the bug

I have seen several people accidentally issue the command `kubectl ws` (with nothing else on the command line), and then be confused from then on.  This happened to me a few times too, I only tracked this down with some effort.

```
mspreitz@mjs12 kubestellar % cp k2.kubeconfig k3.kubeconfig
mspreitz@mjs12 kubestellar % KUBECONFIG=k3.kubeconfig kubectl ws
Current workspace is "kvdk2spgmbix".
Note: 'kubectl ws' now matches 'cd' semantics: go to home workspace. 'kubectl ws -' to go back. 'kubectl ws .' to print current workspace.
mspreitz@mjs12 kubestellar % diff k2.kubeconfig k3.kubeconfig | grep server
<       server: https://yep.yep:6443
<       server: https://yep.yep:6443/clusters/root
<       server: https://yep.yep:6443/clusters/root
<       server: https://yep.yep:6443/clusters/root:compute
>     server: https://yep.yep:6443
>     server: https://yep.yep:6443/clusters/root
>     server: https://10.244.0.9:6443/clusters/kvdk2spgmbix
>     server: https://yep.yep:6443/clusters/root
mspreitz@mjs12 kubestellar % 
```

This was using release v0.11.0 of kcp.

### Steps To Reproduce

1. Make a kubeconfig that uses a domain name for each "cluster" in the kubeconfig relevant to kcp.
2. `kubectl workspace` (nothing else on the command line)
3. Inspect those clusters in that kubeconfig, find that the one named "workspace.kcp.io/current" uses the server's local IP:port in the server URL.

### Expected Behaviour

When all the kcp-relevant clusters use a particular host:port, the `kubectl workspace` plugin should not switch to a different host:port.

embik:
Does the `kcp` process you are running this against set `--external-hostname=yep.yep:6443` as flag? From a cursory review of the plugin code, it seems to pull the workspace URL directly from what kcp advertises the workspace status, which sounds more reliable to me than parsing the existing file and guess what might be right URL.

MikeSpreitzer:
No, I have _not_ given the server `--external-hostname`.  I need a way to support two sets of clients: ones that are close to the server and contact it via a short network path, and those that are farther away and come in through a heavier longer network path.

mjudeikis:
[...]

PR #3085, opened 2024-03-07
:sparkles: Index mounting ordering & few debug nits

mjudeikis:
## Summary

This PR does few things:
1. Fixes usage of HOME workspace where it does not honour host fixes: https://github.com/kcp-dev/kcp/issues/3011
2. A few debug logs for cache and CRD creation (way too many hours spend on those...)
3. Add weight into index routes:
```
// 1. We detect URL for the request and find the correct handler. URL can be
// shard based, virtual workspace or mount. First two are covered by r.URL,
// where mounts are covered by annotation on the workspace with the mount path.
// 2. If mountpoint is found, we rewrite the URL to resolve, else use one in
// request to match with mappings.
// 3. Iterate over mappings and find the one that matches the URL. If found,
// use the handler for that mapping, else use default handler - kcp.
// Mappings are done from most specific to least specific:
// Example: /clusters/cluster1/ will be matched before /clusters/
```

In general, it dynamically overrides mounts URLs (this was not done before), and matches them in order of length. Higher the weight (`weight = length of the path`) first it will get matches.

Without this, map is randomly selected and if you have 2 mappings with
```
clusters/foo
clusters/
```
your traffic will be randomized :)

## Related issue(s)

Fixes #

## Release Notes

```release-note
Orders index path from longest with weight. 
```

### DIFF
--- a/pkg/proxy/mapping.go
+++ b/pkg/proxy/mapping.go
@@ -26,13 +26,14 @@ import (
 	"path"
 	"strings"
 
+	"github.com/kcp-dev/logicalcluster/v3"
+
 	"k8s.io/component-base/metrics/legacyregistry"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/yaml"
 
 	"github.com/kcp-dev/kcp/pkg/proxy/index"
 	proxyoptions "github.com/kcp-dev/kcp/pkg/proxy/options"
-	"github.com/kcp-dev/logicalcluster/v3"
 )
 
 // PathMapping describes how to route traffic from a path to a backend server.
--- a/pkg/proxy/mapping.go
+++ b/pkg/proxy/mapping.go
@@ -23,6 +23,8 @@ import (
 	"net/http/httputil"
 	"net/url"
 	"os"
+	"path"
+	"strings"
 
 	"k8s.io/component-base/metrics/legacyregistry"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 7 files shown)


pkg/cliplugins/workspace/plugin/kubeconfig.go
```diff
@@ -215,7 +215,20 @@ func (o *UseWorkspaceOptions) Run(ctx context.Context) error {
...
+		u, _, err := pluginhelpers.ParseClusterURL(config.Host)
+		if err != nil {
+			return fmt.Errorf("current URL %q does not point to a workspace", config.Host)
+		}
+		uh, err := url.Parse(homeWorkspace.Spec.URL)
+		if err != nil {
+			return fmt.Errorf("invalid home workspace URL %q: %w", homeWorkspace.Spec.URL, err)
+		}
+
+		newServerHost = u.Scheme + "://" + path.Join(u.Host, uh.Path)
```
sttts, line 231: maybe worth a comment what we are doing here, namely: keep the old host, but append the path of the homeworkspace URL.

pkg/proxy/mapping.go
```diff
@@ -102,8 +174,34 @@ func NewHandler(ctx context.Context, o *proxyoptions.Options, index index.Index)
...
+}
+
+func sortMappings(mappings []httpHandlerMapping) []httpHandlerMapping {
+	// sort mappings by weight
+	// higher weight means that the mapping is more specific and should be matched first
+	// Example: /clusters/cluster1/ will be matched before /clusters/
+	for i := range mappings {
+		for j := range mappings {
+			if mappings[i].weight > mappings[j].weight {
+				mappings[i], mappings[j] = mappings[j], mappings[i]
```
sttts, line 202: bubble sort 🎉

mjudeikis: One and only! Why invest something new when this just does the job :D

PR #3092, opened 2024-03-15
:sparkles: Protect APIBindings created during workspace initialization

palnabarun:
## Summary

The APIBindings created during workspace initialization should not be updated or deleted by users as no controller reconciles those. Once they are updated or deleted, no lifecycle happens for those APIBindings.

This creates inconsistencies in the system as there is no guarantee that APIBindings listed in `WorkspaceType.spec.defaultAPIBindings` would be honored across the lifecycle of the workspace.

## Release Notes

```release-note
✨ Protect APIBindings created during Workspace Initialization 
```


embik:
I feel like this is missing something, or is there magic around `internal.kcp.io/*` annotations that I am not aware of?

PR #3101, opened 2024-03-27
:book: Update documentation with CNCF community group

embik:
## Summary

We were notified on Slack that our docs still refer to the community calendar, which we no longer maintain. This replaces the reference with our CNCF community group instead. I've also made the docs index and repository readme consistent.

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,14 +34,16 @@ This community has a [Code of Conduct](./code-of-conduct.md). Please make sure t
 
 There are several ways to communicate with us:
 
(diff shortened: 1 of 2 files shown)


embik:
> We'll probably want to cherry-pick #3086 first

The change in question is in v0.23.0 already, but we haven't updated our documentation link on kcp.io to the docs for that version. Will also do that (partially included in https://github.com/kcp-dev/kcp.io/pull/35 already).

PR #3102, opened 2024-03-27
:book: Deploy most recent release documentation as 'latest' alias

embik:
## Summary

I noticed that on https://docs.kcp.io, the "default" release (the release you get when going to https://docs.kcp.io/kcp/) points to 0.11. The reason for that is that the "default" release was set at _some_ point manually, but it was never automated.

This PR:
- tries to figure out if the pipeline is running for the latest released version series (e.g. v0.23 at the moment)
- if so, updates a version alias "latest" to the version while deploying
- ensures that the default version points to the "latest" alias

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.github/workflows/docs-gen-and-push.yaml
+++ b/.github/workflows/docs-gen-and-push.yaml
@@ -30,6 +30,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - run: git fetch origin gh-pages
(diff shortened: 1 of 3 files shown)


embik:
This should be ready for review now.

PR #3106, opened 2024-03-31
:sparkles:  split go releaser archives

mjudeikis:
## Summary

Release file copies, and not softlink'ed one. Go release does not know how to handle those

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -8,8 +8,6 @@ on:
   push:
     tags:
     - 'v*'
-    - 'cli/v*'
-    - 'sdk/v*'
 
 permissions:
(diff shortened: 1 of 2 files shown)


sttts:
> Why don't we reduce or plugins to a single one

That breaks down with `kubectl create workspace`, xref https://github.com/kcp-dev/kcp/pull/3154.

From all the commands `kubectl ws` is what I use by far most often. IMO it deserves to be its own command.

sttts:
This stops the bleeding.

Follow-up https://github.com/kcp-dev/kcp/pull/3154.

/lgtm
/approve

PR #3111, opened 2024-04-04
:sparkles:  Fix leader election issue with workspace controller and other KCP Controllers

sankar17:
This PR addresses the following, 

Change the workspace controllers start logic inside runners to fix leader election issue.
The way we register controllers and define the runner is problematic, the runner calls start only. but in case leader election is lost start finishes (as it was waiting on <- ctx.Done()) which leads to the defer on the queue.Shutdown() to run. Once you shutdown a queue, there’s no way to restart it

**Background:**
    At times we faced workspace controller creation stuck at scheduling phase and never recovers. Regarding leader election the requests/events queued to both leader and other pods aswell , this makes the queue depth to grow.

### DIFF
--- a/pkg/server/controllers.go
+++ b/pkg/server/controllers.go
@@ -29,13 +29,19 @@ import (
 	kcpkubernetesclientset "github.com/kcp-dev/client-go/kubernetes"
 	kcpmetadata "github.com/kcp-dev/client-go/metadata"
 	"github.com/kcp-dev/logicalcluster/v3"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	admissionregistrationv1alpha1 "k8s.io/api/admissionregistration/v1alpha1"
 
 	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 	certutil "k8s.io/client-go/util/cert"
 	"k8s.io/client-go/util/keyutil"
 	"k8s.io/klog/v2"
@@ -48,6 +54,7 @@ import (
 	configuniversal "github.com/kcp-dev/kcp/config/universal"
 	bootstrappolicy "github.com/kcp-dev/kcp/pkg/authorization/bootstrap"
 	kcpfeatures "github.com/kcp-dev/kcp/pkg/features"
+	"github.com/kcp-dev/kcp/pkg/indexers"
 	"github.com/kcp-dev/kcp/pkg/informer"
 	"github.com/kcp-dev/kcp/pkg/reconciler/apis/apibinding"
 	"github.com/kcp-dev/kcp/pkg/reconciler/apis/apibindingdeletion"
@@ -60,6 +67,7 @@ import (
 	apisreplicateclusterrole "github.com/kcp-dev/kcp/pkg/reconciler/apis/replicateclusterrole"
 	apisreplicateclusterrolebinding "github.com/kcp-dev/kcp/pkg/reconciler/apis/replicateclusterrolebinding"
 	apisreplicatelogicalcluster "github.com/kcp-dev/kcp/pkg/reconciler/apis/replicatelogicalcluster"
+	"github.com/kcp-dev/kcp/pkg/reconciler/cache/labelclusterroles"
 	"github.com/kcp-dev/kcp/pkg/reconciler/cache/replication"
 	logicalclusterctrl "github.com/kcp-dev/kcp/pkg/reconciler/core/logicalcluster"
 	"github.com/kcp-dev/kcp/pkg/reconciler/core/logicalclusterdeletion"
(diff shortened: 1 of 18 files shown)


pkg/server/controllers.go
```diff
@@ -410,24 +431,33 @@ func (s *Server) installWorkspaceScheduler(ctx context.Context, config *rest.Con
...
+		byBase36Sha224Name: indexByBase36Sha224Name,
+	})
+	indexers.AddIfNotPresentOrDie(s.CacheKcpSharedInformerFactory.Tenancy().V1alpha1().WorkspaceTypes().Informer().GetIndexer(), cache.Indexers{
+		indexers.ByLogicalClusterPathAndName: indexers.IndexByLogicalClusterPathAndName,
+	})
 
 	if err := s.registerController(&controllerWrapper{
 		Name: workspace.ControllerName,
 		Runner: func(ctx context.Context) {
+			workspaceController, err := workspace.NewController(
```
yastij, line 447: this issue happens with every controller, we need to to do it for the rest of the controllers

embik: After this PR popped up I _was_ wondering why it would only affect specific controllers, this seemed like a fairly fundamental flaw in how I implemented leader election.

@sankar17 is this something that you want to refactor at a bigger scale to address the architectural problem or would you like me to take a look?

sankar17: @embik This has to be fixed for all the controllers, initially we got stuck with workspace controller creation stopped working with our longevity tests / over night test with 200 workspace creation / deletion . We saw the kcp metrics and the request queued to all 3/3 pods, though the leader election is enabled.

Currently I am working on changing this logic for all the controllers and need to do more tests.

if you think this is going to be fixed architectural level , please take a look at it.

CC: @yastij @palnabarun

embik: Okay, taking a quick look at where queues are being started I think the approach in this PR is overall fine. @sankar17 feel free to carry on, we should change it for all controllers. I can also take over if you don't have time.

sankar17: @embik I did changes to all the controllers, currently testing the fix, if all good I will update this PR in few days

embik:
@sankar17 Please consider not running re-tests when tests are failing consistently, at least not without any code changes pushed. Those tests burn CI cycles without any real reason, we already know that they don't work.

sankar17:
> @sankar17 Please consider not running re-tests when tests are failing consistently, at least not without any code changes pushed. Those tests burn CI cycles without any real reason, we already know that they don't work.

Sure I will make sure it works in local and do retest

embik:
Thanks!

pkg/server/controllers.go
```diff
@@ -600,21 +682,20 @@ func (s *Server) installAPIBindingController(ctx context.Context, config *rest.C
...
+
+	// APIExport indexers
+	indexers.AddIfNotPresentOrDie(s.KcpSharedInformerFactory.Apis().V1alpha1().APIExports().Informer().GetIndexer(), cache.Indexers{
+		indexers.ByLogicalClusterPathAndName: indexers.IndexByLogicalClusterPathAndName,
+		indexAPIExportsByAPIResourceSchema:   indexAPIExportsByAPIResourceSchemasFunc,
+	})
+	indexers.AddIfNotPresentOrDie(s.CacheKcpSharedInformerFactory.Apis().V1alpha1().APIExports().Informer().GetIndexer(), cache.Indexers{
+		indexers.ByLogicalClusterPathAndName: indexers.IndexByLogicalClusterPathAndName,
+		indexAPIExportsByAPIResourceSchema:   indexAPIExportsByAPIResourceSchemasFunc,
+	})
```
embik, line 698: What is the reason for moving the indexer setup out of the controller creation to here (this happens in other parts of the code, so this is more of a placeholder reference for all those changes)?

palnabarun: The core reason we moved registration of indexers outside the `controller.NewFooController` is because with this change, all of the `controller.NewFooController` calls are inside the `Runner` function which is called everytime a KCP pod becomes a leader.

`indexers.AddIfNotPresentOrDie` is supposed to be run only once in the lifecycle of a KCP pod because otherwise it will panic.

The mental model is that:
- installFooController outside of Runner has steps that can or needed to run one-time only in the lifecycle of the KCP pod
- Runner has all the steps that can be run multiple times in the runtime of a Pod.

Note: The same KCP pod can somehow lose leader-election and get elected again in due course of time. The changes in this PR enables that scenario.

palnabarun: @embik this is a little bit complicated to wrap at one go. Maybe we can explain over the community call later.

embik: @palnabarun I first understood `indexers.AddIfNotPresentOrDie` to be idem-potent since it filters out already present indexers (that's why I asked), but my feeling is the problem isn't _that_ part of the logic, it's that calling `indexer.AddIndexers` even with an empty list of indexers (because `indexers.AddIfNotPresentOrDie` filtered out all existing indexers) will panic when the store is already started -- do I get this right? Or is the problem _that_ it's called with an empty list of indexers (because they all already exist)?

I won't be on the next community call unfortunately, but I don't mind anyone else reviewing this and discussing it there either if that works better.

embik (changes requested):
Overall this PR looks good to me but it needs:

1. Squashed commits.
2. DCO sign-off on all commits.
3. a release note in the PR description.

pkg/server/controllers.go
```diff
@@ -80,12 +88,24 @@ import (
...
 )
 
+const (
+	byBase36Sha224Name                      = "byBase36Sha224Name"
+	unschedulable                           = "unschedulable"
+	indexAPIExportsByAPIResourceSchema      = "apiExportsByAPIResourceSchema"
+	indexAPIExportEndpointSliceByAPIExport  = "indexAPIExportEndpointSliceByAPIExport"
+	indexAPIExportEndpointSlicesByPartition = "indexAPIExportEndpointSlicesByPartition"
+)
+
```
sttts, line 108: can we qualify these more? indexWhatByWhat

pkg/server/controllers.go
```diff
@@ -95,6 +115,69 @@ type controllerWrapper struct {
...
+			// this will never happen due to validation
+			return []string{}, fmt.Errorf("cluster information missing")
+		}
+		key := client.ToClusterAwareKey(clusterName, slice.Spec.Partition)
+		return []string{key}, nil
+	}
+
+	return []string{}, nil
+}
+
```
sttts, line 180: would move all of these into an index.go. This file here is getting big.

pkg/server/controllers.go
```diff
@@ -345,15 +428,14 @@ func (s *Server) installTenancyLogicalClusterController(ctx context.Context, con
...
-	)
-
 	return s.registerController(&controllerWrapper{
 		Name: tenancylogicalcluster.ControllerName,
 		Runner: func(ctx context.Context) {
+			controller := tenancylogicalcluster.NewController(
+				kubeClusterClient,
+				s.KcpSharedInformerFactory.Core().V1alpha1().LogicalClusters(),
+				s.KubeSharedInformerFactory.Rbac().V1().ClusterRoleBindings(),
+			)
```
sttts, line 438: 🔴  What is the semantics here now in the Runner? Do we ensure the informer factor is started after all the runners? Looking at the code this can't be the case: `controller.Start` blocks. This means we have a race of the individual informers not being started if the factory is faster than the runners.

pkg/server/controllers.go
sankar17: Sure , we wil do

pkg/server/controllers.go
sankar17: In few controllers the wait for informer cache to sync logic is implemented using wait method before start.Can we just add this for all the controllers for which informers to be sycned before start ?

https://github.com/sankar17/kcp/blob/sankar17/rest-cfg-timeout-and-workspace-ctrl-leader-election-fix/pkg/server/controllers.go#L719-L732

https://github.com/sankar17/kcp/blob/sankar17/rest-cfg-timeout-and-workspace-ctrl-leader-election-fix/pkg/server/controllers.go#L912-L918

https://github.com/sankar17/kcp/blob/sankar17/rest-cfg-timeout-and-workspace-ctrl-leader-election-fix/pkg/server/controllers.go#L991-L997

https://github.com/sankar17/kcp/blob/sankar17/rest-cfg-timeout-and-workspace-ctrl-leader-election-fix/pkg/server/controllers.go#L1217-L1225

https://github.com/sankar17/kcp/blob/sankar17/rest-cfg-timeout-and-workspace-ctrl-leader-election-fix/pkg/server/controllers.go#L1254-L1261

https://github.com/sankar17/kcp/blob/sankar17/rest-cfg-timeout-and-workspace-ctrl-leader-election-fix/pkg/server/controllers.go#L1467-L1478


CC: @yastij @palnabarun

sankar17:
This is implemented with alternative approach https://github.com/kcp-dev/kcp/pull/3132
, hence this PR is no longer needed

PR #3112, opened 2024-04-04
:seedling: Set controller rest config timeout to 30secs

sankar17:
Configure the controller rest config timeout to 30 seconds by default to avoid any client stuck at infinitely.

### DIFF
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -26,6 +26,7 @@ import (
 	_ "net/http/pprof"
 	"net/url"
 	"os"
+	"time"
 
 	kcpapiextensionsclientset "github.com/kcp-dev/client-go/apiextensions/client"
 	kcpapiextensionsinformers "github.com/kcp-dev/client-go/apiextensions/informers"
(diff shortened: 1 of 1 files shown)


palnabarun:
@sankar17 can you please add a release note in the PR description? Check the PR body template for the structure.

pkg/server/server.go
```diff
@@ -387,7 +387,8 @@ func (s *Server) Run(ctx context.Context) error {
 	// TODO: split apart everything after this line, into their own commands, optional launched in this process
 
 	controllerConfig := rest.CopyConfig(s.IdentityConfig)
-
+	// Setting the default restConfig timeout to 30 seconds to avoid any client is waiting infinitely
+	controllerConfig.Timeout = time.Second * 30
```
sttts, line 391: Wondering, we should probably set that much earlier. IdentityConfig comes the loopback config if I am not mistaken. So maybe already the loopback config needs this timeout?

embik:
@sankar17 please re-add the release notes block that is in the PR description template and add a meaningful release note. Thanks!

PR #3118, opened 2024-04-18
:sparkles: Implement cluster-aware OpenAPI v3

sttts:
## Summary

The OpenAPI v3 stack was not yet cluster-aware. This PR adds
1. a controller generating and caching OpenAPI v3 specs for all CRDs, including bound CRD for APIBindings
2. a filter `WithOpenAPIv3` that merges these specs on-demand keeping a cache for different API configurations.

It makes heavy use of the lazy evaluation to avoid computation where possible.

Currently, (2) is implemented with a LRU cache of size 100 holding the (lazily) merged specs for API configurations. This means that with more than a 100 different configurations and constant OpenAPI v3 requests, this will not work well. Either we increase the number of different API configurations in the cache, or we use another kind of cache. But it's probably not common that there are so many different API configurations because some workspace types will be reused a lot.

## Release Notes

```release-note
Add support for /openapi/v3 endpoints for workspaces with awareness of static resources, CRDs and APIBindings.
```

### DIFF
--- a/pkg/server/openapiv3/controller.go
+++ b/pkg/server/openapiv3/controller.go
@@ -1,0 +1,260 @@
+/*
+Copyright 2024 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package openapiv3
+
+import (
+	"context"
+	"crypto/sha512"
+	"encoding/json"
+	"fmt"
+	"sync"
+	"time"
+
+	kcpcache "github.com/kcp-dev/apimachinery/v2/pkg/cache"
+	kcpapiextensionsv1informers "github.com/kcp-dev/client-go/apiextensions/informers/apiextensions/v1"
+	kcpapiextensionsv1listers "github.com/kcp-dev/client-go/apiextensions/listers/apiextensions/v1"
+	"github.com/kcp-dev/logicalcluster/v3"
+
+	apiextensionshelpers "k8s.io/apiextensions-apiserver/pkg/apihelpers"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apiextensions-apiserver/pkg/controller/openapi/builder"
+	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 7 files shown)


pkg/server/openapiv3/controller.go
```diff
@@ -0,0 +1,260 @@
...
+		}
+		return false, nil
+	}
+
+	c.updateCRD(crd)
+
+	return false, nil
+}
+
+func (c *Controller) updateCRD(crd *apiextensionsv1.CustomResourceDefinition) {
```
mjudeikis, line 182: updateCRD where? As I understand this is just cache update method? 
maybe small comment on the function?

pkg/server/openapiv3/controller.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// add new instance with all versions
+	for _, v := range crd.Spec.Versions {
+		if !v.Served {
+			delete(c.byClusterNameVersion[clusterName][crd.Name], v.Name)
+			continue
+		}
+
+		spec := cached.NewStaticSource[*spec3.OpenAPI](func() cached.Result[*spec3.OpenAPI] {
+			spec, err := builder.BuildOpenAPIV3(crd, v.Name, builder.Options{V2: false})
```
mjudeikis, line 196: Just trying to learn k8s machienry code,  what does this do?

pkg/server/openapiv3/servicecache.go
```diff
@@ -0,0 +1,298 @@
...
+
+	// get both real CRDs and bound CRD from APIBindings
+	crds, err := c.crdLister.Cluster(clusterName).List(ctx, labels.Everything())
+	if err != nil {
+		responsewriters.InternalError(w, r, err)
+		return
+	}
+
+	// operate on sorted lists to have deterministic API configuration key
+	orderedCRDs := make([]*apiextensionsv1.CustomResourceDefinition, len(crds))
```
mjudeikis, line 137: Would this make sense to move into apiConfiguration? 
Now `apiConfiguration` expects a sorted list. If at any point in the probable future, somebody want to reuse this function, this is not explicitly clear (function comment might be good too)?

mjudeikis: Ah, line 142 would need to go in. Code becomes offuscated. I think explicit code comment on function would be enough.

mjudeikis (commented):
Looks good. Want to test "in prod"? 
Do you think it worth a feature-gate so we can test in isolation and turn it off if needed?

pkg/server/openapiv3/controller.go
sttts: renamed to processCRD

pkg/server/openapiv3/controller.go
sttts: build a full spec from the openapi v3 schema in the CRD, i.e. add everything like http paths, parameters etc.

pkg/server/openapiv3/servicecache.go
sttts: have renamed the argument to orderedCRDs. Should make it clearer.

PR #3128, opened 2024-05-08
:book: Redirect aliases instead of symlinking to their targets

embik:
## Summary

One last minor follow up to documentation improvements. We now have https://docs.kcp.io/kcp/latest/ and it points to v0.24, but I think it would make more sense if it redirected to https://docs.kcp.io/kcp/v0.24/ over just showing the same content. That way it's clearer what version you are actually looking at and links are more permanent (a link to a page in `/latest/` might no longer work when we update the documentation).

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -80,6 +80,9 @@ plugins:
           name: English
       reconfigure_material: true
       reconfigure_search: true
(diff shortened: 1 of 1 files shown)


xrstf (approved):
/approve

PR #3129, opened 2024-05-10
:sparkles: apiexports: make SubjectAccessReview and LocalSubjectAccessReview claimable

sttts:
## Summary

The list of native resources that are claimable is static and defined at https://github.com/kcp-dev/kcp/blob/main/pkg/virtual/apiexport/schemas/builtin/builtin.go#L88.

This PR adds `SubjectAccessReview` and `LocalSubjectAccessReview` to that list.

## Release Notes

```release-note
Allow claiming `SubjectAccessReview` and `LocalSubjectAccessReview` in apiexports.
```

PR #3131, opened 2024-05-13
:bug: apibindings/reconciler/labeller: fix informer error on non-persisted resources

sttts:
## Summary

Since https://github.com/kcp-dev/kcp/pull/3129, both `SubjectAccessReview` and `LocalSubjectAccessReview` can be claimed. But as these are not persisted, there is no need for labelling of objects. It's not even possible to create such an informer. Hence, skip these resources.

## Release Notes

```release-note
NONE
```


sttts:
Workspace deletion flake, unrelated.

PR #3132, opened 2024-05-16
✨ Add installControllers and uninstallControllers for leaderElectionStopping

ramramu3433:
## Summary

At times we faced workspace controller creation stuck at scheduling phase and never recovers, we found that RCA of the issue is that , when the controllers are started using leaderElection, if the leader is lost once, the Queue processing is never restored ,because workqueue depth increases

* initialize informers before starting the controllers and informers ,and stop controllers when the pod is no longer a leader 
* reinitialize when the pod is elected as leader 
* move the openapi controller to a seperate post hook to avoid stopping when the pod is lost leader election. 

## Related issue(s)

Fixes #

## Release Notes

```release-note
 Fix sequencing of controllers/informers start and leader election 
```

### DIFF
--- a/pkg/server/controllers.go
+++ b/pkg/server/controllers.go
@@ -78,7 +78,6 @@ import (
 	"github.com/kcp-dev/kcp/pkg/reconciler/tenancy/workspacemounts"
 	"github.com/kcp-dev/kcp/pkg/reconciler/tenancy/workspacetype"
 	"github.com/kcp-dev/kcp/pkg/reconciler/topology/partitionset"
-	"github.com/kcp-dev/kcp/pkg/server/openapiv3"
 	initializingworkspacesbuilder "github.com/kcp-dev/kcp/pkg/virtual/initializingworkspaces/builder"
 	corev1alpha1 "github.com/kcp-dev/kcp/sdk/apis/core/v1alpha1"
 	tenancyv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/tenancy/v1alpha1"
@@ -88,6 +87,10 @@ import (
 
 type RunFunc func(ctx context.Context)
 type WaitFunc func(ctx context.Context, s *Server) error
+
+const (
+	waitPollInterval = time.Millisecond * 100
+)
 
 type controllerWrapper struct {
 	Name   string
@@ -146,19 +149,13 @@ func (s *Server) installClusterRoleAggregationController(ctx context.Context, co
 
 	return s.registerController(&controllerWrapper{
 		Name: controllerName,
+		Wait: func(ctx context.Context, s *Server) error {
+			return wait.PollUntilContextCancel(ctx, waitPollInterval, true, func(ctx context.Context) (bool, error) {
+				return s.KubeSharedInformerFactory.Rbac().V1().ClusterRoles().Informer().HasSynced(), nil
+			})
+		},
 		Runner: func(ctx context.Context) {
 			c.Run(ctx, 5)
-		},
-	})
-}
-
-func (s *Server) installOpenAPIv3Controller(ctx context.Context, config *rest.Config) error {
-	controllerName := openapiv3.ControllerName
-
-	return s.registerController(&controllerWrapper{
-		Name: controllerName,
-		Runner: func(ctx context.Context) {
-			s.openAPIv3Controller.Run(ctx)
 		},
 	})
 }
@@ -201,6 +198,11 @@ func (s *Server) installKubeNamespaceController(ctx context.Context, config *res
 
 	return s.registerController(&controllerWrapper{
 		Name: controllerName,
+		Wait: func(ctx context.Context, s *Server) error {
+			return wait.PollUntilContextCancel(ctx, waitPollInterval, true, func(ctx context.Context) (bool, error) {
+				return s.KubeSharedInformerFactory.Core().V1().Namespaces().Informer().HasSynced(), nil
+			})
+		},
 		Runner: func(ctx context.Context) {
 			c.Run(ctx, 10)
 		},
@@ -227,6 +229,12 @@ func (s *Server) installKubeServiceAccountController(ctx context.Context, config
 
 	return s.registerController(&controllerWrapper{
 		Name: controllerName,
+		Wait: func(ctx context.Context, s *Server) error {
+			return wait.PollUntilContextCancel(ctx, waitPollInterval, true, func(ctx context.Context) (bool, error) {
+				return s.KubeSharedInformerFactory.Core().V1().ServiceAccounts().Informer().HasSynced() &&
+					s.KubeSharedInformerFactory.Core().V1().Namespaces().Informer().HasSynced(), nil
+			})
+		},
 		Runner: func(ctx context.Context) {
 			c.Run(ctx, 1)
 		},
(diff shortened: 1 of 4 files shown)


embik:
/ok-to-test

Please restore the [PR description format](https://github.com/kcp-dev/kcp/blob/main/.github/pull_request_template.md) and give a release note!

pkg/server/server.go
```diff
@@ -391,156 +550,14 @@ func (s *Server) Run(ctx context.Context) error {
...
-	}
-
-	if err := s.installRootCAConfigMapController(ctx, s.Apis.GenericAPIServer.LoopbackClientConfig); err != nil {
-		return err
-	}
-
-	if err := s.installApiExportIdentityController(ctx, controllerConfig); err != nil {
+	// Adding the bootup time controllers and informers initialization
+	s.controllerLock.Lock()
+	defer s.controllerLock.Unlock()
```
embik, line 557: Isn't `(s *Server) Run` blocking due to the ` s.MiniAggregator.GenericAPIServer.PrepareRun().Run(ctx.Done())` call at the end? How would the leader election be able to take this lock when it's already taken by the main server method for its whole lifetime?

pkg/server/server.go
```diff
@@ -583,6 +598,19 @@ func (s *Server) Run(ctx context.Context) error {
 		return err
 	}
 
+	// add post start hook that starts all registered controllers.
+	if err := s.AddPostStartHook("kcp-start-openapiv3-server", func(srvcontext genericapiserver.PostStartHookContext) error {
```
embik, line 602: Comment needs updating and I would maybe call this hook `kcp-start-openapiv3-controller`. I don't think it's called the openapiv3-server anywhere.

embik:
Thank you for the contribution @ramramu3433! I've left some comments, overall this looks like a very promising approach to resolving the issue at hand.

pkg/server/server.go
```diff
@@ -178,6 +180,163 @@ func NewServer(c CompletedConfig) (*Server, error) {
 	return s, nil
 }
 
+/* Deregister the controllers when the leader election is lost. */
+func (s *Server) uninstallControllers(ctx context.Context) {
```
embik, line 184: what is the `ctx context.Context` for?

pkg/server/server.go
```diff
@@ -649,9 +677,19 @@ loop:
 				Callbacks: leaderelection.LeaderCallbacks{
 					OnStartedLeading: func(leaderElectionCtx context.Context) {
 						electionLogger.Info("started leading")
+						if len(s.controllers) == 0 {
+							s.controllerLock.Lock()
+							defer s.controllerLock.Unlock()
```
palnabarun, line 682: IIUC the installation of controllers happen only in one thread. Do we need to acquire a lock for that?

pkg/server/server.go
ramramu3433: Moved the unlock after installController to not block startControllers

pkg/server/server.go
ramramu3433: This was added , so there would be no race happens when the uninstallController and installController

pkg/server/server.go
```diff
@@ -649,9 +677,20 @@ loop:
...
 					OnStartedLeading: func(leaderElectionCtx context.Context) {
 						electionLogger.Info("started leading")
+						if len(s.controllers) == 0 {
+							s.controllerLock.Lock()
+							if err = s.installControllers(ctx, config); err != nil {
+								logger.Error(err, "error in re-registering controllers")
+								return
+							}
+							s.controllerLock.Unlock()
+						}
```
palnabarun, line 687: Is there a possibility that this and/or `uninstallControllers` is running in multiple Goroutines? I am wondering why we need the mutex. What am I missing? I feel they would run in the same Goroutine.

pkg/server/controllers.go
```diff
@@ -866,7 +966,9 @@ func (s *Server) installAPIExportController(ctx context.Context, config *rest.Co
 			return wait.PollUntilContextCancel(ctx, time.Millisecond*100, true, func(ctx context.Context) (bool, error) {
 				crdsSynced := s.ApiExtensionsSharedInformerFactory.Apiextensions().V1().CustomResourceDefinitions().Informer().HasSynced()
 				exportsSynced := s.KcpSharedInformerFactory.Apis().V1alpha1().APIExports().Informer().HasSynced()
-				return crdsSynced && exportsSynced, nil
+				namespaceSynced := s.KubeSharedInformerFactory.Core().V1().Namespaces().Informer().HasSynced()
+				secretSynced := s.KubeSharedInformerFactory.Core().V1().Secrets().Informer().HasSynced()
+				return crdsSynced && exportsSynced && namespaceSynced && secretSynced, nil
```
sttts, line 971: would unify the style.

pkg/server/controllers.go
```diff
@@ -1198,6 +1381,11 @@ func (s *Server) installApiExportIdentityController(ctx context.Context, config
 
 	return s.registerController(&controllerWrapper{
 		Name: identitycache.ControllerName,
+		Wait: func(ctx context.Context, s *Server) error {
+			return wait.PollUntilContextCancel(ctx, time.Millisecond*100, true, func(ctx context.Context) (bool, error) {
```
sttts, line 1385: can we turn the interval into a constant?

pkg/server/server.go
```diff
@@ -649,9 +677,20 @@ loop:
 				Callbacks: leaderelection.LeaderCallbacks{
 					OnStartedLeading: func(leaderElectionCtx context.Context) {
 						electionLogger.Info("started leading")
+						if len(s.controllers) == 0 {
+							s.controllerLock.Lock()
+							if err = s.installControllers(ctx, config); err != nil {
+								logger.Error(err, "error in re-registering controllers")
+								return
+							}
+							s.controllerLock.Unlock()
```
sttts, line 686: we have to unlock before the return too.

sttts: Maybe even put into a closure with defer.

pkg/server/server.go
```diff
@@ -178,6 +180,163 @@ func NewServer(c CompletedConfig) (*Server, error) {
 	return s, nil
 }
 
+/* Deregister the controllers when the leader election is lost. */
+func (s *Server) uninstallControllers() {
```
sttts, line 184: these run under the lock. `uninstallControllersLocked()` is typical style to make that clear.

pkg/server/server.go
```diff
@@ -391,156 +550,14 @@ func (s *Server) Run(ctx context.Context) error {
...
-		return err
-	}
-
-	if err := s.installRootCAConfigMapController(ctx, s.Apis.GenericAPIServer.LoopbackClientConfig); err != nil {
-		return err
-	}
-
-	if err := s.installApiExportIdentityController(ctx, controllerConfig); err != nil {
+	// Adding the bootup time controllers and informers initialization
+	s.controllerLock.Lock()
```
sttts, line 556: any reason why we should move the locking into install/uninstallControllers? Looks like the lock is only needed there.

pkg/server/server.go
sttts: +1, very probably it is one go routine. We could then get rid of the locking?

pkg/server/server.go
ramramu3433: removed the lock as suggested above

pkg/server/server.go
ramramu3433: removed locking as suggested

pkg/server/server.go
ramramu3433: removed mutex locks

embik (changes requested):
Code looks good to me now, but please update the PR description before this can be approved and lgtm'd: https://github.com/kcp-dev/kcp/pull/3132#issuecomment-2115520181.

embik (approved):
/lgtm

Will leave final approval to others who have reviewed this.

palnabarun (approved):
/lgtm

PR #3133, opened 2024-05-16
✨ cli: write diagnostics in `kubectl-workspace` to stderr

embik:
## Summary

While working on https://github.com/embik/kubectl-switch-ws I noticed that `kubectl-workspace use` will always print diagnostic information (deprecation notices and warnings about unfinished API bindings), even when using `--short` for inclusion in scripts / terminal prompts. There is no good way to avoid these warnings for scripts that only expect the workspace name returned as output.

I would like to propose that we start writing this supplementary information to stderr, which is specifically for diagnostic information and errors. This way, scripts can avoid the warnings by using something like `kubectl workspace use --short <workspace> 2>/dev/null` to make sure you only get back the workspace name (or an exit code that isn't zero, which signals that something bad happened).

There's also an old `OWNERS` file left here which I'd propose to remove.

## Related issue(s)

Fixes #

## Release Notes

```release-note
Write diagnostics (deprecation notices and warnings) in `kubectl-workspace` to stderr instead of stdout
```

### DIFF
--- a/cli/pkg/OWNERS
+++ b/cli/pkg/OWNERS
@@ -1,6 +1,0 @@
-approvers:
-- ncdc
-- sttts
-reviewers:
-- stevekuznetsov
-- davidfestal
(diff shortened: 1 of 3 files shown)


mjudeikis:
Where I dont quite like abusing stderr but I dont see better way without having bunch of
`if short; dont print` and this sound more of the evil compared to what you propose.
/lgtm
/approve
/hold

PR #3139, opened 2024-06-10
:bug:  Fix race issues in leader relection

ramramu3433:
## Summary

When the controller startup sequence is fixed with https://github.com/kcp-dev/kcp/pull/3132 , there is a race condition introduced which would throw panic as the controller reintialization adds indexer to running informers

## Related issue(s)

Fixes #

## Release Notes

```release-note
Calls intialize indexer only once before the informer starts
```

### DIFF
--- a/pkg/permissionclaim/permissionclaim_labeler.go
+++ b/pkg/permissionclaim/permissionclaim_labeler.go
@@ -60,10 +60,6 @@ func NewLabeler(
 	apiBindingInformer apisv1alpha1informers.APIBindingClusterInformer,
 	apiExportInformer, globalAPIExportInformer apisv1alpha1informers.APIExportClusterInformer,
 ) *Labeler {
-	indexers.AddIfNotPresentOrDie(apiExportInformer.Informer().GetIndexer(), cache.Indexers{
-		indexers.ByLogicalClusterPathAndName: indexers.IndexByLogicalClusterPathAndName,
-	})
-
 	return &Labeler{
 		listAPIBindingsAcceptingClaimedGroupResource: func(clusterName logicalcluster.Name, groupResource schema.GroupResource) ([]*apisv1alpha1.APIBinding, error) {
 			indexKey := indexers.ClusterAndGroupResourceValue(clusterName, groupResource)
@@ -147,4 +143,11 @@ func (l *Labeler) LabelsFor(ctx context.Context, cluster logicalcluster.Name, gr
 	}
 
 	return labels, nil
+}
+
+// InstallIndexers adds the additional indexers that this controller requires to the informers.
+func InstallIndexers(apiExportInformer apisv1alpha1informers.APIExportClusterInformer) {
+	indexers.AddIfNotPresentOrDie(apiExportInformer.Informer().GetIndexer(), cache.Indexers{
+		indexers.ByLogicalClusterPathAndName: indexers.IndexByLogicalClusterPathAndName,
+	})
 }
--- a/pkg/server/controllers.go
+++ b/pkg/server/controllers.go
@@ -32,6 +32,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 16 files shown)


pkg/server/controllers.go
```diff
@@ -1478,3 +1488,188 @@ func (s *Server) WaitForSync(stop <-chan struct{}) error {
...
+	indexers.AddIfNotPresentOrDie(s.CacheKcpSharedInformerFactory.Core().V1alpha1().Shards().Informer().GetIndexer(), cache.Indexers{
+		workspace.ByBase36Sha224Name: workspace.IndexByBase36Sha224Name,
+	})
+	indexers.AddIfNotPresentOrDie(s.CacheKcpSharedInformerFactory.Tenancy().V1alpha1().WorkspaceTypes().Informer().GetIndexer(), cache.Indexers{
+		indexers.ByLogicalClusterPathAndName: indexers.IndexByLogicalClusterPathAndName,
+	})
+	// APIBinding indexers
+	indexers.AddIfNotPresentOrDie(s.KcpSharedInformerFactory.Apis().V1alpha1().APIBindings().Informer().GetIndexer(), cache.Indexers{
+		indexers.APIBindingsByAPIExport: indexers.IndexAPIBindingByAPIExport,
+	})
```
embik, line 1506: I'm concerned about this being very hard to trace back to the respective controller that need this. Is there any chance we can re-organise the code into the respective controllers (where possible) and call that from here?

embik (changes requested):

pkg/server/controllers.go
ramramu3433: @embik probably we can have this in each controllers , but have a condition in installControllers , like bootstrapIndexers if true , we would activate this codepath , if not we would skip it , as this is already bootstrapped? , or probably thinking of getting indexers before calling indexersAddition

embik: If each controller package had a `InstallIndexers` func to call in `addIndexersToInformers`, I think that would already be sufficient.

ramramu3433: Addressed the comment as mentioned above @embik , PTAL

pkg/permissionclaim/permissionclaim_labeler.go
```diff
@@ -148,3 +144,13 @@ func (l *Labeler) LabelsFor(ctx context.Context, cluster logicalcluster.Name, gr
 
 	return labels, nil
 }
+
+/*
+installIndexers method is added, as we are re-electing the controllers
+as the informers are always running, we cannot call this in multiple places.
+*/
```
embik, line 151: Nit: These comments are formatted in a very atypical way for Go. Can you please update them (here and in the other occurrences of `InstallIIndexers`) to have `//` in front of each line and follow the style of _"FunctionName does this or that"_, so it would be something like

> InstallIndexers adds the additional indexers that this controller requires to the informers.

embik (changes requested):
Thanks for the changes! One nit about the comments, apart from that it is looking good to me now.

pkg/permissionclaim/permissionclaim_labeler.go
ramramu3433: Done

pkg/server/controllers.go
```diff
@@ -1356,13 +1360,6 @@ func (s *Server) installKubeQuotaController(
 		Runner: func(ctx context.Context) {
 			c.Start(ctx, 2)
 		},
-	}); err != nil {
-		return err
-	}
-
-	return s.AddPreShutdownHook(kubequota.ControllerName, func() error {
-		close(s.quotaAdmissionStopCh)
-		return nil
```
embik, line 1365: why has this been removed?

embik (changes requested):

pkg/server/controllers.go
embik: @ramramu3433 gentle ping, everything else looks great in this PR.

embik (approved):
/approve

PR #3143, opened 2024-07-30
:seedling: Update go-jose to v3

embik:
## Summary

`gopkg.in/square/go-jose.v2` is archived, there are no updates to it anymore, and GHSA-c5q2-7r4c-mv6g has been opened for this dependency. While Kubernetes isn't affected according to https://github.com/kubernetes/kubernetes/issues/123252, I thought it'd be nice to bump this as v3 seems to work the same way and received a fix for this issue.

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


embik:
Maybe it doesn't make sense to deviate from upstream here, as @xrstf rightfully pointed out. I'm at the moment not sure if we are affected by this anyway, I will take a closer look at the security issue.

/hold

xrstf:
See also https://github.com/kubernetes/kubernetes/pull/123253#issuecomment-1940379993 for why upstream has not updated the dependency [yet?].

PR #3140, opened 2024-07-15
:sparkles: Update to Kubernetes 1.30

embik:
## Summary

This updates kcp to Kubernetes 1.30.0. The corresponding branch for our Kubernetes fork is [kcp-1.30](https://github.com/kcp-dev/kubernetes/tree/kcp-1.30). Mainly this PR adds new APIs or removes feature gates where they have been removed by upstream.

The bigger changes in this PR are:

- Updating to the new ValidatingAdmissionPolicy plugin infrastructure.
- Rewrite the OpenAPI controller we recently added to accommodate for new upstream types.

The following bigger TODOs exist, but I would like to tackle them separately:

- Look at enabling the flags I've disabled in this PR, most notably `--authentication-config`, which seems like a hugely useful feature for a multi-tenant API server like KCP.

## Related issue(s)

Fixes #3068

## Release Notes

```release-note
Update to Kubernetes 1.30.0
```

PR #3146, opened 2024-07-31
🌱  Remove dependency on `cli` module and bump dependencies in `cli` and `sdk`

embik:
## Summary

Since kcp 0.24.0, the new `cli` package is creating a bit of a problem because its a dependency of `github.com/kcp-dev/kcp`, but most users would not import it. Thus, the version from `go.mod` is used for dependency resolution, which is an invalid one: https://github.com/kcp-dev/kcp/blob/f4297bda4dbd1c82883c68eda57af67f34e32984/go.mod#L17

To avoid having a `replace` directive in every project that imports `github.com/kcp-dev/kcp`, I've moved the small API surface used in `github.com/kcp-dev/kcp` (the main module) to `github.com/kcp-dev/kcp/sdk`, a module that both modules depend on anyway.

This still means that to build a project from `github.com/kcp-dev/kcp`, you must have `github.com/kcp-dev/kcp/sdk` as a dependency as well, but that seems to be the more likely case.

In addition, this PR bumps the `cli` and `sdk` modules to be in sync with the changes done to the main module in #3140.

After that, I _also_ ran into the issue that we are using a KCP `apigen` from v0.21.0 (so quite an old one), so I've changed that logic to use the `apigen` that is present in the repository right now. It doesn't make much sense to me to keep using this from previous releases.

## Related issue(s)

Fixes #

## Release Notes

```release-note
Remove the need to put a `replace` directive in place for `github.com/kcp-dev/kcp/cli` when importing `github.com/kcp-dev/kcp`
```


mjudeikis:
/lgtm
looks leggit :)

embik:
/retest

looks like a flake.

PR #3149, opened 2024-08-05
:seedling: Delete OWNERS files to let current maintainers handle all packages

embik:
## Summary

One last PR then I'm done for a bit. I've noticed that we still have _a lot_ of `OWNERS` files that list subsets of the old development team. I'd propose we delete all of these to make clear that people don't have those responsibilities anymore. I've only left those `OWNERS` files in place that reference `api-approvers`, because those have been updated in #3036. All the others should go in my opinion (😢 ).

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


xrstf:
/unhold

whoops, too late 😁

If you want your ownership back, please ping @embik :)

embik:
@mjudeikis @sttts for your awareness that this PR was approved by accident. If you think we should revert the change, please let me know.

sttts:
Looked through the changes. Either the ppl are not active anymore, or they are in the root OWNER file. So it looks like this PR makes the owner fit reality.

PR #3152, opened 2024-08-16
:sparkles: Add inactive annotation to logicalcluster

RedbackThomson:
## Summary
In order to do maintenance on a workspace, without controllers reconciling the resources inside the window during the maintenance period, I need a mechanism by which to disable access to the workspace. This feature adds a new optional annotation to `LogicalCluster` which disables API access through to the resource while present:

```
internal.kcp.io/inactive: "true"
```

All requests to the logical cluster while the annotation is set are denied with a 400 error, except for requests to CRUD the logical cluster itself or the OpenAPI spec (so that you can reasonably disable the annotation).

## Related issue(s)

None

## Release Notes

```release-note
Add support for inactive annotation on logical clusters
```

### DIFF
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -445,6 +445,7 @@ func NewConfig(opts kcpserveroptions.CompletedOptions) (*Config, error) {
 		apiHandler = filters.WithWarningRecorder(apiHandler)
 
 		apiHandler = kcpfilters.WithAuditEventClusterAnnotation(apiHandler)
+		apiHandler = kcpfilters.WithBlockInactiveLogicalClusters(apiHandler, c.KcpSharedInformerFactory.Core().V1alpha1().LogicalClusters())
 
 		// Add a mux before the chain, for other handlers with their own handler chain to hook in. For example, when
 		// the virtual workspace server is running as part of kcp, it registers /services with the mux so it can handle
--- a/pkg/server/filters/filters.go
+++ b/pkg/server/filters/filters.go
@@ -18,6 +18,7 @@ package filters
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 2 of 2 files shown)


pkg/server/config.go
```diff
@@ -445,6 +445,7 @@ func NewConfig(opts kcpserveroptions.CompletedOptions) (*Config, error) {
 		apiHandler = filters.WithWarningRecorder(apiHandler)
 
 		apiHandler = kcpfilters.WithAuditEventClusterAnnotation(apiHandler)
+		apiHandler = kcpfilters.WithBlockInactiveLogicalClusters(apiHandler, c.KcpSharedInformerFactory.Core().V1alpha1().LogicalClusters())
```
sttts, line 448: Wondering, what speaks against putting it just before auth, i.e. line 411 or 426 ?

pkg/server/filters/filters.go
```diff
@@ -78,6 +86,49 @@ func WithAuditEventClusterAnnotation(handler http.Handler) http.HandlerFunc {
...
+			}
+		}
+
+		cluster := request.ClusterFrom(req.Context())
+		if cluster != nil && !cluster.Name.Empty() && !isException {
+			logicalCluster, err := kcpClusterClient.Cluster(cluster.Name).Lister().Get(corev1alpha1.LogicalClusterName)
+			if err == nil {
+				if ann, ok := logicalCluster.ObjectMeta.Annotations[inactiveAnnotation]; ok && ann == "true" {
+					responsewriters.ErrorNegotiated(
+						apierrors.NewBadRequest("logical cluster is marked inactive"),
```
sttts, line 117: 400 is explicitly a client error. Looking through the list

403 Forbidden
423 Locked
503 Service Unavailable

What do we return when accessing a logical cluster without the `LogicalCluster` object? 403 I assume? Maybe that would be the right response here too.

sttts:
Wondering, a small e2e test would be good.

sttts:
What about discovery? Should we allow that? 

Why do you need OpenAPI (am just curious) ?

embik:
What kind of maintenance would this feature allow? If everything except the LogicalCluster is blocked, what can you reasonably do during a maintenance window?

pkg/server/config.go
RedbackThomson: I honestly had no intuition as to where it should go. Let me know :)

RedbackThomson:
> What kind of maintenance would this feature allow? If everything except the LogicalCluster is blocked, what can you reasonably do during a maintenance window?

My use case is that I'm rebuilding the cluster by restoring from an etcd snapshot and in that time I don't want controllers to access resources before they're fully reapplied.

PR #3155, opened 2024-08-21
:sparkles: Add original user/group as extra to the impersonating client used by virtual workspace

turkenh:
## Summary

We have some validations on our CRDs where we would like to act based on the user information, i.e., we want to allow a specific field to only be initialized by a controller but no one else. The current implementation of build virtual workspaces loses the original user info while doing impersonation. 

This PR adds original user and groups information to impersonating client as `extra`, so that our validator can extract the user/group information and make decisions accordingly. 

## Related issue(s)

N/A

## Release Notes

```release-note
Add original user/groups information as extra to the impersonating client used by virtual workspace.
```

### DIFF
--- a/pkg/virtual/apiexport/builder/build.go
+++ b/pkg/virtual/apiexport/builder/build.go
@@ -52,7 +52,16 @@ import (
 	kcpinformers "github.com/kcp-dev/kcp/sdk/client/informers/externalversions"
 )
 
-const VirtualWorkspaceName string = "apiexport"
+const (
+	// VirtualWorkspaceName is the name of the virtual workspace.
+	VirtualWorkspaceName string = "apiexport"
+	// OriginalUserAnnotationKey is the key used in a user's "extra" to
+	// specify the original user of the authenticating request.
+	OriginalUserAnnotationKey = "experimental.authorization.kcp.io/original-username"
+	// OriginalGroupsAnnotationKey is the key used in a user's "extra" to
+	// specify the original groups of the authenticating request.
+	OriginalGroupsAnnotationKey = "experimental.authorization.kcp.io/original-groups"
+)
 
 func BuildVirtualWorkspace(
 	rootPathPrefix string,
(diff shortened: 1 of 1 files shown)


pkg/virtual/apiexport/builder/build.go
```diff
@@ -113,6 +122,15 @@ func BuildVirtualWorkspace(
 						serviceaccount.ClusterNameKey: {cluster.Name.Path().String()},
 					},
 				}
+
+				if user, ok := genericapirequest.UserFrom(ctx); ok {
+					// We pass the original user and groups as extra fields to
+					// the impersonation config so that the receiver can make
+					// decisions based on the original user/groups.
+					impersonationConfig.Impersonate.Extra[OriginalUserAnnotationKey] = []string{user.GetName()}
+					impersonationConfig.Impersonate.Extra[OriginalGroupsAnnotationKey] = user.GetGroups()
```
sttts, line 131: trying to wrap my head around whether we actually want to preserve the original groups as part of the impersonated identity.

What we really want is this:
- inside of the virtualized workspace this user should have the admin CRUD permissions because the authorization is done by the virtual workspace.
- outside of the virtualized workspace (e.g. in admission webhooks) we want the actually export owner identity to be effective.

I think we have that partially somewhere called "soft impersonation". Have to double check our thinking there.

🤔

s-urbaniak: AFAIR I also believe we should leave groups out. we are really only interested in the person's identity. Groups are per-workspace and not global.

Take this with a grain of salt: it's been a while.

turkenh: > outside of the virtualized workspace (e.g. in admission webhooks) we want the actually export owner identity to be effective.

Owner identity is both its username and groups. Even we add its groups to the list of groups, it would be partial, permissions might have been granted to its user only. So, on the admission side, I would rather rely on extra where I can find both username/group.

sttts:
With https://github.com/kcp-dev/kcp/pull/3156 the problem here goes away as the external virtual workspace user is preserved.

pkg/virtual/apiexport/builder/build.go
```diff
@@ -52,7 +52,16 @@ import (
...
-const VirtualWorkspaceName string = "apiexport"
+const (
+	// VirtualWorkspaceName is the name of the virtual workspace.
+	VirtualWorkspaceName string = "apiexport"
+	// OriginalUserAnnotationKey is the key used in a user's "extra" to
+	// specify the original user of the authenticating request.
+	OriginalUserAnnotationKey = "authorization.kcp.io/original-username"
+	// OriginalGroupsAnnotationKey is the key used in a user's "extra" to
+	// specify the original groups of the authenticating request.
+	OriginalGroupsAnnotationKey = "authorization.kcp.io/original-groups"
```
sttts, line 63: I would go with this as a stop-gap until the [warrant topic](https://github.com/kcp-dev/kcp/pull/3156) is resolved. Can you change these to `experimental.authorization.kcp.io/*`? Then I think we can merge.

turkenh: Done 👍

PR #3164, opened 2024-09-17
:bug: admission/webhooks: properly cache webhooks and cleanup

sttts:
## Summary

The old code only added webhook sources, but never cleaned up. This PR bump kube to get CEL optimization for webhook filters, and it properly caches the webhook per cluster and cleans up on deletion.

## Related issue(s)

Fixes #

## Release Notes

```release-note
Fix memory leak in admission webhook caching.
```

### DIFF
--- a/pkg/admission/mutatingwebhook/plugin.go
+++ b/pkg/admission/mutatingwebhook/plugin.go
@@ -28,18 +28,22 @@ import (
 	kcpkubernetesclientset "github.com/kcp-dev/client-go/kubernetes"
 	"github.com/kcp-dev/logicalcluster/v3"
 
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apiserver/pkg/admission"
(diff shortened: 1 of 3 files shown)


pkg/admission/mutatingwebhook/plugin.go
```diff
@@ -56,10 +60,16 @@ type Plugin struct {
...
 	globalKubeSharedInformerFactory kcpkubernetesinformers.SharedInformerFactory
 
-	getAPIBindings func(clusterName logicalcluster.Name) ([]*apisv1alpha1.APIBinding, error)
+	getAPIBinding     func(clusterName logicalcluster.Name, gr schema.GroupResource) (*apisv1alpha1.APIBinding, error)
+	getLogicalCluster func(clusterName logicalcluster.Name) (*corev1alpha1.LogicalCluster, error)
 
-	managerLock   sync.Mutex
-	managersCache map[logicalcluster.Name]generic.Source
+	lock  sync.RWMutex
+	cache map[logicalcluster.Name]map[logicalcluster.Name]clusterCache // by request and hook source cluster.
```
mjudeikis, line 67: Can you elaborate why we need sourceCluster and cluster logicalclusters in this cache?

mjudeikis: Its Source and Target? maybe name them this way if this the case

sttts: because we have two informers giving us webhook configuration: locally in the workspace and from the APIExport workspace (cache server).

PR #3169, opened 2024-09-19
:bug: reconcilers: ignore sync events for secondary informers

sttts:
## Summary

Controllers should only act on sync events for the primary informers, i.e. those whose objects are queued. Secondary informers are those whose objects are mapped to the primary objects before putting the key into the queue. We want to avoid that objects with many secondary objects get resynced a lot more often.

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


yastij (approved):
/approve
/lgtm

PR #3176, opened 2024-10-03
:seedling: Fix bad smells

cupakob:
## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/test/e2e/fixtures/webhook/webhook.go
+++ b/test/e2e/fixtures/webhook/webhook.go
@@ -19,6 +19,7 @@ package webhook
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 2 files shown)


embik:
Hi @cupakob, thank you for the contribution! Before we can accept your contribution, please make sure it is signed off (the `-s` flag on `git commit`) so you signal that you are accepting the [DCO](https://github.com/kcp-dev/kcp/blob/main/DCO). In addition, it would be great if you could write a short summary (how are the things you've changed smells?) so that this PR has the necessary context.

cupakob:
> how are the things you've changed smells?

done

test/e2e/fixtures/webhook/webhook.go
```diff
@@ -59,7 +60,7 @@ func (s *AdmissionWebhookServer) StartTLS(t *testing.T, certFile, keyFile string
 
 	go func() {
 		err := serv.ListenAndServeTLS(certFile, keyFile)
-		if err != nil && err != http.ErrServerClosed {
+		if err != nil && errors.Is(err, http.ErrServerClosed) {
```
embik, line 63: This looks like a change in logic to me (from "err is **not** `http.ErrServerClosed`" to "err _is_ `http.ErrServerClosed`"). Is that intentional?

embik (changes requested):

test/e2e/fixtures/webhook/webhook.go
cupakob: no, its fixed now

embik (approved):
/approve

Thank you for contributing to kcp!

PR #3179, opened 2024-10-11
:bug: Fix probable bug

cupakob:
Possible integer overflow when converting int -> uint16. This bug may never occur, but it is better to have proper handling of this case.

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/topology/partitionset/partitionset_reconcile.go
+++ b/pkg/reconciler/topology/partitionset/partitionset_reconcile.go
@@ -97,7 +97,8 @@ func (c *controller) reconcile(ctx context.Context, partitionSet *topologyv1alph
 	} else {
 		matchLabelsMap = partition(shards, dimensions, nil)
 	}
-	partitionSet.Status.Count = uint16(len(matchLabelsMap))
+
+	partitionSet.Status.Count = uint(len(matchLabelsMap))
 	existingMatches := map[string]struct{}{}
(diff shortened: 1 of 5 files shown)


pkg/reconciler/topology/partitionset/partitionset_reconcile.go
```diff
@@ -97,7 +99,11 @@ func (c *controller) reconcile(ctx context.Context, partitionSet *topologyv1alph
 	} else {
 		matchLabelsMap = partition(shards, dimensions, nil)
 	}
-	partitionSet.Status.Count = uint16(len(matchLabelsMap))
```
sttts, line 100: can we just make this an uint?

cupakob: should not be a problem...but then why not int? what is the advantage of the uint?

stevekuznetsov: Negative numbers of shards doesn't make any sense either :)

cupakob: yeah, sure...but then another convert is needed from uint to int. In this case `len()` should not return a negative number ;)

cupakob: ok, changed to uint

embik (approved):
/approve

PR #3183, opened 2024-10-17
:sparkles:  Add workspace phase update logic

mjudeikis:
## Summary

Add a reconciler to switch workspaces phase to "Unavailable" if any workspace conditions are not ready.
This way, if some of the auxiliary conditions are `Unavailable`, the phase will change.

Adds mounts controller to propagate status to the workspace when mount is not ready. This way we don't use the workspace at all if the mount is not responding. 

Adds code to index/proxy to not serve workspaces which are in `Unavailable` state, so we avoid using workspaces with errorous backends. 

partially based on https://github.com/kcp-dev/enhancements/pull/6 

## Related issue(s)

Fixes #

## Release Notes

```release-note
Add workspace phase reporter reconciler 
Add the Unavailable phase to the API
Implement exclusion of Unavailable workspaces from serving via proxy to avoid serving something which is not supposed to be served. 
```

### DIFF
--- a/pkg/index/index.go
+++ b/pkg/index/index.go
@@ -81,6 +81,12 @@ func (c *State) UpsertWorkspace(shard string, ws *tenancyv1alpha1.Workspace) {
 	if ws.Status.Phase == corev1alpha1.LogicalClusterPhaseScheduling {
 		return
 	}
+	// If the workspace is unavailable, we should delete it from the index,
+	// as it is not usable.
+	if ws.Status.Phase == corev1alpha1.LogicalClusterPhaseUnavailable {
+		c.DeleteWorkspace(shard, ws)
+		return
+	}
 	clusterName := logicalcluster.From(ws)
 
 	c.lock.RLock()
--- a/pkg/index/index.go
+++ b/pkg/index/index.go
@@ -39,6 +39,10 @@ type Result struct {
 	Shard string
 	// Cluster canonical path
 	Cluster logicalcluster.Name
+
+	// ErrorCode is the HTTP error code to return for the request.
+	// If this is set, the URL and Shard fields are ignored.
+	ErrorCode int
 }
 
 // PathRewriter can rewrite a logical cluster path before the actual mapping through
@@ -55,9 +59,13 @@ func New(rewriters []PathRewriter) *State {
 		shardClusterParentCluster: map[string]map[logicalcluster.Name]logicalcluster.Name{},
 		shardBaseURLs:             map[string]string{},
 		// Experimental feature: allow mounts to be used with Workspaces
-		// structure: (clusterName, workspace name) -> string serialized mount objects
+		// structure: (shard, logical cluster, workspace name) -> string serialized mount objects
 		// This should be simplified once we promote this to workspace structure.
-		clusterWorkspaceMountAnnotation: map[logicalcluster.Name]map[string]string{},
+		shardClusterWorkspaceMountAnnotation: map[string]map[logicalcluster.Name]map[string]string{},
+
+		// shardClusterWorkspaceNameErrorCode is a map of shar,logical cluster, workspace to error code when we want to return an error code
+		// instead of a URL.
+		shardClusterWorkspaceNameErrorCode: map[string]map[logicalcluster.Name]map[string]int{},
 	}
 }
 
@@ -74,36 +82,38 @@ type State struct {
 	shardClusterParentCluster map[string]map[logicalcluster.Name]logicalcluster.Name            // (shard name, logical cluster) -> parent logical cluster
 	shardBaseURLs             map[string]string                                                 // shard name -> base URL
 	// Experimental feature: allow mounts to be used with Workspaces
-	clusterWorkspaceMountAnnotation map[logicalcluster.Name]map[string]string // (clusterName, workspace name) -> mount object string
+	shardClusterWorkspaceMountAnnotation map[string]map[logicalcluster.Name]map[string]string // (shard name, logical cluster, workspace name) -> mount object string
+
+	shardClusterWorkspaceNameErrorCode map[string]map[logicalcluster.Name]map[string]int // (shard name, logical cluster, workspace name) -> error code
 }
 
 func (c *State) UpsertWorkspace(shard string, ws *tenancyv1alpha1.Workspace) {
 	if ws.Status.Phase == corev1alpha1.LogicalClusterPhaseScheduling {
 		return
 	}
-	// If the workspace is unavailable, we should delete it from the index,
-	// as it is not usable.
-	if ws.Status.Phase == corev1alpha1.LogicalClusterPhaseUnavailable {
-		c.DeleteWorkspace(shard, ws)
-		return
-	}
+
 	clusterName := logicalcluster.From(ws)
-
-	c.lock.RLock()
-	cluster := c.shardWorkspaceNameCluster[shard][clusterName][ws.Name]
-	mountObjString := c.clusterWorkspaceMountAnnotation[clusterName][ws.Name] // experimental feature
-	c.lock.RUnlock()
-
-	// TODO(mjudeikis): we are allowing upsert in 2 cases:
-	// 1. cluster name is different
-	// 2. mount object string is different (updated, added, or removed)
-	// When we promote this to workspace structure, we should make this check smarter and better tested.
-	if (cluster.String() == ws.Spec.Cluster) && (ws.Annotations[tenancyv1alpha1.ExperimentalWorkspaceMountAnnotationKey] != "" && mountObjString == ws.Annotations[tenancyv1alpha1.ExperimentalWorkspaceMountAnnotationKey]) {
-		return
-	}
 
 	c.lock.Lock()
 	defer c.lock.Unlock()
+
+	// If the workspace is unavailable, we set custom error code for it. And add it to the index as normal.
+	// TODO(mjudeikis): Once we have one more case - move to a separate function.
+	if ws.Status.Phase == corev1alpha1.LogicalClusterPhaseUnavailable {
+		if c.shardClusterWorkspaceNameErrorCode[shard] == nil {
+			c.shardClusterWorkspaceNameErrorCode[shard] = map[logicalcluster.Name]map[string]int{}
+		}
+		if c.shardClusterWorkspaceNameErrorCode[shard][logicalcluster.Name(ws.Spec.Cluster)] == nil {
+			c.shardClusterWorkspaceNameErrorCode[shard][logicalcluster.Name(ws.Spec.Cluster)] = map[string]int{}
+		}
+		// Unavailable workspaces should return 503
+		c.shardClusterWorkspaceNameErrorCode[shard][logicalcluster.Name(ws.Spec.Cluster)][ws.Name] = 503
+	} else {
+		delete(c.shardClusterWorkspaceNameErrorCode[shard][logicalcluster.Name(ws.Spec.Cluster)], ws.Name)
+		if len(c.shardClusterWorkspaceNameErrorCode[shard][logicalcluster.Name(ws.Spec.Cluster)]) == 0 {
+			delete(c.shardClusterWorkspaceNameErrorCode[shard], logicalcluster.Name(ws.Spec.Cluster))
+		}
+	}
 
 	if cluster := c.shardWorkspaceNameCluster[shard][clusterName][ws.Name]; cluster.String() != ws.Spec.Cluster {
 		if c.shardWorkspaceNameCluster[shard] == nil {
@@ -119,11 +129,14 @@ func (c *State) UpsertWorkspace(shard string, ws *tenancyv1alpha1.Workspace) {
 		c.shardClusterParentCluster[shard][logicalcluster.Name(ws.Spec.Cluster)] = clusterName
 	}
 
-	if mountObjString := c.clusterWorkspaceMountAnnotation[clusterName][ws.Name]; mountObjString != ws.Annotations[tenancyv1alpha1.ExperimentalWorkspaceMountAnnotationKey] {
-		if c.clusterWorkspaceMountAnnotation[clusterName] == nil {
-			c.clusterWorkspaceMountAnnotation[clusterName] = map[string]string{}
+	if mountObjString := c.shardClusterWorkspaceMountAnnotation[shard][clusterName][ws.Name]; mountObjString != ws.Annotations[tenancyv1alpha1.ExperimentalWorkspaceMountAnnotationKey] {
+		if c.shardClusterWorkspaceMountAnnotation[shard] == nil {
+			c.shardClusterWorkspaceMountAnnotation[shard] = map[logicalcluster.Name]map[string]string{}
 		}
-		c.clusterWorkspaceMountAnnotation[clusterName][ws.Name] = ws.Annotations[tenancyv1alpha1.ExperimentalWorkspaceMountAnnotationKey]
+		if c.shardClusterWorkspaceMountAnnotation[shard][clusterName] == nil {
+			c.shardClusterWorkspaceMountAnnotation[shard][clusterName] = map[string]string{}
+		}
+		c.shardClusterWorkspaceMountAnnotation[shard][clusterName][ws.Name] = ws.Annotations[tenancyv1alpha1.ExperimentalWorkspaceMountAnnotationKey]
 	}
 }
 
@@ -132,7 +145,7 @@ func (c *State) DeleteWorkspace(shard string, ws *tenancyv1alpha1.Workspace) {
 
 	c.lock.RLock()
 	_, foundCluster := c.shardWorkspaceNameCluster[shard][clusterName][ws.Name]
-	_, foundMount := c.clusterWorkspaceMountAnnotation[clusterName][ws.Name]
+	_, foundMount := c.shardClusterWorkspaceMountAnnotation[shard][clusterName][ws.Name]
 	c.lock.RUnlock()
 
 	if !foundCluster && !foundMount {
@@ -162,11 +175,15 @@ func (c *State) DeleteWorkspace(shard string, ws *tenancyv1alpha1.Workspace) {
 		}
 	}
 
-	if _, foundMount = c.clusterWorkspaceMountAnnotation[clusterName][ws.Name]; foundMount {
-		delete(c.clusterWorkspaceMountAnnotation[clusterName], ws.Name)
-		if len(c.clusterWorkspaceMountAnnotation[clusterName]) == 0 {
-			delete(c.clusterWorkspaceMountAnnotation, clusterName)
+	if _, foundMount = c.shardClusterWorkspaceMountAnnotation[shard][clusterName][ws.Name]; foundMount {
+		delete(c.shardClusterWorkspaceMountAnnotation[shard][clusterName], ws.Name)
+		if len(c.shardClusterWorkspaceMountAnnotation[shard][clusterName]) == 0 {
+			delete(c.shardClusterWorkspaceMountAnnotation[shard], clusterName)
 		}
+	}
+	delete(c.shardClusterWorkspaceNameErrorCode[shard][clusterName], ws.Name)
+	if len(c.shardClusterWorkspaceNameErrorCode[shard][clusterName]) == 0 {
+		delete(c.shardClusterWorkspaceNameErrorCode[shard], clusterName)
 	}
 }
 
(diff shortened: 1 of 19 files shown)


sdk/apis/core/v1alpha1/logicalcluster_types.go
```diff
@@ -59,13 +59,14 @@ const (
...
 //
-// +kubebuilder:validation:Enum=Scheduling;Initializing;Ready
+// +kubebuilder:validation:Enum=Scheduling;Initializing;Ready;NotReady
 type LogicalClusterPhaseType string
 
 const (
 	LogicalClusterPhaseScheduling   LogicalClusterPhaseType = "Scheduling"
 	LogicalClusterPhaseInitializing LogicalClusterPhaseType = "Initializing"
 	LogicalClusterPhaseReady        LogicalClusterPhaseType = "Ready"
+	LogicalClusterPhaseNotReady     LogicalClusterPhaseType = "NotReady"
```
sttts, line 69: can you document what it means? What is the behaviour of authz and the front-proxy?

pkg/reconciler/tenancy/workspacestatuses/README.md
```diff
@@ -0,0 +1,6 @@
+# Workspace statuses controller
+
+The controller has a single purpose - to update the workspace phase based on
+the conditions of the workspace. This way if we have an external process that
+updates the workspace, we can reflect the status of the workspace in the workspace
+object.
```
sttts, line 6: why is this not a reconciler in tenancy/workspace ?

mjudeikis: Idea was that for now this is separate and easier to control via featureflag (not even installed). But I think same can be done in the workspace controller too. less writes

mjudeikis: will move. read the code there, Makes more sense there

mjudeikis:
@sttts @embik so this might need a discussion. 
With Mountpoint, we need a way to prevent users from interacting with the workspace if the mount is dead. The current behaviour is that if the URL does not respond you will get an error from the reverse proxy inside the proxy. But this does not cover  all cases fully, as the URL might be alive but we just don't want the users to interact with the workspace (maintenance, or just temporary off).

In addition, it's very confusing (found this from running this in my fork for some time already), if mountpoint goes down due to network, dead agent, you name it... And Workspace still shows `Ready` but if you try to access it - you get a proxy error, so need a way to indicate to users "This is not healthy".

The solution I proposed was to add a new phase - `NotReady`. So the flow would be:
`Scheduling -> Initializing -> Ready <--> NotReady`. So Workspace can jump between ready and not-ready (like a node).
This is achieved by extending phase reconciliation to detect if any condition in the workspace object is not `True` and flip the phase to it.

Even more, not even serve these workspaces at proxy layer if they are `NotReady/Unavailable`. This allows to do some maintenance modes too at shard level apis. 

This allows some third-party component or in our case mounts controller to inject a condition of its own and based on those change `phase` of the Workspace. It works quite nicely, but there is a risk that somebody adds some non-blocking condition, like `Update is needed for Workspace APIs` which is more informational, and not terminal.

So I think not to go further from the original enhancement https://github.com/kcp-dev/enhancements/pull/6/files#diff-716b46559ae0795860e35aebf72fce98bdf288d935756a40b917410386e10870R265

I think 1 change needs to be made:
1. Flip phase based only on Condition which is Prefixed with `Workspace`. In example `WorkspaceMountNotReady` -> makes phase -> NotReady

And original proposal was:
```
Scheduling -> Initializing -> Ready <-> Unavailable.
```
I somehow think NotReady is better due to its parallels with wider k8s terminology.
```
Scheduling -> Initializing -> Ready <-> NotReady.
```

Comments, concerns?

sdk/apis/core/v1alpha1/logicalcluster_types.go
mjudeikis: done

sttts:
I think we have to distinguish between being switched to the new phase for some hard reason, like maintenance, versus soft reasons like network issues, or when the agent is not reachable. The former is definitely a phase, the later is not. 

There are many actors potentially who might experience problems to reach a mount. Think of some HA env where one component thinks the mount is down and the other side thinks it is up. If one of them can mark the mount unavailable for all components, that amplifies errors.

mjudeikis:
> I think we have to distinguish between being switched to the new phase for some hard reason, like maintenance, versus soft reasons like network issues, or when the agent is not reachable. The former is definitely a phase, the later is not.

I think final result of this is that author needs to decide if accessing workspace would be allowed or not. 
Image situation: My workspace is backed by mount. So situation is like in linux where I have file but I soft link ontop of the file pointing to some other file. So my file still exists but its not accesible as I have link ontop of it. So now if the link malfunctions and workspace is still served via front-proxy I might in-intentionally get access to original "file" and I would not see what I expect to see. In this case I want either to show right file or not show at all. 

So all and all, is there anything you suggest to change in this proposed implementation? If document, would our public docs would be enough? 

```
➜  mjudeikis tree
.
├── test
│   └── test.txt
└── test2
    └── test2.txt
➜  mjudeikis ln -sfn ../test2/test2.txt ./test/test.txt
➜  mjudeikis tree
.
├── test
│   └── test.txt -> ../test2/test2.txt
└── test2
    └── test2.txt
```


> There are many actors potentially who might experience problems to reach a mount. Think of some HA env where one component thinks the mount is down and the other side thinks it is up. If one of them can mark the mount unavailable for all components, that amplifies errors.

I think this now for implemented to deal with. But its authors ownership to make decision.

sttts:
"Unavailable" as a phase seems to cover the soft reasons too. I want to discourage an implementor to implement health checks in some controller and then set that phase indirectly through a condition. 

Note: a condition from a health check is fine. 

Maybe "Inaccessible" is better?

sdk/apis/core/v1alpha1/logicalcluster_types.go
```diff
@@ -59,13 +59,17 @@ const (
...
-// +kubebuilder:validation:Enum=Scheduling;Initializing;Ready
+// +kubebuilder:validation:Enum=Scheduling;Initializing;Ready;Unavailable
 type LogicalClusterPhaseType string
 
 const (
 	LogicalClusterPhaseScheduling   LogicalClusterPhaseType = "Scheduling"
 	LogicalClusterPhaseInitializing LogicalClusterPhaseType = "Initializing"
 	LogicalClusterPhaseReady        LogicalClusterPhaseType = "Ready"
+	// LogicalClusterPhaseUnavailable phase is used to indicate that the logical cluster us unavailable to be used.
+	// It will will not be served via front-proxy when in this state.
```
sttts, line 70: would be more precise: what does front-proxy return? I would expect 503.

pkg/index/index.go
```diff
@@ -81,6 +81,12 @@ func (c *State) UpsertWorkspace(shard string, ws *tenancyv1alpha1.Workspace) {
 	if ws.Status.Phase == corev1alpha1.LogicalClusterPhaseScheduling {
 		return
 	}
+	// If the workspace is unavailable, we should delete it from the index,
+	// as it is not usable.
+	if ws.Status.Phase == corev1alpha1.LogicalClusterPhaseUnavailable {
+		c.DeleteWorkspace(shard, ws)
```
sttts, line 87: will this result in a 503?

pkg/reconciler/tenancy/workspace/workspace_reconcile_phase.go
```diff
@@ -106,7 +116,33 @@ func (r *phaseReconciler) reconcile(ctx context.Context, workspace *tenancyv1alp
...
+			return reconcileStatusStopAndRequeue, nil
+		}
 	}
 
 	return reconcileStatusContinue, nil
 }
+
+// checkConditionAndSetPhase checks if the workspace is ready by checking conditions and sets the phase accordingly.
+// It returns true if the phase was changed, false otherwise.
+func checkConditionAndSetPhase(workspace *tenancyv1alpha1.Workspace) bool {
```
sttts, line 131: the name does not really fit as it is not responsible for the initial phases. Maybe call it `updateTerminalConditionsAndPhase`?

pkg/reconciler/tenancy/workspace/workspace_reconcile_phase.go
```diff
@@ -106,7 +116,33 @@ func (r *phaseReconciler) reconcile(ctx context.Context, workspace *tenancyv1alp
 			logger.Info("workspace content is deleted")
 			return reconcileStatusContinue, nil
 		}
+		logger.Info("workspace is ready, checking conditions", "conditions", workspace.Status.Conditions)
```
sttts, line 119: V(4) ? Could we make it edge based?

pkg/reconciler/tenancy/workspace/workspace_reconcile_phase.go
```diff
@@ -106,7 +116,33 @@ func (r *phaseReconciler) reconcile(ctx context.Context, workspace *tenancyv1alp
...
+		if c.Status == v1.ConditionFalse && strings.HasPrefix(string(c.Type), "Workspace") {
+			notReady = true
+			break
+		}
+	}
+	if notReady && workspace.Status.Phase != corev1alpha1.LogicalClusterPhaseUnavailable {
+		workspace.Status.Phase = corev1alpha1.LogicalClusterPhaseUnavailable
+		return true
+	}
+	if !notReady && workspace.Status.Phase != corev1alpha1.LogicalClusterPhaseReady {
```
sttts, line 143: `== Unavailble`

sdk/apis/core/v1alpha1/logicalcluster_types.go
mjudeikis: 401 as with all not found resources. we would need to change front-proxy to return something else

pkg/index/index.go
mjudeikis: no, like not found, in our case 401... strange too

pkg/reconciler/tenancy/workspace/workspace_reconcile_phase.go
mjudeikis: can you elaborate what do you mean?

pkg/index/index.go
```diff
@@ -57,7 +61,11 @@ func New(rewriters []PathRewriter) *State {
 		// Experimental feature: allow mounts to be used with Workspaces
 		// structure: (clusterName, workspace name) -> string serialized mount objects
 		// This should be simplified once we promote this to workspace structure.
-		clusterWorkspaceMountAnnotation: map[logicalcluster.Name]map[string]string{},
+		clusterWorkspaceMountAnnotation: map[string]map[logicalcluster.Name]map[string]string{},
```
sttts, line 64: no sure I understand the three keys. shard, cluster, workspace? Then it should be called like that.

pkg/index/index.go
```diff
@@ -57,7 +61,11 @@ func New(rewriters []PathRewriter) *State {
...
+
+		// ErrorCodeTracker is a map of logical cluster to error code when we want to return an error code
+		// instead of a URL.
+		errorCodeTracker: map[string]map[logicalcluster.Name]map[string]int{},
```
sttts, line 68: shard, cluster, workspace?

pkg/reconciler/tenancy/workspace/workspace_reconcile_phase.go
sttts: only log when state changes

pkg/index/index.go
mjudeikis: ah, yes,changed last second. will rename.

pkg/index/index.go
```diff
@@ -74,31 +82,39 @@ type State struct {
...
-	mountObjString := c.clusterWorkspaceMountAnnotation[clusterName][ws.Name] // experimental feature
-	c.lock.RUnlock()
 
-	// TODO(mjudeikis): we are allowing upsert in 2 cases:
-	// 1. cluster name is different
-	// 2. mount object string is different (updated, added, or removed)
-	// When we promote this to workspace structure, we should make this check smarter and better tested.
-	if (cluster.String() == ws.Spec.Cluster) && (ws.Annotations[tenancyv1alpha1.ExperimentalWorkspaceMountAnnotationKey] != "" && mountObjString == ws.Annotations[tenancyv1alpha1.ExperimentalWorkspaceMountAnnotationKey]) {
-		return
-	}
```
mjudeikis, line 97: These were "overoptimization". It backfired when now condition and annotations is being updated (in 2 different reconcile cycles due to not try to update both in one) where annotations gets updated and now this prevents it to move forward and update error codes cache. 

Overengineering from my side :/

pkg/index/index.go
```diff
@@ -55,9 +59,13 @@ func New(rewriters []PathRewriter) *State {
 		shardClusterParentCluster: map[string]map[logicalcluster.Name]logicalcluster.Name{},
 		shardBaseURLs:             map[string]string{},
 		// Experimental feature: allow mounts to be used with Workspaces
-		// structure: (clusterName, workspace name) -> string serialized mount objects
+		// structure: (shard, logical cluster, workspace name) -> string serialized mount objects
 		// This should be simplified once we promote this to workspace structure.
-		clusterWorkspaceMountAnnotation: map[logicalcluster.Name]map[string]string{},
+		shardClusterWorkspaceMountAnnotation: map[string]map[logicalcluster.Name]map[string]string{},
```
mjudeikis, line 64: We have few duplicate ones here. We could do some more nicer abstraction here.
Most data is `shard` -> `cluster`-> `workspace` -> `data` - so some in-memory store mini-engine could help. Now its very hard to make changes where and make sure we clean, add, and update when we need to. 

Something to think of during insomnia nights.

mjudeikis:
/retest 
show recociler in tests? looking

pkg/index/index.go
```diff
@@ -74,31 +82,39 @@ type State struct {
 	shardClusterParentCluster map[string]map[logicalcluster.Name]logicalcluster.Name            // (shard name, logical cluster) -> parent logical cluster
 	shardBaseURLs             map[string]string                                                 // shard name -> base URL
 	// Experimental feature: allow mounts to be used with Workspaces
-	clusterWorkspaceMountAnnotation map[logicalcluster.Name]map[string]string // (clusterName, workspace name) -> mount object string
+	shardClusterWorkspaceMountAnnotation map[string]map[logicalcluster.Name]map[string]string // (shard name, logical cluster, workspace name) -> mount object string
+
+	shardCLusterWorkspaceNameErrorCode map[string]map[logicalcluster.Name]map[string]int // (shard name, logical cluster, workspace name) -> error code
```
embik, line 87: Nit:

```suggestion
	shardClusterWorkspaceNameErrorCode map[string]map[logicalcluster.Name]map[string]int // (shard name, logical cluster, workspace name) -> error code
```

embik (approved):
This seems to look fine for now to me, I agree with @sttts that we should clearly document expectations for implementors once we have this stabilised a bit and are ready for people to integrate with it.

sdk/apis/tenancy/v1alpha1/types_mounts.go
```diff
@@ -41,6 +41,14 @@ const (
 	MountPhaseUnknown MountPhaseType = "Unknown"
 )
 
+const (
+	// MountConditionReady is the condition type for MountReady.
+	MountConditionReady conditionsv1alpha1.ConditionType = "WorkspaceMountReady"
+
+	// MountNotReady is reason when the mount is not ready.
+	MountReasonNotReady = "WorkspaceMountNotReady"
```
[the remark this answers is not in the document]
sttts: Could or should this be a reason defined by the controller backing the mount resource?

pkg/reconciler/tenancy/workspace/workspace_reconcile_phase.go
```diff
@@ -106,7 +116,34 @@ func (r *phaseReconciler) reconcile(ctx context.Context, workspace *tenancyv1alp
...
+			return reconcileStatusStopAndRequeue, nil
+		}
 	}
 
 	return reconcileStatusContinue, nil
 }
+
+// updateTerminalConditionsAndPhase checks if the workspace is ready by checking conditions and sets the phase accordingly.
+// It returns true if the phase was changed, false otherwise.
+func updateTerminalConditionsAndPhase(workspace *tenancyv1alpha1.Workspace) bool {
```
sttts, line 132: this does not change the conditions.

Can we express this functionally without side-effects?
```golang
func terminalConditionPhase(workspace *tenancyv1alpha1.Workspace) (string, bool) {
```
Maybe like that. Easier to reason about.

pkg/reconciler/tenancy/workspacemounts/workspacemounts_reconcile_updater.go
```diff
@@ -44,6 +47,11 @@ func (r *workspaceStatusUpdater) reconcile(ctx context.Context, workspace *tenan
 			if err != nil {
 				return reconcileStatusStopAndRequeue, err
 			}
+		} else {
```
sttts, line 50: no need for `if workspace.Annotations != nil {` above. Would remove it.

pkg/reconciler/tenancy/workspacemounts/workspacemounts_reconcile_updater.go
```diff
@@ -77,10 +85,35 @@ func (r *workspaceStatusUpdater) reconcile(ctx context.Context, workspace *tenan
...
+		switch mount.MountStatus.Phase {
+		case tenancyv1alpha1.MountPhaseReady:
+			if current == nil {
+				conditions.MarkTrue(workspace, tenancyv1alpha1.MountConditionReady)
+				return reconcileStatusContinue, nil
+			}
+			if current.Status == v1.ConditionTrue {
+				return reconcileStatusContinue, nil
+			}
+			conditions.MarkTrue(workspace, tenancyv1alpha1.MountConditionReady)
```
sttts, line 111: isn't this and the if above the same?

pkg/reconciler/tenancy/workspacemounts/workspacemounts_reconcile_updater.go
```diff
@@ -77,10 +85,35 @@ func (r *workspaceStatusUpdater) reconcile(ctx context.Context, workspace *tenan
...
+			return reconcileStatusContinue, nil
+		default:
+			msg := fmt.Sprintf("Mount is not reporting ready. See %s %s status for more details", obj.GroupVersionKind().Kind, obj.GetName())
```
sttts, line 114:
```suggestion
			msg := fmt.Sprintf("Mount is not reporting ready. See %s %q status for more details", obj.GroupVersionKind().Kind, obj.GetName())
```

sdk/apis/tenancy/v1alpha1/types_mounts.go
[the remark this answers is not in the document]
mjudeikis: Gonna just drop it. And use Condition above. Make sense

sdk/apis/core/v1alpha1/logicalcluster_types.go
```diff
@@ -59,13 +59,17 @@ const (
...
+	// Possible state transitions are from Ready to Unavailable and from Unavailable to Ready.
+	LogicalClusterPhaseUnavailable LogicalClusterPhaseType = "Unavailable"
```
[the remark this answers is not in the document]
mjudeikis: Updated

pkg/reconciler/tenancy/workspace/workspace_reconcile_phase.go
```diff
@@ -127,9 +127,9 @@ func (r *phaseReconciler) reconcile(ctx context.Context, workspace *tenancyv1alp
 	return reconcileStatusContinue, nil
 }
 
-// updateTerminalConditionsAndPhase checks if the workspace is ready by checking conditions and sets the phase accordingly.
+// terminalConditionPhase checks if the workspace is ready by checking conditions and sets the phase accordingly.
 // It returns true if the phase was changed, false otherwise.
-func updateTerminalConditionsAndPhase(workspace *tenancyv1alpha1.Workspace) bool {
+func terminalConditionPhase(workspace *tenancyv1alpha1.Workspace) bool {
```
sttts, line 132: not really what I meant. Either call it `terminalConditionPhase` and make it side-effect-free (i.e. returning the phase string, or `updateTerminalConditionPhase` (without the "And").

mjudeikis: Taking second one :) As reconcilers already playing a lot with pointers and editing existing objects - this just felt natural.

PR #3186, opened 2024-10-25
:seedling: Rename variables

cupakob:
Some variable names are same like reserved words. This should be avoided

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


embik (approved):
/approve

PR #3188, opened 2024-10-26
:bug: Fix regression when generating deep copy

mjudeikis:
## Summary

At some point we lost this in the last releases. `kube_codegen.sh` does not work the same way as before. 
We will need to migrate the rest of the generators when we can. For now just patching the bleeding. 

## Related issue(s)

Fixes #

## Release Notes

```release-note
Fix regression in DeepCopy generator
```


embik (approved):
/approve

my bad! Thanks for catching this.

PR #3194, opened 2024-10-30
:bug: Fix possible leak (W.i.P.)

cupakob:
Not closing os.File must not, but may lead to a leak. It is better to close it manually instead of waiting for the next GC to be triggered

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/cmd/sharded-test-server/cache.go
+++ b/cmd/sharded-test-server/cache.go
@@ -74,12 +74,12 @@ func startCacheServer(ctx context.Context, logDirPath, workingDir string, synthe
 	}
 
 	logger := klog.FromContext(ctx)
-	// defer func() {
-	// 	err = logFile.Close()
-	// 	if err != nil {
-	// 		logger.Error(err, "failed to close the log file")
-	// 	}
-	// }()
+	defer func() {
+		err = logFile.Close()
+		if err != nil {
+			logger.Error(err, "failed to close the log file")
+		}
+	}()
 
 	writer := helpers.NewHeadWriter(logFile, out)
 	cmd.Stdout = writer
--- a/cmd/sharded-test-server/cache.go
+++ b/cmd/sharded-test-server/cache.go
@@ -73,6 +73,14 @@ func startCacheServer(ctx context.Context, logDirPath, workingDir string, synthe
 		return nil, "", err
 	}
 
+	logger := klog.FromContext(ctx)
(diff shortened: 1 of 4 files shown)


embik:
Hi @cupakob, I'm not sure this is a flaky test. Have you verified that the same test works for you locally?

cupakob:
> Hi @cupakob, I'm not sure this is a flaky test. Have you verified that the same test works for you locally?

nope, in the log output I dont really see a problem related to my changes

i will run it localy, maybe I will see more

cupakob:
@embik currently i don't have any changes in the branch (all changes are commented), but the tests are still failing...it looks like flaky tests somehow, right?

embik:
> @embik currently i don't have any changes in the branch (all changes are commented), but the tests are still failing...it looks like flaky tests somehow, right?

I think you're right, the test occasionally flakes. You can see in [older runs](https://public-prow.kcp.k8c.io/job-history/s3/prow-public-data/pr-logs/directory/pull-kcp-test-e2e-shared?buildId=1849144573578711040) that it used to fail sometimes before this PR. Apologies for the confusion caused by it.

I think the consistent failures we saw earlier in this PR had a differentiation though: They ran into the 2h timeout, so those jobs were running for much longer. The flakes are usually happening within the first 15-20 minutes.

cupakob:
> I think you're right, the test occasionally flakes. You can see in [older runs](https://public-prow.kcp.k8c.io/job-history/s3/prow-public-data/pr-logs/directory/pull-kcp-test-e2e-shared?buildId=1849144573578711040) that it used to fail sometimes before this PR. Apologies for the confusion caused by it.

all good, not a problem at all ;)

> I think the consistent failures we saw earlier in this PR had a differentiation though: They ran into the 2h timeout, so those jobs were running for much longer. The flakes are usually happening within the first 15-20 minutes.

okay, but I don't have an explination, how the defer function, where I close the file, can cause a 2h timeout...any hints are welcome.

sttts:
@cupakob is this blocked on anything? Anything we can help with?

cupakob:
> @cupakob is this blocked on anything? Anything we can help with?

I have no idea what exactly is wrong and how some defer functions can cause 2h timeout. I will rebase on main and will try again to check, if the problem still exists.

PR #3196, opened 2024-10-30
:bug: Terminate kcp process if controller installation fails

embik:
## Summary

As per discussions on Slack (https://kubernetes.slack.com/archives/C021U8WSAFK/p1730131234224829), it makes sense to let the kcp process crash when it wins a leader election but then fails to install controllers. If this happens, it might hold onto the leader election lock but no controllers are actually running to reconcile objects. By making it crash, we make a new leader election possible.

The leader election package unfortunately doesn't have a "native" way to handle such failures.

## Related issue(s)

Fixes #

## Release Notes

```release-note
Purposefully crash if leader election was won but controllers failed to install, allowing another instance to take leadership
```


palnabarun (approved):
/lgtm

PR #3096, opened 2024-03-21
✨ Feature: Re-enable webhook token authentication

ajwdev:
## Summary

This simply re-enables Kubernetes [Webhook Token Authentication](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#webhook-token-authentication) that was previously commented out.

I verified it worked in a local setup against a custom authenticator. See below:

```
$ \cat me.kubeconfig
apiVersion: v1
clusters:
- cluster:
    insecure-skip-tls-verify: true
    server: https://localhost:6443/clusters/root
  name: workspace.kcp.io/current
- cluster:
    insecure-skip-tls-verify: true
    server: https://localhost:6443/clusters/root
  name: workspace.kcp.io/previous
contexts:
- context:
    cluster: workspace.kcp.io/current
    user: test
  name: workspace.kcp.io/current
- context:
    cluster: workspace.kcp.io/previous
    user: test
  name: workspace.kcp.io/previous
current-context: workspace.kcp.io/current
kind: Config
preferences: {}
users:
- name: test
  user:
    token: eyJhbGciO <SNIP> RVA


$ KUBECONFIG=me.kubeconfig kubectl kcp ws create foo --enter
error: workspaces.tenancy.kcp.io is forbidden: User "andrewwi@vmware.com" cannot create resource "workspaces" in API group "tenancy.kcp.io" at the cluster scope: access denied
zsh: exit 1     KUBECONFIG=me.kubeconfig kubectl kcp ws create foo --enter

$ \cat admin.yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: me-admin
subjects:
- kind: User
  name: andrewwi@vmware.com
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-admin
  
  
$ KUBECONFIG=.kcp/admin.kubeconfig kubectl apply -f admin.yaml
clusterrolebinding.rbac.authorization.k8s.io/me-admin created

$ KUBECONFIG=me.kubeconfig kubectl kcp ws create foo --enter
Workspace "foo" (type root:organization) created. Waiting for it to be ready...
Workspace "foo" (type root:organization) is ready to use.
Current workspace is "root:foo" (type root:organization).
```

KCP was started with the following flags:

```
bin/kcp start --authentication-token-webhook-version=v1 --authentication-token-webhook-config-file=/tmp/webhook.yaml
```

## Release Notes

```release-note
[...]

PR #3198, opened 2024-11-13
✨ Enable webhook authorization options

xrstf:
## Summary
This PR allows to configure an external webhook for authorization. Previously the relevant flags have been explicitly disabled in kcp, because kcp has its own authorizer. The webhook authorizer will sit next to the existing authorizers (allowPath, allowGroup) and so is an alternative to the built-in authorizers.

In the documentation, it looks like this:

I chose to let the webhook authorizer run in parallel to the existing RBAC chain so that admins have full freedom to tweak the authorization. This comes with the warning that if your webhook is configured to deny everything, it will block kcp's startup (thankfully, kcp recovers quickly when the webhook stops denying everything and starts to either have no opinion or allow things explicitly).

Since the webhook authorizer is part of a union with the others, and more importantly comes *after* the allowGroups/Paths authorizers, webhooks cannot prevent healthchecks from functioning.

---

Originally I started by simply enabling the authz options in the built-in server options, i.e. remove the `o.GenericControlPlane.Authorization = nil // we have our own` line, to surface the existing webhook options from Kubernetes.

However I did not manage to integrate the webhook well into kcp. When the webhook authorizer is enabled, the code would then _not_ run the RBAC PostStart hook in the server anymore, so RBAC policies like the `cluster-admin` ClusterRole were simply not being created anymore. I fear to break a lot of things if I hack around in the Kubernetes fork and so ultimately decided to not use the authz options in the built-in server options, but instead _add_ them to the kcp authz options (where we also configure --always-allow-paths, for example).

This leads to a bit more code, because some steps like setting up the CLI flags have to be done manually, but allowed much greater control over how the webhook is loaded and was ultimately simpler to do. Now whatever kcp does with the webhook flags, will not affect Kubernetes' internal bootstrapping anymore.

I also renamed a few variables to make things a bit more clear.

---

In addition to the code change, I also revamped the authorizer documentation for kcp. The old one still mentioned non-existing authorizers (like the "top-level org authorizer") or behaviour that simply doesn't exist (like the workspace content authorizer alledgedly injecting data into the auth context for the authorizers downstream, which it doesn't do in reality).

Since the original ASCIIChart looked terrible due to my font settings, I was very glad to see that mkdocs-material comes with a Mermaid integration and so I replaced the oringinal chart with a much easier to maintain Mermaid diagram.

This PR also includes a dependency bump to include https://github.com/kcp-dev/kubernetes/pull/151, which will take care of including the cluster name in the webhook's payload. In that dependency bump is also, slightly hidden, a cleanup for the `bump-k8s` script, which failed at least since the 1.31 bump since the `legacy-cloud-providers` module doesn't exist anymore.

---

Also I vote that we rename the "System CRD Authorizer" to something like "System Resource Authorizer", because the authorizer does not actually handle CustomResourceDefinitions at all. It handles APIBindings and APIExports. Calling it a "CRD Authorizer" is confusing to me.

## Release Notes
```release-note
Add support for external webhook authorization.
```

### DIFF
--- a/cmd/kcp/kcp.go
+++ b/cmd/kcp/kcp.go
@@ -78,8 +78,8 @@ func main() {
 		}
 	}
 
-	serverOptions := options.NewOptions(rootDir)
-	serverOptions.Server.GenericControlPlane.Logs.Verbosity = logsapiv1.VerbosityLevel(2)
+	kcpOptions := options.NewOptions(rootDir)
+	kcpOptions.Server.GenericControlPlane.Logs.Verbosity = logsapiv1.VerbosityLevel(2)
 
 	startCmd := &cobra.Command{
 		Use:   "start",
@@ -101,30 +101,30 @@ func main() {
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// run as early as possible to avoid races later when some components (e.g. grpc) start early using klog
-			if err := logsapiv1.ValidateAndApply(serverOptions.Server.GenericControlPlane.Logs, kcpfeatures.DefaultFeatureGate); err != nil {
+			if err := logsapiv1.ValidateAndApply(kcpOptions.Server.GenericControlPlane.Logs, kcpfeatures.DefaultFeatureGate); err != nil {
 				return err
 			}
 
-			completed, err := serverOptions.Complete()
+			completedKcpOptions, err := kcpOptions.Complete()
 			if err != nil {
 				return err
 			}
 
-			if errs := completed.Validate(); len(errs) > 0 {
+			if errs := completedKcpOptions.Validate(); len(errs) > 0 {
 				return errors.NewAggregate(errs)
 			}
 
 			logger := klog.FromContext(cmd.Context())
-			logger.Info("running with selected batteries", "batteries", strings.Join(completed.Server.Extra.BatteriesIncluded, ","))
+			logger.Info("running with selected batteries", "batteries", strings.Join(completedKcpOptions.Server.Extra.BatteriesIncluded, ","))
 
 			ctx := genericapiserver.SetupSignalContext()
 
-			serverConfig, err := server.NewConfig(ctx, completed.Server)
+			serverConfig, err := server.NewConfig(ctx, completedKcpOptions.Server)
 			if err != nil {
 				return err
 			}
 
-			completedConfig, err := config.Complete()
+			completedConfig, err := serverConfig.Complete()
 			if err != nil {
 				return err
 			}
@@ -146,7 +146,7 @@ func main() {
 
 	// add start named flag sets to start flags
 	fss := cliflag.NamedFlagSets{}
-	serverOptions.AddFlags(&fss)
+	kcpOptions.AddFlags(&fss)
 	globalflag.AddGlobalFlags(fss.FlagSet("global"), cmd.Name(), logs.SkipLoggingConfigurationFlags())
 	startFlags := startCmd.Flags()
 	for _, f := range fss.FlagSets {
--- a/cmd/kcp/kcp.go
+++ b/cmd/kcp/kcp.go
@@ -117,7 +117,9 @@ func main() {
 			logger := klog.FromContext(cmd.Context())
 			logger.Info("running with selected batteries", "batteries", strings.Join(completed.Server.Extra.BatteriesIncluded, ","))
 
-			config, err := server.NewConfig(completed.Server)
+			ctx := genericapiserver.SetupSignalContext()
+
+			serverConfig, err := server.NewConfig(ctx, completed.Server)
 			if err != nil {
 				return err
 			}
@@ -126,8 +128,6 @@ func main() {
 			if err != nil {
 				return err
 			}
-
-			ctx := genericapiserver.SetupSignalContext()
 
 			// the etcd server must be up before NewServer because storage decorators access it right away
 			if completedConfig.EmbeddedEtcd.Config != nil {
--- a/docs/content/concepts/authorization/authorizers.md
+++ b/docs/content/concepts/authorization/authorizers.md
@@ -5,46 +5,89 @@ description: >
 
 # Authorizers
 
-The following authorizers are configured in kcp:
+In kcp, a request has four different ways of being admitted:
 
-| Authorizer                             | Description                                                                       |
-|----------------------------------------|-----------------------------------------------------------------------------------|
-| Top-Level organization authorizer      | checks that the user is allowed to access the organization                        |
-| Workspace content authorizer           | determines additional groups a user gets inside of a workspace                    |
-| Maximal permission policy authorizer   | validates the maximal permission policy RBAC policy in the API exporter workspace |
-| Local Policy authorizer                | validates the RBAC policy in the workspace that is accessed                       |
-| Kubernetes Bootstrap Policy authorizer | validates the RBAC Kubernetes standard policy                                     |
+* It can be made to one of the preconfigured paths that do not require authorization, like `/healthz`.
+* It can be performed by a user in one of the configured always-allow groups, by default `system:masters`.
+* It can pass through the RBAC chain and match configured Roles and ClusterRoles.
+* It can be permitted by an external HTTPS webhook backend.
 
 They are related in the following way:
 
-1. top-level organization authorizer must allow
-2. workspace content authorizer must allow, and adds additional (virtual per-request) groups to the request user influencing the follow authorizers.
-3. maximal permission policy authorizer must allow
-4. one of the local authorizer or bootstrap policy authorizer must allow.
+``` mermaid
+graph TD
+    start(Request):::state --> main_alt[/one of\]:::or
+    main_alt --> aapa[Always Allow Paths Auth]
+    main_alt --> aaga[Always Allow Groups Auth]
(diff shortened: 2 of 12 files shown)


pkg/server/options/options.go
```diff
@@ -114,15 +113,6 @@ func NewOptions(rootDir string) *Options {
...
 	o.GenericControlPlane.SecureServing.ServerCert.CertDirectory = rootDir
-	o.GenericControlPlane.Authentication = kubeoptions.NewBuiltInAuthenticationOptions().
-		WithAnonymous().
-		WithBootstrapToken().
-		WithClientCert().
-		WithOIDC().
-		WithRequestHeader().
-		WithServiceAccounts().
-		WithTokenFile().
-		WithWebHook()
```
embik, line 125: I don't get why this is removed. There's probably a reason, but I'm not sure what it is.

embik (changes requested):

pkg/server/options/options.go
xrstf: Orginally, in #424, we started with 

```go
o.GenericControlPlane.Authentication = kubeoptions.NewBuiltInAuthenticationOptions().
		WithAnonymous().
		WithBootstrapToken().
		WithClientCert().
		WithOIDC().
		WithRequestHeader().
		// WithServiceAccounts().
		WithTokenFile()
	//WithWebHook()
```

In #681 the ServiceAccount option was re-enabled and in #3096 the webhook auth was re-enabled. So now the code in kcp is identical to upstream's `WithAll()` and so in order to cleanup and reduce cognitive load, I chose to remove the code here. I'm not sure if it was intended to not follow upstream.

cmd/kcp/kcp.go
```diff
@@ -78,8 +78,8 @@ func main() {
 		}
 	}
 
-	serverOptions := options.NewOptions(rootDir)
-	serverOptions.Server.GenericControlPlane.Logs.Verbosity = logsapiv1.VerbosityLevel(2)
+	kcpOptions := options.NewOptions(rootDir)
```
mjudeikis, line 81: Not sure how this is cleaner :) maybe `kcpServerOptions` - to implied this is still server options, not some other kcp related options?

docs/content/concepts/authorization/authorizers.md
```diff
@@ -5,46 +5,89 @@ description: >
...
 
-| Authorizer                             | Description                                                                       |
-|----------------------------------------|-----------------------------------------------------------------------------------|
-| Top-Level organization authorizer      | checks that the user is allowed to access the organization                        |
-| Workspace content authorizer           | determines additional groups a user gets inside of a workspace                    |
-| Maximal permission policy authorizer   | validates the maximal permission policy RBAC policy in the API exporter workspace |
-| Local Policy authorizer                | validates the RBAC policy in the workspace that is accessed                       |
-| Kubernetes Bootstrap Policy authorizer | validates the RBAC Kubernetes standard policy                                     |
+* It can be made to one of the preconfigured paths that do not require authorization, like `/healthz`.
+* It can be performed by a user in one of the configured always-allow groups, by default `system:masters`.
```
mjudeikis, line 11: `system:masters` and `system:kcp:admins`. Where first should never be used by users itself. For external users second one is intended to be used.

docs/content/concepts/authorization/authorizers.md
```diff
@@ -5,46 +5,89 @@ description: >
...
+
+### Always Allow Paths Authorizer
+
+Like in vanilla Kubernetes, this authorizer always grants access to the configured URL paths. This is
+used for the health and liveness checks of kcp.
+
+### Always Allow Groups Authorizer
+
+This authorizer always permits access if the user is in one of the configured groups. By default this
+only includes the `system:masters` group.
```
mjudeikis, line 58: and `system:kcp:admins`?

pkg/server/options/authorization.go
```diff
@@ -79,11 +112,48 @@ func (s *Authorization) AddFlags(fs *pflag.FlagSet) {
...
+
+		if config.EgressSelector != nil {
+			egressDialer, err := config.EgressSelector.Lookup(egressselector.ControlPlane.AsNetworkContext())
+			if err != nil {
+				return err
+			}
+			authorizationConfig.CustomDial = egressDialer
+		}
+
+		authorizer, _, err := authorizationConfig.New(ctx, config.APIServerID)
```
mjudeikis, line 149: Is this `ctx` just initializing one or used for runtime? If the first one, can we cap it with ctxWithTimeout? As it might block all startup hanging without a clear error? If second - I don't think we can do anything :/

hack/bump-k8s.sh
```diff
@@ -29,7 +29,7 @@ set -o xtrace
 # Note: setting GOPROXY=direct allows us to bump very quickly after the fork has been committed to.
 GITHUB_USER=${GITHUB_USER:-kcp-dev}
 GITHUB_REPO=${GITHUB_REPO:-kubernetes}
-BRANCH=${BRANCH:-kcp-feature-logical-clusters-1.24-v3}
+BRANCH=${BRANCH:-kcp-1.31.0}
```
mjudeikis, line 32: I didnt even knew this script existed :D worth adding into rebase document!

mjudeikis (changes requested):
Few nits. 
I would really like to have some example of running webhook authz server as example. Or reference to one in upstream? By any chance you had one done while doing this PR in testing?

And maybe adding some e2e for this would great too. Myabe out of scope for this PR but great.

xrstf:
> Or reference to one in upstream? By any chance you had one done while doing this PR in testing?

I used [xrstf/httest](https://github.com/xrstf/httest) to quickly spawn a local HTTPS server with CA/serving cert, which I can then use to inspect the webhook requests (it's a spleen for me, I like to see what's actually going over the wire).

So during development I did

```bash
$ httest --tls --response allow.json --trace
INFO[2024-12-09T10:38:54+01:00] Creating private key…                         file=.httest/ca.key
INFO[2024-12-09T10:38:54+01:00] Creating CA certificate…                      file=.httest/ca.crt
INFO[2024-12-09T10:38:54+01:00] Creating private key…                         file=.httest/serving.key
INFO[2024-12-09T10:38:54+01:00] Creating serving certificate…                 cn=localhost file=.httest/serving.crt
INFO[2024-12-09T10:38:54+01:00] Listening securely…                           address=localhost:8080 ca=.httest/ca.crt domains=[localhost 127.0.0.1]
```

Then I created my auth.kubeconfig using the CA key:

```yaml
apiVersion: v1
kind: Config
clusters:
  - name: httest
    cluster:
      certificate-authority: /home/xrstf/gospace/src/github.com/kcp-dev/kcp/.httest/ca.crt
      server: https://localhost:8080/
current-context: webhook
contexts:
  - name: webhook
    context:
      cluster: httest
```

I had prepared a few standard responses, like the `deny.json`:

```json
{
   "apiVersion": "authorization.k8s.io/v1beta1",
   "kind": "SubjectAccessReview",
   "status": {
     "allowed": false,
     "denied": true,
     "reason": "computer says no"
   }
}
```

I could then observe requests coming in to the webhook, like

```
INFO[2024-12-09T10:41:30+01:00] Request                                       method=POST path=/?timeout=30s protocol=HTTP/2.0 remote=127.0.0.1:46418 ua=Go-http-client/2.0
POST /?timeout=30s HTTP/2.0
Host: localhost:8080
Accept: application/json, */*
Accept-Encoding: gzip
Content-Length: 434
Content-Type: application/json
User-Agent: Go-http-client/2.0

{"kind":"SubjectAccessReview","apiVersion":"authorization.k8s.io/v1beta1","metadata":{"creationTimestamp":null},"spec":{"resourceAttributes":{"verb":"create","group":"tenancy.kcp.io","version":"v1alpha1","resource":"workspacetypes"},"user":"system:kcp:bootstrapper","group":["system:kcp:tenancy:workspace-bootstrapper","system:authenticated"],"extra":{"authorization.kubernetes.io/cluster-name":["root"]}},"status":{"allowed":false}}
```

cmd/kcp/kcp.go
xrstf: I found `serverOptions.Server` to be confusing to read, thinking that this struct here holds all the options you can pass to `kcp`, hence `kcpOptions`, of which the server part is only a part.

But I'm also not super married to this, if you want I can revert it.

docs/content/concepts/authorization/authorizers.md
xrstf: https://github.com/kcp-dev/kcp/blob/main/pkg/server/options/authorization.go#L48 says it's only 1 default group though.

pkg/server/options/authorization.go
xrstf: The context given here is ultimately used to start a goroutine that does

```go
filesystem.WatchUntil(
		ctx,
		r.reloadInterval,
		r.initialConfig.ReloadFile,
		func() {
			r.checkFile(ctx)
		},
		func(err error) {
			klog.ErrorS(err, "watching authorization config file")
		},
	)
```

in the auto-reloading webhook authorizer (it reloads the given webhook-config-file automatically at runtime).

mjudeikis:
/lgtm
/approve 

We can iterate on this. We need to test this "in prod"

PR #3206, opened 2024-12-06
:bug: Fix impersonation for non-system users

mjudeikis:
## Summary

This is an attempt to improve impersonation security stance. 

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


pkg/server/handler.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/emicklei/go-restful/v3"
 	jwt2 "gopkg.in/square/go-jose.v2/jwt"
 
+	authenticationapi "k8s.io/api/authentication/v1"
```
sttts, line 33: authenticationv1

pkg/server/handler.go
```diff
@@ -253,6 +256,86 @@ func WithRequestIdentity(handler http.Handler) http.Handler {
...
+			}
+		}
+
+		responsewriters.ErrorNegotiated(
+			apierrors.NewForbidden(schema.GroupResource{}, "", fmt.Errorf("impersonation is not allowed for the requestor")),
+			errorCodecs, schema.GroupVersion{}, w, req)
+	})
+}
+
+func checkImpersonation(userGroups []string, requestedGroups []string) bool {
```
sttts, line 297:
```suggestion
func validImpersonation(userGroups []string, requestedGroups []string) bool {
```

pkg/server/handler.go
```diff
@@ -253,6 +256,86 @@ func WithRequestIdentity(handler http.Handler) http.Handler {
...
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		// Impersonation check is only done when impersonation is requested.
+		// And impersonations is only allowed for the users, who have metadata in the ctx.
+		impersonationUser := req.Header.Get(authenticationapi.ImpersonateUserHeader)
+		if len(impersonationUser) != 0 {
+			requestor, exists := request.UserFrom(req.Context())
+			if !exists {
+				responsewriters.ErrorNegotiated(
+					apierrors.NewForbidden(schema.GroupResource{}, "", fmt.Errorf("impersonation is not allowed for the requestor")),
+					errorCodecs, schema.GroupVersion{}, w, req)
```
sttts, line 282:
```suggestion
				responsewriters.ErrorNegotiated(
					apierrors.NewForbidden(schema.GroupResource{}, "", fmt.Errorf("impersonation is not allowed for the requestor")),
					errorCodecs, schema.GroupVersion{}, w, req)
					return
```

pkg/server/handler.go
sttts: not sure I can follow here with the levels.

What is the logic? Do we have multiple levels of being privileged?

pkg/authorization/bootstrap/policy.go
```diff
@@ -46,6 +46,21 @@ const (
...
+	// SystemMastersGroup is the group inherited from k8s codebase - all powerful, all knowing!
+	// Users should not be added to this group.
+	SystemMastersGroup = "system:masters"
+)
+
+const (
+	// SystemKcpAdminUser is the user that has all permissions across all workspaces.
+	SystemKcpAdminUser = "kcp-admin"
+	// SystemShardAdmin is the user that has all permissions across all workspaces.
+	SystemShardAdmin = "shard-admin"
```
sttts, line 59: does the name alone has impact?

sttts: same for kcp-admin.

pkg/authorization/bootstrap/policy.go
```diff
@@ -46,6 +46,21 @@ const (
 	SystemKcpWorkspaceAccessGroup = "system:kcp:workspace:access"
 )
 
+const (
+	// SystemMastersGroup is the group inherited from k8s codebase - all powerful, all knowing!
+	// Users should not be added to this group.
+	SystemMastersGroup = "system:masters"
```
sttts, line 52: this exists in Kube.

pkg/authorization/bootstrap/policy.go
```diff
@@ -46,6 +46,21 @@ const (
...
+)
+
+const (
+	// SystemKcpAdminUser is the user that has all permissions across all workspaces.
+	SystemKcpAdminUser = "kcp-admin"
+	// SystemShardAdmin is the user that has all permissions across all workspaces.
+	SystemShardAdmin = "shard-admin"
+	// SystemServiceAccountDefaultRest is the default service account for fake rest client.
+	SystemServiceAccountDefaultRest = "system:serviceaccount:default:kcp-rest"
```
sttts, line 61: how is this special?

pkg/server/handler.go
mjudeikis: yes. `system:masters` allows to impersonate `system:kcp:admin` but not other way around. and `system:kcp:admin` can impersonate anything bellow too. 

overkill? :D

pkg/authorization/bootstrap/policy.go
mjudeikis: with groups flow. it does not anymore. Will see if worth keeping so we can have less strings in code or just drop this.

pkg/authorization/bootstrap/policy.go
mjudeikis: with groups, not anymore too. But in general if we use this fake account in flows, wanted to make sure users cant create one. For edge cases we can't yet see.

sttts: Would try to make all user name irrelavant.

pkg/server/handler.go
```diff
@@ -253,6 +256,86 @@ func WithRequestIdentity(handler http.Handler) http.Handler {
...
 }
 
+var (
+	// impersonationGroups is a set of groups that are not allowed to impersonate
+	// if the user does not have a specific group or lower ID (lower = higher permissions).
+	impersonationGroups = map[string]int{
+		authorizationbootstrap.SystemMastersGroup:  0,
+		authorizationbootstrap.SystemKcpAdminGroup: 1,
+	}
+)
```
sttts, line 266:
```suggestion
type privilege int

const (
	superPrivileged privilege = iota
	priviledged
	unprivileged
)

var (
	// specialGroups specify groups with special meaning kcp. Lower privilege (= higher number)
	// cannot impersonate higher privilege levels.
	specialGroups = map[string]privilege{
		authorizationbootstrap.SystemMastersGroup:  superPrivileged,
		authorizationbootstrap.SystemKcpAdminGroup: priviledged,
	}
)
```

pkg/server/handler.go
sttts: We could drop the levels, yes:

- `system:masters` can do everything.
- Nobody else can impersonate `system:masters`, nor `system:kcp:admin`.

pkg/authorization/bootstrap/policy.go
mjudeikis: You will be very surprised :D The only place where this constant in defined in kube is cmd/kubeadm :D everywhere else it just hardcoded :D I think nobody bothered :D 

If you give me hint where kube keeps constants like this, I can try cleaning :D

pkg/authorization/bootstrap/policy.go
mjudeikis: Just left it as constant. It irrelavant now.

pkg/authorization/bootstrap/policy.go
mjudeikis: deleted those.

pkg/server/handler.go
mjudeikis: Got ya, overkill it is :)

mjudeikis:
Looks like we doing some impersonation in shared model (not sure  yet tests or real). Need to debug "

pkg/server/handler.go
```diff
@@ -253,6 +257,86 @@ func WithRequestIdentity(handler http.Handler) http.Handler {
 	})
 }
 
+type privilege int
+
+const (
+	superPrivileged privilege = iota
+	priviledged
+	unprivileged
+)
```
embik, line 266: Maybe it's just me, but I find this super confusing because it creates some weird situations in the code later on. Should we maybe start from `unprivileged`?

I mean e.g. this code snippet:

```go
// Check if user's max privilege is less than or equal to the requested group's privilege
if userMaxPrivilege > requestedPrivilege {
	return false // Deny impersonation if user's privilege is not sufficient
}
```

The comment is basically saying the inverse of the actual if condition. I feel like this will be prone to errors if we have to modify it at some point.

pkg/virtual/apiexport/builder/build.go
```diff
@@ -116,8 +116,8 @@ func BuildVirtualWorkspace(
 
 				impersonationConfig := rest.CopyConfig(cfg)
 				impersonationConfig.Impersonate = rest.ImpersonationConfig{
-					UserName: "system:serviceaccount:default:rest",
-					Groups:   []string{bootstrap.SystemKcpAdminGroup},
+					UserName: authorizationbootstrap.SystemServiceAccountDefaultRest,
```
embik, line 119: This is changing the user name from `system:serviceaccount:default:rest` to `system:serviceaccount:default:kcp-rest`. What is the reason?

pkg/server/handler.go
embik: cc @sttts since you proposed this change, I think.

pkg/authorization/bootstrap/policy.go
sttts: https://github.com/kubernetes/kubernetes/blob/a499facee693a1a83daadb82d88f7b51d324ffc5/staging/src/k8s.io/apiserver/pkg/authentication/user/user.go#L71

pkg/authorization/bootstrap/policy.go
sttts: Where do we need it?

pkg/server/handler.go
```diff
@@ -253,6 +257,86 @@ func WithRequestIdentity(handler http.Handler) http.Handler {
...
+// if they are valid. If they are not, will return a 403.
+// We check for impersonation in the request headers, early to avoid it being propagated to
+// the backend services.
+func WithImpersonationGatekeeper(handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		// Impersonation check is only done when impersonation is requested.
+		// And impersonations is only allowed for the users, who have metadata in the ctx.
+		// Else just pass the request.
+		impersonationUser := req.Header.Get(authenticationv1.ImpersonateUserHeader)
+		var requestor user.Info
```
sttts, line 287: why as var here? It's unused.

pkg/server/handler.go
```diff
@@ -253,6 +257,86 @@ func WithRequestIdentity(handler http.Handler) http.Handler {
...
+		if len(impersonationUser) != 0 {
+			var exists bool
+			requestor, exists = request.UserFrom(req.Context())
```
sttts, line 290: https://www.grammarpalette.com/requester-or-requestor-which-spelling/

pkg/server/handler.go
```diff
@@ -253,6 +257,86 @@ func WithRequestIdentity(handler http.Handler) http.Handler {
...
+				handler.ServeHTTP(w, req)
+				return
+			}
+
+			responsewriters.ErrorNegotiated(
+				apierrors.NewForbidden(schema.GroupResource{}, "", fmt.Errorf("impersonation is not allowed for the requestor")),
+				errorCodecs, schema.GroupVersion{}, w, req)
+			return
+		}
+		handler.ServeHTTP(w, req)
```
sttts, line 307: return early for `== 0` above.

pkg/admission/reservednames/admission.go
```diff
@@ -80,6 +81,11 @@ func NewReservedNames() *ReservedNames {
 				tenancyv1alpha1.Kind("WorkspaceType"),
 				tenancyv1alpha1.WorkspaceTypeReservedNames()...,
 			),
+			newReservedNameFn(
+				corev1.Resource("serviceaccounts"),
+				corev1.SchemeGroupVersion.WithKind("ServiceAccount").GroupKind(),
+				"kcp-rest",
+			),
```
sttts, line 88: am still not convinced of this. We should not need it.

pkg/server/handler.go
sttts: Commented the same below: invert the order. Makes it simpler.

pkg/cache/server/config.go
```diff
@@ -136,8 +136,11 @@ func NewConfig(opts *cacheserveroptions.CompletedOptions, optionalLocalShardRest
 	}
 
 	serverConfig.Config.BuildHandlerChainFunc = func(apiHandler http.Handler, genericConfig *genericapiserver.Config) (secure http.Handler) {
-		apiHandler = genericapiserver.DefaultBuildHandlerChainFromAuthz(apiHandler, genericConfig)
-		apiHandler = genericapiserver.DefaultBuildHandlerChainBeforeAuthz(apiHandler, genericConfig)
+
+		apiHandler = genericapiserver.DefaultBuildHandlerChainFromAuthzToCompletion(apiHandler, genericConfig)
```
mjudeikis, line 140: Not sure we want to catch impersonation errors in cache server too?

pkg/proxy/config.go
```diff
@@ -97,7 +97,7 @@ func NewConfig(ctx context.Context, opts *proxyoptions.Options) (*Config, error)
 		&clientcmd.ClientConfigLoadingRules{ExplicitPath: c.Options.ShardsKubeconfig},
 		// We override the Server here so that the user doesn't have to specify unused server value
 		// The Server must have HTTPS scheme otherwise CA won't be loaded (see IsConfigTransportTLS method)
-		&clientcmd.ConfigOverrides{ClusterInfo: clientcmdapi.Cluster{Server: "https://fakeserver.io"}}).ClientConfig()
+		&clientcmd.ConfigOverrides{ClusterInfo: clientcmdapi.Cluster{Server: "https://kcp.io/fake"}}).ClientConfig()
```
mjudeikis, line 100: Just cleaning. Once in my career somebody registered domain like this and data got leaked :D

pkg/server/handler.go
```diff
@@ -253,6 +257,86 @@ func WithRequestIdentity(handler http.Handler) http.Handler {
...
+			userMaxPrivilege = priv
+		}
+	}
+
+	// Iterate through each requested group and verify if user's privilege allows impersonation
+	for _, requestedGroup := range requestedGroups {
+		requestedPrivilege, exists := specialGroups[requestedGroup]
+		// If requested group is not recognized, treat it as the lowest privilege level
+		if !exists {
+			requestedPrivilege = unprivileged
```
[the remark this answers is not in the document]
mjudeikis: I was going for kernel logic but make sense

pkg/proxy/config.go
embik: Good idea.

pkg/virtual/apiexport/builder/build.go
sttts: would get rid of the constant. It is really just a random string.

pkg/cache/server/config.go
sttts: How do we authorize the cache-server? Only `system:masters` in the first place?

pkg/server/handler.go
```diff
@@ -253,6 +256,83 @@ func WithRequestIdentity(handler http.Handler) http.Handler {
...
+	return max
+}
+
+// validImpersonation checks if a user can impersonate all requested groups.
+func validImpersonation(userGroups, requestedGroups []string) bool {
+	userMax := maxUserPrivilege(userGroups)
+
+	for _, g := range requestedGroups {
+		p := unprivileged
+		if sp, found := specialGroups[g]; found {
```
sttts, line 325: no need for this. Just do `p := specialGroups[g]`. Zero value is `unprivileged`.

pkg/server/handler.go
```diff
@@ -253,6 +256,83 @@ func WithRequestIdentity(handler http.Handler) http.Handler {
...
+type privilege int
+
+const (
+	unprivileged privilege = iota
+	priviledged
+	superPrivileged
+)
+
+var (
+	// specialGroups specify groups with special meaning kcp. Lower privilege (= higher number)
```
sttts, line 268:
```suggestion
	// specialGroups specify groups with special meaning kcp. Lower privilege (= lower number)
```

embik:
/retest

flake?

sttts:
Looks good to go. Merge the Kube PR and revendor?

Lgtm
Approve.

mjudeikis:
> Looks good to go. Merge the Kube PR and revendor?

already done in this pr

pkg/cache/server/config.go
mjudeikis: right. I think we can revisit this post merge. WIll create issue to check this.

embik (approved):
Leaving final approval to @sttts.

PR #3207, opened 2024-12-06
📖 Add APIBindings example

SimonTheLeg:
## Summary

Adds an APIBinding example to fit with the example previously on the page and adds some more information on APIBindings/turns some of the bullets into full text.

The asciiflow link did not resolve for me and is also not rendered in the documentation at the moment, so I removed it.

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/content/concepts/apis/exporting-apis.md
+++ b/docs/content/concepts/apis/exporting-apis.md
@@ -1,4 +1,4 @@
-# Exporting APIs
+# Exporting and Binding APIs
 
 If you're looking to provide APIs that can be consumed by multiple workspaces, this section is for you!
 
@@ -374,12 +374,33 @@ TODO
 ## Bind to Exported APIs
 
 ### APIBinding
-TODO
-- "imports" all the `APIResourceSchemas` defined in an `APIExport` into its workspace
-- Also provides access to the group/resources defined in an `APIExport`'s `PermissionClaims` slice
-- APIs are "imported" by accepting `PermissionClaims` within the `APIBinding` for each of the `APIExport`'s group/resources
-- The consumer of the `APIExport` must be granted permission to the `bind` verb on that `APIExport` in order to create an `APIBinding`
-- An `APIBinding` is bound to a specific `APIExport` and associated `APIResourceSchema`s via the `APIBinding.Status.BoundResources` field, which will hold the identity information to precisely identify relevant objects.
-- how do I correctly reference an APIExport?
 
-[diagram1]: https://asciiflow.com/#/share/eJyrVspLzE1VssorzcnRUcpJrEwtUrJSqo5RqohRsrI0NdGJUaoEsozMzYCsktSKEiAnRkmBGPBoyh5qoZiYPGKtVFBwzs8rLs1NLVIIzy%2FKLi5ITE6FyJBgyIC4G5cMEYZgtVwhPDMlPbWkWMExwNMpMy8lMy%2BdFAOp5C44BXGNgiMWY6gY4igBgNUBTtgdAGQDw0khoCi%2FLDMFNfHgNMp5gPxCxeSJO4YR8YeqEilVuVYU5BeVKDya3kKCDdj5ONROw68WyS1BqcX5pUXJqcHJGam5iehx1vNoSgM10AT6xHATzlKsiZRcN4dKvl5C1xIDS9DgKMmICQyoqU24ZUgyBEcpRpYh6CURWYagl0EkGDKFSsljRoxSrVItAH%2FrdL4%3D
+`APIBindings` are used to import API resources. They contain a reference to an `APIExport` using the namespace and workspace path of an `APIExport` and will bind all APIs defined in the `APIExport`. The reference path needs to be provided to you by the provider of the API or an external catalog solution.
+
+Furthermore, `APIBindings` provide access to the groups/resources defined in an `APIExport's` `PermissionClaims` list.
+
+Returning to our previous example, we can use the following `APIBinding` to import the widgets api.
+
+```yaml
+apiVersion: apis.kcp.io/v1alpha1
+kind: APIBinding
+metadata:
+  name: example.kcp.io
+spec:
+  reference:
+    export:
+      name: example.kcp.io
+      path: "root:api-provider" # path of your api-provider workspace
+```
+
+It should be noted that `APIBindings` do not create `CRDs` or `APIResourceSchemas`. Instead APIs are directly bound. This means you can query for imported APIs using `kubectl api-resources`. Additionally you can use `kubectl explain` to get a detailed view on all fields of the API.
+
+```sh
+# inside consumer workspace
+kubectl api-resources --api-group='example.kcp.io'
+
+NAME      SHORTNAMES   APIVERSION                NAMESPACED   KIND
+widgets                example.kcp.io/v1alpha1   false        Widget
+```
+
+Furthermore, you can use the `APIBinding.Status.BoundResources` field to precisely identify which `APIResourceSchemas` have been imported.
--- a/docs/content/concepts/apis/exporting-apis.md
+++ b/docs/content/concepts/apis/exporting-apis.md
@@ -377,7 +377,7 @@ TODO
 
 `APIBindings` are used to import API resources. They contain a reference to an `APIExport` using the namespace and workspace path of an `APIExport` and will bind all APIs defined in the `APIExport`. The reference path needs to be provided to you by the provider of the API or an external catalog solution.
 
-Furthermore, `APIBindings` provide access to the groups/resources defined in an `APIExport's` `PermissionClaims` list.
+Furthermore, `APIBindings` provide the `APIExport` owner access to additional resources defined in an `APIExport`'s `PermissionClaims` list. `PermissionClaims` must be accepted by the user explicitly, before this access is granted. The resources can be builtin Kubernetes resources or resources from other `APIExports`.
 
 Returning to our previous example, we can use the following `APIBinding` to import the widgets api.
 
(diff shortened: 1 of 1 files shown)


docs/content/concepts/apis/exporting-apis.md
```diff
@@ -374,12 +374,33 @@ TODO
...
-- An `APIBinding` is bound to a specific `APIExport` and associated `APIResourceSchema`s via the `APIBinding.Status.BoundResources` field, which will hold the identity information to precisely identify relevant objects.
-- how do I correctly reference an APIExport?
-
-[diagram1]: https://asciiflow.com/#/share/eJyrVspLzE1VssorzcnRUcpJrEwtUrJSqo5RqohRsrI0NdGJUaoEsozMzYCsktSKEiAnRkmBGPBoyh5qoZiYPGKtVFBwzs8rLs1NLVIIzy%2FKLi5ITE6FyJBgyIC4G5cMEYZgtVwhPDMlPbWkWMExwNMpMy8lMy%2BdFAOp5C44BXGNgiMWY6gY4igBgNUBTtgdAGQDw0khoCi%2FLDMFNfHgNMp5gPxCxeSJO4YR8YeqEilVuVYU5BeVKDya3kKCDdj5ONROw68WyS1BqcX5pUXJqcHJGam5iehx1vNoSgM10AT6xHATzlKsiZRcN4dKvl5C1xIDS9DgKMmICQyoqU24ZUgyBEcpRpYh6CURWYagl0EkGDKFSsljRoxSrVItAH%2FrdL4%3D
+
+`APIBindings` are used to import API resources. They contain a reference to an `APIExport` using the namespace and workspace path of an `APIExport` and will bind all APIs defined in the `APIExport`. The reference path needs to be provided to you by the provider of the API or an external catalog solution.
+
+Furthermore, `APIBindings` provide access to the groups/resources defined in an `APIExport's` `PermissionClaims` list.
```
sttts, line 380:
```suggestion
Furthermore, `APIBindings` provide the `APIExport` owner access to additional resources defined in an `APIExport`'s `PermissionClaims` list. `PermissionClaims` must be accepted by the user explicitly, before this access is granted. The resources can be builtin Kubernetes resources or resources from other `APIExports`.
```

docs/content/concepts/apis/exporting-apis.md
```diff
@@ -374,12 +374,33 @@ TODO
...
+metadata:
+  name: example.kcp.io
+spec:
+  reference:
+    export:
+      name: example.kcp.io
+      path: "root:api-provider" # path of your api-provider workspace
+```
+
+It should be noted that `APIBindings` do not create `CRDs` or `APIResourceSchemas`. Instead APIs are directly bound. This means you can query for imported APIs using `kubectl api-resources`. Additionally you can use `kubectl explain` to get a detailed view on all fields of the API.
```
sttts, line 396:
```suggestion
It should be noted that `APIBindings` do not create `CRDs` or `APIResourceSchemas`. Instead APIs are directly bound. 

Bound APIs are like any other resources in kcp or Kubernetes. This means you can query for imported APIs using `kubectl api-resources`. Additionally you can use `kubectl explain` to get a detailed view on all fields of the API.
```

docs/content/concepts/apis/exporting-apis.md
sttts: Maybe also:

```
When an `APIExport` is changed after workspaces bound to them, new or changed APIs are automatically propagated to all `APIBindings`. New `PermissionClaims` on the other hand are NOT automatically accepted.

Comment (SimonTheLeg) [reply docs/content/concepts/apis/exporting-apis.md]: this is a really good point! Thanks I have added it

File: docs/content/concepts/apis/exporting-apis.md [hunk 3]
```diff
@@ -374,12 +374,36 @@ TODO
...
-- The consumer of the `APIExport` must be granted permission to the `bind` verb on that `APIExport` in order to create an `APIBinding`
-- An `APIBinding` is bound to a specific `APIExport` and associated `APIResourceSchema`s via the `APIBinding.Status.BoundResources` field, which will hold the identity information to precisely identify relevant objects.
-- how do I correctly reference an APIExport?
-
-[diagram1]: https://asciiflow.com/#/share/eJyrVspLzE1VssorzcnRUcpJrEwtUrJSqo5RqohRsrI0NdGJUaoEsozMzYCsktSKEiAnRkmBGPBoyh5qoZiYPGKtVFBwzs8rLs1NLVIIzy%2FKLi5ITE6FyJBgyIC4G5cMEYZgtVwhPDMlPbWkWMExwNMpMy8lMy%2BdFAOp5C44BXGNgiMWY6gY4igBgNUBTtgdAGQDw0khoCi%2FLDMFNfHgNMp5gPxCxeSJO4YR8YeqEilVuVYU5BeVKDya3kKCDdj5ONROw68WyS1BqcX5pUXJqcHJGam5iehx1vNoSgM10AT6xHATzlKsiZRcN4dKvl5C1xIDS9DgKMmICQyoqU24ZUgyBEcpRpYh6CURWYagl0EkGDKFSsljRoxSrVItAH%2FrdL4%3D
+
+`APIBindings` are used to import API resources. They contain a reference to an `APIExport` using the namespace and workspace path of an `APIExport` and will bind all APIs defined in the `APIExport`. The reference path needs to be provided to you by the provider of the API or an external catalog solution.
```
embik, line 378:
```suggestion
`APIBindings` are used to import API resources. They contain a reference to an `APIExport` using the name and workspace path of an `APIExport` and will bind all APIs defined in the `APIExport`. The reference path needs to be provided to you by the provider of the API or an external catalog solution if you don't have direct access to the workspace containing the `APIExport`.
```

I think the "namespace" part here is wrong, APIExports are cluster-level resources. Maybe "name" was meant? In addition, I suggest a little clarification on how to get the APIExport path (if you have access directly, you don't need someone to give you the path; it's entirely possible that people make this choice).

docs/content/concepts/apis/exporting-apis.md
```diff
@@ -374,12 +374,36 @@ TODO
...
+
+Furthermore, `APIBindings` provide the `APIExport` owner access to additional resources defined in an `APIExport`'s `PermissionClaims` list. `PermissionClaims` must be accepted by the user explicitly, before this access is granted. The resources can be builtin Kubernetes resources or resources from other `APIExports`.
```
embik, line 380: Nit: I think we shouldn't put permission claims in backticks, they aren't a standalone resource type. It would be more useful to put the JSONpath references to where permission claims are configured in `APIExport` and `APIBindings`.

docs/content/concepts/apis/exporting-apis.md
```diff
@@ -374,12 +374,36 @@ TODO
...
+metadata:
+  name: example.kcp.io
+spec:
+  reference:
+    export:
+      name: example.kcp.io
+      path: "root:api-provider" # path of your api-provider workspace
+```
+
+It should be noted that `APIBindings` do not create `CRDs` or `APIResourceSchemas`. Instead APIs are directly bound.
```
embik, line 397:
```suggestion
It should be noted that `APIBindings` do not create `CRDs` or `APIResourceSchemas` in the workspace. Instead APIs are directly bound. 
```

Strictly speaking (as far as I know), CRDs are created somewhere in the shadows, so let's make this sentence a little more precise.

docs/content/concepts/apis/exporting-apis.md
```diff
@@ -374,12 +374,36 @@ TODO
...
+
+```sh
+# inside consumer workspace
+kubectl api-resources --api-group='example.kcp.io'
+
+NAME      SHORTNAMES   APIVERSION                NAMESPACED   KIND
+widgets                example.kcp.io/v1alpha1   false        Widget
+```
+
+Furthermore, you can use the `APIBinding.Status.BoundResources` field to precisely identify which `APIResourceSchemas` have been imported.
```
embik, line 409:
```suggestion
Furthermore, you can use the `.status.boundResources` field of an `APIBinding` to precisely identify which `APIResourceSchemas` have been imported.
```

I would say we should reference via JSONpath, not Go structs. That only makes sense if you develop a controller, and you usually don't write controllers that check an APIBinding status.

embik (changes requested):
Some small changes I would suggest.

docs/content/concepts/apis/exporting-apis.md
janwillies, line 397: > Instead APIs are directly bound.
What does this mean?

docs/content/concepts/apis/exporting-apis.md
jmcshane, line 380: How are these permissions granted? Need an example in line with the `APIExport` sample above.

docs/content/concepts/apis/exporting-apis.md
mjudeikis: `using the name of the APIExport and path to the workspace, where an APIExport is located` 
something like that maybe

docs/content/concepts/apis/exporting-apis.md
mjudeikis: maybe something "apis are reflected". It there, but not in the same way in Kubernetes. So some clearing the waters is needed

docs/content/concepts/apis/exporting-apis.md
SimonTheLeg: fixed

docs/content/concepts/apis/exporting-apis.md
SimonTheLeg: fixed

docs/content/concepts/apis/exporting-apis.md
SimonTheLeg: yes good point!

docs/content/concepts/apis/exporting-apis.md
SimonTheLeg: I have re-worked this section again. I would really like to keep the word "bound" in as it comes directly from how Kubernetes behind the scenes calls it "binding apis".

I have made it more clear that this is something Kubernetes does for you behind the scenes and then in the next sentence I outlined what the practical implications of it are.
I think it is important in this section that we don't go into too many details and only outline what this means for an end-user.

docs/content/concepts/apis/exporting-apis.md
SimonTheLeg: very good point. I have added an example which matches the previous APIExport

SimonTheLeg:
@embik I think this PR is ready for review/merging now. I have addressed all the comments and I think it's a good state to merge. 
More improvements can be added in additional PRs

embik (changes requested):
One last (🤞🏻 ) thing.

embik (approved):
/approve

PR #1798, opened 2022-08-18
Introduce staticcheck linter for deprecations

varshaprasad96:
Follow up of https://github.com/kcp-dev/kcp/pull/1772

PR #3208, opened 2024-12-06
🌱 Bump golangci-lint, remove staticcheck

xrstf:
## Summary
This is some housekeeping for our linters.

* Originally, the idea behind running staticcheck standalone was that "golangci-lint should find deprecation, but it doesn't work" (in #1798). I checked and could not immediately see deprecated code that golangci-lint doesn't find by itself.
* Additionally, #1798 doesn't even enable the necessary check in staticcheck. Instead we configured:

   * ST1019 - duplicate import statements
   * ST1005 - error strings should not be capitalized
 
  Both of these things can be accomplished using `revive`. However we have to lower the minimum confidence level from 0.8 to 0.6 to surface these issues correctly. staticcheck (the standalone binary) has some additional smarts and produces fewer false positives because it recognizes code identifiers in error strings. However a few //nolint comments are to me an acceptable downside, compared to keeping a dedicated linter around.
* A whole bunch of functions in our SDK implicitly do a `fmt.Sprintf()`, which newer golangci-lint versions are complaining about if non-constant format strings are used. I pondered multiple solutions, but ultimately chose the simplest one that doesn't cause an API break (by calling `MarkFalse(reason)` now as `MarkFalse("%s", reason)`). Other options would be possible, like introducing new functions (`MarkFalse` and `MarkFalsef`) or removing the sprintf functionality alltogether, however since it's part of the SDK package, I feared an API break.

## Release Notes

```release-note
Update golangci-lint to 1.26.2, remove dependency on standalone staticcheck binary
```


test/e2e/authorizer/serviceaccounts_test.go
```diff
@@ -162,8 +162,6 @@ func TestServiceAccounts(t *testing.T) {
 		}},
 	}
 	for i, ttc := range testCases {
-		i := i
```
mjudeikis, line 165: I remember reading some go release notes but cant find any about this. We don't need this anymore, right? Or what is the reasons to removing these?

mjudeikis: its a closure thing with variables... Still confused why we removing this?

xrstf: Loop variables are copies since Go 1.22, so creating scoped copies is not necessary anymore.

See also https://go.dev/blog/go1.22#language-changes

embik (approved):
Leaving final approval to @mjudeikis.

PR #3212, opened 2024-12-10
:seedling: Update to Go 1.22.10

embik:
## Summary

What it says on the title. Little housekeeping update to the Go version we use.

## Related issue(s)

Fixes #

## Release Notes

```release-note
kcp is built with Go 1.22.10
```

PR #3213, opened 2024-12-10
[release-0.26] :seedling: Update to Go 1.22.10

embik:
## Summary

A manual backport of #3212. Since 0.27 is still out a bit, it would be nice to have some basic maintenance on the active release branch since we should get into the habit of patch releases.

## Related issue(s)

Fixes #

## Release Notes

```release-note
kcp is built with Go 1.22.10
```


xrstf (approved):
/approve

PR #3219, opened 2024-12-15
🐛 fix check `deleteError != nil` but return an other nil value `err`

alingse:
## Summary

it's a harmless bug, but may cause some bug in future.

the code check the `deleteError != nil` but return an other nil `err`
return a other nil `err` , but it was check the 

## Related issue(s)

https://github.com/alingse/sundrylint/issues/4

I have an idea to detect the return nilness error bug, and check the top 1000 github go repos, and found this.

Fixes #

## Release Notes

```release-note
NONE
```


embik (approved):
/approve

PR #3220, opened 2024-12-18
:sparkles:  Add prefix to dev images for gc

mjudeikis:
## Summary

This prefixes non-branch or tag images with `dev-` so we can start publishing them (follow-up pr). and later purge them

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/hack/build-image.sh
+++ b/hack/build-image.sh
@@ -40,8 +40,8 @@ fi
 repository=ghcr.io/kcp-dev/kcp
 architectures="amd64 arm64"
 
-# when building locally, just tag with the current HEAD hash with a "dev-" prefix for GC.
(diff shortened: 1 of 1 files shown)


hack/build-image.sh
```diff
@@ -60,14 +60,19 @@ if [ -n "${PULL_BASE_REF:-}" ]; then
   fi
 fi
 
+# Prefix with "dev-" if not on a tag or branch
+if [ -z "$branchName" ]; then
+  version="dev-$version"
```
sttts, line 65: double prefix? dev-dev-23hkj23n4

mjudeikis: not anymore.

PR #3223, opened 2024-12-18
:sparkles: Decouple cli module from main codebase

mjudeikis:
## Summary

Not sure why we had this replace. It same as not having a module. 

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


sttts:
> It same as not having a module.

Not completely true. The replace means that locally when building the cli, it would go directly to the directory. Without the replace it means it would follow the go.mod require version of the SDK.

When depending on the cli source code (which probably nobody does), it would ignore the replace.

So I am not really sure we actually want to remove it 🤔

mjudeikis:
If that is the case (TIL) - this is right

PR #3124, opened 2024-05-02
:bug: admission/webhooks: set kcp.io/cluster annotation on create

sttts:
## Summary

On webhook admission on create, it is unclear where the object comes from. This PR sets the `kcp.io/cluster` annotation on webhook dispatch.

## Release Notes

```release-note
Set the kcp.io/cluster annotation on objects passed to an admission webhook on create.
```

PR #3229, opened 2024-12-20
:bug: Fix undo function of cluster annotation setter wiping annotations

### LINKED ISSUE #3182: bug: Patched annotations by admission controller get removed by KCP
### Describe the bug

We have an admission controller that mutates our custom objects. The admission controller adds 1 annotation and 1 label with JsonPatch. Problem is that the label gets added successfully to the object but the annotation doesn't. I discovered that if I manually add a kcp.io/cluster annotation to the created object then our custom annotation also gets added successfully. I'm suspecting that the process which adds the kcp.io/cluster annotation. Note, annotations that have been manually added to the object aren't affected by this bug.

### Steps To Reproduce

1. Create a mutating admission controller that returns annotation patches. For example:
```go
pt := admission.PatchTypeJSONPatch
patch := []byte(`[{ "op": "add", "path": "/metadata/annotations/foo", "value": "bar" }]`)
responseAdmissionReview := &admission.AdmissionReview{}
...
responseAdmissionReview.Response = &admission.AdmissionResponse{Allowed: true, PatchType: &pt, Patch: patch}
...
```
2. Register the controller with MutatingWebhookConfiguration.
3. Create a custom object that matches the configured webhook rule.
4. Observe that the only annotations for the object are:
```yaml
kcp.io/cluster: 2go2ya0dhfrlvb9f
kubectl.kubernetes.io/last-applied-configuration: "..."
```
5. Delete the created object.
6. Manually add the kcp.io/cluster annotation to the object yaml, value has to match the actual cluster:
```yaml
annotations:
  kcp.io/cluster: 2go2ya0dhfrlvb9f
```
7. Re-create the object.
8. Observe that the custom foo annotation got added successfully:
```yaml
foo: bar
kcp.io/cluster: 2go2ya0dhfrlvb9f
kubectl.kubernetes.io/last-applied-configuration: "..."
```

### Expected Behaviour

Admission controller should be able to successfully add the custom annotation.

### Additional Context

I'm not 100% sure that this is caused by KCP but the admission controller does work as expected when used in a local kubernetes cluster without KCP. I have verified that the behaviour described above at least happens in our KCP.

KCP version: v0.26.0

ntnn:
## Summary

The undo function of the admission webhook wiped annotations if the cluster annotation was not present.

## Related issue(s)

Fixes #3182 introduced in #3124 

## Release Notes

```release-note
Fix external modifications to annotations being reverted by admission webhook
```

### DIFF
--- a/pkg/admission/validatingwebhook/plugin_test.go
+++ b/pkg/admission/validatingwebhook/plugin_test.go
@@ -1,0 +1,92 @@
+/*
+Copyright 2022 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package validatingwebhook
+
+import (
+	"testing"
+
+	"github.com/kcp-dev/logicalcluster/v3"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const clusterName = logicalcluster.Name("test-cluster")
+
+func TestSetClusterAnnotation(t *testing.T) {
+	tests := []struct {
+		name string
+		in   *corev1.ConfigMap
+	}{
+		{
+			name: "no annotations",
+			in: &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{},
+			},
+		},
+		{
+			name: "with annotations",
+			in: &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						"foo": "bar",
+					},
+				},
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			origAnnotations := test.in.GetAnnotations()
+
+			undo := SetClusterAnnotation(test.in, clusterName)
+			assert.NotNil(t, undo)
+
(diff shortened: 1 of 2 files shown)


embik:
/ok-to-test

Thank you for the contribution @ntnn! Can you please also add a release note to the PR description (something like "Stop removing annotations added by mutating webhooks" that describes what is fixed here)?

pkg/admission/validatingwebhook/plugin_test.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			origAnnotations := test.in.GetAnnotations()
+
+			undo := SetClusterAnnotation(test.in, clusterName)
+			assert.NotNil(t, undo)
+
+			require.NotNil(t, test.in.GetAnnotations())
```
stevekuznetsov, line 62: Zero-value of `GetAnnotations()` not being `nil` even though you don't set it to anything in an empty `ObjectMeta` is enforced by what? Doesn't Go semantic not do that? https://go.dev/play/p/vYJPy8IEjXq

ntnn: `SetClusterAnnotation` is supposed to set annotations if no annotations are set initially:
https://github.com/kcp-dev/kcp/pull/3229/files#diff-21219bad7e49f362d06f1668718121ef01005c2a0a8133188b31a32b9944b097R215

stevekuznetsov: Ah, I see - d'oh. The only difference I can see in your impl which might be unintended is that if originally annotations were `nil`, does your `undo` return it to `nil`? Does it matter?

ntnn: I was thinking about resetting to nil if the map would be left empty as well, but I think it doesn't matter.
On the wire it would be omitted either way looking into the json encoding:

https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/encoding/json/encode.go;l=73-76
https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/encoding/json/encode.go;l=704
https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/encoding/json/encode.go;l=307-310

stevekuznetsov: That was my suspicion, thank you.

mjudeikis:
/lgtm
/approve 

thanks for this!

PR #3232, opened 2024-12-21
:memo: Mount API POC

mjudeikis:
## Summary

Thinking how best to promote mounts into more native way of doing things in kcp.
Somehow logical steps in my head currently are:
1. Extend WorkspaceType with `Mount *metav1.TypeMeta `json:"mount,omitempty"`` telling which flavour it is.
Maybe even add "SubType: LogicalCluster,Mount"? But think this is already too much. If Mount is set, we expect everything else to be nil/not set. It would specific Kind for mount object, but not name.

2. Extend Workspace with `Mount *metav1.ObjectMeta `json:"mount,omitempty"`` where we specify what specific object we refer to. It would when be resolved into statuses as we do now. 

3. We would need to relax URL validation for `mount` types and basically use `spec.url` for mount URL if its mountpoint.

This is very much brain-dump to kick-off discussions. @sttts @embik 

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/sdk/apis/tenancy/v1alpha1/types_workspace.go
+++ b/sdk/apis/tenancy/v1alpha1/types_workspace.go
@@ -158,8 +158,9 @@ type WorkspaceSpec struct {
 	// MountName is the name of the mount that this workspace is mounted under.
 	//
 	// +optional
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="mountName is immutable"
-	MountName *string `json:"mountName,omitempty"`
+	// +kubebuilder:validation:XValidation:rule="!has(oldSelf.name) || !has(self.name) || self.name == oldSelf.name",message="name is immutable"
+	// +kubebuilder:validation:XValidation:rule="!has(oldSelf.namespace) || !has(self.namespace) || self.namespace == oldSelf.namespace",message="namespace is immutable"
+	Mount *metav1.ObjectMeta `json:"mount,omitempty"`
 
 	// location constraints where this workspace can be scheduled to.
 	//
--- a/sdk/apis/tenancy/v1alpha1/types_workspace.go
+++ b/sdk/apis/tenancy/v1alpha1/types_workspace.go
@@ -155,11 +155,17 @@ type WorkspaceSpec struct {
 	// +kubebuilder:validation:XValidation:rule="!has(oldSelf.path) || !has(self.path) || self.path == oldSelf.path",message="path is immutable"
 	Type WorkspaceTypeReference `json:"type,omitempty"`
 
+	// MountName is the name of the mount that this workspace is mounted under.
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="mountName is immutable"
+	MountName *string `json:"mountName,omitempty"`
+
 	// location constraints where this workspace can be scheduled to.
 	//
 	// If the no location is specified, an arbitrary location is chosen.
 	//
-	// +optional
+	// +optionalz``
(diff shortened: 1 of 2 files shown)


sdk/apis/tenancy/v1alpha1/types_workspace.go
```diff
@@ -155,11 +155,18 @@ type WorkspaceSpec struct {
 	// +kubebuilder:validation:XValidation:rule="!has(oldSelf.path) || !has(self.path) || self.path == oldSelf.path",message="path is immutable"
 	Type WorkspaceTypeReference `json:"type,omitempty"`
 
+	// MountName is the name of the mount that this workspace is mounted under.
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="!has(oldSelf.name) || !has(self.name) || self.name == oldSelf.name",message="name is immutable"
+	// +kubebuilder:validation:XValidation:rule="!has(oldSelf.namespace) || !has(self.namespace) || self.namespace == oldSelf.namespace",message="namespace is immutable"
+	Mount *metav1.ObjectMeta `json:"mount,omitempty"`
```
sttts, line 163: Doesn't this need some higher-level CEL rule to make it mutual exclusive and bound to type?

sdk/apis/tenancy/v1alpha1/types_workspacetype.go
```diff
@@ -116,6 +116,12 @@ type WorkspaceTypeSpec struct {
 	//
 	// +optional
 	DefaultAPIBindings []APIExportReference `json:"defaultAPIBindings,omitempty"`
+
+	// Mount is the mount configuration for the workspace type. Single type can have only one mount kind.
+	// If set, the workspace will be backed by a mount.
+	//
+	// +optional
+	Mount *metav1.TypeMeta `json:"mount,omitempty"`
```
sttts, line 124: how is this related to workspace types?

mjudeikis: This would be something like `kind: mounts.contrib.kcp.io` so every workspace initiated via this type just needs to carry name only for the kind. basically bounts workspaceType to mount implementation kind. Still thinking if this is a good idea. 

But if we keep Kind of mount object inside workspace spec, workspace Type is not even used.

maybe at the workspace level:
`Type WorkspaceTypeReference` should become with `Kind: Mount,LogicalCluster` ?

mjudeikis: something like:
```
spec:
  URL: https://192.168.3.55:6443/clusters/root:consumer
  cluster: ofebwh0c93o6swd3
  type:
    name: organization
    path: root
```

Would become:
```
spec:
  URL: https://192.168.3.55:6443/clusters/root:consumer
  cluster: ofebwh0c93o6swd3
  type:
    kind: WorkspaceType
    name: organization
    path: root
```

and 
```
spec:
  URL: https://192.168.3.55:6443/clusters/root:consumer
  cluster: ofebwh0c93o6swd3
  type:
    kind: Mount
    name: mountBob
    kind: mounts.contrib.kcp.io
    version: v1alpha1
```

sttts: In my mind type and mount are mutual exclusive, i.e. WorkspaceTypes are actually LogicalClusterTypes.

mjudeikis: So do you think its root level new field in Workspace? Any suggestions how this could be named considering type is taken? :) Or just `mount:`

PR #3243, opened 2025-01-09
:bug:  Add additional impersonation check

mjudeikis:
## Summary

It looks like you can do only groups impersonation, without users. So this should not allow this.

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/server/handler.go
+++ b/pkg/server/handler.go
@@ -283,10 +283,20 @@ func WithImpersonationGatekeeper(handler http.Handler) http.Handler {
 		// And impersonations is only allowed for the users, who have metadata in the ctx.
 		// Else just pass the request.
 		impersonationUser := req.Header.Get(authenticationv1.ImpersonateUserHeader)
-		if len(impersonationUser) == 0 {
+		impersonationGroups := req.Header[authenticationv1.ImpersonateGroupHeader]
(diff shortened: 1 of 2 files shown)


pkg/server/handler.go
```diff
@@ -283,18 +283,20 @@ func WithImpersonationGatekeeper(handler http.Handler) http.Handler {
 		// And impersonations is only allowed for the users, who have metadata in the ctx.
 		// Else just pass the request.
 		impersonationUser := req.Header.Get(authenticationv1.ImpersonateUserHeader)
-		if len(impersonationUser) == 0 {
+		impersonationGroups := req.Header[authenticationv1.ImpersonateGroupHeader]
+		if len(impersonationUser) == 0 && len(impersonationGroups) == 0 { // No user and group to impersonate - just pass the request.
```
sttts, line 287: extra as well

mjudeikis: how could one use this?

~~Scenario:~~
~~1. I set ONLY extras in the headers (no, groups, no user) - It will check for `validImpersonation` and will allow as groups are nil~~
~~2.  I set groups and extras - it will check it and normal logic applies.~~

Later on scopes and warrants need to be added here.

PR #3244, opened 2025-01-11
[MVP] Scopes. Not for merging

mjudeikis:
## Summary

Just need CI input. 

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


pkg/server/handler.go
```diff
@@ -305,6 +306,44 @@ func WithImpersonationGatekeeper(handler http.Handler) http.Handler {
...
+				apierrors.NewForbidden(schema.GroupResource{}, "", fmt.Errorf("no cluster found in the context")),
+				errorCodecs, schema.GroupVersion{}, w, req)
+			return
+		}
+
+		extra := user.GetExtra()
+		if extra == nil {
+			extra = map[string][]string{}
+		}
+		extra["authentication.kcp.io/scopes"] = append(extra["authentication.kcp.io/scopes"], fmt.Sprintf("cluster:%s", cluster.Name))
```
sttts, line 332: this is too much. We only have to do this if impersonation happened.

pkg/server/handler.go
```diff
@@ -305,6 +306,44 @@ func WithImpersonationGatekeeper(handler http.Handler) http.Handler {
...
+		handler.ServeHTTP(w, req.WithContext(request.WithUser(req.Context(), copyUserWithExtra(user, extra))))
+	})
+}
+
+// copyUserWithExtra returns a copy of the user with the extra field set to the provided value.
+func copyUserWithExtra(u user.Info, extra map[string][]string) user.Info {
```
sttts, line 338: not a fan of constructor wrapping. Would rather inline this. Makes it clearer.

pkg/server/handler.go
```diff
@@ -305,6 +306,44 @@ func WithImpersonationGatekeeper(handler http.Handler) http.Handler {
 	})
 }
 
+// WithScoping scopes the request to the cluster it is intended for.
+func WithScoping(handler http.Handler) http.Handler {
```
sttts, line 310: does this have to be public?

pkg/server/config.go
```diff
@@ -442,6 +442,12 @@ func NewConfig(ctx context.Context, opts kcpserveroptions.CompletedOptions) (*Co
 			apiHandler = WithVirtualWorkspacesProxy(apiHandler, shardVirtualWorkspaceURL, virtualWorkspaceServerProxyTransport, proxy)
 		}
 
+		// There is ordering here in play:
+		// 1. Default handlers up to impersonation gatekeeper preventing impersonation of the privileged user.
+		// 2. Rest of the handlers up to Authz
+		// 3. Scoping handlers to ensure that the request is scoped to the user's clusters before authz is done.
+		// 4. Rest of the handlers.
+		apiHandler = WithScoping(apiHandler)
```
sttts, line 450: we have to e2e test the effectiveness of this. It is kind of brittle.

pkg/server/handler.go
mjudeikis: Yes, was thinking that need to set some metadata for this and check it here.

PR #3059, opened 2023-12-12
✨  Add OpenAPI Schema support to virtual workspace framework

palnabarun:
## Summary

Add OpenAPI Schema support to virtual workspace framework.

The implementation taps into the `CreateServingInfo` path to generate the spec and adds a handler to VW apiserver for handling the request and updating the OpenAPI v3 service caches.

TODO: OpenAPI v2 support and structuring the code for merge.

## Related issue(s)

None

## Release Notes

```release-note
TODO: Before merge.
```

PR #3246, opened 2025-01-13
✨ Add OpenAPI v3 schema support to the Virtual Workspace framework

### LINKED ISSUE #3242: feature: OpenAPIv3 support in virtual workspace framework
### Feature Description

Virtual workspaces act as "virtual" Kubernetes-like API endpoints, but they do not support OpenAPI to discover API schemas. This makes virtual workspaces not feel like a "fully fledged" endpoint, and especially if users interact with kcp through virtual workspaces they might be missing the usual ergonomics of interacting with a Kubernetes API (e.g. `kubectl explain` won't work).

It would therefore be nice if OpenAPIv3 information was exposed by virtual workspaces.

### Proposed Solution

An initial solution was sketched out in #3059. The virtual workspace framework should be capable of generating and updating OpenAPIv3 information and serve it.

### Alternative Solutions

_No response_

### Want to contribute?

### Additional Context

_No response_

xmudrii:
## Summary

This PR continues work done in #3059. Huge thanks to @palnabarun for all the work on this feature so far!

This PR adds OpenAPI v3 schema support to the Virtual Workspace dynamic framework.

A new `openAPIHandler` has been implemented that utilizes the `kube-openapi` library to generate and serve the OpenAPI v3 specs. The specs are generated for the following routes:

- `/api` and `/apis` - information about API groups
- `/apis/<group>` - information about API versions
- `/api/v1` and `/apis/<group>/<version>` - specs for the concrete GroupVersion

Example calls:

```shell
# The root spec containing all other OpenAPI v3 URLs
k get --raw '/services/apiexport/<Cluster ID>/<APIExport Name>/clusters/<Cluster ID>/openapi/v3

# Information about API groups
k get --raw '/services/apiexport/<Cluster ID>/<APIExport Name>/clusters/<Cluster ID>/openapi/v3/apis

# Specs for APIs in stable.example.com/v1
k get --raw '/services/apiexport/<Cluster ID>/<APIExport Name>/clusters/<Cluster ID>/openapi/v3/stable.example.com/v1
```

Specs are generated for all `APIDefintions` available in that virtual workspace. In other words, everything that's served by [`GetAPIDefinitionSet` function](https://github.com/kcp-dev/kcp/blob/4c2d1061d8411e32adbe8184fdd694a5e830eaf9/pkg/virtual/framework/dynamic/apidefinition/types.go#L59-L62) is available in the OpenAPI v3 spec in that virtual workspace.

The implementation is a bit different to what the regular API servers are doing due to the way how are virtual workspaces implemented. The regular API servers are building the OpenAPI v3 specs from web services registered in that API server. However, this is not an option for the virtual workspaces API server because it only has three web services registered: `/api`, `/apis`, and `/version`. None of these three web services have any subroute registered, so there's no information about e.g. `/apis/<group>` or `/apis/<group>/<version>`. That makes generating the OpenAPI v3 specs using the "usual approach" impossible.

I mitigated that by:

- Generating specs for `/api`, `/apis/`, and `/version` from the respective web services
- Utilizing API discovery to generate web services for each available `/apis/<group>`, and then generate the OpenAPI v3 specs for these web services
- Converting the available `APIDefinitions` to `CustomResourceDefintions`, and then using the appropriate function from the `kube-openapi` library to generate the specs

At the end, all these specs are merged into one and served. This is heavily inspired by #3246

This is also a bit different than idea in #3059, as #3059 tried to manually implement each OpenAPI v3 route, while this heavily relies on the `kube-openapi` library.

To make this more performant, a simple caching solution has been implemented. I'm generating a hash out of the APIDefinition's spec, and then use it to generate a key. That key is used for the LRU cache where I store the OpenAPI v3 service. This is almost copy-paste of implementation from #3246.

As requested by the community, I implemented a simple way to disable the OpenAPI v3 generation by setting the `SkipOpenAPIInstallation` extra configuration option to `true`.

Finally, unlike the original implementation in #3059, this implementation doesn't tap into the `CreateServingInfoFor` function. Thinking about that, I think it would have been possible and it could have enabled more effective caching, but because of the complexity, I decided to go with a "simpler" initial implementation, and we can consider doing this in the future.

The only missing part are E2E tests. This is something that I plan on adding later, potentially as a follow up PR if that's okay.

## Related issue(s)

Fixes #3242

## Release Notes
```release-note
Add OpenAPI v3 schema support to the Virtual Workspace framework
```

### DIFF
--- a/pkg/virtual/framework/dynamic/apiserver/apiserver.go
+++ b/pkg/virtual/framework/dynamic/apiserver/apiserver.go
@@ -180,18 +180,21 @@ func (c completedConfig) New(virtualWorkspaceName string, delegationTarget gener
 
 	s.GenericAPIServer.Handler.GoRestfulContainer.Add(discovery.NewLegacyRootAPIHandler(c.GenericConfig.DiscoveryAddresses, s.GenericAPIServer.Serializer, "/api").WebService())
 
-	getOpenAPIDefinitions := openapi.GetOpenAPIDefinitionsWithoutDisabledFeatures(generatedopenapi.GetOpenAPIDefinitions)
-	namer := openapinamer.NewDefinitionNamer(scheme)
-	openapiv3Config := genericapiserver.DefaultOpenAPIV3Config(getOpenAPIDefinitions, namer)
-	openapiv3Config.Info.Title = "Kubernetes"
+	if !c.GenericConfig.SkipOpenAPIInstallation {
+		getOpenAPIDefinitions := openapi.GetOpenAPIDefinitionsWithoutDisabledFeatures(generatedopenapi.GetOpenAPIDefinitions)
+		namer := openapinamer.NewDefinitionNamer(scheme)
+		openapiv3Config := genericapiserver.DefaultOpenAPIV3Config(getOpenAPIDefinitions, namer)
+		openapiv3Config.Info.Title = "Kubernetes"
 
-	openAPIHandler := newOpenAPIHandler(s.APISetRetriever, s.GenericAPIServer.Handler.GoRestfulContainer, openapiv3Config, delegateHandler)
+		openAPIHandler := newOpenAPIHandler(s.APISetRetriever, s.GenericAPIServer.Handler.GoRestfulContainer, openapiv3Config, DefaultServiceCacheSize, delegateHandler)
+
+		s.GenericAPIServer.Handler.NonGoRestfulMux.Handle("/openapi", openAPIHandler)
+		s.GenericAPIServer.Handler.NonGoRestfulMux.HandlePrefix("/openapi/", openAPIHandler)
+	}
 
 	s.GenericAPIServer.Handler.NonGoRestfulMux.Handle("/apis", crdHandler)
 	s.GenericAPIServer.Handler.NonGoRestfulMux.HandlePrefix("/apis/", crdHandler)
 	s.GenericAPIServer.Handler.NonGoRestfulMux.Handle("/api/v1", crdHandler)
 	s.GenericAPIServer.Handler.NonGoRestfulMux.HandlePrefix("/api/v1/", crdHandler)
-	s.GenericAPIServer.Handler.NonGoRestfulMux.Handle("/openapi", openAPIHandler)
-	s.GenericAPIServer.Handler.NonGoRestfulMux.HandlePrefix("/openapi/", openAPIHandler)
 	return s, nil
 }
--- a/pkg/virtual/framework/dynamic/apiserver/apiserver.go
+++ b/pkg/virtual/framework/dynamic/apiserver/apiserver.go
@@ -29,7 +29,10 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apiserver/pkg/endpoints/discovery"
+	openapinamer "k8s.io/apiserver/pkg/endpoints/openapi"
 	genericapiserver "k8s.io/apiserver/pkg/server"
+	"k8s.io/apiserver/pkg/util/openapi"
+	generatedopenapi "k8s.io/kubernetes/pkg/generated/openapi"
 
 	virtualcontext "github.com/kcp-dev/kcp/pkg/virtual/framework/context"
 	"github.com/kcp-dev/kcp/pkg/virtual/framework/dynamic/apidefinition"
@@ -177,7 +180,12 @@ func (c completedConfig) New(virtualWorkspaceName string, delegationTarget gener
 
 	s.GenericAPIServer.Handler.GoRestfulContainer.Add(discovery.NewLegacyRootAPIHandler(c.GenericConfig.DiscoveryAddresses, s.GenericAPIServer.Serializer, "/api").WebService())
 
-	openAPIHandler := newOpenAPIHandler(s.APISetRetriever, delegateHandler)
+	getOpenAPIDefinitions := openapi.GetOpenAPIDefinitionsWithoutDisabledFeatures(generatedopenapi.GetOpenAPIDefinitions)
+	namer := openapinamer.NewDefinitionNamer(scheme)
+	openapiv3Config := genericapiserver.DefaultOpenAPIV3Config(getOpenAPIDefinitions, namer)
+	openapiv3Config.Info.Title = "Kubernetes"
+
+	openAPIHandler := newOpenAPIHandler(s.APISetRetriever, s.GenericAPIServer.Handler.GoRestfulContainer, openapiv3Config, delegateHandler)
 
 	s.GenericAPIServer.Handler.NonGoRestfulMux.Handle("/apis", crdHandler)
 	s.GenericAPIServer.Handler.NonGoRestfulMux.HandlePrefix("/apis/", crdHandler)
--- a/pkg/virtual/framework/dynamic/apiserver/apiserver.go
+++ b/pkg/virtual/framework/dynamic/apiserver/apiserver.go
@@ -177,14 +177,13 @@ func (c completedConfig) New(virtualWorkspaceName string, delegationTarget gener
 
 	s.GenericAPIServer.Handler.GoRestfulContainer.Add(discovery.NewLegacyRootAPIHandler(c.GenericConfig.DiscoveryAddresses, s.GenericAPIServer.Serializer, "/api").WebService())
 
+	openAPIHandler := newOpenAPIHandler(s.APISetRetriever, delegateHandler)
+
 	s.GenericAPIServer.Handler.NonGoRestfulMux.Handle("/apis", crdHandler)
 	s.GenericAPIServer.Handler.NonGoRestfulMux.HandlePrefix("/apis/", crdHandler)
 	s.GenericAPIServer.Handler.NonGoRestfulMux.Handle("/api/v1", crdHandler)
 	s.GenericAPIServer.Handler.NonGoRestfulMux.HandlePrefix("/api/v1/", crdHandler)
-
-	// TODO(david): plug OpenAPI if necessary. For now, according to the various virtual workspace use-cases,
-	// it doesn't seem necessary.
-	// Of course this requires using the --validate=false argument with some kubectl command like kubectl apply.
-
+	s.GenericAPIServer.Handler.NonGoRestfulMux.Handle("/openapi", openAPIHandler)
+	s.GenericAPIServer.Handler.NonGoRestfulMux.HandlePrefix("/openapi/", openAPIHandler)
 	return s, nil
 }
--- a/pkg/virtual/framework/dynamic/apiserver/openapi.go
+++ b/pkg/virtual/framework/dynamic/apiserver/openapi.go
@@ -41,6 +41,7 @@ import (
 	"k8s.io/kube-openapi/pkg/common/restfuladapter"
 	"k8s.io/kube-openapi/pkg/handler3"
 	"k8s.io/kube-openapi/pkg/spec3"
+	"k8s.io/utils/keymutex"
 	"k8s.io/utils/lru"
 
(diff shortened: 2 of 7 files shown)


sttts:
/retest

e2e-sharded flake real?

xmudrii:
@sttts I believe it's a flake, it's now green and I didn't touch anything related to that (I think). I'm now working on adding an E2E test for this feature too and it appears to be pretty stable (I still didn't push it, will likely do it tomorrow).

pkg/virtual/framework/dynamic/apiserver/openapi.go
```diff
@@ -0,0 +1,331 @@
...
+	// Get the OpenAPI service from cache or create a new one
+	key, err := apiConfigurationKey(apiSet)
+	if err != nil {
+		responsewriters.InternalError(w, req, err)
+		return
+	}
+	log = log.WithValues("key", key)
+	entry, ok := o.services.Get(key)
+	if !ok {
+		log.V(7).Info("Generating OpenAPI v3 specs")
```
sttts, line 127: this is a heavy operation, isn't it? Shouldn't we lock the saved key somehow and let other parties with the same interest wait? This even applies to one client collecting spec for example.

A common pattern for that: store an item struct in the lru cache with a wait channel:
```golang
type item struct {
  done <-chan struct{}
  err error
  service http.Handler
}
```
Every cache hit will wait for `done` before accessing `err` or `service`. The go routine owning this will add it to the lru cache before starting the heavy work. It will set `err` or `service` and then close `done` to wake all other waiters up.

sttts:
This needs an e2e test.

pkg/virtual/framework/dynamic/apiserver/openapi.go
```diff
@@ -0,0 +1,331 @@
...
+	"k8s.io/utils/lru"
+
+	"github.com/kcp-dev/kcp/pkg/virtual/framework/dynamic/apidefinition"
+	dyncamiccontext "github.com/kcp-dev/kcp/pkg/virtual/framework/dynamic/context"
+	apisv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/apis/v1alpha1"
+)
+
+const (
+	// DefaultServiceCacheSize is the default size of the OpenAPI service cache.
+	DefaultServiceCacheSize = 50
```
sttts, line 53: We could also have a ttl cache 🤔

pkg/virtual/framework/dynamic/apiserver/apiserver.go
```diff
@@ -177,14 +180,21 @@ func (c completedConfig) New(virtualWorkspaceName string, delegationTarget gener
 
 	s.GenericAPIServer.Handler.GoRestfulContainer.Add(discovery.NewLegacyRootAPIHandler(c.GenericConfig.DiscoveryAddresses, s.GenericAPIServer.Serializer, "/api").WebService())
 
+	if !c.GenericConfig.SkipOpenAPIInstallation {
```
embik, line 183: Just for my understanding here to match it with the PR description: Your intention was that developers of virtual workspaces that use the `virtual` framework can disable OpenAPI, not that kcp admins can disable OpenAPI on their default kcp VWs, right?

xmudrii: To put it the other way around, what do we actually want and/or need more? My understanding of this implementation is that it's more towards those who run virtual workspaces, i.e. kcp-admins. Do we want both options?

pkg/virtual/framework/dynamic/apiserver/openapi.go
xmudrii: I decided to go with the LRU cache for now as I wasn't able to find an existing TTL cache implementation.

pkg/virtual/framework/dynamic/apiserver/openapi.go
xmudrii: I implemented something based on that. I would very appreciate if you can take a look and let me know if it makes any sense.

pkg/virtual/framework/dynamic/apiserver/openapi.go
```diff
@@ -0,0 +1,383 @@
...
+	//      (that's why we have servicesLock)
+	o.servicesLock.LockKey(key)
+	entry, ok := o.services.Get(key)
+	var item *openAPIServiceItem
+
+	// If we found the struct, and the channel is closed, and the spec was generated successfully, serve the already
+	// generated spec, otherwise regenerate it.
+	if ok {
+		item = entry.(*openAPIServiceItem)
+		if item != nil {
```
sttts, line 154: can it be nil?

pkg/virtual/framework/dynamic/apiserver/openapi.go
```diff
@@ -0,0 +1,383 @@
...
+	}
+
+	item = &openAPIServiceItem{
+		done: make(chan struct{}, 1),
+	}
+	o.services.Add(key, item)
+	o.servicesLock.UnlockKey(key)
+
+	// If we got here, we have nothing in cache or cache is invalid, so we just generate everything again and serve it.
+	log.V(7).Info("Generating OpenAPI v3 specs")
```
sttts, line 172: would be interesting to print this at the end (with a defer) and the time it took.

pkg/virtual/framework/dynamic/apiserver/openapi.go
```diff
@@ -0,0 +1,383 @@
...
+				o.servicesLock.UnlockKey(key)
+				item.service.ServeHTTP(w, req)
+				return
+			}
+		}
+
+	}
+
+	item = &openAPIServiceItem{
+		done: make(chan struct{}, 1),
```
sttts, line 166:
```suggestion
	doneCh := make(chan struct{}),
	item = &openAPIServiceItem{
		done: doneCh,
```

pkg/virtual/framework/dynamic/apiserver/openapi.go
```diff
@@ -0,0 +1,383 @@
...
+	apisv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/apis/v1alpha1"
+)
+
+const (
+	// DefaultServiceCacheSize is the default size of the OpenAPI service cache.
+	DefaultServiceCacheSize = 50
+)
+
+type openAPIServiceItem struct {
+	done    chan struct{}
```
sttts, line 58:
```suggestion
	done    <- chan struct{}
```
Makes clear who closes it.

pkg/virtual/framework/dynamic/apiserver/openapi.go
xmudrii: I looked at `kubectl -v=6` and cold cache request takes up to 15 ms, cached request is around 9 ms

pkg/virtual/framework/dynamic/apiserver/openapi.go
xmudrii: I tried that but it throws:

> invalid operation: cannot close receive-only channel item.done (variable of type <-chan struct{})compiler[InvalidClose](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#InvalidClose)

pkg/virtual/framework/dynamic/apiserver/openapi.go
```diff
@@ -0,0 +1,383 @@
...
+				return
+			}
+		}
+
+	}
+
+	item = &openAPIServiceItem{
+		done: make(chan struct{}, 1),
+	}
+	o.services.Add(key, item)
```
sttts, line 168: missing removal when something goes wrong. Also it is brittle that the channel is never closed e.g. in case of panic. I would rather put the handling into a defer, such that it is always run. If you set the `err` with some placeholder from the beginning, you are sure it always sends the error to the other waiters.

pkg/virtual/framework/dynamic/apiserver/openapi.go
```diff
@@ -0,0 +1,383 @@
...
+	entry, ok := o.services.Get(key)
+	var item *openAPIServiceItem
+
+	// If we found the struct, and the channel is closed, and the spec was generated successfully, serve the already
+	// generated spec, otherwise regenerate it.
+	if ok {
+		item = entry.(*openAPIServiceItem)
+		if item != nil {
+			_, ok := <-item.done
+			if !ok && item.err == nil {
```
sttts, line 156: why the ok check? Why not just `<- item.done`?

xmudrii: I thought for some reason that `<-item.done` is blocking even if the channel is closed, but now that I tested it again, turns out it isn't blocking. I'll change that.

pkg/virtual/framework/dynamic/apiserver/openapi.go
xmudrii: > missing removal when something goes wrong.

Why do we need removal? If something goes wrong, the next caller will figure that out from `item.err` and it'll just generate a new item with a new channel. If anyone else is waiting on the old channel, they'll eventually pick up the new channel (because the old one will get closed, we have locks around getting the item, so we shouldn't end up in any race condition).

> Also it is brittle that the channel is never closed e.g. in case of panic. I would rather put the handling into a defer, such that it is always run.

I agree with that, I was thinking about that, but I didn't went with it because `defer` would run after serving, so it could add some overhead. But it's much more safer, I'll change that.

> If you set the `err` with some placeholder from the beginning, you are sure it always sends the error to the other waiters.

I'm not sure that I fully understand this, but I'll try something.

pkg/virtual/framework/dynamic/apiserver/openapi.go
xmudrii: Removed

pkg/virtual/framework/dynamic/apiserver/openapi.go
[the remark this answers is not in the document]
xmudrii: Removed

xmudrii: Changed

pkg/virtual/framework/dynamic/apiserver/openapi.go
sttts: > and it'll just generate a new item with a new channel.

Sounds ok. Is it tested?

sttts: > but I didn't went with it because defer would run after serving

Put the protected part into a closure. The serving code outside.

xmudrii: @sttts This should be addressed now, PTAL when you can and let me know if it's okay. The E2E test is flaking, I'm investigating that at the moment.

xmudrii:
/hold
The E2E test is flaking and not working in the sharded setup.

xmudrii:
I'm testing the fix for flakiness. I'm not yet sure this is the fix we want, but I want to check if it's effective.
/test all

pkg/virtual/framework/dynamic/apiserver/apiserver.go
embik: I would say for now we should leave it to the implementors, not the admins. If someone wants to make it optional in their virtual workspace, they can add a cmd flag exposing this setting.

pkg/virtual/framework/dynamic/apiserver/openapi.go
embik: Problem here is that you are doing `close(item.done)` later on I suppose. For @sttts' suggestion, I suppose you could close `doneCh`, not `item.done`, in the `defer` block in the later part of the code.

pkg/virtual/framework/dynamic/apiserver/openapi.go
xmudrii: Added logging for requests that take too long (over 1 second)

embik (approved):
/approve

Leaving lifting the hold to @sttts.

Issue #3221, opened 2024-12-18
feature: APIExports VirtualWorkspace without consumer enhancement 


mjudeikis:
### Feature Description

Currently, if you create APIExport, until there is a first consumer (APIBinding), the endpoint is not usable. This means that there are no CRDs, informers can't start and the controllers will not work.

This leads to behaviour where platforms engineers are creating hacks around this by creating fake apibindings to enable controllers to start.
platform engineers create hacks around this by creating fake API bindings

This is not nice DevExerience and we should fix this. 

### Proposed Solution

We need to investigate:
> fake an endpoint until there actually is a CRD. We can decide whether it is technically possible to adopt the existing watches from the real storage, or cut them off.

### Want to contribute?

- [ ] I would like to work on this issue.

### Additional Context

For more info see slack: https://kubernetes.slack.com/archives/C021U8WSAFK/p1734437782238829

PR #3256, opened 2025-01-18
:sparkles: make APIExportEndpointSlice consumer aware

mjudeikis:
## Summary

This part makes APIExportEndpointSlices consumers aware. Meaning it will add or remove URLs if consumers are present or goes away.

In this PR:
1. Current APIExportEndpointSlice reconcile is split into 2 - `status` and `url`

`status` - status controller checks if API export is valid, partitions are set right, and it can be queried. Most of these objects operate in a single-shard context, so this reconciliation works only in a single-shard context. CAVITE: Based on the partition setup - the shards involved will be different. Hence we store the final computed selector into status for the `URL` reconciler to pick up. 

`url` - URL reconcile operates cross-shards setup. It takes `APIExportEndpointSlice` from the cache and local shard and observes `APIBindings` and `APIExports` and, based on consumers, will update URLs. 

In addition, it uses not a commuter but Server-Side Apply patch to update only its own field, which is owned by shard. 

If we want to remove the selector from status - it would require an overload single reconciliation quite a lot. And filter when we can update statuses and when not. Overall, I tried, and the code readability was very poor. 

## Related issue(s)

Fixes # https://github.com/kcp-dev/kcp/issues/3221

## Release Notes

```release-note
make APIExportEndpointSlices consumer aware
```

### DIFF
--- a/pkg/indexers/apiexport.go
+++ b/pkg/indexers/apiexport.go
@@ -17,6 +17,8 @@ limitations under the License.
 package indexers
 
 import (
+	"fmt"
+
 	kcpcache "github.com/kcp-dev/apimachinery/v2/pkg/cache"
 	"github.com/kcp-dev/logicalcluster/v3"
 
@@ -33,6 +35,8 @@ const (
 	// APIExportByClaimedIdentities is the indexer name for retrieving APIExports that have a permission claim for a
 	// particular identity hash.
 	APIExportByClaimedIdentities = "APIExportByClaimedIdentities"
+	// APIExportEndpointSliceByAPIExport is the indexer name for retrieving APIExportEndpointSlices by their APIExport's Reference Path and Name.
+	APIExportEndpointSliceByAPIExport = "APIExportEndpointSliceByAPIExport"
 )
 
 // IndexAPIExportByIdentity is an index function that indexes an APIExport by its identity hash.
@@ -71,4 +75,24 @@ func IndexAPIExportByClaimedIdentities(obj interface{}) ([]string, error) {
 		claimedIdentities.Insert(claim.IdentityHash)
 	}
 	return sets.List[string](claimedIdentities), nil
+}
+
+// IndexAPIExportEndpointSliceByAPIExportFunc indexes the APIExportEndpointSlice by their APIExport's Reference Path and Name.
+func IndexAPIExportEndpointSliceByAPIExport(obj interface{}) ([]string, error) {
+	apiExportEndpointSlice, ok := obj.(*apisv1alpha1.APIExportEndpointSlice)
+	if !ok {
+		return []string{}, fmt.Errorf("obj %T is not an APIExportEndpointSlice", obj)
+	}
+
+	var result []string
+	pathRemote := logicalcluster.NewPath(apiExportEndpointSlice.Spec.APIExport.Path)
+	if !pathRemote.Empty() {
+		result = append(result, pathRemote.Join(apiExportEndpointSlice.Spec.APIExport.Name).String())
+	}
+	pathLocal := logicalcluster.From(apiExportEndpointSlice).Path()
+	if !pathLocal.Empty() {
+		result = append(result, pathLocal.Join(apiExportEndpointSlice.Spec.APIExport.Name).String())
+	}
+
+	return result, nil
 }
--- a/pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
+++ b/pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
@@ -19,14 +19,15 @@ package apiexportendpointslice
 import (
 	"context"
 	"fmt"
-	"reflect"
 	"time"
 
+	"github.com/go-logr/logr"
 	kcpcache "github.com/kcp-dev/apimachinery/v2/pkg/cache"
 	"github.com/kcp-dev/logicalcluster/v3"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -41,12 +42,10 @@ import (
 	"github.com/kcp-dev/kcp/pkg/reconciler/events"
 	apisv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/apis/v1alpha1"
 	"github.com/kcp-dev/kcp/sdk/apis/core"
-	corev1alpha1 "github.com/kcp-dev/kcp/sdk/apis/core/v1alpha1"
 	topologyv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/topology/v1alpha1"
 	kcpclientset "github.com/kcp-dev/kcp/sdk/client/clientset/versioned/cluster"
 	apisv1alpha1client "github.com/kcp-dev/kcp/sdk/client/clientset/versioned/typed/apis/v1alpha1"
-	apisinformers "github.com/kcp-dev/kcp/sdk/client/informers/externalversions/apis/v1alpha1"
-	corev1alpha1informers "github.com/kcp-dev/kcp/sdk/client/informers/externalversions/core/v1alpha1"
+	apisv1alpha1informers "github.com/kcp-dev/kcp/sdk/client/informers/externalversions/apis/v1alpha1"
 	topologyinformers "github.com/kcp-dev/kcp/sdk/client/informers/externalversions/topology/v1alpha1"
 )
 
@@ -57,9 +56,8 @@ const (
 // NewController returns a new controller for APIExportEndpointSlices.
 // Shards and APIExports are read from the cache server.
 func NewController(
-	apiExportEndpointSliceClusterInformer apisinformers.APIExportEndpointSliceClusterInformer,
-	globalShardClusterInformer corev1alpha1informers.ShardClusterInformer,
-	globalAPIExportClusterInformer apisinformers.APIExportClusterInformer,
+	apiExportEndpointSliceClusterInformer apisv1alpha1informers.APIExportEndpointSliceClusterInformer,
+	globalAPIExportClusterInformer apisv1alpha1informers.APIExportClusterInformer,
 	partitionClusterInformer topologyinformers.PartitionClusterInformer,
 	kcpClusterClient kcpclientset.ClusterInterface,
 ) (*controller, error) {
@@ -73,11 +71,8 @@ func NewController(
 		listAPIExportEndpointSlices: func() ([]*apisv1alpha1.APIExportEndpointSlice, error) {
 			return apiExportEndpointSliceClusterInformer.Lister().List(labels.Everything())
 		},
-		listShards: func(selector labels.Selector) ([]*corev1alpha1.Shard, error) {
-			return globalShardClusterInformer.Lister().List(selector)
-		},
-		getAPIExportEndpointSlice: func(clusterName logicalcluster.Name, name string) (*apisv1alpha1.APIExportEndpointSlice, error) {
-			return apiExportEndpointSliceClusterInformer.Lister().Cluster(clusterName).Get(name)
+		getAPIExportEndpointSlice: func(path logicalcluster.Path, name string) (*apisv1alpha1.APIExportEndpointSlice, error) {
+			return indexers.ByPathAndName[*apisv1alpha1.APIExportEndpointSlice](apisv1alpha1.Resource("apiexportendpointslices"), apiExportEndpointSliceClusterInformer.Informer().GetIndexer(), path, name)
 		},
 		getAPIExport: func(path logicalcluster.Path, name string) (*apisv1alpha1.APIExport, error) {
 			return indexers.ByPathAndName[*apisv1alpha1.APIExport](apisv1alpha1.Resource("apiexports"), globalAPIExportClusterInformer.Informer().GetIndexer(), path, name)
@@ -104,50 +99,38 @@ func NewController(
 		commit:                                committer.NewCommitter[*APIExportEndpointSlice, Patcher, *APIExportEndpointSliceSpec, *APIExportEndpointSliceStatus](kcpClusterClient.ApisV1alpha1().APIExportEndpointSlices()),
 	}
 
+	logger := logging.WithReconciler(klog.Background(), ControllerName)
+
 	_, _ = apiExportEndpointSliceClusterInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
-			c.enqueueAPIExportEndpointSlice(obj)
+			c.enqueueAPIExportEndpointSlice(objOrTombstone[*apisv1alpha1.APIExportEndpointSlice](obj), logger, "")
 		},
 		UpdateFunc: func(_, newObj interface{}) {
-			c.enqueueAPIExportEndpointSlice(newObj)
+			c.enqueueAPIExportEndpointSlice(objOrTombstone[*apisv1alpha1.APIExportEndpointSlice](newObj), logger, "")
 		},
 		DeleteFunc: func(obj interface{}) {
-			c.enqueueAPIExportEndpointSlice(obj)
+			c.enqueueAPIExportEndpointSlice(objOrTombstone[*apisv1alpha1.APIExportEndpointSlice](obj), logger, "")
 		},
 	})
 
 	_, _ = globalAPIExportClusterInformer.Informer().AddEventHandler(events.WithoutSyncs(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
-			c.enqueueAPIExportEndpointSlicesForAPIExport(obj)
+			c.enqueueAPIExportEndpointSlicesForAPIExport(objOrTombstone[*apisv1alpha1.APIExport](obj), logger)
 		},
 		DeleteFunc: func(obj interface{}) {
-			c.enqueueAPIExportEndpointSlicesForAPIExport(obj)
-		},
-	}))
-
-	_, _ = globalShardClusterInformer.Informer().AddEventHandler(events.WithoutSyncs(cache.ResourceEventHandlerFuncs{
-		AddFunc: func(obj interface{}) {
-			c.enqueueAllAPIExportEndpointSlices(obj)
-		},
-		UpdateFunc: func(oldObj, newObj interface{}) {
-			if filterShardEvent(oldObj, newObj) {
-				c.enqueueAllAPIExportEndpointSlices(newObj)
-			}
-		},
-		DeleteFunc: func(obj interface{}) {
-			c.enqueueAllAPIExportEndpointSlices(obj)
+			c.enqueueAPIExportEndpointSlicesForAPIExport(objOrTombstone[*apisv1alpha1.APIExport](obj), logger)
 		},
 	}))
 
 	_, _ = partitionClusterInformer.Informer().AddEventHandler(events.WithoutSyncs(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
-			c.enqueuePartition(obj)
+			c.enqueuePartition(objOrTombstone[*topologyv1alpha1.Partition](obj), logger)
 		},
 		UpdateFunc: func(_, newObj interface{}) {
-			c.enqueuePartition(newObj)
+			c.enqueuePartition(objOrTombstone[*topologyv1alpha1.Partition](newObj), logger)
 		},
 		DeleteFunc: func(obj interface{}) {
-			c.enqueuePartition(obj)
+			c.enqueuePartition(objOrTombstone[*topologyv1alpha1.Partition](obj), logger)
 		},
 	}))
 
@@ -166,45 +149,34 @@ type CommitFunc = func(context.Context, *Resource, *Resource) error
 type controller struct {
 	queue workqueue.TypedRateLimitingInterface[string]
 
-	listShards                            func(selector labels.Selector) ([]*corev1alpha1.Shard, error)
 	listAPIExportEndpointSlices           func() ([]*apisv1alpha1.APIExportEndpointSlice, error)
-	getAPIExportEndpointSlice             func(clusterName logicalcluster.Name, name string) (*apisv1alpha1.APIExportEndpointSlice, error)
+	getAPIExportEndpointSlice             func(path logicalcluster.Path, name string) (*apisv1alpha1.APIExportEndpointSlice, error)
 	getAPIExport                          func(path logicalcluster.Path, name string) (*apisv1alpha1.APIExport, error)
 	getPartition                          func(clusterName logicalcluster.Name, name string) (*topologyv1alpha1.Partition, error)
 	getAPIExportEndpointSlicesByPartition func(key string) ([]*apisv1alpha1.APIExportEndpointSlice, error)
 
-	apiExportEndpointSliceClusterInformer apisinformers.APIExportEndpointSliceClusterInformer
+	apiExportEndpointSliceClusterInformer apisv1alpha1informers.APIExportEndpointSliceClusterInformer
 	commit                                CommitFunc
 }
 
 // enqueueAPIExportEndpointSlice enqueues an APIExportEndpointSlice.
-func (c *controller) enqueueAPIExportEndpointSlice(obj interface{}) {
-	key, err := kcpcache.DeletionHandlingMetaClusterNamespaceKeyFunc(obj)
+func (c *controller) enqueueAPIExportEndpointSlice(slice *apisv1alpha1.APIExportEndpointSlice, logger logr.Logger, logSuffix string) {
+	key, err := kcpcache.DeletionHandlingMetaClusterNamespaceKeyFunc(slice)
 	if err != nil {
 		utilruntime.HandleError(err)
 		return
 	}
 
-	logger := logging.WithQueueKey(logging.WithReconciler(klog.Background(), ControllerName), key)
-	logger.V(4).Info("queueing APIExportEndpointSlice")
+	logger.V(4).Info(fmt.Sprintf("queueing APIExportEndpointSlice%s", logSuffix))
 	c.queue.Add(key)
 }
 
 // enqueueAPIExportEndpointSlicesForAPIExport enqueues APIExportEndpointSlices referencing a specific APIExport.
-func (c *controller) enqueueAPIExportEndpointSlicesForAPIExport(obj interface{}) {
-	if d, ok := obj.(cache.DeletedFinalStateUnknown); ok {
-		obj = d.Obj
-	}
-	export, ok := obj.(*apisv1alpha1.APIExport)
-	if !ok {
-		utilruntime.HandleError(fmt.Errorf("obj is supposed to be a APIExport, but is %T", obj))
-		return
-	}
-
+func (c *controller) enqueueAPIExportEndpointSlicesForAPIExport(export *apisv1alpha1.APIExport, logger logr.Logger) {
 	// binding keys by full path
 	keys := sets.New[string]()
 	if path := logicalcluster.NewPath(export.Annotations[core.LogicalClusterPathAnnotationKey]); !path.Empty() {
-		pathKeys, err := c.apiExportEndpointSliceClusterInformer.Informer().GetIndexer().IndexKeys(indexAPIExportEndpointSliceByAPIExport, path.Join(export.Name).String())
+		pathKeys, err := c.apiExportEndpointSliceClusterInformer.Informer().GetIndexer().IndexKeys(indexers.APIExportEndpointSliceByAPIExport, path.Join(export.Name).String())
 		if err != nil {
 			utilruntime.HandleError(err)
 			return
@@ -212,7 +184,7 @@ func (c *controller) enqueueAPIExportEndpointSlicesForAPIExport(obj interface{})
 		keys.Insert(pathKeys...)
 	}
 
-	clusterKeys, err := c.apiExportEndpointSliceClusterInformer.Informer().GetIndexer().IndexKeys(indexAPIExportEndpointSliceByAPIExport, logicalcluster.From(export).Path().Join(export.Name).String())
+	clusterKeys, err := c.apiExportEndpointSliceClusterInformer.Informer().GetIndexer().IndexKeys(indexers.APIExportEndpointSliceByAPIExport, logicalcluster.From(export).Path().Join(export.Name).String())
 	if err != nil {
 		utilruntime.HandleError(err)
 		return
@@ -227,35 +199,12 @@ func (c *controller) enqueueAPIExportEndpointSlicesForAPIExport(obj interface{})
 		} else if !exists {
 			continue
 		}
-		logger := logging.WithObject(logging.WithReconciler(klog.Background(), ControllerName), obj.(*apisv1alpha1.APIExport))
-		logging.WithQueueKey(logger, key).V(4).Info("queuing APIExportEndpointSlices because of referenced APIExport")
-		c.enqueueAPIExportEndpointSlice(slice)
-	}
-}
-
-// enqueueAllAPIExportEndpointSlices enqueues all APIExportEndpointSlices.
-func (c *controller) enqueueAllAPIExportEndpointSlices(shard interface{}) {
-	list, err := c.listAPIExportEndpointSlices()
-	if err != nil {
-		utilruntime.HandleError(err)
-		return
-	}
-
-	logger := logging.WithObject(logging.WithReconciler(klog.Background(), ControllerName), shard.(*corev1alpha1.Shard))
-	for i := range list {
-		key, err := kcpcache.DeletionHandlingMetaClusterNamespaceKeyFunc(list[i])
-		if err != nil {
-			utilruntime.HandleError(err)
-			continue
-		}
-
-		logging.WithQueueKey(logger, key).V(4).Info("queuing APIExportEndpointSlice because Shard changed")
-		c.queue.Add(key)
+		c.enqueueAPIExportEndpointSlice(objOrTombstone[*apisv1alpha1.APIExportEndpointSlice](slice), logger, " because of referenced APIExport")
 	}
 }
 
 // enqueuePartition maps a Partition to APIExportEndpointSlices for enqueuing.
-func (c *controller) enqueuePartition(obj interface{}) {
+func (c *controller) enqueuePartition(obj *topologyv1alpha1.Partition, logger logr.Logger) {
 	key, err := kcpcache.DeletionHandlingMetaClusterNamespaceKeyFunc(obj)
 	if err != nil {
 		utilruntime.HandleError(err)
@@ -268,10 +217,8 @@ func (c *controller) enqueuePartition(obj interface{}) {
 		return
 	}
 
-	logger := logging.WithObject(logging.WithReconciler(klog.Background(), ControllerName), obj.(*topologyv1alpha1.Partition))
-	logging.WithQueueKey(logger, key).V(4).Info("queuing APIExportEndpointSlices because Partition changed")
 	for _, slice := range slices {
-		c.enqueueAPIExportEndpointSlice(slice)
+		c.enqueueAPIExportEndpointSlice(objOrTombstone[*apisv1alpha1.APIExportEndpointSlice](slice), logger, " because of Partition change")
 	}
 }
 
@@ -328,7 +275,7 @@ func (c *controller) process(ctx context.Context, key string) error {
 		utilruntime.HandleError(err)
 		return nil
 	}
-	obj, err := c.getAPIExportEndpointSlice(clusterName, name)
+	obj, err := c.getAPIExportEndpointSlice(clusterName.Path(), name)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			return nil // object deleted before we handled it
@@ -353,6 +300,7 @@ func (c *controller) process(ctx context.Context, key string) error {
 	// If the object being reconciled changed as a result, update it.
 	oldResource := &Resource{ObjectMeta: old.ObjectMeta, Spec: &old.Spec, Status: &old.Status}
 	newResource := &Resource{ObjectMeta: obj.ObjectMeta, Spec: &obj.Spec, Status: &obj.Status}
+
 	if err := c.commit(ctx, oldResource, newResource); err != nil {
 		errs = append(errs, err)
 	}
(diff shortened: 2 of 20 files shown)


pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -57,11 +60,14 @@ const (
 // NewController returns a new controller for APIExportEndpointSlices.
 // Shards and APIExports are read from the cache server.
 func NewController(
+	shandName string,
```
sttts, line 63: shard

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -172,22 +194,54 @@ type controller struct {
 	getAPIExport                          func(path logicalcluster.Path, name string) (*apisv1alpha1.APIExport, error)
 	getPartition                          func(clusterName logicalcluster.Name, name string) (*topologyv1alpha1.Partition, error)
 	getAPIExportEndpointSlicesByPartition func(key string) ([]*apisv1alpha1.APIExportEndpointSlice, error)
+	getShardName                          func() string
```
sttts, line 197: why does this need a function?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -353,11 +413,16 @@ func (c *controller) process(ctx context.Context, key string) error {
...
 	newResource := &Resource{ObjectMeta: obj.ObjectMeta, Spec: &obj.Spec, Status: &obj.Status}
-	if err := c.commit(ctx, oldResource, newResource); err != nil {
-		errs = append(errs, err)
+
+	// TODO: Maybe we can use different commiters? When is local or remote? Or we dont care on performance here yet?
+	if err := c.commitExternal(ctx, oldResource, newResource); err != nil {
+		if errors.IsConflict(err) {
+			logger.V(4).Info("requeueing APIExportEndpointSlice because of conflict", err)
+			return true, nil
+		}
```
sttts, line 422: why does it matter? Different client?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
```diff
@@ -18,237 +18,74 @@ package apiexportendpointslice
...
-	"path"
 
 	"github.com/kcp-dev/logicalcluster/v3"
 
-	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/util/sets"
-	"k8s.io/klog/v2"
+	utilserrors "k8s.io/apimachinery/pkg/util/errors"
```
sttts, line 25: utilerrors

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
```diff
@@ -18,237 +18,74 @@ package apiexportendpointslice
...
-		getAPIExport: c.getAPIExport,
-		getPartition: c.getPartition,
-	}
-
-	return r.reconcile(ctx, apiExportEndpointSlice)
+type endpointsUrlsReconciler struct {
+	listShards   func(selector labels.Selector) ([]*corev1alpha1.Shard, error)
+	getAPIExport func(path logicalcluster.Path, name string) (*apisv1alpha1.APIExport, error)
+	getPartition func(clusterName logicalcluster.Name, name string) (*topologyv1alpha1.Partition, error)
+	getShardName func() string
```
sttts, line 43: as before, this is constant. No need for a function.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
mjudeikis: this was try to see why its failing. will remove. Conflicts where due to cache vs real object

pkg/server/controllers.go
```diff
@@ -1533,40 +1578,63 @@ func (s *Server) WaitForSync(stop <-chan struct{}) error {
 
 // addIndexerstoInformers is separated out from controllers as the re-election calls for controller re-initialization,
 // it would panics in indexer addition to informers as they are already started at bootup.
-func (s *Server) addIndexersToInformers(ctx context.Context) map[schema.GroupVersionResource]replication.ReplicatedGVR {
-	permissionclaimlabel.InstallIndexers(s.KcpSharedInformerFactory.Apis().V1alpha1().APIExports(), s.KcpSharedInformerFactory.Apis().V1alpha1().APIBindings())
-	permissionclaimlabler.InstallIndexers(s.KcpSharedInformerFactory.Apis().V1alpha1().APIExports())
+func (s *Server) addIndexersToInformers(_ context.Context) map[schema.GroupVersionResource]replication.ReplicatedGVR {
```
mjudeikis, line 1581: noop. just readability thing.

sdk/apis/apis/v1alpha1/types_apiexportendpointslice.go
```diff
@@ -73,10 +73,19 @@ type APIExportEndpointSliceStatus struct {
...
+	APIExportEndpoints []APIExportEndpoint `json:"endpoints"`
 
-	// endpoints contains all the URLs of the APIExport service.
-	APIExportEndpoints []APIExportEndpoint `json:"endpoints,omitempty"`
+	// +optional
+
+	// shardSelector is the selector used to filter the shards. It is used to filter the shards
+	// when determining partition scope when deriving the endpoints. This is set by owning shard,
+	// and is used by follower shards to determine if its inscope or not.
+	ShardSelector string `json:"shardSelector,omitempty"`
```
mjudeikis, line 88: This is needed so we can cleanly split reconcilers. One uses committer, another - SSA client. Keeping them together gets very much non-readable code.

cmd/sharded-test-server/main.go
```diff
@@ -320,7 +320,7 @@ func start(proxyFlags, shardFlags []string, logDirPath, workDirPath string, numb
 		if i >= len(regions) {
 			break
 		}
-		patch := fmt.Sprintf(`{"metadata":{"labels":{"region":%q}}}`, regions[i])
+		patch := fmt.Sprintf(`{"metadata":{"labels":{"region":%q,"shared": "true"}}}`, regions[i])
```
mjudeikis, line 323: to have unified selector for shards to do partition testing

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -237,15 +277,15 @@ func (c *controller) enqueueAPIExportEndpointSlicesForAPIExport(obj interface{})
 func (c *controller) enqueueAllAPIExportEndpointSlices(shard interface{}) {
 	list, err := c.listAPIExportEndpointSlices()
 	if err != nil {
-		runtime.HandleError(err)
+		utilruntime.HandleError(err)
```
sttts, line 280: please do this kind of changes always in a separate commit.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -57,13 +57,16 @@ const (
 // NewController returns a new controller for APIExportEndpointSlices.
 // Shards and APIExports are read from the cache server.
 func NewController(
-	apiExportEndpointSliceClusterInformer apisinformers.APIExportEndpointSliceClusterInformer,
+	shardName string,
+	apiExportEndpointSliceClusterInformer apisv1alpha1informers.APIExportEndpointSliceClusterInformer,
+	apiBindingInformer apisv1alpha1informers.APIBindingClusterInformer,
```
sttts, line 62: is this needed in the non-url controller?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -164,24 +202,26 @@ type CommitFunc = func(context.Context, *Resource, *Resource) error
 // controller reconciles APIExportEndpointSlices. It ensures that the shard endpoints are populated
 // in the status of every APIExportEndpointSlices.
 type controller struct {
-	queue workqueue.TypedRateLimitingInterface[string]
+	queue     workqueue.TypedRateLimitingInterface[string]
+	shardName string
```
sttts, line 206: still needed?

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointslice_reconcile.go
```diff
@@ -0,0 +1,154 @@
+/*
```
sttts, line 1: filename should match package (for logs....)

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,496 @@
...
+			obj, err := indexers.ByPathAndNameWithFallback[*apisv1alpha1.APIExportEndpointSlice](apisv1alpha1.Resource("apiexportendpointslices"), apiExportEndpointSliceClusterInformer.Informer().GetIndexer(), globalAPIExportEndpointSliceClusterInformer.Informer().GetIndexer(), path, name)
+			if err != nil {
+				return nil, err
+			}
+			return obj, err
+		},
+		getAPIExport: func(path logicalcluster.Path, name string) (*apisv1alpha1.APIExport, error) {
+			return indexers.ByPathAndName[*apisv1alpha1.APIExport](apisv1alpha1.Resource("apiexports"), globalAPIExportClusterInformer.Informer().GetIndexer(), path, name)
+		},
+		listAPIBindingsByAPIExport: func(export *apisv1alpha1.APIExport) ([]*apisv1alpha1.APIBinding, error) {
```
sttts, line 91: see how we add the indexes elsewhere. It is easier to read the code when they are separate, maybe even in a file like indexes.go.

sttts: nvmd. Misread.

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,496 @@
...
+			keys := sets.New[string]()
+			if path := logicalcluster.NewPath(export.Annotations[core.LogicalClusterPathAnnotationKey]); !path.Empty() {
+				pathKeys, err := apiBindingInformer.Informer().GetIndexer().IndexKeys(indexers.APIBindingsByAPIExport, path.Join(export.Name).String())
+				if err != nil {
+					return nil, err
+				}
+				keys.Insert(pathKeys...)
+			}
+
+			clusterKeys, err := apiBindingInformer.Informer().GetIndexer().IndexKeys(indexers.APIBindingsByAPIExport, logicalcluster.From(export).Path().Join(export.Name).String())
```
sttts, line 102: there is a helper `ByIndex` or so.

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,496 @@
...
+			c.enqueueAPIExportEndpointSlicesForAPIExport(obj)
+		},
+		DeleteFunc: func(obj interface{}) {
+			c.enqueueAPIExportEndpointSlicesForAPIExport(obj)
+		},
+	}))
+
+	_, _ = globalShardClusterInformer.Informer().AddEventHandler(events.WithoutSyncs(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			c.enqueueAllAPIExportEndpointSlices(obj)
```
sttts, line 173: why does this controller care about other shards?

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,496 @@
...
+		if path := logicalcluster.NewPath(binding.Spec.Reference.Export.Path); !path.Empty() {
+			pathKeys, err := c.globalApiExportEndpointSliceClusterInformer.Informer().GetIndexer().IndexKeys(indexAPIExportEndpointSliceByAPIExport, path.Join(binding.Spec.Reference.Export.Name).String())
+			if err != nil {
+				utilruntime.HandleError(err)
+				return
+			}
+			keys.Insert(pathKeys...)
+		}
+
+		for _, key := range sets.List[string](keys) {
```
sttts, line 250: isn't `sets.List` ordering? We don't need an order.

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointslice_reconcile.go
```diff
@@ -0,0 +1,154 @@
...
+func (r *endpointsReconciler) updateEndpoints(ctx context.Context,
+	apiExportEndpointSlice *apisv1alpha1.APIExportEndpointSlice,
+	apiExport *apisv1alpha1.APIExport,
+	shards []*corev1alpha1.Shard) (*result, error) {
+	logger := klog.FromContext(ctx)
+	var rs result
+	for _, shard := range shards {
+		if shard.Name != r.shardName {
+			continue
+		}
```
sttts, line 107: this is strange. Why watch shards? Why loop here? We know which shard we are.

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointslice_reconcile.go
```diff
@@ -0,0 +1,154 @@
...
+		if shard.Spec.VirtualWorkspaceURL == "" {
+			continue
+		}
+
+		// Check if we have local consumers
+		bindings, err := r.listAPIBindingsByAPIExport(apiExport)
```
sttts, line 113: you don't care about the actual bindings, only about whether there is one, or at best the number. It would be enough to just get the keys and return the number of keys.

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointslice_controller.go
```diff
@@ -0,0 +1,496 @@
...
+
+	patch := apisv1alpha1apply.APIExportEndpointSlice(obj.Name)
+	if result.remove {
+		patch.WithStatus(apisv1alpha1apply.APIExportEndpointSliceStatus())
+	} else {
+		patch.WithStatus(apisv1alpha1apply.APIExportEndpointSliceStatus().
+			WithAPIExportEndpoints(apisv1alpha1apply.APIExportEndpoint().WithURL(result.url)))
+	}
+
+	_, err = c.clusterClient.ApisV1alpha1().APIExportEndpointSlices().Cluster(clusterName.Path()).ApplyStatus(
```
sttts, line 438: I would move that into a method in the reconciler. The code here should look 100% standard. It's not the place for logic.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
mjudeikis: It is. When binding changed, it might be sign of new consumer or one going away

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointslice_reconcile.go
mjudeikis: I was thinking in the future this could extend to "failed" bindings. meaning there might not be a consumer which is ready.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
mjudeikis: no its not.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
mjudeikis: no

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointslice_reconcile.go
mjudeikis: The problem here is that partition might be based on selector:
```
 eu-shards_only: true
```
 We propagate this selector via APIExportEndpointSlice. But now `Shard X`  does not know if I'm on the list of this selector or not. So it needs to list all shards based on selector from status and check "if I'm on the list". 
 
Maybe there is nicer way to do this. need to think

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
mjudeikis: will rebase with https://github.com/kcp-dev/kcp/pull/3278

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointslice_controller.go
mjudeikis: this pattern is being used all around the code. I think we either leave it or change everywhere.

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointslice_controller.go
mjudeikis: ack

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointslice_controller.go
mjudeikis: In general, if shards changed - their labels might be changed. Hence, slices might need to be re-calculated.

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointslice_controller.go
mjudeikis: I lost this comment somehow in github :/ what was it about?

mjudeikis:
/retest 
flakes flakes flakes :/

pkg/indexers/apiexport.go
```diff
@@ -72,3 +76,17 @@ func IndexAPIExportByClaimedIdentities(obj interface{}) ([]string, error) {
 	}
 	return sets.List[string](claimedIdentities), nil
 }
+
+// IndexAPIExportEndpointSliceByAPIExportFunc indexes the APIExportEndpointSlice by their APIExport's Reference Path and Name.
+func IndexAPIExportEndpointSliceByAPIExport(obj interface{}) ([]string, error) {
```
mjudeikis, line 81: Moving as its being reused. No change.

pkg/indexers/apiexport.go
```diff
@@ -72,3 +76,17 @@ func IndexAPIExportByClaimedIdentities(obj interface{}) ([]string, error) {
...
+	apiExportEndpointSlice, ok := obj.(*apisv1alpha1.APIExportEndpointSlice)
+	if !ok {
+		return []string{}, fmt.Errorf("obj %T is not an APIExportEndpointSlice", obj)
+	}
+
+	path := logicalcluster.NewPath(apiExportEndpointSlice.Spec.APIExport.Path)
+	if path.Empty() {
+		path = logicalcluster.From(apiExportEndpointSlice).Path()
```
sttts, line 89: don't we need both ways to address it?

sttts: Depends on how you look it up later in the code.

mjudeikis: Not sure I understand this comment.

sdk/apis/apis/v1alpha1/types_apiexportendpointslice.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The KCP Authors.
+Copyright 2025 The KCP Authors.
```
sttts, line 2: nit: we usually don't update when just changing files.

sdk/apis/apis/v1alpha1/types_apiexportendpointslice.go
```diff
@@ -93,10 +102,16 @@ type APIExportEndpoint struct {
 }
 
 func (in *APIExportEndpointSlice) GetConditions() conditionsv1alpha1.Conditions {
+	if in == nil {
+		return nil
+	}
```
sttts, line 107: Not sure we want this. It encourages brittle code.

sdk/apis/apis/v1alpha1/types_apiexportendpointslice.go
```diff
@@ -106,7 +121,10 @@ const (
 	// PartitionValid is a condition for APIExportEndpointSlice that reflects the validity of the referenced Partition.
 	PartitionValid conditionsv1alpha1.ConditionType = "PartitionValid"
 
-	APIExportEndpointSliceURLsReady conditionsv1alpha1.ConditionType = "EndpointURLsReady"
+	// EndpointSliceReadyForURLs is a condition for APIExportEndpointSlice that reflects the readiness of the slice to
+	// provide URLs. It is set to True when the slice is ready to provide URLs which will be lifecycled based on existing consumers in the different
+	// shards.
+	APIExportEndpointSliceReadyForURLs conditionsv1alpha1.ConditionType = "EndpointReadyForURLs"
```
sttts, line 127: `ReadyForEndpoints`.

Actually not clear why we need this. When is it not ready?

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_reconcile.go
```diff
@@ -0,0 +1,168 @@
...
+	return r.reconcile(ctx, apiExportEndpointSlice)
+}
+
+func (r *endpointsReconciler) reconcile(ctx context.Context, apiExportEndpointSlice *apisv1alpha1.APIExportEndpointSlice) error {
+	if !conditions.IsTrue(apiExportEndpointSlice, apisv1alpha1.APIExportEndpointSliceReadyForURLs) {
+		return nil
+	}
+
+	s := apiExportEndpointSlice.Status.ShardSelector
+	if s == "" { // should never happen.
```
sttts, line 69: is this what the condition above is guarding?

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_reconcile.go
```diff
@@ -0,0 +1,168 @@
...
+		return nil
+	}
+
+	selector, err := labels.Parse(s)
+	if err != nil {
+		return err
```
sttts, line 75: err means just requeue, right? That's not helping, is it? The string won't change.

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_reconcile.go
```diff
@@ -0,0 +1,168 @@
...
+	apiExportPath := logicalcluster.NewPath(apiExportEndpointSlice.Spec.APIExport.Path)
+	if apiExportPath.Empty() {
+		apiExportPath = logicalcluster.From(apiExportEndpointSlice).Path()
+	}
+	apiExport, err := r.getAPIExport(apiExportPath, apiExportEndpointSlice.Spec.APIExport.Name)
+	if err != nil {
+		return err
+	}
+
+	shards, err := r.listShards(selector)
```
sttts, line 87: why do we need all shards? Just get "our" shard.

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller_test.go
```diff
@@ -0,0 +1,367 @@
+/*
+Copyright 2024 The KCP Authors.
```
sttts, line 2: 25

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
```diff
@@ -0,0 +1,481 @@
...
+	err = c.reconcile(ctx, obj)
+	if err != nil {
+		errs = append(errs, err)
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+// filterShardEvent returns true if the event passes the filter and needs to be processed false otherwise.
+func filterShardEvent(oldObj, newObj interface{}) bool {
```
sttts, line 438: Give it some proper name that says what it does.

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
```diff
@@ -0,0 +1,481 @@
...
+			continue
+		}
+		logger := logging.WithObject(logging.WithReconciler(klog.Background(), ControllerName), obj.(*apisv1alpha1.APIExport))
+		logging.WithQueueKey(logger, key).V(4).Info("queuing APIExportEndpointSlices because of referenced APIExport")
+		c.enqueueAPIExportEndpointSlice(slice)
+	}
+}
+
+// enqueueAllAPIExportEndpointSlices enqueues all APIExportEndpointSlices.
+func (c *controller) enqueueAllAPIExportEndpointSlices(shard interface{}) {
```
sttts, line 342: This is `enqueueShard` and should be strictly typed.

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
```diff
@@ -0,0 +1,481 @@
...
+		return
+	}
+
+	logger := logging.WithQueueKey(logging.WithReconciler(klog.Background(), ControllerName), key)
+	logger.V(4).Info("queueing APIExportEndpointSlice from cache")
+	c.queue.Add(key)
+}
+
+// enqueueAPIExportEndpointSlicesForAPIExport enqueues APIExportEndpointSlices referencing a specific APIExport.
+func (c *controller) enqueueAPIExportEndpointSlicesForAPIExport(obj interface{}) {
```
sttts, line 299: make all of these strictly typed please. This is brittle.

sttts, line 299: why do we care about reconciling when an export changes? Won't APIBindings eventually change and that is enough for us to be triggered?

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
```diff
@@ -0,0 +1,481 @@
...
+// enqueueAPIExportEndpointSlice enqueues an APIExportEndpointSlice.
+func (c *controller) enqueueAPIExportEndpointSlice(obj interface{}) {
+	key, err := kcpcache.DeletionHandlingMetaClusterNamespaceKeyFunc(obj)
+	if err != nil {
+		utilruntime.HandleError(err)
+		return
+	}
+
+	logger := logging.WithQueueKey(logging.WithReconciler(klog.Background(), ControllerName), key)
+	logger.V(4).Info("queueing APIExportEndpointSlice")
```
sttts, line 281: can we add a suffix argument to the method and then do `logger.V(4).Info("queueing APIExportEndpointSlice"+suffix)` to get `queueing APIExportEndpointSlice because APIExport X` (we have some examples already), or use contextual logging with `reason` keys.

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
```diff
@@ -0,0 +1,481 @@
...
+	if d, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+		obj = d.Obj
+	}
+	binding, ok := obj.(*apisv1alpha1.APIBinding)
+	if !ok {
+		utilruntime.HandleError(fmt.Errorf("obj is supposed to be a APIBinding, but is %T", obj))
+		return
+	}
+
+	{ // local to shard
```
sttts, line 222: why double code?

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_reconcile.go
mjudeikis: In theory yes, but I wanted to avoid cases where somebody patched status and now it might give false positive. 
And I think now that maybe this be `omitempty` pointer... Need to check if `empty selector` is valid case.

sdk/apis/apis/v1alpha1/types_apiexportendpointslice.go
mjudeikis: Its more like handover/signal to secondary url controller. I used this a trigger. We could treat all conditions ready as signal too and just drop this all togheter.

sttts: Was thinking about the label selector. It must be known, right?

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_reconcile.go
sttts: yep

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
mjudeikis: But this means we need to do something like:
```
AddFunc: func(obj interface{}) {
			c.enqueueAPIExportEndpointSlicesForAPIExport(obj.(*apisv1alpha1.APIExport))
		},
```
So, is it the same casting, just in a different place? Im fine either way, just this is how we do this all around the code base 😢

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
```diff
@@ -0,0 +1,481 @@
...
+			c.enqueueAPIExportEndpointSlicesForAPIExport(obj)
+		},
+		DeleteFunc: func(obj interface{}) {
+			c.enqueueAPIExportEndpointSlicesForAPIExport(obj)
+		},
+	}))
+
+	_, _ = globalShardClusterInformer.Informer().AddEventHandler(events.WithoutSyncs(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			c.enqueueAllAPIExportEndpointSlices(obj)
```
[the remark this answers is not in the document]
mjudeikis: I was thinking about a case where a change in shards could impact APIExportSlices flow. But wrong assumptions

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
mjudeikis: It's one of those wrong assumptions, I think. WIll try

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
mjudeikis: readability. It isolates variables and makes it very clear to read two blocks. Its all about reading code and reducing mental overhead when reading the code.
// local to shard
// from cache

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
mjudeikis: ah, found a tombstone pattern. looks better

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
mjudeikis: this went away after removal of some code

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -76,8 +78,8 @@ func NewController(
 		listShards: func(selector labels.Selector) ([]*corev1alpha1.Shard, error) {
 			return globalShardClusterInformer.Lister().List(selector)
 		},
-		getAPIExportEndpointSlice: func(clusterName logicalcluster.Name, name string) (*apisv1alpha1.APIExportEndpointSlice, error) {
-			return apiExportEndpointSliceClusterInformer.Lister().Cluster(clusterName).Get(name)
+		getAPIExportEndpointSlice: func(ctx context.Context, path logicalcluster.Path, name string) (*apisv1alpha1.APIExportEndpointSlice, error) {
+			return indexers.ByPathAndName[*apisv1alpha1.APIExportEndpointSlice](apisv1alpha1.Resource("apiexportendpointslices"), apiExportEndpointSliceClusterInformer.Informer().GetIndexer(), path, name)
```
sttts, line 82:
```suggestion
		getAPIExportEndpointSliceByAPIExportPath: func(ctx context.Context, path logicalcluster.Path, name string) (*apisv1alpha1.APIExportEndpointSlice, error) {
			return indexers.ByPathAndName[*apisv1alpha1.APIExportEndpointSlice](apisv1alpha1.Resource("apiexportendpointslices"), apiExportEndpointSliceClusterInformer.Informer().GetIndexer(), path, name)
```

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -104,50 +106,52 @@ func NewController(
 		commit:                                committer.NewCommitter[*APIExportEndpointSlice, Patcher, *APIExportEndpointSliceSpec, *APIExportEndpointSliceStatus](kcpClusterClient.ApisV1alpha1().APIExportEndpointSlices()),
 	}
 
+	logger := logging.WithReconciler(klog.Background(), ControllerName)
+
 	_, _ = apiExportEndpointSliceClusterInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
-			c.enqueueAPIExportEndpointSlice(obj)
+			c.enqueueAPIExportEndpointSlice(objOrTombstone[*apisv1alpha1.APIExportEndpointSlice](obj), logger, "")
```
sttts, line 113: tombstones only exist for `DeleteFunc`.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -227,35 +221,34 @@ func (c *controller) enqueueAPIExportEndpointSlicesForAPIExport(obj interface{})
 		} else if !exists {
 			continue
 		}
-		logger := logging.WithObject(logging.WithReconciler(klog.Background(), ControllerName), obj.(*apisv1alpha1.APIExport))
 		logging.WithQueueKey(logger, key).V(4).Info("queuing APIExportEndpointSlices because of referenced APIExport")
```
sttts, line 224: now you log twice, don't you?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -227,35 +221,34 @@ func (c *controller) enqueueAPIExportEndpointSlicesForAPIExport(obj interface{})
...
-		c.enqueueAPIExportEndpointSlice(slice)
+		c.enqueueAPIExportEndpointSlice(objOrTombstone[*apisv1alpha1.APIExportEndpointSlice](slice), logger, " because of referenced APIExport")
 	}
 }
 
 // enqueueAllAPIExportEndpointSlices enqueues all APIExportEndpointSlices.
-func (c *controller) enqueueAllAPIExportEndpointSlices(shard interface{}) {
+func (c *controller) enqueueAllAPIExportEndpointSlices(shard *corev1alpha1.Shard, logger logr.Logger, logSuffix string) {
```
sttts, line 230: why react to all shards?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
```diff
@@ -268,10 +261,9 @@ func (c *controller) enqueuePartition(obj interface{}) {
 		return
 	}
 
-	logger := logging.WithObject(logging.WithReconciler(klog.Background(), ControllerName), obj.(*topologyv1alpha1.Partition))
 	logging.WithQueueKey(logger, key).V(4).Info("queuing APIExportEndpointSlices because Partition changed")
```
sttts, line 264: now we log twice

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
```diff
@@ -165,90 +127,24 @@ func (r *endpointsReconciler) reconcile(ctx context.Context, apiExportEndpointSl
...
 	}
 	if selector == nil {
 		selector = labels.Everything()
 	}
+
 	conditions.MarkTrue(apiExportEndpointSlice, apisv1alpha1.PartitionValid)
 
-	// Get shards
-	shards, err := r.listShards(selector)
+	// Check if listing shards works. This is a good indicator that the cache is up to date.
```
sttts, line 139: it always works if we synced. Don't think there is value.

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
```diff
@@ -0,0 +1,379 @@
...
+		queue: workqueue.NewTypedRateLimitingQueueWithConfig(
+			workqueue.DefaultTypedControllerRateLimiter[string](),
+			workqueue.TypedRateLimitingQueueConfig[string]{
+				Name: ControllerName,
+			},
+		),
+		getMyShard: func() (*corev1alpha1.Shard, error) {
+			return globalShardClusterInformer.Cluster(core.RootCluster).Lister().Get(shardName)
+		},
+		getAPIExportEndpointSlice: func(ctx context.Context, path logicalcluster.Path, name string) (*apisv1alpha1.APIExportEndpointSlice, error) {
```
sttts, line 78: no `ctx`.

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
```diff
@@ -0,0 +1,379 @@
...
+		},
+		apiExportEndpointSliceClusterInformer:       apiExportEndpointSliceClusterInformer,
+		globalApiExportEndpointSliceClusterInformer: globalAPIExportEndpointSliceClusterInformer,
+	}
+
+	logger := logging.WithReconciler(klog.Background(), ControllerName)
+
+	_, _ = apiExportEndpointSliceClusterInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			c.enqueueAPIExportEndpointSlice(objOrTombstone[*apisv1alpha1.APIExportEndpointSlice](obj), logger, "")
```
sttts, line 133: tombstones only for `DeleteFunc`.

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
```diff
@@ -0,0 +1,379 @@
...
+		},
+		UpdateFunc: func(_, newObj interface{}) {
+			c.enqueueAPIExportEndpointSliceByAPIBinding(objOrTombstone[*apisv1alpha1.APIBinding](newObj), logger)
+		},
+		DeleteFunc: func(obj interface{}) {
+			c.enqueueAPIExportEndpointSliceByAPIBinding(objOrTombstone[*apisv1alpha1.APIBinding](obj), logger)
+		},
+	})
+
+	_, _ = globalAPIExportEndpointSliceClusterInformer.Informer().AddEventHandler(events.WithoutSyncs(cache.ResourceEventHandlerFuncs{
```
sttts, line 155: we have to watch our slide, don't we?

sdk/apis/apis/v1alpha1/types_apiexportendpointslice.go
```diff
@@ -97,6 +106,9 @@ func (in *APIExportEndpointSlice) GetConditions() conditionsv1alpha1.Conditions
 }
 
 func (in *APIExportEndpointSlice) SetConditions(conditions conditionsv1alpha1.Conditions) {
+	if in == nil {
+		return
+	}
```
sttts, line 111: we don't want this

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
mjudeikis: Not sure I get the comment. Is it not what this function does?
Return obj or tombstone if it was deleted?

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
mjudeikis: This helper is friendly for consistency. Are you suggesting just to cast to type where we don't need tombstone?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
mjudeikis: Shard might change (label changed), so it got removed from partition somewhere else, so we need to recompute the partitions and selectors every time shard changed.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
mjudeikis: We been using this pattern everywhere (updated,created, deletes) and it keeps it consistent (easier to read too).
So why to break this here?

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
mjudeikis: Dont think so. We watch all as we might need to add our URL to foreign APIExportEndpointSlice if our shard is part of the partition, but we don't host it. This is why the patch client uses an external client.

pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_controller.go
sttts: oh, this also casts. Ok.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
```diff
@@ -81,13 +69,6 @@ func (r *endpointsReconciler) reconcile(ctx context.Context, apiExportEndpointSl
 				apiExportPath,
 				apiExportEndpointSlice.Spec.APIExport.Name,
 			)
-			conditions.MarkFalse(
-				apiExportEndpointSlice,
-				apisv1alpha1.APIExportEndpointSliceURLsReady,
-				apisv1alpha1.ErrorGeneratingURLsReason,
-				conditionsv1alpha1.ConditionSeverityError,
-				"",
-			)
```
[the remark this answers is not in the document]
sttts: I meant whether it is actively removed. If we don't update it anymore, we have to remove it once. Or is this controller using SSA?

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
sttts: I don't understand. I thought the non-url controller is for the local slices only. It does not depend (in this controller) from shards anymore.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_reconcile.go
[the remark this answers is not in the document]
mjudeikis: Added removal and deprecation. Took me a while to get this.

pkg/reconciler/apis/apiexportendpointslice/apiexportendpointslice_controller.go
mjudeikis: I think you are right. I just needed to re-read this a bit myself to understand how this is wired. And Im not sure why Shard was in the picture for this controller in the first place. 

Setup usually is like this:
```
shard1 -->
shard2 --> PartitionSet --> Partitions
shard3 --> 
```

this bubbles to downstream:
```
Partition --> APIExportEndpointSlices <-- APIExport (global as it might be from outside)
```

The non-url controller should work only on `Partitions`, `APIExport` or `APIExportEndpointSlices` changes to recompute the result. 

URL should act on `APIBindings` and `APIExportEndpointSlices`.

PR #3263, opened 2025-01-23
:bug: Correctly check original groups against `system:masters` in scope filter

embik:
## Summary

We discovered that the `WithImpersonationScoping` filter had a flaw: It was checking the already impersonated user identity to determine whether adding a scope should be skipped for `system:masters` impersonators. But this check needs to run against the original user identity, not the impersonated identity (since that will never include `system:masters`).

This PR stores the user identity during `WithImpersonationGatekeeper` in the request context to fetch it at a later stage in `WithImpersonationScoping`.

As a bonus, I discovered that the check for extra impersonation headers is not correct: It was checking the HTTP header values, but it should be checking the name/key of the header. So the list of extra impersonations was always empty. The tests didn't check this.

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/server/filters/impersonation.go
+++ b/pkg/server/filters/impersonation.go
@@ -52,12 +52,13 @@ var (
 	}
 )
 
-// impersonationContextType is a context key for impersonation marker. It's true
-// if the request is impersonated.
+// impersonationContextType is a context key for impersonation markers.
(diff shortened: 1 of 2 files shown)


embik:
/hold

just ... checking some things.

PR #3267, opened 2025-01-24
:bug: impersonation: add missing privileged groups

sttts:
## Summary

There are some more priviliged groups that this PR is adding to the impersonation guard. Only `system:master` suffices to impersonate as any of them if the incoming user does not have that group already.

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


sttts:
Flake.

/retest

PR #3272, opened 2025-01-26
:seedling: Configure prow artifacts correctly

sttts:
## Summary

Prow collects artifacts from the directory in `$(ARTIFACTS)`. In kcp we use `ARTIFACT_DIR`. This PR sets it up correctly that `ARTIFACT_DIR` is set to `$(ARTIFACTS)`.

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -109,6 +109,8 @@ presubmits:
             - ./hack/run-with-prometheus.sh
             - make
             - test-e2e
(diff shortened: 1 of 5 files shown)


embik (approved):
/approve

JUnit results show up in Prow, so seems pretty safe to say this works.

PR #3273, opened 2025-01-26
:bug: index: fix how error codes are stored for unavailable workspaces

sttts:
## Summary

The old code stored the error code under the logical cluster which was supposed to fail, not the workspace pointing to it. This kind of worked on one shard, but with multiple it failed when workspace and target cluster were not colocated.

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


mjudeikis (approved):

mjudeikis:
/retest
looks like we still have flakes

PR #3274, opened 2025-01-26
WIP: :bug: authentication: add custom SA lookup with ttl cache for non-local clusters

sttts:
## Summary

With service accounts valid (even as `system:anonymous`+`system:authenticated`) on other clusters, we have to validate them on other shards. As there are many service accounts, we cannot distribute them via the cache server. But luckily there are not many those service accounts. We can use some ttl cache for that and use a real external client.

### TODO

- [ ] wire external client

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


mjudeikis:
We need to revist this

mjudeikis:
will close ths one as we have enough reference to this poc

PR #2595, opened 2023-01-11
:seedling: remove tenancy v1beta1

stevekuznetsov:
pkg/apis: add Workspace to tenancy.kcp.io/v1alpha1

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

*: use v1alpha1.Workspaces everywhere

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

*: remove tenancy.kcp.io/v1beta1

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

/assign @sttts @ncdc 
Implements #2563

PR #3282, opened 2025-02-02
:sparkles: Promote mounts to spec

mjudeikis:
## Related issue(s)

Fixes #

## Release Notes

```release-note
Deprecate experimental mount annotations
Promote mounts to workspace.spec.mount
```


sdk/apis/tenancy/v1alpha1/types_workspace.go
```diff
@@ -174,6 +176,23 @@ type WorkspaceSpec struct {
 	//
 	// +kubebuilder:format:uri
 	URL string `json:"URL,omitempty"`
+
+	// Mount is a reference to a mount that is used to mount the workspace.
+	// If specified, logicalcluster will not be created and the workspace will be mounted
+	// using reference mount object.
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="mount is immutable"
```
sttts, line 185: is this enough to stop you from removing the field? I think we need something like line 153 (on the higher level) with `has`.

sdk/apis/tenancy/v1alpha1/types_workspace.go
```diff
@@ -174,6 +176,23 @@ type WorkspaceSpec struct {
...
+	Mount *Mount `json:"mount,omitempty"`
+}
+
+func (in *WorkspaceSpec) IsMounted() bool {
+	return in.Mount != nil
+}
+
+type Mount struct {
+	// Reference is an ObjectReference to the object that is mounted.
+	Reference corev1.ObjectReference `json:"ref,omitempty"`
```
sttts, line 195: please use our own variant. This has far too many fields and makes the semantics unclear.

sttts:
Have not read all the controller changes. Have you changed it to not have the shadow logical cluster? I.e. a mount is a mount, always. There is not this hybrid state depending on readiness.

mjudeikis:
> Have not read all the controller changes. Have you changed it to not have the shadow logical cluster? I.e. a mount is a mount, always. There is not this hybrid state depending on readiness.

I think I did, need to check I pushed it out, but basically goal is "shallow workspace object" with no logicalcluster backing it.

sdk/apis/tenancy/v1alpha1/types_workspace.go
mjudeikis: done

sdk/apis/tenancy/v1alpha1/types_workspace.go
mjudeikis: Added. Somehow, it fails to start when I add a filter to `namespace` Do you have any idea why the `omit empty` optional field fails to start?

sdk/apis/tenancy/v1alpha1/types_workspace.go
```diff
@@ -174,6 +175,52 @@ type WorkspaceSpec struct {
...
+	Kind string `json:"kind"`
+	// Name is the name of the object.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+	// Namespace is the namespace of the object.
+	//
+	// +optional
+	// +kubebuilder:validation:Optional
```
sttts, line 222: I believe kubebuilder understands both markers.

sdk/apis/tenancy/v1alpha1/types_workspace.go
```diff
@@ -174,6 +175,52 @@ type WorkspaceSpec struct {
...
+	// +kubebuilder:validation:XValidation:rule="has(oldSelf.name) == has(self.name)",message="name is immutable"
+	Reference ObjectReference `json:"ref,omitempty"`
+}
+
+type ObjectReference struct {
+	// APIVersion is the API group and version of the object.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	APIVersion string `json:"apiVersion"`
```
sttts, line 208: minLength=1

sttts: same below

sdk/apis/tenancy/v1alpha1/types_workspace.go
```diff
@@ -174,6 +175,52 @@ type WorkspaceSpec struct {
...
+// Mount is a reference to a mount that is used to mount the workspace.
+type Mount struct {
+	// Reference is an ObjectReference to the object that is mounted.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:XValidation:rule="has(oldSelf.apiVersion) == has(self.apiVersion)",message="apiVersion is immutable"
+	// +kubebuilder:validation:XValidation:rule="has(oldSelf.kind) == has(self.kind)",message="kind is immutable"
+	// +kubebuilder:validation:XValidation:rule="has(oldSelf.name) == has(self.name)",message="name is immutable"
+	Reference ObjectReference `json:"ref,omitempty"`
```
sttts, line 200: no omitempty

sdk/apis/tenancy/v1alpha1/types_workspace.go
```diff
@@ -117,7 +117,7 @@ const LogicalClusterTypeAnnotationKey = "internal.tenancy.kcp.io/type"
 // +kubebuilder:resource:scope=Cluster,categories=kcp,shortName=ws
 // +kubebuilder:printcolumn:name="Type",type=string,JSONPath=`.spec.type.name`,description="Type of the workspace"
 // +kubebuilder:printcolumn:name="Region",type=string,JSONPath=`.metadata.labels['region']`,description="The region this workspace is in"
-// +kubebuilder:printcolumn:name="Phase",type=string,JSONPath=`.metadata.labels['tenancy\.kcp\.io/phase']`,description="The current phase (e.g. Scheduling, Initializing, Ready, Deleting)"
+// +kubebuilder:printcolumn:name="Phase",type=string,JSONPath=`.status.phase`,description="The current phase (e.g. Scheduling, Initializing, Ready, Deleting)"
```
sttts, line 120: is this a no-op? Is the code removed for that label? Is the label removed?

sdk/apis/tenancy/v1alpha1/types_workspace.go
```diff
@@ -134,6 +134,7 @@ type Workspace struct {
 // WorkspaceSpec holds the desired state of the Workspace.
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.URL) || has(self.URL)",message="URL cannot be unset"
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.cluster) || has(self.cluster)",message="cluster cannot be unset"
+// +kubebuilder:validation:XValidation:rule="!(has(self.mount) && has(self.type))",message="spec.mount and spec.type cannot both be set"
```
sttts, line 137: what about immutability?

test/e2e/mounts/mounts_machinery_test.go
```diff
@@ -49,6 +49,12 @@ import (
 //go:embed *.yaml
 var testFiles embed.FS
 
+// Test is exercising the mounts api. In production we should be configurating front-proxy
+// with urls for routing requests to right backend, potentially virtual-workspaces using mappings file.
+// This is due to reason front-proxy will not route traffic to not-trusted urls.
+// To test this in the text, we will route traffic to another workspace. This kinda simulates mounts in
+// symbolic links fashion.
+
 func TestMountsMachinery(t *testing.T) {
```
sttts, line 58:
```suggestion
// TestMountsMachinery is exercising the mounts api. In production we should be configurating front-proxy
// with urls for routing requests to right backend, potentially virtual-workspaces using mappings file.
// This is due to reason front-proxy will not route traffic to not-trusted urls.
// To test this in the text, we will route traffic to another workspace. This kinda simulates mounts in
// symbolic links fashion.
func TestMountsMachinery(t *testing.T) {
```

test/e2e/mounts/mounts_machinery_test.go
```diff
@@ -146,50 +156,51 @@ func TestMountsMachinery(t *testing.T) {
...
 		if err != nil {
 			return false, err.Error()
 		}
 
-		initialized := conditions.IsTrue(current, tenancyv1alpha1.WorkspaceInitialized)
 		ready := conditions.IsTrue(current, tenancyv1alpha1.MountConditionReady)
-		return initialized && ready, yamlMarshal(t, current)
-	}, wait.ForeverTestTimeout, 100*time.Millisecond, "waiting for WorkspaceMountReady and WorkspaceInitialized conditions to be true")
+		return ready, yamlMarshal(t, current)
+	}, wait.ForeverTestTimeout, 100*time.Millisecond, "waiting for WorkspaceMountReady conditions to be true")
```
sttts, line 173:
```suggestion
	}, wait.ForeverTestTimeout, 100*time.Millisecond, "waiting for WorkspaceMountReady condition to be true")
```

sdk/apis/tenancy/v1alpha1/types_workspace.go
sttts: nvmd if you have admission in place

sdk/apis/tenancy/v1alpha1/types_workspace.go
mjudeikis: I dont understand why label was used in the first place. Mounts dont have this label. 
Why label was used in the first place but not status itself as reference for column?

sttts: Maybe there are states not represented in the phase?

mjudeikis: No idea. I was not able to find a case of why this was done. Just looks like some side conversations which was not documented 🤷 
It was introduced in bootstrapping PR: https://github.com/kcp-dev/kcp/pull/2595

sdk/apis/tenancy/v1alpha1/types_workspace.go
mjudeikis: admission does this.

pkg/admission/workspacetypeexists/admission.go
```diff
@@ -132,8 +137,8 @@ func (o *workspacetypeExists) Admit(ctx context.Context, a admission.Attributes,
 	}
 
 	// if the user has not provided any type, use the default from the parent workspace
-	empty := tenancyv1alpha1.WorkspaceTypeReference{}
-	if ws.Spec.Type == empty {
+	// We default to workspaceType only if not mount.
+	if ws.Spec.Type == nil || *ws.Spec.Type == (tenancyv1alpha1.WorkspaceTypeReference{}) {
```
embik, line 141: This feels a little odd as a comparison, shouldn't we simply check for `ws.Spec.Type.Name == ""`?

pkg/index/index.go
```diff
@@ -271,28 +275,30 @@ func (c *State) Lookup(path logicalcluster.Path) (Result, bool) {
...
+			// Else we get back to default behavior.
+			wsSpec, foundMount := c.shardClusterWorkspaceMount[shard][originalCluster][s] // experimental feature
+			if foundMount {
+				if wsSpec.Mount != nil && wsSpec.URL != "" {
+					u, err := url.Parse(wsSpec.URL)
+					if err != nil {
+						// default to workspace itself.
+					} else {
+						return Result{URL: u.String(), ErrorCode: errorCode}, true
+					}
```
embik, line 296: isnt this simply

```suggestion
					if err == nil {
						return Result{URL: u.String(), ErrorCode: errorCode}, true
					}
```

pkg/reconciler/tenancy/workspacemounts/workspacemounts_reconcile_updater.go
```diff
@@ -35,102 +36,74 @@ import (
...
-	mount, err := tenancyv1alpha1.ParseTenancyMountAnnotation(v)
+
+	if !workspace.DeletionTimestamp.IsZero() {
+		return reconcileStatusContinue, nil
+	}
+
+	obj, err := r.getMountObject(ctx, logicalcluster.From(workspace).Path(), workspace.Spec.Mount.Reference)
 	if err != nil {
+		return reconcileStatusStopAndRequeue, err
+	} else if kerrors.IsNotFound(err) {
```
embik, line 54: This doesn't look right, it would never enter the `else if` branch, no?

sdk/apis/tenancy/v1alpha1/types_workspace.go
```diff
@@ -174,6 +175,48 @@ type WorkspaceSpec struct {
...
+	//
+	// +optional
+	Mount *Mount `json:"mount,omitempty"`
+}
+
+func (in *WorkspaceSpec) IsMounted() bool {
+	return in.Mount != nil
+}
+
+// Mount is a reference to a mount that is used to mount the workspace.
```
embik, line 191: This sentence makes me deeply unhappy 😢

embik (changes requested):
Few minor notes.

pkg/reconciler/tenancy/workspacemounts/workspacemounts_reconcile_updater.go
yastij: +1

mjudeikis:
> Because if we merge it for 0.27, I wanna make sure (with a follow-up ticket?) we document this. The current comments on the struct only really make sense if you already know how this works.

At this point when it goes it goes :) no preference. Docs will be next ticket once this gets final rubber stamp for merging.

sdk/apis/tenancy/v1alpha1/types_workspace.go
mjudeikis: changed :D holy chatGPT :D

pkg/admission/workspacetypeexists/admission.go
mjudeikis: I think this is better now: `if ws.Spec.Type == nil || (ws.Spec.Type != nil && ws.Spec.Type.Name == "") {`

embik: > There is a semi-breaking change where type now becomes a pointer and optional, so it can be nil now.

Per se yes, but because you are already checking `ws.Spec.Type == nil` in the first part of the condition, the part after `||` will only ever be evaluated if `Type` is not `nil`. So we don't need another check for that.

pkg/reconciler/tenancy/workspacemounts/workspacemounts_reconcile_updater.go
```diff
@@ -85,52 +61,50 @@ func (r *workspaceStatusUpdater) reconcile(ctx context.Context, workspace *tenan
...
-		// Only Spec or Status can be updated, not both.
-		if mount.MountStatus.Phase != tenancyv1alpha1.MountPhaseType(statusPhase) || mount.MountStatus.URL != statusURL {
-			mount.MountStatus.Phase = tenancyv1alpha1.MountPhaseType(statusPhase)
-			mount.MountStatus.URL = statusURL
+		return reconcileStatusContinue, nil //nolint:nilerr // we ignore the error intentionally. Not helpful.
+	}
 
-			workspace.Annotations[tenancyv1alpha1.ExperimentalWorkspaceMountAnnotationKey] = mount.String()
+	// url might not be there if the mount is not ready
+	statusURL, _, _ := unstructured.NestedString(obj.Object, "status", "URL")
```
yastij, line 85: can we pull the workspaceType and Cluster too? otherwise adopting this against code that already exists would be pretty hard. the other option is to relax admission with the ability to set those when mount is specified

yastij: nvm, looks like validation is relaxed, still would really be nice to pull the type and cluster back. otherwise all existing integrations would have to be reworked. that could slow/prevent adoption of mounts

pkg/reconciler/tenancy/workspacemounts/workspacemounts_reconcile_updater.go
```diff
@@ -109,3 +102,33 @@ func (r *workspaceStatusUpdater) reconcile(ctx context.Context, workspace *tenan
...
+	statusURL, found, err := unstructured.NestedString(obj.Object, "status", "URL")
+	if !found || err != nil {
+		return fmt.Errorf("unable to read .status.URL, found %v, err: %w", found, err)
+	}
+	workspace.Spec.URL = statusURL
+
+	clusterName, _, _ := unstructured.NestedString(obj.Object, "status", "cluster")
+	workspace.Spec.Cluster = clusterName
+
+	wsTypeName, found, err := unstructured.NestedString(obj.Object, "status", "type", "name")
```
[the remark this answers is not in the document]
yastij: @sttts - not sure they should be part of the spec of the object satisfying the contract, as that object gathers that info. but no strong opinions. cc @mjudeikis thoughts?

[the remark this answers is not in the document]
mjudeikis: You mean spec of the "mount object"? Not sure I understand the proposal.

pkg/reconciler/tenancy/workspacemounts/workspacemounts_controller.go
```diff
@@ -197,20 +197,61 @@ func (c *Controller) process(ctx context.Context, key string) (bool, error) {
...
 
 	// If the object being reconciled changed as a result, update it.
 	oldResource := &workspaceResource{ObjectMeta: old.ObjectMeta, Spec: &old.Spec, Status: &old.Status}
-	newResource := &workspaceResource{ObjectMeta: workspace.ObjectMeta, Spec: &workspace.Spec, Status: &workspace.Status}
+	newResource := &workspaceResource{ObjectMeta: workspace.ObjectMeta, Spec: &old.Spec, Status: &workspace.Status}
+	if err := c.commit(ctx, oldResource, newResource); err != nil {
+		return false, err
+	}
+
+	// reconcole the spec
```
mjudeikis, line 235: `reconcile the spec`

pkg/reconciler/tenancy/workspacemounts/workspacemounts_controller.go
```diff
@@ -197,20 +197,61 @@ func (c *Controller) process(ctx context.Context, key string) (bool, error) {
...
+	specUpdater := &workspaceSpecUpdater{
+		getMountObject: getMountObjectFunc,
+	}
+	status, err = specUpdater.reconcile(ctx, workspace)
```
mjudeikis, line 239: Can we get at least the comment why this breaks the standard patter? 
is this to avoid status and spec update in the same loop?

palnabarun: Yes, the standard pattern is limited to updating only either spec or status.

I will add a comment to explain.

palnabarun: Done.

palnabarun:
/test pull-kcp-test-e2e

An unrelated test failed.

pkg/reconciler/tenancy/workspacemounts/workspacemounts_controller.go
```diff
@@ -197,20 +197,65 @@ func (c *Controller) process(ctx context.Context, key string) (bool, error) {
...
+		resource := strings.ToLower(ref.Kind) + "s"
+		gvr := schema.GroupVersionResource{Resource: resource}
+		cs := strings.SplitN(ref.APIVersion, "/", 2)
+		if len(cs) == 2 {
+			gvr.Group = cs[0]
+			gvr.Version = cs[1]
+		} else {
+			gvr.Version = ref.APIVersion
+		}
+		return c.dynamicClusterClient.Cluster(cluster).Resource(gvr).Get(ctx, ref.Name, metav1.GetOptions{})
```
dakraus, line 211: It looks like the namespace of the reference isn't being used here. Is this intentional, or am I missing something?

mjudeikis: Idea is that mounts are cluster-scoped objects. Dont support namespace one for now

dakraus: Okay, got it. Thanks for the clarification.

mjudeikis: But good point. We need to document this.

pkg/reconciler/tenancy/workspacemounts/workspace_indexes.go
```diff
@@ -0,0 +1,94 @@
...
+func indexWorkspaceByURL(obj interface{}) ([]string, error) {
+	ws, ok := obj.(*tenancyv1alpha1.Workspace)
+	if !ok {
+		return nil, fmt.Errorf("obj %T is not an Workspace", obj)
+	}
+
+	return []string{ws.Spec.URL}, nil
+}
+
+const workspaceMountsReferenceIndex = "WorkspacesByMountReference"
```
mjudeikis, line 42: Everything below this is copy-paste from another file to keep index-related stuff in one file.

test/e2e/mounts/mount_ws.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: tenancy.kcp.io/v1alpha1
```
mjudeikis, line 1: Will delete. Not used

embik:
/retest

flake (#3371)

mjudeikis:
Need to rebase and see how we can add "References" into this. This was split and core got merged.

mjudeikis:
Will close this as we have mvp merged. This adds type resolution but original requestor of this is not responsive on this.

PR #3298, opened 2025-02-13
:bug: fix(admission): rename allOf to anyOf for type extension

thevilledev:
## Summary

The function `allOfTheFormerExistInTheLater` is renamed to `anyOfTheFormerExistInTheLater` to better reflect its behavior. The function implements additive type extension where a type gains capabilities of types it extends.

The implementation was correct, but the name suggested an "all" check when it performs an "any" check. This aligns with documented behavior where a type can be used anywhere its extended type is allowed.

This is a naming-only change with no behavioral modifications.

## Related issue(s)

Fixes `nil`

## Release Notes

```release-note
NONE
```


embik:
Hi @thevilledev, apologies for the confusion - I misread your initial report as this being a field on `WorkspaceTypes`. Since that is obviously not the case and your PR renames a function appropriately, feel free to mark the PR as ready, we're happy to merge it immediately :)

embik (approved):
/approve

Thank you so much for the contribution!

PR #3308, opened 2025-02-24
:bug: Dockerfile: fix constructing URL for kubectl

gman0:
## Summary

`go list -m -json k8s.io/kubernetes | jq -r .Version` outputs a version string in format `v0.<Minor>.<Patch>`, but binaries at dl.k8s.io/release are versioned in line with real K8s versions. This commit replaces the "v0" part with "v1" so that the URL is correct.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@ COPY cli/go.sum cli/go.sum
 USER 0
 
 # Install kubectl.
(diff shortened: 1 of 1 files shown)


gman0:
https://github.com/kcp-dev/kcp/blob/main/go.mod#L41 actually has `v1.31.0`

PR #3313, opened 2025-02-26
:book: remove superflous output from make install command

Skarlso:
## Summary

The output from `make install` no longer matches the output in the docs. So we just remove that to prevent confusion as to what to execute.

## Related issue(s)

Fixes #

## Release Notes

```release-note
remove superflous output from make install command
```


embik (approved):
/approve

PR #3317, opened 2025-03-03
:seedling:  hack/verify-go-modules.sh: don't run pager with git diff

gman0:
## Summary

Depending on which pager git is configured to use, it may clear terminal's screen, including the error message from verify-go-modules.sh. This then may be confusing for the user where is this `git diff` coming from.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/hack/verify-go-modules.sh
+++ b/hack/verify-go-modules.sh
@@ -107,8 +107,8 @@ for dir in "${DIRS[@]}"; do
     echo "Verifying ${dir}"
 
     if ! git diff --quiet HEAD -- go.mod go.sum; then
-      echo "go module files are out of date"
-      git diff HEAD -- go.mod go.sum
+      git --no-pager diff HEAD -- go.mod go.sum
+      echo "Error: go.mod and/or go.sum in ${dir} files have been modified, inspect and commit them before continuing" 1>&2
       exit 1
     fi
 


hack/verify-go-modules.sh
```diff
@@ -107,8 +107,8 @@ for dir in "${DIRS[@]}"; do
     echo "Verifying ${dir}"
 
     if ! git diff --quiet HEAD -- go.mod go.sum; then
-      echo "go module files are out of date"
-      git diff HEAD -- go.mod go.sum
+      git --no-pager diff HEAD -- go.mod go.sum
+      echo "Error: go.mod and/or go.sum in ${dir} files have been modified, inspect and commit them before continuing" 1>&2
```
gman0, line 111: Having no explanation about the error, I'm just assuming that this is what we want the user to do but I'm not really sure that that's the workflow we want to enforce.

`make verify-modules` first runs `hack/update-go-modules.sh` which may have modified go mod files and we want the user to notice somehow. But nothing stops the user from running `hack/verify-go-modules.sh` alone and they wouldn't notice anything.

Instead, I'd propose we run [`go mod tidy -diff`](https://go.dev/ref/mod#go-mod-tidy) and fail if the go mod files are outdated and the user needs to clean up first.

embik (approved):
/approve

PR #3318, opened 2025-03-03
✨ create apis/v1alpha2, add new ResourceSchemas

### LINKED ISSUE #3301: feature: Implement API change for APIExport latestResourceSchemas
### Feature Description

Following https://docs.google.com/document/d/18Xh2-VFnH23bDiAHPP_JXm4qAG4QK6GjTKcpszPGApQ/edit?tab=t.0

This is the first ticket in the implementation list.

Epic: https://github.com/kcp-dev/kcp/issues/3292

### Proposed Solution

Add resourceSchemas new field, which would be used in favor of latestResourceSchemas field. 
Anything added to the latestResourceSchemas field would be automatically added to `resourceSchemas` via mutation and set to storage: CRD. The latestResourceShemas field would be marked as deprecated, and resource schemas would be the source of truth. Bump API version by minor version

### Alternative Solutions

_No response_

### Want to contribute?

### Additional Context

_No response_

xrstf:
## Summary
This PR adds a new version to the `apis.kcp.io` API, `v1alpha2`. This new version only contains the slightly updated APIExports, which now support a more elaborate configuration structure to enable virtual resources in the future.

All the code was updated to refer to `v1alpha2.APIExports`.

Conversion between the two versions is handled by a wrapper around the stock CR Converter, see individual commits for more information.

## Related issue(s)
Fixes #3301

TODO:
- [x] Add admission to check valid json annotations
- [x] Add admission to check for duplicates

Confirmed that if you add a new type into a new API and don't sort converters, all stuff just explodes in tests and generation :) which is good! 

## Release Notes
```release-note
Introduce `apis.kcp.io/v1alpha2`, containing solely `APIExport`
```


xrstf:
> I would really like it to be split in commits: All API changes and webhook into 1 commit, Sed replace changes to another.

Will do, for rebasing it was just simpler to keep it locally as 1 commit for now.

sdk/apis/apis/v1alpha2/types_apiexport.go
```diff
@@ -0,0 +1,353 @@
...
+	// have no effect on existing APIBindings, but only on newly bound ones.
+	//
+	// For updating existing APIBindings, use an APIDeployment keeping bound
+	// workspaces up-to-date.
+	//
+	// Deprecated: use resourceSchemas instead
+	//
+	// +optional
+	// +listType=set
+	LatestResourceSchemas []string `json:"latestResourceSchemas,omitempty"`
```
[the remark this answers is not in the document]
sttts: remove. Having duplicated info is not healthy.

sdk/apis/apis/v1alpha2/types_apiexport.go
```diff
@@ -0,0 +1,328 @@
...
+	//
+	// +optional
+	// +listType=map
+	// +listMapKey=group
+	// +listMapKey=resource
+	PermissionClaims []PermissionClaim `json:"permissionClaims,omitempty"`
+}
+
+type ResourceSchema struct {
+	Schema  string                `json:"schema"`
```
sttts, line 166: IMO this is not helpful as a map key. We should add resource and group now I think.

sdk/apis/apis/v1alpha2/types_apiexport.go
```diff
@@ -0,0 +1,328 @@
...
+	Storage ResourceSchemaStorage `json:"storage"`
```
sttts, line 167: go doc. And it needs a default.

sdk/apis/apis/v1alpha2/types_apiexport.go
```diff
@@ -0,0 +1,328 @@
...
+}
+
+// ResourceSchemaStorage must set either CRD or Virtual, but never both.
+type ResourceSchemaStorage struct {
+	CRD     *ResourceSchemaStorageCRD     `json:"crd,omitempty"`
+	Virtual *ResourceSchemaStorageVirtual `json:"virtual,omitempty"`
```
sttts, line 173: go doc

sdk/apis/apis/v1alpha2/types_apiexport.go
```diff
@@ -0,0 +1,353 @@
...
+}
```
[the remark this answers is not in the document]
sttts: Do we expect additional information? I could imagine we do.

sdk/apis/apis/v1alpha2/types_apiexport.go
```diff
@@ -0,0 +1,328 @@
...
+
+type ResourceSchemaStorageCRD struct{}
+
+type ResourceSchemaStorageVirtual struct {
+	Reference corev1.TypedLocalObjectReference `json:"reference"`
```
sttts, line 179: go doc

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+		out.PermissionClaims = newClaims
+	}
+
+	if schemas := in.ResourceSchemas; schemas != nil {
+		hubSchemas := []string{}
+		for _, schema := range schemas {
+			// TODO: Should this return an error if non-CRD-based resource schemas are used, or
+			// resort to an annotation?
+			hubSchemas = append(hubSchemas, schema.Schema)
```
sttts, line 57: annotation. This must roundtrip. Please add some fuzz tests.

sdk/apis/apis/v1alpha2/doc.go
```diff
@@ -0,0 +1,21 @@
+/*
+Copyright 2022 The KCP Authors.
```
sttts, line 2: nit: 2025 for new files

sdk/apis/apis/v1alpha2/types_apiexport.go
```diff
@@ -0,0 +1,330 @@
...
+	// +optional
+	// +listType=map
+	// +listMapKey=group
+	// +listMapKey=resource
+	PermissionClaims []PermissionClaim `json:"permissionClaims,omitempty"`
+}
+
+type ResourceSchema struct {
+	// Schema is the name of the referenced APIResourceSchema.
+	Schema string `json:"schema"`
```
sttts, line 152: this stutters. `spec.resourceSchemas.name` or `spec.resources.schema`. My vote goes for the second.

In Slack we discussed:
```
spec:
  resources:
  - name:
    group:
    schema:
```
I see you list-map-key above reflects that. Here I see only scheme 🤔

sdk/apis/apis/v1alpha2/types_apiexport.go
sttts: The implementation for this will come later, right? So either remove it for now, or at least use CEL or OpenAPI to forbid it.

sdk/apis/apis/v1alpha2/types_apiexport.go
sttts: So this object will live beside the APIExport? Will shards need it to serve virtual resources? Who will replicate it?

sdk/apis/apis/v1alpha2/types_apiexport_conversion_test.go
```diff
@@ -0,0 +1,162 @@
...
+						Virtual: &ResourceSchemaStorageVirtual{
+							Reference: corev1.TypedLocalObjectReference{
+								Kind: "test",
+							},
+						},
+					},
+				}},
+			},
+		},
+	}
```
sttts, line 89: Would like to see fuzzing tests.

pkg/server/conversion_webhook.go
```diff
@@ -0,0 +1,104 @@
...
+	apisv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/apis/v1alpha1"
+	apisv1alpha2 "github.com/kcp-dev/kcp/sdk/apis/apis/v1alpha2"
+)
+
+type CRConverterFactory struct {
+	delegate *conversion.CRConverterFactory
+	scheme   *runtime.Scheme
+}
+
+func NewCRConverterFactory(serviceResolver webhook.ServiceResolver, authResolverWrapper webhook.AuthenticationInfoResolverWrapper) (*CRConverterFactory, error) {
```
sttts, line 37: go doc

sdk/apis/apis/v1alpha2/types_apiexport.go
mjudeikis: I think lets remove it for now.

sdk/apis/apis/v1alpha2/types_apiexport.go
xrstf: On Slack you said not to do this in this PR, so now I'm confused.

sdk/apis/apis/v1alpha2/types_apiexport.go
xrstf: What default values do you expect here?

sdk/apis/apis/v1alpha2/types_apiexport.go
xrstf: @mjudeikis This is your proposal and idea. :grin:

sdk/apis/apis/v1alpha2/types_apiexport.go
xrstf: But on Slack we also discussed "of course" not to do this in this PR, so now I'm confused.

sdk/apis/apis/v1alpha2/types_apiexport.go
mjudeikis: Reference allows BYO virtual resource implementation. In example
```
   ref:
     kind: ReferenceObject
     resource: references.compute.corp.io
     name: instance-references 
```
It would be an implementor-created and maintained object, exposing VirtualWorkspace URL at status.URL:
```
apiVersion: references.compute.corp.io/v1alpha1
kind: ReferenceObject
metadata:
  name: instance-references 
spec:
  filters: sku=premium
status:
  endpoints:
   - url: https://192.168.3.14:7444/services/apiexport-resources/2zge3fnkzj3s8kmk/cowboys
   - url: https://192.168.3.14:7445/services/apiexport-resources/2zge3fnkzj3s8kmk/cowboys
```

This URL would be used in aggregated server mode. We can do this the same way we do now—one per shard. 

In addition, we could have a kcp driven implementation where only schema and source (workspace) are provided, and we replicate it via a cache server and create our own implementation. Idea was to follow mounts approach where URL is dynamically populated

sttts: This should move to EndpointSlices on day one, please. No endpoints in the object please.

And then we follow the same approach as with exports, similar controller and replication.

sttts: But am still not sure that is enough. We still need the replicated ReferenceObject, don't we?

This is kind of out-of-scope of this PR and maybe worth to discuss in a follow-up about virtual resources.

sdk/apis/apis/v1alpha2/types_apiexport_conversion_test.go
xrstf: I am not sure how fuzzing would be useful in this case. I tried to, with your help, look at the Kubernetes code to learn how it does fuzzing tests, but what I saw were

* All of the tests are based on internal types (i.e. a `__internal` API version). We do not have such internal types (i.e. we do not have a shared APIExport struct that is the combination of all APIExport versions).
* "Roundtrips" in the apitesting package from k/k are defined as "object -> JSON -> object", _not_ "v1 -> JSON -> v2". So I doubt that the roundtrip tests in k/k are a good example of how to fuzz version conversions.
* Even if we were to use fuzzing, how exactly does that help and what exactly is it testing? The conversion between v1alpha1 and v1alpha2 are mostly interesting for ResourceSchema conversions. This requires that the types have at least somewhat sensible values for schema, resources, names etc. -- the default fuzzing code in k/k however simply generates raw binary data. With binary data, none of the conversions can ever work, and all we could assert is that "all fuzzing tests must fail", which is not a useful assertion.
* Fuzzing makes sense to me for parsers, like JSON/YAML. Here I can see a clear usecase and a distinction between "invalid input" (good) and "program crashes" (bad, test should fail). However this does simply not apply to CRD conversion in Kubernetes.
* To use fuzzing, a smart(er) fuzzer/generator would be required, that understands how to generate resource schema names (a string with dots in it, basically). However when you think about what the generator needs to be able to generate, all the cornercases, it's easier to just hand roll some unit tests that express these corner cases, rather than implementing the corner cases in Go, and hoping to not make any mistakes and also to be exhaustive. The fuzzer would here be useless overhead, compared to handwritten unit tests.
* Looking deeper into the k/k tests, I even wonder what exactly Kubernetes is trying to test. That the JSON/YAML packages work?
* Even k/k uses handwritten tests to test version conversions, like here: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1/conversion_test.go

I think we should not use fuzzing for these tests.

sdk/apis/apis/v1alpha2/types_apiexport.go
xrstf: Okay, storage is now defaulted to CRD storage.

mjudeikis:
@sttts  wanna check this out again before tagging?

config/crds/apis.kcp.io_apiexports.yaml
```diff
@@ -479,6 +479,8 @@ spec:
                         "<version>.<name>.<group>".
                       type: string
                     storage:
+                      default:
+                        crd: {}
```
sttts, line 483: this is not enough. You need a default for storage as well (can be `{}`).

sttts: nvmd. Wrong indention.

config/crds/apis.kcp.io_apiexports.yaml
```diff
@@ -467,8 +467,16 @@ spec:
...
+                      description: Group is the API group of the resource.
+                      type: string
+                    name:
+                      description: Name is the name of the resource.
+                      type: string
                     schema:
-                      description: Schema is the name of the referenced APIResourceSchema.
+                      description: |-
+                        Schema is the name of the referenced APIResourceSchema. This must be of the format
+                        "<version>.<name>.<group>".
```
sttts, line 479: nit: version is misleading. Name does not really mean anything. Would use `<prefix>.<resource>.<group>`.

config/crds/apis.kcp.io_apiexports.yaml
```diff
@@ -489,12 +497,15 @@ spec:
                           type: object
                       type: object
                   required:
+                  - group
+                  - name
                   - schema
                   - storage
```
sttts, line 503: this is wrong. Storage has a default, but is not required. A client verifying this client-side does not see the default and will fail.

pkg/admission/apiexport/admission.go
```diff
@@ -99,5 +100,25 @@ func (e *APIExportAdmission) Validate(ctx context.Context, a admission.Attribute
...
+		}
+	}
+
+	return nil
+}
+
+func validateResourceSchema(resourceSchema apisv1alpha2.ResourceSchema, path *field.Path) *field.Error {
+	group := resourceSchema.Group
+	if group == "" {
+		group = "core"
```
sttts, line 115: does this mean group should have a `""` default? Is it documented in the API docs that empty group means core?

pkg/admission/apiexport/admission.go
```diff
@@ -99,5 +100,25 @@ func (e *APIExportAdmission) Validate(ctx context.Context, a admission.Attribute
...
+	}
+
+	expectedSuffix := fmt.Sprintf(".%s.%s", resourceSchema.Name, group)
```
sttts, line 118: this feel wrong: either call that field `resources` (then "name" makes sense), or call it `resourceSchemas` (I don't like that, but then this resource field must be called `resource` and the schema field must be `name`.

sdk/apis/apis/v1alpha2/types_apiexport.go
```diff
@@ -148,7 +149,12 @@ type APIExportSpec struct {
 }
 
 type ResourceSchema struct {
-	// Schema is the name of the referenced APIResourceSchema.
+	// Name is the name of the resource.
+	Name string `json:"name"`
```
sttts, line 153: this seems off, compare https://github.com/kcp-dev/kcp/pull/3318/commits/d5e6f866419346c71e16397929091b87108d8d8c#r2006280842.

Would rename the field to `spec.resource`. Then you naming makes sense.

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
```diff
@@ -0,0 +1,186 @@
...
+
+const (
+	resourceSchemasAnnotation = "apis.v1alpha2.kcp.io/resource-schemas"
+)
+
+func Convert_v1alpha2_APIExport_To_v1alpha1_APIExport(in *APIExport, out *apisv1alpha1.APIExport, s kubeconversion.Scope) error {
+	out.ObjectMeta = in.ObjectMeta
+
+	// before converting the spec, figure out which ResourceSchemas could not be represented in v1alpha1 and
+	// retain them via an annotation
```
sttts, line 37: surprising. Which are that?

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
```diff
@@ -0,0 +1,186 @@
...
+	"encoding/json"
+	"fmt"
+
+	kubeconversion "k8s.io/apimachinery/pkg/conversion"
+
+	apisv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/apis/v1alpha1"
+)
+
+const (
+	resourceSchemasAnnotation = "apis.v1alpha2.kcp.io/resource-schemas"
```
sttts, line 30: this must be internal. Then it is protected automatically. This is not meant to be user changeable, is it?

sttts: Actually, not sure my comment makes sense. Everybody can and must be able to change it, right? A conversion happens as normal user. We must validate it to make sense.

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
sttts: ic, this is in preparation of virtual resources? But does not yet apply?

sdk/apis/apis/v1alpha2/types_apiexport_conversion_test.go
```diff
@@ -0,0 +1,162 @@
+/*
+Copyright 2022 The KCP Authors.
```
sttts: 2025

sdk/apis/apis/v1alpha2/types_apiexport_conversion_test.go
sttts: > Even if we were to use fuzzing, how exactly does that help and what exactly is it testing?

You will learn after implementing it. There are always edge cases one does not expect ;-)

sttts: I don't see an annotation with overhangs in any of the tests. is that coming with virtual resources? Should there at least be an empty resource? And should the annotation be removed in v1alpha2? These are the things are tested in fuzzing.

sdk/apis/apis/v1alpha2/types_apiexport_conversion_test.go
```diff
@@ -0,0 +1,162 @@
...
+func TestConvertV1Alpha1APIExports(t *testing.T) {
+	testcases := []apisv1alpha1.APIExport{
+		{
+			Spec: apisv1alpha1.APIExportSpec{},
+		},
+		{
+			Spec: apisv1alpha1.APIExportSpec{
+				LatestResourceSchemas: []string{
+					"foo",
+					"bar",
```
sttts: these don't look like schemas.

sdk/apis/apis/v1alpha2/types_apiexport_conversion_test.go
```diff
@@ -0,0 +1,162 @@
...
+		{
+			Spec: APIExportSpec{},
+		},
+		{
+			Spec: APIExportSpec{
+				ResourceSchemas: []ResourceSchema{{
+					Schema: "foo",
+					Storage: ResourceSchemaStorage{
+						CRD: &ResourceSchemaStorageCRD{},
+					},
```
sttts, line 42: I don't see valid ResourceSchemas. This would be tested with fuzzing.

sdk/apis/apis/v1alpha2/types_apiexport_conversion_test.go
```diff
@@ -0,0 +1,162 @@
...
+					Schema: "foo",
+					Storage: ResourceSchemaStorage{
+						Virtual: &ResourceSchemaStorageVirtual{
+							Reference: corev1.TypedLocalObjectReference{
+								Kind: "test",
+							},
+						},
+					},
+				}},
+			},
```
sttts, line 75: how do we know that everything in the spec is copied over to v1alpha1 and back? This would be tested with fuzzing. We must have these tests. Somebody adds a field and forgets the conversion. Then we have dataloss 😞

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
```diff
@@ -0,0 +1,186 @@
...
+		if len(out.Annotations) == 0 {
+			out.Annotations = nil
+		}
+	}
+
+	return nil
+}
+
+func Convert_v1alpha1_APIExportSpec_To_v1alpha2_APIExportSpec(in *apisv1alpha1.APIExportSpec, out *APIExportSpec, s kubeconversion.Scope) error {
+	if in.Identity != nil {
```
sttts, line 146: this is untested

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
```diff
@@ -0,0 +1,186 @@
...
+		if err := Convert_v1alpha1_Identity_To_v1alpha2_Identity(in.Identity, out.Identity, s); err != nil {
+			return err
+		}
+	}
+
+	if in.MaximalPermissionPolicy != nil {
```
sttts, line 152: this is untested

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
```diff
@@ -0,0 +1,186 @@
...
+		if err := Convert_v1alpha1_MaximalPermissionPolicy_To_v1alpha2_MaximalPermissionPolicy(in.MaximalPermissionPolicy, out.MaximalPermissionPolicy, s); err != nil {
+			return err
+		}
+	}
+
+	if claims := in.PermissionClaims; claims != nil {
```
sttts, line 158: this is untested

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
```diff
@@ -0,0 +1,186 @@
...
+
+func Convert_v1alpha1_APIExport_To_v1alpha2_APIExport(in *apisv1alpha1.APIExport, out *APIExport, s kubeconversion.Scope) error {
+	if err := autoConvert_v1alpha1_APIExport_To_v1alpha2_APIExport(in, out, s); err != nil {
+		return err
+	}
+
+	if overhanging, ok := in.Annotations[resourceSchemasAnnotation]; ok {
+		resourceSchemas := []ResourceSchema{}
+		if err := json.Unmarshal([]byte(overhanging), &resourceSchemas); err != nil {
+			return fmt.Errorf("failed to decode schemas from JSON: %w", err)
```
sttts, line 123: do we have admission that checks this not to happen?

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+
+		if len(resourceSchemas) > 0 {
+			if out.Spec.ResourceSchemas == nil {
+				out.Spec.ResourceSchemas = []ResourceSchema{}
+			}
+
+			out.Spec.ResourceSchemas = append(out.Spec.ResourceSchemas, resourceSchemas...)
```
sttts, line 131: this can lead to duplicated without admission checking that there is no duplicate.

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+
+		delete(out.Annotations, resourceSchemasAnnotation)
```
sttts, line 134: All these mutations look suspicious. Wondering whether we might cache mutations here.

Look at the invariant defined here:

```
	// Convert attempts to convert one object into another, or returns an error. This
	// method does not mutate the in object, but the in and out object might share data structures,
	// i.e. the out object cannot be mutated without mutating the in object as well.
	// The context argument will be passed to all nested conversions.
	Convert(in, out, context interface{}) error
```

In other words, in and out share data structure. But when you mutate out, you must do a (hopefully shallow, avoid deep copies) copy.

config/crds/apis.kcp.io_apiexports.yaml
xrstf: I don't understand. If this is defaulted and we need some sort of storage defined for a resource schema, how exactly is storage then not a required field? How do we deal with resource schemas without a storage? Which, by not making it a require field, you imply is possible and valid.

sdk/apis/apis/v1alpha2/types_apiexport.go
mjudeikis: Renamed to resources, but object stays `ResourceSchema` due to fact we already have func named `Resource`

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
mjudeikis: yes, this is for one which we don't have "where to put in old API"

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
mjudeikis: This is fuzzing test where we check that we didn't modified the input itself, right? Or you had something else inmind with `cache` ?

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
mjudeikis: By untested, you mean if it's being converted roundtrip?

pkg/admission/apiexport/admission.go
mjudeikis: renamed

config/crds/apis.kcp.io_apiexports.yaml
mjudeikis: `this is wrong. Storage has a default, but is not required` - not sure what this means? Should it be optional if it has default? or what?

sttts: Can must not require defaulted fields. The client (kubectl or others) will fails with their client-side validation.

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
sttts: Yes. If I get these lines wrong, no unit test will fail (before roundtrip tests).

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
sttts: Right. If you roundtrip test checks the input, it is fine.

sdk/apis/apis/v1alpha2/types_apiexport.go
```diff
@@ -0,0 +1,329 @@
...
+	// +listMapKey=resource
+	PermissionClaims []PermissionClaim `json:"permissionClaims,omitempty"`
+}
+
+// ResourceSchema defines the resource schemas that are exposed with this APIExport.
+type ResourceSchema struct {
+	// Name is the name of the resource.
+	Name string `json:"name"`
+	// Group is the API group of the resource.
+	Group string `json:"group"`
```
sttts, line 156: this needs a work about core group.

sdk/apis/apis/fuzzer/fuzzer.go
```diff
@@ -0,0 +1,113 @@
...
+			c.FuzzNoCustom(r)
+			r.TypeMeta = metav1.TypeMeta{}
+			r.Kind = ""
+			r.APIVersion = ""
+		},
+		func(r *v1alpha1.APIExportSpec, c fuzz.Continue) {
+			c.FuzzNoCustom(r)
+
+			r.LatestResourceSchemas = []string{
+				"v1." + nonEmptyString(c.RandString) + "." + nonEmptyString(c.RandString),
```
sttts, line 55: this should be multiple, random number >= 1

sdk/apis/apis/fuzzer/fuzzer.go
```diff
@@ -0,0 +1,113 @@
...
+
+			r.ResourceSchemas = []v1alpha2.ResourceSchema{
+				{
+					Group:  group,
+					Name:   name,
+					Schema: schema,
+					Storage: v1alpha2.ResourceSchemaStorage{
+						CRD: &v1alpha2.ResourceSchemaStorageCRD{},
+					},
+				},
```
sttts, line 73: same here. Needs to be a slice with small random number of entries.

sdk/apis/apis/fuzzer/fuzzer.go
```diff
@@ -0,0 +1,113 @@
...
+		},
+		func(r *v1alpha1.APIExportList, c fuzz.Continue) {
+			c.FuzzNoCustom(r)
+			r.TypeMeta = metav1.TypeMeta{}
+			r.Kind = ""
+			r.APIVersion = ""
+		},
+		func(r *v1alpha1.APIResourceSchemaSpec, c fuzz.Continue) {
+			r.Conversion = &v1alpha1.CustomResourceConversion{
+				Strategy: v1alpha1.ConversionStrategyType("None"),
```
sttts, line 96: what about other strategies.

mjudeikis: As we need 1 or another added random of 2 + webhook random if webhook was picked.

sdk/apis/apis/fuzzer/fuzzer.go
mjudeikis: I think we dont enforce (and we should not) version to be v{1,2,3} so replaced this with string too.

sdk/apis/apis/v1alpha2/types_apiexport.go
mjudeikis: added

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
mjudeikis: added duplicate checking based on schema (not name)

config/crds/apis.kcp.io_apiexports.yaml
mjudeikis: reduced validation on `storage`

mjudeikis:
Most issues addressed. Just checkiner where I broke restmapper :/

pkg/admission/apiexport/admission.go
mjudeikis: Added as doc line in the api

mjudeikis:
/retest
Somehow can't replicate locally :/

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
mjudeikis: now we do.

mjudeikis:
/hold 
want to test one nit before merging, but should be good'ish now.

pkg/admission/apiexport/admission.go
```diff
@@ -115,10 +149,46 @@ func validateResourceSchema(resourceSchema apisv1alpha2.ResourceSchema, path *fi
...
 
 	return nil
 }
+
+func validateOverhangingAnnotations(_ context.Context, _ admission.Attributes, ae *apisv1alpha1.APIExport) error {
+	// TODO(mjudeikis): Remove this once we are sure that all APIExport objects are
+	// converted to v1alpha2.
+	if ae.Annotations == nil {
+		return nil
+	}
```
sttts, line 167: you don't need this. The if clause below goes through even with nil.

pkg/admission/apiexport/admission.go
```diff
@@ -62,23 +65,62 @@ var _ = admission.ValidationInterface(&APIExportAdmission{})
...
+		if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, ae); err != nil {
+			return fmt.Errorf("failed to convert unstructured to APIExport: %w", err)
+		}
+
+		// Before we convert to v1alpha2, we need to validate the annotations overhanging:
+		if err := validateOverhangingAnnotations(ctx, a, ae); err != nil {
+			return admission.NewForbidden(a, err)
+		}
+
+		var v2 *apisv1alpha2.APIExport
```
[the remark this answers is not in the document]
mjudeikis: will update in follow-up dear AI :D

PR #3319, opened 2025-03-05
🐛 do not create executable YAML files from crd-puller

xrstf:
## Summary
The crd-puller currently generates YAML files with mode 0777, which is an odd choice for YAML files. This PR changes the mode to 0644, something IMHO more sensible.

## Release Notes
```release-note
crd-puller will generate files with 0644 permissions instead of 0777.
```


embik (approved):
/approve

Issue #3289, opened 2025-02-09
bug: global service account flakes


mjudeikis:
### Describe the bug

We have flakes in Service account test now in global setup:

https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3256/pull-kcp-test-e2e-sharded/1888625034432352256

```
=== FAIL: test/e2e/authorizer TestServiceAccounts/Bound_service_token/Access_another_workspace_in_the_same_org (33.87s)
    serviceaccounts_test.go:186: Create workspace with the same name 
    serviceaccounts_test.go:187: Created root:universal workspace root:e2e-workspace-7rl6l:e2e-workspace-twx26 as /clusters/2up324saf6ctlqu4 on shard "root"
    serviceaccounts_test.go:195: Accessing workspace with the service account
    serviceaccounts_test.go:199: Giving the access to configmaps in the other workspace
    serviceaccounts_test.go:236: Accessing other workspace with the (there foreign) service account should eventually work because it is authenticated
    assertions.go:1902: Waiting for condition, but got: Unauthorized
    serviceaccounts_test.go:237: 
        	Error Trace:	/home/prow/go/src/github.com/kcp-dev/kcp/test/e2e/framework/util.go:352
        	            				/home/prow/go/src/github.com/kcp-dev/kcp/test/e2e/authorizer/serviceaccounts_test.go:237
        	Error:      	Condition never satisfied
        	Test:       	TestServiceAccounts/Bound_service_token/Access_another_workspace_in_the_same_org
```

This is basically due to fact it can't find this service account as its lives somewhere else :D 

Ticket to track when we will fix it. 

### Steps To Reproduce

Fix lookups. 

### Expected Behaviour

To work.

ntnn:
Can this be reopened?

This test is flaking both in CI and on my local machine:

https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3392/pull-kcp-test-e2e-shared/1918712190810984448

https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3392/pull-kcp-test-e2e-sharded/1918720468274122752

https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3392/pull-kcp-test-e2e-sharded/1918727553225330688

[...]

PR #3328, opened 2025-03-12
🌱 Add kube feature gate for global service account 

cnvergence:
## Summary

Hiding the ability to use a global ServiceAccount behind a feature flag until it will become ready with  https://github.com/kcp-dev/kcp/pull/3274.
This PR aims to unblock the release, bumping k8s fork with https://github.com/kcp-dev/kubernetes/pull/165.

## Related issue(s)

Part of https://github.com/kcp-dev/kcp/issues/3289 and https://github.com/kcp-dev/kcp/pull/3274

## Release Notes

```release-note
Provide a feature gate GlobalServiceAccount that enables cross-workspace ServiceAccount authorization (requires --service-account-lookup=false in sharded environments).
```


embik:
@cnvergence Can we please add a release note naming the feature gate and what it does, so we have this on the release notes?

pkg/authorization/resolver_test.go
```diff
@@ -128,16 +129,19 @@ func TestResolverWithWarrants(t *testing.T) {
 			wantResourceRules:    []authorizer.ResourceRuleInfo{getServices},
 			wantNonResourceRules: nil, // global service accounts do no work without a cluster.
 		},
+		// TODO(cnvergence): restore the commented lines once we drop the global service account feature flag
 		{
-			name:                 "service account with this cluster",
-			user:                 &user.DefaultInfo{Name: "system:serviceaccount:default:sa", Groups: []string{"system:serviceaccounts", user.AllAuthenticated}, Extra: map[string][]string{authserviceaccount.ClusterNameKey: {"this"}}},
-			wantResourceRules:    []authorizer.ResourceRuleInfo{getServices},
-			wantNonResourceRules: []authorizer.NonResourceRuleInfo{getReadyz},
+			name:              "service account with this cluster",
```
sttts, line 134: this doesn't look like a foreign service account. Why does it break?

sttts: Nvmd. It uses the global service account sa name.

pkg/authorization/resolver_test.go
```diff
@@ -128,16 +129,19 @@ func TestResolverWithWarrants(t *testing.T) {
 			wantResourceRules:    []authorizer.ResourceRuleInfo{getServices},
 			wantNonResourceRules: nil, // global service accounts do no work without a cluster.
 		},
+		// TODO(cnvergence): restore the commented lines once we drop the global service account feature flag
```
sttts, line 132: couldn't we just make this dependent on the feature gate? So the test is active again when the gate is on or the code here breaks when the gate goes away?

sttts: like add a `skip bool` field and make that dependent on the gate.

cnvergence: That also makes a perfect sense

embik (approved):
/approve

embik:
/retest

?

PR #3329, opened 2025-03-12
:seedling: Switch to golangci-lint gci for import sorting

sttts:
## Summary

gci is pretty much standard in many projects. Let's use it instead of some openshift tool.

Opinions about the order?

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


embik (approved):
/approve

PR #3331, opened 2025-03-14
:seedling: Update to Go 1.23.7

embik:
## Summary

Go 1.22 is [EOL](https://endoflife.date/go). It seems that it already didn't get some recent `net/http` fixes, and trying to update `golang.org/x/net` I ran into issues with Go 1.22 as well. We should upgrade to Go 1.23 before the release to make sure we release with a supported Go version and include the latest stdlib fixes.

## Related issue(s)

Fixes #

## Release Notes

```release-note
kcp is built with Go 1.23.7
```


embik:
/retest

flake?

PR #3335, opened 2025-03-15
:seedling: Make exposed testing package work outside

sttts:
## Summary

Some fixes and more exposed options to be usable outside. Successfully tested in github.com/kcp-dev/multicluster-provider 🎉

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/sdk/testing/server/config.go
+++ b/sdk/testing/server/config.go
@@ -31,29 +31,43 @@ type Config struct {
 }
 
 // Option a function that wish to modify a given kcp configuration.
-type Option func(*Config) *Config
+type Option func(*Config)
 
 // WithScratchDirectories adds custom scratch directories to a kcp configuration.
 func WithScratchDirectories(artifactDir, dataDir string) Option {
-	return func(cfg *Config) *Config {
+	return func(cfg *Config) {
 		cfg.ArtifactDir = artifactDir
 		cfg.DataDir = dataDir
-		return cfg
 	}
 }
 
 // WithCustomArguments applies provided arguments to a given kcp configuration.
 func WithCustomArguments(args ...string) Option {
-	return func(cfg *Config) *Config {
+	return func(cfg *Config) {
 		cfg.Args = args
-		return cfg
 	}
 }
 
-// WithClientCADir sets the client CA directory for a given kcp configuration.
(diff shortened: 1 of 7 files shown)


sdk/testing/server/fixture.go
```diff
@@ -91,8 +92,9 @@ func NewFixture(t *testing.T, cfgs ...Config) Fixture {
 	// Launch kcp servers and ensure they are ready before starting the test
 	start := time.Now()
 	t.Log("Starting kcp servers...")
-	wg := sync.WaitGroup{}
-	wg.Add(len(servers))
+	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
+	g, ctx := errgroup.WithContext(ctx)
```
sttts, line 97: For the reviewer: error groups fail when the first go routine fails. Before the `wg.Wait()` was blocking forever.

sdk/testing/server/fixture.go
```diff
@@ -250,6 +251,8 @@ func Command(executableName, identity string) []string {
 	binary := executableName
 	if binDir := os.Getenv(kcpBinariesDirEnvDir); binDir == "" && inKcp {
 		binary = filepath.Join(kcptestinghelpers.RepositoryBinDir(), executableName)
+	} else if binDir != "" {
+		binary = filepath.Join(binDir, executableName)
```
sttts, line 255: for the reviewer: had forgotten this case that adds the `KCP_BINARIES_DIR`.

sdk/testing/server/config.go
```diff
@@ -31,29 +31,43 @@ type Config struct {
 }
 
 // Option a function that wish to modify a given kcp configuration.
-type Option func(*Config) *Config
```
stevekuznetsov, line 34: For my understanding - what about this was blocking external users?

sttts: It wasn't. But it's also unnecessary and I have seen the non-functional style elsewhere.

sttts: Note: this code was not exposed before. So we are just cleaning up stuff on the go.

stevekuznetsov: Makes sense, thanks!

PR #3336, opened 2025-03-16
:seedling: Replace *testing.T with interface for integration with test frameworks

sttts:
## Summary

Ginkgo and others have their own `*testing.T` interface, implementing the majority of methods, but not all, and in particular not using the `testing.T` struct from Go. This PR replaces `*testing.T` with a smaller variant of the interface.

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


sdk/testing/server/testing.go
```diff
@@ -0,0 +1,33 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package server
+
+// TestingT is implemented by *testing.T and potentially other test frameworks.
+type TestingT interface {
```
mjudeikis, line 20: Why its duplicate with one in `sdk/testing/testing.go` ?

sdk/testing/helpers/testing.go
```diff
@@ -0,0 +1,27 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package helpers
+
+// TestingT is implemented by *testing.T and potentially other test frameworks.
+type TestingT interface {
```
mjudeikis, line 20: We nw have 3 versions of this (2 same and this). Not sure I understand why?

sttts: Because this is not an interface anybody consumes as a promise, but it is input, i.e. a requirement. Hence, it is defined where the code lives that consumes it.

PR #3337, opened 2025-03-18
:bug: cli/bind: fix parsing permission claims passed by cli args

gman0:
## Summary

This PR fixes parsing permission claims passed through `--{accept,reject}-permission-claim` flags.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/cli/pkg/bind/plugin/bind.go
+++ b/cli/pkg/bind/plugin/bind.go
@@ -215,7 +215,7 @@ func (b *BindOptions) Run(ctx context.Context) error {
 }
 
 func (b *BindOptions) parsePermissionClaim(claim string, accepted bool) error {
-	claimParts := strings.Split(claim, ".")
+	claimParts := strings.SplitN(claim, ".", 2)
 	if len(claimParts) != 2 {
 		return fmt.Errorf("invalid permission claim %q", claim)
 	}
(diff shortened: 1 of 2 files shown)


cli/pkg/bind/plugin/bind.go
```diff
@@ -221,8 +221,8 @@ func (b *BindOptions) parsePermissionClaim(claim string, accepted bool) error {
 	}
 
 	parsedClaim := apisv1alpha1.AcceptablePermissionClaim{}
-	resource := claimParts[0]
```
mjudeikis, line 224: Resource should be first `resource.group`. Why change this? Maybe doc/example was wrong

gman0: I see, in that case indeed the docs string is wrong: https://github.com/kcp-dev/kcp/blob/main/cli/pkg/bind/cmd/cmd.go#L35

embik: Looking at the code and the docstring, this will only work for the core API anyway, no? The code assumes the format `X.Y`, but you can create permission claims for groups with a dot as well, so the parsing here would probably fail at those (e.g. `apibindings.apis.kcp.io`).

gman0: Would `<group>/<resource>` be better?

embik: The syntax is fine, the parser just needs to take the element at index 0 as the resource name and concat the remaining elements into the group.

PR #3349, opened 2025-03-25
:bug: Bump WorkspaceType `APIResourceSchema` name

embik:
## Summary

This [Slack thread](https://kubernetes.slack.com/archives/C021U8WSAFK/p1742888913738159) mentions that we have a problem with the 0.26 to 0.27 upgrade with the WorkspaceType `APIResourceSchema`. The reason is that we changed the file in #3233 but it looks like changes in that PR were done manually with search and replace, which evaded detection by `apigen` and by any other validation logic we have.

This PR, intended to be backported to `release-0.27`, does two things: It manually resets the `APIResourceSchema` name in question so that upgrades can proceed.

It _also_ adds a small `hack/verify-apiresourceschemas.sh` script that compares the `APIResourceSchemas` against the pull request target branch and if there is a diff in the file, it errors if the object name is the same (a diff is fine, but any changes mean the object name should have been updated). The script is probably not ideal, but it should prevent us from making this mistake again.

## Related issue(s)

Fixes #

## Release Notes

```release-note
Fix `APIResourceSchema` name for the `WorkspaceTypes` resource to unblock upgrade from previous versions
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -224,6 +224,8 @@ verify-codegen: ## Verify codegen
 		exit 1; \
 	fi
 
+	./hack/verify-apiresourceschemas.sh
+
 .PHONY: imports
 imports: WHAT ?=
 imports: $(GOLANGCI_LINT) verify-go-versions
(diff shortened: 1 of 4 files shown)


embik:
/hold

Could this be in `apigen`? It requires a git base ref to compare against. WDYT?

PR #3354, opened 2025-03-26
:seedling: Pin GitHub Actions workflows to commits

embik:
## Summary

It's a [best practice](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) to pin workflows used in GitHub Actions to specific commits. We should do that.

In addition, I feel like our dependabot configuration doesn't work (is it maybe not enabled?), so I'm copying the one we are using in kubernetes-sigs/multicluster-runtime.

## Related issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/.github/workflows/pr-verifier.yaml
+++ b/.github/workflows/pr-verifier.yaml
@@ -16,8 +16,28 @@ jobs:
       pull-requests: read
     runs-on: ubuntu-latest
     steps:
-    - name: Verifier action
-      id: verifier
-      uses: kubernetes-sigs/kubebuilder-release-tools@012269a88fa4c034a0acf1ba84c26b195c0dbab4 # tag=v0.4.3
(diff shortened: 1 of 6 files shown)


.github/workflows/pr-verifier.yaml
```diff
@@ -10,14 +10,34 @@ on:
 
 jobs:
   verify:
-    name: verify PR contents
+    name: verify PR title
```
xrstf, line 13: I would keep the original title, this job has multiple steps of which the title is only one. Currently the only one, but still.

embik: fair enough!

xrstf (approved):
/approve

PR #3357, opened 2025-03-27
🐛 fix: mermaid mkdocs error and add make target to run mkdocs locally

cnvergence:
## Summary

Markdown docs were displayed properly for me in the editor, however they broke in the mkdocs website.
This PR fixes the mermaid structure and adds a local mkdocs target that serves the kcp.io docs locally.

## Release Notes

```release-note
NONE
```


Makefile
```diff
@@ -175,6 +175,12 @@ generate-api-docs: ## Generate api docs
 VENVDIR=$(abspath docs/venv)
 REQUIREMENTS_TXT=docs/requirements.txt
 
+.PHONY: local-docs
+local-docs: venv ## Run mkdocs serve
+	cd docs
+	. $(VENV)/activate; \
+	VENV=$(VENV) docs/scripts/run-mkdocs.sh
```
embik, line 182: Can't we run `mkdocs serve` from here directly? 🤔  Also, did you get the `serve-docs` target to run?

cnvergence: Right, it was leftover, we can run it from main directory :)
I did try but I think for editing locally and then viewing `mkdocs serve` is the better option
It is also explained in the `mike` github repo: https://github.com/jimporter/mike?tab=readme-ov-file#viewing-your-docs

embik (approved):
/approve

PR #3359, opened 2025-04-04
:seedling: Update go to v1.23.1

pepov:
## Summary

Fixes CVE-2024-34156 by updating go to its latest patch version

## Release notes

Fixes CVE-2024-34156 by updating go to its latest patch version

### DIFF
--- a/cli/go.mod
+++ b/cli/go.mod
@@ -1,6 +1,6 @@
 module github.com/kcp-dev/kcp/cli
 
-go 1.23.0
+go 1.23.1
 
 require (
 	github.com/google/go-cmp v0.6.0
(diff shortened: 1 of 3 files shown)


pepov:
I wasn't sure if I should create a PR against main first or can patch the release and will a forward port PR created, please advise! I approached this against the release branch as I would prefer to see this in a patch release if possible.

embik:
Hi @pepov, thank you for trying to kcp safe! We really appreciate that.

The `go` directive in `go.mod` files denotes the _minimum_ version needed to compile the module. We are actually building kcp containers and binaries with Go 1.23.7 already, see #3331. We tend to keep the minimum version at the minor release we are using and don't bump it for Go patch releases unless we are really using a language or stdlib feature added in a patch release.

I hope this helps, but be assured that kcp releases are already patched against https://github.com/advisories/GHSA-crqm-pwhx-j97f.

PR #3372, opened 2025-04-14
Add `test/integration`

ntnn:
## Summary

Adding a framework for integration tests.

Startup takes a long time - I am contemplating adding support for a shared embedded etcd (or an external etcd like in kube) if that significantly improves startup.

## What Type of PR Is This?

/kind feature

## Related Issue(s)

-

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/server/options/controllers.go
+++ b/pkg/server/options/controllers.go
@@ -43,18 +43,12 @@ type Controllers struct {
 	SAController kcmoptions.SAControllerOptions
 }
 
-var kcmDefaults *kcmoptions.KubeControllerManagerOptions
-
-func init() {
-	var err error
-
-	kcmDefaults, err = kcmoptions.NewKubeControllerManagerOptions()
+func NewControllers() *Controllers {
+	kcmDefaults, err := kcmoptions.NewKubeControllerManagerOptions()
 	if err != nil {
 		panic(err)
 	}
-}
 
-func NewControllers() *Controllers {
 	return &Controllers{
 		EnableAll: true,
 
--- a/sdk/testing/server/fixture.go
+++ b/sdk/testing/server/fixture.go
@@ -310,9 +310,6 @@ func (c *kcpServer) Run(t TestingT) error {
 }
 
 func (c *kcpServer) Stop() {
-	c.lock.Lock()
-	defer c.lock.Unlock()
-
 	if c.cancel == nil {
 		return
 	}
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -95,6 +95,33 @@ presubmits:
               memory: 4Gi
               cpu: 2
 
+  - name: pull-kcp-test-integration
+    decorate: true
+    # only run integration tests if code changed.
+    run_if_changed: "(cmd|config|pkg|sdk|test|go.mod|go.sum|Makefile|.prow.yaml)"
+    clone_uri: "https://github.com/kcp-dev/kcp"
+    labels:
+      preset-goproxy: "true"
+    spec:
+      containers:
+        - image: ghcr.io/kcp-dev/infra/build:1.23.7-2
+          command:
+            - ./hack/run-with-prow.sh
+            - ./hack/run-with-prometheus.sh
+            - make
+            - test-integration
+          env:
+            - name: USE_GOTESTSUM
+              value: '1'
+            - name: KUBE_CACHE_MUTATION_DETECTOR
+              value: '1'
+            - name: E2E_PARALLELISM
+              value: '3'
+          resources:
+            requests:
+              memory: 6Gi
+              cpu: 4
+
   - name: pull-kcp-test-e2e
     decorate: true
     # only run e2e tests if code changed.
--- a/Makefile
+++ b/Makefile
@@ -368,8 +368,16 @@ endif
 test: WHAT ?= ./...
 # We will need to move into the sub package, of sdk to run those tests.
 test: ## Run tests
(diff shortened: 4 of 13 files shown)


pkg/server/options/controllers.go
```diff
@@ -43,18 +43,12 @@ type Controllers struct {
...
-
-func init() {
-	var err error
-
-	kcmDefaults, err = kcmoptions.NewKubeControllerManagerOptions()
+func NewControllers() *Controllers {
+	kcmDefaults, err := kcmoptions.NewKubeControllerManagerOptions()
 	if err != nil {
 		panic(err)
 	}
```
ntnn, line 50: I'd prefer to bubble up the error, but that'd require to add error handling to every `NewOptions` that is calling it:

https://github.com/kcp-dev/kcp/blob/90605ce2317922cf3580cab2743c929a76c4541a/pkg/server/options/options.go#L99

https://github.com/kcp-dev/kcp/blob/90605ce2317922cf3580cab2743c929a76c4541a/cmd/kcp/options/options.go#L41

https://github.com/kcp-dev/kcp/blob/90605ce2317922cf3580cab2743c929a76c4541a/cmd/kcp/kcp.go#L90

Which would be a good change - but would also touch a lot of the codebase.

ntnn: I'd rather to that in a separate PR tbh.

ntnn:
Tests with race condition testing fails. I think kube also doesn't race test their integration tests.

embik:
> Startup takes a long time - I am contemplating adding support for a shared embedded etcd (or an external etcd like in kube) if that significantly improves startup.

This is kind of a known problem with starting kcp the first time, so we should probably tackle this in kcp itself, not the integration test framework.

ntnn:
Yeah, I just measured the startup times - etcd takes less than a second, kcp phase1 takes roughly 22s. And using a shared kcd for the integration tests would make less sense imho - if the kcp instance could be shared without tests affecting each other they could likely also be e2e tests.

ntnn:
Running with race detection works. Root cause was that this method could be called from multiple goroutines if no `DialContext` is set: 
https://github.com/kcp-dev/kcp/blob/a2014d6bc9c7585fa55dc1d1b4f8476a805e8b86/pkg/network/dialer.go#L39-L48

This originates from the server config here:
https://github.com/kcp-dev/kcp/blob/a2014d6bc9c7585fa55dc1d1b4f8476a805e8b86/pkg/server/config.go#L208-L210

So either the wrapper function is wrapped with a mutex to prevent concurrent writes or the wrapper function acts atomically.
I don't think the GenericAPIServer can be persuaded to call this initially.

I don't particulalry like either option. Another option could be to just remove that config - TCP timeout should be 10m anyhow iirc.

ntnn:
Found out why `.Transport` isn't used. When this is set this if takes hold:

https://github.com/kubernetes/kubernetes/blob/30469e180361d7da07b0fee6d47c776fa2cf3e86/staging/src/k8s.io/client-go/transport/transport.go#L34-L40

```go
// New returns an http.RoundTripper that will provide the authentication
// or transport level security defined by the provided Config.
func New(config *Config) (http.RoundTripper, error) {
	// Set transport level security
	if config.Transport != nil && (config.HasCA() || config.HasCertAuth() || config.HasCertCallback() || config.TLS.Insecure) {
		return nil, fmt.Errorf("using a custom transport with TLS certificate options or the insecure flag is not allowed")
	}
```

So the TLS configuration would have to be removed, which then causes connections to fail due to the missing TLS configuration:

```
E0415 14:28:52.537341   14738 storage_rbac.go:191] "Unhandled Error" err="unable to initialize clusterroles: Get \"https://127.0.0.1:57059/clusters/system:admin/apis/rbac.authorization.k8s.io/v1/clusterroles\": tls: failed to verify certificate: x509: certificate signed by unknown authority" logger="UnhandledError"
F0415 14:28:52.538522   14738 hooks.go:210] PostStartHook "kcp-bootstrap-policy" failed: unable to initialize roles: timed out waiting for the condition
FAIL    github.com/kcp-dev/kcp/test/integration/framework       39.978s

```

_Technically_ a bug, but there's even more oddities in that library - e.g. when the `.Transport` is set the wrappers are not applied.

For a fix on the KCP side for now we'll have to go with either of the workarounds.

pkg/server/options/controllers.go
mjudeikis: Panic here is not that bad. It at the point of startup config validation so its quick feedback. Not something happening at runtime 🤷

ntnn:
Marking as WIP as I want to change how the server is started

ntnn:
The goleak ignore list is pretty long but I'm fairly certain that quite a few are just downstream of other issues.

sdk/testing/server/fixture.go
```diff
@@ -310,9 +310,6 @@ func (c *kcpServer) Run(t TestingT) error {
 }
 
 func (c *kcpServer) Stop() {
-	c.lock.Lock()
-	defer c.lock.Unlock()
```
embik, line 314: Just a minor q, why are we removing this?

ntnn: That was this commit: https://github.com/kcp-dev/kcp/pull/3372/commits/206456112353c6dfd873ba1a189898b0f826b990
The lock is removed because `.cancel` is locking so calling `.Stop` was actually producing a deadlock :D

embik (approved):
LGTM, happy to also approve if none of the other usual suspects in this PR (@sttts and @mjudeikis) want to take a look.

Issue #3255, opened 2025-01-18
feature: Extend permission claims with verbs


mjudeikis:
### Feature Description

Extend permissions claims with verbs:

Description here: https://docs.google.com/document/d/1toEU8_tldfc-kIWDgKfROyPVJGhhWDrr2cU0ayRj76I/edit?tab=t.0#heading=h.24igyjbjpe3n

### Proposed Solution

https://docs.google.com/document/d/1toEU8_tldfc-kIWDgKfROyPVJGhhWDrr2cU0ayRj76I/edit?tab=t.0#heading=h.24igyjbjpe3n 

### Want to contribute?

- [ ] I would like to work on this issue.

xmudrii:
I would like to give it a try!
/assign

xmudrii:
The implementation PR has been merged, but we're yet to:
- Discuss how do we want to handle some edge cases (scheduled for the kcp community meeting on May 22)
- Update docs

I'll keep this issue open until these items are solved.

xmudrii:
This has been fully implemented and documented
/close

PR #3384, opened 2025-04-25
Extend `apis.kcp.io/v1alpha2` with `APIBinding`

xmudrii:
## Summary

This PR extends the `apis.kcp.io/v1alpha2` with `APIBindings` and related types. This new version is a 1:1 copy of the current `v1alpha1` API at the moment. The new `v1alpha2` version will be changed after this PR is merged (to make this PR easier to review) as part of implementing #3255 

All the code was updated to refer to `v1alpha2.APIBindings`.

https://github.com/kcp-dev/kcp/pull/3318 has been used as the inspiration for this PR.

## What Type of PR Is This?

/kind feature

## Related Issue(s)

None

## Release Notes
```release-note
Extended `apis.kcp.io/v1alpha2` with `APIBinding`
```

### DIFF
--- a/config/crds/apis.kcp.io_apibindings.yaml
+++ b/config/crds/apis.kcp.io_apibindings.yaml
@@ -5,6 +5,8 @@ metadata:
     controller-gen.kubebuilder.io/version: v0.17.3
   name: apibindings.apis.kcp.io
 spec:
+  conversion:
+    strategy: None
   group: apis.kcp.io
   names:
     categories:
--- a/cli/pkg/bind/plugin/bind.go
+++ b/cli/pkg/bind/plugin/bind.go
@@ -35,7 +35,7 @@ import (
 
 	"github.com/kcp-dev/kcp/cli/pkg/base"
 	pluginhelpers "github.com/kcp-dev/kcp/cli/pkg/helpers"
-	apisv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/apis/v1alpha1"
+	apisv1alpha2 "github.com/kcp-dev/kcp/sdk/apis/apis/v1alpha2"
 	kcpclientset "github.com/kcp-dev/kcp/sdk/client/clientset/versioned/cluster"
 )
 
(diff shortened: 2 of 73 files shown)


sttts:
Sgtm.

config/crds/apis.kcp.io_apibindings.yaml
```diff
@@ -429,6 +431,420 @@ spec:
         - spec
         type: object
     served: true
+    storage: false
```
sttts, line 434: we are breaking rollback with this

sttts:
While in general this looks good, the timing is questionable. Why don't we first get the API in place properly with the field we want, and AFTER that do the "now everything will use it" PR ? Seems to be much more natural rather than having a limbo phase where the version is not ready, but already used.

mjudeikis:
> While in general this looks good, the timing is questionable. Why don't we first get the API in place properly with the fields we want, and AFTER that do the "now everything will use it" PR ? Seems to be much more natural rather than having a limbo phase where the version is not ready, but already used.

I think the bot works IF we can get the next PR "chained to this PR " and merge them together. Would want to see "implementation" first before start merging.

config/crds/apis.kcp.io_apibindings.yaml
xmudrii: @sttts I believe this is coming from: https://github.com/kcp-dev/kcp/blob/b564d605b6fe03db57eb2d703aefd549f9bb891a/sdk/apis/apis/v1alpha2/types_apibinding.go#L34

I took that from APIExports v1alpha2: https://github.com/kcp-dev/kcp/blob/e2ba960b69964aa42e6923710f766fce62dc2024/sdk/apis/apis/v1alpha2/types_apiexport.go#L48

I don't know if this is needed for APIExports and not needed for APIBindings, but my thought was to stay consistent. Given that, this would break rollback for APIExports too, right?

Do you have any alternative idea what to do about this?

xmudrii:
> While in general this looks good, the timing is questionable. Why don't we first get the API in place properly with the fields we want, and AFTER that do the "now everything will use it" PR ? Seems to be much more natural rather than having a limbo phase where the version is not ready, but already used.

I'm fine with that, I'm already working on applying the implementation to v1alpha2 and I'll do as @mjudeikis proposed.

config/crds/apis.kcp.io_apibindings.yaml
xmudrii: @sttts I discussed this with @xrstf (since he did the APIExports/v1alpha2 implementation) and came to the following realisation. I think the idea here is that we might change this API in the backwards incompatible way especially in the development phase (i.e. until the v0.27.0 release). Making such changes while the version is the storage version means it could break reading from the etcd, as the API server could run into fields and types that it can't recognize. And implementing a migration mechanism for that case specifically sounds like lots of work, especially that we already have conversion from v1alpha1 to v1alpha2.

That said, I think the idea is that the **previous** version should be the storage version at least until just before the release or until we're confident into the new API. In this case, both APIBindings and APIExports should use v1alpha1 as the storage version.

Did I get this correctly?

xmudrii:
@sttts @mjudeikis I created #3402 based on this PR with implementation for #3255. The PR is in the WIP state because I'm yet to add some more tests, but other than that, the implementation is complete and ready for review.

mjudeikis:
/lgtm 
/approve

Usage will follow in minutes from another PR. So this is no-op for now.

PR #3388, opened 2025-04-29
Update sdk/testing/server

ntnn:
## Summary

Consolidated running in-process and external test instances into `KcpRunner` and consolidated using the existing `Config` rather than duplicating the same values into the server.

Also passing the `Config` to the runner allows adding more features without changing the interface.

The important part was enabling stopping the server for in-process instances for leak tests.

~~The only breaking change should be that the servers will listen on localhost by default, rather than all interfaces.~~

## What Type of PR Is This?

/kind cleanup
/kind feature

## Related Issue(s)

-

## Release Notes

```release-note
Added `sdk/testing/server.ContextRunInProcessFunc`
Deprecated `sdk/testing/server.RunInProcessFunc`
```

### DIFF
--- a/sdk/testing/server/config.go
+++ b/sdk/testing/server/config.go
@@ -16,6 +16,8 @@ limitations under the License.
 
 package server
 
+import "path/filepath"
+
 // Config qualify a kcp server to start
 //
 // Deprecated for use outside this package. Prefer PrivateKcpServer().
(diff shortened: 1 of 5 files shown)


ntnn:
Not sure if any of the changes breaks stuff for prow, at least running the tests locally is not producing any errors.

embik:
Let's see!

/ok-to-test

ntnn:
Ah no, the server is panicing: `        fatal error: concurrent map writes`

ntnn:
Took out the breaking change with settings the interface to localhost - I saw some erroring tests due to TLS issues because the certificates have the wrong subject/SAN. 
I put that onto my TODO list as that would touch more than I'd like to in this PR (also I didn't immediately see where the "wrong" sub/SAN came from :D).

Issue #3320, opened 2025-03-08
feature: statis root APIExport identities hash'es


mjudeikis:
### Feature Description

I want to have static root identity hashes. Either for testing or some specific production usecase.

### Proposed Solution

Add new flag to main server:
```
--core-identity-secret <some-file-name> 
```
and load yaml file in the form of:
```
export-name: <hash>
```

if provided - it should be used as values

### Want to contribute?

- [ ] I would like to work on this issue.

gman0:
I can give this a go,

/assign

mjudeikis:
As discussed in the call: 
Lets just do the `name and value` yaml files and nothing else. Simple and no much other things

PR #3400, opened 2025-05-07
Bootstrap with pre-defined identities

gman0:
## Summary

This PR adds a way to inject identities from a file while performing phase0 bootstrapping, making it possible to override core identities with pre-defined ones.

The file is expected to by YAML-formatted, passed via `--root-identities-file` kcp command line flag; its structure is rather simple, see the linked issue for reference.

## What Type of PR Is This?

/kind feature

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3320

## Release Notes

```release-note
Adds a new --root-identities-file kcp cli flag used to bootstrap APIExport identities inside the root workspace
```

### DIFF
--- a/pkg/server/options/options.go
+++ b/pkg/server/options/options.go
@@ -67,6 +67,7 @@ type ExtraOptions struct {
 	LogicalClusterAdminKubeconfig         string
 	ExternalLogicalClusterAdminKubeconfig string
 	ConversionCELTransformationTimeout    time.Duration
+	RootIdentitiesFile                    string
 	BatteriesIncluded                     []string
 	// DEVELOPMENT ONLY. AdditionalMappingsFile is the path to a file that contains additional mappings
 	// for the mini-front-proxy to use. The file should be in the format of the
@@ -166,6 +167,7 @@ func (o *Options) AddFlags(fss *cliflag.NamedFlagSets) {
 	fs.StringVar(&o.Extra.ShardClientKeyFile, "shard-client-key-file", o.Extra.ShardClientKeyFile, "Path to a client certificate key file the shard uses to communicate with other system components.")
 	fs.StringVar(&o.Extra.LogicalClusterAdminKubeconfig, "logical-cluster-admin-kubeconfig", o.Extra.LogicalClusterAdminKubeconfig, "Kubeconfig holding system:kcp:logical-cluster-admin credentials for connecting to other shards. Defaults to the loopback client")
 	fs.StringVar(&o.Extra.ExternalLogicalClusterAdminKubeconfig, "external-logical-cluster-admin-kubeconfig", o.Extra.ExternalLogicalClusterAdminKubeconfig, "Kubeconfig holding system:kcp:external-logical-cluster-admin credentials for connecting to the external address (e.g. the front-proxy). Defaults to the loopback client")
+	fs.StringVar(&o.Extra.RootIdentitiesFile, "root-identities-file", "", "Path to a YAML file used to bootstrap APIExport identities inside the root workspace. The YAML file must be structured as {\"identities\": [ {\"export\": \"<APIExport name>\", \"identity\": \"<APIExport identity>\"}, ... ]}. If a secret with matching APIExport name already exists inside kcp-system namespace, it will be left unchanged. Defaults to empty, i.e. no identities are bootstrapped.")
 
 	fs.BoolVar(&o.Extra.ExperimentalBindFreePort, "experimental-bind-free-port", o.Extra.ExperimentalBindFreePort, "Bind to a free port. --secure-port must be 0. Use the admin.kubeconfig to extract the chosen port.")
 	fs.MarkHidden("experimental-bind-free-port") //nolint:errcheck
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -47,6 +47,8 @@ import (
 	"github.com/kcp-dev/logicalcluster/v3"
 
 	configroot "github.com/kcp-dev/kcp/config/root"
+	configrootidentities "github.com/kcp-dev/kcp/config/root-identities"
+	configrootidentitiesns "github.com/kcp-dev/kcp/config/root-identities/namespace"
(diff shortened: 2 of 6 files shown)


config/root-core-identities/bootstrap.go
```diff
@@ -0,0 +1,74 @@
...
+	identitiesFilename string,
+) error {
+	var coreIdentities coreIdentitiesFile
+	bs, err := os.ReadFile(identitiesFilename)
+	if err != nil {
+		return fmt.Errorf("failed to read core identities file %s: %v", identitiesFilename, err)
+	}
+	if err = yaml.UnmarshalStrict(bs, &coreIdentities); err != nil {
+		return fmt.Errorf("failed to parse YAML file %s: %v", identitiesFilename, err)
+	}
```
ntnn, line 61:
```suggestion
	bs, err := os.ReadFile(identitiesFilename)
	if err != nil {
		return fmt.Errorf("failed to read core identities file %s: %v", identitiesFilename, err)
	}
	var coreIdentities coreIdentitiesFile
	if err := yaml.UnmarshalStrict(bs, &coreIdentities); err != nil {
		return fmt.Errorf("failed to parse YAML file %s: %v", identitiesFilename, err)
	}
```

pkg/server/server.go
```diff
@@ -434,6 +436,31 @@ func (s *Server) Run(ctx context.Context) error {
...
+					logger.Error(err, "failed to bootstrap identity secrets namespace")
+					return nil // don't klog.Fatal. This only happens when context is cancelled.
+				}
+				logger.Info("bootstrapped system namespace into root workspace", "file", s.Options.Extra.CoreIdentitiesFile)
+				logger.Info("injecting core identities into root workspace", "file", s.Options.Extra.CoreIdentitiesFile)
+				if err := configrootcoreidentities.Bootstrap(hookCtx,
+					s.BootstrapApiExtensionsClusterClient.Cluster(core.RootCluster.Path()).Discovery(),
+					s.DynamicClusterClient.Cluster(core.RootCluster.Path()),
+					s.KubeClusterClient.Cluster(core.RootCluster.Path()),
+					s.Options.Extra.CoreIdentitiesFile,
```
ntnn, line 456: E.g. here the file is read `pkg/server` and then the contents are handed off:
https://github.com/kcp-dev/kcp/blob/e2ba960b69964aa42e6923710f766fce62dc2024/pkg/server/config.go#L369-L389

Might also make it easier to test `config` without having to write the contents out to a file rather than just passing the bytes.

ntnn (changes requested):

pkg/server/options/options.go
```diff
@@ -166,6 +167,7 @@ func (o *Options) AddFlags(fss *cliflag.NamedFlagSets) {
 	fs.StringVar(&o.Extra.ShardClientKeyFile, "shard-client-key-file", o.Extra.ShardClientKeyFile, "Path to a client certificate key file the shard uses to communicate with other system components.")
 	fs.StringVar(&o.Extra.LogicalClusterAdminKubeconfig, "logical-cluster-admin-kubeconfig", o.Extra.LogicalClusterAdminKubeconfig, "Kubeconfig holding system:kcp:logical-cluster-admin credentials for connecting to other shards. Defaults to the loopback client")
 	fs.StringVar(&o.Extra.ExternalLogicalClusterAdminKubeconfig, "external-logical-cluster-admin-kubeconfig", o.Extra.ExternalLogicalClusterAdminKubeconfig, "Kubeconfig holding system:kcp:external-logical-cluster-admin credentials for connecting to the external address (e.g. the front-proxy). Defaults to the loopback client")
+	fs.StringVar(&o.Extra.CoreIdentitiesFile, "core-identities-file", "", "Path to a YAML file with APIExport - Identity secret Identity key pairs. Defaults to empty, i.e. all identities are generated.")
```
sttts, line 170: the format must be documented. In the docs, and If it's super simple, then here too.

sttts, line 170:
```suggestion
	fs.StringVar(&o.Extra.CoreIdentitiesFile, "core-identities-file", "", "Path to a YAML file with APIExport - identity secret identity key pairs. Defaults to empty, i.e. all identities are generated.")
```

sttts: Maybe you mean that in the description. Maybe `.... with a map where keys are X and values are Y. What are the keys? What is a identity secret? Are these write 1:1 to the secret of the same name? Try to be clear and precise.

sttts: Also it must be very clear what happens if the secrets are pre-existing.

config/root-core-identities/identity-secret-namespace/bootstrap.go
```diff
@@ -0,0 +1,43 @@
+/*
+Copyright 2022 The KCP Authors.
```
sttts, line 2: 2025

config/root-core-identities/bootstrap.go
gman0: Done, thanks!

pkg/server/server.go
gman0: Good point, reading in the caller function now.

pkg/server/options/options.go
gman0: Just for context, I decided to skip-on-exists when creating the secrets so that we don't break existing APIExports.

gman0:
I've renamed `core-identities-file` to `root-identities-file` because with current implementation of this feature, nothing stops the user to bootstrap identities for exports outside of `*.kcp.io`. Also, AFAIK we don't have a clear definition of what "core" refers to in this context, and so "root" is more fitting -- especially as it refers to the fact that the bootstrapping is happening inside the root workspace. Let me know if you disagree, I can change it back.

gman0:
The format suggested in https://github.com/kcp-dev/kcp/issues/3320 (and the one that's currently in use) is a bit more verbose than it has to be. Alternatively we could use a simple dictionary of `"<APIExport name>": "<APIExport identity>"`. This however would make it harder should we ever want to change/add fields (e.g. namespace), so I kept this more verbose option.

PR #3402, opened 2025-05-09
Implement support for verbs in PermissionClaims

### LINKED ISSUE #3255: feature: Extend permission claims with verbs

xmudrii:
## Summary

This PR implements verbs support for PermissionClaims in APIExports and APIBindings as described in #3255. This PR and its implementation require the v1alpha2 API version of APIBindings (#3384).

Prior to this PR, it wasn't possible to limit the access scope (i.e. verbs) for resources provided via the PermissionClaims mechanism. Providing access to resources via PermissionClaims always assumed all verbs (`*`), i.e. being able to do anything with those resources. This was seen as a problem because a consumer often wants to give only read-only access to resources to the service provider.

Starting with this PR, users are **required** to define verbs for all PermissionClaims in both APIExports and APIBindings. This only affects the v1alpha2 version of APIExports and APIBindings. The v1alpha1 version does **not** have support for verbs, instead, the prior behavior is in effect (i.e. all verbs are allowed).

Example permission claim:

```yaml
    - group: ""
      resource: "secrets"
      all: true
      verbs: ["get", "list", "create"]
      state: Accepted
```

The verbs field follows the same semantics as verbs in the native Roles/ClusterRoles:

- `verbs: null` and `verbs: []` are **not** permitted
- verbs must have at least one non-empty element
- there are no validation on the elements themselves (i.e. you can put anything as a verb)

This is implemented by taping into the binding authorizer in the APIExport Virtual Workspace. The authorizer has been extended to check if the requested verb is in the set of the allowed verbs. However, it's important to note that the set of the allowed verbs is calculated by making an intersection between verbs in the APIExport and verbs in the APIBinding.

There are at least two reasons for introducing this intersection behavior:

- We want to encourage users to properly document their needs (in this case verbs) in APIExports
- We want to encourage a good security hygiene by implementing the least privilege access mechanism

The PR implements a roundtrip conversion between v1alpha2 and v1alpha for PermissionClaims (i.e. for both APIExports and APIBindings). If there non-wildcard verbs in PermissionClaims, these PermissionClaims will be put in an annotation upon converting from v1alpha2 to v1alpha1. When converting from v1alpha1 to v1alpha2, kcp will read that annotation if present and properly populate verbs for these PermissionClaims. Similarly, also when converting from v1alpha1 to v1alpha2, PermissionClaims that don't have verbs (`verbs == nil`) will have verbs defaulted to `["*"]`.

To manually test the PR, you can follow the steps from the `TestAPIBindingPermissionClaimsVerbs` test case.

## What Type of PR Is This?

/kind feature
/kind api-change

## Related Issue(s)

xref #3255

## Release Notes
```release-note
Implement support for verbs in PermissionClaims
```


sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
```diff
@@ -239,3 +281,71 @@ func Convert_v1alpha1_LatestResourceSchema_To_v1alpha2_ResourceSchema(in []strin
...
+func Convert_v1alpha2_APIBinding_To_v1alpha1_APIBinding(in *APIBinding, out *apisv1alpha1.APIBinding, s kubeconversion.Scope) error {
+	out.ObjectMeta = in.ObjectMeta
+
+	if len(in.Spec.PermissionClaims) > 0 {
+		encoded, err := json.Marshal(in.Spec.PermissionClaims)
+		if err != nil {
+			return fmt.Errorf("failed to encode claims as JSON: %w", err)
+		}
+
+		if out.Annotations == nil {
```
mjudeikis, line 294: same here as above

pkg/virtual/apiexport/authorizer/binding.go
```diff
@@ -35,18 +37,25 @@ import (
...
-func NewBoundAPIAuthorizer(delegate authorizer.Authorizer, apiBindingInformer apisv1alpha2informers.APIBindingClusterInformer, kubeClusterClient kcpkubernetesclientset.ClusterInterface) authorizer.Authorizer {
+func NewBoundAPIAuthorizer(delegate authorizer.Authorizer, apiBindingInformer apisv1alpha2informers.APIBindingClusterInformer, apiExportInformer apisv1alpha2informers.APIExportClusterInformer, kubeClusterClient kcpkubernetesclientset.ClusterInterface) authorizer.Authorizer {
+	apiExportLister := apiExportInformer.Lister()
 	apiBindingLister := apiBindingInformer.Lister()
 
 	return &boundAPIAuthorizer{
 		delegate: delegate,
+		getAPIExport: func(clusterName, apiExportName string) (*apisv1alpha2.APIExport, error) {
+			return apiExportLister.Cluster(logicalcluster.Name(clusterName)).Get(apiExportName)
+		},
```
mjudeikis, line 58: Remind me, will this work if APIExport is on another shard? Is Lister here cache aware?

pkg/virtual/apiexport/authorizer/binding_test.go
```diff
@@ -0,0 +1,517 @@
...
+							},
+						},
+					},
+				}, nil
+			},
+			expectedDecision: authorizer.DecisionDeny,
+			expectedReason:   "failed to find suitable reason to allow access in APIBinding",
+		},
+		{
+			name: "list request to the API bound via permission claims but not allowing list (2)",
```
mjudeikis, line 390: Why do we have 3 test for same? can you make description bit more detail?

pkg/virtual/apiexport/authorizer/binding_test.go
```diff
@@ -0,0 +1,517 @@
...
+							},
+						},
+					},
+				}, nil
+			},
+			expectedDecision: authorizer.DecisionDeny,
+			expectedReason:   "failed to find suitable reason to allow access in APIBinding",
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
```
mjudeikis, line 496: I think we are missing tests for:
1. I give/consent more verbs/permissions than apiexport needs - should allow
2. custom verbs like "connect" or "proxy".

sdk/apis/apis/fuzzer/fuzzer.go
```diff
@@ -71,6 +71,22 @@ func Funcs(codecs runtimeserializer.CodecFactory) []interface{} {
 					},
 				})
 			}
+			r.PermissionClaims = nil
+			for i := 0; i < c.Intn(5); i++ {
+				group := nonEmptyString(c.RandString)
+				resource := nonEmptyString(c.RandString)
+				identityHash := nonEmptyString(c.RandString)
+				verb := nonEmptyString(c.RandString)
```
mjudeikis, line 79: Can we add this to have more than 1?

sdk/apis/roundtrip_test.go
```diff
@@ -131,7 +132,7 @@ func testGenericConversion[V1 runtime.Object, V2 runtime.Object](
 		err = toV1(intermediate, result, nil)
 		require.NoError(t, err)
 
-		require.Equal(t, original, result)
+		require.True(t, apiequality.Semantic.DeepEqual(original, result), "expects original to equal result")
 		require.Equal(t, originalCopy, original)
```
mjudeikis, line 136: Lets make this one same as above too.

pkg/virtual/apiexport/authorizer/binding_test.go
xmudrii: If the APIExport author removes a permission, that permission is no longer allowed even though it's still allowed via APIBinding. Only permissions that are allowed both in APIExport and APIBinding are allowed/permitted. However, I don't see (in the ideal scenario) how removing some permissions in APIExport might break things down. Sure, if they remove permissions that are still actually used, but then don't remove those permissions, I think that's the behavior that we want to encourage here.

test/e2e/apibinding/apibinding_permissionclaims_test.go
```diff
@@ -291,3 +290,128 @@ func bindConsumerToProvider(ctx context.Context, t *testing.T, consumerWorkspace
 	require.NoError(t, err, "error retrieving consumer workspace %q API discovery", consumerWorkspace)
 	require.True(t, resourceExists(resources, "cowboys"), "consumer workspace %q discovery is missing cowboys resource", consumerWorkspace)
 }
+
+func TestAPIBindingPermissionClaimsVerbs(t *testing.T) {
```
mjudeikis, line 294: Just for sake of sanity - can we extend this test to have mixed permissions:
1. I can create configMap to start with, but not a secret
2. Updated Export allows creating secrets now too
3. User removed permission - Export should still function, just not with full permissions. Warning raised on binding?

Basically, checking updates on both ends makes a difference, and mixed permissions work. 

Wdyt? Not overkill?

xmudrii: Makes sense to me! I'll push an update tomorrow.

xmudrii: @mjudeikis Addressed and PR is ready for another review. 🙂

pkg/virtual/apiexport/authorizer/binding_test.go
mjudeikis: > I don't see (in the ideal scenario) how removing some permissions in APIExport might break things down.
this is what I want to prove in the test. That having more permissions on APIBinding side than its actually still works. 

Other way around - yes, it will break. But I assume case like this should work.

xmudrii: I think we do have tests for these scenarios in this file (`pkg/virtual/apiexport/authorizer/binding_test.go`):

- `list request when APIExport allows list but APIBinding does not`
- `list request when APIBinding allows list but APIExport does not`

Have you checked these two cases? Both are resulting in a failure and that's correct.

> But I assume case like this should work.

I'm not sure what do you exactly mean by that, can you please clarify?

mjudeikis:
/lgtm
/approve
/hold 

Ping me Monday once we can start merging the previous one and this one. So we can do all on same go.

Issue #3371, opened 2025-04-14
flake: TestAPIBinding - "not found exactly on one shard"


embik:
It looks like my PR #3338 added a certain flakiness to `TestAPIBinding`.

```
apibinding_test.go:427: Error: cowboys.wildwest.dev is forbidden: User "kcp-admin" cannot list resource "cowboys" in API group "wildwest.dev" at the cluster scope: access denied
apibinding_test.go:439: 
        Error Trace:	/home/prow/go/src/github.com/kcp-dev/kcp/test/e2e/apibinding/apibinding_test.go:439
        Error:      	Not equal: 
       	            	expected: 1
       	            	actual  : 0
        Test:       	TestAPIBinding
        Messages:   	cowboys not found exactly on one shard, but on 0/1 
```

Here's a couple of failed runs:

- https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3282/pull-kcp-test-e2e-shared/1910191974799904768
- https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3368/pull-kcp-test-e2e-shared/1910037926343872512
- https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3353/pull-kcp-test-e2e-shared/1904846765551521792

- https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3368/pull-kcp-test-e2e-sharded/1910060684582850560
- https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3356/pull-kcp-test-e2e-sharded/1905169547917266944

And more.

It looks like the user in question doesn't get immediate access to the virtual workspace endpoint for APIExports, gets rejected, and the test fails. This checks needs to be made more robust and wrapped in a wait/retry, probably.

PR #3405, opened 2025-05-13
test: APIBinding: wrap listing VWs between shards with Eventually

gman0:
## Summary

The user gets Forbidden error even though it should have access to the resources. Retrying listing on the correct shard will eventually make it succeed.

Btw, the diff looks larger than it really is.

## What Type of PR Is This?

/kind flake

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3371

## Release Notes

```release-note
NONE
```

### DIFF
--- a/test/e2e/apibinding/apibinding_test.go
+++ b/test/e2e/apibinding/apibinding_test.go
@@ -442,29 +442,46 @@ func TestAPIBinding(t *testing.T) {
 	export2, err := kcpClusterClient.Cluster(provider2Path).ApisV1alpha2().APIExports().Get(ctx, exportName, metav1.GetOptions{})
 	require.NoError(t, err)
 
-	foundOnShards := 0
-	//nolint:staticcheck // SA1019 VirtualWorkspaces is deprecated but not removed yet
-	for _, vw := range export2.Status.VirtualWorkspaces {
-		vw2ClusterClient, err := kcpdynamic.NewForConfig(apiexportVWConfig(t, rawConfig, vw.URL))
-		require.NoError(t, err)
+	kcptestinghelpers.Eventually(t, func() (bool, string) {
+		foundOnShards := 0
+		var listErrs []error
 
-		t.Logf("Listing %s|%s cowboys via virtual workspace %s/clusters/%s", provider2Path, exportName, vw.URL, consumer3ClusterName)
-		gvr := wildwestv1alpha1.SchemeGroupVersion.WithResource("cowboys")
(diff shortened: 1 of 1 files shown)


ntnn (approved):

test/e2e/apibinding/apibinding_test.go
```diff
@@ -442,29 +442,46 @@ func TestAPIBinding(t *testing.T) {
...
-		t.Logf("Listing %s|%s cowboys via virtual workspace %s/clusters/%s", provider2Path, exportName, vw.URL, consumer3ClusterName)
-		gvr := wildwestv1alpha1.SchemeGroupVersion.WithResource("cowboys")
-		list, err := vw2ClusterClient.Cluster(consumer3ClusterName.Path()).Resource(gvr).Namespace("").List(ctx, metav1.ListOptions{})
-		if err != nil {
-			t.Logf("Error: %v", err)
-			continue
+			t.Logf("Listing %s|%s cowboys via virtual workspace wildcard list", provider2Path, exportName)
+			list, err = vw2ClusterClient.Resource(gvr).List(ctx, metav1.ListOptions{})
+			require.NoError(t, err, "error listing through virtual workspace wildcard")
+			require.Equal(t, 1, len(list.Items), "unexpected # of cowboys through virtual workspace with wildcard")
```
embik, line 468: I suspect all `require` calls (here and one below) should instead be returning `false` to the `Eventually` loop, no? Wouldn't it fail the test case in case of flakes otherwise?

gman0: Yes that's intentional, with three cases to cover:

* zero elements found: that's the flake, and we want to retry,
* one element: we're good and can return with success,
* more than one: definitely should not happen, and retrial won't help so we fail the test.

Does this sound reasonable?

gman0: Similarly with the # of `cowboys`, if it's not one then something else must be broken and we fail the test.

ntnn: An fail raised by `require` is collected by `Eventually`: https://github.com/stretchr/testify/blob/v1.10.0/assert/assertions.go#L2016
I had the same question when I worked on flakes but earlier `require` fails are ignored once the condition goes through.

cnvergence (approved):

test/e2e/apibinding/apibinding_test.go
embik: I see! TIL. Thanks for educating me.

embik (approved):
/approve

PR #3406, opened 2025-05-14
Update to code-generator v3

xrstf:
## Summary
The kcp code-generator was recently rewritten and now follows upstream much closer. This PR integrates this new generator into kcp and rebuilds all the generated code.

The new generated code is fully compatible with the old one, but now makes use of Kubernetes' `gentype` package to use Go Generics for fake clients.

Apart from that change, this PR also restores sorting the import statements. This was effectively broken and I would call it broken because "make codegen" always runs "make imports", indicating a clear wish that generated code should have sorted imports.

I tried to migrate the codegen logic to `kube_codegen.sh`, but since we have multiple inputs args to tools like applyconfiguration-gen, this is simply not possible (`kube::codegen::gen_client` expects exactly 1 source directory for APIs).

Updating to `kube_codegen.sh` would have given us a cleanup routine for free, which we are apparently lacking because we accidentally committed 

* `sdk/client/applyconfiguration/apis/v1alpha2/resource.go`
* `sdk/client/applyconfiguration/apis/v1alpha2/resourceschemastoragevirtual.go`

Both files claim to be generated, but are leftovers. If you checkout kcp's main branch and run `rm -rf sdk/client; make codegen` you will notice they are not being created [anymore]. To prevent this in the future I added a manual cleanup step into the codegen, expecting that `sdk/client` will for now only ever contain generated code (except, ironically, the `sdk/client/key_hack.go` :grin:).

## What Type of PR Is This?
/kind feature

## Release Notes
```release-note
Update to kcp/code-generator v3
```


sdk/apis/apis/v1alpha2/zz_generated.conversion.go
```diff
@@ -24,11 +24,12 @@ package v1alpha2
...
 	unsafe "unsafe"
 
-	v1alpha1 "github.com/kcp-dev/kcp/sdk/apis/apis/v1alpha1"
-	conditionsv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/third_party/conditions/apis/conditions/v1alpha1"
 	v1 "k8s.io/api/core/v1"
 	conversion "k8s.io/apimachinery/pkg/conversion"
 	runtime "k8s.io/apimachinery/pkg/runtime"
+
+	v1alpha1 "github.com/kcp-dev/kcp/sdk/apis/apis/v1alpha1"
+	conditionsv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/third_party/conditions/apis/conditions/v1alpha1"
```
mjudeikis, line 32: What I dont quite understand is why these are linted right (2 blocks), and 1 bellow - in reverse? Is this because one bellow (2) is third_party?

sdk/apis/third_party/conditions/util/conditions/getter.go
```diff
@@ -17,11 +17,10 @@ limitations under the License.
 package conditions
 
 import (
+	conditionsapi "github.com/kcp-dev/kcp/sdk/apis/third_party/conditions/apis/conditions/v1alpha1"
```
mjudeikis, line 20: (2) this is one below :D

mjudeikis (approved):
just a question on import changes.

PR #3408, opened 2025-05-15
Fix DynamicRestMapper bash

mjudeikis:
## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


ntnn (approved):

mjudeikis:
> FYI: https://formulae.brew.sh/formula/gnu-sed With `PATH="$HOMEBREW_PREFIX/opt/gnu-sed/libexec/gnubin:$PATH"`

Yes :) but I don't use it locally just so I can catch things like this :D I have installed but not aliased :D

ntnn:
Ahh ok :D

gman0 (approved):
We should have a replacement script for the files in https://github.com/kcp-dev/kcp/tree/main/hack/boilerplate to have these horrible things in one place.

PR #3418, opened 2025-05-20
Fix `--bind-address`

### LINKED ISSUE #3404: bug: Flag `--bind-address` is not honored
### Describe the bug

Regardless of the `--bind-address` flag, `kcp start` will always bind to all networks and template your IP-address into any generated urls (see below). I assume this flag comes directly from api-server code, but I still think we should honor the bind-address (or alternatively remove the flag, which I don't favor)

 I think this is particularly annoying during local development if your IP-address changes (for example when switching WIFI)

### Steps To Reproduce

```sh
$ go run ./cmd/kcp start --bind-address=127.0.0.1
```

kcp will still bind on all networks

```sh
$ lsof -Pi :6443
COMMAND   PID      USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
kcp     54920 simonbein    6u  IPv6 0x5dbac9139d8cd1c7      0t0  TCP *:6443 (LISTEN)
```

In addition, the generated admin-kubeconfig will still point to the ip address assigned on the default network interface instead of the bind-address.

```sh
$ grep "server" .kcp/admin.kubeconfig
    server: https://192.168.0.230:6443
    server: https://192.168.0.230:6443/clusters/root
    server: https://192.168.0.230:6443/clusters/system:admin
```

---

On a side note: I think when working on this, we should also take a look to update any other templating we have in place. For example for initializingworkspaces, it should also become the bind address I think.

```sh
$ k get workspacetypes -o yaml | yq '.[] | .[] | .status.virtualWorkspaces'
- url: https://192.168.0.230:6443/services/initializingworkspaces/root:home
- url: https://192.168.0.230:6443/services/initializingworkspaces/root:organization
- url: https://192.168.0.230:6443/services/initializingworkspaces/root:root
- url: https://192.168.0.230:6443/services/initializingworkspaces/root:team
- url: https://192.168.0.230:6443/services/initializingworkspaces/root:universal
```

### Expected Behaviour

All templating should be done based on the supplied bind address and kcp should only bind on the specified bind-address.

### Additional Context

_No response_

ntnn:
## Summary

Fix `--bind-address`

The fix is really only this commit, that sets the ExternalAddress if BindAddress is set:  [`eff92dd` (#3418)](https://github.com/kcp-dev/kcp/pull/3418/commits/eff92ddf7200ea05a19416193fbcaced39038cbc)

The others are changing the tests to run on localhost, which I had wanted to implement previously but failed one one hand because the ExternalAddress was the default IP rather than the BindAddress and because the loopback addresses were missing the in the certificates.

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #3404 

## Release Notes

```release-note
Fix `--bind-address` not being honoured in some generated configuration files
```

### DIFF
--- a/pkg/server/options/options.go
+++ b/pkg/server/options/options.go
@@ -287,6 +287,22 @@ func (o *Options) Complete(ctx context.Context, rootDir string) (*CompletedOptio
 		}
 	}
 
+	// ExternalAddress is the address used e.g. when generating
+	// kubeconfigs. It defaults to the default interface, usually the
+	// first non-loopback interface, e.g. 192.168.0.1.
+	// BindAddress is the address the server binds to, it defaults to
+	// 0.0.0.0 or ::.
+	//
+	// If BindAddress is set to a specific address, e.g. the loopback
+	// 127.0.0.1 the ExternalAddress is invalid and all URLs generated
+	// from it will not work.
+	//
+	// To prevent this ExternalAddress is set to the value of
+	// BindAddress if it wasn't set to a specific address.
+	if o.GenericControlPlane.GenericServerRunOptions.ExternalHost == "" && !o.GenericControlPlane.SecureServing.BindAddress.IsUnspecified() {
+		o.GenericControlPlane.GenericServerRunOptions.ExternalHost = o.GenericControlPlane.SecureServing.BindAddress.String()
+	}
+
 	if o.Extra.ExperimentalBindFreePort {
 		listener, _, err := genericapiserveroptions.CreateListener("tcp", fmt.Sprintf("%s:0", o.GenericControlPlane.SecureServing.BindAddress), net.ListenConfig{})
 		if err != nil {
--- a/pkg/server/options/options.go
+++ b/pkg/server/options/options.go
@@ -327,7 +327,34 @@ func (o *Options) Complete(ctx context.Context, rootDir string) (*CompletedOptio
 		o.GenericControlPlane.ServiceAccountSigningKeyFile = o.Controllers.SAController.ServiceAccountKeyFile
 	}
 
-	completedGenericOptions, err := o.GenericControlPlane.Complete(ctx, nil, nil)
+	// o.GenericControlPlane.Complete creates self-signed certificates
+	// with the advertise address by default. This can cause spurious
+	// errors if the server binds on multiple interfaces.
+	possibleIPs := []net.IP{
+		o.GenericControlPlane.GenericServerRunOptions.AdvertiseAddress,
+		o.GenericControlPlane.SecureServing.BindAddress,
+		o.GenericControlPlane.SecureServing.ExternalAddress,
+	}
+	if o.GenericControlPlane.SecureServing.Listener != nil {
+		host, _, err := net.SplitHostPort(o.GenericControlPlane.SecureServing.Listener.Addr().String())
+		if err != nil {
+			return nil, err
+		}
+		possibleIPs = append(possibleIPs, net.ParseIP(host))
+	}
+
+	alternateIPs := []net.IP{}
+	alternateDNS := []string{}
+
+	for _, ip := range possibleIPs {
+		if ip == nil || ip.IsUnspecified() {
+			continue
+		}
+		alternateIPs = append(alternateIPs, ip)
+		alternateDNS = append(alternateDNS, ip.String())
+	}
+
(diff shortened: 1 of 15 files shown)


pkg/server/options/options.go
```diff
@@ -287,6 +287,15 @@ func (o *Options) Complete(ctx context.Context, rootDir string) (*CompletedOptio
 		}
 	}
 
+	// ExternalAddress defaults to the default interface if unset and
```
gman0, line 290: Can you please re-phrase this comment? It reads a bit awkwardly.

ntnn:
```suggestion
	// ExternalAddress is the address used e.g. when generating kubeconfigs
	// or the virtual workspace URLs. It defaults to the default interface.
	// If the user provides BindAddress but not ExternalAddress the generated
	// kubeconfigs and URLs might not be valid.
	// Hence ExternalAddress is set here to the value of BindAddress only if
	// BindAddress is set but not ExternalAddress.
	// If neither are provided then the server listens on all interfaces but advertises
	// only the default interface.
```

Oh yeah, you're right. I think I had squashed two different comments together and that's why its so rough.

pkg/server/options/options.go
```diff
@@ -311,7 +320,44 @@ func (o *Options) Complete(ctx context.Context, rootDir string) (*CompletedOptio
 		o.GenericControlPlane.ServiceAccountSigningKeyFile = o.Controllers.SAController.ServiceAccountKeyFile
 	}
 
-	completedGenericOptions, err := o.GenericControlPlane.Complete(ctx, nil, nil)
+	// o.GenericControlPlane.Complete creates self-signed certificates
+	// with the advertise address by default. This can cause spurious
+	// errors if the server binds on multiple interfaces but only
+	// advertises for one.
```
ntnn, line 326:
```suggestion
	// o.GenericControlPlane.Complete creates self-signed certificates
	// with the advertise address by default. This can cause spurious
	// errors if the server binds on multiple interfaces.
```

pkg/server/options/options.go
```diff
@@ -311,7 +320,44 @@ func (o *Options) Complete(ctx context.Context, rootDir string) (*CompletedOptio
...
+		possibleIPs = append(possibleIPs, net.ParseIP(host))
+	}
+
+	alternateIPs := []net.IP{
+		// Always pass loopback addresses as part of the alternates.
+		// E.g. the generate admin.kubeconfig can set the IPv4 or IPv6
+		// loopback in the URIs if the server is bound to localhost.
+		// Since these alternateDNS and alternateIPs are used to
+		// generate the self-signed certificates connections will then
+		// fail if these loopbacks are not present in the SANs.
```
ntnn, line 346:
```suggestion
		// Always pass loopback addresses as part of the alternate IPs.
		// This prevents errors when the server is listening on `localhost`
		// and clients connect to `127.0.0.1`.
```

pkg/server/options/options.go
```diff
@@ -311,7 +327,40 @@ func (o *Options) Complete(ctx context.Context, rootDir string) (*CompletedOptio
...
+	if o.GenericControlPlane.SecureServing.Listener != nil {
+		host, _, err := net.SplitHostPort(o.GenericControlPlane.SecureServing.Listener.Addr().String())
+		if err != nil {
+			return nil, err
+		}
+		possibleIPs = append(possibleIPs, net.ParseIP(host))
+	}
+
+	alternateIPs := []net.IP{
+		// Always pass loopback addresses as part of the alternate IPs.
```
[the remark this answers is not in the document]
ntnn: The function that generates the certificates (that we also use indirectly) is using `localhost` as a DNS SAN by default if not specified:
https://github.com/kubernetes/kubernetes/blob/013c745c2d7c21eddfe7c0d337f72d403979e03b/staging/src/k8s.io/apiserver/pkg/server/options/serving.go#L366

I'm not sure if this is still required; I originally added this because I had some failures in the sharded tests with clients trying to connect to 127.0.0.1 but the certificate only being valid for localhost.

I'll run a test without the loopback and see if there are still failures - arguably it would be better to fix the tests to use the same address rather than add the loopbacks universally.

embik:
/approve

Looking good, leaving the lgtm to @gman0 to see if he has any follow-up concerns.

pkg/server/options/options.go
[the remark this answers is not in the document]
gman0: Great, looks better now!

PR #3429, opened 2025-05-26
Add verbs bundle validation

mjudeikis:
## Summary

Add verbs bundles validation for permissionsClaims

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Add PermissionClaims verbs bundles admission.
```


pkg/admission/apiexport/admission.go
```diff
@@ -146,6 +150,85 @@ func (e *APIExportAdmission) validatev1alpha2(_ context.Context, a admission.Att
 	return nil
 }
 
+// validateVerbsBundle validates that the verbs are a valid bundle according
+// to the following rules:
+// get | possible standalone
+// list -> get, watch  | if list requested, always requires get and watch too
```
embik, line 156: Not to bikeshed this list into all of eternity, but I wonder if `watch` should be a requirement for `list` now? A script/tool that just fetches all objects doesn't necessarily create a `watch` on the objects.

mjudeikis: But idea is that you can get objects data. So its still "read" type of action

xmudrii: +1 to what @embik said, I can see valid cases where you want to allow `list`, but not `watch`, there are in a way distinct verbs.

mjudeikis: In my head its more about "what operations you can do to the object, and less about the kube world". Does it give you same access as other verbs in the bundle, when yes. 

And again, we can always relax it later, where making it stricted it would be breaking change.

PR #3430, opened 2025-05-27
Serve both v1alpha1 and v1alpha2 APIBindings in APIExport VW, fix virtual workspace OpenAPI v3 panics

xmudrii:
## Summary

At the moment, `kubectl api-resources` in the APIExport Virtual Workspace shows that we serve APIBindings v1alpha2. However, getting an APIBinding (`kubectl get apibinding`) results into getting the v1alpha1 version of the object (it's impossible to get the `v1alpha2` version at all).

This PR addresses this by ensuring both v1alpha1 and v1alpha2 are available via the APIExport Virtual Workspace.

Additionally, fixing this bug undercovered an issue with the virtual workspace OpenAPI v3 endpoint/implementation. If there's a resource with two different versions, it'll cause a panic because we try to register the same GR (GroupResource) two times. This is fixed in this PR as well and it needs to be cherry-picked to `release-0.27`.

Tests are added to cover both of these bugs.

## What Type of PR Is This?

/kind bug

## Related Issue(s)

None

## Release Notes
```release-note
- Serve both v1alpha1 and v1alpha2 of APIBindings in the APIExport Virtual Workspace
- Fix a panic in the OpenAPI v3 endpoint for Virtual Workspaces happening if there's a resource with two or more versions
```

### DIFF
--- a/pkg/virtual/apiexport/builder/build.go
+++ b/pkg/virtual/apiexport/builder/build.go
@@ -179,7 +179,7 @@ func BuildVirtualWorkspace(
 						cancelFn:      cancelFn,
 					}, nil
 				},
-				func(ctx context.Context, clusterName logicalcluster.Name, apiExportName string) (apidefinition.APIDefinition, error) {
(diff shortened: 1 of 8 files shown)


embik (approved):
/approve

PR #3433, opened 2025-06-02
Rework test/integration framework

### LINKED ISSUE #3420: feature: Revise integration tests
### Feature Description

Revise integration tests to expose the underlying `pkg/server.Server`.

Things like the DynamicRESTMapper cannot be tested properly at the moment as the structure is not exposed to any of the tests.

When adding `test/integration` I originally had a separate setup function but adjusted `sdk/testing` instead as this was already doing most of what was needed to start and stop the server.

However proper integration tests - like for DynamicRESTMapper - would require exposing the underlying server struct.

### Proposed Solution

The easiest solution:
Restructure the framework to allow for `runner` to optionally return the server struct and store this in `kcpServer`. For external servers this can be simply nil.
The `RunningServer` interface could then have a method to return this server and it's up to the user to use this correctly.

The slightly harder solution:
Rewrite `test/integration/framework` to expose the underlying server struct.

The harder solution:
Redesign `sdk/testing`, `test/e2e/framework` and `test/integration/framework` with all this in mind.

### Alternative Solutions

_No response_

### Want to contribute?

### Additional Context

Personally I'm not a big fan of the easiest solution because `sdk/testing...` is already messy as is. 
Then again this would probably just be a <10sloc change<sup>citation needed</sup>.

Then again I'm also not a fan of rewriting `test/integration/framework` because it would probably duplicate a good chunk of code from `sdk/testing` - which is what I wanted to avoid intially by using `test/e2e/framework/inprocess` and `sdk/testing`.

ntnn:
## Summary

Rework test/integration to expose the underlying server struct.

## What Type of PR Is This?

/kind feature
/kind cleanup

## Related Issue(s)

Fixes #3420 
Closes #3422 

## Release Notes

```release-note
`test/e2e/framework` no longer sets the in-process function in `sdk/testing/server`, import `test/integration/framework` instead
```


ntnn:
@gman0 this should allow testing `DynamicRESTMapper`.

ntnn:
I removed the inprocess function from the e2e framework - it wasn't used. Instead it is now set in the integration framework.
This might be a problem if people are importing the e2e just to set this function. LMK if you want to keep this behaviour, but I didn't want to just leave a shim there that imports the integration framework when it's not actually used in e2e tests.

ntnn:
/retest 

o.O

PR #3435, opened 2025-06-02
fix docs validation in Pull Requests

xrstf:
## Summary
This PR makes sure our docs build within a PR actually verifies the docs. By default mkdocs runs in non-strict mode and in that mode it just ignores errors. To make strict mode work, I also had to bump one plugin.

I also

* fixed the make target to clean the generated CRDs,
* adjusted some links to mkdocs strict mode,
* reformatted the CLI help section from `SEE ALSO` to `See Also` because this is not a manfile, but Markdown,
* unindented the kcp CLI examples, which were intended too much: 
 which made them look awful in the generated docs
* prevent leftover binaries by using `go run` to generate the CLI docs and
* skip _create_ plugins in the docs to not generate stray links to a nonexisting `k create` command.

## What Type of PR Is This?
/kind documentation

## Release Notes
```release-note
NONE
```

### DIFF
--- a/cli/pkg/bind/cmd/cmd.go
+++ b/cli/pkg/bind/cmd/cmd.go
@@ -28,15 +28,15 @@ import (
 
 var (
 	bindExampleUses = `
-	# Create an APIBinding named "my-binding" that binds to the APIExport "my-export" in the "root:my-service" workspace.
-	%[1]s bind apiexport root:my-service:my-export --name my-binding
+# Create an APIBinding named "my-binding" that binds to the APIExport "my-export" in the "root:my-service" workspace.
+%[1]s bind apiexport root:my-service:my-export --name my-binding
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -163,12 +163,11 @@ update-contextual-logging: $(LOGCHECK) ## Update contextual logging
 .PHONY: generate-cli-docs
 generate-cli-docs: ## Generate cli docs
 	git clean -fdX docs/content/reference/cli
-	pushd . && cd docs/generators/cli-doc && go build . && popd
-	./docs/generators/cli-doc/cli-doc -output docs/content/reference/cli
+	cd ./docs/generators/cli-doc && go run . -output ../../content/reference/cli
```
embik, line 166: Why change this? It was working perfectly fine and now it's changing directories 🤷🏻

PR #2510, opened 2022-12-19
⚠️  ClusterWorkspace => LogicalCluster refactor

sttts:
High-level overview: https://hackmd.io/Qb3HX2XGRV63Rt-83dZXNg

TODOs:
- [x] Rename `ThisWorkspace` to `LogicalCluster`, group `core.kcp.dev`.
- [x] Rename `ClusterWorkspaceShard` to `Shard`, group `core.kcp.dev`.
- [x] Rename `ClusterWorkspaceType` to `WorkspaceType`.
- [x] Rename `tenancy.kcp.dev/path` to `kcp.dev/paths`, and move into core package.
- [x] create `ThisWorkspace` in some logical cluster that is NOT the workspace path: `base36(sha224(random))` and name with cluster name.
- [x] Rename `LogicalCluster` objects to `"cluster"`

Revisit:
- [x] @s-urbaniak [REVISIT(sur): Revert "SQUASH: enhance godoc"](https://github.com/kcp-dev/kcp/pull/2510/commits/6b7eb2a776d2a6c46f5fd8edea29526c8022de62)
- [x] @s-urbaniak [REVISIT(sur): Revert "pkg/authorization: add delegation reason in audit"](https://github.com/kcp-dev/kcp/pull/2510/commits/1c20db9a4f4480815bd0be8bf890e4ad4fb4ccc6)
- [x] @p0lyn0mial [REVISIT(lukasz): skip quota and homeworkspace checks for thisworkspac…](https://github.com/kcp-dev/kcp/pull/2510/commits/a8c641d64501b5bcbdae2cc2c5bec8b35c4dfb9f)
- [x] @davidfestal @jmprusi [REVISIT(dfestal): e2e/virtual/syncer: disable test depending on root:…](https://github.com/kcp-dev/kcp/pull/2510/commits/17009290838b5d3d05b499dcca91337d611c3fd4)
- [x] @sttts REVISIT(sttts): special case thisworkspaces as identity-less

Follow-ups:
- [ ] @sttts update https://hackmd.io/Qb3HX2XGRV63Rt-83dZXNg and copy into enhancement repo
- [ ] Rename `kcp.dev` everywhere (including groups) to `*.kcp.io`
- [ ] Remove `ClusterWorkspace` and the projection.
- [ ] Replace kcp.dev/cluster in logicalcluster/v3.
- [ ] Move workspace.status.cluster to spec
- [ ] Add an indexer helper for getting by path, path+name
[...]

PR #3438, opened 2025-06-03
Do not set the owner annotation on `LogicalClusters` to empty value

embik:
## Summary

It looks like all the way back in #2510, we added some logic to mirror the `experimental.tenancy.kcp.io/owner` annotation from `Workspace` objects to `LogicalClusters`. Unfortunately, this is how `pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling.go` ended up looking like:

https://github.com/kcp-dev/kcp/blob/68596e1e59e966cfc44b1c3c1e6287193f27ca38/pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling.go#L246-L270

There is a check if `ExperimentalWorkspaceOwnerAnnotationKey` is set on the workspace later on in the code, but the annotation is already set by this time (see the initial set of annotations). This produces an annotation with an empty value, which trips up the owner annotation parser.

Since there might be existing kcp installations out there with bad owner annotations on `LogicalCluster` objects caused by this bug, this PR also skips parsing the annotation if it's empty.

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Only set `experimental.tenancy.kcp.io/owner` annotation on `LogicalCluster` if `Workspace` has the annotation
```

### DIFF
--- a/pkg/reconciler/tenancy/logicalcluster/logicalcluster_controller.go
+++ b/pkg/reconciler/tenancy/logicalcluster/logicalcluster_controller.go
@@ -201,9 +201,11 @@ func (c *Controller) process(ctx context.Context, key string) error {
 		return nil
 	}
 
-	// need to create
-	ownerAnnotation, found := logicalCluster.Annotations[tenancyv1alpha1.ExperimentalWorkspaceOwnerAnnotationKey]
-	if !found {
+	// need to create ClusterRoleBinding for owner.
+	ownerAnnotation := logicalCluster.Annotations[tenancyv1alpha1.ExperimentalWorkspaceOwnerAnnotationKey]
+	// some older installations of kcp might have produced an annotation with empty value, so we should
+	// not care whether the annotation is set or if it's empty.
+	if ownerAnnotation == "" {
 		// no owner - can't create
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


pkg/reconciler/tenancy/logicalcluster/logicalcluster_controller.go
```diff
@@ -203,7 +203,8 @@ func (c *Controller) process(ctx context.Context, key string) error {
 
 	// need to create
 	ownerAnnotation, found := logicalCluster.Annotations[tenancyv1alpha1.ExperimentalWorkspaceOwnerAnnotationKey]
-	if !found {
+	// some older installations of kcp might have produced an annotation with empty value.
+	if !found || ownerAnnotation == "" {
 		// no owner - can't create
```
xrstf: If this is the case, then we do not need `found` anymore and can just write

```go
owner := logicalCluster.Annotations[tenancyv1alpha1.ExperimentalWorkspaceOwnerAnnotationKey]
if owner == "" {
   return nil
}
```

embik: Fixed.

embik:
known flake (#3444).

/retest

embik:
/retest

same flake ...

PR #3411, opened 2025-05-18
Disable APIExport VirtualWorkspace URL population

mjudeikis:
## Summary

Disable APIExport URL population. 

## What Type of PR Is This?
/kind feature
/kind deprecation

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3366

## Release Notes

```release-note
Deprecated APIExport VirtualWorksapce URL population
Add FeatureFlag to re-enable deprecated APIExport VirtualWorkspace URLs. 
```

PR #3439, opened 2025-06-03
Authorize server-side apply as create in `forwardingregistry`

xmudrii:
## Summary

At the moment, server-side apply requests in the virtual workspaces are authorized exclusively as `patch`. This doesn't match the upstream Kubernetes behavior where such requests are authorized as `create`. This is documented here: https://kubernetes.io/docs/reference/using-api/api-concepts/#patch-and-apply

The initial idea for mitigating this issue, for the APIExport Virtual Workspace, was to bundle `patch` and `create` verbs. In other words, if you request `patch`, you must request `create` as well. However, this doesn't solve the issue that exclusive `patch` can create new resources using SSA, it only "hides" the issue.

This PR takes another approach by running the admission chain if it is determined that the object needs to be created. The admission chain in this case includes the authorizers, but instead of running them with `patch`, it runs them with `create`. If that fails, we are not proceeding with creation but instead return 403 Forbidden (as is done in Kubernetes).

This taps into the `forwardingregistry` which is used by other Virtual Workspaces. However, for all other workspaces (i.e. workspaces that are not APIExport VW), this change is supposedly INOP as long as they do not forbid `create` request. The same admission chain is anyways being used before handling the request, just with a different verb (`patch` instead of `create`). This only ensures we run the admission chain with the correct verb (`create`) once it is determined that we want to create a new object.

## What Type of PR Is This?

/kind feature

## Related Issue(s)

xref #3429 

## Release Notes
```release-note
Run the admission chain in the virtual workspace (forwarding) registry with the `create` verb upon creating a new object using server-side apply. As a result, running the server-side apply for a claimed resource in the APIExport Virtual Workspace requires the `create` verb
```

/assign @sttts @mjudeikis @embik

### DIFF
--- a/test/e2e/virtual/apiexport/binding_test.go
+++ b/test/e2e/virtual/apiexport/binding_test.go
@@ -18,6 +18,7 @@ package apiexport
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"strings"
@@ -29,7 +30,9 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/rest"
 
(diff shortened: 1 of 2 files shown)


xmudrii:
/hold
for discussion

test/e2e/virtual/apiexport/binding_test.go
```diff
@@ -404,6 +407,162 @@ func TestAPIBindingPermissionClaimsVerbs(t *testing.T) {
...
+
+	t.Logf("Validate that the permission claims were all applied")
+	kcptestinghelpers.EventuallyCondition(t, func() (conditions.Getter, error) {
+		return kcpClusterClient.Cluster(consumerPath).ApisV1alpha2().APIBindings().Get(ctx, "cowboys", metav1.GetOptions{})
+	}, kcptestinghelpers.Is(apisv1alpha2.PermissionClaimsApplied), "unable to see claims applied")
+
+	t.Logf("Waiting for APIExport to have a virtual workspace URL for the bound workspace %q", consumerPath)
+	apiExportVWCfg := rest.CopyConfig(cfg)
+	kcptestinghelpers.Eventually(t, func() (bool, string) {
+		apiExport, err := kcpClusterClient.Cluster(providerPath).ApisV1alpha2().APIExports().Get(ctx, "today-cowboys", metav1.GetOptions{})
```
mjudeikis: This will stop working once https://github.com/kcp-dev/kcp/pull/3411 merges

mjudeikis (commented):
just nit

test/e2e/virtual/apiexport/binding_test.go
xmudrii: I tried to see if I can already change this, but it requires #3411 to get merged. If #3411 gets merged first, I'll rebase and update this PR, otherwise if this PR gets merged first, you can handle this in #3411.

xmudrii: Updated now that #3411 is merged

PR #3441, opened 2025-06-04
Use 'core' group to replicate core API objects

gman0:
## Summary

Resources from core API use empty group, but this breaks
label selectors that end up being empty, and fail validation.
This PR replaces empty group with 'core'.

## What Type of PR Is This?

/kind bug

## Release Notes

```release-note
NONE
```


gman0:
PTAL @mjudeikis

PR #3457, opened 2025-06-19
Improve linting config

xrstf:
## Summary
This PR cleans up the golangci-lint config a bit. Linters that were disabled but actually not trigger on any errors anymore have been removed, some other linters have been enabled and their issues have been fixed (like `intrange`). 95% of this PR is just enabling the filepathJoin linter in gocritic, which I chose to fix simply because it did actually point out a few odd lines that were worth fixing.

## What Type of PR Is This?
/kind cleanup

## Release Notes
```release-note
NONE
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -41,27 +41,20 @@ linters:
   settings:
     gocritic:
       disabled-checks:
-        - appendAssign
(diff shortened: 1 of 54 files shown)


embik (approved):
/approve

PR #3465, opened 2025-06-24
CachedResource replication reconciler fixes

gman0:
## Summary

This PR fixes various things around CacheResource replication:

* (bug) If the `CachedObject` is missing, and `Create` fails with `AlreadyExists`, the reconciler now returns an error and lets the work item to be requeued. The previous code would, in that case, attempt to do an `Update` but without setting the object's meta, the Update call would always fail (missing `resourceVersion` and others...). Letting the item to be requeued is cleaner and fixes object updates.
* (clean up) The reconciler should rely on the GVR passed from the controller instead of having to reach out to REST mapper. It already has all the information it needs, no need to do any extra resource-kind resolution.
* (feature) CachedObject is now namespace-aware. ~~CachedObjects now follow `<Version>.<Group>.<Resource>.<Name>.<Namespace>` naming scheme.~~
  * CachedObject uses hash of GVR+Namespace+Name as its name.

These three changes are split into three respective commits.

## What Type of PR Is This?

/kind cleanup

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/cache/cachedresources/replication/replication_reconcile.go
+++ b/pkg/reconciler/cache/cachedresources/replication/replication_reconcile.go
@@ -17,11 +17,15 @@ limitations under the License.
 package replication
 
 import (
+	"bytes"
 	"context"
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
+
+	"github.com/martinlindhe/base36"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -48,6 +52,18 @@ const (
 	LabelKeyObjectOriginalName      = "cache.kcp.io/object-original-name"
 	LabelKeyObjectOriginalNamespace = "cache.kcp.io/object-original-namespace"
 )
+
+func GenCachedObjectName(gvr schema.GroupVersionResource, namespace, name string) string {
+	buf := bytes.Buffer{}
+	buf.WriteString(gvr.String())
+	buf.WriteString(namespace)
+	buf.WriteString(name)
+
+	hash := sha256.Sum256([]byte(name))
+	base36hash := strings.ToLower(base36.EncodeBytes(hash[:]))
+
+	return base36hash
+}
 
 func (c *Controller) reconcile(ctx context.Context, gvrKey string) error {
 	if c.deleted {
@@ -142,7 +158,7 @@ func (c *Controller) reconcile(ctx context.Context, gvrKey string) error {
 					APIVersion: cachev1alpha1.SchemeGroupVersion.String(),
 				},
 				ObjectMeta: metav1.ObjectMeta{
-					Name:              gvr.Version + "." + gvr.Resource + "." + gvr.Group + "." + obj.GetName(), // TODO: handle namespace
+					Name:              GenCachedObjectName(gvr, obj.GetNamespace(), obj.GetName()),
 					Labels:            obj.GetLabels(),
 					Annotations:       obj.GetAnnotations(),
 					CreationTimestamp: metav1.NewTime(time.Now()),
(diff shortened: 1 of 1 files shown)


embik:
> (feature) CachedObject is now namespace-aware. CachedObjects now follow `<Version>.<Group>.<Resource>.<Name>.<Namespace>`  naming scheme.
> Is this ok, or is there some idiomatic naming convention that I should follow?

My natural instinct would be `<Group>.<Version>.<Resource>.<Namespace>.<Name>` as an order, @mjudeikis any thoughts?

Also, at this point I would be concerned about object name length for the `CachedObject`. Is it time to hash names and have all this information in labels?

pkg/reconciler/cache/cachedresources/replication/replication_reconcile.go
```diff
@@ -149,42 +151,41 @@ func (c *Controller) reconcile(ctx context.Context, gvrKey string) error {
...
-				return nil, err
-			}
 			cacheObj := &cachev1alpha1.CachedObject{
 				TypeMeta: metav1.TypeMeta{
 					Kind:       cache.CachedObjectKind,
 					APIVersion: cachev1alpha1.SchemeGroupVersion.String(),
 				},
 				ObjectMeta: metav1.ObjectMeta{
-					Name:            gvr.Version + "." + mapper.Resource.Resource + "." + gvr.Group + "." + obj.GetName(),
+					Name:            gvr.Version + "." + gvr.Resource + "." + gvr.Group + "." + obj.GetName(),
```
mjudeikis: namespace?

mjudeikis:
> > (feature) CachedObject is now namespace-aware. CachedObjects now follow `<Version>.<Group>.<Resource>.<Name>.<Namespace>`  naming scheme.
> > Is this ok, or is there some idiomatic naming convention that I should follow?
> 
> My natural instinct would be `<Group>.<Version>.<Resource>.<Namespace>.<Name>` as an order, @mjudeikis any thoughts?
> 
> Also, at this point I would be concerned about object name length for the `CachedObject`. Is it time to hash names and have all this information in labels?

So we have 2 problems:
1. Naming of the object - we dont care that much as its just needs to be unique and readable.
2. Label selector particular GVR (Example - Give me all `KubeClusters` in group `Foo.Bar.Potato`, version `v1`. And this is where challenges comes

When naming is `gvr.Version + "." + gvr.Resource + "." + gvr.Group` you can 
split first 2 dots and you know `version` and `resource` as they are always without `.` in the name. Where group might be multiple-dots. So hence  pattern `gvr.Version + "." + gvr.Resource + "." + gvr.Group` came to life.

So everything else was derived from it to make it easier to read:
So I see label `v1.kubeclusters.foo.bar.potato` and name is `v1.kubeclusters.foo.bar.potato.default.default-potato` 

meaning its not mentally different from label schema `foo.bar.potato.v1.kubeclusters.default.default-potato` 

TL:DR: Schema was forced to be able to do label selectors queries easier, everything else just readability.

mjudeikis:
Regards hashing - this could be an issue. maybe worth doing this now.

pkg/reconciler/cache/cachedresources/replication/replication_reconcile.go
gman0: Namespace is appended conditionally to the name just below, L197.

gman0:
Do we care about name readability if we have labels? There are two options:

* We just use a hash
* We construct a name as `<partial name>-<hash>`:
  * Where `<partial name> = substr("<GVR>-<NS>-<Name>", 0, <Max object name length> - <Length of hash> - 1)`

gman0:
I could also see just using GenerateName with `<GVR>-` prefix.

mjudeikis:
> I could also see just using GenerateName with `<GVR>-` prefix.

I think we need hash, to avoid conflicts if some errors is made. else we might end up with duplicates. My vote goes to hash.

gman0:
True. Do we want any human-readable prefix for the name, or is a plain hash (of the name we have currently in use) sufficient? An admin inspecting the cache server might have a hard time identifying what's currently stored without extra tooling.

mjudeikis:
Its already dedicated type from `cache` object. 
And you can add column into CRD to print Name, group, etc as kubectl. But not like anybody ever should talk to cache server via kubectl.

gman0:
/hold

I'll address the comments ASAP.

gman0:
The name is now `base36(sha256(GVR+namespace+name))`. E.g. `62kbbygmbxs7b3bfv4qqa5iv9vwwu1roqi0295mdnt8pojkj4x`, with length of 50 characters.

PR #3468, opened 2025-06-26
:bug: [release-0.27] Fix virutal workspace OpenAPI v3 endpoint panics

xmudrii:
## Summary

Partial cherry-pick of #3430

Fixes an issue with the virtual workspace OpenAPI v3 endpoint/implementation. If there's a resource with two different versions, it'll cause a panic because we try to register the same GR (GroupResource) twice.

## What Type of PR Is This?

/kind bug

## Release Notes
```release-note
Fix a panic in the OpenAPI v3 endpoint for Virtual Workspaces happening if there's a resource with two or more versions
```

/assign @xrstf @embik


embik (approved):
/approve

Issue #3469, opened 2025-06-30
bug: CachedResource GVR uniqueness


gman0:
### Describe the bug

There should be only one CachedResource per GVR in a workspace to prevent duplicate replication, e.g. when sets of objects to replicate overlap because of used label selectors.

### Steps To Reproduce

Create ConfigMaps hello-green and hello-blue:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: hello-green
  labels:
      color: green
data:
  hello: world
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: hello-blue
  labels:
      color: blue
data:
  hello: world
```

Define CachedResources hello-cache-all and hello-cache-green:

```yaml
apiVersion: cache.kcp.io/v1alpha1
kind: CachedResource
metadata:
  name: hello-cache-all
spec:
  group: ""
  version: v1
  resource: configmaps
---
apiVersion: cache.kcp.io/v1alpha1
kind: CachedResource
metadata:
  name: hello-cache-green
spec:
  group: ""
  version: v1
  resource: configmaps
  labelSelector:
    matchLabels:
      color: green
```

Observe duplicate replication due to overlapping selectors (all and `color: green`):

```yaml
apiVersion: cache.kcp.io/v1alpha1
kind: CachedResource
metadata:
  annotations:
    kcp.io/cluster: root
  creationTimestamp: "2025-06-30T07:51:19Z"
  finalizers:
  - cachedresource.cache.kcp.dev
  generation: 2
  name: hello-cache-all
  resourceVersion: "412"
  uid: 36a92eae-3e8b-4fab-8aa9-808822fcf717
spec:
  group: ""
  identity:
    secretRef:
      name: hello-cache-all
      namespace: kcp-system
  resource: configmaps
  version: v1
status:
  conditions:
  - lastTransitionTime: "2025-06-30T07:51:20Z"
    status: "True"
    type: IdentityValid
  - lastTransitionTime: "2025-06-30T07:51:20Z"
    status: "True"
    type: ReplicationStarted
  - lastTransitionTime: "2025-06-30T07:51:20Z"
    status: "True"
    type: ResourceValid
  identityHash: a922ffde13cc8005f6b3323bb921ae9294453f46a7566867ad084e162e105fb8
  phase: Ready
  resourceCounts:
    cache: 4
    local: 4
---
apiVersion: cache.kcp.io/v1alpha1
kind: CachedResource
metadata:
  annotations:
[...]

PR #3476, opened 2025-07-07
Add admission plugin for CachedResources

gman0:
## Summary

The admission plugin validates CachedResources creation, and ensures that there is no other such CachedResource in the cluster that acts on the same GVR.

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3469

## Release Notes

```release-note
Make CachedResources GVR immutable and unique in the logicalcluster 
```

### DIFF
--- a/pkg/admission/cachedresource/admission_test.go
+++ b/pkg/admission/cachedresource/admission_test.go
@@ -1,0 +1,200 @@
+/*
+Copyright 2025 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cachedresource
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apiserver/pkg/admission"
+	"k8s.io/apiserver/pkg/authentication/user"
+	genericapirequest "k8s.io/apiserver/pkg/endpoints/request"
+
+	"github.com/kcp-dev/logicalcluster/v3"
+
+	"github.com/kcp-dev/kcp/pkg/admission/helpers"
+	cachev1alpha1 "github.com/kcp-dev/kcp/sdk/apis/cache/v1alpha1"
+)
+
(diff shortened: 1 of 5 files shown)


pkg/virtual/framework/transforming/transformer_test.go
```diff
@@ -165,8 +165,15 @@ func (c *resourceClient) RawDelete(ctx context.Context, name string, opts metav1
...
+func titleASCII(s string) string {
+	if s == "" {
+		return ""
+	}
+	return strings.ToUpper(s[0:1]) + s[1:]
+}
+
 func (c *resourceClient) RawDeleteCollection(ctx context.Context, opts metav1.DeleteOptions, listOptions metav1.ListOptions) ([]byte, int, error) {
-	gvk := c.resource.GroupVersion().WithKind(strings.TrimRight(strings.Title(c.resource.Resource), "s")) //nolint:staticcheck
+	gvk := c.resource.GroupVersion().WithKind(strings.TrimRight(titleASCII(c.resource.Resource), "s"))
```
gman0, line 176: Unrelated to this PR, but `make lint` wouldn't pass.

gman0: An alternative to the `titleASCII` function would be to use the experimental `cases.Title`. See https://pkg.go.dev/strings#Title

pkg/admission/cachedresource/admission_test.go
```diff
@@ -0,0 +1,160 @@
...
+							Group:    "example.org",
+							Version:  "v1",
+							Resource: "objects",
+						}),
+					},
+				},
+			},
+			cluster: logicalcluster.Name("cluster-1"),
+		},
+		"AlreadyExists": {
```
ntnn, line 125: Can you add a test updating a `CachedResource`? e.g. by setting/changing the `secretRef`.

gman0: The handler is defined only on Create ops so it shouldn't be triggered by mutating the object, but an additional check is ok.

ntnn: Since the GVR on the CachedObject is not immutable, maybe we should hook this plugin on update as well?
Otherwise a user could create an object with one GVR and then update it to another existing one

gman0: @mjudeikis I also think GVR should be immutable. Do we put it here in the admission?

ntnn: A validation on the spec should be enough:
https://github.com/kcp-dev/kcp/blob/802765fdead32f88824b23b7de075b9043137eca/sdk/apis/cache/v1alpha1/types_cachedresource.go#L79-L99

But would it make sense to make it immutable? Would there be something that could potentially break if the GVR of a CachedResource is updated rather than deleted and created?

gman0: I've talked with @mjudeikis about this and indeed the GVR should be immutable. The way things are planned out currently is tha tthe replication VW will be exposed at `/services/replication/<CachedResource cluster>/<CachedResource>` API endpoint with read-only access to the cached objects described by the CachedResource object in the workspace. If the GVR changes, the "old" cached objects wouldn't be reachable.

Btw, the same is true if the label selector changes, e.g. from no selector -- meaning all objects -- into any non-empty selector.

Issue #3417, opened 2025-05-20
feature: implement PublishedResourcesEndpointSlices


mjudeikis:
### Feature Description

As per document https://docs.google.com/document/d/18Xh2-VFnH23bDiAHPP_JXm4qAG4QK6GjTKcpszPGApQ/edit?tab=t.0 

### Proposed Solution

We need to implement VirtualWorkspace & EndpointSlices object for PublishedResources (introduced in https://github.com/kcp-dev/kcp/issues/3367 ) to serve published resources via shard aware URLs from cache server.

It should be something like /services/publishedresources/shards/*/clusters/*/publishedresource/<name> type of URL where we serve content based on PublishedResource identity.  

### Want to contribute?

- [ ] I would like to work on this issue.

gman0:
The GVR reference in CachedResource is a bit too loose. Where does the schema come from, and how can the replication VW ensure that whatever it reads from the CachedObject can be fitted onto the same schema that was used while creating that CachedObject?

A solution: when creating a CachedResource, we should ensure that the GVR reference points to something that can be resolved to APIResourceSchema or core k8s/kcp APIs. WDYT, @mjudeikis ?

mjudeikis:
Maybe we need to change (I think maybe might be too weak of the statmenet) but CachedResources should be using APIResourceSchema. So the propagation pattern is the same as in normal API Export i using.

gman0:
I think the currently proposed Replication VW URL, `/services/replication/<CachedResource cluster>/<CachedResource>/clusters/<Target cluster>/<tail>`, might need a change.

CachedResource is just a GVR+Label selector on which to trigger replication. The real ownership of the schema of the objects we are replicating belongs to the respective APIExports, and I think we should tie the VW with that.

Case 1:
* CachedResource `cowboys-blue` for wildwest.dev/cowboys with label selector `color=blue`, and
* CachedResource `cowboys-green` for wildwest.dev/cowboys with label selector `color=green`.
* Both point to the same schema defined in an APIResourceSchema, both should be accessible from the same Replication VW URL.
* See also https://github.com/kcp-dev/kcp/issues/3469

Case 2:
* CachedResource `cowboys` for wildwest.dev/cowboys in workspace `root:a`, and
* CachedResource `cowboys` for wildwest.dev/cowboys in workspace `root:b`.
* Wildcard requests should list cached Cowboys from all workspaces that define CachedResource with wildwest.dev/cowboys GR.

[...]

PR #3480, opened 2025-07-11
Replication VW for CachedResources

gman0:
## Summary

This PR adds Replication VW for CachedResources. The VW exposes objects replicated by a CachedResource as read-only, with support for GET, LIST and WATCH verbs.

Creation of CachedResource automatically triggers creation of a CachedResourceEndpointSlice (with the same name as its CachedResource parent) where the URL for the VW is displayed.

No partitions or accessing replicated objects across workspaces. This will be done as the next step when integrating with APIExports.

## What Type of PR Is This?

/kind api-change
/kind feature

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3417

## Release Notes

```release-note
* Adds Replication virtual workspace for CachedResources
* Endpoint for the Replication VW is now displayed in its respective CachedResourceEndpointSlice
```

### DIFF
--- a/pkg/virtual/replication/builder/build.go
+++ b/pkg/virtual/replication/builder/build.go
@@ -1,0 +1,395 @@
+/*
+Copyright 2025 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package builder
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apiserver/pkg/authorization/authorizer"
+	genericapirequest "k8s.io/apiserver/pkg/endpoints/request"
+	genericapiserver "k8s.io/apiserver/pkg/server"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+
+	kcpdynamic "github.com/kcp-dev/client-go/dynamic"
+	kcpkubernetesclientset "github.com/kcp-dev/client-go/kubernetes"
+	"github.com/kcp-dev/logicalcluster/v3"
+
+	"github.com/kcp-dev/kcp/pkg/authorization"
+	"github.com/kcp-dev/kcp/pkg/indexers"
+	"github.com/kcp-dev/kcp/pkg/reconciler/apis/apibinding"
+	cachedresourcesreplication "github.com/kcp-dev/kcp/pkg/reconciler/cache/cachedresources/replication"
+	"github.com/kcp-dev/kcp/pkg/virtual/apiexport/schemas/builtin"
+	"github.com/kcp-dev/kcp/pkg/virtual/framework"
+	virtualworkspacesdynamic "github.com/kcp-dev/kcp/pkg/virtual/framework/dynamic"
+	"github.com/kcp-dev/kcp/pkg/virtual/framework/dynamic/apidefinition"
+	"github.com/kcp-dev/kcp/pkg/virtual/framework/dynamic/apiserver"
+	dynamiccontext "github.com/kcp-dev/kcp/pkg/virtual/framework/dynamic/context"
+	"github.com/kcp-dev/kcp/pkg/virtual/framework/forwardingregistry"
+	"github.com/kcp-dev/kcp/pkg/virtual/framework/rootapiserver"
+	"github.com/kcp-dev/kcp/pkg/virtual/replication"
+	"github.com/kcp-dev/kcp/pkg/virtual/replication/apidomainkey"
+	replicationauthorizer "github.com/kcp-dev/kcp/pkg/virtual/replication/authorizer"
+	apisv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/apis/v1alpha1"
+	apisv1alpha2 "github.com/kcp-dev/kcp/sdk/apis/apis/v1alpha2"
+	corev1alpha1 "github.com/kcp-dev/kcp/sdk/apis/core/v1alpha1"
+	kcpclientset "github.com/kcp-dev/kcp/sdk/client/clientset/versioned/cluster"
+	kcpinformers "github.com/kcp-dev/kcp/sdk/client/informers/externalversions"
+)
+
+func BuildVirtualWorkspace(
+	cfg *rest.Config,
+	rootPathPrefix string,
+	kcpClusterClient kcpclientset.ClusterInterface,
+	dynamicClusterClient kcpdynamic.ClusterInterface,
+	kubeClusterClient kcpkubernetesclientset.ClusterInterface,
+	localKcpInformers kcpinformers.SharedInformerFactory,
+	globalKcpInformers kcpinformers.SharedInformerFactory,
+) ([]rootapiserver.NamedVirtualWorkspace, error) {
+	if !strings.HasSuffix(rootPathPrefix, "/") {
+		rootPathPrefix += "/"
+	}
+
+	readyCh := make(chan struct{})
+
+	cachedResourceContent := &virtualworkspacesdynamic.DynamicVirtualWorkspace{
+		RootPathResolver: framework.RootPathResolverFunc(func(urlPath string, requestContext context.Context) (accepted bool, prefixToStrip string, completedContext context.Context) {
+			targetCluster, apiDomain, prefixToStrip, ok := digestURL(urlPath, rootPathPrefix)
+			if !ok {
+				return false, "", requestContext
+			}
+
+			if targetCluster.Wildcard {
+				return false, "", requestContext
+			}
+
+			parsedKey, err := apidomainkey.Parse(apiDomain)
+			if err != nil {
+				return false, "", requestContext
+			}
+
+			if targetCluster.Name != parsedKey.CachedResourceCluster {
+				return false, "", requestContext
+			}
+
+			// We only accept requests for CachedResource's local cluster.
+
+			completedContext = genericapirequest.WithCluster(requestContext, targetCluster)
+			completedContext = dynamiccontext.WithAPIDomainKey(completedContext, apiDomain)
+			return true, prefixToStrip, completedContext
+		}),
+		Authorizer: newAuth(kubeClusterClient),
+		ReadyChecker: framework.ReadyFunc(func() error {
+			select {
+			case <-readyCh:
+				return nil
+			default:
+				return errors.New("replication virtual workspace controllers are not started")
+			}
+		}),
+		BootstrapAPISetManagement: func(mainConfig genericapiserver.CompletedConfig) (apidefinition.APIDefinitionSetGetter, error) {
+			if err := mainConfig.AddPostStartHook(replication.VirtualWorkspaceName, func(hookContext genericapiserver.PostStartHookContext) error {
+				defer close(readyCh)
+
+				indexers.AddIfNotPresentOrDie(
+					globalKcpInformers.Cache().V1alpha1().CachedObjects().Informer().GetIndexer(),
+					cache.Indexers{
+						cachedresourcesreplication.ByGVRAndLogicalClusterAndNamespace: cachedresourcesreplication.IndexByGVRAndLogicalClusterAndNamespace,
+					},
+				)
+				indexers.AddIfNotPresentOrDie(
+					globalKcpInformers.Apis().V1alpha2().APIExports().Informer().GetIndexer(),
+					cache.Indexers{
+						indexers.ByLogicalClusterPathAndName: indexers.IndexByLogicalClusterPathAndName,
+					},
+				)
+				indexers.AddIfNotPresentOrDie(
+					localKcpInformers.Apis().V1alpha2().APIExports().Informer().GetIndexer(),
+					cache.Indexers{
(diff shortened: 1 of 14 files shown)


pkg/reconciler/cache/cachedresourceendpointslice/cachedresourceendpointslice_controller.go
```diff
@@ -0,0 +1,250 @@
...
+			return globalShardClusterInformer.Cluster(core.RootCluster).Lister().Get(shardName)
+		},
+		getLogicalCluster: func(clusterName logicalcluster.Name) (*corev1alpha1.LogicalCluster, error) {
+			return lcClusterInformer.Cluster(clusterName).Lister().Get("cluster")
+		},
+		getAPIBinding: func(clusterName logicalcluster.Name, bindingName string) (*apisv1alpha2.APIBinding, error) {
+			return apiBindingClusterInformer.Cluster(clusterName).Lister().Get(bindingName)
+		},
+		cachedResourceEndpointSliceClusterInformer: cachedResourceEndpointSliceClusterInformer,
+		commit: committer.NewCommitter[*CachedResourceEndpointSlice, Patcher, *CachedResourceEndpointSliceSpec, *CachedResourceEndpointSliceStatus](kcpClusterClient.CacheV1alpha1().CachedResourceEndpointSlices()),
```
mjudeikis: This looks a bit strange here. Why are they not in the reconciler itself?

pkg/reconciler/cache/cachedresourceendpointslice/cachedresourceendpointslice_controller.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	logger := logging.WithReconciler(klog.Background(), ControllerName)
+
+	_, _ = cachedResourceEndpointSliceClusterInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			c.enqueueCachedResourceEndpointSlice(objOrTombstone[*cachev1alpha1.CachedResourceEndpointSlice](obj), logger, "")
```
mjudeikis: Drop the logsuffix if we dont use it. Usually we use it where are multipe enqueue entrypoints and we want to note reason for it.

pkg/reconciler/cache/cachedresourceendpointslice/cachedresourceendpointslice_reconcile.go
```diff
@@ -0,0 +1,133 @@
...
+		// Should never happen
+		logger = logging.WithObject(logger, shard)
+		logger.Error(
+			err, "error parsing shard.spec.virtualWorkspaceURL",
+			"VirtualWorkspaceURL", shard.Spec.VirtualWorkspaceURL,
+		)
+		return reconcileStatusStop, nil
+	}
+
+	addr.Path = path.Join(
```
mjudeikis, line 112: please add comment how this url looks like? Just so reader knowns

pkg/reconciler/cache/cachedresourceendpointslice/cachedresourceendpointslice_reconcile.go
```diff
@@ -0,0 +1,133 @@
...
+	addrUrl := addr.String()
+
+	for _, endpoint := range endpoints.Status.CachedResourceEndpoints {
+		if endpoint.URL == addrUrl {
+			// Already in endpoints slice, nothing to do.
+			return reconcileStatusContinue, nil
+		}
+	}
+
+	endpoints.Status.CachedResourceEndpoints = append(endpoints.Status.CachedResourceEndpoints, cachev1alpha1.CachedResourceEndpoint{
```
mjudeikis: What about delete? Who removes urls? 
And is this right to assume that now only 1 shard will append urls? so we will not have 2 urls for 2 shards?

mjudeikis: I think this needs to have bit TODO to follow-up with "consumer awarness" same as apiexportsendpoint slides has it. 
Is this the case here?

pkg/reconciler/cache/cachedresources/cachedresources_reconcile_endpointslice.go
```diff
@@ -0,0 +1,79 @@
...
+				Spec: cachev1alpha1.CachedResourceEndpointSliceSpec{
+					CachedResource: cachev1alpha1.CachedResourceReference{
+						Name: cachedResource.Name,
+					},
+				},
+			}
+			if err = r.createEndpointSlice(ctx, clusterName.Path(), &cachedResourceEndpointSlice); err != nil {
+				return reconcileStatusStopAndRequeue, fmt.Errorf("error creating CachedResourceEndpointSlice for CachedResource %s|%s: %w", clusterName, cachedResource.Name, err)
+			}
+		} else {
```
mjudeikis, line 73: If somebody creates same name slice, and we get into this error with AlreadyExists - we basically get into hot loop.

We need to check if one already exists - if it does and we own it - update, if we dont own it - maybe raise a condition on CachedResource with warrning that we failed to create endpointslide due to conflict. and move on.

pkg/server/controllers.go
```diff
@@ -1724,6 +1726,7 @@ func (s *Server) installCacheController(ctx context.Context, config *rest.Config
 		s.DiscoveringDynamicSharedInformerFactory,
 		s.CacheKcpSharedInformerFactory,
 		cachedResourceInformer,
+		cachedResourceEndpointSliceInformer,
```
mjudeikis, line 1729: just do inline:
`s.KcpSharedInformerFactory.Cache().V1alpha1().CachedResourceEndpointSlices()`

mjudeikis: I think one above was my mistake :)

mjudeikis: It bit strange now that namespaces and secrets are inline, other not :)

pkg/virtual/replication/builder/build.go
```diff
@@ -0,0 +1,395 @@
...
+					return localKcpInformers.Core().V1alpha1().LogicalClusters().Cluster(cluster).Lister().Get(name)
+				},
+
+				getAPIBinding: func(cluster logicalcluster.Name, name string) (*apisv1alpha2.APIBinding, error) {
+					return localKcpInformers.Apis().V1alpha2().APIBindings().Cluster(cluster).Lister().Get(name)
+				},
+
+				getAPIExportByPath: func(path logicalcluster.Path, name string) (*apisv1alpha2.APIExport, error) {
+					return indexers.ByPathAndNameWithFallback[*apisv1alpha2.APIExport](
+						apisv1alpha1.Resource("apiexports"),
```
mjudeikis: dont we have this function in `apisv1alpha2`?

pkg/virtual/replication/builder/build.go
```diff
@@ -0,0 +1,395 @@
...
+					return globalKcpInformers.Apis().V1alpha1().APIResourceSchemas().Cluster(cluster).Lister().Get(name)
+				},
+
+				config:               mainConfig,
+				dynamicClusterClient: dynamicClusterClient,
+				storageProvider: func(ctx context.Context, dynamicClusterClientFunc forwardingregistry.DynamicClusterClientFunc, sch *apisv1alpha1.APIResourceSchema, version string) (apiserver.RestProviderFunc, error) {
+					return forwardingregistry.ProvideReadOnlyRestStorage(
+						ctx,
+						dynamicClusterClientFunc,
+						withUnwrapping(sch, version, globalKcpInformers),
```
mjudeikis: sch what? :D just name it `apiResourceSchema`

mjudeikis (changes requested):
This is complicated PR and I expect it to have it fair issues but for now looks good. Mostly:
1. We know we need to start working on replication to see full picture, and smaller prs are hard in this case. Once we work on aggregation, we can fix what is not right here. Same as we did with caching.
2. Its feature-flag gates - so no expected usage yet.

in general few nits and lets move to final building block! 

P.s. Amazing work :) kubos for pushing it to the line!

pkg/reconciler/cache/cachedresourceendpointslice/cachedresourceendpointslice_controller.go
mjudeikis: but I think its ok. Easier to read here once I went via all pr. lets leave it.

pkg/server/controllers.go
ntnn: > I think one above was my mistake :)

That one is actually required because it's used in the returned wrapper below

pkg/server/controllers.go
```diff
@@ -1732,7 +1735,50 @@ func (s *Server) installCacheController(ctx context.Context, config *rest.Config
...
+	config = rest.CopyConfig(config)
+	config = rest.AddUserAgent(config, logicalclusterdeletion.ControllerName)
+	kcpClusterClient, err := kcpclientset.NewForConfig(config)
+	if err != nil {
+		return err
+	}
+	cachedResourceEndpointSliceInformer := s.KcpSharedInformerFactory.Cache().V1alpha1().CachedResourceEndpointSlices()
+	cachedResourceInformer := s.KcpSharedInformerFactory.Cache().V1alpha1().CachedResources()
+	lcClusterInformer := s.KcpSharedInformerFactory.Core().V1alpha1().LogicalClusters()
+	apiBindingClusterInfomer := s.KcpSharedInformerFactory.Apis().V1alpha2().APIBindings()
```
ntnn, line 1761: The apiBindingClusterInformer one can be inlined, or? I don't see it used in the controller wrapper

pkg/reconciler/cache/cachedresourceendpointslice/cachedresourceendpointslice_reconcile.go
ntnn: > What about delete? Who removes urls? And is this right to assume that now only 1 shard will append urls? so we will not have 2 urls for 2 shards?

I haven't read about the CachedResources too much yet - but I from what I understand at least for now there's a one-to-one relationship between shard, workspace, cachedresource and cachedresourceendpointslice:

```mermaid
flowchart TD

subgraph Shard
subgraph ProvderWS
CR1 --> Resource1
CR2 --> Resource2
CR1[CachedResource: Resource 1] --> CRES1[CachedResourceEndpointSlice: Resource 1]
CR2[CachedResource: Resource 2] --> CRES2[CachedResourceEndpointSlice: Resource 2]
end
end

CR1 -.-> |CachedObject1| Cache
CR2 -.-> |CachedObject2| Cache
Cache --> Shard2
Shard2 --> CRES1
Shard2 --> CRES2
```

So each CR has one CRES and each CRES has one URI (for now).
And if a CR is deleted the corresponding CRES is deleted as well.

If multiple shards should publish the same GVR that would have to be changed

mjudeikis: Its more like this:
image (https://link.excalidraw.com/l/2pL0lvjcXfb/6jW0OV53PIi) 

Where every shard has url pointing to local VirtualWorkspace instance for cachedResources. Hence - url per shard. But they are served from cache always, even locally to the source shard.

ntnn: Thanks - when a shard is responsible for cleaning up their own URLs upon shutdown we can expect orphans when a shard is non-gracefully shut down and never restored.
Other instances where this could happen is if a shard changes their URL, if that is possible (e.g. changed hostname, port, ...).

Question is how likely that is, but some housekeeping wouldn't be amiss. 

Maybe it would make sense that the "owning" shard of the CRES performs housekeeping and validates that all URLs are valid?
Although that would kind of defeat the original intent of distributing the load by having the shards register their own URLs.

Alternatively only the owning shard could maintain the VW URLs in the CRES. That would circumvent the problem entirely, with the impact being  that shards with many CachedResources would be busier.

mjudeikis: Currently, this is based on consumers - If there is a consumer, the URL is appended. 
For gradual shutdown - yes, it's a problem now. Not addressed in APIExportEndpointSlices too. 

So its winder problem (for now hypotetical, but problem), not related to this implementation. If we follow same what we have in APIExport now, we can address this in one go.

gman0: I've added a TODO comment there as discussed. This is to be done when implementing aggregation.

pkg/reconciler/cache/cachedresources/cachedresources_reconcile_endpointslice.go
gman0: Regarding the hot loop:

1. `getEndpointSlice`
2. `if err`
3. `if IsNotFound`
4. `createEndpointSlice`

If 4. fails with `AlreadyExists`, the work queue will retry. When the function is run again, 1. would succeed, making the function return success. No hot loop I think?

gman0: Regarding the condition: there is none for APIExports / APIExportEndpointSlices either:

https://github.com/kcp-dev/kcp/blob/1da313caf4b8d9af94cc0f0112038abcaae1f6cb/pkg/reconciler/apis/apiexport/apiexport_reconcile.go#L103-L135

Please let me know if you'd like me to add it anyway, or I open a ticket to add a condition for both endpoint slice implementations. But in general, is it necessary? If the point of auto-creating the slice is just for convenience, maybe it's good enough if it already exists?

pkg/server/controllers.go
gman0: It's just that I added a `HasSynced()` call into the wait function below.

gman0:
Before approving I can squash the commits from the reviews.

pkg/server/controllers.go
gman0: I forgot to add a `HasSynced()` call for that.

pkg/virtual/replication/builder/build.go
gman0: Done.

gman0:
Squashed.

mjudeikis:
/lgtm
/approve
We gonna come back to this once we start aggregation part for sure!

PR #3481, opened 2025-07-11
WIP - Per workspace authn

### LINKED ISSUE #3428: feature: per-workspace authentication configuration
### Feature Description

Currently, kcp allows to configure authentication (authN) via command line flags on the `kcp` binary or via a `AuthenticationConfiguration` file passed via a flag.

To make kcp more of a multi-tenant platform, it would be great to be able to configure authentication strategies on a per-workspace basis. This way, different workspaces could be configured for different OIDC providers (e.g. on a per-customer basis).

### Proposed Solution

Configuration of workspace-level authentication should be typed, i.e. we should probably introduce a new `WorkspaceAuthenticationConfiguration` resource in the tenancy API. `Workspaces` and `WorkspaceTypes` should be able to reference it, even across workspace boundaries (which means we need a new RBAC verb for this, i.e. `use`, that can be configured on the `WorkspaceAuthenticationConfiguration` objects.

An example could look like this:

```yaml
kind: WorkspaceAuthenticationConfiguration
metadata:
    name: oidc-tenant-1
spec:
    jwt:
    - issuer:
        url: https://example.com # Same as --oidc-issuer-url.
        discoveryURL: https://discovery.example.com/.well-known/openid-configuration
        certificateAuthority: <PEM encoded CA certificates>    
        audiences:
        - my-app # Same as --oidc-client-id.
        - my-other-app
```

Given that `WorkspaceTypes` would support this, we might skip inheritance from one workspace into child workspaces by default and instead allow configuring child workspaces via workspace type allowed children. This likely means that this resource needs to be replicated to the cache server.

Authentication should be additive, i.e. both the globally configured authentication and the per-workspace authentication should allow for accessing a workspace with this configured. The logical conclusion from this behaviour is that global authentication should only include the smallest possible set of users.

Special consideration should be given to RBAC across workspace boundaries. We might want to look into introducing some pseudo-groups that allow giving access to any user able to authenticate into a specific workspace instead of specific users or groups, since these might be arbitrary given the high amount of independence workspaces achieve with their own authentication configuration.

### Alternative Solutions

_No response_

### Want to contribute?

### Additional Context

_No response_

xrstf:
## What Type of PR Is This?
/kind feature

## Related Issue(s)
Fixes #3428

## Release Notes
```release-note
NONE
```


sdk/apis/tenancy/v1alpha1/types_workspacetype.go
```diff
@@ -147,6 +153,16 @@ type APIExportReference struct {
...
 
+// AuthenticationConfigurationReference provides the fields necessary to resolve a WorkspaceAuthenticationConfiguration.
+type AuthenticationConfigurationReference struct {
+	// configuration is the name of the WorkspaceAuthenticationConfiguration.
+	//
+	// +required
+	// +kubebuilder:validation:Required
+	// +kube:validation:MinLength=1
+	Configuration string `json:"configuration"`
+}
```
embik: not super happy with this type, can we change the field to `Name` to at least resemble an object reference?

test/e2e/authentication/workspace_test.go
```diff
@@ -0,0 +1,154 @@
...
+		// Initialization of the JWT authenticator happens asynchronously and depends both
+		// on the cluster index knowing the WorkspaceType and the authenticator itself
+		// performing OIDC discovery.
+
+		require.Eventually(t, func() bool {
+			_, err := authenticatedKubeClusterClient.Cluster(team1Path).CoreV1().ConfigMaps("default").List(ctx, metav1.ListOptions{})
+
+			return err == nil
+		}, wait.ForeverTestTimeout, 500*time.Millisecond)
+	}
```
embik: I think we should add a test to this (perhaps a dedicated test case, not sure) where we grant the token identity permissions in another workspace and we try to use the token against that workspace. It should fail to authenticate against that.

pkg/authentication/authentication_controller.go
```diff
@@ -0,0 +1,318 @@
...
+		clientGetter: clientGetter,
+		authIndex:    *NewIndex(ctx, baseAudiences),
+
+		shardIndexer: shardInformer.Informer().GetIndexer(),
+		shardLister:  shardInformer.Lister(),
+
+		shardWatchers: map[string]*shardWatcher{},
+	}
+
+	_, _ = shardInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
embik: This can return an error, let's catch it please.

pkg/authentication/authentication_controller.go
```diff
@@ -0,0 +1,318 @@
...
+}
+
+func NewShardWatcher(
+	ctx context.Context,
+	shardName string,
+	shardClient kcpclientset.ClusterInterface,
+	state *state,
+) *shardWatcher {
+	wacInformer := tenancyv1alpha1informers.NewWorkspaceAuthenticationConfigurationClusterInformer(shardClient, resyncPeriod, nil)
+	_, _ = wacInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
embik: Same here, this can error out.

pkg/authentication/authentication_controller.go
```diff
@@ -0,0 +1,318 @@
...
+			if final, ok := obj.(cache.DeletedFinalStateUnknown); ok {
+				obj = final.Obj
+			}
+			wac := obj.(*tenancyv1alpha1.WorkspaceAuthenticationConfiguration)
+			state.DeleteWorkspaceAuthenticationConfiguration(shardName, wac)
+		},
+	})
+
+	wtInformer := tenancyv1alpha1informers.NewWorkspaceTypeClusterInformer(shardClient, resyncPeriod, nil)
+	_, _ = wtInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
embik: Here as well.

pkg/server/config.go
```diff
@@ -457,6 +460,22 @@ func NewConfig(ctx context.Context, opts kcpserveroptions.CompletedOptions) (*Co
 			apiHandler = WithVirtualWorkspacesProxy(apiHandler, shardVirtualWorkspaceURL, virtualWorkspaceServerProxyTransport, proxy)
 		}
 
+		// Wrap authenticator with a per-workspace authenticator if desired. This authenticator
+		// requires the WorkspaceAuth middleware to have looked up and injected the relevant
+		// authenticator into the request context already.
+		var authIndex authentication.AuthenticatorIndex
+		if kcpfeatures.DefaultFeatureGate.Enabled(kcpfeatures.WorkspaceAuthentication) {
+			authIndexState := authentication.NewIndex(ctx, genericConfig.Authentication.APIAudiences)
+			authentication.NewShardWatcher(ctx, c.Options.Extra.ShardName, c.KcpClusterClient, authIndexState)
```
embik: Just to have this said out loud, we're creating a shard watcher here but not storing it anywhere nor queuing up a `.Done()` call. So we are relying on the context closing?

embik (commented):
A couple of notes. In addition, I have two important topics I couldn't quite place in the code:

1. We should make sure we attach a [scope](https://docs.kcp.io/kcp/main/concepts/authorization/authorizers/#scopes) to authenticated users in a workspace, so that their identity is only valid in the workspace they have authenticated for.
2. What happens with the mini-front-proxy in the server when a WorkspaceType / AuthenticationConfig in a different shard is referenced? Is that (not) supposed to work?

test/e2e/authentication/workspace_test.go
xrstf: Yes I was planning on extending the test to run 2 independent mockoidc issuers and have 2 different workspaces and try to use tokens against the other ones. Didn't get to it yet.

pkg/server/config.go
xrstf: Yes. I left a comment on this code now to mention that.

xrstf:
1. I've added that functionality and added an e2e test to verify it.
2. Yes, that's correct. The localproxy auth functionality only works if Workspace, WorkspaceType, LogicalCluste and WorkspaceAuthentication exist on the local shard. There is purposefully no intention to change this, as I would simply recommend using the front-proxy if you want this feature across all your shards. That's at least how I intended it.

Makefile
```diff
@@ -354,7 +355,7 @@ test-run-sharded-server: LOG_DIR ?= $(WORK_DIR)/.kcp
 test-run-sharded-server:
 	mkdir -p "$(LOG_DIR)" "$(WORK_DIR)/.kcp"
 	rm -f "$(WORK_DIR)/.kcp/ready-to-test"
-	UNSAFE_E2E_HACK_DISABLE_ETCD_FSYNC=true NO_GORUN=1 ./bin/sharded-test-server --quiet --v=2 --log-dir-path="$(LOG_DIR)" --work-dir-path="$(WORK_DIR)" --shard-run-virtual-workspaces=false --shard-feature-gates=$(TEST_FEATURE_GATES) $(TEST_SERVER_ARGS) --number-of-shards=2 2>&1 & PID=$$!; echo "PID $$PID" && \
+	UNSAFE_E2E_HACK_DISABLE_ETCD_FSYNC=true NO_GORUN=1 ./bin/sharded-test-server --quiet --v=2 --log-dir-path="$(LOG_DIR)" --work-dir-path="$(WORK_DIR)" --shard-run-virtual-workspaces=false --shard-feature-gates=$(TEST_FEATURE_GATES) --proxy-feature-gates=$(PROXY_FEATURE_GATES) $(TEST_SERVER_ARGS) --number-of-shards=2 2>&1 & PID=$$!; echo "PID $$PID" && \
```
mjudeikis, line 358: With second flag this looks strange now. 2 featureflags flags. Should we maybe get ticket to deprecate `--shard-feature-gates` in favor of `--feature-gates` and use use one everywhere?

cmd/kcp-front-proxy/options/options.go
```diff
@@ -43,6 +46,11 @@ func NewOptions() *Options {
 func (o *Options) AddFlags(fss *cliflag.NamedFlagSets) {
 	o.Proxy.AddFlags(fss)
 	logsapiv1.AddFlags(o.Logs, fss.FlagSet("logging"))
+
+	// add flags that are filtered out from upstream, but overridden here with our own version
+	fss.FlagSet("KCP").Var(kcpfeatures.NewFlagValue(), "feature-gates", ""+
```
mjudeikis, line 51: This this should be default ? appended?

docs/content/concepts/authentication/workspace.md
```diff
@@ -0,0 +1,143 @@
...
+
+To enable this, kcp supports per-workspace authentication. In this model, a `WorkspaceType` configures a set of additional OIDC validators that are then used by kcp in addition to the global authentication mechanisms configured with CLI flags. Every workspace using these custom workspace types will then have these additional auth methods available.
+
+This document describes how to enable and use this feature. Please refer to [OIDC Configuration](./oidc.md) for more information about the global OIDC configuration.
+
+## Feature Gate
+
+The feature is guarded by a feature gate called `WorkspaceAuthentication`, which is disabled by default. It can be independently enabled on any front-proxy and/or any kcp shard servers, though it is recommended and intended to enable it on front-proxies only. The shard support is mainly for testing and developing.
+
+Add `--feature-gates=WorkspaceAuthentication=true` to the CLI flags on the front-proxy to enable the feature.
```
mjudeikis, line 18: not `--proxy-feature-gates=`?

docs/content/concepts/authentication/workspace.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Overview
+
+Extra authentication for a workspace is configured using `WorkspaceAuthenticationConfiguration` (colloquially called "auth configs") objects, which can be thought of as CRD variants of the Kubernetes authentication configuration (as described in [OIDC Configuration](./oidc.md)). Each auth config contains a set of JWT validators that are capable of validating an incoming JWT bearer token.
+
+Workspace types then reference a set of auth configs, and their configuration will apply to all their workspaces/logicalclusters. Compared to many other settings in a `WorkspaceType` that work only as a preset for _new_ workspaces, the configured auth configs will continue to affect workspaces, so when a `WorkspaceType` is changed, this will impact existing workspaces, too.
```
mjudeikis, line 24: What is behaviour when we have nesting of WorkspaceTypes?

docs/content/concepts/authentication/workspace.md
```diff
@@ -0,0 +1,143 @@
...
+
+For every incoming HTTPS request, kcp will then resolve the logicalcluster, determine the used workspace type, assemble the list of auth configs and create an authenticator suitable for exactly the one logicalcluster targeted by the request. This workspace authenticator is an *alternative* to kcp's regular authentication (i.e. it forms a union with it).
```
mjudeikis, line 26: What is behaviour when you dont have set any core authentication and only AuthConfig?
One might not even have auth flags in kcp and just AuthConfig?

docs/content/concepts/authentication/workspace.md
```diff
@@ -0,0 +1,143 @@
...
+
+For example, suppose kcp is started with `--api-audiences=https://kcp.example.com` and there is a `WorkspaceAuthenticationConfiguration` that defines a JWT validator using the audience `https://corp.initech.com`. For a token to be admitted into a workspace that uses this auth config, the token will have to contain *both* audiences. This is to ensure the token is actually meant to be used in kcp, regardless of which audiences are then configured per workspace.
+
+## Limitations
+
+This feature has some small limitations that users should keep in mind:
+
+* As mentioned above, the JWT validation for a workspace is not 100% independent from the global kcp authentication: tokens will need to contain kcp's global API audience (configured with `--api-audiences`) and any audience configured in the auth configs. You cannot have a token not contain kcp's global audience.
+* `WorkspaceAuthenticationConfiguration` objects must reside in the same logicalcluster as the `WorkspaceType`.
```
mjudeikis, line 44: Do we have example how to make sure this happens or what does it mean from the "hands on experience"?

docs/content/concepts/authentication/workspace.md
```diff
@@ -0,0 +1,143 @@
...
+* The workspace authentication in the localproxy, as part of a single shard server, only knows about the data on the local shard and cannot handle cross-shard authentication. Users are advised to use the front-proxy instead.
+* Even when the feature is disabled on all shards and all front-proxies, the API (CRDs) are always available in kcp. Admins might uses RBAC or webhooks to prevent creating `WorkspaceAuthenticationConfiguration` objects if needed.
+
+## Example
+
+In this example we want to create a workspace where users with tokens from our local OIDC provider are admitted to.
+
+### Step 0: Enabling the Feature
+
+Add `--feature-gates=WorkspaceAuthentication=true` to the CLI flags on the front-proxy to enable the feature. When developing or just testing, you can also add the feature gate to the kcp process like
```
mjudeikis, line 55: again, is this right flag?

docs/content/concepts/authentication/workspace.md
```diff
@@ -0,0 +1,143 @@
...
+metadata:
+  name: make-externals-admins
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
+subjects:
+  - apiGroup: rbac.authorization.k8s.io
+    kind: Group
+    name: oidc:admins
```
mjudeikis, line 129: Your example does not have `oidc:` prefix. This will not work, right?

mjudeikis: Can one create group `system:masters` or system:kcp:admin` in external authentication OIDC and so get wider access than one should? 

Thinking if we should force everybody to use prefix for authentication configs? Or prevent certain groups in these authConfigs. 

@embik wdyt?

pkg/index/index.go
```diff
@@ -41,6 +42,9 @@ type Result struct {
 	Shard string
 	// Cluster canonical path
 	Cluster logicalcluster.Name
+	// Type is not set for mounted workspaces. For all others this value is the
+	// fully-qualified name of the type, e.g. "root:universal".
+	Type logicalcluster.Path
```
mjudeikis, line 47: Why not `LogicalClusterType` ? Type read like it was from "WorkspaceType"?

pkg/proxy/server.go
```diff
@@ -45,44 +49,91 @@ type Server struct {
...
+		s.completedConfig.AuthenticationInfo.Authenticator,
 		s.CompletedConfig.AdditionalAuthEnabled)
 
+	// Make the per-workspace authenticator available to the previous middleware
+	// by hooking up a handler and a runtime index.
+	hasWorkspaceAuth := hasShardMapping && kcpfeatures.DefaultFeatureGate.Enabled(kcpfeatures.WorkspaceAuthentication)
+
+	if hasWorkspaceAuth {
+		// This controller is similar to the index controller, but keeps track of the per-workspace authenticators.
+		ctrl, err := authentication.NewController(ctx, s.KcpSharedInformerFactory.Core().V1alpha1().Shards(), getClientFunc, nil)
```
mjudeikis, line 118: Do you ever see this merging into index controller or we keep it as featureflag forever?

mjudeikis (commented):
More questions than issues :) looks very promising!

docs/content/concepts/authentication/workspace.md
embik: Hm, good question. We might have to do that indeed. I wonder if we should have a prefix prepended that is simply the cluster ID, even though that would be massively annoying for creating RBAC ...

Makefile
xrstf: Because I needed to be able to control the feature independently between shard and front-proxy, because some tests are meant to ensure that the front-proxy does the authentication. That's why **the sharded-test-server** has a new CLI flag. The front-proxy's flag is called `--feature-gate`, like the shard's CLI flag is. Only on the test server do we need 2 sets of feature gates, hence 2 CLI flags.

docs/content/concepts/authentication/workspace.md
xrstf: No, the `proxy-` prefix is only on the sharded-test-server, not on the front-proxy itself.

docs/content/concepts/authentication/workspace.md
xrstf: I considered implementing support for extending workspaces, but deemed it too complicated for the initial implementation and was also not sure if it's really needed. I think if we really want it, we could add support for that in a dedicated PR.

docs/content/concepts/authentication/workspace.md
xrstf: If you do not have any auth flags configured on the front-proxy, it will pass through all requests targetting clusters without AuthConfigs. For clusters with AuthConfigs (via their types), the front-proxy will perform authentication and reject the request if it cannot be authenticated.

Without any auth flags, it will be difficult to access the root workspace I think.

docs/content/concepts/authentication/workspace.md
xrstf: For the first bullet point, yes, there are e2e tests that test that your tokens need to have kcp's global audience configured to be admitted.

For the second bullet point, it's simply the fact that references from WorkspaceTypes to AuthConfigs cannot have a path. All you can configure is the name of the AuthConfig, so they logically must be in the same logicalcluster as the WorkspaceType that is referring to them.

docs/content/concepts/authentication/workspace.md
xrstf: I'm pretty sure it is :)

docs/content/concepts/authentication/workspace.md
xrstf: I'm not sure this is a problem. Yes the example could show the username claim with the "oidc:" prefix, but if you have AuthConfigs that do not configure a claim prefix, then yes, users could get any arbitrary group assigned to them. But their authentication is limitted (via the scope annotation thingy on the UserInfo) to just one single logicalcluster, so it should be fine. And who are we to prevent kcp users from giving someone poweruser permissions inside their clusters?

But I'm open to explicit rules you guys want to see here.

pkg/index/index.go
xrstf: Because in my world when I implemented this, only workspaces had types, not logicalclusters. Nowhere in kcp code have I yet seen the wording "logical cluster type", but if that's what I should use, sure, if you want :D

pkg/proxy/server.go
xrstf: Since the feature gate is not just about gradually enabling a feature that matures, but about adding extra authentication, I would consider this feature gate to be permanent (originally it was a CLI flag, but handling it as a FG was simpler). In my idea admins should forever be able to turn this on or off based on each pod, because the "open auth" front-proxy might be behind a dedicated firewall or something.

One could think about merging the two indexes together, but I did not want to conflate Lookup() doing both shard/cluster looks *and* constructing authenticators, so it seemed to make sense to have dedicated indexes/controllers for this.

cmd/kcp-front-proxy/options/options.go
xrstf: How do you mean? My thought was that there were no FGs on the front-proxy yet, so I can just copy-paste this into it.

docs/content/concepts/authentication/workspace.md
embik: We have to avoid privileged identities (the one @mjudeikis mentioned) because they can reach *past* the workspace and circumvent authorization. I'm not sure that scopes are enough for keeping them in check. Maybe the authenticator should reject anything with a `system:` prefix.

docs/content/concepts/authentication/workspace.md
mjudeikis: ack. make sense, as long as we know the limitations. So last one wins

docs/content/concepts/authentication/workspace.md
mjudeikis: This is the point, They might be scoped to workspace, but Im not even sure what other "holes" they could poke with this.

pkg/proxy/server.go
mjudeikis: ack

pkg/index/index.go
mjudeikis: ok, all this is internal :) we can always rename :D

docs/content/concepts/authentication/workspace.md
mjudeikis: > Without any auth flags, it will be difficult to access the root workspace I think.
you could just use admin kubeconfig as admin. Its a security feature - you know there is no OIDC access to anything else apart workspaces, using it.

docs/content/concepts/authentication/workspace.md
xrstf: Do we want to reject non-prefixed AuthSettings or eventually forbid the WorkspaceAuthenticator from returning a set of protected groups, regardless of the OIDC settings?

xrstf: I've added a wrapper that removes system: groups and rejects system: names outright. A new e2e test ensures that this works.

embik (approved):
/approve

Issue #3444, opened 2025-06-06
flake: TestAPIExportBindingAuthorizer - waiting for virtual workspace URLs to be available


embik:
It looks like we have a new flake in `TestAPIExportBindingAuthorizer` on our hands:

- https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3381/pull-kcp-test-e2e-sharded/1930557252733767680
- https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3441/pull-kcp-test-e2e-sharded/1930218163291557888
- https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3441/pull-kcp-test-e2e-sharded/1930208212900909056

The error in question is:

```
    authorizer_test.go:624: Create virtual workspace client for "today-sherriffs" APIExport in workspace "root:e2e-workspace-5xk59:service-provider"
    assertions.go:1949: Waiting for condition, but got: waiting for virtual workspace URLs to be available: [{https://127.0.0.1:7445/services/apiexport/medira8q7zlk16xz/wild.wild.west}]
    authorizer_test.go:628: 
        	Error Trace:	/home/prow/go/src/github.com/kcp-dev/kcp/sdk/testing/helpers/eventually.go:40
        	            				/home/prow/go/src/github.com/kcp-dev/kcp/test/e2e/virtual/apiexport/authorizer_test.go:628
        	Error:      	Condition never satisfied
        	Test:       	TestAPIExportBindingAuthorizer

```

PR #3486, opened 2025-07-14
Move permission claim object selectors from APIExport to APIBinding

### LINKED ISSUE #3456: api: Clean up the PermissionClaims API in v1alpha2
We introduced new API versions of APIExports and APIBindings (v1alpha2) which includes PermissionClaims as well. We should use this opportunity to define this API properly and clean up the deprecated/unused fields.

- Consider removing the `All` field in favor of an empty label selector https://github.com/kcp-dev/kcp/blob/e747d521d6f087e7f20b2d0343e977374f6b6903/sdk/apis/apis/v1alpha2/types_apiexport.go#L226
- Remove the `ResourceSelector` field as it's no implemented https://github.com/kcp-dev/kcp/blob/e747d521d6f087e7f20b2d0343e977374f6b6903/sdk/apis/apis/v1alpha2/types_apiexport.go#L236-L239
- Add a new field that will be used for all of: `all` (empty label selector), label selectors, resource selectors

embik:
## Summary

This PR implements the API changes we wanted to get into `v1alpha2` before we cut the release solidifying the API version. It's been mainly discussed on Slack: https://kubernetes.slack.com/archives/C021U8WSAFK/p1749552577773639. I don't think a strong consensus has been reached but the general direction is what this PR proposes.

It removes the `all` and `resourceSelector` fields from the `PermissionClaim` type (i.e. remove the object selector configuration from the `APIExport` on the service provider side). It instead adds a `selector` field in `AcceptablePermissionClaim` (the wrapper type used for permission claims in `APIBindings`) which allows the API consumer to scope access of an API provider to a specific resource.

Conversions have been added so that if a permission claim in an `APIBinding` had `all: true`, it would automatically translate into `selector.matchAll: true` (and back). To allow API roundtripping without loss of information, permission claims are now stored in an annotation when converting from v1alpha1 to v1alpha2, so that `resourceSelector` can be restored upon converting back to v1alpha1.

A few more things this PR does:

- Update the usage of `APIExport` and `APIBindings` across the code base.
- Split conversion tests in `v1alpha2` package into two files.
- Rename `PermissionClaim.Equal` to `PermissionClaim.EqualGRI`. `Equal` is used by `cmp` to compare objects, so (acceptable) permission claims were only compared partially, which shadowed some bugs in my code early on.
- Switch compared objects in conversion tests so that the diff output was more readable.

## What Type of PR Is This?
/kind cleanup
/kind api-change

## Related Issue(s)

Fixes #3456

## Release Notes

```release-note
NONE
```


embik:
/hold

Since I want to discuss this PR with @xmudrii.

sdk/apis/roundtrip_test.go
```diff
@@ -131,6 +142,9 @@ func testGenericConversion[V1 runtime.Object, V2 runtime.Object](
 		err = toV1(intermediate, result, nil)
 		require.NoError(t, err)
 
+		// remove annotations that we expect on the object due to conversion data stored.
+		removeAnnotations(t, result, v1ExpectedAnnotations)
```
ntnn, line 175: Nitpicky but `removeAnnotations` and `expectedAnnotation` reads antithetical to me. I'd expect `expectedAnnotations` to be the annotations I expect on the final object.

Maye `v1TrimAnnotations` to explicitly say that these will be removed?

embik: the idea here was that for the purpose of comparison, we "expect" this annotation to be there, so we need to remove it before checking equality. But I don't have strong feelings about renaming it.

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
```diff
@@ -0,0 +1,203 @@
...
+
+	apisv1alpha1 "github.com/kcp-dev/kcp/sdk/apis/apis/v1alpha1"
+)
+
+const (
+	AcceptablePermissionClaimsAnnotation = "apis.v1alpha2.kcp.io/acceptable-permission-claims"
+)
+
+// v1alpha2 -> v1alpha1
+
```
ntnn:
```suggestion
```

Nitpick but I think the functions are named clear enough, or?

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
```diff
@@ -0,0 +1,203 @@
...
+	return nil
+}
+
+func Convert_v1alpha2_ScopedPermissionClaim_To_v1alpha1_PermissionClaim(in *ScopedPermissionClaim, out *apisv1alpha1.PermissionClaim, s kubeconversion.Scope) error {
+	return Convert_v1alpha2_PermissionClaim_To_v1alpha1_PermissionClaim(&in.PermissionClaim, out, s)
+}
+
+// v1alpha1 -> v1alpha2
+
```
ntnn:
```suggestion
```

As above

sdk/apis/apis/v1alpha2/types_apibinding.go
```diff
@@ -96,6 +115,17 @@ const (
...
 )
 
+// PermissionClaimSelector configures scoped access to objects
+// of a claimed resource.
+//
+// +kubebuilder:validation:XValidation:rule="(has(self.matchAll) && self.matchAll)",message="a selector is required. Only \"matchAll\" is currently implemented"
+type PermissionClaimSelector struct {
+	metav1.LabelSelector `json:",inline"`
+
+	// matchAll grants full access to all objects of the claimed resource.
```
xmudrii: The `full` access depends on verbs, it only grants access to all objects.
```suggestion
	// matchAll grants access to all objects of the claimed resource.
```

sdk/apis/roundtrip_test.go
ntnn: Ahh okay that makes sense. Yeah under that lens the name is appropriate.
But then I'd turn that around and have `removeAnnotations` yield an error if the annotation is missing but expected.

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
embik: this comment basically denotes the whole block following. I found it easier to have especially the second comment as a break to signal "it's changing now from v2 -> v1 to v1 -> v2". Before it, this file was a bit messy 🤷🏻

xmudrii: +1, I find this comment to help with reading this file even though we already have it in function names

sdk/apis/roundtrip_test.go
ntnn: Although it could be argued that if the annotation is missing that the conversion overall failed and that the test will fail anyhow.

sdk/apis/apis/v1alpha2/types_apiexport_conversion_test.go
```diff
@@ -115,9 +114,15 @@ func TestConvertV1Alpha2APIExports(t *testing.T) {
 				t.Fatalf("Failed to convert v1alpha1 back to v1alpha2: %v", err)
 			}
 
+			// Drop annotation we know this object will have from v1alpha1 -> v1alpha2 conversion.
 			v2.GetObjectKind().SetGroupVersionKind(testcase.GroupVersionKind())
+			v2obj := v2.(*APIExport)
+			delete(v2obj.Annotations, PermissionClaimsV1Alpha1Annotation)
+			if len(v2obj.Annotations) == 0 {
+				v2obj.Annotations = nil
+			}
```
xmudrii, line 123: This is fine, but we should have a quick test to make sure the behavior around this annotation is correct.

sdk/apis/roundtrip_test.go
xmudrii: I'd also extend the comment to clarify this properly, at the moment, it's not very clear out from the comment alone.

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
```diff
@@ -0,0 +1,203 @@
...
+	if originalPermissionClaims, ok := in.Annotations[PermissionClaimsV1Alpha1Annotation]; ok {
+		permissionClaims := []apisv1alpha1.AcceptablePermissionClaim{}
+		if err := json.Unmarshal([]byte(originalPermissionClaims), &permissionClaims); err != nil {
+			return fmt.Errorf("failed to decode schemas from JSON: %w", err)
+		}
+
+		for _, pc := range permissionClaims {
+			for i, opc := range out.Spec.PermissionClaims {
+				if pc.PermissionClaim.EqualGRI(opc.PermissionClaim) {
+					out.Spec.PermissionClaims[i].ResourceSelector = pc.ResourceSelector
```
xmudrii, line 75: Is it expected that we don't have `All` here as in the APIExports conversion?

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
```diff
@@ -0,0 +1,203 @@
...
+func Convert_v1alpha2_AcceptablePermissionClaims_To_v1alpha1_AcceptablePermissionClaims(in []AcceptablePermissionClaim, s kubeconversion.Scope) (out []apisv1alpha1.AcceptablePermissionClaim, overhanging []AcceptablePermissionClaim, err error) {
+	for _, apc := range in {
+		if len(apc.PermissionClaim.Verbs) == 1 && apc.PermissionClaim.Verbs[0] == "*" {
+			var v1pc apisv1alpha1.PermissionClaim
+			err := Convert_v1alpha2_PermissionClaim_To_v1alpha1_PermissionClaim(&apc.PermissionClaim, &v1pc, s)
+			if err != nil {
+				return nil, nil, err
+			}
+
+			out = append(out, apisv1alpha1.AcceptablePermissionClaim{PermissionClaim: v1pc})
```
xmudrii: We are missing `state` here or do we handle it somewhere else?

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	return Convert_v1alpha2_APIBindingStatus_To_v1alpha1_APIBindingStatus(&in.Status, &out.Status, s)
+}
+
+func Convert_v1alpha2_AcceptablePermissionClaims_To_v1alpha1_AcceptablePermissionClaims(in []AcceptablePermissionClaim, s kubeconversion.Scope) (out []apisv1alpha1.AcceptablePermissionClaim, overhanging []AcceptablePermissionClaim, err error) {
+	for _, apc := range in {
+		if len(apc.PermissionClaim.Verbs) == 1 && apc.PermissionClaim.Verbs[0] == "*" {
+			var v1pc apisv1alpha1.PermissionClaim
+			err := Convert_v1alpha2_PermissionClaim_To_v1alpha1_PermissionClaim(&apc.PermissionClaim, &v1pc, s)
```
xmudrii: Why is this not calling `Convert_v1alpha2_AcceptablePermissionClaim_To_v1alpha1_AcceptablePermissionClaim` as this is about `AcceptablePermissionClaims`?

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
xmudrii: Ignore, it's expected, it's handled in `Convert_v1alpha2_AcceptablePermissionClaim_To_v1alpha1_AcceptablePermissionClaim`

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
```diff
@@ -0,0 +1,203 @@
...
+			out = append(out, apisv1alpha1.AcceptablePermissionClaim{PermissionClaim: v1pc})
+		} else {
+			overhanging = append(overhanging, apc)
+		}
+	}
+
+	return
+}
+
+func Convert_v1alpha2_AcceptablePermissionClaim_To_v1alpha1_AcceptablePermissionClaim(in *AcceptablePermissionClaim, out *apisv1alpha1.AcceptablePermissionClaim, s kubeconversion.Scope) error {
```
xmudrii, line 217: This should have a comment that this is not a loseless conversion

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
```diff
@@ -0,0 +1,203 @@
...
+	if err := Convert_v1alpha2_PermissionClaim_To_v1alpha1_PermissionClaim(&in.PermissionClaim, &out.PermissionClaim, s); err != nil {
+		return err
+	}
+	out.State = apisv1alpha1.AcceptablePermissionClaimState(in.State)
+	out.All = in.Selector.MatchAll
+
+	return nil
+}
+
+func Convert_v1alpha2_ScopedPermissionClaim_To_v1alpha1_PermissionClaim(in *ScopedPermissionClaim, out *apisv1alpha1.PermissionClaim, s kubeconversion.Scope) error {
```
xmudrii, line 227: This should have a comment that this is not a loseless conversion

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
```diff
@@ -0,0 +1,203 @@
...
+		} else {
+			overhanging = append(overhanging, apc)
+		}
+	}
+
+	return
+}
+
+func Convert_v1alpha2_AcceptablePermissionClaim_To_v1alpha1_AcceptablePermissionClaim(in *AcceptablePermissionClaim, out *apisv1alpha1.AcceptablePermissionClaim, s kubeconversion.Scope) error {
+	if err := Convert_v1alpha2_PermissionClaim_To_v1alpha1_PermissionClaim(&in.PermissionClaim, &out.PermissionClaim, s); err != nil {
```
xmudrii: This should call `Convert_v1alpha2_ScopedPermissionClaim_To_v1alpha1_PermissionClaim` instead

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
```diff
@@ -0,0 +1,203 @@
...
+		return err
+	}
+	out.State = apisv1alpha1.AcceptablePermissionClaimState(in.State)
+	out.All = in.Selector.MatchAll
```
xmudrii: This should go in `Convert_v1alpha2_ScopedPermissionClaim_To_v1alpha1_PermissionClaim`

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
```diff
@@ -0,0 +1,203 @@
...
+
+	if err := Convert_v1alpha1_APIBindingSpec_To_v1alpha2_APIBindingSpec(&in.Spec, &out.Spec, s); err != nil {
+		return err
+	}
+	if err := Convert_v1alpha1_APIBindingStatus_To_v1alpha2_APIBindingStatus(&in.Status, &out.Status, s); err != nil {
+		return err
+	}
+
+	// store acceptable permission claims in annotation. this is necessary for a clean conversion of
+	// FieldSelector, which went away in v1alpha2.
```
xmudrii:
```suggestion
	// ResourceSelector, which went away in v1alpha2.
```

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
```diff
@@ -0,0 +1,203 @@
...
+			out = append(out, v2apc)
+		} else {
+			overhanging = append(overhanging, apc)
+		}
+	}
+
+	return
+}
+
+func Convert_v1alpha1_AcceptablePermissionClaim_To_v1alpha2_AcceptablePermissionClaim(in *apisv1alpha1.AcceptablePermissionClaim, out *AcceptablePermissionClaim, s kubeconversion.Scope) error {
```
xmudrii, line 443: This function appears to be unused, do we need it?

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
```diff
@@ -0,0 +1,203 @@
...
+	out.Selector.MatchAll = in.PermissionClaim.All
+	if len(out.PermissionClaim.Verbs) == 0 {
+		out.PermissionClaim.Verbs = []string{"*"}
+	}
+
+	return nil
+}
+
+func Convert_v1alpha1_PermissionClaim_To_v1alpha2_ScopedPermissionClaim(in *apisv1alpha1.PermissionClaim, out *ScopedPermissionClaim, s kubeconversion.Scope) error {
+	return Convert_v1alpha1_PermissionClaim_To_v1alpha2_PermissionClaim(in, &out.PermissionClaim, s)
```
xmudrii: Shouldn't we also set `MatchAll` in this conversion given its available?

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
embik: Changing it but it doesnt matter because this func is only used to determine the overhanging AcceptablePermissionClaims. The actual conversion uses `Convert_v1alpha2_AcceptablePermissionClaim_To_v1alpha1_AcceptablePermissionClaim` instead. Don't ask me why, it's basically been like that before.

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
embik: Not unused, it's used right above in `Convert_v1alpha1_AcceptablePermissionClaims_To_v1alpha2_AcceptablePermissionClaims`.

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
embik: Updated.

sdk/apis/roundtrip_test.go
embik: Updated.

sdk/apis/apis/v1alpha2/types_apibinding.go
embik: Updated.

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
embik: Updated.

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
embik: Updated.

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
embik: Updated.

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
embik: Updated.

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
embik: Updated.

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
embik: Updated.

mjudeikis:
/hold

I think this is wrong. This moves permission claim responsibility to APIBinding from APIExport. This means that now you can have a need to check " accepted claims and state of them" on each binding. This will put big operational strain on the provider. Meaning on each reconcile loop provider needs to check what is the state of APIBinding and adjust reconciler for this. If user says 'selector=label=foo' provider will need to inject this.

mjudeikis:
/hold cancel
Plan is to mutate on provider side so should be ok there,

sdk/apis/apis/fuzzer/fuzzer.go
```diff
@@ -86,7 +86,6 @@ func Funcs(codecs runtimeserializer.CodecFactory) []interface{} {
 						Group:    group,
```
mjudeikis, line 86: Im suprised you didnt needed to add some fields above here for APIBindings, does fuzzing test runs for these?

cli/pkg/bind/plugin/bind.go
```diff
@@ -235,7 +235,7 @@ func (b *BindOptions) parsePermissionClaim(claim string, accepted bool) error {
 		parsedClaim.State = apisv1alpha2.ClaimRejected
 	}
 	// TODO(mjudeikis): Once we add support for selectors/
-	parsedClaim.All = true
+	parsedClaim.Selector = apisv1alpha2.PermissionClaimSelector{MatchAll: true}
```
xrstf, line 238: What's this TODO left to hint at?

sdk/apis/roundtrip_test.go
```diff
@@ -60,11 +62,13 @@ func TestRoundTripTypes(t *testing.T) {
...
+	name                  string
+	v1                    runtime.Object
+	v2                    runtime.Object
+	toV2                  func(in, out runtime.Object, s kubeconversion.Scope) error
+	toV1                  func(in, out runtime.Object, s kubeconversion.Scope) error
+	v1ExpectedAnnotations []string
+	v2ExpectedAnnotations []string
 }
 
 func TestConversion(t *testing.T) {
```
mjudeikis, line 75: Should we add APIBindings there as test cases too?

sdk/apis/apis/v1alpha2/types_apibinding.go
```diff
@@ -74,12 +74,31 @@ type APIBindingSpec struct {
...
+	// +listMapKey=identityHash
 	PermissionClaims []AcceptablePermissionClaim `json:"permissionClaims,omitempty"`
 }
 
+// ScopedPermissionClaim embeds a PermissionClaim and adds a selector to
+// scope down access to objects of the claimed resource.
+type ScopedPermissionClaim struct {
+	PermissionClaim `json:",inline"`
+
+	// selector configures which objects for the claimed resource
```
xrstf, line 89: Full sentence should start with a capital letter.

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
```diff
@@ -0,0 +1,239 @@
...
+		if len(out.Annotations) == 0 {
+			out.Annotations = nil
+		}
+	}
+
+	return Convert_v1alpha2_APIBindingStatus_To_v1alpha1_APIBindingStatus(&in.Status, &out.Status, s)
+}
+
+// Convert_v1alpha2_AcceptablePermissionClaims_To_v1alpha1_AcceptablePermissionClaims converts v1alpha2.AcceptablePermissionClaims
+// to v1alpha1.AcceptablePermissionClaims. This is not a loseless conversion, verbs and selectors are lost in this conversion.
```
xrstf: *lossless* not *loseless*

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
```diff
@@ -0,0 +1,239 @@
...
+			overhanging = append(overhanging, apc)
+		}
+	}
+
+	return
+}
+
+// Convert_v1alpha2_AcceptablePermissionClaim_To_v1alpha1_AcceptablePermissionClaim converts v1alpha2.AcceptablePermissionClaim
+// to v1alpha1.AcceptablePermissionClaim. This is not a lossless conversion, selectors are lost in this conversion.
+// For loseless conversion use Convert_v1alpha2_APIBinding_To_v1alpha1_APIBinding.
```
xrstf: lossless

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
xrstf: Oh THIS is where the "loseless" typo came from 😁

sdk/apis/apis/v1alpha2/types_apibinding_conversion.go
```diff
@@ -0,0 +1,239 @@
...
+			out.Annotations = make(map[string]string)
+		}
+		encoded, err := json.Marshal(overhangingV1PC)
+		if err != nil {
+			return err
+		}
+		out.Annotations[PermissionClaimsV1Alpha1Annotation] = string(encoded)
+	}
+
+	// store v1alpha2 permission claims in annotation. this is necessary for a clean conversion of
```
xrstf: Mixing up so many sentences with lowercase characters.

sdk/apis/apis/v1alpha2/types_apibinding_test.go
```diff
@@ -105,3 +105,110 @@ func TestAPIBindingPermissionClaimCELValidation(t *testing.T) {
...
+				"matchAll": true,
+			},
+		},
+		{
+			name:    "nothing set",
+			current: map[string]any{},
+			old:     map[string]any{},
+			wantErrs: []string{
+				"openAPIV3Schema.properties.spec.properties.permissionClaims.items.properties.selector: Invalid value: \"object\": a selector is required. Only \"matchAll\" is currently implemented",
+			},
```
xrstf, line 131: Just for the record, I hate asserting exact error strings.

sdk/apis/apis/v1alpha2/types_apibinding_test.go
```diff
@@ -105,3 +105,110 @@ func TestAPIBindingPermissionClaimCELValidation(t *testing.T) {
...
+			errs := validator(tc.current, tc.old)
+			if len(errs) > 0 {
+				for _, err := range errs {
+					t.Logf("'%s': observed validation error: %s", tc.name, err)
+				}
+			}
+
+			if got := len(errs); got != len(tc.wantErrs) {
+				t.Errorf("'%s': expected %d errors, got %d", tc.name, len(tc.wantErrs), len(errs))
+				return
```
xrstf: You were looking for t.Fatalf()

sdk/apis/apis/v1alpha2/types_apiexport_conversion.go
```diff
@@ -149,10 +179,37 @@ func Convert_v1alpha2_APIExportSpec_To_v1alpha1_APIExportSpec(in *APIExportSpec,
...
+
+	if err := Convert_v1alpha1_APIExportSpec_To_v1alpha2_APIExportSpec(&in.Spec, &out.Spec, s); err != nil {
 		return err
 	}
+	if err := Convert_v1alpha1_APIExportStatus_To_v1alpha2_APIExportStatus(&in.Status, &out.Status, s); err != nil {
+		return err
+	}
+
+	// store permission claims in annotation. this is necessary for a clean conversion of
+	// field All and FieldSelector, which went away in v1alpha2.
```
xrstf: Sentences have capital letters :)

sdk/apis/apis/fuzzer/fuzzer.go
embik: Not yet.

cli/pkg/bind/plugin/bind.go
embik: This is for when we add support for matching labels in the selector. Since that is going to happen in a follow-up PR, it doesn't make sense to add this functionality in the cli right now, ergo I would leave this TODO in.

sdk/apis/apis/v1alpha2/types_apibinding.go
embik: No they don't, none of the field comments in this file have capital letters, which is because we're using the json representation of the field names to generate helpful API descriptions.

sdk/apis/apis/v1alpha2/types_apibinding_test.go
embik: ACK!

sdk/apis/roundtrip_test.go
embik: Added `APIBindings` and `APIBindingList`.

embik:
/retest

Known flake: #3444

xrstf:
Here's a copy of the Slack thread linked in the PR description. This conversion happened between June 10th and 16th 2025.

<details>
<dl>
<dt>xmudrii:</dt>
<dd>One more thing that we discussed at the previous kcp community meeting was the PermissionClaims label selector behavior.

The first main problem that we have is what if someone attempts to create a new object (for a resource which has a label selector) via the appropriate APIExport Virtual Workspace but without specifying these labels. There are two options:

* Reject such request
  This is at the moment the preferred approach unless there are strong opinions against it
* Mutate the request to add the labels
  This approach has a benefit that the service provider doesn't need to know the labels needed in advance. A consumer might add a custom label selector to the permission claim if they want the resources created by the service provided to be recognized, i.e. labeled, in some way. Such label might be completely random, e.g. the app name, or some internal identifier. This is still mitigable as the service provider can look into the APIBinding itself and see the label selector. I'm not sure if we want to allow this behavior though.

However, this approach has a huge drawback because we don't have support for admission in the virtual workspaces. We would first have to create such a framework to add support for admission, then implement this, this will require significant effort for questionable gains. Because of that, we're thinking of going with rejecting requests, and we can always add mutation if turned out to be needed.

The question here is: is everyone okay with rejecting the request, i.e. are there strong opinions against?

The second main problem is the security aspect. What if the resource that's being created via the appropriate APIExport Virtual Workspace already exists in the target workspace/logical cluster but without labels specified in the label selector? In such case, we need to somehow fail. However, this is opening a security hole, as someone might use it to discover if there is a specific resource in the target logical cluster. It is important to figure out if we're accepting this, and if yes, to decide how we want to error.

And finally the third question is what to do if someone changes the label selector? This will orphan resources, so far, we have thought about just leaving orphan resources if that happens and leave it up to the user to clean up.</dd>

<dt>embik:</dt>
<dd>this is a verrrry important topic we should figure out: is a consumer allowed to set up their own label matching or do they have to adhere to what the service provider sets up in their permission claim

I think so far the assumption was that permission claim in APIExport and APIBinding have to match (?) - I would personally prefer label selectors to be a consumer choice, so they can organise providers with access to their workspace as they see fit

but that might require the whole admission and label selector mutation bits and pieces you've mentioned</dd>

<dt>xrstf:</dt>
<dd>if kcp added the labels based on the APIBinding, then the service owner would not know about them and so the service could not make use of these labels (at worst, it might interfere with a badly written service).</dd>

<dt>embik:</dt>
<dd>my take would be that the service owner doesn't have to know about them</dd>

<dt>xrstf:</dt>
<dd>so one usecase of the labels is for the service owners to calm down the user and proclaim "don't worry, i won't read all secrets, just those with this label."
  
the other usecase is the opposite, the enduser forcing/scoping a service into a small corner by forcefully applying labels to it.</dd>

<dt>embik:</dt>
<dd>I feel like that latter part is my understanding of the feature, the confinement that the enduser applies to a service provider</dd>

<dt>xrstf:</dt>
<dd>if you, the enduser, can change the labels on the binding, this means objects can appear and disappear from the view of the service owner.
  
i am not sure if any of this is good or bad.</dd>

<dt>embik:</dt>
<dd>I wonder if the label selectors should be immutable, just like they are on deployment specs :thinking:</dd>

<dt>xrstf:</dt>
<dd>by immutable you mean "must match exactly what's in the APIExport"? or truly immutable, i.e. unchangeable on both APIExport and APIBinding?</dd>

<dt>embik:</dt>
<dd>so in my mind, I'm wondering if the APIExport should even include a label selector or if it's a purely APIBinding side setting.

and yes, purely immutable

unchangeable</dd>

<dt>xrstf:</dt>
<dd>i think it's wishful thinking to expect every user to scope down every service in every workspace.

that would make the feature an "secret security feature only kcp pro's know about #likeshareandsubscribe"</dd>

<dt>embik:</dt>
<dd>I mean, you don't have to, I agree</dd>

<dt>xrstf:</dt>
<dd>but didn't we want to ensure that 2 services do not see each other by default?</dd>

<dt>xmudrii</dt>
<dd>I'm thinking about this, but I think the label choice should be up to the service provider, as they're going to create these resources and they'll have to eventually specify or reconcile labels. I'm worried about reconciliation, as even well written controllers might run into the problem that there are unexpected labels. If consumers want to risk with that, I'd say they can do that, but risk is up to them and we don't support that.

Like you want to grant access to secrets, but you don't want all secrets. The top priority is to ensure that the service provider can't access non-relevant secrets. If consumer wants to additionally label secrets created by the provider, that comes as the second priority and maybe we shouldn't handle it at all.</dd>

<dt>mjudeikis</dt>
<dd>&gt; <em>I wonder if the label selectors should be immutable, just like they are on deployment specs :thinking:</em>

If this is the case - its basically a status field (you accept these labels claims by binding the permission. No need for spec in the apibinding

&gt; <em>I'm thinking about this, but I think the label choice should be up to the service provider,</em>

And labels being service provider side enables 2 providers share resources (intentionally or not intentionally) and this feels like nice feture. I can see usecases for this already.,

I somehow leaning towards:

* its APIExport setting. It part of full resource claim on binding (take it or leave it, no deviation from what provider suggests, no partial accepts for this).
* Immutable on provider side.
* Provider must provide this with payload else - reject
* If resource exists - reject (same as no label, basically reject all around the place so its harder to tell if one exists or payload is wrong)</dd>

<dt>xrstf:</dt>
<dd>&gt; <em>If this is the case - its basically a status field (you accept these labels claims by binding the permission. No need for spec in the apibinding</em>

well... this would mean you accept "the service uses labels" once and then the service gets updated and specifies an empty label selector (i.e. the privilege to read all objects). so i think we still at least need to record what the user agreed to.</dd>

<dt>embik:</dt>
<dd>just for the record, I meant the label selector in the APIBinding to be immutable. I'm still - if technically feasible - more and more inclined to say that APIExports should not provide the label selector

but ... I might be the only one thinking that right now</dd>

<dt>xrstf:</dt>
<dd>i am not sure either, but i could see us agreeing that label selectors are a single unit (i.e. you cannot add or remove labels, you take the entire map as-is).

this however does not answer what happens if an inevitable drift between APIExport and APIBinding happens.

maybe similar to the old logic, where "the apiexport specified resources: [x,y,z], but you didn't accept exactly that ,so your binding is invalid in its entirety" is the logic i think

or do we in general promote a "you can accept just parts of an APIExport" mentality in kcp?</dd>

<dt>mjudeikis</dt>
<dd>You can already accept partial export - only 2 resources from 3. And labels would mean just subset of that. Where it ok for individual resource, it might not be ok for contract as entity: "I can create database but not its secret". So feel like its can of worms</dd>

<dt>xmudrii</dt>
<dd>First and foremost, we need to agree that label selectors are a security feature. They should exclusively be used to limit what you can see as a service provider. Any organizational outcomes (e.g. that consumers can use label selectors to organize resources in their workspaces) come as a side-effect and we should NOT officially support such outcomes. If we want to have an option for consumers to organize resources created by service providers, we should have a dedicated feature where they can provide labels and annotations to be applied on resources created via the APIExport Virtual Workspace.

Security contract (as in APIExport and APIBinding) shall at all times be a two-side contract, i.e. both sides must sign it. That is done by specifying label selectors on both sides. Needs might evolve over times, I'm sure service providers might come into the situation where they need to adjust labels. Once that happens, how are they going to do it if we make the label selector immutable? They would have to delete the APIExport and create a new one, however, how is that going to reflect to APIBindings?

We should keep one thing in mind. So far, the philosophy for the permission claims has been that the APIExport is purely a document describing what's needed. The actual set of resources that you get depends on what's in the APIBinding. So, APIExport documents what resources they need, aka what labels they need. Users, via APIBinding, chooses what they want to allow or not. Yes, this might open the can of worms, but there still legitimate use cases for that. And then again, the responsibility is up to the user, you didn't accept everything, we warn you, you take risks.</dd>

<dt>mjudeikis</dt>
<dd>More I think about this more I think - its same as exiting verbs & claims. You can change them, and if it ends-up in partial view from provider side - its providers responsibility in the way. We might need to work on signal back - X number of bindings have missmatch of the claims.
maybe its not that bit of the problem as we think it is.

basically - agree</dd>

<dt>xmudrii</dt>
<dd>I'm wondering, if we are allowing partial permission claims, should we have a property such as `required` and if it's set to `true`, you must in fully accept the permission claim to be able to bind the API at all. That way providers can be very explicit in what they needed and what's optional.</dd>

<dt>mjudeikis</dt>
<dd>I think thing like this is something we can add later. If nobody ask, dont deliver features just for sake of features :smile: (as much as engineers want to do this )</dd>

<dt>xmudrii</dt>
<dd>We (embik, xrstf, and cnvergence) had an internal discussion about the API and I'm sharing our findings.

We are agreeing that the use case for this feature is to scope down (limit) access to resources provided via the permission claims. However, we're thinking that this should be done by the consumer. In other words, the label selector should be present only on the APIBinding side. To put it in a different way, how the service provider can know what label selector is a consumer willing to give access to? Answer: they can't. Hence, it doesn't make much sense to have it on the APIExport side.

As of the service provider creating resources via the APIExport Virtual Workspace, they don't need to know the needed labels, we'll put mutation that's going to apply these needed labels.

However, one thing that I'm personally concerned about is related to connecting multiple service providers. I can see label selectors on the APIExport to be very useful for the purpose of documenting the _required_ label selectors. For example, "service provider A" creates some Secrets in the consumer workspace. The "service provider B" accesses these Secrets. How can a consumer know, when creating an APIBinding for the "service provider B", that the "service provider B" needs Secrets created by the "service Provider A"? That's why I find useful to have an option to document, in APIExport, similar to verbs, what is the minimum/required label selectors.

The counter-argument to that is that this can't always work because the "service provider B" can't know in advance (when creating the APIExport) what label selectors are going to be used by the consumer for binding the "service provider A". This can only work if the "service provider A" is using a universal label for all resources created by that provider in any workspace (e.g. something like `a.com/provider=""`).

So, we're wondering, would anyone find useful to have label selectors on the APIExport side? These would be for the documentation purposes as described above, the authority is still within the label selectors defined in the APIBinding.</dd>

<dt>mjudeikis</dt>
<dd>This is still on my list to think of. Im leaning towards different view than yours above:

1. Provider tells claims they suggest including labels.
2. Either accept it or not. No partials. So either take it or leave it.

Provider must use labels it exposes in the payload.

Why? Its simple, pragmatic and unblocks us. Everything else has so many edge cases Im afraid even to open these can of worms :confused:, like partial-views, some resources cross-providers-exposures, etc. And im most cases - users dont have any idea what they doing, hence provider driven approach</dd>

<dt>embik</dt>
<dd>The one big problem with that is if providers chose labels and not consumers, you might create conflicts (because in general, providers don't really know each other)</dd>

<dt>mjudeikis</dt>
<dd>I think uniqueness should be dictated by overall platform builder (use random UUID). But agree. I have no good solution yet in my head</dd>

<dt>embik</dt>
<dd>I do believe that only the consumer can even know how to slice and dice their workspace, so I think this should be user responsibility. We need to provide proper tools to do this right, but I think "users might be too lazy to do it" would also be an argument against RBAC :smile:</dd>

<dt>mjudeikis</dt>
<dd>I agree with your statement, just afraid if we put this into user control it will create more damage to how provider are interacting than value :confused:</dd>

<dt>xmudrii</dt>
<dd>I would just say one thing regarding RBAC. Users might be too lazy to do it has to do something with RBAC being too complicated in its design. We need to see if we can avoid it. If we make this complicated just like RBAC, guess what, consumers will just allow full access and trust the service provider. :upside_down_face:</dd>

<dt>embik</dt>
<dd>a tale as old as time, but I'm not sure we can avoid this. Security needs consideration, if people don't want to take that moment to secure things we will never be able to help them
13:14
and shifting this to the provider renders it useless as a security feature, because users that just blindly accept what the provider gives them will be caught in any bad actor's scheme anyway</dd>

<dt>mjudeikis</dt>
<dd>Devils advocate: Do we really need label selectors? Would names be not enough?</dd>

<dt>luxas</dt>
<dd>
https://github.com/kubernetes/kubernetes/issues/44703
</dd>

<dt>mjudeikis</dt>
<dd>maybe we do namespace level selectors only? You get access to namespace and that is it?</dd>

<dt>embik</dt>
<dd>label selectors make it dynamic, so to share an object with an api provider I just label it and that's it. much easier than going into the APIBinding and appending an object name to a static list</dd>

<dt>mjudeikis</dt>
<dd>I dont think we have consensus here. I tempted to go with APIBinding approach as it feel more intuitive while service provider still dictating "expected values", like `providerA: true` . Because without desired state it will get very hard to debug when somebody does not have access.

APIExport provides what it wants (expects). And any object writen need to have these labels already present

APIBinding can either commit to these or not by accepting claim or partially accepting.</dd>

<dt>embik</dt>
<dd>I feel like that would be even more confusing because then we expect providers to set a label, but that label might not even be the right one because I can choose to not follow the recommendation from the APIExport</dd>

<dt>mjudeikis</dt>
<dd>So just to outline UX:

1. Service provider wants access to `ConfigMap` with `*`
2. User accepting binding says - sure, you get `*` but only on `providerName: potatoService` labeled objects.
3. kcp mutated objects to include this label into objects
4. If user removed this label - object is orphaned and if provider tried to re-create it gets conflict?

first 2 are happy path, 2 last are not</dd>

<dt>xmudrii</dt>
<dd>Let me put it in different way -- why is that a problem? If service providers gives recommendations via the APIExport, and you choose what to use in your APIBinding, why do we care if you go with a wrong choice? I don't see much difference to verbs, you can partially accept verbs, if that breaks something for you, :shrug: I'm sorry, but we can't help you there</dd>

<dt>mjudeikis</dt>
<dd>I think problem is when provider wants `providerName: foo` but consumer says `foo` is already taken in my "world", you get `providerName: fooExternal` and we have miss-match. So it. has to be one or another.

who is right in this scenario? is claim invalid? What is final state?</dd>

<dt>embik</dt>
<dd>I also don't see a huge UX issue with your steps ... in this scenario the provider would give feedback to the user (e.g. via an event on the object that triggered creating a ConfigMap) that the target configmap name is not accessible to the provider, please fix provider permission claims or label the object appropriately, done

yes, you can mess this up if you do something like that, but I would think this is analog to removing a resourceName from a ClusterRole - it will break, but it's literally designed to do so</dd>

<dt>mjudeikis</dt>
<dd>I mean, if we allow configuring on both ends - reconciliation (solving conflict, not code) is harder. So it must be one or another (provider or consumer), not both. So if we say - APIBinding only - its fine. But not both.</dd>

<dt>xmudrii</dt>
<dd>If we're going that way, yes -- APIBinding only is much easier IMO.</dd>

<dt>embik</dt>
<dd>I agree with that, and for me it makes more sense on the consumer than the provider side</dd>

<dt>xmudrii</dt>
<dd>And the provider can always look into the APIBinding to see what labels are there, then do whatever you want, emit an event, fail, etc.</dd>

<dt>embik</dt>
<dd>and to be honest, I would prefer getting this right over implementing it in time for 0.28. So we can say "this needs admission in VWs, we don't have that yet", we add an issue to the backlog and link the two issues</dd>

<dt>mjudeikis</dt>
<dd>Ok, lets put it this way:

Assuming we go with ONLY APIBinding concept.

* Label adding is dynamic - kcp is responsible for this. I dont think we need full blown admission machinery and this can be done in VW code on serving layer. Right way is admission, but I think  long way.

What edge cases we might have in this way?</dd>

<dt>xmudrii</dt>
<dd>The biggest road blocker we'll run into is that a lot of VW code is around APIExports, not APIBindings. Well, I think I'm too light here, all of the code is involving APIExports.

So we'll have a bunch of issues trying to figure out the correct labels and APIBindings.</dd>

<dt>embik</dt>
<dd>the authorization layer already uses APIBindings, I think they should be accessible, no?</dd>

<dt>xmudrii</dt>
<dd>But this is not going to be on the authorization layer.</dd>

<dt>embik</dt>
<dd>I know, but my point is that access to APIBindings is in the general vicinity of the code

Edge cases I can think of:

1. Provider is trying to create an object under a specific name, but an object that doesn't match labels, so it's hidden from the provider. Provider fails to create object, needs to report this.
2. Consumer updates label selectors on APIBinding, making all claim objects vanish. I'm inclined to say these are immutable.
3. What kind of labels do we attach when matchExpressions is used over matchLabels? In  those cases I think the consumer would need to provide a "default label set" that gets attached to objects that match the given expressions</dd>

<dt>mjudeikis</dt>
<dd>&gt; <em>making all claim objects vanish. I'm inclined to say these are immutable.</em>

labels on APIBindings claims?

&gt; <em>matchExpressions</em>

dont support yet. solve later.</dd>

<dt>embik</dt>
<dd>&gt; <em>labels on APIBindings claims?</em>

Yes, matchLabels would be immutable.

&gt; <em>dont support yet. solve later.</em>

Fine by me.</dd>

<dt>mjudeikis</dt>
<dd>so we just need to find nice way to mutate payload now based on apibindings state and it should be relatively easy. if not - its kinda can of worms</dd>

<dt>embik</dt>
<dd>exactly</dd>

<dt>xmudrii</dt>
<dd>I don't see this being possible but I would be more than extremely happy to be proven wrong :upside_down_face: The only option that I'm seeing is the full blown admission, everything else is just... impossible, not even the can of worms.

The way authorization works, and why we have APIBinding there, is that you already have a request, and you can extract the target cluster from there and find the APIBinding based on the APIExport. Easy.

If we would tap into the storage layer or any existing controller, this is just impossible. All controllers, and the whole logic behind serving the VW, is around the fact that we use APIExport to create the storage layer (before any APIBinding is actually created), and then extract info from the request to match the correct cluster.

That said, the only way to find the APIBinding is when you have the concrete request which includes the cluster name. Only admission is doable for that IMO.

However, this actually opens the can of worms, admission solves CREATE and UPDATE/PATCH, but it doesn't solve read calls. We had that solved because we had selector on the APIExport side, now that's not the case any longer, so we need a solution for that too.

Again, tapping into the storage layer is not an option because we just don't know label selectors at the time we create the storage layer.

We'll need some kind of interceptor that's going to filter the response.</dd>
</dl>
</details>

Issue #3488, opened 2025-07-17
bug: Readiness probes not realiable in testing


xrstf:
### Describe the bug

I recently noticed that specifically the `test/e2e/authorizer` tests were failing (timing out and getting panic'ed by Go) on my laptop. If you were to put print statements around every context cancellation and t.Cleanup() call, the log would look like this:

```
=== RUN   TestAuthorizationOrder/Authorization_order_2
    utils.go:35: Starting webhook with kubernetes:authz:allow policy...
    authorizationorder_test.go:137: Saving test artifacts for test "TestAuthorizationOrder/Authorization_order_2" under "/home/xrstf/gospace/src/github.com/kcp-dev/kcp/arty/TestAuthorizationOrder/Authorization_order_2/3544293867".
I0717 14:07:56.855521 3417156 crypto.go:609] "generating new CA for cert" certName="kcp-client-ca" certFile="/tmp/TestAuthorizationOrderAuthorization_order_21792655564/003/client-ca.crt" keyFile="/tmp/TestAuthorizationOrderAuthorization_order_21792655564/003/client-ca.key"
    fixture.go:56: [Jul 17 12:07:57.020] Starting kcp servers...
    fixture.go:317: running: /home/xrstf/gospace/src/github.com/kcp-dev/kcp/bin/kcp start --root-directory /tmp/TestAuthorizationOrderAuthorization_order_21792655564/001/kcp/shared --secure-port=42593 --embedded-etcd-client-port=38555 --embedded-etcd-peer-port=44613 --embedded-etcd-wal-size-bytes=5000 --kubeconfig-path=/tmp/TestAuthorizationOrderAuthorization_order_21792655564/001/kcp/shared/admin.kubeconfig --feature-gates= --audit-log-path /home/xrstf/gospace/src/github.com/kcp-dev/kcp/arty/TestAuthorizationOrder/Authorization_order_2/3544293867/test-authorization-order_authorization_order_2/kcp/shared/kcp.audit --v=4 --bind-address=127.0.0.1 --token-auth-file /home/xrstf/gospace/src/github.com/kcp-dev/kcp/test/e2e/framework/auth-tokens.csv --feature-gates=WorkspaceMounts=true,CacheAPIs=true --audit-policy-file /tmp/TestAuthorizationOrderAuthorization_order_21792655564/002/audit-policy.yaml --client-ca-file /tmp/TestAuthorizationOrderAuthorization_order_21792655564/003/client-ca.crt
[...]

PR #3489, opened 2025-07-17
Improve authorizer webhook tests

xrstf:
## Summary
While refactoring the authorder tests in #3281 we added a small mistake: in each subtest the webhook would get started and its shutdown function registered for cleanup. But then during the test, as intended, we shutdown the webhook and restart it with some other settings. To achieve this, the test manually calls the cancelFunc of the webhook (good) and then starts another webhook (also good).

But for the 2nd instance of the webhook, we do not keep track of its cancellation. Instead when the test finishes, Go will call the cancelFunc for the *1st* webhook *again*, which is pointless at this time.

Additionally, the cancel func for the webhook does not take into account that the webhook will technically *die* because we cancel its context. So the "signal: killed" error that cmd.Wait() returns was printed in the test log, causing slight confusion in me.

If you were to run the tests with `-v`, you'd see this in the logs:

```
fixture.go:119: Waiting for readiness for server at https://127.0.0.1:42123/
metrics.go:66: PROMETHEUS_URL environment variable unset, skipping Prometheus scrape config generation
authorizationorder_test.go:125: Started kcp servers after 30.612133142s
authorizationorder_test.go:50: Admin should be allowed to list Workspaces.
utils.go:70: Stopping webhook...
utils.go:73: error waiting for webhook to finish: signal: killed
utils.go:34: Starting webhook with kubernetes:authz:deny policy...
authorizationorder_test.go:58: Admin should not be allowed to list ConfigMaps.
authorizationorder_test.go:63: Verify that it is allowed to access one of AllowAllPaths endpoints.
authorizationorder_test.go:150: Verifying access to: https://127.0.0.1:42123/healthz
fixture.go:145: Gathering metrics from kcp servers...
fixture.go:150: Gathering metrics for kcp server main
metrics.go:52: error getting metrics for server main: context canceled
fixture.go:269: cleanup: canceling context
fixture.go:273: cleanup: waiting for shutdownComplete
fixture.go:278: cleanup: received shutdownComplete
utils.go:70: Stopping webhook...
utils.go:73: error waiting for webhook to finish: exec: Wait was already called
```

To fix this, this PR makes sure the 2nd webhook also gets its cancel func called. Also, the cancelFunc is now wrapped in sync.OnceFunc(), making it generally much safer to use.

---

Additionally, these tests specifically were failing on my machine. After a lot of debugging, we found the root cause being that these tests are simply too fast. So fast that between kcp's readiness check and us killing it, not enough time passes and kcp freaks out. See #3488 for more information.

Since this weirdness only affects tests that are very, very fast, I chose to work around it by introducing an artificial delay. This delay will only affect the AuthOrder tests and so won't have a noticeable impact on the entire test suite.

## What Type of PR Is This?
/kind flake

## Release Notes
```release-note
NONE
```

### DIFF
--- a/test/e2e/authorizer/authorizationorder_test.go
+++ b/test/e2e/authorizer/authorizationorder_test.go
@@ -43,8 +43,8 @@ func TestAuthorizationOrder(t *testing.T) {
 		webhookPort := "8080"
 		ctx, cancelFunc := context.WithCancel(context.Background())
 		t.Cleanup(cancelFunc)
-		webhookStop := RunWebhook(ctx, t, webhookPort, "kubernetes:authz:allow")
-		t.Cleanup(webhookStop)
+		webhook1Stop := RunWebhook(ctx, t, webhookPort, "kubernetes:authz:allow")
+		t.Cleanup(webhook1Stop)
 
 		server, kcpClusterClient, kubeClusterClient := setupTest(t, "AlwaysAllowGroups,AlwaysAllowPaths,Webhook,RBAC", "testdata/webhook1.kubeconfig")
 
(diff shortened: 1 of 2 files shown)


test/e2e/authorizer/authorizationorder_test.go
```diff
@@ -53,7 +54,8 @@ func TestAuthorizationOrder(t *testing.T) {
 
 		// stop the webhook and switch to a deny policy
 		webhookStop()
-		RunWebhook(ctx, t, webhookPort, "kubernetes:authz:deny")
+		webhookStop = RunWebhook(ctx, t, webhookPort, "kubernetes:authz:deny")
```
ntnn: Nitpicky, but I think instead of shadowing the webhook cancel functions should get their own name, e.g. `denyWebhookStop` and `allowWebhookStop` (or `denyWebhookCancel`, ...).

It's purely cosmetic but I think it'd look better

xrstf: I had the same feeling in my gut. That decides it.

PR #3492, opened 2025-07-24
Make deletion relationship between workspace and logicalcluster one way

### LINKED ISSUE #3490: Change deletion relationship between Workspaces and Logicalclusters to one-way
As discussed in Community Meeting Jul 17 2025, we want to change the deletion relationship between Workspaces and Logicalclusters.

**Problem Statement**
If you have a LogicalCluster with custom finalizers and try to delete its workspace, you will end up in a situation where your workspace is gone, but your logicalcluster remains as a dangling object. The reason behind this is that on workspace deletion we [basically mark the corresponding logicalcluster for deletion](https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/tenancy/workspace/workspace_reconcile_deletion.go#L110-L112); This in turn triggers the logicalcluster deletion controller, [which modifies its owner workspace object by removing the last finalizer](https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/core/logicalclusterdeletion/logicalcluster_deletion_controller.go#L323-L330), thus deleting the workspace, while still leaving the logicalcluster dangling.
This problem blocks the introduction of custom finalizers (= opposite of kcp initializers).

**Solution Proposal**
Make the deletion logic between a workspace and a logicalcluster one-way, with the workspace marking the desired state. 

Specifically if a workspace is marked for deletion, the workspaceDeletionReconciler checks if any logicalcluster exists and if yes marks the logicalcluster for deletion and then re-queues the workspace deletion event. If the logicalcluster does not exist anymore, the workspace controller would remove its finalizer from the workspace.

**Alternatives considered**

- Owner-references -> As far as I can see, we would need to have owner references across different api servers, which we don’t yet have.
- Synchronized finalizers: Mirroring finalizers between Workspace and LogicalCluster objects. I think this would requir

SimonTheLeg:
## Summary

This PR changes the deletion relationship between workspaces and logicalclusters to be one way. This means that the workspace will remain as the parent object of a logicalcluster and does not get deleted until the logicalcluster object is completely done. Prior to this, it was possible for a workspace object to be completely removed, even if it still had a logicalcluster (see the issue linked below for more info).

This PR also introduces an E2E test to verify the behavior. Commits are ordered accordingly: If you run the E2E test against the old code, it will fail; with the new code it will pass.

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #3490 

## Release Notes

```release-note
NONE
```

### DIFF
--- a/test/e2e/workspace/deletion_test.go
+++ b/test/e2e/workspace/deletion_test.go
@@ -37,9 +37,9 @@ import (
 	"github.com/kcp-dev/kcp/test/e2e/framework"
 )
 
-// TestWorkspaceLogicalclusterRelationship tests that a workspace gets not
+// TestWorkspaceLogicalClusterRelationship tests that a workspace gets not
 // deleted prematurely if its LogicalCluster has not yet been deleted.
-func TestWorkspaceLogicalclusterRelationship(t *testing.T) {
+func TestWorkspaceLogicalClusterRelationship(t *testing.T) {
 	t.Parallel()
 	framework.Suite(t, "control-plane")
 
@@ -65,14 +65,16 @@ func TestWorkspaceLogicalclusterRelationship(t *testing.T) {
 
 	// add a finalizer to the cluster, mimicking an external finalizer
 	customFinalizer := "example.com/test"
-	lcsUpd := lc.DeepCopy()
-	lcsUpd.Finalizers = append(lcsUpd.Finalizers, customFinalizer)
-	_, err = clientset.Cluster(testPath).CoreV1alpha1().LogicalClusters().Update(ctx, lcsUpd, v1.UpdateOptions{})
-	require.NoError(t, err, "error updating logicalcluster")
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		lcsUpd := lc.DeepCopy()
+		lcsUpd.Finalizers = append(lcsUpd.Finalizers, customFinalizer)
+		_, err = clientset.Cluster(testPath).CoreV1alpha1().LogicalClusters().Update(ctx, lcsUpd, v1.UpdateOptions{})
+		require.NoError(c, err, "error updating logicalcluster")
 
-	// delete the workspace
-	err = clientset.Cluster(fixtureRoot).TenancyV1alpha1().Workspaces().Delete(ctx, wsName, v1.DeleteOptions{})
-	require.NoError(t, err, "error deleting workspace")
+		// delete the workspace
+		err = clientset.Cluster(fixtureRoot).TenancyV1alpha1().Workspaces().Delete(ctx, wsName, v1.DeleteOptions{})
+		require.NoError(c, err, "error deleting workspace")
+	}, wait.ForeverTestTimeout, 100*time.Millisecond, "waiting for finalizer to be applied and workspace to be deleted")
 
 	// ensure that the deletion has propagated to the logicalcluster, meaning:
 	// - it should have a deletion timestamp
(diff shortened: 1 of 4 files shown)


embik:
Minor note, we don't do emojis in the title anymore 😄

test/e2e/workspace/deletion_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	lc, err := clientset.Cluster(testPath).CoreV1alpha1().LogicalClusters().Get(ctx, corev1alpha1.LogicalClusterName, v1.GetOptions{})
+	require.NoError(t, err, "error getting logicalcluster")
+
+	// add a finalizer to the cluster, mimicking an external finalizer
+	customFinalizer := "example.com/test"
+	lcsUpd := lc.DeepCopy()
+	lcsUpd.Finalizers = append(lcsUpd.Finalizers, customFinalizer)
+	_, err = clientset.Cluster(testPath).CoreV1alpha1().LogicalClusters().Update(ctx, lcsUpd, v1.UpdateOptions{})
+	require.NoError(t, err, "error updating logicalcluster")
```
embik: This can flake, can we wrap into a `require.Eventually`? You can look at other `Update` calls in the e2e tests, they should mostly be wrapped.

test/e2e/workspace/deletion_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// delete the workspace
+	err = clientset.Cluster(fixtureRoot).TenancyV1alpha1().Workspaces().Delete(ctx, wsName, v1.DeleteOptions{})
+	require.NoError(t, err, "error deleting workspace")
```
embik: Same here, we should wrap this into `require.Eventually`.

test/e2e/workspace/deletion_test.go
```diff
@@ -0,0 +1,132 @@
...
+	corev1alpha1 "github.com/kcp-dev/kcp/sdk/apis/core/v1alpha1"
+	kcpclientset "github.com/kcp-dev/kcp/sdk/client/clientset/versioned/cluster"
+	kcptesting "github.com/kcp-dev/kcp/sdk/testing"
+	kcptestinghelpers "github.com/kcp-dev/kcp/sdk/testing/helpers"
+	"github.com/kcp-dev/kcp/test/e2e/framework"
+)
+
+// TestWorkspaceLogicalclusterRelationship tests that a workspace gets not
+// deleted prematurely if its LogicalCluster has not yet been deleted.
+func TestWorkspaceLogicalclusterRelationship(t *testing.T) {
```
embik: Nit:


```suggestion
// TestWorkspaceLogicalClusterRelationship tests that a workspace gets not
// deleted prematurely if its LogicalCluster has not yet been deleted.
func TestWorkspaceLogicalClusterRelationship(t *testing.T) {
```

test/e2e/workspace/deletion_test.go
```diff
@@ -0,0 +1,132 @@
...
+		require.Contains(c, ws.Finalizers, corev1alpha1.LogicalClusterFinalizer)
+	}, wait.ForeverTestTimeout, 100*time.Millisecond, "waiting for workspace to be marked for deletion")
+
+	// remove the custom finalizer from the logicalcluster object
+	lc, err = clientset.Cluster(testPath).CoreV1alpha1().LogicalClusters().Get(ctx, corev1alpha1.LogicalClusterName, v1.GetOptions{})
+	require.NoError(t, err, "error getting logicalcluster")
+	lcsUpd = lc.DeepCopy()
+	lcsUpd.Finalizers = removeByValue(lcsUpd.Finalizers, customFinalizer)
+	_, err = clientset.Cluster(testPath).CoreV1alpha1().LogicalClusters().Update(ctx, lcsUpd, v1.UpdateOptions{})
+	require.NoError(t, err, "error updating logicalcluster")
```
embik: Same here.

SimonTheLeg:
all comments are addressed. PR is ready for review again

test/e2e/workspace/deletion_test.go
```diff
@@ -0,0 +1,136 @@
...
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		lcsUpd := lc.DeepCopy()
+		lcsUpd.Finalizers = append(lcsUpd.Finalizers, customFinalizer)
+		_, err = clientset.Cluster(testPath).CoreV1alpha1().LogicalClusters().Update(ctx, lcsUpd, v1.UpdateOptions{})
+		require.NoError(c, err, "error updating logicalcluster")
+
+		// delete the workspace
+		err = clientset.Cluster(fixtureRoot).TenancyV1alpha1().Workspaces().Delete(ctx, wsName, v1.DeleteOptions{})
+		require.NoError(c, err, "error deleting workspace")
+	}, wait.ForeverTestTimeout, 100*time.Millisecond, "waiting for finalizer to be applied and workspace to be deleted")
```
embik: These two things should be in separate `Eventually` wraps, they are two distinct operations that can fail (Update and Delete).

Can we also (here and in other places) use `kcptestinghelpers.Eventually`? It starts logging the condition if it's not met after some time, plus it's used later in this code as well, and it would be good to use this consistently.

I think the latter part is on me, I said `require.Eventually` in the previous review. Apologies for that :/

embik (changes requested):
One further nit on the test and then we're good to go! 🚀

embik:
/retest

flake

Issue #3479, opened 2025-07-10
feature: logical cluster count metrics per shard


embik:
### Feature Description

It would be great to provide kcp admins with metrics on how many logical cluster are running on a shard.

### Proposed Solution

Provide metrics (via `/clusters/root/metrics`, most likely). The metric should probably be called `kcp_logicalcluster_count` or similar.

### Want to contribute?

- [ ] I would like to work on this issue.

cnvergence:
I will take a look :)

PR #3496, opened 2025-07-28
Add metrics for logical clusters count

cnvergence:
## Summary

Introduce `kcp_logicalcluster_count` metric, that will hold the value of currently running logical clusters in the shard.

## What Type of PR Is This?

/kind feature

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3479

## Release Notes

```release-note
Add metrics for logical clusters count
```

### DIFF
--- a/pkg/reconciler/tenancy/logicalcluster/logicalcluster_controller.go
+++ b/pkg/reconciler/tenancy/logicalcluster/logicalcluster_controller.go
@@ -269,7 +269,7 @@ func (c *Controller) process(ctx context.Context, key string) error {
 	return err
 }
 
-func (c *Controller) handleMetrics(obj any) {
+func (c *Controller) handleMetricsOnAdd(obj any) {
 	logicalCluster, ok := obj.(*corev1alpha1.LogicalCluster)
 	if !ok {
 		return


pkg/reconciler/tenancy/logicalcluster/logicalcluster_controller.go
```diff
@@ -252,3 +268,42 @@ func (c *Controller) process(ctx context.Context, key string) error {
...
 	return err
 }
+
+func (c *Controller) handleMetrics(obj any) {
+	logicalCluster, ok := obj.(*corev1alpha1.LogicalCluster)
+	if !ok {
+		return
+	}
+
+	if logicalCluster.Status.Phase == corev1alpha1.LogicalClusterPhaseReady {
```
mjudeikis: Can we add metrics for all states:
```
const (
	LogicalClusterPhaseScheduling   LogicalClusterPhaseType = "Scheduling"
	LogicalClusterPhaseInitializing LogicalClusterPhaseType = "Initializing"
	LogicalClusterPhaseReady        LogicalClusterPhaseType = "Ready"
	// LogicalClusterPhaseUnavailable phase is used to indicate that the logical cluster is unavailable to be used.
	// It will not be served via front-proxy when in this state.
	// Possible state transitions are from Ready to Unavailable and from Unavailable to Ready.
	// This should be used when we really can't serve the logical cluster content and not some
	// temporary flakes, like readiness probe failing.
	LogicalClusterPhaseUnavailable LogicalClusterPhaseType = "Unavailable"
)
```
Basically if I have bug where my logicalcluster hangs in failed or initializing state - I want to know

cnvergence: valid point, generic count can be helpful in this situation, let's keep it simpler as well

pkg/reconciler/tenancy/logicalcluster/logicalcluster_controller.go
```diff
@@ -186,11 +203,10 @@ func (c *Controller) process(ctx context.Context, key string) error {
 
 	logicalCluster, err := c.logicalClusterLister.Cluster(clusterName).Get(corev1alpha1.LogicalClusterName)
 	if err != nil {
-		if !apierrors.IsNotFound(err) {
+		if apierrors.IsNotFound(err) {
```
embik: What's going on here, why are we changing the condition?

pkg/reconciler/tenancy/logicalcluster/logicalcluster_controller.go
embik: This doesn't look implemented @cnvergence or am I missing something?

pkg/reconciler/tenancy/logicalcluster/logicalcluster_controller.go
cnvergence: good question, changed

pkg/reconciler/tenancy/logicalcluster/logicalcluster_controller.go
cnvergence: I dropped the Readiness check - we just fetch the count of all logical clusters.
Or do we want to add a logical cluster count metric for ready and other separately? 🙂

embik: I think that was the idea. So it would be a `phase` label on the metric:

```
kcp_logicalcluster_count{phase="Ready"} 5
kcp_logicalcluster_count{phase="Scheduling"} 3
kcp_logicalcluster_count{phase="Unavailable"} 0
```

cnvergence: I will consider adding the phase label and processing it during updates. Possibly, we can also split the metrics.
This will align with how, for example, kube-state-metrics processes the count of the pods

pkg/reconciler/tenancy/logicalcluster/logicalcluster_controller.go
```diff
@@ -69,26 +73,36 @@ func NewController(
...
+		countedClusters:          make(map[string]string),
 	}
 
 	_, _ = logicalClusterInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    func(obj interface{}) { c.enqueue(obj) },
-		UpdateFunc: func(obj, _ interface{}) { c.enqueue(obj) },
-		DeleteFunc: func(obj interface{}) { c.enqueue(obj) },
+		AddFunc: func(obj any) {
+			c.enqueue(obj)
+			c.handleMetrics(obj)
```
ntnn:
```suggestion
			c.handleMetricsOnAdd(obj)
```

To keep with the naming of the other methods

pkg/reconciler/tenancy/logicalcluster/logicalcluster_controller.go
```diff
@@ -252,3 +268,76 @@ func (c *Controller) process(ctx context.Context, key string) error {
 	_, err = c.kubeClusterClient.Cluster(clusterName.Path()).RbacV1().ClusterRoleBindings().Update(ctx, newBinding, metav1.UpdateOptions{})
 	return err
 }
+
+func (c *Controller) handleMetrics(obj any) {
```
ntnn:
```suggestion
func (c *Controller) handleMetricsOnAdd(obj any) {
```

As above

ntnn (changes requested):

ntnn:
/retest

integration flake, it's been getting worse =/

ntnn (approved):

PR #3506, opened 2025-08-01
Add some basic documentation for writing kcp-aware controllers

embik:
## Summary

This adds a basic page in our docs documenting how to write controllers against kcp. It's not much, but we haven't had any mention of multicluster-runtime in kcp docs yet, so this makes our recommended way of writing controllers discoverable.

## What Type of PR Is This?

/kind documentation

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/content/developers/controllers/client-go.md
+++ b/docs/content/developers/controllers/client-go.md
@@ -1,0 +1,25 @@
+# client-go
+
+kcp maintains its own `client-go` implementation at [github.com/kcp-dev/client-go](https://github.com/kcp-dev/client-go) for multi-cluster-aware Kubernetes clients, listers and informers.
+
+## Keys for Objects in Listers/Indexers
+
+When you need to get an object from a kcp-aware lister or an indexer, you can't just pass the object's name to the
+`Get()` function, like you do with a typical controller targeting Kubernetes. Projects using kcp's copy of client-go
+are using a modified key function.
+
+Here are what keys look like for an object `foo` for both cluster-scoped and namespace-scoped varieties:
+
+|Organization|Workspace|Logical Cluster|Namespace|Key|
+|-|-|-|-|-|
+|-|-|root|-|root|foo|
+|-|-|root|default|default/root|foo|
+|root|my-org|root:my-org|-|root:my-org|foo|
+|root|my-org|root:my-org|default|default/root:my-org|foo|
+|my-org|my-workspace|my-org:my-workspace|-|my-org:my-workspace|foo|
+|my-org|my-workspace|my-org:my-workspace|default|default/my-org:my-workspace|foo|
+
+## Encoding/Decoding Keys
+
+Use the `github.com/kcp-dev/apimachinery/pkg/cache` package to encode and decode keys.
+
--- a/docs/content/developers/controllers/writing-kcp-aware-controllers.md
+++ b/docs/content/developers/controllers/writing-kcp-aware-controllers.md
@@ -7,23 +7,33 @@ description: >
 
 # Writing kcp-aware Controllers
 
-## Keys for Objects in Listers/Indexers
+While kcp follows the Kubernetes Resource Model (KRM) closely, its multi-tenancy capabilities via workspaces / logical clusters
+make normal Kubernetes controllers somewhat limited in usefulness, since they generally expect a single Kubernetes API endpoint
+(which would correspond to a single logical cluster). While kcp exposes something called [virtual workspaces](../../concepts/workspaces/virtual-workspaces.md),
+their API semantics are also different from standard Kubernetes API endpoints and thus, controllers written for Kubernetes
+cannot be used against them.
 
-When you need to get an object from a kcp-aware lister or an indexer, you can't just pass the object's name to the
-`Get()` function, like you do with a typical controller targeting Kubernetes. Projects using kcp's copy of client-go
-are using a modified key function.
+!!! info
+    kcp has previously maintained a [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime) fork under [kcp-dev/controller-runtime](https://github.com/kcp-dev/controller-runtime).
+    While it likely still works, we recommend migration to multicluster-runtime (see below).
 
-Here are what keys look like for an object `foo` for both cluster-scoped and namespace-scoped varieties:
+## multicluster-runtime
 
-|Organization|Workspace|Logical Cluster|Namespace|Key|
-|-|-|-|-|-|
-|-|-|root|-|root|foo|
-|-|-|root|default|default/root|foo|
-|root|my-org|root:my-org|-|root:my-org|foo|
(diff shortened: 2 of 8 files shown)


docs/content/developers/controllers/client-go.md
```diff
@@ -0,0 +1,25 @@
...
+Here are what keys look like for an object `foo` for both cluster-scoped and namespace-scoped varieties:
+
+|Organization|Workspace|Logical Cluster|Namespace|Key|
+|-|-|-|-|-|
+|-|-|root|-|root|foo|
+|-|-|root|default|default/root|foo|
+|root|my-org|root:my-org|-|root:my-org|foo|
+|root|my-org|root:my-org|default|default/root:my-org|foo|
+|my-org|my-workspace|my-org:my-workspace|-|my-org:my-workspace|foo|
+|my-org|my-workspace|my-org:my-workspace|default|default/my-org:my-workspace|foo|
```
SimonTheLeg: q: I am not sure I fully understand this part. I thought so far that the kcp-dev/client-go was supposed to be used by directly scoping down to the logicalcluster. So something like this

```Go

myCm, err := clientset.CoreV1().ConfigMaps().Cluster(logicalcluster.NewPath("mylogicalcluster")).Namespace("myns").Get(ctx, "mycm", metav1.GetOptions{})

// or with a dynamic client

res, err := client.Cluster(logicalcluster.NewPath("mylogicalcluster")).Resource(corev1.SchemeGroupVersion.WithResource("configmaps")).Get(ctx, "mycm", metav1.GetOptions{})
```

Can you maybe share a snipped what the specialized key looks like? Or would I also in the aforementioned examples need to replace "mycm" with the specialized key?

docs/content/developers/controllers/writing-kcp-aware-controllers.md
```diff
@@ -7,23 +7,33 @@ description: >
...
-|my-org|my-workspace|my-org:my-workspace|-|my-org:my-workspace|foo|
-|my-org|my-workspace|my-org:my-workspace|default|default/my-org:my-workspace|foo|
+[multicluster-runtime](https://github.com/kubernetes-sigs/multicluster-runtime) is an effort initiated by kcp developers in the upstream
+community to provide a framework for building multi-cluster controllers. multicluster-runtime is **not** a fork of controller-runtime,
+instead it is a sort of "addon" built on top of controller-runtime's excellent Go generic support. While the project is relatively young
+and deemed experimental, we are confident in its ability to provide a great controller development experience for kcp.
 
-## Encoding/Decoding Keys
+The runtime itself is generic and needs a so-called provider that provides it with information about the fleet of clusters
+it is supposed to reconcile over. A small ecosystem of providers exists already. Check out the upstream repository for more information.
```
SimonTheLeg: I think this sentence could lead to confusion. I think readers of this doc are interested in kcp providers directly and the next paragraph is very interesting for them. I think we should not put them on the wrong trail by (potentially) indirecting them to go to multicluster provider first, scroll through the Readme, hoping they find the "Provider Ecosystem" section and then return back to this guide to finish our "Providers for kcp" section.

Instead I think it is better to leave this out, so readers get to the "Providers for kcp" section without being sidetracked.

docs/content/developers/storage-to-rest-patterns.md
```diff
@@ -1,12 +1,14 @@
-### workspaces
+# Storage to REST Patterns
 
-kcp promises to support 1 mln of workspaces. 
-A workspace is like a Kubernetes endpoint, i.e. an endpoint usual Kubernetes client tooling (client-go, controller-runtime and others) 
+## Logical Clusters
+
+kcp promises to support 1 mln of logical clusters.
```
SimonTheLeg: nit: I don't think this is the correct abbreviation. I think it's either "M" or super formal "MM". Nonetheless I think I would prefer to just write out million.

docs/content/developers/controllers/client-go.md
embik: I have no idea, this is "previous content" I mostly wanted to retain while moving it out of the document in question. I wonder if this is still accurate.

docs/content/developers/controllers/writing-kcp-aware-controllers.md
```diff
@@ -7,23 +7,33 @@ description: >
...
 
-Use the `github.com/kcp-dev/apimachinery/pkg/cache` package to encode and decode keys.
+For more information, also feel free to check out the [talk introducing multicluster-runtime at KubeCon + CloudNativeCon EU in London, 2025](https://www.youtube.com/watch?v=Tz8IcMSY7jw).
+
+### Providers for kcp
+
+kcp hosts a repository for kcp-specific provider implementations at [kcp-dev/multicluster-provider](https://github.com/kcp-dev/multicluster-provider).
+
+Which provider to choose depends on the controller that is to be written, but there is a good chance you want to write a controller for resources
+distributed via an [APIExport](../../concepts/apis/exporting-apis.md). in that case, you should chose the `apiexport` provider from the repository.
```
xrstf: Uppercase last sentence. Also "choose" instead of "chose".

docs/content/developers/controllers/client-go.md
xrstf: I think at the very least the org column should disappear, orgs are not their own concept in kcp, right?

xrstf: In general this section of the docs was very low-level. I think the main point in using our client-go is that it makes use of kcp-dev/kubernetes, which has the extended etcd handling IIRC.

We might need to add a sentence like

> This however is a low-level concept and as a developer you should rarely be concerned with it. Instead use the provided cluster-aware clientsets from kcp-dev/client-go and/or generate your own cluster-aware code using [kcp's code-generator](https://github.com/kcp-dev/code-generator).

docs/content/developers/controllers/client-go.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Encoding/Decoding Keys
+
+Use the `github.com/kcp-dev/apimachinery/pkg/cache` package to encode and decode keys.
```
xrstf: This I think is also extremely low-level. I never had to deal with encoding keys when working on or with kcp. I would vote to remove it or mark it as something no normal kcp consuming dev would need to know.

embik: This is just content that was in the page previously. I didn't want to silently drop it. I can add a warning.

docs/content/developers/controllers/client-go.md
embik: Just dropped the whole thing. I think it might have been an earlier version of client-go. I don't think this is relevant to people developing kcp controllers in any way.

docs/content/developers/controllers/writing-kcp-aware-controllers.md
embik: Changed this.

xrstf (approved):
/approve

Issue #2603, opened 2023-01-11
bug? flake? TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled


ncdc:
From https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/kcp-dev_kcp/2602/pull-ci-kcp-dev-kcp-main-e2e-shared/1613279921029779456

After all the shards are deleted, we create a workspace, which should be unschedulable, but somehow it gets scheduled and then initialized?

```
=== RUN   TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled
=== PAUSE TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled
=== CONT  TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled
    controller_test.go:170: Saving test artifacts for test "TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled" under "/logs/artifacts/TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled/1636451888".
=== CONT  TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled
    controller_test.go:170: Starting kcp servers...
    kcp.go:430: running: /go/src/github.com/kcp-dev/kcp/bin/kcp start --root-directory /tmp/TestWorkspaceControlleradd_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled3772238024/001/kcp/main --secure-port=37495 --embedded-etcd-client-port=43281 --embedded-etcd-peer-port=39257 --embedded-etcd-wal-size-bytes=5000 --kubeconfig-path=/tmp/TestWorkspaceControlleradd_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled3772238024/001/kcp/main/admin.kubeconfig --feature-gates=CustomResourceValidationExpressions=true --audit-log-path /logs/artifacts/TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled/1636451888/artifacts/kcp/main/kcp.audit
[...]

Issue #688, opened 2022-03-10
Test flake: TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled


ncdc:
From https://github.com/kcp-dev/kcp/runs/5498833067?check_suite_focus=true

```
--- FAIL: TestWorkspaceController (0.00s)
    --- FAIL: TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled (42.82s)
        util.go:84: Saving test artifacts for test "TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled" under "/tmp/e2e/TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled/3343451077".
        fixture.go:98: Starting kcp servers...
        kcp.go:168: running: /home/runner/go/bin/kcp start --root-directory /tmp/e2e/TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled/3343451077/data/kcp/main --secure-port=37237 --embedded-etcd-client-port=41013 --embedded-etcd-peer-port=36411 --embedded-etcd-wal-size-bytes=5000 --kubeconfig-path=admin.kubeconfig --v=7
        kcp.go:389: success contacting https://10.1.0.48:37237/livez
        kcp.go:389: success contacting https://10.1.0.48:37237/readyz
        fixture.go:125: Started kcp servers after 9.606150873s
        controller_test.go:76: Delete all pre-configured shards, we have to control the creation of the workspace shards in this test
        controller_test.go:80: Create a workspace without shards
        controller_test.go:87: Expect workspace to be unschedulable
        controller_test.go:89: 
            	Error Trace:	controller_test.go:89
            	            	controller_test.go:214
            	Error:      	Received unexpected error:
[...]

PR #3511, opened 2025-08-04
Rebase to kube 1.33.3

ntnn:
## Summary

Rebase to 1.33.3

Diff PRs:
- https://github.com/kcp-dev/apimachinery/pull/90
- https://github.com/kcp-dev/embeddedetcd/pull/4
- https://github.com/kcp-dev/code-generator/pull/115
- https://github.com/kcp-dev/client-go/pull/50
- https://github.com/kcp-dev/kubernetes/pull/170
- https://github.com/kcp-dev/kcp/pull/3514

Notable Changes:

- WatchList disabled by default https://github.com/kubernetes/kubernetes/pull/131359
When starting to watch resources the initial state of objects will no longer be sent, instead watchers only receive updates from when the watch was started
As a result one test had to be adjusted to remove the expectation of the initial objects: [`f3a4fee` (#3511)](https://github.com/kcp-dev/kcp/pull/3511/commits/f3a4fee666392d7be3880ec398cb2d83f7c460e7)
While this can be reenabled I think we should follow upstream with this. Clients that need it can still explicitly call list.
- The overall size of dependencies has grown by ~370M:
```bash
kcp-1.33.3 > GOMODCACHE=/tmp/user/kcp-1.33.3/mod go mod download
kcp-1.33.3 > du -hs /tmp/user/kcp-1.33.3
995M    /tmp/user/kcp-1.33.3

main > GOMODCACHE=/tmp/user/kcp-main/mod go mod download
main > du -hs /tmp/user/kcp-main
620M    /tmp/user/kcp-main
```
This causes goreleaser to fail on the free actions runners due to running out of disk space: https://github.com/kcp-dev/kcp/actions/runs/16750243000/job/47418386309?pr=3511
Letting goreleaser build sequentially alleviates the issue and keeps the same runtime: https://github.com/kcp-dev/kcp/actions/runs/16751809185/job/47423741078?pr=3511
Most likely because it can reuse the built archives of previous builds.
@embik and I discussed the possibility of moving goreleaser to a prow job in the future but for this rebase this is fine.

Some planned post-steps/discussion points:

- updating kube rebase docs taking the move to monorepo in account
- protobuf for kube-native types and CBOR for all
KCP can technically use protobuf (and has no problems doing so when the client sends protobuf [or CBOR], see e.g. https://github.com/ntnn/kcp-cbor-proto-tester/) - however having `PrefersProtobuf` enabled in kcp-dev/kubernetes errors in the apiextension server where it is trying to decode protobuf as JSON. I'm not sure why as selecting the correct decoder should have been handled by the headers.
While protobuf is less interesting as it only benefits kube native types CBOR is more interesting and our patches currently do not take CBOR in account, e.g. here: https://github.com/kcp-dev/kubernetes/pull/170/files#diff-39e2ff2eeff49f0b9c7d72720178f54e0c233226f912db61d3d08dcd551278de
 I made some tests with both and ultimately decided that it is better to leave this is a separate issue.
- gc worker lock
Kube switched to using a lock-less GC in 1.32 which was patched back in for KCP; when talking about performance problems it came up that we could just drop that patch and see if that produces any problems. Having just tagged 0.28 this might be a good opportunity to do that.
- converter
The converter has many changes and the patches for it do not apply without conflicts even though there were no changes upstream. Also three of four helper functions are identical to upstream helpers, one only has minor changes. Much of the logic in the converter is also present in the webhook_converter. This patch could probably be rewritten to be less intrusive and to merge easier for future rebases.
- applyconfiguration
In KCPs [hack/update-codegen-clients.sh](https://github.com/kcp-dev/kcp/blob/79715d6e8ff26f13fe389c4c2cbd8f04284b7bad/hack/update-codegen-clients.sh#L32-L34) is a comment about using `generate-groups.sh` once https://github.com/kubernetes/kubernetes/pull/114987 is merged. The PR has been merged a while ago, maybe this can be addressed.
- validation-gen
See https://github.com/kcp-dev/kubernetes/pull/169 and specifically the commit [kcp-dev/kubernetes@`934d467` (#170)](https://github.com/kcp-dev/kubernetes/pull/170/commits/934d4679df3a52e1cf06c3c1667fba6b47a5d119) for details - in essence k/k has added validation-gen for declarative field validation from struct tags rather than the handwritten validation. But that fails in the kcp-dev/k fork as the go.mod rewrites do not work in the fork.
From what I gather this is only intended for kube-native types.

## What Type of PR Is This?

/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Rebase to kubernetes v1.33.3
WatchList has been disabled upstream, following this Watchers will no longer receive the state of objects when starting a watch
```

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -48,7 +48,7 @@ jobs:
       with:
         distribution: goreleaser
         version: latest
-        args: release --timeout 60m
+        args: release --timeout 60m --parallelism 1 --fail-fast
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         KREW_GITHUB_TOKEN: ${{ secrets.KREW_GITHUB_TOKEN }}
@@ -59,7 +59,7 @@ jobs:
       with:
         distribution: goreleaser
         version: latest
-        args: release --timeout 60m --snapshot
+        args: release --timeout 60m --snapshot --parallelism 1 --fail-fast
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         KREW_GITHUB_TOKEN: ${{ secrets.KREW_GITHUB_TOKEN }}


ntnn:
/retest 

https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3511/pull-kcp-test-e2e-sharded/1952381142900412416
Known flake TestAPIExportBindingAuthorizer #3444

Dockerfile
```diff
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 # Build the binary
-FROM --platform=${BUILDPLATFORM} docker.io/golang:1.23.10 AS builder
+FROM --platform=${BUILDPLATFORM} docker.io/golang:1.24 AS builder
```
embik: Can we please set the full Go version to make sure we're in line with the build images? Otherwise it's hard to understand what Go version an image was built from and the builds are less reproducible too.

.github/workflows/docs-gen-and-push.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
 
       - uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # tag=v5.5.0
         with:
-          go-version: v1.23.10
+          go-version-file: go.mod
```
embik: I have similar concerns here, will this simply pick 1.24.0 because that's what is in the `go.mod`? I would prefer to keep this explicit to be honest.

embik (commented):
Mostly looking good, thanks for all the hard work! I mostly have concerns around the Go versioning happening in the GitHub workflows and the Dockerfile.

Dockerfile
ntnn: I agree that the versioning is a bit all over the place. While we have the script that checks that the minor version is equivalent everywhere we don't actually enforce the patch level.

TBH I'd rather set the minimum version in the go.mod and the patch level in a comment, e.g.

```go
go 1.24 // 1.24.5
```

And then use that in the verify-go-version script to ensure that the `go.mod`s have the same minor version and actions, prow, docker etcpp use the same patch level

embik: I would be okay with that as well.

ntnn:
/retest

flake test/e2e/reconciler/workspace TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled

That one has been flaking as well. Seems to have flaked in the past:
https://github.com/kcp-dev/kcp/issues/2603
https://github.com/kcp-dev/kcp/issues/688

https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3511/pull-kcp-test-e2e-multiple-runs/1954278760253493248

test/e2e/virtual/replication/virtualworkspace_test.go
```diff
@@ -277,17 +277,10 @@ func TestCachedResourceVirtualWorkspace(t *testing.T) {
 			}
 		}
 
-		t.Logf("Verify that the first watched event is the second cowboy with labels %v", cowboyLabels)
-		e, next := waitForEvent()
-		checkEvent(e, watch.Added, true, next, func(cowboy *wildwestv1alpha1.Cowboy) {
-			require.Equal(t, cowboyName2, cowboy.Name, "expected to receive the second cowboy")
-			require.Equal(t, cowboyLabels, cowboy.GetLabels(), "expected the cowboy to have labels defined")
-		})
```
embik, line 285: Why remove this?

ntnn: That is detailed in the commit: [`69b8a77` (#3511)](https://github.com/kcp-dev/kcp/pull/3511/commits/69b8a779a796322eb3793cfc71e1d8fa5cc53d0b)
Upstream disabled the (beta) feature gate that watchers get the initial state of watched objects:
https://github.com/kubernetes/kubernetes/pull/131359

ntnn: I also put that in the release note so users will see it in the changelog

embik: Ah, I see. I saw the notes on this but didn't connect the dots, thanks for explaining!

mjudeikis:
/lgtm
/approve
/hold
@embik this looks ok :) 
Anything else blocking this now?

embik:
Nothing blocking this from my perspective.

PR #3523, opened 2025-08-10
Fix bind cli

mjudeikis:
## What Type of PR Is This?

`kubectl kcp bind` stopped working. 

/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Fix `kubectl kcp bind` command after verbs permission claims migration
```

PR #3519, opened 2025-08-08
Enable and fix cli tests

ntnn:
## Summary

When running `kubectl create-workspace <name>` on a workspace that already exists a panic is thrown:

```bash
> go run ./cli/cmd/kubectl-create-workspace test --ignore-existing
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x18 pc=0x105bbefd8]

goroutine 1 [running]:
github.com/kcp-dev/kcp/cli/pkg/workspace/plugin.(*CreateWorkspaceOptions).Run(0x140001170e0, {0x1062e5d98, 0x10722f4e8})
        /Users/I567861/SAPDevelop/code/kcp-work/kcp/cli/pkg/workspace/plugin/create.go:184 +0x5a8
github.com/kcp-dev/kcp/cli/pkg/workspace/cmd.NewCreate.func1(0x140000d4008, {0x140001a1cc0, 0x1, 0x2})
        /Users/I567861/SAPDevelop/code/kcp-work/kcp/cli/pkg/workspace/cmd/cmd.go:228 +0x78
github.com/spf13/cobra.(*Command).execute(0x140000d4008, {0x1400019a3a0, 0x2, 0x2})
        /Users/I567861/SAPDevelop/golang/pkg/mod/github.com/spf13/cobra@v1.9.1/command.go:1015 +0x844
github.com/spf13/cobra.(*Command).ExecuteC(0x140000d4008)
        /Users/I567861/SAPDevelop/golang/pkg/mod/github.com/spf13/cobra@v1.9.1/command.go:1148 +0x384
github.com/spf13/cobra.(*Command).Execute(0x14000519f38?)
        /Users/I567861/SAPDevelop/golang/pkg/mod/github.com/spf13/cobra@v1.9.1/command.go:1071 +0x1c
main.main()
        /Users/I567861/SAPDevelop/code/kcp-work/kcp/cli/cmd/kubectl-create-workspace/main.go:33 +0x88
exit status 2
zsh: exit 1     go run ./cli/cmd/kubectl-create-workspace test --ignore-existing
```

Reason is that in c4eb73553 the type for structuredWorkspaceType was changed to a pointer.

I was wondering why that didn't come up in the tests and discovered two things:

1. The cli tests aren't even executed
2. Even if they were executed this wouldn't have come up since the `.Type` is set by the test

[...]

PR #3525, opened 2025-08-11
[WIP] Add cli e2e suite

ntnn:
## Summary

Add e2e tests for cli plugins. 

I noted in #3519 that some of the tests would be better as e2e tests and #3523 supports that I believe.

This is just a quick WIP to get some feedback and opinions. I'll flesh it out and move some tests over from `cli/pkg/..._test` unless we want another approach.

## What Type of PR Is This?

/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


test/e2e/framework/kubectl.go
```diff
@@ -20,125 +20,53 @@ import (
...
-	t.Logf("running: %s", strings.Join(cmdParts, " "))
-
-	output, err := cmd.CombinedOutput()
-	if err != nil {
-		t.Logf("kubectl output:\n%s", output)
-	}
-	require.NoError(t, err)
-
-	return output
-}
```
ntnn: I removed these as they are unused and I don't think they will be useful for the plugin tests.

ntnn:
Closing for now; it doesn't need to constantly trigger the CI.

PR #3543, opened 2025-08-19
Fix APIBinding admission mishandling v1alpha1 API version

### LINKED ISSUE #3527: bug: regression: APIBinding v1alpha1 fails to apply
### Describe the bug

I was trying to upgrade to kcp 0.28.0 and noticed that during my tests, APIBindings are not accepted anymore.

### Steps To Reproduce

1. Start a kcp on version 0.28.0.
2. Apply the `apibinding.yaml` from below.

### Expected Behaviour

The APIBinding should be accepted and stored internally as v1alpha2.

Instead I receive

```
The APIBinding "foo" is invalid:
* spec.permissionClaims[0]: Invalid value: "object": either "all" or "resourceSelector" must be set
* spec.permissionClaims[1]: Invalid value: "object": either "all" or "resourceSelector" must be set
* spec.permissionClaims[2]: Invalid value: "object": either "all" or "resourceSelector" must be set
```

### Additional Context

`apibinding.yaml`:

```yaml
apiVersion: apis.kcp.io/v1alpha1
kind: APIBinding
metadata:
  name: foo
spec:
  reference:
    export:
      path: root
      name: cache.kcp.io
  permissionClaims:
    - group: ""
      resource: "namespaces"
      all: true
      state: Accepted
    - group: ""
      resource: "secrets"
      all: true
      state: Accepted
    - group: ""
      resource: "configmaps"
      all: true
      state: Accepted
```

xmudrii:
## Summary

As of v0.28.0, it became impossible to apply v1alpha1 APIBindings. The error would look like this:

```
The APIBinding "foo" is invalid:
* spec.permissionClaims[0]: Invalid value: "object": either "all" or "resourceSelector" must be set
* spec.permissionClaims[1]: Invalid value: "object": either "all" or "resourceSelector" must be set
* spec.permissionClaims[2]: Invalid value: "object": either "all" or "resourceSelector" must be set
```

We found out that this is happening because the APIBinding admission is mishandling the v1alpha1 API version. The admission plugin would receive the v1alpha1 version of the object, but it would unpack it into the v1alpha2 struct leading to data loss. At the end, the admission plugin would write back the v1alpha2 object, causing further issues in addition to the data loss.

This PR modifies the admission plugin to be able to handle both v1alpha1 and v1alpha2 API versions. The implementation is based on #3539, i.e. I implemented an interface for APIBindings with functions needed by the admission plugin, then there are two implementations of that interface, for v1alpha1 and v1alpha2. I also added some simple unit tests to the APIBindings admission plugin to make sure we test this scenario in the CI.

Alternative considered was that we do conversion inside the APIBindings admission plugin, that would:
- Add some unnecessary load (convert v1alpha1 to v1alpha2, eventually convert old object to v1alpha2 in case of updates, then convert both back to v1alpha1)
- Introduce some more edge cases (conversion expects already valid object, however CEL validation happens in validating admission, so after mutating admission is done, which is too late in this case)

This is technically a workaround for not having internal types, which should be our goal for the future. I'll create a ticket to track that.

This will need to be cherry-picked to v0.28.

## What Type of PR Is This?

/kind bug
/kind regression

## Related Issue(s)

Fixes #3527 

## Release Notes
```release-note
Fix APIBinding admission mishandling v1alpha1 API version. This fixes the bug where it was impossible to apply v1alpha1 APIBindings
```

/assign @xrstf @embik @mjudeikis

### DIFF
--- a/pkg/admission/apibinding/apibinding_admission.go
+++ b/pkg/admission/apibinding/apibinding_admission.go
@@ -22,11 +22,9 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"reflect"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/apiserver/pkg/admission"
 	"k8s.io/apiserver/pkg/authentication/user"
@@ -93,15 +91,18 @@ func (o *apiBindingAdmission) Admit(ctx context.Context, a admission.Attributes,
 		return apierrors.NewInternalError(err)
 	}
 
-	if a.GetResource().GroupResource() == apisv1alpha1.Resource("apibindings") {
+	// Skip if the object is not an APIBinding
+	// (v1alpha2 doesn't mean the resource version, we're only taking the resource from that package)
+	if a.GetResource().GroupResource() != apisv1alpha2.Resource("apibindings") {
+		return nil
+	}
+
+	// If we're working with v1alpha1 object, check the overhanging permission claims
+	if a.GetResource().GroupVersion() == apisv1alpha1.SchemeGroupVersion {
 		ab := &apisv1alpha1.APIBinding{}
 		if err := validateOverhangingPermissionClaims(ctx, a, ab); err != nil {
 			return admission.NewForbidden(a, err)
 		}
-	}
-
-	if a.GetResource().GroupResource() != apisv1alpha2.Resource("apibindings") {
-		return nil
 	}
 
 	u, ok := a.GetObject().(*unstructured.Unstructured)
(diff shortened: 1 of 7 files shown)


pkg/admission/apibinding/apibinding_interface.go
```diff
@@ -0,0 +1,53 @@
...
+package apibinding
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+type APIBinding interface {
```
mjudeikis: Does these need to be exported? if so can we make a comment that nobody should use these outside admisision? Having same name bit confusing. Maybe `APIBindingInternal`

mjudeikis:
/hold
unhold when ready

pkg/admission/apibinding/apibinding_interface.go
xmudrii: Is it enough to make it unexported or do you prefer that we also rename it? I'm okay with both, I'll definitely make it unexported, just wondering if I should do both.

pkg/admission/apibinding/apibinding_admission.go
```diff
@@ -93,86 +91,91 @@ func (o *apiBindingAdmission) Admit(ctx context.Context, a admission.Attributes,
...
+		a.GetOperation() == admission.Update && apiBinding.Labels()[apisv1alpha1.InternalAPIBindingExportLabelKey] != oldAPIBinding.Labels()[apisv1alpha1.InternalAPIBindingExportLabelKey]:
 
 		// unified forbidden error that does not leak workspace existence
 		action := "create"
 		if a.GetOperation() == admission.Update {
 			action = "update"
 		}
 		forbidden := admission.NewForbidden(a, fmt.Errorf("unable to %s APIBinding: no permission to bind to export %s", action,
-			logicalcluster.NewPath(apiBinding.Spec.Reference.Export.Path).Join(apiBinding.Spec.Reference.Export.Name).String()))
+			logicalcluster.NewPath(apiBinding.BindingReference().ExportPath()).Join(apiBinding.BindingReference().ExportName()).String()))
```
xrstf: apiBinding.BindingReference().ExportPath() is called so often, can we just put it in a variable or is there a chance it changes between all those calls?

pkg/admission/apibinding/apibinding_interface.go
xrstf: Yeah the only reason they're exported in the CLI module is because multiple commands use the same abstraction. This one here can and should be private.

pkg/admission/apibinding/apibinding_admission.go
xmudrii: You're right, this is now fixed.

pkg/admission/apibinding/apibinding_interface.go
xmudrii: The type is now private. 🔒

xmudrii:
/cherrypick release-0.28
(this should be a straightforward cherry-pick)
_(I hope we have the cherrypick bot here)_

hack/gen-patch-defaultrestmapper.sh
```diff
@@ -57,7 +57,7 @@ touch "${DEFAULTRESTMAPPER_PATCH_FILEPATH}"
 
 cat "${REPO_ROOT}/hack/boilerplate/boilerplate.go.txt" > "${DEFAULTRESTMAPPER_PATCH_FILEPATH}"
 if [[ "$OSTYPE" == "darwin"* ]]; then
-  sed -i '' "s/YEAR/$(date +'%Y')/" "${DEFAULTRESTMAPPER_PATCH_FILEPATH}"
+  sed -i "s/YEAR/$(date +'%Y')/" "${DEFAULTRESTMAPPER_PATCH_FILEPATH}"
```
xrstf: If this is identical now, we can get rid of the if statement alltogether.

ntnn: It shouldn't, sed on Mac hasn't changed. But you can install gnu utils and priorities GNU sed instead of MacOS sed, I presume that led to this change?

Wasn't there a helper function for this?

ntnn: Well not a helper function but this one works whether sed is GNU or MacOS sed:

https://github.com/kcp-dev/kcp/blob/a54412e7cafd7863f45615d2f762d3efa7a900df/hack/bump-k8s.sh#L37-L42

xmudrii: I committed this by mistake.

> But you can install gnu utils and priorities GNU sed instead of MacOS sed, I presume that led to this change?

This is my case, I have GNU utils by default and this command doesn't work with the GNU sed. I didn't bother trying to get it fixed, I just hack it locally to use the non-MacOS command. Reverted for now, we can get back to this later.

xmudrii:
/test pull-kcp-test-integration
TestWorkspaceDeletionLeak flake (a known issue)

pkg/admission/apibinding/apibinding_interface.go
mjudeikis: Private is enough. With age of AI autocomplete this is just question when some llm picks this and goes full hallucination mode on :D thanks

PR #3546, opened 2025-08-22
Bump to kcp-1.33.1-1

### LINKED ISSUE #3545: bug: API conversion does not work for built-in types (APIBindings...)
### Describe the bug

This only affects the main branch since we merged the rebase.

Upon creating an object using an older API version, it's expected that the conversion logic will kick in to convert from the old API version to the new API version (e.g. from v1alpha1 to v1alpha2). This is not happening which is leading to various sort of issues (e.g. data loss, validation failures...).

@xrstf, @embik, and I worked on debugging this and we found out that:
- Converter factory is created here: https://github.com/kcp-dev/kcp/blob/339e173f0b33d49da1d1d58791a14524e6d11c64/pkg/server/conversion_webhook.go
- It's passed to the server here: https://github.com/kcp-dev/kcp/blob/339e173f0b33d49da1d1d58791a14524e6d11c64/pkg/server/config.go#L601-L607
- The server does not utilize this converter factory in any meaningful way though

We managed to trace down this to https://github.com/kcp-dev/kubernetes/commit/764af8d6ac67aa1dee78a91bb0d1e28dc0995fdb

It appears that this patch is missing since the rebase. We should verify that and reapply it if needed.

### Steps To Reproduce

Try to create an object with an older API version and observe if the conversion code is being called (e.g. by adding some logging or using a debugger).

### Expected Behaviour

The conversion is working properly.

### Additional Context

_No response_

ntnn:
## Summary

Update to kcp-1.33.3-1, the new rebase fixing #3545 
The rebase includes @mjudeikis PR #3540 

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #3545 

## Release Notes

```release-note
NONE
```


mjudeikis:
/lgtm
/approve 
🚢

PR #3553, opened 2025-08-27
Add CachedResourceSchemaSource for CachedResource

gman0:
## Summary

Replication VW needs to retrieve the schema when constructing APIDefinition. At the moment the schema lookup makes an assumption that the CachedResource, the APIExport and the schema are all co-located in the same shard, which is incorrect as was implemented just as a PoC.

This PR adds CachedResourceSchemaSource to CachedResource's status, and modifies the Replication VW to use that reference to retrieve the schema.

The CachedResourceSchemaSource can be either a CRDSchemaSource or an APIResourceSchemaSource. In both cases, the source schema is cloned to CachedResource's workspace (and then automatically replicated).

This serves two purposes:
* Schema is now disconnected from its original location,
  improving resiliance against provider unavailability.
* Replication VW is greatly simplified, as it does
  schema lookup in a single step.

## What Type of PR Is This?

/kind feature

## Release Notes

```release-note
NONE
```


gman0:
We don't need to bump the cache API version nor implement conversion because cache.kcp.io is not yet usable (i.e. users cannot create consumers). /cc @mjudeikis

gman0:
Hmm...

[pull-kcp-test-integration](https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3553/pull-kcp-test-integration/1960668690961666048)

```
I0827 11:47:14.533075   37584 dynamic_serving_content.go:195] "Failed to remove file watch, it may have been deleted" file="/tmp/TestServerCreateConfigMap3813238405/002/artifacts/apiserver.crt" err="fsnotify: can't remove non-existent watch: /tmp/TestServerCreateConfigMap3813238405/002/artifacts/apiserver.crt"
E0827 11:47:14.533710   37584 dynamic_serving_content.go:144] "Failed to watch cert and key file, will retry later" err="error adding watch for file /tmp/TestServerCreateConfigMap3813238405/002/artifacts/apiserver.crt: no such file or directory"
E0827 11:47:14.535074   37584 dynamic_serving_content.go:221] "Unhandled Error" err="key failed with : open /tmp/TestServerCreateConfigMap3813238405/002/artifacts/apiserver.crt: no such file or directory" logger="UnhandledError"
I0827 11:47:14.535737   37584 controller.go:116] "Shutting down controller" component="kcp" reconciler="kcp-openapiv3"
{"level":"fatal","ts":"2025-08-27T11:47:14.537255Z","caller":"etcdserver/raft.go:234","msg":"failed to save Raft hard state and entries","error":"rename /tmp/TestServerCreateConfigMap3813238405/002/artifacts/etcd-server/member/wal/1.tmp /tmp/TestServerCreateConfigMap3813238405/002/artifacts/etcd-server/member/wal/000000000000008e-000000000000023b.wal: no such file or directory","stacktrace":"go.etcd.io/etcd/server/v3/etcdserver.(*raftNode).start.func1\n\t/home/prow/go/pkg/mod/go.etcd.io/etcd/server/v3@v3.5.21/etcdserver/raft.go:234"}
```

gman0:
/hold

We need a schema provider for either an APIResourceSchema or a CRD.

gman0:
Added support for CRD schema source.

/hold cancel

pkg/reconciler/cache/cachedresources/cachedresources_controller.go
```diff
@@ -144,18 +171,125 @@ func NewController(
...
+		},
+		DeleteFunc: func(obj interface{}) {
+			c.enqueueCachedResourcesInCluster(obj.(metav1.Object), " because of APIResourceSchema update")
+		},
 	})
 
 	return c, nil
 }
 
+func objOrTombstone[T runtime.Object](obj any) T {
```
mjudeikis: This now is in the separate library/package. We should use it.

pkg/reconciler/cache/cachedresources/cachedresources_reconcile_schema_replication.go
```diff
@@ -0,0 +1,279 @@
...
+
+	gvr := schema.GroupVersionResource{
+		Group:    cachedResource.Spec.Group,
+		Version:  cachedResource.Spec.Version,
+		Resource: cachedResource.Spec.Resource,
+	}
+
+	// We need to find out if we have the source schema replicated.
+
+	// Get only the local schema. Global informer may be lagging behind
```
mjudeikis, line 84: Why do we care if its lagging? Can you explain this a bit?

gman0: Helper functions with fallback (e.g. [`NewScopedGetterWithFallback`](https://github.com/kcp-dev/kcp/blob/c4f16ce988182237509ed873ae91a36eeb6d924f/pkg/informer/fallback_lookup.go#L88)) first try local informer. If they fail with a NotFound error, they fall back to the global informer:

https://github.com/kcp-dev/kcp/blob/c4f16ce988182237509ed873ae91a36eeb6d924f/pkg/informer/fallback_lookup.go#L91-L97

In this reconciler we're checking for the existence of an APIResourceSchema, and if it's still in cache while already gone locally, we could miss a Deleted event. This is a real issue I had during debugging :D

* Replicated APIResourceSchema is deleted
* Controller notices, invokes the schema replication reconciler
* [`NewScopedGetterWithFallback`](https://github.com/kcp-dev/kcp/blob/c4f16ce988182237509ed873ae91a36eeb6d924f/pkg/informer/fallback_lookup.go#L88) reports it's still present
  * Already gone locally
  * Global informer hasn't updated yet, the schema is still there
* The reconciler does nothing because it thinks the schema is present
* We've just missed a delete event

gman0:
Seems like the test infra is not having a great day today.

```
✓  test/e2e/authorizer (4m13.629s)
✖  test/e2e/cache (16m56.517s)
{"component":"entrypoint","file":"k8s.io/test-infra/prow/entrypoint/run.go:168","func":"k8s.io/test-infra/prow/entrypoint.Options.ExecuteProcess","level":"error","msg":"Entrypoint received interrupt: terminated","severity":"error","time":"2025-08-30T15:39:37Z"}
```

/retest

pkg/reconciler/cache/cachedresources/cachedresources_controller.go
gman0: Done.

mjudeikis:
/lgtm
/approve
this is behind feature flag so we can be bit more adventures until we prove right path forward

Issue #3513, opened 2025-08-05
feature: add logical cluster pseudo groups to cross-workspace authorization checks


embik:
### Feature Description

When doing cross-workspace authorization (`SubjectAccessReviews`, most often), e.g. when checking for the `bind` verb on an `APIExport` while creating an `APIBinding`, it would be nice to be able to assign permissions on a per-workspace basis (i.e. I want to allow access not based on individuals, but where they are trying to bind my API).

As such, I would like to be able to give permissions to groups like `system:cluster:xadasdwerawf` instead of individuals.

### Proposed Solution

This check is happening e.g. in the `apibinding` admission plugin, where there is an authorizer called to check if the `bind` verb exists. We should be able to inject a group into the user info for that particular authorization check.

Given this, APIExport providers should simply be able to assign permissions to groups named after the schema of `system:cluster:xadasdweda`. 

### Want to contribute?

- [ ] I would like to work on this issue.

### Additional Context

When implementing this, make sure to consider the ramifications of such special groups. Can someone in control of the OIDC server inject those groups into JWT tokens? Should we maybe reject JWT tokens that decode to include groups with the `system:` prefix (if we don't do that already, that is)?

ntnn:
Blocking until https://github.com/kcp-dev/kcp/pull/3481 is done to make sure the cross-workspace auth will also work for oidc users.

ntnn:
Followup ticket for docs: https://github.com/kcp-dev/kcp/issues/3623

ntnn:
Followup ticket to remove `authorization.kubernetes.io/cluster-name`: https://github.com/kcp-dev/kcp/issues/3621

PR #3571, opened 2025-09-05
Filter extra keys containing `kcp.io` on per-workspace authentication

ntnn:
## Summary

Filters keys containing `kcp.io` from user extra entries.
The `kcp.io` keys should only be set by KCP when necessary. 
For the future it will still be possible to explicitly allow explicit keys.

Relevant to #3513 

## Alternatives

CEL validation: Would make later adjustments like allowing explicit keys, harder

## What Type of PR Is This?

/kind cleanup

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/authentication/extra.go
+++ b/pkg/authentication/extra.go
@@ -1,0 +1,77 @@
+/*
+Copyright 2025 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


pkg/authentication/extra.go
```diff
@@ -0,0 +1,77 @@
...
+	"k8s.io/apiserver/pkg/authentication/user"
+)
+
+// ExtraFilter is a filter that filters out extra fields that are not
+// allowed.
+type ExtraFilter struct {
+	Authenticator authenticator.Request
+
+	// AllowExtraKeys is a list of exact keys to allow through.
+	// It takes precedence over DropExtraContains.
```
ntnn:
```suggestion
	// It takes precedence over DropExtraKeyContains.
```

ntnn:

TestImpersonation flake due to require.X inside of require.Eventually

https://github.com/kcp-dev/kcp/blob/896e5e9969a57ae0e707b70535799b6b4eef4d54/test/e2e/authorizer/impersonate_test.go#L65-L72

Issue #3578, opened 2025-09-17
Remove make install from plugin installation docs


olamilekan000:
Description:
Currently, the documentation suggests two ways to install the kcp plugins:

- Local installation via `make install`
- Installation via krew

The `make install` path is unnecessary for end users and mainly useful for development purposes. To avoid confusion and streamline the experience, references to `make install ` can be removed and only the krew installation method can be documented.

PR #3579, opened 2025-09-17
remove make install refrence for plugins from docs

olamilekan000:
## Summary
```
Change removes "make install" reference for plugins from the documentation.
```

## What Type of PR Is This?
/kind documentation

## Related Issue(s)

[Fixes](https://github.com/kcp-dev/kcp/issues/3578)

## Release Notes

```release-note
NONE
```


mjudeikis:
/lgtm
/approve

I think this should not be in user facing docs Makes it nice and clean

Issue #3535, opened 2025-08-14
Update to use `testing.T.Context`


ntnn:
In a lot of places in the tests we use the construct 

```go
ctx, cancel := context.WithCancel(context.Background())`
t.Cleanup(cancel)
```

This can be omitted in go1.25 and everywhere where a context is required `t.Context()` can be used.

/kind cleanup
/kind chore
/label "good first issue"

PaulOberg1:
Hi @ntnn , I'm interested in contributing. Could I be assigned to this please?

ntnn:
That'd be great, thank you!
/assign @PaulOberg1

olamilekan000:
Hi @ntnn, I noticed that this issue was pending for a while, so I opened a PR for it.

xrstf:
Note though the different semantics. t.Context() is cancelled *before* the cleanup functions run, whereas with the custom cancel func we have full control over when we stop the context. Just blindly replacing one with the other might lead to flakes or weird test shutdown issues.

PR #3583, opened 2025-09-19
update test context.WithCancel with testing.T.Context

olamilekan000:
## Summary
```
change replaces context.WithCancel with testing.T.Context
```

## What Type of PR Is This?
/kind cleanup

## Related Issue(s)

[Fixes #
](https://github.com/kcp-dev/kcp/issues/3535)
## Release Notes

```release-note
NONE
```


sdk/testing/kcp.go
```diff
@@ -75,7 +75,7 @@ func PrivateKcpServer(t TestingT, options ...kcptestingserver.Option) kcptesting
 // `--kcp-kubeconfig` or `--use-default-kcp-server` is supplied to the test
 // runner. Otherwise a test-managed server will be started. Only tests
 // that are known to be hermetic are compatible with shared fixture.
-func SharedKcpServer(t TestingT) kcptestingserver.RunningServer {
```
ntnn: Please revert the changes for TestingT - these are here for compatibility with testing frameworks like Ginkgo.
At most we could think about implementing testing.TB; but that has been present since 1.2 and most testing frameworks still don't have feature parity with that.

olamilekan000: done.

pkg/authentication/index_test.go
```diff
@@ -39,8 +37,7 @@ func TestCrossShardWorkspaceType(t *testing.T) {
 		teamCluster = "logicalteamcluster"
 	)
 
-	ctx, cancel := context.WithCancelCause(context.Background())
-	defer cancel(errors.New("test has ended"))
+	ctx := t.Context()
```
ntnn, line 42: Please replace uses of `ctx` with `t.Context()` - cases where we used `ctx, cancel := ...` was only due to needing access to `cancel`.

olamilekan000: There's only one occurance of` context.WithCancelCause` at the moment and the reason it wasn't changed is because the cancel function is being assigned as a field in `authenticatorState`.

```sh
	c.authConfigAuthenticators[shard][mapKey] = authenticatorState{
		cancel:        cancel,
		authenticator: authn,
	}
```

should it be removed?

xrstf: The cancel func needs to remain in the authState, or else the front-proxy will not be able to stop OIDC authenticators at runtime. But this should not affect this test function at all.

I would say you changed this test function here exactly as expected.

olamilekan000: Actually, the part of the code I shared doesn't need to change. There are no longer usage of `context.WithCancelCause` for test in this branch.

mjudeikis:
lint failure looks leggit

ntnn:
Superseded by https://github.com/kcp-dev/kcp/pull/3795

PR #3585, opened 2025-09-22
Fix GenCachedObjectName in CachedResource replication reconciler

gman0:
## Summary

GenCachedObjectName generated an object name just from the "name" arg. This PR fixes that and adds a unit test.

## What Type of PR Is This?

/kind bug

## Release Notes

```release-note
NONE
```


embik (approved):
/approve

Issue #3586, opened 2025-09-22
bug: DynamicRESTMapper doesn't see updates to resource versions


gman0:
### Describe the bug

If a bound / CRD resource is updated with a new resource version, DynamicRESTMapper won't see that change and will not update its mappings.

This is because DynamicRESTMapper only watches `internal.apis.kcp.io/resource-bindings` annotation on LogicalClusters, and it contains only group-resource information, without version -- and so updates to that are never reconciled.

### Steps To Reproduce

For bound resources:

* Bind an export.
* Create a new resource version of the APIResourceSchema, mark it as served.
* Patch the export to refer to this new schema.
* Observe that the binding now uses the new version but DynamicRESTMapper is not updated (visible e.g. in logs)

For CRDs:

* Create a CRD.
* Patch it with a new resource version.
* Observe that DynamicRESTMapper is not updated.

### Expected Behaviour

DynamicRESTMapper should reconcile version changes in bound resources and CRDs.

PR #3587, opened 2025-09-22
DynamicRESTMapper now reconciles version changes in bound resources and CRDs

gman0:
## Summary

This PR adds version reconciliation for bound resources and CRDs to DynamicRESTMapper. It was a bug to ignore changes in resource versions -- see the linked issue.

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3586

## Release Notes

```release-note
NONE
```


pkg/reconciler/dynamicrestmapper/dynamicrestmapper_controller.go
```diff
@@ -204,6 +210,110 @@ func diffResourceBindingsAnn(oldAnn, newAnn apibinding.ResourceBindingsAnnotatio
...
+	}
+
+	boundResourcesAnn, err := apibinding.GetResourceBindings(lc)
+	if err != nil {
+		utilruntime.HandleError(err)
+		return
+	}
+
+	if _, hasCRD := boundResourcesAnn[gr.String()]; !hasCRD {
+		// The CRD is not "ready" yet (maybe conflicts, not established yet, etc.).
```
mjudeikis: If its not established, what gonna happen in line 216 where we check status?

test/integration/dynamicrestmapper/dynamicrestmapper_test.go
```diff
@@ -113,6 +111,39 @@ func TestDynamicRestMapper(t *testing.T) {
...
 	require.Equal(t, "KubeCluster", gvk.Kind)
+
+	t.Logf("Updating the KubeCluster kind with v1alpha2, not serving v1alpha1 anymore")
+	err = helpers.CreateResourceFromFS(ctx, dynamicClusterClient.Cluster(logicalClusterName.Path()), mapper, nil, "crd_kubecluster_updates.yaml", assets.EmbeddedResources)
+	require.NoError(t, err)
+
+	t.Logf("Waiting for CRD to be established")
+	kcptestinghelpers.Eventually(t, func() (bool, string) {
+		crd, err := extensionsClusterClient.Cluster(logicalClusterName.Path()).CustomResourceDefinitions().Get(ctx, crdName, metav1.GetOptions{})
+		require.NoError(t, err)
```
ntnn: Wasn't this something that can produce false positives in .Eventually?

```suggestion
    if err != nil {
        return false, err.Error()
    }
```

pkg/reconciler/dynamicrestmapper/dynamicrestmapper_controller.go
ntnn: Nothing should happen, or?
None of the fields is a pointer so at most `Resource` will be empty string, meaning `gr.String` would at most return `.$group` (or `` if group is empty as well).

But to be safe there could be a check that verifies that both group and resource are not empty string.

Ignore, this should be a comment to @mjudeikis's comment here: https://github.com/kcp-dev/kcp/pull/3587#discussion_r2371631664

ntnn: Nothing should happen, or?
None of the fields is a pointer so at most `Resource` will be empty string, meaning `gr.String` would at most return `.$group` (or `` if group is empty as well).

But to be safe there could be a check that verifies that both group and resource are not empty string.

test/integration/dynamicrestmapper/dynamicrestmapper_test.go
gman0: Ah right, thanks for reminding!

pkg/reconciler/dynamicrestmapper/dynamicrestmapper_controller.go
gman0: True, I've added a check for the Established condition (which includes the population of `.Status.AcceptedNames`), and the reconciler will now skip that CRD if it's `false`.

PR #3606, opened 2025-09-28
add lima integration step to doc

olamilekan000:
## Summary
```
change adds interactions on how kcp can be integrated into Lima VMs
```

## What Type of PR Is This?
/kind documentation

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


docs/content/setup/integrations.md
```diff
@@ -59,3 +59,96 @@ them to OpenFGA queries.
...
+
+    mkdir -p /var/.kcp/
+    sudo chmod 755 /var/.kcp
+    nohup kcp start --root-directory=/var/.kcp/ --bind-address=127.0.0.1 > /var/log/kcp.log 2>&1 &
+
+probes:
+- script: |
+    #!/bin/bash
+    set -eux -o pipefail
+    if ! timeout 30s bash -c "until sudo test -f /var/.kcp/admin.kubeconfig; do sleep 3; done"; then
```
mjudeikis: I think this might not be the best way to check readiness. We would be better off curling readyz endpoints (go version of this). This is due to the fact that kubeconfig is created early, and everything else might fail. 
```
res := client.RESTClient().Get().AbsPath("/readyz").Do(ctx)
		if _, err := res.Raw(); err != nil {
			unreadyComponents := unreadyComponentsFromError(err)
			if !lastSeenUnready.Equal(unreadyComponents) {
				logger.Error(err, "control plane not ready", "unreadyComponents", sets.List[string](unreadyComponents), "error", err)
				lastSeenUnready = unreadyComponents
			}
		}
```

mjudeikis (commented):
small nit on readiness checking

docs/content/setup/integrations.md
olamilekan000: Fixed

docs/content/setup/integrations.md
```diff
@@ -59,3 +59,96 @@ them to OpenFGA queries.
...
+
+    wget https://github.com/kcp-dev/kcp/releases/download/${KCP_VERSION}/kcp_${KCP_VERSION_NO_V}_linux_arm64.tar.gz
+    tar -xzf kcp_${KCP_VERSION_NO_V}_linux_arm64.tar.gz
+    mv bin/kcp /usr/local/bin/
+    chmod +x /usr/local/bin/kcp
+    rm -f kcp_${KCP_VERSION_NO_V}_linux_arm64.tar.gz
+
+    mkdir -p /var/.kcp/
+    sudo chmod 755 /var/.kcp
+    nohup kcp start --root-directory=/var/.kcp/ --bind-address=127.0.0.1 > /var/log/kcp.log 2>&1 &
```
embik: does provisioning run every time you start the VM, or would kcp not start on a subsequent VM start?

docs/content/setup/integrations.md
```diff
@@ -59,3 +59,96 @@ them to OpenFGA queries.
 
 !!! info "Third Party Solutions"
     A third-party example of such a webhook would be Platform Mesh's [rebac-authz-webhook](https://github.com/platform-mesh/rebac-authz-webhook).
+
+## Lima
+You can run kcp inside a [Lima](https://github.com/lima-vm/lima)-managed VM, which makes it portable across macOS, Linux, and Windows (via WSL2). This setup gives you a disposable kcp control plane that integrates smoothly with your host kubectl.
+
+1. Create a Lima template for kcp
```
embik: Let's not enumerate steps here, it should be clear this is sequential. So no reason to have "1.", "2.", etc.

docs/content/setup/integrations.md
```diff
@@ -59,3 +59,96 @@ them to OpenFGA queries.
...
+2. Initialize the VM
+```sh
+limactl create --name=kcp ./kcp.yaml
+```
+
+3. Start the VM
+```sh
+limactl start kcp --vm-type=qemu
+```
+Note: On macOS, Lima may default to vz (Apple Virtualization), while on Linux it defaults to qemu, and on Windows to wsl2. If you want consistency across environments, you can explicitly pass --vm-type=qemu when starting the VM.
```
embik: Let's use an [admonition](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) here:

```suggestion
!!! info
    On macOS, Lima may default to vz (Apple Virtualization), while on Linux it defaults to qemu, and on Windows to wsl2. If you want consistency across environments, you can explicitly pass --vm-type=qemu when starting the VM.
```

embik (changes requested):
Couple of notes. Please make sure to amend any changes to your existing commit / squash any work commits you make before pushing.

docs/content/setup/integrations.md
olamilekan000: no it does not run all the time. I have updated the script so kcp can be managed with systemd.

mjudeikis:
That is ok. 
Going forward, I want us to think about this integration page a bit more, as it's not quite right now.
Meaning we should collect - developer level integration (tilt, lima, etc) and ones where one uses with kcp - postress operator openFPG, etc. 

But this is not for this PR :)

docs/content/setup/integrations.md
```diff
@@ -59,3 +59,116 @@ them to OpenFGA queries.
 
 !!! info "Third Party Solutions"
     A third-party example of such a webhook would be Platform Mesh's [rebac-authz-webhook](https://github.com/platform-mesh/rebac-authz-webhook).
+
+## Lima
+You can run kcp inside a [Lima](https://github.com/lima-vm/lima)-managed VM, which makes it portable across macOS, Linux, and Windows (via WSL2). This setup gives you a disposable kcp control plane that integrates smoothly with your host kubectl.
```
mjudeikis, line 64: Can we add a minimal README on what this integration does? 

This is essentially a playground/dev environment, where one can start a single instance of kcp (kcp in the box) for testing or limited-scope use cases. And I would like to note that this is in no way intended for production usage so we dont send wrong signal for the reader.

mjudeikis (approved):
/lgtm
/approve

PR #3615, opened 2025-10-01
WIP: add finalizer virtualworkspace

SimonTheLeg:
WIP pre rebase. Do not review

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


embik:
Getting the ball rolling on the discussion items:

1. I started out writing a comment in favour of the the dot conversion, but I'm realizing this might have (security) implications since objects and workspaces can have dots in their name too. Can you please explore if it is possible to confuse dot-converted finalizers with a workspace name imitating it (e.g. you have a WorkspaceType in `root:a:b`, and someone creates `root:a.b` and tries to mess with your finalizers)?

2. I suspect that finalizers without access to the workspace are mostly useless (they only make sense for external system integration), so I would be in favour of such a proxy here as well. I'm interested in more opinions on this.

mjudeikis:
> Getting the ball rolling on the discussion items:
> 
> 1. I started out writing a comment in favour of the the dot conversion, but I'm realizing this might have (security) implications since objects and workspaces can have dots in their name too. Can you please explore if it is possible to confuse dot-converted finalizers with a workspace name imitating it (e.g. you have a WorkspaceType in `root:a:b`, and someone creates `root:a.b` and tries to mess with your finalizers)?

Im somehow tempted to say - let's use the same behaviour - patch status. Feels like the finalisers field should be protected from user modification here. And `.` does not make it easier to argue, too. Otherwise, we have 2 different behaviours for initialising and finalising, and I don't like this either. It might have been a mistake doing it this way, but we have it now and I think we're better off sticking with it. 

If we agree otherwise, I would make the same change to both virtual workspaces to have consistent behaviour. 

> 2. I suspect that finalizers without access to the workspace are mostly useless (they only make sense for external system integration), so I would be in favour of such a proxy here as well. I'm interested in more opinions on this.

+1 on having proxy. In my head consistent sysmte beahaviour is above trying to be smart :)

pkg/reconciler/core/logicalcluster/logicalcluster_reconcile_metadata.go
```diff
@@ -62,13 +64,36 @@ func (r *metaDataReconciler) reconcile(ctx context.Context, logicalCluster *core
 		}
 	}
 
+	// add finalizers from the status as hashed labels
+	finalizerKeys := sets.New[string]()
```
mjudeikis: is. the hashed thing here due to `.` in the name?

sdk/apis/core/v1alpha1/logicalcluster_types.go
```diff
@@ -52,9 +52,9 @@ const (
 	// LogicalClusterName is the name of the LogicalCluster singleton.
 	LogicalClusterName = "cluster"
 
-	// LogicalClusterFinalizer attached to the owner of the LogicalCluster resource (usually a Workspace) so that we can control
+	// LogicalClusterFinalizerName attached to the owner of the LogicalCluster resource (usually a Workspace) so that we can control
 	// deletion of LogicalCluster resources.
-	LogicalClusterFinalizer = "core.kcp.io/logicalcluster"
+	LogicalClusterFinalizerName = "core.kcp.io/logicalcluster"
```
mjudeikis, line 57: This could be separate PR and diff for this would be way way smaller.

mjudeikis: for next time. lets leave it for now.

pkg/virtual/finalizingworkspaces/builder/build.go
```diff
@@ -0,0 +1,317 @@
...
+	logicalPath string,
+	accepted bool,
+) {
+	if !strings.HasPrefix(urlPath, rootPathPrefix) {
+		return genericapirequest.Cluster{}, dynamiccontext.APIDomainKey(""), "", false
+	}
+	withoutRootPathPrefix := strings.TrimPrefix(urlPath, rootPathPrefix)
+
+	// Incoming requests to this virtual workspace will look like:
+	//  /services/finalizingworkspace/<finalizer>/clusters/<something>/apis/workload.kcp.io/v1alpha1/synctargets
```
mjudeikis: can we remove synctarget from example? :D

pkg/virtual/finalizingworkspaces/doc.go
```diff
@@ -0,0 +1,33 @@
...
+
+// Package finalizingworkspaces and its sub-packages provide the Finalizing Workspace Virtual Workspace.
+//
+// It allows for one basic functions:
+// - cross-cluster LIST + WATCH of LogicalCluster which:
+//   - are marked for deletion via a DeletionTimestamp
+//   - request finalization by a specific controller
+//
+// That is, a request for
+// GET /services/finalizingworkspaces/<finalizer>/clusters/*/apis/core.kcp.io/v1alpha1/logicalclusters
```
mjudeikis: This would change if we go proxy way? I mean we would get access to the logical cluster via similar url?

pkg/virtual/framework/forwardingregistry/wrappers.go
```diff
@@ -87,3 +88,76 @@ func WithLabelSelector(labelSelectorFrom func(ctx context.Context) labels.Requir
...
+		}
+	})
+}
+
+// ObjectFilterFunc is a function that decides whether an object should remain / be kept.
+type ObjectFilterFunc func(ctx context.Context, obj runtime.Object) bool
+
+// HasDeletionTimestamp returns whether a runtime.object has a deletion timestamp
+// wrapping the meta.Objects deletiontimestamp functionality.
+var HasDeletionTimestamp ObjectFilterFunc = func(ctx context.Context, obj runtime.Object) bool {
```
mjudeikis: need to be public?

mjudeikis (commented):
looks close :)

pkg/reconciler/core/logicalcluster/logicalcluster_reconcile_metadata.go
SimonTheLeg: Main motivation was to keep it consistent with how initializers do it (see code block directly above this one).
I think the idea behind this was:
https://github.com/kcp-dev/kcp/blob/459fa3b0e551e48abc8beb60fc4e21f2601d04e5/sdk/apis/tenancy/initialization/utils.go#L87-L89

pkg/virtual/finalizingworkspaces/builder/build.go
SimonTheLeg: Ah good old copy pasting stuff. will change it :D

pkg/virtual/finalizingworkspaces/doc.go
SimonTheLeg: Not directly. Initializers use 3 distinct virtual workspaces under the hood to handle this:

1. handles all the LIST requests (aka get `clusters/*`)
2. handles GET/PATCH requests to a single Logicalcluster (aka update `clusters/asdfasdfa`). I think this mainly exists so we can restrict which fields of a LogicalCluster a user can edit, because with 3., I think you can just do about anything you want with the objects
3. handles all other requests and proxies them straight through to kcp

But I agree that maybe this doc.go is written a bit misleading. I just copy pasted it from initializers and never looked back 😅. I'll update it to be a bit more precise once I have built virtualworkspace 3. 

Let's leave this comment open for now.

pkg/virtual/framework/forwardingregistry/wrappers.go
SimonTheLeg: Oh yeah that is a good point, there is no big value in exposing this and it is only a helper func for `WithDeletionTimestamp`

SimonTheLeg:
This is ready for review again @embik @mjudeikis

mjudeikis:
/retest

but looks somehow related?

mjudeikis:
/retest
pod got killed O_o

PR #3616, opened 2025-10-01
test/e2e/fixtures/wildwest/client: sherifves -> sheriffs

gman0:
## Summary

Fixes a typo in e2e test fixtures: it should be `sheriffs` instead of `sherifves`. It made the wildwest.dev client fail with unknown resource.

## What Type of PR Is This?

/kind cleanup

## Release Notes

```release-note
NONE
```


embik:
Where is the change in this PR coming from? It seems all generated files, so I'm confused what file you changed to fix this.

gman0:
Sorry, this was just something I `sed`'d to fix a test, but yes now I see it's actually being re-generated and I can't find an easy fix to tell client-gen to do correct plurals :/ in the meantime I made a workaround in my test so I'm closing this.

PR #3617, opened 2025-10-02
don't assume sed revision based on OS

SimonTheLeg:
## Summary

We should not assume the sed revision based on the Operating system. For example on my Mac, I am using gnu sed, which causes the script to fail. I think in this case the fix is rather straightforward, but if you ever run into this in a more complicated setting you can make use of is_gnu_sed snippet https://github.com/kcp-dev/kcp/blob/e3c2d234c1e93f7e643ae600276090aaaeb7d7eb/hack/verify-contextual-logging.sh#L40-L45

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


SimonTheLeg:
note: I have assigned assigned three of your, since you were on the original PR which introduced this.

Issue #3487, opened 2025-07-17
feature: Virtual resources aggregation


gman0:
### Feature Description

Continuation of Part II of [Provider resource reflection - Virtual Resources](https://docs.google.com/document/d/18Xh2-VFnH23bDiAHPP_JXm4qAG4QK6GjTKcpszPGApQ/) design document.

The aggregation layer allows referring to the EndpointSlice object in the APIExport VirtualResource stanza to provide data source for aggregation (where to forward certain requests in the handler chain) for workspace view. 

### Proposed Solution

We can use a caching server pattern to distribute the load on the implementation side. From the provider's perspective, we would replicate data to the cache server and serve it to consumers. The provider would have the option to bring their cache server or allow the platform to provide a dedicated cache for some providers.

APIExport example:

```yaml
apiVersion: apis.kcp.io/v1alpha2
kind: APIExport
metadata:
  name: cowboys
spec:
  identity:
    secretRef:
      name: cowboys
      namespace: kcp-system
  resources: 
  - schema:  today.cowboys.wildwest.dev
    storage: crd
 - schema: instances.compute.corp.io
   storage: virtual
   endpointSliceSelector: 
      kind: PublishedResourceEndpointSlice
      apiVersion: apis.kcp.io/v1alpha1 
      path: root:foo:instances
      identitySecretRef:
        name: secret-name-for-identity
        namespace: default 
      resourceSelector:
        name: 
        namespace: 
        labelSelector: 
           matchLabels: kcp.io/published-resource-name: premium-instance-types
  permissionClaims:
    - all: true
      group: rbac.authorization.k8s.io
      resource: clusterrolebindings
    - all: true
      group: apiextensions.k8s.io
      resource: custom resource definitions
  identityHash: supersecure
status:
  conditions:
    - lastTransitionTime: "2025-02-09T10:38:27Z"
      status: "True"
      type: IdentityValid
  identityHash: 20d38642fc137069154e84e7f450a622ac300542e5c3f3a4c403aad8db9ec6ad
```

APIBinding example:

```yaml
apiVersion: apis.kcp.io/v1alpha1
kind: APIBinding
metadata:
  name: kube-bind.io
spec:
  permissionClaims:
  - all: true
    group: rbac.authorization.k8s.io
    resource: clusterrolebindings
    state: Accepted
  - all: true
    group: apiextensions.k8s.io
    resource: customresourcedefinitions
    state: Accepted
    identityHash: supersecurehash
  - all: true
[...]

Issue #3478, opened 2025-07-10
bug: CachedObjects are populated with PartialObjectMetadata


gman0:
### Describe the bug

When the replication for CachedResources writes the objects to cache, they only contain PartialObjectMetadata instead of full objects.

### Steps To Reproduce

1. Create a ConfigMap and a CachedResource for it:
   ```yaml
   apiVersion: v1
   kind: ConfigMap
   metadata:
     name: hello-green
     labels:
         color: green
   data:
     hello: world
   ---
   apiVersion: cache.kcp.io/v1alpha1
   kind: CachedResource
   metadata:
     name: hello-cache-green
   spec:
     group: ""
     version: v1
     resource: configmaps
     labelSelector:
       matchLabels:
         color: green
   ```
2. Read the object from the cache with a cache client, observe it contains metadata.

### Expected Behaviour

The cached objects should be replicated fully.

### Additional Context

This is happening because the replication controller for CachedResources gets its data from [`DiscoveringDynamicSharedInformerFactory`](https://github.com/kcp-dev/kcp/blob/main/pkg/server/config.go#L133) which is instantiated with a metadata client.

https://github.com/kcp-dev/kcp/blob/802765fdead32f88824b23b7de075b9043137eca/pkg/server/server.go#L155-L175

Normally, this would be okay because DiscoveringDynamicSharedInformerFactory is heavy and metadata is usually sufficient for where this factory is being used at the moment. In context of replication this is however not enough and we need full objects, with a real dynamic client that doesn't strip objects.

PR #3620, opened 2025-10-03
Add virtual resources

gman0:
## Summary

This PR adds Virtual resources (VRs) feature, and modifies the Replication virtual workspace to be in line with VRs.

Unlike regular, CRD-backed resources, VRs don't reside in a workspace, and they don't consume additional etcd storage. They exist at a virtual workspace endpoint, and that endpoint is then aggregated with the rest of the GVRs in a workspace, and exposed like that to the client. This way, producers can project resources to consumers, and their changes to the objects are visible immediately (=as soon as the object is propagated to cache, and from there to consumer's shard and informers).

A virtual resource is consumable through the usual APIExport - APIBinding relationship:

```yaml
kind: APIExport
metadata:
  name: cowboys-cr
spec:
  resources:
  - name: cowboys
    group: wildwest.dev
    schema: today.cowboys.wildwest.dev
    storage:
      virtual:
        reference:
          apiGroup: cache.kcp.io
          kind: CachedResourceEndpointSlice
          name: cowboys
        identityHash: cd2eb0837c9b4c962c22d2ff8b5441b7b45805887f051d39bf133b583baf6860
---
kind: APIBinding
metadata:
  name: cowboys-from-cache
spec:
  reference:
    export:
      path: root:provider-rw:consumer-rw
      name: cowboys-cr
```

The export defines a resource with `storage.virtual`, and refers to an endpoint slice object in `reference` by API group, Kind, and name of that object. Binding that export then exposes it in the consumer's workspace.

### Binding

Binding a VR uses the same code as with CRD-storage resources: a bound CRD is created from the APIResourceSchema, and that CRD must become Established before the resource shows up in discovery and can be used. The only difference is that VRs don't set any `APIBinding.Status.BoundResources[].StorageVersions[]` -- this is the only user-visible way to distinguish VRs from other resources. It was necessary to leave StorageVersions as `nil` so that binding deletion works correctly: the `crdcleanup` controller shouldn't touch the virtual resources.

### Discovery and OpenAPI

Since VRs still use bound CRDs, group discovery and OpenAPI works out-of-the-box, without changes. Version discovery however required some work: see `pkg/server/aggregatingcrdversiondiscovery`. A new apiserver was added to handle aggregation of CRDs (system/apibinding/normal) with VRs.

This was necessary because: the apiextensions apiserver would gladly advertise all bound CRDs, even those that belong to VRs, but all APIResources would get the same verbs; this is a problem because a virtual workspaces that backs up a VR can offer a different set of verbs. AggregatingCRDVersionDiscovery lists CRDs using the `apiBindingAwareCRDLister`, and then modifies the resulting APIResourceList so that all entries have correct verbs.

How does it know which bound CRDs are exported as VRs? `apiBindingAwareCRDLister` was modified to decorate them with  `apis.kcp.io/schema-storage` annotation. This is never persisted in etcd.

Another approach would be to modify the apiextensions apiserver to handle the discovery, but I wasn't sure how to do that cleanly without making a lot of changes. So for now, order of resolving a discovery request is like so:

* MiniAggregator handles group discovery (no changes here)
* AggregatingCRDVersionDiscovery handles `/apis/<Group>/<Resource>`:
  * Only if the group is not reserved, i.e. `*.k8s.io` or `*.kubernetes.io`. This was done to also prevent hijacking discovery for `/apis/apiextensions.k8s.io/v1` which is handled by the apiextensions apiserver. I'm not completely sure this is the best way to do it.
  * The handler code is actually very similar to the one in apiextensions, only there is `verbsProvider` to get the correct verbs for the crd.
* The next delegate is the VR apiserver.
* If nothing matches, the request is delegated to apiextensions.
* If nothing matches, NotFound handler.

### Resource handling

The VR apiserver checks whether the resource is defined as VR in the export: either through digging the associated APIBinding, or through the export identity. If it is, it retrieves the endpoint URL from the endpoint slice, and proxies the request to there.

The handling VW must be able to decode this URL: `<VW endpoint>:<APIExport identity>/<Path...>`, e.g. `https://shard-1.kcp.io:7443/services/replication/<CachedResource cluster>/<CachedResource>:<APIExport identity>/apis/wildwest.dev/v1alpha1/cowboys/namespaces/default/cowboy-1`, where `<APIExport identity>` is the identity of the export owning the schema of the VR.

I'm not completely sure this is ok, but I didn't find a better way to handle local and wildcard requests. The export is the source of the schema (https://github.com/kcp-dev/kcp/pull/3553 didn't work well enough), and since a VR (a CachedResource for example) can be exported by multiple exports, there must be a way to refer to a specific one in the request.
* To match a specific APIExport against an identity, the export must define the requested GVR, and that GVR must be `storage.virtual`. Then the code considers it as a good match.
* There is a caveat where there could be an export that matches, but the schema for that resource is wrong. I have the impression this is normal and expected. Is that so?

## What's missing in this PR

The code is behind `CacheAPIs` feature flag, disabled by default. Also, at the moment the virtual workspaces server must be running as a separate process -- the tls struct expects the certs to be in `opts.Extra.ShardVirtualWorkspaceCAFile, ...`: probably easy to fix, but I haven't tested this properly yet.

* Missing validation for APIExport's `storage.virtual`
  * To be added in an admission plugin
  * Maybe also disallow `storage.virtual` if the feature flag is disabled?
* `CachedResourceEndpointSlice.Spec.CachedResource` is missing an optional `Path` in the reference
  * CachedResource objects would need an admission plugin to insert `kcp.io/path` annotation
* If provider creates an object that's replicated with CachedResource in a namespace that the consumer doesn't have, that object is accessible regardless.
  * How do we want to handle this? The current behavior is obviously undesired. One option is to simply check for the ns, and serve the objects only if the consumer has it.
* CachedResource reconciler replicates only partial metadata: https://github.com/kcp-dev/kcp/issues/3478
  * We need a new dynamic informer with full objects
* There are changes in apiexport API
  * We'll need to coordinate version bump, conversion...

## What Type of PR Is This?

/kind api-change
/kind feature

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3487

## Release Notes

```release-note
* Changes in APIExport API: resource schema storage `virtual`
* Added Virtual resources support
```


gman0:
### Testing

Prerequsites:
* `make test-run-sharded-server`
* Have binding/CRD for cowboys.v1alpha1.wildwest.dev ready

Producer:
* Create a CachedResource:
  ```yaml
  apiVersion: v1
  kind: Secret
  metadata:
    name: cowboys-cr
    namespace: default
  stringData:
    key: xxx
  ---
  apiVersion: cache.kcp.io/v1alpha1
  kind: CachedResource
  metadata:
    name: cowboys
  spec:
    group: wildwest.dev
    version: v1alpha1
    resource: cowboys
    # schema: today.cowboys.wildwest.dev
    # For easier testing...
    identity:
      secretRef:
        name: cowboys-cr
        namespace: default
  ```
* Create an export:
  ```yaml
  kind: APIExport
  metadata:
    name: cowboys-cr
  spec:
    resources:
    - name: cowboys
      group: wildwest.dev
      schema: today.cowboys.wildwest.dev
      storage:
        virtual:
          reference:
            apiGroup: cache.kcp.io
            kind: CachedResourceEndpointSlice
            name: cowboys
          identityHash: cd2eb0837c9b4c962c22d2ff8b5441b7b45805887f051d39bf133b583baf6860
  ```
* Create a Cowboy:
  ```yaml
  apiVersion: wildwest.dev/v1alpha1
  kind: Cowboy
  metadata:
    name: cowgirl
  spec:
    intent: hehe
  ```

Consumer:
* Bind the export:
  ```yaml
  kind: APIBinding
  metadata:
    name: cowboys-from-cache
  spec:
    reference:
      export:
        path: root
        name: cowboys-cr
  ```
* `kubectl api-resources`, `kubectl get cowboys`

config/crds/bootstrap.go
```diff
@@ -124,6 +125,35 @@ func CRD(fs embed.FS, gr metav1.GroupResource) (*apiextensionsv1.CustomResourceD
 	return crd, nil
 }
 
+func AllCRDs() ([]*apiextensionsv1.CustomResourceDefinition, error) {
```
mjudeikis: go docs, public function.

mjudeikis: And its not quite clear what function does? `GetAllCRDs`?

pkg/endpointslice/endpointslice.go
```diff
@@ -0,0 +1,72 @@
...
+package endpointslice
+
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func ListURLsFromUnstructured(endpointSlice unstructured.Unstructured) ([]string, error) {
```
mjudeikis, line 28: go docs.

pkg/endpointslice/endpointslice.go
```diff
@@ -0,0 +1,72 @@
...
+			return nil, fmt.Errorf("missing url in endpoint at index %d", i)
+		}
+
+		urls = append(urls, url)
+	}
+
+	return urls, nil
+}
+
+func FindOneURL(prefix string, urls []string) (string, error) {
```
mjudeikis, line 60: go docs.

pkg/indexers/apibinding.go
```diff
@@ -26,6 +26,29 @@ import (
 	apisv1alpha2 "github.com/kcp-dev/kcp/sdk/apis/apis/v1alpha2"
 )
 
+const (
+	APIBindingByIdentityAndGroupResource = "apibinding-byIdentityGroupResource"
+)
+
+func IndexAPIBindingByIdentityGroupResource(obj interface{}) ([]string, error) {
```
mjudeikis, line 35: go doc

pkg/indexers/apiexport.go
```diff
@@ -97,3 +101,44 @@ func IndexAPIExportEndpointSliceByAPIExport(obj interface{}) ([]string, error) {
 
 	return result, nil
 }
+
+func IndexAPIExportByVirtualResourceIdentities(obj interface{}) ([]string, error) {
```
mjudeikis, line 107: go docs.

pkg/indexers/apiexport.go
```diff
@@ -97,3 +101,44 @@ func IndexAPIExportEndpointSliceByAPIExport(obj interface{}) ([]string, error) {
...
+	for _, res := range apiExport.Spec.Resources {
+		if res.Storage.Virtual != nil {
+			virtualResourceIdentities.Insert(res.Storage.Virtual.IdentityHash)
+		}
+	}
+
+	return sets.List[string](virtualResourceIdentities), nil
+}
+
+func IndexAPIExportByVirtualResourceIdentitiesAndGRs(obj interface{}) ([]string, error) {
```
mjudeikis, line 126: go docs.

pkg/reconciler/cache/cachedresourceendpointslice/cachedresourceendpointslice_controller.go
```diff
@@ -70,32 +68,32 @@ func NewController(
...
-			return globalShardClusterInformer.Cluster(core.RootCluster).Lister().Get(shardName)
+		getPartition: func(clusterName logicalcluster.Name, name string) (*topologyv1alpha1.Partition, error) {
+			return partitionClusterInformer.Lister().Cluster(clusterName).Get(name)
 		},
-		getLogicalCluster: func(clusterName logicalcluster.Name) (*corev1alpha1.LogicalCluster, error) {
-			return lcClusterInformer.Cluster(clusterName).Lister().Get("cluster")
-		},
-		getAPIBinding: func(clusterName logicalcluster.Name, bindingName string) (*apisv1alpha2.APIBinding, error) {
-			return apiBindingClusterInformer.Cluster(clusterName).Lister().Get(bindingName)
+		getCachedResourceEndpointSlicesByPartition: func(key string) ([]*cachev1alpha1.CachedResourceEndpointSlice, error) {
```
mjudeikis, line 83: So we gonna support partitions too from the start?

pkg/reconciler/cache/cachedresourceendpointslice/cachedresourceendpointslice_controller.go
```diff
@@ -70,32 +68,32 @@ func NewController(
...
+			list, err := cachedResourceEndpointSliceClusterInformer.Informer().GetIndexer().ByIndex(indexCachedResourceEndpointSlicesByPartition, key)
+			if err != nil {
+				return nil, err
+			}
+			var slices []*cachev1alpha1.CachedResourceEndpointSlice
+			for _, obj := range list {
+				slice, ok := obj.(*cachev1alpha1.CachedResourceEndpointSlice)
```
mjudeikis: index already has this check. Do we need to have it here too? 
If index returned something else - we have bigger problem? no?

pkg/reconciler/cache/cachedresourceendpointslice/cachedresourceendpointslice_indexers.go
```diff
@@ -22,28 +22,47 @@ import (
...
-	return []string{key}, nil
+	return []string{}, nil
 }
 
-func cachedResourceEndpointSliceByCachedResourceAndLogicalCluster(cachedResourceRef *cachev1alpha1.CachedResourceReference, cluster logicalcluster.Name) string {
-	var key string
-	key += cachedResourceRef.Name
-	if !cluster.Empty() {
-		key += "|" + cluster.String()
+func IndexCachedResourceEndpointSliceByCachedResourceFunc(obj interface{}) ([]string, error) {
```
mjudeikis, line 57: go doc.

pkg/reconciler/cache/cachedresourceendpointsliceurls/cachedresourceendpointsliceurls_controller.go
```diff
@@ -0,0 +1,369 @@
...
+		return
+	}
+
+	logger = logging.WithObject(logger, obj)
+
+	for _, resource := range export.Spec.Resources {
+		if resource.Storage.Virtual == nil ||
+			ptr.Deref(resource.Storage.Virtual.Reference.APIGroup, "") != cachev1alpha1.SchemeGroupVersion.Group ||
+			resource.Storage.Virtual.Reference.Kind != "CachedResourceEndpointSlice" {
+			logger.V(4).Info("skipping APIBinding its referenced APIExport does not export CachedResourceEndpointSlice virtual resources")
```
mjudeikis, line 216: Does this mean we will not yet support other types of xEndpointSlices here?

pkg/reconciler/cache/cachedresourceendpointsliceurls/cachedresourceendpointsliceurls_controller.go
```diff
@@ -0,0 +1,369 @@
...
+
+func (c *controller) enqueueCachedResourceEndpointSlice(obj *cachev1alpha1.CachedResourceEndpointSlice, logger logr.Logger, logSuffix string) {
+	logger = logging.WithObject(logger, obj)
+	key, err := kcpcache.DeletionHandlingMetaClusterNamespaceKeyFunc(obj)
+	if err != nil {
+		utilruntime.HandleError(err)
+		return
+	}
+
+	logger.Info("queueing CachedResourceEndpointSlice", "key", key) // V4
```
mjudeikis: V4

pkg/reconciler/dynamicrestmapper/dynamicrestmapper_controller.go
```diff
@@ -70,6 +72,26 @@ func init() {
...
+	kcpCRDs, err := bootstrapcrds.AllCRDs()
+	if err != nil {
+		panic(err)
+	}
+	for _, crd := range kcpCRDs {
+		for _, version := range crd.Spec.Versions {
+			if !version.Served {
+				continue
+			}
+			builtinGVKRs = append(builtinGVKRs, newTypeMeta(
```
mjudeikis: Whad does this do? Why do we need it?

pkg/server/aggregatingcrdversiondiscovery/server.go
```diff
@@ -0,0 +1,293 @@
...
+		getAPIExportsByVirtualResourceIdentity: func(vrIdentity string) ([]*apisv1alpha2.APIExport, error) {
+			return indexers.ByIndexWithFallback[*apisv1alpha2.APIExport](
+				c.Extra.LocalAPIExportInformer.Informer().GetIndexer(),
+				c.Extra.GlobalAPIExportInformer.Informer().GetIndexer(),
+				indexers.APIExportByVirtualResourceIdentities,
+				vrIdentity,
+			)
+		},
+
+		knownVirtualResourceVerbs: map[string][]string{
```
mjudeikis, line 110: Feel like this need code comment why we doing this verbProvider thingy. I know we talked about this but other readers might not be aware of reasons.

pkg/server/aggregatingcrdversiondiscovery/server.go
```diff
@@ -0,0 +1,293 @@
...
+			"CachedResourceEndpointSlice.cache.kcp.io": {"get", "list", "patch"},
+		},
+		knownVirtualResourceStatusVerbs: map[string][]string{
+			"CachedResourceEndpointSlice.cache.kcp.io": {"get"},
+		},
+		knownVirtualResourceScaleVerbs: map[string][]string{
```
mjudeikis, line 116: and wondering, if we should ahve "catch all", where if somebody adds customer sub-resource we give `get` automatically? if this even possilbe

pkg/server/virtualresources/server.go
```diff
@@ -0,0 +1,399 @@
...
+	}
+
+	vrHandler.ServeHTTP(w, r)
+}
+
+func (s *Server) getVirtualResourceURL(ctx context.Context, apiExportCluster logicalcluster.Name, virtual *apisv1alpha2.ResourceSchemaStorageVirtual) (string, error) {
+	sliceMapping, err := s.drm.ForCluster(apiExportCluster).RESTMapping(schema.GroupKind{
+		Group: ptr.Deref(virtual.Reference.APIGroup, ""),
+		Kind:  virtual.Reference.Kind,
+	}, "v1alpha1") // HACK: we need to be able to discover the latest version.
```
mjudeikis: what is this?

pkg/server/virtualresources/server.go
```diff
@@ -0,0 +1,399 @@
...
+	handler.Transport = &http.Transport{
+		TLSClientConfig: vwTLSConfig,
+	}
+
+	return handler, nil
+}
+
+func virtualResourceURLWithCluster(vwURL, apiExportIdentity string, clusterNameOrWildcard string) string {
+	// Formats the URL like so:
+	//     <Virtual resource VW endpoint>:<APIExport identity>/clusters/<Target cluster>
```
mjudeikis, line 391: Can you add full example how this would look like?

sdk/apis/cache/v1alpha1/types_cachedresourceendpointslice.go
```diff
@@ -53,16 +55,32 @@ type CachedResourceEndpointSliceSpec struct {
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="CachedResource reference must not be changed"
 	CachedResource CachedResourceReference `json:"cachedResource"`
+
+	// partition (optional) points to a partition that is used for filtering the endpoints
```
mjudeikis:
```suggestion
	// partition points to a partition that is used for filtering the endpoints
```

mjudeikis:
I pushed 1 commit to you branch (my miskate but lets keep it), where I started wiring in the example for VirtualMachines.
https://github.com/kcp-dev/kcp/pull/3620/commits/16c0acc40b3a18dba141430606d67ad17097fd0e

But having hard time to make it work :/ 
```
~/go/src/github.com/kcp-dev/kcp virtual-resources ❯ kubectl get instances.machines.svm.io                                                                                                                   11:32:19
Error from server (InternalError): Internal error occurred: error resolving resource: no matches for kind "CachedResourceEndpointSlice" in version "cache.kcp.io/v1alpha1"
~/go/src/github.com/kcp-dev/kcp virtual-resources ❯                                     
```

test/e2e/virtualresources/cachedresources/vr_cachedresources_test.go
```diff
@@ -0,0 +1,577 @@
...
+	kcpclientset "github.com/kcp-dev/kcp/sdk/client/clientset/versioned/cluster"
+	kcptesting "github.com/kcp-dev/kcp/sdk/testing"
+	kcptestinghelpers "github.com/kcp-dev/kcp/sdk/testing/helpers"
+	"github.com/kcp-dev/kcp/test/e2e/fixtures/wildwest"
+	wildwestv1alpha1 "github.com/kcp-dev/kcp/test/e2e/fixtures/wildwest/apis/wildwest/v1alpha1"
+	wildwestclientset "github.com/kcp-dev/kcp/test/e2e/fixtures/wildwest/client/clientset/versioned/cluster"
+	"github.com/kcp-dev/kcp/test/e2e/framework"
+)
+
+const XXX_timeout = math.MaxInt64
```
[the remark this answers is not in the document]
gman0: This is a leftover from when I was testing the test. I'll get rid of it.

config/crds/bootstrap.go
gman0: This was a hacky way to make DynamicRESTMapper resolve resources coming from system CRDs (e.g. the CachedResourceEndpointSlice). It's superseded by https://github.com/kcp-dev/kcp/pull/3630 and once it gets in, this change (+ all other rest mapper changes in this PR) won't be necessary anymore.

pkg/reconciler/cache/cachedresourceendpointsliceurls/cachedresourceendpointsliceurls_controller.go
gman0: We do, but this is a CachedResourceEndpointSlice controller, and so it selects only exported resources with an object reference for this resource.

pkg/reconciler/dynamicrestmapper/dynamicrestmapper_controller.go
gman0: See the comment in https://github.com/kcp-dev/kcp/pull/3620#discussion_r2413023626 -- this is dropped in favor of https://github.com/kcp-dev/kcp/pull/3630.

pkg/server/virtualresources/server.go
gman0: See the comment in https://github.com/kcp-dev/kcp/pull/3620#discussion_r2413023626 -- this is to be dropped, because https://github.com/kcp-dev/kcp/pull/3630 adds support for version defaulting.

pkg/reconciler/cache/cachedresourceendpointslice/cachedresourceendpointslice_controller.go
gman0: Just to be on par with APIExportEndpointSlices. Does this have some implications that should be taken in consideration?

pkg/server/aggregatingcrdversiondiscovery/server.go
gman0: We could do real discovery, and see what verbs are supported by the VW.

gman0: Maybe we leave it for later? We either hard-code `get` as you suggested, or do discovery, but in either case it's a functionality that has no use-case for now. It's fairly easy to implement it, but it would be pretty complex to write a test for it. The whole virtual resources codebase is ready for custom `*EndpointSlice`, but I think we need to iron out specifics of how we want users to add their own virtual resources / backing virtual workspaces etc.

pkg/endpointslice/endpointslice.go
gman0: Added in https://github.com/gman0/kcp/blob/154927d3b56d7283fb6cefd1f8cb05001d891b83/pkg/endpointslice/endpointslice.go#L26-L28.

pkg/endpointslice/endpointslice.go
gman0: Added in https://github.com/gman0/kcp/blob/154927d3b56d7283fb6cefd1f8cb05001d891b83/pkg/endpointslice/endpointslice.go#L58-L60

pkg/indexers/apibinding.go
gman0: Added in https://github.com/gman0/kcp/blob/154927d3b56d7283fb6cefd1f8cb05001d891b83/pkg/indexers/apibinding.go#L33-L35

pkg/indexers/apiexport.go
gman0: Added https://github.com/gman0/kcp/blob/154927d3b56d7283fb6cefd1f8cb05001d891b83/pkg/indexers/apiexport.go#L105-L107

pkg/indexers/apiexport.go
gman0: Added https://github.com/gman0/kcp/blob/154927d3b56d7283fb6cefd1f8cb05001d891b83/pkg/indexers/apiexport.go#L124-L126

pkg/reconciler/cache/cachedresourceendpointslice/cachedresourceendpointslice_controller.go
gman0: Cleaned up https://github.com/gman0/kcp/blob/154927d3b56d7283fb6cefd1f8cb05001d891b83/pkg/reconciler/cache/cachedresourceendpointslice/cachedresourceendpointslice_controller.go#L88-L91

pkg/reconciler/cache/cachedresourceendpointslice/cachedresourceendpointslice_indexers.go
gman0: Added https://github.com/gman0/kcp/blob/154927d3b56d7283fb6cefd1f8cb05001d891b83/pkg/reconciler/cache/cachedresourceendpointslice/cachedresourceendpointslice_indexers.go#L55-L57

pkg/reconciler/cache/cachedresourceendpointsliceurls/cachedresourceendpointsliceurls_controller.go
gman0: Oops, this was there for debugging and I specifically added a comment to not forget to remove it :D https://github.com/gman0/kcp/blob/154927d3b56d7283fb6cefd1f8cb05001d891b83/pkg/reconciler/cache/cachedresourceendpointsliceurls/cachedresourceendpointsliceurls_controller.go#L234

pkg/server/aggregatingcrdversiondiscovery/server.go
gman0: Added a few comments:

* https://github.com/gman0/kcp/blob/154927d3b56d7283fb6cefd1f8cb05001d891b83/pkg/server/aggregatingcrdversiondiscovery/server.go#L107-L109
* https://github.com/gman0/kcp/blob/154927d3b56d7283fb6cefd1f8cb05001d891b83/pkg/server/aggregatingcrdversiondiscovery/server.go#L66-L72

pkg/server/virtualresources/server.go
gman0: Added https://github.com/gman0/kcp/blob/154927d3b56d7283fb6cefd1f8cb05001d891b83/pkg/server/virtualresources/server.go#L395-L401

gman0:
/retest

Flaky goroutine leak detector?

pkg/virtual/replication/authorizer/content.go
```diff
@@ -0,0 +1,224 @@
...
+		export, err := a.getAPIExportByPath(path, binding.Spec.Reference.Export.Name)
+		if err != nil {
+			return authorizer.DecisionNoOpinion, "APIExport not found", err
+		}
+		exports = append(exports, export)
+	}
+
+	// Make sure the user has apiexports/content permissions to the exports that refer to this CachedResource resource.
+
+	SARAttributes := authorizer.AttributesRecord{
```
gman0, line 166: As discussed, Replication VW needs consumers (bound resources through APIBindings) to expose the resources. Because of that we've decided to stick with what's done in APIExport VW: the request is checked for perms to `apiexports/content` and allowed if all is well.

* See https://github.com/kcp-dev/kcp/blob/main/pkg/virtual/apiexport/authorizer/content.go#L67-L86
* And then https://github.com/kcp-dev/kcp/blob/main/pkg/virtual/apiexport/authorizer/maximal_permission_policy.go#L86-L91

pkg/virtual/replication/builder/unwrap.go
```diff
@@ -47,16 +52,48 @@ import (
...
+}
+
 func unwrapCachedObject(obj *cachev1alpha1.CachedObject) (*unstructured.Unstructured, error) {
 	inner := &unstructured.Unstructured{}
 	if err := inner.UnmarshalJSON(obj.Spec.Raw.Raw); err != nil {
 		return nil, fmt.Errorf("failed to decode inner object: %w", err)
 	}
-	inner.SetResourceVersion(obj.GetResourceVersion())
+
+	// Set the original UID and RV.
```
gman0, line 71: This is to be in line with https://github.com/kcp-dev/kcp/pull/3648

pkg/endpointslice/endpointslice.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// ListURLsFromUnstructured retrieves list of endpoint URLs from an unstructured object.
+// The URLs are expected to be present at `.status.endpoints[].url` path inside the object.
+func ListURLsFromUnstructured(endpointSlice unstructured.Unstructured) ([]string, error) {
```
mjudeikis, line 28: Can we generate test for this? Should be quick but I want to make sure we dont nil pointer exception here

pkg/server/aggregatingcrdversiondiscovery/server.go
mjudeikis: Follow-up issue?

mjudeikis (approved):
/lgtm
/approve

Amazing work :)

So few follow-up we discussed:
1. Invalid identity validation for XEndpointSlice
2. Public docs with example (we can ues example I added as base). But needs expanding why we need schema, etc. AI should help with this as skeleton is already there,
3. Refactor informer to be full (not metadatada)

I think these are ones we need now.

gman0:
/retest

CI infra is having a bad day every day lately :'(

pkg/server/aggregatingcrdversiondiscovery/server.go
gman0: https://github.com/kcp-dev/kcp/issues/3660 ^^

PR #3622, opened 2025-10-06
bump to latest Kubernetes fork commit

xrstf:
## Summary
This bumps our Kube dependency to our latest commit in the `kcp-1.33.3-1` branch, where we just merged https://github.com/kcp-dev/kubernetes/pull/180. This PR is required to fix #3614.

## What Type of PR Is This?
/kind cleanup

## Release Notes
```release-note
NONE
```

PR #3624, opened 2025-10-06
Fix TokenReviews when using WorkspaceAuthentication

### LINKED ISSUE #3614: bug: TokenReviews not working with WorkspaceAuthentication
### Describe the bug

The TokenReview handler in the k/k code has no knowledge about the cluster, and so the workspace authenticator cannot determine the authn. This makes it so that a TokenReview is always resulting in a HTTP 401 "invalid bearer token".

We should adjust our k/k fork to ensure the target cluster is available in the fakeRequest that the TokenReview handler creates and hands over to the tokenAuthenticator.

### Steps To Reproduce

N/A

### Expected Behaviour

TokenReviews should work as expected.

### Additional Context

_No response_

xrstf:
## Summary

The original design for per-workspace-auth was in the front-proxy only, only later did I add it to each shard, primarily to make writing e2e tests easier.

In the front-proxy, there is a chain of middleware handlers that

* resolve the cluster path
* find the authenticator for a workspace
* inject the authenticator into the request context
* the handle the "optional auth", where the authenticator is read from the context again

This means the actual authenticator used (i.e. in genericConfig.Authentication.Authenticator) is one that simply reads the authenticator from step 3 (out of the context) and calls it.

In the shard variant, this concept was kept the same, i.e. there are handlers that are involved to make the per-workspace-authenticator work. This however means that any other component in kcp that uses the .Authenticator has one that has no concept of per-workspace auth anymore, since the middlewares are not being called.

Thankfully the concept of middlewares is not required to make per-workspace-auth working on the shard level, so this commit condenses the logic on the shards to instead provide a "standalone" authenticator that does all steps listed above in one big function. This required the Kube patch introduced via #3622 (https://github.com/kcp-dev/kubernetes/pull/180)

Required for this refactoring is untangling the previous LocalProxy initialisation routine. Now the cluster index is started separatedly (since the new standalone auther needs it) and then handed into the middleware.

## What Type of PR Is This?
/kind bug

## Related Issue(s)
Fixes #3614

## Release Notes
```release-note
Fix TokenReviews when using WorkspaceAuthentication
```

### DIFF
--- a/test/e2e/authentication/workspace_test.go
+++ b/test/e2e/authentication/workspace_test.go
@@ -567,3 +567,76 @@ func TestAcceptableWorkspaceAuthenticationConfigurations(t *testing.T) {
 		})
 	}
 }
+
+func TestWorkspaceOIDCTokenReview(t *testing.T) {
+	framework.Suite(t, "control-plane")
+
+	ctx := context.Background()
+
+	// start kcp and setup clients
+	server := kcptesting.SharedKcpServer(t)
+
+	if len(server.ShardNames()) > 1 {
+		t.Skip("This feature currently does not support multi shards because AuthConfigs are not replicated yet.")
+	}
+
+	baseWsPath, _ := kcptesting.NewWorkspaceFixture(t, server, logicalcluster.NewPath("root"), kcptesting.WithNamePrefix("workspace-auth-token-review"))
+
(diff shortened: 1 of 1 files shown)


xrstf:
what is going on

/retest

embik (approved):

PR #3627, opened 2025-10-07
Add docker build image for local dev.

mjudeikis:
## Summary

Originally, I modified the same script to have 2 builders, but it just grew too big. 
Adding a second script for those who don't use `buildah` but still want to build images locally. 

## What Type of PR Is This?
/kind cleanup

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/hack/build-image-docker.sh
+++ b/hack/build-image-docker.sh
@@ -1,0 +1,219 @@
+#!/usr/bin/env bash
+
+# Copyright 2025 The KCP Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Build container images for KCP using Docker
+#
+# This script builds container images using Docker (with or without buildx).
+#
+# Usage examples:
+#   # Build locally with default settings (uses current git commit hash)
+#   ./hack/build-image-docker.sh
+#   
+#   # Build locally with custom repository name
+#   REPOSITORY=my-registry/kcp ./hack/build-image-docker.sh
+#   
+#   # Build locally without pushing (dry run)
+#   DRY_RUN=1 ./hack/build-image-docker.sh
+#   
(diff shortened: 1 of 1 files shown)


hack/build-image-docker.sh
```diff
@@ -0,0 +1,215 @@
...
+#   - docker + buildx: Multi-arch support with intelligent platform handling  
+#   - docker only: Single architecture fallback
+
+set -euo pipefail
+
+# make git available
+if ! [ -x "$(command -v git)" ]; then
+  echo "Installing git ..."
+  yum install -y git
+fi
```
embik, line 50: This was specific to the buildah image, so maybe we should drop it from this script?

hack/build-image-docker.sh
```diff
@@ -0,0 +1,215 @@
...
+if docker buildx version >/dev/null 2>&1; then
+  echo "Docker buildx is available for multi-arch builds"
+  DOCKER_BUILDX=true
+else
+  echo "Docker buildx not available, falling back to single-arch builds"
+  DOCKER_BUILDX=false
+fi
+
+
+repository=${REPOSITORY:-ghcr.io/kcp-dev/kcp}
```
embik: Does it make sense to have this default here? We aren't going to use this in CI after all, and I would like to avoid someone with maintainer permissions to run this by mistake and push images.

hack/build-image-docker.sh
```diff
@@ -0,0 +1,215 @@
...
+if [ -z "${DRY_RUN:-}" ]; then
+  echo "Logging into GHCR ..."
+  
+  if [ "$DOCKER_BUILDX" = true ]; then
+    # buildx with --push already pushed during build
+    echo "Images already pushed during buildx build"
+  else
+    # Regular docker - need to login and push
+    if [ -n "${KCP_GHCR_USERNAME:-}" ] && [ -n "${KCP_GHCR_PASSWORD:-}" ]; then
+      echo "$KCP_GHCR_PASSWORD" | docker login ghcr.io -u "$KCP_GHCR_USERNAME" --password-stdin
```
embik: Similar thing here: If this isn't meant for CI, let's just run through these automated login steps?

embik (commented):
Few notes.

hack/build-image-docker.sh
mjudeikis: We still need it to get tag/version

hack/build-image-docker.sh
mjudeikis: we still need it. I know I do :) Just removing KCP_

hack/build-image-docker.sh
mjudeikis: made it required param

PR #3628, opened 2025-10-07
turn cluster resolver middleware into a mux that can recognize cluster names in path mappings

xrstf:
## Summary
This PR implements an extension to the routing capabilities of the front-proxy. Previously it could only perform prefix matches (and those were mostly manually implemented), like sending everything beginning with `/services/` to a virtual-workspace endpoint.

With this PR, the configured path in the front-proxy's mapping is treated a a path pattern like `http.ServerMux` expects it, i.e. placeholders like `{cluster}` can be used.

The use of these placeholders is limited since only core kcp code can access and make use of them. For now, there is a single special placeholder called `{cluster}` which can be used to provide a cluster context especially to virtual workspace endpoints (i.e. make the front-proxy understand that `/services/myservice/clusters/abc123/apis/` refers to cluster `abc123`). This cluster context is then important when the per-workspace authentication feature is used.

I could see this being extended to help in per-shard routing for virtual workspaces, which is currently still an unsolved TODO: now more of the codebase uses a standard `http.ServerMux`, so it forms a good basis for extensions in the future.

However, the whole topic of routing in kcp is IMHO still quite confusing. There are special routes like `/clusters/`, there are multiple places throughout the codebase that use differnt methods of parsing URLs. I would at some point like to reorganize and clean it up a bit.

---

Code-wise, this PR replaces the ClusterResolver's custom string parsing to instead use a mux. This allows it to set the cluster in the request context, like we already do in the `/clusters/`-handler. Note that all configured mappings are added twice to the mux in order to be compatible with the old prefix-matching (configuring a path mapping `/foo` must match the literal `/foo` and `/foo/bar`, so two mux routes are necessary).

The proxy.HttpHandler was thinned out massively by simply relying on the already resolved data in the request context, rather than doing an index lookup and string parsing all over again. The trick to make this work was to not accidentally lose the query string parameters, which the ClusterResolver middleware up until now did when constructing the sharded target URL.

The mapping types were moved into a standalone package to solve an import loop.

## What Type of PR Is This?
/kind feature

## Release Notes
```release-note
Path mappings in the front-proxy are treated as standard Go ServerMux patterns and can make use of the `{cluster}` placeholder to provide a cluster context to the WorkspaceAuthentication for virtual workspaces (e.g. `/services/myservice/clusters/{cluster}`).
```

### DIFF
--- a/cmd/sharded-test-server/frontproxy.go
+++ b/cmd/sharded-test-server/frontproxy.go
@@ -77,6 +77,14 @@ func startFrontProxy(
 			ProxyClientKey:  filepath.Join(workDirPath, ".kcp-front-proxy", "requestheader.key"),
 		},
 		{
+			Path:            "/e2e/clusters/{cluster}/",
+			Backend:         "https://localhost:2443",
+			BackendServerCA: filepath.Join(workDirPath, ".kcp", "serving-ca.crt"),
+			// in the existing testcases, these two do not matter, but have to be non-empty
+			ProxyClientCert: filepath.Join(workDirPath, ".kcp-front-proxy", "requestheader.crt"),
+			ProxyClientKey:  filepath.Join(workDirPath, ".kcp-front-proxy", "requestheader.key"),
+		},
+		{
 			Path: "/clusters/",
 			// this path is not actually used, since shard URLs are determined based on the Shard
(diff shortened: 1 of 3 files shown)


embik (approved):
/approve

PR #3630, opened 2025-10-07
DynamicRESTMapper: add mapping for core k8s and kcp types

gman0:
DynamicRESTMapper only watches for bound and regular CRDs, while CRDs from `system:system-crds` are not included at all. This PR adds the mappings for these too.

The mapper is now split into two controllers:
* DynamicTypesController: manages dynamic, cluster-local types added through bindings or local CRDs,
* BuiltinTypesController: manages core k8s and kcp types, common to all clusters.

Additionally, default group-versions are now populated automatically, making resource resolution where version is optional work.

The built-in mappings are now shared, no need to copy them to each cluster as it was before (see `DynamicRESTMapper.builtin`).

## What Type of PR Is This?

/kind feature

## Release Notes

```release-note
NONE
```


pkg/reconciler/dynamicrestmapper/defaultrestmapper_mutable.go
```diff
@@ -41,6 +43,24 @@ func (m *DefaultRESTMapper) add(typeMeta typeMeta) {
 
 	m.kindToPluralResource[kind] = plural
 	m.kindToScope[kind] = meta.RESTScopeRoot
+
+	foundDefaultVersion := false
+	for i := range m.defaultGroupVersions {
+		if m.defaultGroupVersions[i].Group == typeMeta.Group {
+			if typeMeta.Version > m.defaultGroupVersions[i].Version {
```
mjudeikis, line 50: How does this work? Do we have served=true here for type? Not quite sure why highest wins here.

pkg/reconciler/dynamicrestmapper/dynamicrestmapper_builtin_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	return c, nil
+}
+
+func (c *BuiltinTypesController) enqueueCRD(crd *apiextensionsv1.CustomResourceDefinition, logger logr.Logger) {
+	if !apiextensionshelpers.IsCRDConditionTrue(crd, apiextensionsv1.Established) {
+		// The CRD is not ready yet. Nothing to do, we'll get notified on the next update event.
+		return
+	}
+
+	key := fmt.Sprintf("%s.%s/%s", crd.Spec.Group, crd.Spec.Names.Plural, crd.Name)
```
mjudeikis: This will be in foramt `example.corp.com.machines/machine`

Can you add comment that first part must be Group.Resource. Basically explain key structure as its custom key

pkg/reconciler/dynamicrestmapper/dynamicrestmapper_builtin_controller.go
```diff
@@ -0,0 +1,235 @@
...
+	gr := schema.ParseGroupResource(parts[0])
+	crdName := parts[1]
+
+	crd, err := c.getCRD(systemCRDClusterName, crdName)
+	if err != nil && apierrors.IsNotFound(err) {
+		return err
+	}
+
+	// Remove and add the mapping -- this way we can refresh any existing mappings.
+
```
mjudeikis: remove new line

pkg/server/controllers.go
```diff
@@ -1663,8 +1663,29 @@ func (s *Server) installGarbageCollectorController(ctx context.Context, config *
 	})
 }
 
-func (s *Server) installDynamicRESTMapper(ctx context.Context, config *rest.Config) error {
-	c, err := dynamicrestmapper.NewController(ctx, s.DynRESTMapper,
+func (s *Server) installDynamicRESTMapper(ctx context.Context, _ *rest.Config) error {
```
mjudeikis: Can we remove the parameter rest.config? Dont remember if its "standard function name" or we can remove it....

mjudeikis (commented):
few nits

pkg/reconciler/dynamicrestmapper/dynamicrestmapper_builtin_controller.go
gman0: Ohh that's a bug, thanks for spotting this! It indeed should be `<Resource>.<Group>/<CRD name>`. We don't have a test that covers deletion for built-in types so this went unchecked. I'll try to think of a way to test this.

pkg/reconciler/dynamicrestmapper/defaultrestmapper_mutable.go
gman0: Yes, by the time the type info gets here, it was already checked by the controllers:

* BuiltinTypesController:
  * https://github.com/gman0/kcp/blob/drm-builtin-types/pkg/reconciler/dynamicrestmapper/dynamicrestmapper_builtin_controller.go#L176-L184
* DynamicTypesController:
  * https://github.com/gman0/kcp/blob/drm-builtin-types/pkg/reconciler/dynamicrestmapper/dynamicrestmapper_dynamic_controller.go#L403-L411
  * https://github.com/gman0/kcp/blob/drm-builtin-types/pkg/reconciler/dynamicrestmapper/dynamicrestmapper_dynamic_controller.go#L436-L447

The version must be `Served=true` to be added to the `gvkrs` slice, and only this `gvkrs` slice is then added to the mapper.

pkg/server/controllers.go
gman0: Sure, I don't think there is any standard for this.

pkg/reconciler/dynamicrestmapper/dynamicrestmapper_builtin_controller.go
gman0: Fixed in https://github.com/gman0/kcp/blob/drm-builtin-types/pkg/reconciler/dynamicrestmapper/dynamicrestmapper_builtin_controller.go#L125

Also I've added a test that checks for adding, updating and deleting CRDs in `system:system-crds`: https://github.com/gman0/kcp/blob/drm-builtin-types/test/integration/dynamicrestmapper/dynamicrestmapper_test.go#L163-L243

I confirm that without the fix above, this test wouldn't pass.

pkg/reconciler/dynamicrestmapper/dynamicrestmapper_builtin_controller.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+		return nil, err
+	}
+	return gvkrs, nil
+}
+
+func (c *BuiltinTypesController) process(ctx context.Context, key string) error {
+	logger := logging.WithQueueKey(klog.FromContext(ctx), key)
+
+	parts := strings.Split(key, "/")
```
mjudeikis: Still want a comment with example key so reader does not need to read code to understand what type of custom key is this. same in the place where key is consturcted.
Simple:
// Key example <resource.group>/<crd-name>: machines.corp.example.com/machines

gman0: I ended up using the standard `kcpcache.MetaClusterNamespaceKeyFunc()`  because CRD names in system-crds are validated to be `<resource>.<group>` anyway.

gman0: So in `BuiltinTypesController.process()` we can just extract the GR from the CRD name, and it's fine even if the CRD itself is deleted.

gman0:
/retest

Both pull-kcp-test-e2e-sharded and pull-kcp-test-e2e-multiple-runs failed at

```
=== Failed
=== FAIL: test/e2e/workspacetype TestWorkspaceTypes/create_a_workspace_without_an_explicit_type,_get_default_type (31.79s)
    controller_test.go:340: Created root:organization workspace root:e2e-workspace-npsrh as /clusters/36qnjtmawxj1u4b7 on shard "shard-1"
    controller_test.go:346: Create a workspace without explicit type
    controller_test.go:346: Saving test artifacts for test "TestWorkspaceTypes/create_a_workspace_without_an_explicit_type,_get_default_type" under "/logs/artifacts/TestWorkspaceTypes/create_a_workspace_without_an_explicit_type,_get_default_type/935858795".
    controller_test.go:346: 
        	Error Trace:	/home/prow/go/src/github.com/kcp-dev/kcp/test/e2e/workspacetype/controller_test.go:74
        	            				/home/prow/go/src/github.com/kcp-dev/kcp/test/e2e/workspacetype/controller_test.go:346
        	Test:       	TestWorkspaceTypes/create_a_workspace_without_an_explicit_type,_get_default_type
        	Messages:   	workspace should be ready

=== FAIL: test/e2e/workspacetype TestWorkspaceTypes (0.04s)
    controller_test.go:331: Shared kcp server will target configuration "/home/prow/go/src/github.com/kcp-dev/kcp/.kcp/admin.kubeconfig"
    controller_test.go:331: Waiting for readiness for server at https://127.0.0.1:6444
```

PR #3636, opened 2025-10-08
Update shard validation

mjudeikis:
## Summary

Align shard validation to be same in admission and config.go

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Consistently use the user-provided base URL as the default for ShardBaseURL and VirtualWorkspacesURL
```

### DIFF
--- a/pkg/admission/shard/admission.go
+++ b/pkg/admission/shard/admission.go
@@ -65,6 +65,15 @@ func (o *shard) Admit(_ context.Context, a admission.Attributes, _ admission.Obj
 		return fmt.Errorf("failed to convert unstructured to Shard: %w", err)
 	}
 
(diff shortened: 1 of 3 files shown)


embik:
I'm not super happy with the release note. That only makes sense to us as developers, what does it fix for users? I think the answer is something along the lines of "Configure user-provided base URL when defaulting shard external and virtual workspace URLs"?

mjudeikis:
> I'm not super happy with the release note. That only makes sense to us as developers, what does it fix for users? I think the answer is something along the lines of "Configure user-provided base URL when defaulting shard external and virtual workspace URLs"?

Updated. Feel free to change if you think it still needs changing

embik:
Looks great now, thanks!

Issue #3653, opened 2025-10-17
bug: sdk is not go install'able anymore


mjudeikis:
### Describe the bug

```
/tmp/test ❯ go install github.com/kcp-dev/kcp/sdk/cmd/apigen@v0.0.0-20251015071049-e6ae71b74882                                                                                                                                           10:59:21
go: downloading github.com/kcp-dev/kcp v0.0.0-20251015071049-e6ae71b74882
go: github.com/kcp-dev/kcp/sdk/cmd/apigen@v0.0.0-20251015071049-e6ae71b74882 (in github.com/kcp-dev/kcp/sdk@v0.0.0-20251015071049-e6ae71b74882):
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.
```

### Steps To Reproduce

Due to existing of `replace` in go.mod it is not installable now.

### Expected Behaviour

We should move it to stagging repository as with othe repos. 
See https://kubernetes.slack.com/archives/C09C7UP1VLM/p1760601820003479

xmudrii:
This is now working as expected with the new module (`github.com/kcp-dev/sdk`):
```
$ go install github.com/kcp-dev/sdk/cmd/apigen@v0.0.0-20251103193518-385b036e4c87
go: downloading github.com/kcp-dev/sdk v0.0.0-20251103193518-385b036e4c87
```
/close

PR #3656, opened 2025-10-20
Make SDK installable

mjudeikis:
## Summary

This makes `sdk` installable again. This got broken after the monorepo migration.
We have a follow-up issue https://github.com/kcp-dev/kcp/issues/3653, but it's blocking me now on a different project. 

Both replaces are not changing often, so this should be ok until we get the 3653 fixes. 

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Make SDK `go installable` after monorepo migration. This is a temporary solution. 
```


embik (approved):
/approve

Makes me unhappy but we should have thought about this a bit more before migrating us into the situation.

PR #3670, opened 2025-10-21
Implement support for references in permission claims

### LINKED ISSUE #3409: feature: Extend permissions claims with references
### Feature Description

Extend permissions claims with references
https://docs.google.com/document/d/1toEU8_tldfc-kIWDgKfROyPVJGhhWDrr2cU0ayRj76I/edit?tab=t.0#heading=h.24igyjbjpe3n

Related to https://github.com/kcp-dev/kcp/issues/1937

TL:DR:
instead of `all: true` user should be able to do reference:
```
    resourceSelector:
     - reference:
         resource: MangoDBInstance
         group: mangodb.amazingcor.com
        type: corev1.ObjectReference
        JSONPath: spec.configuration
```

### Proposed Solution

https://docs.google.com/document/d/1toEU8_tldfc-kIWDgKfROyPVJGhhWDrr2cU0ayRj76I/edit?tab=t.0#heading=h.24igyjbjpe3n 

### Alternative Solutions

_No response_

### Want to contribute?

### Additional Context

_No response_

xrstf:
## Summary
This PR attempts to implement reference-based access to claimed resources in the apiexport virtual workspace. Where instead of the APIBinding directly naming the (for example) allowed Secrets to access, it would instead say "allow all secrets that are mentioned in this JSONPath in the following other resources". Think of allowing all Secrets that are mentioned in `spec.secretName` in Certificate objects.

The idea is to handle most of the access control by extending the permissionclaim labeler, which will calculate for each and every claimed object, whether that object should be included in an APIExport's virtual workspace. The resullt ofthat calculation is then stored as a label and this label is silently injected into requests to the apiexport virtual workspace, thereby filtering the objects down to the desired set.

With `matchAll`, this is easy, just label all the claimed objects. For label selectors this is also still somewhat easy because for calcualting if one object should be labeled, only that one object needs to be looked at.

However for references, suddenly to calculate the label for objects of resource X, we need to list all objects of resource Y (in full, not just their metadata) and evaluate JSONPaths against all of them to determine the set of allowed object names. This means the labels will need to be recalculated for all X objects whenever any Y object changes.

In effect that means kcp would need to either (like in this primitive unfinished PR) list all the Y objects dynamically or keep runtime informers around for basically all bound resources. And changes to one object can suddenly balloon into label changes on many objects (this was already true with APIBindings, but those change much less frequently than the "interesting" workload resources that users bound into their workspaces), and also changes to a claimed object (resource X) will force kcp to re-evaluate all Y objects.

None of this touches yet on the authorization that would need to happen for when objects are created (I am not sure when a new object is labeled and when authz for that action happens).

---

Since we currently have no strong usecase for this feature and fear the implications in terms of complexity and performance, we abandoned this approach for now.

## What Type of PR Is This?
/kind feature

## Related Issue(s)
Fixes #3409

## Release Notes
```release-note
TBD
```


xrstf:
I pushed my branch into this repository, but cannot change this PR and am loo tazy to create a new PR just to instantly close it. :grin:

PR #3671, opened 2025-10-21
fix: kubectl-kcp fail with error message instead of panic if crd defines webhook with service ref

m-szalik:
# Summary

## kubectl kcp crd snapshot

In the CRD Webhook can be defined as URL or reference to a service. 
The 2nd variant is not supported. This PR adds error message instead of crash with panic.

### CRD Example:
```yaml
spec:
  conversion:
    strategy: Webhook
    webhook:
      clientConfig:
        caBundle: "..."
        service:
          name: some-service
          namespace: some-namespace
          path: /convert
          port: 443
```
# What Type of PR Is This?
/kind bug

# Related Issue(s)

# Release Notes

```release-note
`kubectl kcp` returns error instead of panic when converting CRD with service webhook reference
```

### DIFF
--- a/sdk/apis/apis/v1alpha1/crd_to_apiresourceschema.go
+++ b/sdk/apis/apis/v1alpha1/crd_to_apiresourceschema.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"errors"
 	"fmt"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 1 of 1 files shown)


embik:
I believe that the lack of support for service-based webhooks is on purpose. kcp doesn't have `Service` objects (or any service routing, for that matter), thus you can't reference a service URL (like this PR generates) because you can't expect it to be there and accessible to the kcp apiserver.

embik:
I think it's a little unclear (at least to me) what scenario / use case you are trying to solve here. If you take a CRD with a service URL as webhook, it will by design not function (unless the webhook is running in the same cluster as kcp itself, which is not a given), even after merging this PR.

m-szalik:
@embik 
You are right. 
I have change my commit and PR description.

embik:
Thank you very much! Can you please also update the PR title to clarify the intention and make sure that the release notes block in the PR description is properly formatted? See [here](https://github.com/kcp-dev/kcp/blob/main/.github/pull_request_template.md?plain=1#L39).

/kind bug

PR #3672, opened 2025-10-21
Add docs for virtual resources and CachedResources

### LINKED ISSUE #3662: docs: document virtual resources with Replication VW as an example
We're missing docs on VRs.

gman0:
## Summary

This PR adds docs for  virtual resources and CachedResources.

## What Type of PR Is This?

/kind documentation

## Related Issue(s)

Fixes #3662 

## Release Notes

```release-note
NONE
```

### DIFF
--- a/config/examples/virtualresources/README.md
+++ b/config/examples/virtualresources/README.md
@@ -1,7 +1,7 @@
 # VirtualResources Example
 
-This example shows usage of VirtualResources, together with CachedResources.
-The goal of VirtualResources is to distribute static, read-only resources to multiple clusters
+This example shows usage of virtual resources, together with CachedResources.
+The goal of CachedResources is to distribute static, read-only resources to multiple clusters
 in a scalable way.
 
 ## Setup
@@ -19,15 +19,16 @@ in a scalable way.
    kubectl ws create provider --enter
 
    kubectl create -f config/examples/virtualresources/crd-instances.yaml
-   # this this to work we always require apiresource schema to be present
+   # for this to work we always require apiresource schema to be present
    kubectl create -f config/examples/virtualresources/apiresourceschema-instances.yaml
    kubectl create -f config/examples/virtualresources/instances.yaml
 
-   # create caching for the resources
+   # create caching for the resources with a pre-made CachedResources identity
+   kubectl create -f config/examples/virtualresources/cached-resource-identity.yaml
    kubectl create -f config/examples/virtualresources/cached-resource-instances.yaml
    ```
 
-3. Create a an APIResourceSchema for actual virtual machines to be distributed,
+3. Create an APIResourceSchema for actual virtual machines to be distributed,
    which will be using instance types. 
 
    ```bash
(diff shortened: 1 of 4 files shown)


embik:
/retest

CI eviction.

gman0:
I've rephrased a few sentences to make them read a bit better.

docs/content/concepts/apis/exporting-apis.md
```diff
@@ -534,3 +534,58 @@ widgets                example.kcp.io/v1alpha1   false        Widget
...
+
+1. The `storage` block with `crd` declaration below is implicit. No need to define it manually.
+
+
+### Virtual resources
+
+!!! warning
+    As of 0.29, this feature is of alpha-version quality. To use it, enable the `CachedAPIs` feature gate.
+
+An APIExport resource may be defined with `virtual` storage. This means the resource is provided by a [virtual workspace](../workspaces/virtual-workspaces.md), and is projected into APIBinding's workspace.
```
mirzakopic, line 565: Looks good to me, I would have just explained maybe from a consumers point of view what can be done with VRs. Maybe in a follow up?

docs/content/concepts/apis/cached-resources.md
```diff
@@ -0,0 +1,232 @@
...
+kind: CachedResource
+metadata:
+  name: my-cached-widgets
+spec:
+  group: example.kcp.io
+  version: v2
+  resource: widgets
+```
+
+The snippet above shows an example where all `widgets.v2.example.kcp.io` objects in the workspace are replicated to the cache. There are some constraints on what resources may be replicated:
```
mjudeikis: We talk about instances, lets use instances as example too.

docs/content/concepts/apis/exporting-apis.md
mjudeikis: Lets just refer to cachedResource document for more details how this can work with CachedResources

docs/content/concepts/apis/exporting-apis.md
```diff
@@ -534,3 +534,58 @@ widgets                example.kcp.io/v1alpha1   false        Widget
...
+spec:
+  resources:
+    - group: example.kcp.io
+      name: managedwidgets
+      schema: v220801.managedwidgets.example.kcp.io
+      storage:
+        virtual:
+          reference: # (1)
+            apiGroup: example.kcp.io
+            kind: ManagedWidgetsEndpointSlice
```
mjudeikis: Lets not overcomplicate this for now. Lets use CachedResourceEndpoint slice, and add note in the bottom that virtual resources machinery is generic, and while we use it CachedResources feature, one can implement its own {any}EndpointSlice and as long as they comply with EndpointSlice machinery - it would work. 


But lets leave this as foot note ^

docs/content/concepts/apis/cached-resources.md
gman0: I've switched to `cpuflavors.v1.cloud.example.com` in the whole document. 

The CachedResource object is renamed to `cpuflavors-v1` because if the user ever needs to update to v2, they should already be using versioned naming. I think we may even want to enforce some naming scheme on CachedResources -- in any case there can be only one CachedResource per GVR.

docs/content/concepts/apis/exporting-apis.md
gman0: Updated, hopefully it's better now.

PR #3674, opened 2025-10-23
add bind address to quickstart guide

SimonTheLeg:
## What Type of PR Is This?

/kind documentation

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/docs/content/setup/quickstart.md
+++ b/docs/content/setup/quickstart.md
@@ -27,7 +27,7 @@ Extract `kcp` and `kubectl-kcp-plugin` and place all the files in the `bin` dire
 You can start kcp using this command:
 
 ```shell
-kcp start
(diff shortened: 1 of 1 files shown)


docs/content/setup/quickstart.md
```diff
@@ -27,7 +27,7 @@ Extract `kcp` and `kubectl-kcp-plugin` and place all the files in the `bin` dire
 You can start kcp using this command:
 
 ```shell
-kcp start
+kcp start --bind-address=127.0.0.1
```
embik, line 30: Should we also include `--external-hostname 127.0.0.1` if we are doing this already?

embik: Did some quick testing but I think this flag might be redundant in this situation.

PR #3677, opened 2025-10-23
fix initializer update error kind

SimonTheLeg:
The object which is modified through this virtualworkspace is the LogicalCluster, not the workspace. Therefore the invalidUpdateErr should also be on the LogicalCluster.

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


embik (approved):
/approve

PR #3682, opened 2025-10-24
fix concurrent map read access and write panic

xrstf:
## Summary

The controller already uses the RWMutex to guard the calls to state.apply() (which is important because we have 2 worker threads in this controller), but failed to also guard the state.getGVKRs() call. apply() would write to the underlying map in the DefaultRESTMapper, getGVKRs() would read. Boom.

This panic happens extremely rarely and only surfaced when I moved the initialisation of the DiscoveringDynamicSharedInformerFactory up from server.NewServer() into server.NewConfig(). With that change, kcp begun to panic in maybe 30% of cases within the first 5 seconds of runtime.

While moving the DDSIF might not make sense, a race condition is a race condition, so we should fix it.

## What Type of PR Is This?
/kind bug

## Release Notes
```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/dynamicrestmapper/dynamicrestmapper_builtin_controller.go
+++ b/pkg/reconciler/dynamicrestmapper/dynamicrestmapper_builtin_controller.go
@@ -201,9 +201,6 @@ func (c *BuiltinTypesController) gatherGVKRsForCRD(crd *apiextensionsv1.CustomRe
 }
 
 func (c *BuiltinTypesController) gatherGVKRsForMappedGroupResource(gr schema.GroupResource) ([]typeMeta, error) {
-	c.state.lock.RLock()
-	defer c.state.lock.RUnlock()
-
 	gvkrs, err := c.state.builtin.getGVKRs(gr)
 	if err != nil {
 		if meta.IsNoMatchError(err) {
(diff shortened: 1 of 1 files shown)


pkg/reconciler/dynamicrestmapper/dynamicrestmapper_builtin_controller.go
```diff
@@ -201,6 +201,9 @@ func (c *BuiltinTypesController) gatherGVKRsForCRD(crd *apiextensionsv1.CustomRe
 }
 
 func (c *BuiltinTypesController) gatherGVKRsForMappedGroupResource(gr schema.GroupResource) ([]typeMeta, error) {
+	c.state.lock.RLock()
```
gman0: Let's just move the `BuiltinTypesController.gatherGVKRsForMappedGroupResource` call in `BuiltinTypesController.process` after `c.state.lock.Lock()` otherwise there's a race between reading (by `getGVKRs`) and writing (by `apply`).

I see the same issue in `DynamicTypesController` too where write-after-read is not atomic, but this can be fixed in a separate PR.

pkg/reconciler/dynamicrestmapper/dynamicrestmapper_builtin_controller.go
gman0, line 238: This `gatherGVKRsForCRD` call could go out of the critical section as it doesn't touch the shared `c.state` data, but let's leave it for a well-deserved locks refactor later.

PR #3683, opened 2025-10-24
CachedResources: replicate full objects

gman0:
## Summary

DDSIF, the informer for retrieving local objects, yields only PartialMetadataObjects. This PR adds a live GET call to retrieve the full object for replication.

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3478

## Release Notes

```release-note
NONE
```

### DIFF
--- a/test/e2e/virtualresources/cachedresources/vr_cachedresources_test.go
+++ b/test/e2e/virtualresources/cachedresources/vr_cachedresources_test.go
@@ -367,16 +367,29 @@ func TestCachedResources(t *testing.T) {
 	// Eventually we should see two of each in the resourcesCounter,
 	// i.e. one for each consumer workspace.
 	t.Logf("Creating a Cowboy in %q", providerPath)
-	cowboyOne, err := wildwestClusterClient.Cluster(providerPath).WildwestV1alpha1().Cowboys("default").Create(ctx, &wildwestv1alpha1.Cowboy{
+	cowboyOne := createCowboy(t, ctx, wildwestClusterClient, providerPath, &wildwestv1alpha1.Cowboy{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "cowboys-1",
+			Name:      "cowboys-1",
+			Namespace: "default",
 		},
-	}, metav1.CreateOptions{})
+		Spec: wildwestv1alpha1.CowboySpec{
+			Intent: "cowboys-1-spec",
+		},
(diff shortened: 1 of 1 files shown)


pkg/reconciler/cache/cachedresources/replication/replication_reconcile.go
gman0: This early-return with selector mismatch is an additional opportunity to optimize away the GET call: we could feed `r.getLocalCopy` with the selector, and skip on mismatch before getting the full object.

Let me know if you'd like to do it still in this PR or separately.

pkg/reconciler/cache/cachedresources/replication/replication_reconcile.go
```diff
@@ -237,6 +211,54 @@ func (c *Controller) reconcile(ctx context.Context, gvrKey string) error {
 			return c.kcpCacheClient.Cluster(cluster.Path()).CacheV1alpha1().CachedObjects().Delete(ctx, cachedObjName, metav1.DeleteOptions{})
 		},
 	}
+	// r.getLocalCopy is defined separately because it calls r.getGlobalCopy internally.
+	r.getLocalCopy = func(cluster logicalcluster.Name, namespace, name string) (*unstructured.Unstructured, error) {
```
mjudeikis: Name now does now make sense. Can we rename to something better as it calls both now?

pkg/reconciler/cache/cachedresources/replication/replication_reconcile.go
mjudeikis: Lets do that, and rename the function so its something clearer names what it actually does now

pkg/reconciler/cache/cachedresources/replication/replication_controller.go
```diff
@@ -88,10 +90,33 @@ func NewController(
...
+	globalHandler, err := c.replicated.Global.AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: func(obj interface{}) bool {
+			cachedObj := obj.(*cachev1alpha1.CachedObject)
+			labels := cachedObj.Labels
+			if labels == nil {
+				return false
+			}
+
+			// Skip CachedObjects that are not coming from the source CachedResource.
+
```
mjudeikis, line 102: nit: dont need these new lines :)

PR #3688, opened 2025-10-29
Bump golangci-lint to 2.6.0

xrstf:
## Summary
This is just maintenance. I also bumped logcheck because 0.9 doesn't work on Go 1.25 systems.

gocritic now complains about the misplaced Deprecation notices, and after fixing them, I had to adjust a bit of code in our codebase. However I did not remove the deprecated CreateSheriff, even though it was meant to be deprecated allll the way back in #1642 (2022). But since that never worked, nobody noticed and so I would declare today the official deprecation day :grin: 

## What Type of PR Is This?
/kind chore

## Release Notes
```release-note
NONE
```


mjudeikis (approved):
/lgtm
/approve

Issue #3673, opened 2025-10-22
feature: add kcp.io/path to apibindings


mjudeikis:
### Feature Description

Currently, https://github.com/kcp-dev/multicluster-provider accepts only a hash value for the cluster.
Which is a poor usage experience when the user has only a canonical path available (from `logicalcluster.From`).

We should extend multicluster-provider to accept both hash and canonical paths to resolve to cluster. 

This can be done already if one claims LogicalCluster, but this elevated apibinding/export contract quite high.

### Proposed Solution

We should add `kcp.io/path` with logicalcluster annotation to every non system apibinding (there might be some clash trying to annotate system ones) in admission (github.com/kcp-dev/kcp/pkg/admission/pathannotation/pathannotation_admission.go) so every VW could be resolved both to representations of cluster.

Extend MCR to index clusters on both values and accept them.

We should not support mixed values of hash and canonical paths. 

### Want to contribute?

- [ ] I would like to work on this issue.

PR #3691, opened 2025-10-30
Add path annotation to apibindings

mjudeikis:
## Summary

This adds path annotation to apibindings. APIbindings are accessible via VirtualWorkspaces. This means providers will be able to resolve the canonical path and cluster more easily. 

```
- apiVersion: apis.kcp.io/v1alpha2
  kind: APIBinding
  metadata:
    annotations:
      kcp.io/cluster: hezfu00lhrvcdnls
      kcp.io/path: root:test
```

Caveat: there is a race condition with logicalcluster. But it resolves quite quickly due to the binding handshake, which is followed by the update. Not ideal, but the alternatives (the ones I could think of) were even more hacky. 

## What Type of PR Is This?

/kind feature 

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3673 

## Release Notes

```release-note
Add kcp.io/path annotation to APIBindings
```

### DIFF
--- a/pkg/admission/pathannotation/pathannotation_admission.go
+++ b/pkg/admission/pathannotation/pathannotation_admission.go
@@ -71,6 +71,7 @@ type pathAnnotationPlugin struct {
 var pathAnnotationResources = sets.New[string](
 	apisv1alpha2.Resource("apiexports").String(),
 	cachev1alpha1.Resource("cachedresources").String(),
+	apisv1alpha2.Resource("apibindings").String(),
 	tenancyv1alpha1.Resource("workspacetypes").String(),
 )
 
@@ -107,6 +108,10 @@ func (p *pathAnnotationPlugin) Admit(ctx context.Context, a admission.Attributes
 
 	logicalCluster, err := p.getLogicalCluster(clusterName, corev1alpha1.LogicalClusterName)
 	if err != nil {
+		// We skip adding for system bindings if the logical cluster is not found during creation. This is racy during workspace bootstrap.
+		if apierrors.IsNotFound(err) && a.GetResource().GroupResource() == apisv1alpha2.Resource("apibindings") {
+			return nil
+		}
 		return admission.NewForbidden(a, fmt.Errorf("cannot get this workspace: %w", err))
 	}
 	thisPath := logicalCluster.Annotations[core.LogicalClusterPathAnnotationKey]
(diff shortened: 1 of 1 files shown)


mjudeikis:
moving to draft. some strange errors pop up. Investigating.

gman0:
Another approach would be to:

1. Skip mutating/validating objects whose logical cluster resolves to NotFound:
    * The reasoning would be that it's probably not a job of this plugin to deny requests with missing LogicalCluster -- something else should catch and validate this if we really need such a check.
2. Modify the APIBinding admission plugin to add an empty `kcp.io/path` annotation on the object if it's missing during create/update.
    * The pathannotation admission plugin would then catch this, and update the path accordingly once the APIBinding's logical cluster is ready.

Maybe this divides the responsibilities a bit better? But it may be that I'm missing some knowledge about why it is done like that in the first place :P

gman0:
Also, existing APIBindings are not updated right? Is that a problem?

mjudeikis:
> Another approach would be to:
> 
> 1. Skip mutating/validating objects whose logical cluster resolves to NotFound:
>    
>    * The reasoning would be that it's probably not a job of this plugin to deny requests with missing LogicalCluster -- something else should catch and validate this if we really need such a check.
> 2. Modify the APIBinding admission plugin to add an empty `kcp.io/path` annotation on the object if it's missing during create/update.
>    
>    * The pathannotation admission plugin would then catch this, and update the path accordingly once the APIBinding's logical cluster is ready.
> 
> Maybe this divides the responsibilities a bit better? But it may be that I'm missing some knowledge about why it is done like that in the first place :P

So my code is attempting to do option 2:
But I didn't want to do an empty annotation as some of the other code relies on this, and empty is invalid. So skipping kinda looked like a better alternative. 

And it still gets updated/added once a logical cluster is added and the next resync/ status update.

gman0:
I agree :+1: 

/lgtm

gman0:
/lgtm

We know the limitations, but it's better than the alternatives. And the limitations don't seem to be very serious.

xrstf:
For my own understanding: For just resolving the cluster ID to the workspace path, one could (like the syncagent does) simply also claim logicalclusters, right?

/approve

xrstf:
> This can be done already if one claims LogicalCluster, but this elevated apibinding/export contract quite high.

Yeah you said that already in the ticket. Nevermind. ^^

PR #3696, opened 2025-11-03
VR apiserver retrieves endpoint slices by a get call

### LINKED ISSUE #3695: bug: virtual resources apiserver resolves to wrong endpoint slice
### Describe the bug

VR apiserver needs to retrieve VW endpoint slice where it forwards the request, and it does so by listing.

https://github.com/kcp-dev/kcp/blob/fa4e118e4471de3c1714a229382d1af24cc95b99/pkg/server/virtualresources/server.go#L79-L100

The request's context is missing shard name, and so the listing needs to do a wildcard search.

Secondly, when the list is being constructed, cluster annotation is set incorrectly:

https://github.com/kcp-dev/kubernetes/blob/ffe1d7c8649b8fef60c69cf91924fb34a2c64fbb/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go#L830-L833

Each item is annotated with the cluster name from the request instead of their real cluster of origin.

Both of these issues then lead to a situation where, if there are more endpoint slices with the same GVR and name (in different workspaces), VR apiserver will fail the slice resolution because the endpoint slice list used for searching contains more then one match.

### Steps To Reproduce

* Create a CachedResource `my-cached-resource` in workspaces `root:a` and `root:b`. An associated CachedResourceEndpointSlice is created automatically with the same name.
* Create an APIExport in `b`, and bind to it in `root:c`
* List the bound cached resource in `c`.
* The request will fail with `Error from server (InternalError): Internal error occurred: error resolving resource: Internal error occurred: multiple objects found` because there was more than one endpoint slice with matching name.

### Expected Behaviour

VR apiserver should always pick correct endpoint slice.

### Additional Context

_No response_

gman0:
## Summary

Endpoint slices are now retrieved by GET-ting them from cache. The slice must be co-located with the APIExport that exports the VR, and so the export's shard and cluster name is used.

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #3695 

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/server/config.go
+++ b/pkg/server/config.go
@@ -696,6 +696,7 @@ func NewConfig(ctx context.Context, opts kcpserveroptions.CompletedOptions) (*Co
 			&virtualResourcesConfig,
 			vwClientConfig,
 			c.CacheDynamicClient,
+			c.Options.Extra.ShardName,
 			c.ShardVirtualWorkspaceURL,
 			c.ApiExtensionsSharedInformerFactory.Apiextensions().V1().CustomResourceDefinitions(),
 			c.KcpSharedInformerFactory.Apis().V1alpha2().APIBindings(),
--- a/pkg/server/virtualresources/config.go
+++ b/pkg/server/virtualresources/config.go
@@ -38,6 +38,7 @@ type ExtraConfig struct {
 	VWClientConfig       *rest.Config
 	DynamicClusterClient kcpdynamic.ClusterInterface
 
+	ShardName                      string
 	ShardVirtualWorkspaceURLGetter func() string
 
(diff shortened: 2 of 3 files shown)


gman0:
/retest

Maybe this resets the tide job?

gman0:
/retest

A bunch of tests failed with "Pod scheduling timeout."

gman0:
There was a `panic: cache *v1alpha1.LogicalCluster modified` in  test/e2e/apibinding TestAPIBindingCRDs. Unrelated, but something we should have a look later: https://github.com/kcp-dev/kcp/issues/3700

xrstf:
/retest

DNS hiccup in the CI cluster

PR #3708, opened 2025-11-08
Use non-versioned modules for staging repo dependencies

xmudrii:
## Summary

This is a somewhat simple PR but it's hard to recap its intention in a short title:

- There should be `replace` directives for all dependencies to other staging repositories. For example, if some module depends on `client-go` which depends on `apimachinery` and `code-generator/v3`, there should be `replace` directives for all three modules. It's kind of intuitive, but it's very easy to miss.
- Modules within `kcp-dev/kcp` shall not refer to staging repositories using versions or commit references (pseudo-versions). Instead, `v0.0.0` or `vX.0.0-00010101000000-000000000000` (if `X >= 2`) shall be used. Within `kcp-dev/kcp`, `replace` directives will point to the proper Go module within `staging/src/github.com/kcp-dev`. In published repositories (e.g. `github.com/kcp-dev/apimachinery`), publishing-bot will update all these references upon publishing from staging to published repositories.

## What Type of PR Is This?

/kind cleanup

## Release Notes
```release-note
NONE
```

/assign @xrstf @embik @mjudeikis

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,11 +20,11 @@ require (
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
 	github.com/kcp-dev/apimachinery/v2 v2.0.1-0.20250728122101-adbf20db3e51
-	github.com/kcp-dev/client-go v0.0.0-20250728134101-0355faa9361b
-	github.com/kcp-dev/code-generator/v3 v3.0.0-20250728122101-5b4ff5c24054
+	github.com/kcp-dev/client-go v0.0.0
+	github.com/kcp-dev/code-generator/v3 v3.0.0-00010101000000-000000000000
 	github.com/kcp-dev/embeddedetcd v1.0.3-0.20250805142358-a4839a83564a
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -20,11 +20,11 @@ require (
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
 	github.com/kcp-dev/apimachinery/v2 v2.0.1-0.20250728122101-adbf20db3e51
```
xmudrii, line 22: I'm aware that this is not non-versioned import, but I'm not able to turn it into `v2.0.0-00010101000000-000000000000` (`go mod tidy` reverts back to this version). I assume this is because `github.com/kcp-dev/apimachinery/v2` is a dependency in `kcp-dev/kubernetes` which we import here (`go mod why` tells nothing, but some search magic uncovers this ✨).

Nonetheless, this is the main/root module and it's not that important, so we just leave it as it is. There's anyways `replace` directive for it below.

Issue #3658, opened 2025-10-20
feature: add path to CachedResourceReference API


gman0:
### Feature Description

Currently the code around CachedResourceEndpointSlices assume their associated CachedResource is co-located in the same workspace. We should support external CachedResources too.

### Proposed Solution

Add `path` into [`CachedResourceReference struct`](https://github.com/kcp-dev/kcp/blob/adc738049d209e67adedf810db05cc37166ae64b/sdk/apis/cache/v1alpha1/types_cachedresource.go#L226) and integrate it where necessary.

### Want to contribute?

- [ ] I would like to work on this issue.

PR #3726, opened 2025-11-20
add path to cachedresource api

olamilekan000:
## Summary
```
change implements adds path to the cachedresource api so that a CachedResourceEndpointSlice can reference a `CachedResource` located in another workspace.
```
image

## Related Issue(s)

Fixes [3658](https://github.com/kcp-dev/kcp/issues/3658)

## Release Notes

```release-note
Added to path to cachedresource so that CachedResourceEndpointSlice can reference a CachedResource in another workspace
```

### DIFF
--- a/pkg/reconciler/cache/cachedresources/cachedresources_reconcile_endpointslice.go
+++ b/pkg/reconciler/cache/cachedresources/cachedresources_reconcile_endpointslice.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/kcp-dev/logicalcluster/v3"
 	cachev1alpha1 "github.com/kcp-dev/sdk/apis/cache/v1alpha1"
+	"github.com/kcp-dev/sdk/apis/core"
 )
 
 // endpointSlice creates CachedResourceEndopointSlice for the published resource.
@@ -43,6 +44,7 @@ func (r *endpointSlice) reconcile(ctx context.Context, cachedResource *cachev1al
 	}
 
 	clusterName := logicalcluster.From(cachedResource)
+	clusterPath := cachedResource.Annotations[core.LogicalClusterPathAnnotationKey]
 
 	_, err := r.getEndpointSlice(ctx, clusterName, cachedResource.Name)
 	if err != nil {
@@ -62,6 +64,7 @@ func (r *endpointSlice) reconcile(ctx context.Context, cachedResource *cachev1al
 				},
 				Spec: cachev1alpha1.CachedResourceEndpointSliceSpec{
 					CachedResource: cachev1alpha1.CachedResourceReference{
+						Path: clusterPath,
 						Name: cachedResource.Name,
 					},
 				},
--- a/test/e2e/reconciler/cachedresourceendpointslice/cachedresourceendpointslice_test.go
+++ b/test/e2e/reconciler/cachedresourceendpointslice/cachedresourceendpointslice_test.go
@@ -1,0 +1,198 @@
+/*
+Copyright 2025 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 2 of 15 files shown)


staging/src/github.com/kcp-dev/sdk/apis/cache/v1alpha1/register.go
```diff
@@ -34,7 +34,8 @@ func Kind(kind string) schema.GroupKind {
 
 // Resource takes an unqualified resource and returns a Group qualified GroupResource.
 func Resource(resource string) schema.GroupResource {
-	return SchemeGroupVersion.WithResource(resource).GroupResource()
+	res := SchemeGroupVersion.WithResource(resource).GroupResource()
```
gman0: Why change this?

gman0:
Thanks, looks good.

Can you please add an e2e test to `cachedresourceendpointslice/` in https://github.com/kcp-dev/kcp/tree/main/test/e2e/reconciler , similar to the one implemented in https://github.com/kcp-dev/kcp/tree/main/test/e2e/reconciler/apiexportendpointslice ?

If you're ok with that, do it only for the `path` param, and leave `partition` for now because we might want to change it later.

staging/src/github.com/kcp-dev/sdk/apis/cache/v1alpha1/register.go
olamilekan000: It was changed when I was debugging. I'll revert it.

test/e2e/reconciler/cachedresourceendpointslice/cachedresourceendpointslice_test.go
```diff
@@ -0,0 +1,220 @@
...
+	kcpClusterClient, err := kcpclientset.NewForConfig(cfg)
+	require.NoError(t, err, "failed to construct kcp cluster client for server")
+
+	dynamicClusterClient, err := kcpdynamic.NewForConfig(cfg)
+	require.NoError(t, err, "failed to construct dynamic cluster client for server")
+
+	// Create a CRD in the provider workspace
+	t.Logf("Creating CRD in provider workspace %q", providerPath)
+	mapper := restmapper.NewDeferredDiscoveryRESTMapper(memory.NewMemCacheClient(kcpClusterClient.Cluster(providerPath).Discovery()))
+	err = helpers.CreateResourceFromFS(ctx, dynamicClusterClient.Cluster(providerPath), mapper, nil, "crd_cpuflavors.yaml", testFiles)
```
gman0: Instead of adding a completely new CRD, can you please reuse one that already exists? 

There's a helper function for that too: https://github.com/kcp-dev/kcp/blob/main/test/e2e/fixtures/wildwest/bootstrap.go#L40

gman0:
Great, thanks!

~/lgtm~

Actually one more thing.

pkg/reconciler/cache/cachedresources/cachedresources_reconcile_endpointslice.go
```diff
@@ -62,6 +62,7 @@ func (r *endpointSlice) reconcile(ctx context.Context, cachedResource *cachev1al
 				},
 				Spec: cachev1alpha1.CachedResourceEndpointSliceSpec{
 					CachedResource: cachev1alpha1.CachedResourceReference{
+						Path: clusterName.Path().String(),
```
gman0: The implicitly created endpoint slice should have its `path` empty. We don't pass the cluster name to spec, only the path.

gman0: Or pass the cluster path, similar to what the APIExport reconciler is doing: https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/apis/apiexport/apiexport_reconcile.go#L124

This would actually be preferable, so that the behavior is consistent.

Issue #3729, opened 2025-11-24
bug: CachedResource fails to delete when resource does not exists or isn't valid


olamilekan000:
When trying to delete a `CachedResource`  with a resource that doesn’t exist, the deletion process hangs in the terminal and times out after a while.
The log below is what is being logged when the error happens
```
E1123 05:32:43.352648   35874 cachedresources_reconcile_replication.go:90] "Failed to get Kind for resource" err="no matches for apps/v1, Resource=deployments" component="kcp" postStartHook="kcp-start-controllers" reconciler="kcp-cached-resources-controller" key="1l5sxijy0c6hwyoh|test-bad" cachedresource.workspace="1l5sxijy0c6hwyoh" cachedresource.namespace="" cachedresource.name="test-bad" cachedresource.apiVersion="" resource="apps/v1, Resource=deployments"

```

PR #3730, opened 2025-11-24
fix bug when trying to delete a cached resource

olamilekan000:
## Summary
```
change fixes bug that prevents the deletion of a cachedresource that makes a
reference to a resource that doesn't have gvr
```
image

## Related Issue(s)

Fixes [3729](https://github.com/kcp-dev/kcp/issues/3729)

## Release Notes

```release-note
Fixed a bug that prevents the deletion of a cachedresource that makes a
reference to a resource that doesn't have GVR.
```

### DIFF
--- a/pkg/reconciler/cache/cachedresources/cachedresources_reconcile_replication.go
+++ b/pkg/reconciler/cache/cachedresources/cachedresources_reconcile_replication.go
@@ -72,7 +72,13 @@ func (r *replication) reconcile(ctx context.Context, cachedResource *cachev1alph
 	controller := r.controllerRegistry.get(controllerName)
 	// We setup controller even if we are deleting. This is to ensure that we can purge the cache.
 	// If for some reason was dead, we will recreate it.
+	danglingResources := cachedResource.Status.ResourceCounts != nil && cachedResource.Status.ResourceCounts.Cache > 0
 	if controller == nil {
+		if cachedResource.DeletionTimestamp != nil && !danglingResources {
+			cachedResource.Status.Phase = cachev1alpha1.CachedResourcePhaseDeleted
+			return reconcileStatusStopAndRequeue, nil
+		}
+
 		// Global informer is based on the CachedResource type and we construct index based on the schema labels.
 		controllerCtx, cancel := context.WithCancel(ctx)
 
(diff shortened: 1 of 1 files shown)


pkg/reconciler/cache/cachedresources/cachedresources_reconcile_replication.go
```diff
@@ -81,14 +87,24 @@ func (r *replication) reconcile(ctx context.Context, cachedResource *cachev1alph
 		// Local informer is based on the specific types we want to replicate.
 		local, err := r.discoveringDynamicKcpInformers.ClusterWithContext(ctx, cluster).ForResource(gvr)
 		if err != nil {
-			logger.Error(err, "Failed to get local informer for resource", "resource", gvr)
 			cancel()
+			if cachedResource.DeletionTimestamp != nil && !danglingResources {
```
gman0: Isn't it enough that this is already checked and handled at https://github.com/kcp-dev/kcp/pull/3730/files#diff-9723dbe0e395b66baceb9f3ae33514b289f83e93d441b248c1018b01d40df25fR77 ?

pkg/reconciler/cache/cachedresources/cachedresources_reconcile_replication.go
```diff
@@ -81,14 +87,24 @@ func (r *replication) reconcile(ctx context.Context, cachedResource *cachev1alph
...
+				return reconcileStatusStopAndRequeue, nil
+			}
+			logger.Error(err, "Failed to get local informer for resource", "resource", gvr)
 			return reconcileStatusStopAndRequeue, err
 		}
 		replicatedKind, err := r.dynRESTMapper.ForCluster(clusterName).KindFor(gvr)
 		if err != nil {
-			logger.Error(err, "Failed to get Kind for resource", "resource", gvr)
 			cancel()
+			if cachedResource.DeletionTimestamp != nil && !danglingResources {
```
gman0: The same as the comment above: isn't it enough that this is already checked and handled at https://github.com/kcp-dev/kcp/pull/3730/files#diff-9723dbe0e395b66baceb9f3ae33514b289f83e93d441b248c1018b01d40df25fR77 ?

pkg/reconciler/cache/cachedresources/cachedresources_reconcile_replication.go
olamilekan000: Those duplicate checks were from an earlier iteration where I was also inspecting the `Phase` of the cachedresource (e.g. handling Initializing  and deleting). A few CachedResources were still stuck in Initializing, so I added the guard around each error site to cover that. I Later simplified the logic to rely on DeletionTimestamp + .cache, but I forgot to remove the redundant branches. However, the early return should now be enough as you've pointed out.

PR #3731, opened 2025-11-24
fix missing livereload in local-docs makefile target

xrstf:
## Summary
I thought it was me, but no, apparently an upstream issue in mkdocs/clicks/something is accidentally disabling the default livereload functionality: https://github.com/mkdocs/mkdocs/issues/4032

The suggested workaround is to pin one of the dependencies to an older version.

## What Type of PR Is This?
/kind bug

## Release Notes
```release-note
NONE
```


xmudrii (approved):
/lgtm
/approve

Issue #3736, opened 2025-12-01
Add version banner warning for unreleased main branch documentation


olamilekan000:
At times, users view documentation for the unreleased "main" branch without realizing it contains features and APIs that are not yet available in stable releases. This causes confusion when users try to reference features described in `main` branch docs but find they don't exist in their installed version. There needs to be a clear distinction between unreleased development documentation and stable release documentation.

PR #3737, opened 2025-12-01
add warning banner for unreleased branch or version in the documentation

olamilekan000:
## Summary

```
change adds a warning banner in the documentation to ensure that there's a clear
distinction between the released and unreleased versions of the doc
```
image

## What Type of PR Is This?
/kind documentation

## Related Issue(s)

Fixes [3736](https://github.com/kcp-dev/kcp/issues/3736)

## Release Notes

```release-note
Added sticky warning banner on "main" branch documentation to alert users they're viewing unreleased content.
```

### DIFF
--- a/docs/overrides/main.html
+++ b/docs/overrides/main.html
@@ -1,0 +1,155 @@
+{% extends "base.html" %} {% block extrahead %} {{ super() }}
+<style>
+  #version-banner {
+    display: none;
+    position: sticky;
+    top: 0;
+    left: 0;
+    right: 0;
+    width: 100%;
+    background-color: #00bcd4;
+    color: white;
+    padding: 8px 16px;
+    text-align: center;
+    font-size: 14px;
+    line-height: 1.5;
+    z-index: 10000;
+    box-shadow: 0 2px 4px rgba(0, 0, 0, 0.1);
+  }
+
+  #version-banner a {
+    color: white;
+    text-decoration: underline;
+    margin-left: 8px;
+  }
+
+  #version-banner a:hover {
+    text-decoration: none;
+  }
+
+  .md-header {
+    transition: top 0.2s ease;
+  }
+</style>
+
+<script>
+  (function () {
+    "use strict";
+
+    const CONFIG = {
+      UNRELEASED_VERSION: "main",
+      RETRY_DELAY: 500,
+      HEADER_ADJUST_DELAY: 100,
+    };
+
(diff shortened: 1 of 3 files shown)


ntnn:
I like the header but the orange is a bit krass. Maybe a light blue, lilac or cyan?

xmudrii:
From Slack: https://kubernetes.slack.com/archives/C09C7UP1VLM/p1764769123353489?thread_ts=1764631722.014269&cid=C09C7UP1VLM

> I think the banner is nice, but I find the script is quite complex. I think we should first solve that `https://docs.kcp.io/kcp` redirects to the latest stable version. Then you can just use that link in the banner, it would always lead to the latest stable version, no need for complex JS code.
>
> I'm pretty sure that `https://docs.kcp.io/kcp` used to work like that, to lead to the latest stable version. I don't know why/when it broke.

olamilekan000:
> From Slack: https://kubernetes.slack.com/archives/C09C7UP1VLM/p1764769123353489?thread_ts=1764631722.014269&cid=C09C7UP1VLM
> 
> > I think the banner is nice, but I find the script is quite complex. I think we should first solve that `https://docs.kcp.io/kcp` redirects to the latest stable version. Then you can just use that link in the banner, it would always lead to the latest stable version, no need for complex JS code.
> > I'm pretty sure that `https://docs.kcp.io/kcp` used to work like that, to lead to the latest stable version. I don't know why/when it broke.

This works perfectly fine at the moment.
> I think we should first solve that `https://docs.kcp.io/kcp` redirects to the latest stable version

I think we can focus on the banner?

docs/overrides/main.html
```diff
@@ -0,0 +1,155 @@
...
+
+    return banner;
+  }
+
+  function updateBannerLink(banner) {
+    const link = banner.querySelector("#latest-version-link");
+    if (!link) return;
+
+    const latestPath = getLatestVersionPath();
+    console.log("latestPathlatestPath", latestPath)
```
xmudrii: Nit: I don't think this is needed in the prod.

xmudrii (changes requested):
Let's please remove `console.log` lines, I don't think these are needed.

docs/overrides/main.html
mjudeikis: this and should be good to go

olamilekan000: done

Issue #3753, opened 2025-12-10
fix TestValidatingAdmissionPolicyCrossWorkspaceAPIBinding flaky test


olamilekan000:
The TestValidatingAdmissionPolicyCrossWorkspaceAPIBinding test sometimes throws an error about "default" workspace not existing.

```
=== Failed
=== FAIL: test/e2e/conformance TestValidatingAdmissionPolicyCrossWorkspaceAPIBinding (63.70s)
    validatingadmissionpolicy_test.go:192: Shared kcp server will target configuration "/home/prow/go/src/github.com/kcp-dev/kcp/.kcp/admin.kubeconfig"
    validatingadmissionpolicy_test.go:192: Waiting for readiness for server at https://127.0.0.1:6444/
    validatingadmissionpolicy_test.go:207: Created root:organization workspace root:e2e-workspace-f9ptd as /clusters/2x2j0qne703tbzlb on shard "shard-1"
    validatingadmissionpolicy_test.go:208: Created root:universal workspace root:e2e-workspace-f9ptd:e2e-workspace-2pfwk as /clusters/30avdkeidcvqj5jd on shard "shard-1"
    validatingadmissionpolicy_test.go:209: Created root:universal workspace root:e2e-workspace-f9ptd:e2e-workspace-vj9g7 as /clusters/12wv83trvc3wt3dq on shard "root"
    validatingadmissionpolicy_test.go:220: Install a cowboys APIResourceSchema into workspace "root:e2e-workspace-f9ptd:e2e-workspace-2pfwk"
    validatingadmissionpolicy_test.go:273: Create an APIExport for it
    validatingadmissionpolicy_test.go:294: Create an APIBinding in workspace "root:e2e-workspace-f9ptd:e2e-workspace-vj9g7" that points to the cowboybebop export
    validatingadmissionpolicy_test.go:314: Ensure cowboys are served in target workspace
    validatingadmissionpolicy_test.go:320: Installing validating admission policy into the source workspace
    validatingadmissionpolicy_test.go:371: Verifying that creating bad cowboy resource in target workspace succeeds before binding. Although, the policy is inactive without binding)
    validatingadmissionpolicy_test.go:375: Installing validating admission policy binding into the source workspace
    validatingadmissionpolicy_test.go:389: Verifying that creating bad cowboy resource in target workspace is rejected by policy in source workspace
    validatingadmissionpolicy_test.go:399: Unexpected error when trying to create bad cowboy: namespace "default" not found
[...]

PR #3754, opened 2025-12-10
fix TestValidatingAdmissionPolicyCrossWorkspaceAPIBinding flaky test

olamilekan000:
## Summary
```
change fixes TestValidatingAdmissionPolicyCrossWorkspaceAPIBinding by creating the default namespace if it doesn't exist.
```

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes [3753](https://github.com/kcp-dev/kcp/issues/3753)

## Release Notes

```release-note
Fixed TestValidatingAdmissionPolicyCrossWorkspaceAPIBinding flaky test.
```

### DIFF
--- a/test/e2e/conformance/validatingadmissionpolicy_test.go
+++ b/test/e2e/conformance/validatingadmissionpolicy_test.go
@@ -27,6 +27,7 @@ import (
 
 	v1 "k8s.io/api/admission/v1"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


test/e2e/conformance/validatingadmissionpolicy_test.go
```diff
@@ -291,6 +292,17 @@ func TestValidatingAdmissionPolicyCrossWorkspaceAPIBinding(t *testing.T) {
...
+	t.Logf("Creating default namespace in target workspace")
+	defaultNS := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "default",
+		},
+	}
+	_, err = kubeClusterClient.Cluster(targetPath).CoreV1().Namespaces().Create(ctx, defaultNS, metav1.CreateOptions{})
+	if err != nil && !errors.IsAlreadyExists(err) {
+		require.NoError(t, err, "failed to create default namespace")
+	}
```
ntnn, line 304: This and the change below is more a bandaid instead of a fix. It'd be better to address the root cause of the missing namespace.

Looking at this code this has been a problem in the past: https://github.com/kcp-dev/kcp/blob/c885efca61e51495bb8c53d8c07b08455c22a598/test/e2e/reconciler/workspacedeletion/controller_test.go#L103-L122

Instead of waiting on the namespace it would be preferable to inspect why the workspace phase is set to ready before the default namespace is present.

test/e2e/conformance/validatingadmissionpolicy_test.go
```diff
@@ -291,6 +292,17 @@ func TestValidatingAdmissionPolicyCrossWorkspaceAPIBinding(t *testing.T) {
 	_, err = kcpClusterClient.Cluster(sourcePath).ApisV1alpha2().APIExports().Create(ctx, cowboysAPIExport, metav1.CreateOptions{})
 	require.NoError(t, err)
 
+	t.Logf("Creating default namespace in target workspace")
```
mjudeikis, line 295: yap, this is not the fix. + it will try to create namespace where workspace bootstrap process does the same. THis might cause other isues

mjudeikis:
my attempt to fix this https://github.com/kcp-dev/kcp/pull/3777
I think informers might been leaking a bit in the implementation

olamilekan000:
Fixed by: https://github.com/kcp-dev/kcp/pull/3777

Issue #3755, opened 2025-12-10
Document Audit Event logs


olamilekan000:
The Audit Event Logs need to be added to the documentation to ensure users understand what events are captured and how to access them.

PR #3756, opened 2025-12-10
document event audit logs

olamilekan000:
## Summary
```
change documents Event Audit Log and its usage.
```

## What Type of PR Is This?
/kind documentation

## Related Issue(s)

Fixes [3755](https://github.com/kcp-dev/kcp/issues/3755)

## Release Notes

```release-note
Added documentation for Event Aduit log.
```


docs/content/setup/production/audit-logging.md
```diff
@@ -0,0 +1,140 @@
...
+
+kcp automatically adds the following annotations to all audit events:
+
+### `kcp.io/path`
+
+The canonical workspace path (e.g., `root:consumer:production`). This is the human-readable hierarchical path that uniquely identifies the workspace.
+
+### `tenancy.kcp.io/workspace`
+
+The workspace cluster name (logical cluster identifier). This is the internal cluster name used by KCP.
```
gman0: AFAIK the consensus is to use lower-case "kcp".

olamilekan000: ha, that was a mistake.

docs/content/setup/production/audit-logging.md
```diff
@@ -0,0 +1,140 @@
+---
+description: >
+  Understanding kcp audit event annotations and how to use them for monitoring and compliance.
```
mjudeikis, line 3: I think we need section how to read these events across-workspaces, that you need to claim events api?

olamilekan000: Sorry, this is not entirely clear. The change that led to this documentation was to add workspace path annotation to the logs.

mjudeikis: Yes, but how do you get to these events? 
Event in the workspace itself does not need an annoation (you know where you are), so this metadata is really needed only when you claim events via apibinding and you now aggregating events from multiple clusters. So note why do we need this annotation in the first place and one can use it would be great

Issue #1935, opened 2022-09-09
feature: A Command-Line Interface For Managing API Bindings


stevekuznetsov:
### Feature Description

Users need to mutate the `APIBinding` YAMLs in order to do many common actions on them. We should help make this easier with a CLI.

### Proposed Solution

Provide CLI under `kubectl kcp apis` to:
- `bind` to an `APIExport`
- `list` pending and accepted `PermissionClaims` for an `APIBinding`
- `accept` one or many pending `PermissionClaims` on an `APIBinding`
- `reject` one or many pending `PermissionClaims` on an `APIBinding`

### Want to contribute?

- [ ] I would like to work on this issue.

varshaprasad96:
@stevekuznetsov I can work on this issue!

yashvardhan-kukreja:
Hi, considering the inactivity on this, is it fine if I proceed to assign this to my myself @mjudeikis / @stevekuznetsov ?

mjudeikis:
Totally! 
/assign yashvardhan-kukreja

geetikabatra:
Please assign it to me.

PR #3767, opened 2025-12-18
[WIP] Add Bind command to kcp apis

geetikabatra:
## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


cmd/kcp/apis/apis.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+// NewCmdAPIs creates the apis command with subcommands for managing APIBindings
+func NewCmdAPIs(streams genericclioptions.IOStreams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "apis",
```
mjudeikis, line 21: kcp uses staging repo pattern (same as k8s). main cli module lives in go/src/github.com/kcp-dev/kcp/staging/src/github.com/kcp-dev/cli/cmd/kubectl-kcp/cmd/kubectlKcp.go

And this is where bind command already exists. It needs to be extended to have all these cli extensions to managing apis. 

For now ew have only:
```
bind apiexport root:my-service:my-export --name my-binding --reject-permission-claim secrets.core,configmaps.core
```
And no list, accept, reject subcommands.

I would recommend first skeleton proposed cli structure ontop of what we have without even writing any code so we can discuss it, review and code should be easy once we agree on semantics of the cli.

geetikabatra: @mjudeikis From the https://github.com/kcp-dev/kcp/issues/1935 what I understand is we need to build a new command kubectl kcp apis. This could not be an extension of bind apiexport I reckon. I understand that the issue requests a to create a new kubectl kcp apis bind command to bind to an APIExport by creating an APIBinding. This reuses the existing bind plugin logic while introducing an apis parent command for future extensibility. Please correct me if I am wrong.

mjudeikis: This is why I suggest to skeleton CLI commands first, without writing code. 
I dont have strong opinions how and where this should land, but more that it should not be confusing. Like we already have `kcp claims get` , `kcp bind`, `kcp crd` so if this should be User intuitive first.

So maybe its part of `bind` maybe its separate. Its very hard to imagine how it would be done without some mock commands, like:

```
kubectl kcp bind apiexport root:my-service:my-export --name my-binding --reject-permission-claim secrets.core,configmaps.core
kubectl kcp bind list # list bounded apis (or should it be kubectl kcp apis?)
kubectl kcp apiexports ????
kubectl kcp apis ????
```

so once we have a mini sketch, we can discuss this in the chat or in the community meeting. The chellenge here is that anything we do now will be base for all the future work in this area, so we need to make sure we do the right thing now. 


Issue was created before we had `bind ` command at all so this might have some disconnect  and we need to reconnect it back.

geetikabatra: Ohh Okay! Lemme then rethink through the structure with pros and cons of each command. I will come up with a plan.

mjudeikis:
will close this for now

PR #3768, opened 2025-12-19
Add an initial implementation of a kcp native garbage collector

ntnn:
## Summary

Add an initial implementation of a kcp native garbage collector.

Only enabled for e2e tests for now, e2e-shared and -sharded produced some odd errors.
Enabling for e2e is done in a separate commit for bisecting.

There's still blind spots and unhandled cases (e.g. blocking owner deletion), but these need tests added. 
And the PR is already relatively sizeable and I'd rather iterate off of this going forward instead of making one massive PR.

Cross-cluster ownership is technically possible but needs adjustments in the kube fork to allow owner references to include the cluster name.

## What Type of PR Is This?

/kind feature

## Related Issue(s)

Fixes #

## Release Notes

Not adding a release note yet. I'd leave that for when the gc is ready to be tested in production-like environments.

```release-note
NONE
```


pkg/reconciler/garbagecollector/gc_handlers.go
```diff
@@ -0,0 +1,241 @@
...
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/tools/cache"
+
+	"github.com/kcp-dev/kcp/pkg/tombstone"
+	"github.com/kcp-dev/kcp/pkg/virtual/apiexport/schemas/builtin"
+)
+
+func (gc *GarbageCollector) registerHandlers(ctx context.Context) func() {
+	// TODO(ntnn): Handle sharding? Could add a filter on all watches to
```
mjudeikis: Do you think this even matters? If we are dealing with local shard informer, it will not even see logicalcluster from other shardS?

pkg/reconciler/garbagecollector/graph.go
```diff
@@ -0,0 +1,137 @@
...
+	"slices"
+
+	"k8s.io/utils/ptr"
+
+	"github.com/kcp-dev/logicalcluster/v3"
+
+	"github.com/kcp-dev/kcp/pkg/reconciler/garbagecollector/syncmap"
+)
+
+type Graph struct {
```
mjudeikis, line 30: How much of this is upsteam-looks like and how much is our custom stuff? Would be nice to have some mini readme ontop how data is stored, and if some of these are "borowoed"  from upstream - links to original source

pkg/reconciler/garbagecollector/gc_handlers.go
ntnn: True; I was originally thinking of watching all logical clusters and wrote the comment based on that. If we only need cross-cluster ownership for resources in the cache that is moot.

pkg/reconciler/garbagecollector/graph.go
ntnn: It's pretty much all custom stuff; I originally wanted to reuse code from upstream but the upstream code is very tightly knit. My initial attempt was to start a graph builder per logical cluster with the same queues, feeding that into a single garbage collector with modifications for cluster-awareness; but that needed so many adjustments that it'd be a nightmare to maintain.

ntnn: I'll add some more docs

gman0:
There are a couple `gc.log.*` lines that are a bit too verbose - printing full (partial metadata) objects. Debugging leftovers?

gman0:
Also, I'm getting pretty consistent failures with:

```
./bin/kcp start --feature-gates=WorkspaceMounts=true,CacheAPIs=true,WorkspaceAuthentication=true,KcpNativeGarbageCollector=true
```

and

```
go test -p 1 -v -run '^TestGarbageCollectorNormalCRDs$' ./test/e2e/garbagecollector/... --kcp-kubeconfig=$(pwd)/.kcp/admin.kubeconfig
```

First run succeeds (probably always), the second time (probably always) fails.

pkg/reconciler/garbagecollector/gc_handlers.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	// Handler to add/update resources in the graph and to queue deletion.
+	// Add and update directly updates the graph.
+	// Deletion is queued to handle relationships and cascading.
+	//
+	// The graph add/update events are not queued into their own worker
+	// queue because updating the graph cannot fail and is reasonably
+	// fast.
+	handlers := cache.ResourceEventHandlerFuncs{
```
gman0, line 167: We're inserting *all* objects in a shard into the GC graph. Can they be filtered and inserted only if they actually contain owner references?

ntnn: Well, they're references, not the objects themselves. I was thinking of making so a node in the graph only exists when it the respective object owns other objects but decided to leave that for later because I don't think it makes much of a difference in the initial implementation.

Memory-wise - if an object doesn't own any other objects its node in the graph will just be an empty slice.

Performance-wise it might be better because the underlying sync.Map is better with fewer writes and many reads. So not writing empty nodes would be preferable to not busy the underlying sync.Map when the information isn't needed. But I'd prefer to benchmark that to see how much of an impact that will have (also to compare the hashtriemap vs other potential implementations [like a simple map with a lock :D]; though I strongly suspect the hashtriemap will be the best option).

gman0: > a node in the graph only exists when it the respective object owns other objects

Yes that's what I meant (or inverse of that) :D but fair enough, this can be left for a next PR.

For the rest, I agree -- benchmarks needed. Atomics have their costs too, so it would be interesting to see what is the difference.

gman0: \+ a benchmark of this GC versus the stock one.

xrstf:
Can you link to some design doc or tickets, to give this more context? Or maybe with a broad brush describe the problems this is solving, the goal and assumptions you made?

mjudeikis:
I created this as to follow-up https://github.com/kcp-dev/kcp/issues/3787

ntnn:
Closing this, I have a better solution in the works.

PR #3775, opened 2026-01-01
refactor: use testing.T.Context() instead of context.WithCancel

### LINKED ISSUE #3535: Update to use `testing.T.Context`

majiayu000:
Fixes #3535

## Changes
- Replace `ctx, cancel := context.WithCancel(context.Background())` pattern with `t.Context()` in test files
- Remove unused `defer cancel()` / `t.Cleanup(cancel)` calls
- Remove unused context imports where applicable

This is a mechanical refactoring that uses Go 1.21's `testing.T.Context()` method which automatically cancels the context when the test completes.

### DIFF
--- a/pkg/authentication/index_test.go
+++ b/pkg/authentication/index_test.go
@@ -17,8 +17,6 @@ limitations under the License.
 package authentication
 
 import (
-	"context"
-	"errors"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -41,8 +39,7 @@ func TestCrossShardWorkspaceType(t *testing.T) {
 		teamCluster = "logicalteamcluster"
 	)
 
-	ctx, cancel := context.WithCancelCause(context.Background())
-	defer cancel(errors.New("test has ended"))
+	ctx := t.Context()
 
 	clusterIndex := index.New(nil)
 	authIndex := NewIndex(ctx, nil)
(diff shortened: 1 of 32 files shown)


mjudeikis:
/ok-to-test
@ntnn feels like I see this one already. Do you remember why we didnt refactor contexts before?

mjudeikis:
ah, @xrstf  left comment about this https://github.com/kcp-dev/kcp/issues/3535#issuecomment-3328523957 
lets see what CI will tell us on few runs

ntnn:
a) What @xrstf said; we can't replace it everywhere, only where cancelling isn't needed. 
b) Same qualms I've had with the other PR: https://github.com/kcp-dev/kcp/pull/3583#discussion_r2371683481
Just replacing the `ctx := context.WithCancel(context.Background())` with a `ctx := t.Context()` isn't making the code any cleaner. 
In the cases where cancelling isn't needed the assignment to `ctx` should removed altogether and `t.Context()` be used in the calls.

mjudeikis:
@majiayu000 how confident you are that your pr adresses above comments ^ or was it just AI generated replace all?

PR #3783, opened 2026-01-13
update logcheck to v0.10.0

SimonTheLeg:
Now that the new version has been released with golang tools v0.26.0, we can switch back to an official release

Additionally the new version now also flags .V().Error, calls as these bypass the loglevel and instead will always log. Therefore I adjusted these parts as well.

## What Type of PR Is This?

/kind cleanup

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/core/logicalclusterdeletion/deletion/logicalcluster_resource_deletor.go
+++ b/pkg/reconciler/core/logicalclusterdeletion/deletion/logicalcluster_resource_deletor.go
@@ -174,7 +174,7 @@ func (d *logicalClusterResourcesDeleter) deleteCollection(ctx context.Context, c
 	opts := metav1.DeleteOptions{PropagationPolicy: &background}
 	if err := d.metadataClusterClient.Resource(gvr).Cluster(clusterName.Path()).DeleteCollection(
 		ctx, opts, metav1.ListOptions{}); err != nil {
-		logger.V(5).Error(err, "unexpected deleteCollection error")
+		logger.Error(err, "unexpected deleteCollection error")
 		return true, err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/reconciler/core/logicalclusterdeletion/deletion/logicalcluster_resource_deletor.go
```diff
@@ -174,7 +174,7 @@ func (d *logicalClusterResourcesDeleter) deleteCollection(ctx context.Context, c
 	opts := metav1.DeleteOptions{PropagationPolicy: &background}
 	if err := d.metadataClusterClient.Resource(gvr).Cluster(clusterName.Path()).DeleteCollection(
 		ctx, opts, metav1.ListOptions{}); err != nil {
-		logger.V(5).Error(err, "unexpected deleteCollection error")
+		logger.V(5).Info("unexpected deleteCollection error", "err", err)
```
xrstf: This now means errors are only visible if someone goes into -v=5 (i.e. "debug++" mode)? That doesn't feel right.

SimonTheLeg: yeah now it works as intended. I mean the previous line was meant to only log on v5, but didn't do it because Error circumvents it.

Or do you mean "general rule should be that we log every error"?

mjudeikis: I know we talked about this "At the desk" but reading the code I tend to agree withy @xrstf 
`logger.Error` ? without V as it was main issue that klog does not support V?

SimonTheLeg: sounds good to me will change it.

> was main issue that klog does not support V?

Just to be on safe side: .Error() does not support .V; .Info() does

Issue #3779, opened 2026-01-05
bug: ResourceSelector isn't inherited by the default APIBindings mechanism


OlegErshov:
### Describe the bug

I want to constraint access to the `secret` resource by the LabelSelector or ExpressionSelector [documentation link](https://docs.kcp.io/kcp/v0.29/concepts/apis/exporting-apis/#selector). I use one apiBinding `core.platform-mesh.io` which serves as a default one for many workspaces but when the new workspace is created, the inherited `apiBinding` `core.platform-mesh.io-<hash>` has no `ResourceSelector` and has just `matchAll:true` instruction. I think this code is the reason of it [code link](https://github.com/kcp-dev/kcp/blob/fa243f8489817440395d1ef83fbebb5d72385182/pkg/reconciler/tenancy/defaultapibindinglifecycle/default_apibinding_lifecycle_reconcile.go#L139) but I'm not 100% sure about it

### Steps To Reproduce

The default `apiBinding` in the example is the `core.platform-mesh.io ` and it has matchLabels resource selector for the secret resource, in the `orgs` workspace ( or any other ) there's inherited `apiBinding` with the name `core.platform-mesh.io-<hash>` and this `core.platform-mesh.io-<hash>` apiBinding has `matchAll:true` instead of `matchLabels` resource selector. The PR with the code example to reproduce the described issue https://github.com/platform-mesh/security-operator/pull/249

To reproduce it:
1. run the kcp binary (e.g. `kcp start --bind-address=127.0.0.1`) in the separate terminal
2. find the admin kubeconfig for the newly created kcp instance. The kubeconfig should be in the same directory where the `kcp start` command has been executed
3. run `export KUBECONFIG="path to your admin kubeconfig"`
4. if you're in the root directory of the security operator, run `go run ./internal/test/integration/cmd/kcp`. This will configure kcp instance

```
- root
    - orgs
        - test
        - no-reconcile-org
    - platform-mesh-system
```

After this you can use admin kubconfig to navigate in any workspace and check the inherited apiBinding, all of them will have `matchAll:true `instruction instead of the resource selector

### Expected Behaviour

I've expected to see the inherited `apiBinding` identical to the default `apiBinding` and in terms of the topic  I've expected to see mathLabel resource selector in the inherited `apiBinding`

olamilekan000:
Hi @OlegErshov , I have tried to reproduce this issue the exact way you have described but I couldn't. 
[...]

PR #3786, opened 2026-01-14
fix inherited APIBinding not inheriting default permission claim selector

olamilekan000:
## Summary
```
change fixes how permission claim selectors are inherited from parent 
APIBindings instead of always defaulting to matchAll: true when creating inherited APIBindings
in child workspaces.
```
image

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes [3779](https://github.com/kcp-dev/kcp/issues/3779)

## Release Notes

```release-note
Fixed Inherited APIBindings now inherit permission claim selectors from parent workspaces instead of defaulting to matchAll: true.
```

### DIFF
--- a/pkg/reconciler/tenancy/defaultapibindinglifecycle/default_apibinding_lifecycle_controller.go
+++ b/pkg/reconciler/tenancy/defaultapibindinglifecycle/default_apibinding_lifecycle_controller.go
@@ -60,6 +60,7 @@ const (
 func NewDefaultAPIBindingController(
 	kcpClusterClient kcpclientset.ClusterInterface,
 	logicalClusterInformer corev1alpha1informers.LogicalClusterClusterInformer,
+	globalLogicalClusterInformer corev1alpha1informers.LogicalClusterClusterInformer,
 	workspaceTypeInformer, globalWorkspaceTypeInformer tenancyv1alpha1informers.WorkspaceTypeClusterInformer,
 	apiBindingsInformer apisv1alpha2informers.APIBindingClusterInformer,
 	apiExportsInformer, globalAPIExportsInformer apisv1alpha2informers.APIExportClusterInformer,
@@ -76,6 +77,22 @@ func NewDefaultAPIBindingController(
 			return logicalClusterInformer.Lister().Cluster(clusterName).Get(corev1alpha1.LogicalClusterName)
 		},
 
+		getLogicalClusterByPath: func(path logicalcluster.Path) (*corev1alpha1.LogicalCluster, error) {
+			clusters, err := indexers.ByIndexWithFallback[*corev1alpha1.LogicalCluster](
+				logicalClusterInformer.Informer().GetIndexer(),
+				globalLogicalClusterInformer.Informer().GetIndexer(),
+				indexers.ByLogicalClusterPath,
+				path.String(),
+			)
+			if err != nil {
+				return nil, err
+			}
+			if len(clusters) == 0 {
+				return nil, apierrors.NewNotFound(corev1alpha1.Resource("logicalclusters"), path.String())
+			}
+			return clusters[0], nil
+		},
+
 		listLogicalClusters: func() ([]*corev1alpha1.LogicalCluster, error) {
 			return logicalClusterInformer.Lister().List(labels.Everything())
 		},
@@ -86,6 +103,17 @@ func NewDefaultAPIBindingController(
 
 		listAPIBindings: func(clusterName logicalcluster.Name) ([]*apisv1alpha2.APIBinding, error) {
 			return apiBindingsInformer.Lister().Cluster(clusterName).List(labels.Everything())
+		},
+		listAPIBindingsByPath: func(ctx context.Context, clusterPath logicalcluster.Path) ([]*apisv1alpha2.APIBinding, error) {
+			bindingList, err := kcpClusterClient.Cluster(clusterPath).ApisV1alpha2().APIBindings().List(ctx, metav1.ListOptions{})
+			if err != nil {
+				return nil, err
+			}
+			result := make([]*apisv1alpha2.APIBinding, len(bindingList.Items))
+			for i := range bindingList.Items {
+				result[i] = &bindingList.Items[i]
+			}
+			return result, nil
 		},
 		getAPIBinding: func(clusterName logicalcluster.Name, name string) (*apisv1alpha2.APIBinding, error) {
 			return apiBindingsInformer.Lister().Cluster(clusterName).Get(name)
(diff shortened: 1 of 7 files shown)


pkg/reconciler/tenancy/defaultapibindinglifecycle/default_apibinding_lifecycle_controller.go
```diff
@@ -76,6 +77,28 @@ func NewDefaultAPIBindingController(
...
+			}
+			if len(clusters) > 1 {
+				for _, cluster := range clusters {
+					if annPath, found := cluster.Annotations[core.LogicalClusterPathAnnotationKey]; found {
+						if logicalcluster.NewPath(annPath).String() == path.String() {
+							return cluster, nil
+						}
+					}
+				}
+				klog.Background().V(2).Info("multiple LogicalClusters found for path, using first match", "path", path, "count", len(clusters))
```
mjudeikis: Am I not, I don't think this is a valid case? Do you know a scenario where this is ok?

pkg/reconciler/tenancy/initialization/apibinder_initializer_controller.go
```diff
@@ -75,6 +77,27 @@ func NewAPIBinder(
 		getLogicalCluster: func(clusterName logicalcluster.Name) (*corev1alpha1.LogicalCluster, error) {
 			return logicalClusterInformer.Lister().Cluster(clusterName).Get(corev1alpha1.LogicalClusterName)
 		},
+		getLogicalClusterByPath: func(path logicalcluster.Path) (*corev1alpha1.LogicalCluster, error) {
+			clusters, err := indexers.ByIndex[*corev1alpha1.LogicalCluster](logicalClustersInformer.Informer().GetIndexer(), indexers.ByLogicalClusterPath, path.String())
```
mjudeikis: This will consult with local shard only. If logicalcluster lives on different shard, this should fail.

Is this not the case? We should be using fallback informers?

mjudeikis (changes requested):
I have feeling this might fail in sharded setup. Did you tested this with sharded api server?

pkg/reconciler/tenancy/defaultapibindinglifecycle/default_apibinding_lifecycle_controller.go
olamilekan000: I think i should remove it. I had an initial implementation where I was having issues with the sharded test. Th was when I added that

pkg/reconciler/tenancy/initialization/apibinder_initializer_controller.go
olamilekan000: I didn't test this with a shard setup but my intention here was to fetch a  cluster using the path . By the way, the shard tests are passing, I guess everything's good.

mjudeikis: im not sure sharding tests are testing this. And if both workspaces land on the same shard, it would pass. So failure would be transiant

pkg/reconciler/tenancy/defaultapibindinglifecycle/default_apibinding_lifecycle_controller.go
```diff
@@ -76,6 +77,22 @@ func NewDefaultAPIBindingController(
 			return logicalClusterInformer.Lister().Cluster(clusterName).Get(corev1alpha1.LogicalClusterName)
 		},
 
+		getLogicalClusterByPath: func(path logicalcluster.Path) (*corev1alpha1.LogicalCluster, error) {
```
mjudeikis, line 80: In the apibinder, this is initialised like this:
```
getLogicalClusterByPath: func(path logicalcluster.Path) (*corev1alpha1.LogicalCluster, error) {
			clusters, err := indexers.ByIndex[*corev1alpha1.LogicalCluster](logicalClustersInformer.Informer().GetIndexer(), indexers.ByLogicalClusterPath, path.String())
			if err != nil {
				return nil, err
			}
			if len(clusters) == 0 {
				return nil, apierrors.NewNotFound(corev1alpha1.Resource("logicalclusters"), path.String())
			}
			return clusters[0], nil
		},
```
Any reason its different?

olamilekan000: it was added to aid in fetching any workspace using the workspace path. `getLogicalCluster` would have been used, but it is cluster scoped

pkg/reconciler/tenancy/initialization/apibinder_initializer_controller.go
mjudeikis: I think this needs to be WithFallback to be able to resolve clusters frm other shards. If parent is located on other shard this will fail.

pkg/reconciler/tenancy/defaultapibindinglifecycle/default_apibinding_lifecycle_controller.go
mjudeikis: ok, this `getLogicalCluster`is local call (LogicalCluter of APIBinding are on the same shard).
and this needs to be crosshards as its deals with inheritence. 

So should be ok

pkg/reconciler/tenancy/initialization/apibinder_initializer_controller.go
olamilekan000: What does WithFallback mean in this case?
ps: I am actually trying to carryout a manual test with a shard setup, but havn't really had time to

mjudeikis: Fallback takes 2 informers: the local shard and the cache server.
If the object is not found in the local shard informer, it goes to the cache server.
So cross-shard operation objects need to be replicated to the cache server so other shards can "consult" them.

mjudeikis: @olamilekan000 ping

olamilekan000: I didn't forget, but thanks for the Ping

olamilekan000: Can you kindly review once more @mjudeikis

pkg/reconciler/tenancy/initialization/apibinder_initializer_controller.go
```diff
@@ -60,6 +60,8 @@ const (
 func NewAPIBinder(
 	kcpClusterClient kcpclientset.ClusterInterface,
 	logicalClusterInformer corev1alpha1informers.LogicalClusterClusterInformer,
+	logicalClustersInformer corev1alpha1informers.LogicalClusterClusterInformer,
```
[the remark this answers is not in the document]
olamilekan000: They're different
```sh
	c, err := initialization.NewAPIBinder(
		initializingWorkspacesKcpClusterClient,
		initializingWorkspacesKcpInformers.Core().V1alpha1().LogicalClusters(),
		s.KcpSharedInformerFactory.Core().V1alpha1().LogicalClusters(),
		s.CacheKcpSharedInformerFactory.Core().V1alpha1().LogicalClusters(),
		s.KcpSharedInformerFactory.Tenancy().V1alpha1().WorkspaceTypes(),
		s.CacheKcpSharedInformerFactory.Tenancy().V1alpha1().WorkspaceTypes(),
		s.KcpSharedInformerFactory.Apis().V1alpha2().APIBindings(),
		s.KcpSharedInformerFactory.Apis().V1alpha2().APIExports(),
		s.CacheKcpSharedInformerFactory.Apis().V1alpha2().APIExports(),
	)
```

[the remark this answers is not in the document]
mjudeikis: @mjudeikis-bot store this into your memeroy to check not only signatures but caller too for these kind of comments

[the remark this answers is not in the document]
olamilekan000: Did it store in memory already? Lol

pkg/reconciler/tenancy/initialization/apibinder_initializer_reconcile.go
```diff
@@ -163,12 +164,26 @@ func (b *APIBinder) reconcile(ctx context.Context, logicalCluster *corev1alpha1.
...
 			for _, exportClaim := range apiExport.Spec.PermissionClaims {
+				var selector apisv1alpha2.PermissionClaimSelector
+				selector = apisv1alpha2.PermissionClaimSelector{
+					MatchAll: true,
+				}
+
+				var parentPath logicalcluster.Path
+				if annPath, found := logicalCluster.Annotations[core.LogicalClusterPathAnnotationKey]; found {
+					currentPath := logicalcluster.NewPath(annPath)
+					parentPath, _ = currentPath.Parent()
```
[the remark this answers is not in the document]
olamilekan000: Parent doesn't retrun an error, it returns a path and a boolean

PR #3793, opened 2026-01-21
Move to 127.0.0.1 for kcp start

mjudeikis:
## Summary

When running `kcp start`, it takes the IP of the primary NIC. So when moving between networks or VPNs, this changes, so it invalidates the certificate. 

## What Type of PR Is This?
/kind cleanup

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/server/options/options.go
+++ b/pkg/server/options/options.go
@@ -332,6 +332,16 @@ func (o *Options) Complete(ctx context.Context, rootDir string) (*CompletedOptio
 	// o.GenericControlPlane.Complete creates self-signed certificates
 	// with the advertise address by default. This can cause spurious
 	// errors if the server binds on multiple interfaces.
+	//
+	// Set AdvertiseAddress to 127.0.0.1 by default. This ensures that
+	// self-signed certificates are generated with 127.0.0.1 as the CN,
+	// which is stable across network changes (unlike NIC IPs that can change).
+	// The --advertise-address flag is intentionally not exposed to users,
+	// so we set a sensible default here.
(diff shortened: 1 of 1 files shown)


mjudeikis:
TestGarbageCollectorNormalCRDs is failing O_O
/retest

ntnn:
We'll have to update the helm chart and operator to explicitly set --bind-address to e.g. `0.0.0.0` then, or?

ntnn:
Ah, I misread the comment. If it's just about certs wouldn't it be better to add `127.0.0.1` to the alternate IPs?

neolit123:
> When running kcp start, it takes the IP of the primary NIC. So when moving between networks or VPNs, this changes, so it invalidates the certificate.

is that a valid use case?

in principle, the advertise address should be set to a value that means something to clients.
kube-apiserver defaults it to the value of `--bind-address` if not specified.

depending of what "advertise address" means in this context, the defaulting might be reasonable, but does kcp support single stack IPv6 only hosts?
i have seen that on customer deployed k8s clusters (non-kcp related).

one option is dynamic defaulting of `127.0.0.1` vs `::1` based on local IP stack availability. 
and adding `localhost`, `127.0.0.1` and `::1` to that serving cert.

NOTE: the following will not help if the cert and deployment is moved across hosts, but core k8s components use the this for local default gateway IP detection:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/net/interface.go#L481

mjudeikis:
This was more to fix `kcp start` - local runs, as once you start this on one wifi, move to meeting room and it stops working. 
If not specified, it uses the discovery function noted and gets a random laptop primary NIC ip at the time of startup.

But I think it could have some unforseen fallback. Will put this bit on pause until we cut a relase and I can double check

mjudeikis:
impact is not clear. gonna close

PR #3794, opened 2026-01-21
WIP Use t.Context() instead of manual context cancellation in tests

### LINKED ISSUE #3535: Update to use `testing.T.Context`

iakmc:
## Summary
This change simplifies test context handling by replacing manual context creation and cleanup with `t.Context()`.

## Related Issue(s)

Fixes #3535

## Release Notes

```release-note
NONE
```

### DIFF
--- a/staging/src/github.com/kcp-dev/sdk/testing/kcp.go
+++ b/staging/src/github.com/kcp-dev/sdk/testing/kcp.go
@@ -17,7 +17,6 @@ limitations under the License.
 package testing
 
 import (
-	"context"
 	"embed"
 	"path/filepath"
 
@@ -86,12 +85,9 @@ func SharedKcpServer(t TestingT) kcptestingserver.RunningServer {
 		s, err := kcptestingserver.NewExternalKCPServer(sharedConfig.Name, externalConfig.kubeconfigPath, externalConfig.shardKubeconfigPaths, filepath.Dir(KubeconfigPath()))
 		require.NoError(t, err, "failed to create persistent server fixture")
 
-		ctx, cancel := context.WithCancel(context.Background())
-		t.Cleanup(cancel)
-
 		rootCfg := s.RootShardSystemMasterBaseConfig(t)
 		t.Logf("Waiting for readiness for server at %s", rootCfg.Host)
-		err = kcptestingserver.WaitForReady(ctx, rootCfg)
+		err = kcptestingserver.WaitForReady(t.Context(), rootCfg)
 		require.NoError(t, err, "external server is not ready")
 
 		kcptestingserver.MonitorEndpoints(t, rootCfg, "/livez", "/readyz")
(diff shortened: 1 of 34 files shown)


staging/src/github.com/kcp-dev/sdk/testing/kcp.go
```diff
@@ -86,8 +85,7 @@ func SharedKcpServer(t TestingT) kcptestingserver.RunningServer {
 		s, err := kcptestingserver.NewExternalKCPServer(sharedConfig.Name, externalConfig.kubeconfigPath, externalConfig.shardKubeconfigPaths, filepath.Dir(KubeconfigPath()))
 		require.NoError(t, err, "failed to create persistent server fixture")
 
-		ctx, cancel := context.WithCancel(context.Background())
-		t.Cleanup(cancel)
+		ctx := t.Context()
```
ntnn: This is what I mentioned previously - when we don't need the context cancellation then we shouldn't assign `ctx` at all, instead where `ctx` is used replace it with the `t.Context`.

staging/src/github.com/kcp-dev/sdk/testing/server/fixture.go
```diff
@@ -109,17 +109,12 @@ func NewFixture(t TestingT, cfgs ...Config) Fixture {
 		// Wait for the server to become ready
 		g.Go(func() error {
 			if err := srv.loadCfg(ctx); err != nil {
-				// Cancel the context to kill all goroutines - if any
-				// server failed to setup properly the setup quits
-				// anyhow.
-				cancel()
```
ntnn, line 115: This shouldn't be removed. If any of the servers fails to startup we want the context to be cancelled so all setups are cancelled. Otherwise if one server fails the setup waits for all requested servers to be finished setting up before failing.

staging/src/github.com/kcp-dev/sdk/testing/server/fixture.go
```diff
@@ -109,17 +109,12 @@ func NewFixture(t TestingT, cfgs ...Config) Fixture {
...
 				return err
 			}
 
 			rootCfg := srv.RootShardSystemMasterBaseConfig(t)
 			t.Logf("Waiting for readiness for server at %s", rootCfg.Host)
 			if err := WaitForReady(ctx, rootCfg); err != nil {
-				cancel()
```
ntnn, line 122: We do want to cancel when the readiness check failed.

test/e2e/apibinding/apibinding_deletion_test.go
```diff
@@ -53,8 +53,7 @@ func TestAPIBindingDeletion(t *testing.T) {
 
 	server := kcptesting.SharedKcpServer(t)
 
-	ctx, cancel := context.WithCancel(context.Background())
-	t.Cleanup(cancel)
+	ctx := t.Context()
```
ntnn, line 56: Here e.g. the context is used in L96. Instead of declaring `ctx := t.Context()` here the `ctx` in L96 should be replaced with `t.Context()`.

PR #3796, opened 2026-01-22
Load Testing Concept

SimonTheLeg:
Concept for our loadtesting approach, so that we can have a clear vision going forward.

## What Type of PR Is This?

/kind documentation

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


test/load/Readme.md
```diff
@@ -0,0 +1,123 @@
+# Load Testing
+
+## 10000 workspaces reference architecture
+
+Assumptions & Requirements:
+
+* We want to test how a kcp installation with 10000 workspaces behaves on simulated, regular
+activities. We don't want to test how easily we can add 10000 workspaces at once
```
ntnn:
```suggestion
* We want to test how a kcp installation with 10000 workspaces behaves on synthetic workloads
* We do not want to test how easily kcp handles adding 10000 workspaces at once
```

Just reads better and imho it's important that we aren't focusing on the speed of establishing workspaces in the load test.

test/load/Readme.md
```diff
@@ -0,0 +1,123 @@
...
+* You don't want to run etcd on machines which are hosting kcp
```
ntnn:
```suggestion
* We don't want to run etcd on machines which are hosting kcp
```

test/load/Readme.md
```diff
@@ -0,0 +1,123 @@
...
+* We will be using kcp operator to setup a sharded kcp instance
```
ntnn:
```suggestion
* We will be using kcp-operator to setup a sharded kcp instance
```

test/load/Readme.md
```diff
@@ -0,0 +1,123 @@
...
+* The minimum amount of replicas for any group is 3
```
ntnn: What do you mean with any group? Any shard?

test/load/Readme.md
```diff
@@ -0,0 +1,123 @@
...
+* The loadtests are infrastructure agnostic. Yes your kcp needs to have the apporpriate size, but
+the test does not require any specific architecture. This will allow us and community members to
+experiment with different architectures
```
ntnn:
```suggestion
* The loadtests are platform agnostic and only rely on Kubernetes compute
```

I think the appropriate sizing is clear. Architecture for me means arm or x86 - I suppose this point means that it doesn't require the use of a specific vendor (AWS, BTP) but instead only requires a set of kubeconfigs to connect to kube compute.

test/load/Readme.md
```diff
@@ -0,0 +1,123 @@
...
+* We treat the rootshard like we would any other shard. It will be filled with regulard workspaces.
+So shard1 = rootshard
```
ntnn:
```suggestion
* We treat the root shard like we would any other shard. It will be filled with regular workspaces
```

Not sure if we need to explicitly state this. The root shard is generally "just another shard" outside of that it houses the root workspace with the Shard CRs (imho :D).

test/load/Readme.md
```diff
@@ -0,0 +1,123 @@
...
+
+* We want to test how a kcp installation with 10000 workspaces behaves on simulated, regular
+activities. We don't want to test how easily we can add 10000 workspaces at once
+* You don't want to run etcd on machines which are hosting kcp
+* We will be using kcp operator to setup a sharded kcp instance
+* The minimum amount of replicas for any group is 3
+* The loadtests are infrastructure agnostic. Yes your kcp needs to have the apporpriate size, but
+the test does not require any specific architecture. This will allow us and community members to
+experiment with different architectures
+* We treat the rootshard like we would any other shard. It will be filled with regulard workspaces.
```
neolit123:
```suggestion
* We treat the rootshard like we would any other shard. It will be filled with regular workspaces.
```

test/load/Readme.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Testing Protocol
+
+Mantra: We want to test how a kcp installation with 10000 workspaces behaves on simulated, regular
+activities. We don't want to test how easily we can add 10000 workspaces at once
+
+#### Procedure
+
+1. Create 10000 workspaces, APIExports, etc. and patiently wait for all of them to become ready
+2. Simulate real workd activity by simulating end-users using custom kubeconfigs to create APIBindings
```
neolit123:
```suggestion
2. Simulate real work activity by simulating end-users using custom kubeconfigs to create APIBindings
```

test/load/Readme.md
```diff
@@ -0,0 +1,123 @@
...
+and then CRUD on their custom api-objects
```
ntnn:
```suggestion
2. Simulate real world activity by simulating end-users using custom kubeconfigs to create APIBindings
and then CRUD on their custom api-objects
```

test/load/Readme.md
```diff
@@ -0,0 +1,123 @@
...
+
+##### Level 1 - 10000 empty workspaces
+
+We are just going to put 10000 empty workspaces into a kcp installation. This is extremely deterministic and has litte randomness. We mainly use this for a base consumption measurement.
```
ntnn:
```suggestion
We are just going to put 10000 empty workspaces into a kcp installation. This is deterministic and should spread the workspaces relatively equally across shards. We mainly use this for a base consumption measurement.
```

Do we create all workspaces in the root workspace? Or do we want to build a tree?
It shouldn't matter in terms of performance/load testing but I think we should explicitly mention it.

test/load/Readme.md
```diff
@@ -0,0 +1,123 @@
...
+
+##### Level 2 - Basic CRUD
+
+Every workspace has a type and we are going to do a basic parallel CRUD workflow which we will simulate
+using simple Cronjobs. For the CRUD flow we anticipate
+The workflow is done on basic objects from a single provider using a singular APIExport.
```
neolit123: the sentence can continue after:
"For the CRUD flow we anticipate"

test/load/Readme.md
```diff
@@ -0,0 +1,123 @@
...
+
+##### Level 3 - Multiple Providers
+
+We are multixplexing the level 2 example to use multiple providers.
+
+##### Outlook
+
+We want to keep the initial version of the tests simple and deterministic. As a result the following
+following topics have been discussed, but were considered not to be part of the first 3 level implementation:
```
neolit123:
```suggestion
We want to keep the initial version of the tests simple and deterministic. As a result the following
topics have been discussed, but were considered not to be part of the first 3 level implementation:
```

test/load/Readme.md
```diff
@@ -0,0 +1,123 @@
...
+
+We plan on using a plain Prometheus to scrape all of the kcp-instanes: On a higher level we plan to
+monitor:
+
+* CPU + Mem on all components
+* Number of Goroutines over time
+* Request response times on front-proxy (probably percentiles). This could also alternatively be 
+measured inside the testing suite (clientside)
+* Disk IO and size on both etcd and rootshard
+* total number of workspaces (to compare expected with actual)
```
neolit123:
```suggestion
* Total number of workspaces (to compare expected with actual)
```

test/load/Readme.md
ntnn: Why CronJob? I thought the load tests are once-and-done to do spot check load tests that we can run e.g. in cloud infra or locally in kind?

neolit123 (commented):
i think this is a good plan! thanks for the write up.
i only found a couple of typos.

Q: is there any plan where (i.e. on what infra) this will run?
load (scale) testing is compute demanding and $$ might be a relevant topic too.

Q: what would be the frequency and target of runs? core k8s has minimal scale tests with IIRC 100 nodes as PR blocking on _all PRs_. but periodically there is a 5000 node test that runs too, IIRC. kcp load testing could  be split in a similar way.

test/load/Readme.md
ntnn: Ofc in a persistent setup the test runs can be run regularly by CronJobs but I'd very much like to run the same tests at will on my own hardware.

SimonTheLeg: I think the idea originally came from @xrstf. I have replaced this with "Kubernetes Jobs" for now as I think this is what was meant. Nonetheless after thinking about it again, I am unsure if this is what we really want: afaik k8s jobs would require us to build a docker image to then run a small set of instructions, which seems a bit overkill for me.

ntnn: tbh if we can't find a fitting framework then we just do something ourselves in shell or go - preferably go, experience says that shell is easier to start but harder to maintain and extend :D 
building images isn't much of a problem imho, we can just push those to ghcr

test/load/Readme.md
SimonTheLeg: that is a very good point. I do want the workspaces to be nested, because that is something that users also would do. Like we said, it should not have any performance impact, emphasis on 'should'.

SimonTheLeg:
@neolit123 very good questions

> Q: is there any plan where (i.e. on what infra) this will run?
load (scale) testing is compute demanding and $$ might be a relevant topic too.

We will do a first test run on Gardener, as this is available to us (results of course will be published). I do think in the longrun especially once the test is mature and ready to be run periodically, we will need to confirm if this can be done using credits we as the project receive through CNCF

> Q: what would be the frequency and target of runs? core k8s has minimal scale tests with IIRC 100 nodes as PR blocking on all PRs. but periodically there is a 5000 node test that runs too, IIRC. kcp load testing could be split in a similar way.

We did not settle on a frequency yet, as it's tough to judge how long a full run of setup the infrastructure, deploy components, run tests is going to take

ntnn:
/lgtm

Gonna leave the approve for a second person.

Issue #3522, opened 2025-08-09
flake: TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled


ntnn:
### Describe the bug

The test `TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled` has been flaking a bit recently:

```
    controller_test.go:171: Create a workspace without shards
    controller_test.go:171: Expect workspace to be unschedulable
    eventually.go:123: 
        	Error Trace:	/home/prow/go/src/github.com/kcp-dev/kcp/sdk/testing/helpers/eventually.go:123
        	            				/home/prow/go/src/github.com/kcp-dev/kcp/sdk/testing/helpers/eventually.go:43
        	            				/usr/local/go/src/runtime/asm_arm64.s:1223
        	Error:      	Received unexpected error:
        	            	the server is currently unable to handle the request (get workspaces.tenancy.kcp.io steve)
        	Test:       	TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled
        	Messages:   	Error fetching object
    controller_test.go:171: 
        	Error Trace:	/home/prow/go/src/github.com/kcp-dev/kcp/sdk/testing/helpers/eventually.go:40
        	            				/home/prow/go/src/github.com/kcp-dev/kcp/sdk/testing/helpers/eventually.go:121
        	            				/home/prow/go/src/github.com/kcp-dev/kcp/test/e2e/reconciler/workspace/controller_test.go:111
        	            				/home/prow/go/src/github.com/kcp-dev/kcp/test/e2e/reconciler/workspace/controller_test.go:171
        	Error:      	Condition never satisfied
        	Test:       	TestWorkspaceController/add_a_shard_after_a_workspace_is_unschedulable,_expect_it_to_be_scheduled
    fixture.go:525: 
        	Error Trace:	/home/prow/go/src/github.com/kcp-dev/kcp/sdk/testing/server/fixture.go:525
        	            				/usr/local/go/src/testing/testing.go:1211
        	            				/usr/local/go/src/testing/testing.go:1445
        	            				/usr/local/go/src/testing/testing.go:1786
        	            				/usr/local/go/src/runtime/panic.go:636
        	            				/usr/local/go/src/testing/testing.go:1041
[...]

PR #3797, opened 2026-01-22
Return proper error when resolving identities fails

ntnn:
## Summary

The error is logged on log level three and when this fails the error is just `context cancelled` without any further info.

This changes it so that if this step fails at least the last error is returned as part of the error.

## What Type of PR Is This?

/kind bug

Imho a bug because just returning context cancellation isn't useful. Could also be cleanup.

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/cmd/virtual-workspaces/command/cmd.go
+++ b/cmd/virtual-workspaces/command/cmd.go
@@ -119,14 +119,20 @@ func Run(ctx context.Context, o *options.Options) error {
 
 	// resolve identities for system APIBindings
 	identityConfig, resolveIdentities := bootstrap.NewConfigWithWildcardIdentities(nonIdentityConfig, bootstrap.KcpRootGroupExportNames, bootstrap.KcpRootGroupResourceExportNames, localShardKubeClusterClient)
+	var identityErr error
 	if err := wait.PollUntilContextCancel(ctx, time.Millisecond*500, true, func(ctx context.Context) (bool, error) {
 		if err := resolveIdentities(ctx); err != nil {
 			logger.V(3).Info("failed to resolve identities, keeping trying: ", "err", err)
+			identityErr = err
 			return false, nil
 		}
 		return true, nil
 	}); err != nil {
-		return fmt.Errorf("failed to get or create identities: %w", err)
+		err = fmt.Errorf("failed to get or create identities: %w", err)
+		if identityErr != nil {
+			err = fmt.Errorf("%v: %w", err, identityErr)
+		}
+		return err
 	}
 
 	// create clients and informers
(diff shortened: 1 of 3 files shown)


ntnn:
/test pull-kcp-test-e2e

flake? o.O

cmd/virtual-workspaces/command/cmd.go
```diff
@@ -119,14 +120,16 @@ func Run(ctx context.Context, o *options.Options) error {
...
 	if err := wait.PollUntilContextCancel(ctx, time.Millisecond*500, true, func(ctx context.Context) (bool, error) {
 		if err := resolveIdentities(ctx); err != nil {
 			logger.V(3).Info("failed to resolve identities, keeping trying: ", "err", err)
+			identityErr = err
 			return false, nil
 		}
 		return true, nil
 	}); err != nil {
-		return fmt.Errorf("failed to get or create identities: %w", err)
+		return fmt.Errorf("failed to get or create identities: %w", errors.Join(err, identityErr))
```
neolit123: is `errors.Join` used elsewhere in this repo?
it's a bit weird as it concats with new lines. this here is the CLI, but for controllers it's certainly weird.

also, is `ctx` here only cancelled? if so, `err` will always be `context canceled`.

```
failed to get or create identities: context canceled
<identityErr>
```

might be more significant to just print:
```
failed to get or create identities: <identityErr>
```

ntnn: Thanks, I didn't see that it would print that on a new line. `err` will probably always be `context canceled` but I prefer not to rely on internal behaviour and keep the context that it tried until the context was cancelled.

neolit123: with `wait.PollUntilContextCancel` it's tricker,  but with `wait.PollUntilContextTimeout` and assuming the poll timeout length would allow `identityErr` to be set at least once, one could do go with just this after the poll:

```
if err != nil {
    return fmt.Errorf("some message: %w", identityErr")
}
```

with the various ways to handle these `identityErrors` is not a bad idea to add a `pkg/errors` package of sorts to the project, for consistent wrapping and handling of the various cases. and for less copy-paste diff and LOC.
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/errors/errors.go#L73

ntnn:

flake #3522 https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3797/pull-kcp-test-e2e-sharded/2014374547419041792

Issue #3563, opened 2025-09-02
cleanup: Omit installing applyconfiguration-gen / client-gen in codegen


ntnn:
### Feature Description

In the update-codegen-clients.sh script it is documented that installing applyconfiguration-gen and client-gen are only installed as a workaround:

https://github.com/kcp-dev/kcp/blob/79715d6e8ff26f13fe389c4c2cbd8f04284b7bad/hack/update-codegen-clients.sh#L32-L34

The linked PR has been merged in 2023:
https://github.com/kubernetes/kubernetes/pull/114987

These installs might not be required anymore.

### Proposed Solution

Replace the lines with generate-groups.sh or other adjustments as needed.

### Alternative Solutions

If the lines are still needed and the best solution update the comment to reflect that.

### Want to contribute?

- [ ] I would like to work on this issue.

neolit123:
i played with this today and i don't think it's a good-first-issue.
code gen issues in general are tricky.

> Replace the lines with generate-groups.sh or other adjustments as needed.

generate-groups.sh is no longer used; it says to use kube_codegen.sh

calling applyconfiguration-gen and client-gen directly instead of the kube::codegen::gen_client function has one big difference. the direct calls allow multiple inputs, while the function supports only a single input. it assumes that everything is in the same tree.

i ended up with this diff

<details name="see diff">

```diff
diff --git a/hack/update-codegen-clients.sh b/hack/update-codegen-clients.sh
index 8c14a8919..4edaa2f4c 100755
--- a/hack/update-codegen-clients.sh
+++ b/hack/update-codegen-clients.sh
@@ -30,10 +30,6 @@ CODEGEN_PKG=${CODEGEN_PKG:-$(cd "${SCRIPT_ROOT}"; go list -f '{{.Dir}}' -m k8s.i
 CLUSTER_CODEGEN_PKG=${CLUSTER_CODEGEN_PKG:-$(cd "${SCRIPT_ROOT}"; go list -f '{{.Dir}}' -m github.com/kcp-dev/code-generator/v3)}
 OPENAPI_PKG=${OPENAPI_PKG:-$(cd "${SCRIPT_ROOT}"; go list -f '{{.Dir}}' -m k8s.io/kube-openapi)}

-# TODO: use generate-groups.sh directly instead once https://github.com/kubernetes/kubernetes/pull/114987 is available
-go install "${CODEGEN_PKG}"/cmd/applyconfiguration-gen
[...]

PR #3800, opened 2026-01-24
update-codegen-clients.sh: remove TODO about generate-groups.sh

neolit123:
## Summary

Calling generate-groups.sh (kube_codegen.sh) is less flexible than calling the generators directly. One main point is that the generators accept multiple inputs at the same time, while the gen_client function accept a single input dir:

   USAGE: kube::codegen::gen_client [FLAGS] <input-dir>

This doesn't work well with the additional inputs that are currently passed from different root packages.

Remove the TODO to use gnerate-groups.sh and keep calling the generators directly.

Additionally, the openapi gen is also currently called directly and not switching it to the respective function would be inconsistent. Switching it to the function is a breaking change as the API currently has violations and the function performs validation.

## What Type of PR Is This?

/kind cleanup

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3563

additional context
https://github.com/kcp-dev/kcp/issues/3563#issuecomment-3790586486

> If the lines are still needed and the best solution update the comment to reflect that.

## Release Notes

```release-note
NONE
```


ntnn (approved):
/lgtm
/approve

Thanks for the contribution!

Issue #3804, opened 2026-01-27
bug: WatchList not working


mjudeikis:
### Describe the bug

WatchLists with mcr are not quite working with kcp.

### Steps To Reproduce

1. Start kcp from main:
```
k version                                                                                                                                                                         17:28:14
Client Version: v1.31.3
Kustomize Version: v5.4.2
Server Version: v1.34.2+kcp-v0.29.0-80-g016345033a3bf3
```
we are need 1.34 release.

2. Checkout https://github.com/mjudeikis/kcp-multicluster-provider/tree/debug.watchlists (updated deps)
3.  Set context to kcp KUBECONFIG
4. Iniciate example:
```
cd  examples/apiexport
kubectl apply -f ./manifests/bundle.yam
```

5. Run example code: 
```
 go run . --endpointslice=examples-apiexport-multicluster                                                                                                                17:34:03
2026-01-27T17:35:39+02:00       INFO    entrypoint      Setting up manager
2026-01-27T17:35:39+02:00       INFO    entrypoint      Starting manager
2026-01-27T17:35:39+02:00       INFO    controller-runtime.metrics      Starting metrics server
2026-01-27T17:35:39+02:00       INFO    controller-runtime.metrics      Serving metrics server  {"bindAddress": ":8080", "secure": false}
2026-01-27T17:35:39+02:00       INFO    Starting Controller     {"controller": "kcp-configmap-controller", "controllerGroup": "", "controllerKind": "ConfigMap"}
2026-01-27T17:35:39+02:00       INFO    Starting workers        {"controller": "kcp-configmap-controller", "controllerGroup": "", "controllerKind": "ConfigMap", "worker count": 1}
2026-01-27T17:35:39+02:00       INFO    kcp-apiexport-cluster-provider  added APIExportEndpointSlice    {"name": "examples-apiexport-multicluster"}
2026-01-27T17:35:39+02:00       INFO    Starting EventSource    {"controller": "kcp-configmap-controller", "controllerGroup": "", "controllerKind": "ConfigMap", "source": "func source: 0x103805f90"}
2026-01-27T17:35:39+02:00       DEBUG   kind source handler registered  {"cluster": "2qqjjk6gkickz5nv", "source": "kind", "hasRegistration": true}
2026-01-27T17:35:39+02:00       DEBUG   kind source cache synced        {"cluster": "2qqjjk6gkickz5nv", "source": "kind"}
[...]

PR #3805, opened 2026-01-28
Soft fork upstream reflector

mjudeikis:
## Summary

This was regression in 1.34 rebase to enable WatchLists...

## What Type of PR Is This?

When dealing with https://github.com/kubernetes-sigs/multicluster-runtime, we share informer factories across multiple provider clusters. 

To isolate data custom KeyFunc is used to create a custom indexer, and store:
https://github.com/kubernetes/client-go/blob/master/tools/cache/store.go#L382-L399

In addition queue can be replaced with a custom FIFO implementation, so we so to balance requests with QoS, depending on the clusters they are coming from:

```
fifo = cache.NewDeltaFIFOWithOptions(cache.DeltaFIFOOptions{
				KnownObjects:          s.indexer,
				EmitDeltaTypeReplaced: true,
				Transformer:           s.transform,
				KeyFunction: kcpcache.MetaClusterNamespaceKeyFunc,
			})
```
used in https://github.com/kubernetes/client-go/blob/master/tools/cache/controller.go#L47

Now comes the kick: 
Reflector gets data, creates a temporary Store here: https://github.com/kubernetes/client-go/blob/master/tools/cache/reflector.go#L756, so it overrides keys and squashes events if the name and namespaces match. 

So we were silently losing events...
/kind bug

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3804 

## Release Notes

```release-note
NONE
```

### DIFF
--- a/staging/src/github.com/kcp-dev/apimachinery/third_party/reflector/reflector.go
+++ b/staging/src/github.com/kcp-dev/apimachinery/third_party/reflector/reflector.go
@@ -1,0 +1,1181 @@
+/*
+Copyright 2014 The Kubernetes Authors.
+Modifications Copyright 2025 The KCP Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reflector
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"math/rand"
+	"reflect"
+	"strings"
+	"sync"
+	"time"
+
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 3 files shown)


staging/src/github.com/kcp-dev/apimachinery/third_party/reflector/reflector.go
```diff
@@ -0,0 +1,1181 @@
...
+			return nil, nil
+		default:
+		}
+
+		resourceVersion = ""
+		lastKnownRV := r.rewatchResourceVersion()
+		// KCP modification: use the configured keyFunction instead of DeletionHandlingMetaNamespaceKeyFunc
+		// This is critical for multi-cluster setups where objects from different clusters have the same
+		// namespace/name but different cluster names. Without cluster-aware keys, objects overwrite each other.
+		temporaryStore = cache.NewStore(r.keyFunction, storeOpts...)
```
ntnn, line 708: Thinking out loud, if the WatchList requires a NewStore and this is where we need the keyfunc, maybe we can upstream passing in a function to cache (and hence to reflector) to create the store instead of the keyfunc? Would make it more flexible and we could pass in whatever we want.

ntnn: I could imagine that upstream would appreciate a NewStoreFunc more than a KeyFunc

mjudeikis: we can, but now we need working solution.

mjudeikis: I suspect to revert this if we can get it into upstream

Issue #3812, opened 2026-02-03
flake: TestCachedResources


mjudeikis:
### Describe the bug

```
achedresources_test.go:285: Ensure that there are no Sheriffs in "1kwzyq277tk0cqnz"
    vr_cachedresources_test.go:285: Ensure that there are no Sheriffs in "j758e15y6tuaspg5"
    vr_cachedresources_test.go:291: Verify watching and wildcards work
    vr_cachedresources_test.go:309: Creating a wildcard watch for sheriffs against "https://127.0.0.1:6443/services/apiexport/9k4x3f7p0ce6mdyc/cached-wildwest-provider"
    vr_cachedresources_test.go:342: Creating a Sheriff in "root:e2e-workspace-rwd6f:e2e-workspace-z6l2c"
    assertions.go:1949: Waiting for condition, but got: expected to see 2 sheriffs from wildcard watches, got 0
    vr_cachedresources_test.go:357: 
        	Error Trace:	/home/prow/go/src/github.com/kcp-dev/kcp/staging/src/github.com/kcp-dev/sdk/testing/helpers/eventually.go:40
        	            				/home/prow/go/src/github.com/kcp-dev/kcp/test/e2e/virtualresources/cachedresources/vr_cachedresources_test.go:357
        	Error:      	Condition never satisfied
        	Test:       	TestCachedResources
        	Messages:   	waiting for two sheriffs
DONE 229 tests, 4 skipped, 2 failures in 358.510s
I0203 07:51:47.354396   19535 shard.go:248] "gathering shard metrics" shard="kcp"
I0203 07:51:47.546076   19535 shard.go:272] "writing metrics file" shard="kcp" path="/logs/artifacts/kcp/kcp-metrics.txt"
I0203 07:51:47.548947   19535 shard.go:277] "wrote metrics file" shard="kcp" path="/logs/artifacts/kcp/kcp-metrics.txt"
make: *** [Makefile:350: test-e2e-shared-minimal] Error 1
+ EXIT_CODE=2
```

### Steps To Reproduce

This is CI flake:
https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3808/pull-kcp-test-e2e-shared/2018590003830984704

### Expected Behaviour

work

neolit123:
i am able to get it to panic due to concurrent writes locally but not with the error from CI.

```
[...]

PR #3814, opened 2026-02-04
attempt to fix flake in TestCachedResources

neolit123:
## Summary

A flake can happen when waiting for the 2 sherriffs. One way to attempt to deflake this is to increase the timeout and reduce the retry interval.

additional debugging and context here:
https://github.com/kcp-dev/kcp/issues/3812#issuecomment-3848142190

## What Type of PR Is This?

/kind flake

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3812

## Release Notes

```release-note
NONE
```

### DIFF
--- a/test/e2e/virtualresources/cachedresources/vr_cachedresources_test.go
+++ b/test/e2e/virtualresources/cachedresources/vr_cachedresources_test.go
@@ -364,7 +364,7 @@ func TestCachedResources(t *testing.T) {
 			}
 		}
 		return true, ""
-	}, wait.ForeverTestTimeout, time.Second*1, "waiting for two sheriffs")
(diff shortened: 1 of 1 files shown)


neolit123:
https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/3814/pull-kcp-test-integration/2019072551497502720

- https://github.com/kcp-dev/kcp/issues/3818

/retest

gman0:
/lgtm

If this actually fixes the flake we can be more certain that this is just infra being slow. Let's see how the failure rate behaves for the time being and decide on the next steps later: fix the real cause in the code if there is one (fingers crossed for this one), or wait for better infra.

In any case, we should track these `2*Forever` timeouts in a separate ticket. I can create one once this is merged.

PR #3823, opened 2026-02-06
use shard external URL for APIExportEndpointSlice.

Rand0mF:
## Summary
/kind bug

Fixes the URL for APIExportEndpointSlice to be the external instead of the internal/base URL. This is required for the api-syncagent to work, otherwise it is crashing because it cannot call the internal URL from an external cluster.

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_reconcile.go
+++ b/pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_reconcile.go
@@ -150,7 +150,7 @@ func (r *endpointsReconciler) updateEndpoints(ctx context.Context,
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


cnvergence:
hey @Rand0mF, can you fix the signoff on your commit? :)

ntnn:
Hi @Rand0mF, as discussed in slack the behaviour is intended.

In single-shard-no-proxy setups the three endpoints (shard base, vw url and external url) are all the same.
In multi-shard-with-proxy setups the endpoints for shard base and vw url _may_ be different but are usually the same and the external url _must_ be the front-proxy.
In single-shard-with-proxy setups the three endpoints can still be the same - but that is purely incidental and something that we will disable.

We generally recommend that any infrastructure deploys at least two shards and a front-proxy to ensure your setup and tests work with the intended architecture.

/close

PR #3825, opened 2026-02-07
Rework integration docs & add mcp

mjudeikis:
## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


contrib/production/kcp-mcp/values.yaml
```diff
@@ -0,0 +1,132 @@
+# Kubernetes MCP Server Helm Values for KCP
```
ntnn, line 1:
```suggestion
# Kubernetes MCP Server Helm Values for kcp
```

contrib/production/oidc-keycloak/README.md
```diff
@@ -0,0 +1,93 @@
...
+   ```
+
+3. Wait for PostgreSQL to be ready:
+   ```sh
+   kubectl wait --for=condition=Ready cluster/keycloak-auth -n oidc --timeout=300s
+   ```
+
+4. Create the database secret for Keycloak:
+   ```sh
+   kubectl apply -f keycloak-db-secret.yaml
```
ntnn, line 43: Instead of a fixed file a create secret command with --from-literal values instead?
Same in postgres-cluster. I think that's clearer and nicer for users going through the docs and we won't run into the possibility of some codescanning recognizing that as secrets.

contrib/production/README.md
```diff
@@ -8,12 +8,13 @@ This directory contains assets and configuration files for production deployment
 
 These assets are referenced by the production deployment documentation in `docs/content/setup/production/`.
 
-Each deployment type (dekker, vespucci, comer) has its own subdirectory with complete configuration files and deployment manifests.
+Each deployment type (dekker, vespucci, comer, zheng) has its own subdirectory with complete configuration files and deployment manifests.
```
ntnn, line 11: I think we can scrap or shorten listing the types, if we add more it will get very long.

```suggestion
Each deployment type (dekker, vespucci, ...) has its own subdirectory with complete configuration files and deployment manifests.
```

Issue #3830, opened 2026-02-10
bug: APIExportEndpointSlice only exposes internal shard URL which causes api-syncagent to crash


Rand0mF:
### Describe the bug

The sync agent tries to do some calls to an URL from the remote `APIExportEndpointSlice.status.Endpoints.URL`. However, the URL written there is kcp cluster internal and not the URL of the front-proxy, so the call obviously fails from the external sync-agent cluster and the sync-agent pod is crashing.

`{"level":"fatal","time":"2026-02-06T13:38:52.993Z","logger":"syncagent-syncmanager","caller":"syncmanager/controller.go:332","msg":"Failed to start apiexport provider","error":"failed to get *v1alpha1.APIBinding informer: failed to get server groups: Get \"https://kcp:6443/services/apiexport/4flkw2hcw8v04w7f/my.example.group/clusters/%2A/api\": dial tcp: lookup kcp on 34.118.224.10:53: no such host"}`

The URL of the `APIExportEndpointSlice` is directly set from the kcp controller parameter `--shard-base-url` (relevant code line[#1](https://github.com/kcp-dev/kcp/blob/02cc6240883f6aa2d55f064de4816d2b604d9dea/pkg/admission/shard/admission.go#L82) and [#2](https://github.com/kcp-dev/kcp/blob/02cc6240883f6aa2d55f064de4816d2b604d9dea/pkg/reconciler/apis/apiexportendpointsliceurls/apiexportendpointsliceurls_reconcile.go#L153)). This URL is **not** the front-proxy url. 
The parameter is configured [here](https://github.com/kcp-dev/helm-charts/blob/main/charts/kcp/templates/server-deployment.yaml#L157) in the helm values and it differs from the external URL, which is defined one line below.

### Steps To Reproduce

1. install kcp on a cluster with the [helm chart](https://github.com/kcp-dev/helm-charts/tree/main/charts/kcp) and values:
```
externalHostname: my.external.hostname
kcpFrontProxy:
  service:
    type: LoadBalancer
```
configure the externalHostname DNS record to point to the front-proxy loadbalancer.
Now KCP works normally.

2. Add an (empty) APIExport, which will later be referenced by the api-syncagent

3. install the kcp api-syncagent on a second cluster and point it to the APIExport on KCP.

4. Create a Binding on KCP for the APIExport.

Now you will notice, the `APIExportEndpointSlice` gets filled with a URL: (removed some fields for brevity)
```
Name:         my.example.group
[...]

PR #3832, opened 2026-02-11
Deprecate --external-hostname and set based on --shard-base-url or --bind-address

ntnn:
## Summary

Disallow setting `--external-hostname` and set based on `--shard-base-url` or `--bind-address` instead.

In kcp-operator we are already forcing `--external-hostname` to `--shard-base-url`: https://github.com/kcp-dev/kcp-operator/blob/850a5f963929303bc1b3c311aa7a23360ab7b871/internal/resources/shard/deployment.go#L203-L204

## What Type of PR Is This?

/kind cleanup

## Related Issue(s)

Somewhat related to #3830, but not directly.

## Release Notes

```release-note
Deprecate `--external-hostname`, determined based on `--shard-base-url` or `--bind-address` instead
```

### DIFF
--- a/pkg/server/options/flags.go
+++ b/pkg/server/options/flags.go
@@ -122,7 +122,6 @@ var (
 		"shutdown-delay-duration",                 // Time to delay the termination. During that time the server keeps serving requests normally. The endpoints /healthz and /livez will return success, but /readyz immediately returns failure. Graceful termination starts after this delay has elapsed. This can be used to allow load balancer to stop sending traffic to this server.
 		"shutdown-send-retry-after",               // If true the HTTP Server will continue listening until all non long running request(s) in flight have been drained, during this window all incoming requests will be rejected with a status code 429 and a 'Retry-After' response header, in addition 'Connection: close' response header is set in order to tear down the TCP connection when idle.
 		"strict-transport-security-directives",    // List of directives for HSTS, comma separated. If this list is empty, then HSTS directives will not be added. Example: 'max-age=31536000,includeSubDomains,preload'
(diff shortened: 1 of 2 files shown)


ntnn:
Mhm, saw this while looking through other things: https://github.com/xrstf/kcp/blob/570d782411548335dd9aa41fb98bbf1e94b4a935/pkg/server/config.go#L583-L589

So `.ShardBaseURL` defaults to `.GenericConfig.ExternalAddress` if not set. Could be ok. Feels off.

ntnn:
No I think that is ok, those seems to be two different things.

ntnn:
Looking through references of that the `.ExternalAddress` is only used to build shard-local URLs:

https://github.com/ntnn/kcp/blob/8f5b9af8a054c481822a811c93d240e6d4ed5938/cmd/virtual-workspaces/command/cmd.go#L221-L222

https://github.com/ntnn/kcp/blob/8f5b9af8a054c481822a811c93d240e6d4ed5938/pkg/server/controllers.go#L969-L970

https://github.com/ntnn/kcp/blob/8f5b9af8a054c481822a811c93d240e6d4ed5938/pkg/server/virtual.go#L113-L114

This one could be intended to be for the front-proxy? But even here it's fine to write the shard URL instead of whatever the user sets:
https://github.com/ntnn/kcp/blob/8f5b9af8a054c481822a811c93d240e6d4ed5938/pkg/server/options/authentication.go#L154-L155

Insofar I don't think even allowing `--external-hostname` to set something other than the shard URL would make sense.

PR #3833, opened 2026-02-11
prevent concurrent map access in kube-openapi dependencies

### LINKED ISSUE #3820: bug: concurrent map access panic (openapi?)
### Describe the bug

I've encountered following panics:

* fatal error: concurrent map writes
* fatal error: concurrent map iteration and map write

See the logs in the attachment. There is quite a few mentions of openapi.

[concurrent-map-writes-panic-logs.tar.gz](https://github.com/user-attachments/files/25091037/concurrent-map-writes-panic-logs.tar.gz)

### Steps To Reproduce

This is a race, very hard to reproduce :C

### Expected Behaviour

No concurrent access to unprotected maps.

### Additional Context

_No response_

xrstf:
## Summary
Claude suggested this as the fix for the rare panic during bootup, reported in #3820. To be honest, I have no real idea if this is correct, but the tests sure do pass...

## What Type of PR Is This?
/kind bug

## Related Issue(s)
Fixes #3820

## Release Notes
```release-note
Fix very rare openapi-related panic during startup.
```

### DIFF
--- a/pkg/server/openapiv3/servicecache.go
+++ b/pkg/server/openapiv3/servicecache.go
@@ -104,7 +104,15 @@ func (c *ServiceCache) RegisterStaticAPIs(cont *restful.Container) error {
 	for gvPath, ws := range byGVPath {
 		c.staticSpecs[gvPath] = cached.Once(cached.Func[*spec3.OpenAPI](
 			func() (value *spec3.OpenAPI, etag string, err error) {
-				spec, err := builder3.BuildOpenAPISpecFromRoutes(restfuladapter.AdaptWebServices(ws), c.config)
+				// Copy the config and clear Definitions so that newOpenAPI() calls
+				// GetDefinitions, creating a fresh definitions map per invocation.
+				// Without this, the caching layer's listMerger evaluates multiple
+				// closures in parallel and they all share the same Definitions map.
+				// buildDefinitionRecursively mutates Extension maps reachable
+				// through that shared map, causing concurrent map writes.
+				cfg := *c.config
+				cfg.Definitions = nil
+				spec, err := builder3.BuildOpenAPISpecFromRoutes(restfuladapter.AdaptWebServices(ws), &cfg)
 				if err != nil {
 					return nil, "", fmt.Errorf("failed to build OpenAPI v3 spec for %s: %w", gvPath, err)
 				}


pkg/server/openapiv3/servicecache.go
```diff
@@ -104,7 +104,15 @@ func (c *ServiceCache) RegisterStaticAPIs(cont *restful.Container) error {
...
-				spec, err := builder3.BuildOpenAPISpecFromRoutes(restfuladapter.AdaptWebServices(ws), c.config)
+				// Copy the config and clear Definitions so that newOpenAPI() calls
+				// GetDefinitions, creating a fresh definitions map per invocation.
+				// Without this, the caching layer's listMerger evaluates multiple
+				// closures in parallel and they all share the same Definitions map.
+				// buildDefinitionRecursively mutates Extension maps reachable
+				// through that shared map, causing concurrent map writes.
+				cfg := *c.config
+				cfg.Definitions = nil
+				spec, err := builder3.BuildOpenAPISpecFromRoutes(restfuladapter.AdaptWebServices(ws), &cfg)
```
ntnn, line 115: Makes sense, though many of the members of OpenAPIV3Config are pointers. I wondered if the tests only pass by accident and a deepcopy may be better, but all this setup is in the `NewConfig` and the OpenAPIV3Config is modified in the `.Complete`, which would be unavailable here if it is deep copied.

I also wonder if there isn't a reason the `.Definitions` is set, since it takes precedent over `.GetDefinitons`: https://pkg.go.dev/k8s.io/kube-openapi/pkg/common#OpenAPIV3Config

Since this in a `cached.Once` - maybe it'd be better to put in a lock?

xrstf: I really couldn't tell ya, personally I've never encountered this panic before.

mjudeikis: I would be tempted to try and if we see issues again - revert while we mid release cycle

ntnn: I'd merge and review before tagging the next release. It's safer than before but I think we wouldn't get a major slowdown if we mutexted these functions.
It also looks like this only runs once for a startup ondemand. We could also make it sequential and move in a goroutine. That feels better than hoping that the concurrency is solved by random tries.

ntnn:
/lgtm
/approve

@gman0 please see if you can forcibly reproduce the issue somehow. Given that it is related to building the openapispec maybe you were doing many requests during startup?

PR #3835, opened 2026-02-11
WIP Fix self-subject reviews returning 403 even with the right RBAC permissions

### LINKED ISSUE #3809: Bug: Permissions for authorization introspection APIs not behaving as expected (SSRR, SSAR)
### Describe the bug

Creating SelfSubjectAccessReveiws and SelfSubjectRulesReviews fails with unauthorized even if the user should have access according to ClusterRole+ClusterRoleBinding in a workspace.

### Steps To Reproduce

1. Create new workspace `kubectl create-workspace ssrr-poc`
2. Deploy the ClusterRole + ClusterRoleBinding below:
```yaml
kind: ClusterRole
metadata:
  name: introspection
rules:
- apiGroups:
  - authorization.k8s.io
  resources:
  - selfsubjectaccessreviews
  - selfsubjectrulesreviews
  verbs:
  - create

---

kind: ClusterRoleBinding
metadata:
  name: introspection-binding
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: introspection
subjects:
- kind: User
  name: oidc:example@example.com
- kind: Group
  name: system:authenticated
```

---
SelfSubjectAccessReviews:

run `kubectl auth can-i create selfsubjectaccessreviews --as=oidc:example@example.com --as-group=system:authenticated`

---

SelfSubjectRulesReviews:

```bash
kubectl create --as=oidc:example@example.com -oyaml --validate=false  -f - <<EOF 
apiVersion: authorization.k8s.io/v1
kind: SelfSubjectRulesReview
spec:
  namespace: default
EOF
```

### Expected Behaviour

SelfSubjectAccessReviews:

Expected: Valid response with result.
Observed: Fails with HTTP status 403

---

SelfSubjectRulesReviews:

Expected: Valid response with result
Observed: Error from server (Forbidden): error when creating "STDIN": selfsubjectrulesreviews.authorization.k8s.io is forbidden: User "oidc:example@example.com" cannot create resource "selfsubjectrulesreviews" in API group "authorization.k8s.io" at the cluster scope: access denied

### Additional Context

Creating a ClusterRoleBinding to `cluster-admin` expectedly allows the user to perform both actions.

iakmc:
## Summary
- SelfSubjectAccessReviews, SelfSubjectRulesReviews, and SelfSubjectReviews were getting 403 even when users had the right RBAC permissions

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #3809

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/authorization/workspace_content_authorizer.go
+++ b/pkg/authorization/workspace_content_authorizer.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"strings"
 
+	authenticationv1 "k8s.io/api/authentication/v1"
+	authorizationv1 "k8s.io/api/authorization/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apiserver/pkg/authorization/authorizer"
 	genericapirequest "k8s.io/apiserver/pkg/endpoints/request"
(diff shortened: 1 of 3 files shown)


mjudeikis:
Do you tracked which authoriser in the code returns 403? I suspect this https://github.com/iakmc/kcp/blob/ddc083d6989df952cf884fb9d5e5cdd7ef4c5b66/pkg/authorization/workspace_content_authorizer.go#L146  maybe?

I would like to understand which authorizer is supposed to evaluate this and didn't? RBAC is senstivie topic, so before we stab something like this I want to understand where the failure is

PR #3847, opened 2026-02-19
Parallelise installing resources

ntnn:
## Summary

I've been fiddling with the startup because 30s cold start for an instance is a bit much for my liking.
This cuts a bit off - not a lot though. On average parallelizing this cuts down ~5s startup time.

Other functions are already parallelizing in the same way, so it's more of a cleanup to adhere to standards set in other functions.

## What Type of PR Is This?

/kind cleanup

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/config/helpers/bootstrap.go
+++ b/config/helpers/bootstrap.go
@@ -103,7 +103,7 @@ func CreateResourcesFromFS(ctx context.Context, client dynamic.Interface, mapper
 	}
 
 	groups := GroupObjectsByDefaultHierarchy(resources)
-	if err := CreateResourcesFromGroups(ctx, client, mapper, groups); err != nil {
+	if err := CreateGroupedResources(ctx, client, mapper, groups); err != nil {
 		return fmt.Errorf("could not create resources from groups: %w", err)
 	}
 	return nil
@@ -123,14 +123,14 @@ func CreateResourceFromFS(ctx context.Context, client dynamic.Interface, mapper 
 	}
 
 	groups := GroupObjectsByDefaultHierarchy(resources)
-	if err := CreateResourcesFromGroups(ctx, client, mapper, groups); err != nil {
+	if err := CreateGroupedResources(ctx, client, mapper, groups); err != nil {
 		return fmt.Errorf("could not create resources from groups: %w", err)
 	}
 	return nil
 }
 
-// CreateResourcesFromGroups sequentially runs CreateResources for each group of resources.
-func CreateResourcesFromGroups(ctx context.Context, client dynamic.Interface, mapper meta.RESTMapper, groups [][]*unstructured.Unstructured) error {
+// CreateGroupedResources sequentially runs CreateResources for each group of resources.
+func CreateGroupedResources(ctx context.Context, client dynamic.Interface, mapper meta.RESTMapper, groups [][]*unstructured.Unstructured) error {
 	for i, resources := range groups {
 		if err := CreateResources(ctx, client, mapper, resources); err != nil {
 			return fmt.Errorf("failed to create resources from group %d: %w", i, err)


config/helpers/bootstrap.go
```diff
@@ -130,28 +142,45 @@ func CreateResourceFromFS(ctx context.Context, client dynamic.Interface, mapper
...
-	var errs []error
+	docs := [][]byte{}
 	for i := 1; ; i++ {
 		doc, err := d.Read()
 		if errors.Is(err, io.EOF) {
 			break
 		} else if err != nil {
-			return err
-		}
-		if len(bytes.TrimSpace(doc)) == 0 {
```
gman0, line 142: Why move the empty-doc check from here into the `for ... := range docs` loop below?

gman0:
Otherwise LGTM.

config/helpers/bootstrap.go
ntnn: True that is better, at the time I was more focused on the startup in pkg/server so I didn't think twice about it.

config/helpers/bootstrap.go
```diff
@@ -106,14 +107,25 @@ func CreateResourcesFromFS(ctx context.Context, client dynamic.Interface, mapper
...
-		if err := CreateResourceFromFS(ctx, client, mapper, batteriesIncluded, f.Name(), fs, transformers...); err != nil {
-			errs = append(errs, err)
-		}
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			if f.IsDir() {
+				return
+			}
+			if err := CreateResourceFromFS(ctx, client, mapper, batteriesIncluded, f.Name(), fs, transformers...); err != nil {
```
mjudeikis: I remember some bootstrap files used a file in order. Can we maybe add some high-level ordering? Meaning - namespaces first, cluster object second, and the rest follows? 

This lib is heavy used outside kcp in our other repos and I think it might break things if no ordering are done?

ntnn: Mhm, we could do weights but I'd rather wrap `createResourceFromFS` in an eventually so ordering doesn't matter.
So basically `CreateResourcesFromFS` - multiplex -> `CreateResourceFromFS` - multiplex -> `createRetryResourceFromFS` - retry -> `createResourceFromFs`.
I was already thinking that wrapping `CreateResourcesFromFS` in an eventually in the startup was a bit too expensive because on every failure it would apply _everything_ again.

mjudeikis: But does this mean if you have 100 files, it will fire 100 create requests, and 99 will fail due to missing namespace, and 1 namespace succeeds, and after that, 99 retries succeed. 

vs: 1 namespace request, and 99 following. 

Basically, 199 requests vs 100 requests.

ntnn: True, but then we should redesign the resource installation; so we weight the resources being installing, group them by weight and install in chunks, with resources in one chunk being parallelised.
I think @xrstf already made a similar weighing for the init-agent, we could reuse that code and go from there.

ntnn:
/hold 

I've parallelised installing resources in a few more spaces. Works fine to startup etc.pp. initially but now the batteries kcp-admin can't access workspace types:

```text
Error from server (Forbidden): workspacetypes.tenancy.kcp.io is forbidden: User "kcp-admin" cannot list resource "workspacetypes" in API group "tenancy.kcp.io" at the cluster scope: access denied
```

I'll revisit the changes later.

config/crds/bootstrap.go
```diff
@@ -124,6 +124,32 @@ func CRD(fs embed.FS, gr metav1.GroupResource) (*apiextensionsv1.CustomResourceD
...
+		}()
+	}
+	wg.Wait()
+	close(errsChan)
+
+	errs := make([]error, 0, len(rawCRDs))
+	for err := range errsChan {
+		errs = append(errs, err)
+	}
+	return utilerrors.NewAggregate(errs)
```
xrstf: This PR contains this exact code (create slice, crate worker routines, wait for them to finish, collect the errors into a slice) 4 times. Can/should we make this a helper somewhere that I can just call with `doInParallel(f ...func(ctx) error) error`?

ntnn: > This PR contains this exact code (create slice, crate worker routines, wait for them to finish, collect the errors into a slice) 4 times. Can/should we make this a helper somewhere that I can just call with `doInParallel(f ...func(ctx) error) error`?

I was thinking using errgroup for that: https://pkg.go.dev/golang.org/x/sync/errgroup 
The only drawback there is that it returns the first error and not all errors. Then again during bootstrapping we don't particularly cared what failed, we just want to bootstrap and exit on an error.

ntnn: @mjudeikis-bot Create a PR that adds a utility function like that in a suitable package in kcp. If there is no suitable package create a new package in `pkg`.
The function should take the `exitEarly` boolean that cancel all goroutines when one returns an error.

xrstf: @mjudeikis-bot Please open a browser tab and play any random Harlem shake video on YouTube on full volume.

xrstf: Clbuttic!

ntnn:
/retest 

Now that codeberg is back up

ntnn:
@mjudeikis-bot review!

ntnn:
/hold cancel

Okay, I finally worked out all of the nits that were wrong. I also moved a bunch of code so it's not _everything_ in the one bootstrap.go file.

config/helpers/bootstrap.go
```diff
@@ -88,121 +53,112 @@ func Bootstrap(ctx context.Context, discoveryClient discovery.DiscoveryInterface
...
 	return wait.PollUntilContextCancel(ctx, time.Second, true, func(ctx context.Context) (bool, error) {
-		if err := CreateResourcesFromFS(ctx, dynamicClient, mapper, batteriesIncluded, fs, transformers...); err != nil {
+		if err := CreateResourcesFromChunk(ctx, dynamicClient, mapper, chunk); err != nil {
 			klog.FromContext(ctx).WithValues("err", err).Info("failed to bootstrap resources, retrying")
 			// invalidate cache if resources not found
 			// xref: https://github.com/kcp-dev/kcp/issues/655
-			cache.Invalidate()
+			// After adding ordering to the installation this
+			// technically isn't needed anymore - however other projects
+			// are also using the Bootstrap function and they resource
```
ntnn:
```suggestion
			// are also using the Bootstrap function and their resource
```

ntnn:
@mjudeikis-bot Apply each commit in this PR and run the lint and e2e tests every time to verify that each commit is atomic.

config/helpers/bootstrap.go
```diff
@@ -88,121 +53,112 @@ func Bootstrap(ctx context.Context, discoveryClient discovery.DiscoveryInterface
...
-			}
-		}
+// CreateResourcesFromChunk creates all resources from a chunk of resources.
+func CreateResourcesFromChunk(ctx context.Context, client dynamic.Interface, mapper meta.RESTMapper, chunk []*unstructured.Unstructured) error {
+	g := errgroup.WithContext(ctx)
 
-		if err := createResourceFromFS(ctx, client, mapper, doc, batteriesIncluded); err != nil {
-			errs = append(errs, fmt.Errorf("failed to create resource %s doc %d: %w", filename, i, err))
-		}
+	for _, obj := range chunk {
```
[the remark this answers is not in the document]
ntnn: That is very old, Go isn't reusing the same loopvar object since 1.21.

config/helpers/binds.go
```diff
@@ -0,0 +1,103 @@
...
+
+func BindRootAPIs(ctx context.Context, kcpClient kcpclient.Interface, exportNames ...string) error {
+	g := errgroup.WithContext(ctx)
+	for _, exportName := range exportNames {
+		g.Go(func(ctx context.Context) error {
+			if err := BindRootAPI(ctx, kcpClient, exportName); err != nil {
+				return fmt.Errorf("failed to bind root API %s: %w", exportName, err)
+			}
+			return nil
+		})
```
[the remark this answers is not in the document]
ntnn: As above, that hasn't been the case since go1.21

config/helpers/read.go
```diff
@@ -0,0 +1,129 @@
...
+	for _, resource := range resources {
+		value, found := resource.GetAnnotations()[annotationBattery]
+		if !found {
+			// resource has no batteries annotation, add it to be
+			// installed
+			ret = append(ret, resource)
+			continue
+		}
+
+		annotated := sets.New[string](strings.Split(value, ",")...)
```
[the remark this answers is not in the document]
ntnn: That's fine, that's user error.

ntnn:
/test pull-kcp-test-e2e-shared

That looks like flakes.

config/helpers/bootstrap.go
```diff
@@ -88,121 +53,112 @@ func Bootstrap(ctx context.Context, discoveryClient discovery.DiscoveryInterface
...
-		doc, err := d.Read()
-		if errors.Is(err, io.EOF) {
-			break
-		} else if err != nil {
-			return err
-		}
-		if len(bytes.TrimSpace(doc)) == 0 {
-			continue
+// CreateResourcesFromChunks sequentially runs CreateResourcesFromChunk for each chunk of resources.
+func CreateResourcesFromChunks(ctx context.Context, client dynamic.Interface, mapper meta.RESTMapper, chunks [][]*unstructured.Unstructured) error {
```
xrstf: I find the name "Chunk" to be a bit cumbersome for an exported function. Chunking is a concept really only done internally in this package, and from an outsider's perspective, when I see "FromChunks" I will wonder "what chunks? I have this list of objects here, do I have to chunk it first? My list _is_ a chunk!".

If not needed, I wouldn't export this function at all.

config/helpers/bootstrap.go
```diff
@@ -88,121 +53,112 @@ func Bootstrap(ctx context.Context, discoveryClient discovery.DiscoveryInterface
...
+}
 
-		for _, transformer := range transformers {
-			doc, err = transformer(doc)
-			if err != nil {
-				return err
-			}
-		}
+// CreateResourcesFromChunk creates all resources from a chunk of resources.
+func CreateResourcesFromChunk(ctx context.Context, client dynamic.Interface, mapper meta.RESTMapper, chunk []*unstructured.Unstructured) error {
```
xrstf: Same here. This function to an outside caller would make much more sense if it's simply `CreateResources(ctx, client, mapper, resources)`.

config/helpers/read.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error templating manifest %q: %w", filename, err)
+	}
+
+	return raw, nil
+}
+
+// ReadResourceFromFS uses ReadRawFromFS to read a file and returns it
+// based on the batteries input as a slice of unstructured.Unstructured.
+func ReadResourceFromFS(
```
xrstf, line 58: Is your intention to make this a re-usable kcp component for third party apps? There are so many public API functions now, I am beginning to wonder if this whole package should be in the SDK now?

config/helpers/sort.go
```diff
@@ -0,0 +1,100 @@
...
+	"slices"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+// ChunkObjectsByHierarchy returns the given objects chunked by their
+// hierarchy as noted in SortObjectsByHierarchy.
+// The objects slice is modified in-place.
+func ChunkObjectsByHierarchy(objects []*unstructured.Unstructured) [][]*unstructured.Unstructured {
```
xrstf: Who is meant to use this function? They cannot influence the weighting, so ... is this really that re-usable? 🤔

xrstf (commented):
cc

config/helpers/bootstrap.go
ntnn: Mhm true; it was the first term that came to my mind. `CreateGroupedResources` maybe? Then the slice of slice still makes sense.

config/helpers/read.go
ntnn: @mjudeikis mentioned that at least the `Bootstrap` is also used by other components, at least in kcp.
Checking pkg.go.dev we are at least using this package in kube-bind and tmc: https://pkg.go.dev/github.com/kcp-dev/kcp@v0.30.1/config/helpers?tab=importedby 
Moving to the sdk wouldn't be a terrible idea - there's also an issue to e.g. install the CRDs in kcp-operator, which these functions could make very easy.

ntnn:
/retest

All flakes

ntnn:
/retest

ci infra

config/helpers/bootstrap.go
xrstf: I still wonder why the whole "grouped" aspect needs to be exposed to the callers. Can you outline an example usecase where someone would want a grouped/chunked API to work with?

ntnn: @mjudeikis-bot That's enough, you can ignore this PR now.



> I still wonder why the whole "grouped" aspect needs to be exposed to the callers. Can you outline an example usecase where someone would want a grouped/chunked API to work with?

I can see using this in e.g. kube-bind where the same problem of `APIServiceExport` -> `APIServiceBinding` etc.pp. exists, so we could reuse the same functions in kube-bind and just use weights from kube-bind; admittedly I only got that idea when looking into where else these helpers are used :D 

Outside of that anything where resources have a hierarchical order to install can make use of those grouped installs.

PR #3849, opened 2026-02-20
Deprecate `--shard-external-url` for virtual-workspace

ntnn:
## Summary

As noted here the flag is never actually used for virtual-workspace: https://github.com/kcp-dev/kcp-operator/issues/139#issuecomment-3935234308
It doesn't make sense to require a flag that isn't actually being used.

We can also accept but not use that flag and mark it deprecated. Mostly wanna see what breaks when removing it entire.

## What Type of PR Is This?

/kind cleanup

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Deprecated the unused flag `--shard-external-url` for `virtual-workspace` 
```

### DIFF
--- a/cmd/sharded-test-server/virtual.go
+++ b/cmd/sharded-test-server/virtual.go
@@ -146,7 +146,6 @@ func newVirtualWorkspace(ctx context.Context, index int, servingCA *crypto.CA, h
 
 	args := []string{
 		fmt.Sprintf("--kubeconfig=%s", kubeconfigPath),
-		fmt.Sprintf("--shard-external-url=https://%s:%d", hostIP, 6443),
 		fmt.Sprintf("--cache-kubeconfig=%s", cacheServerConfigPath),
 		fmt.Sprintf("--authentication-kubeconfig=%s", authenticationKubeconfigPath),
 		fmt.Sprintf("--client-ca-file=%s", clientCAFilePath),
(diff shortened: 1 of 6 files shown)


mjudeikis:
If somebody provides this flag, and we remove it, it will throw "Unknown flag"? 
And im surpsied this is not used.

ntnn:
> If somebody provides this flag, and we remove it, it will throw "Unknown flag"?

Yeah, that's why I'm contemplating just making it simply not required. And if we want e.g. kcp-operator to also deploy other VWs it would be good to have a "standard set" of flags to pass to set the front-proxy (for singleton VWs) and the shard url (for per-shard VWs).

> And im surpsied this is not used.

I'm not that surprised; the VW can retrieve all information they need from the shard.
And this might've been added at the start before knowing what would be needed.

ntnn:
After thinking some more I don't think it makes sense to keep the flag around.

Both singleton and per-shard VWs would need a kubeconfig for the front-proxy and their shard respectively and can infer the correct URL from there, so it makes little sense to make a new convention to pass a URL when there is already a conventional way to retrieve said URL.
Custom VWs that don't need a kubeconfig can still get the respective URL passed from the person deploying them - we don't need to interfere and set another standard.

mjudeikis:
If we want to remove it, we should deprecate and remove it later. I don't want to have "Flag not defined" errors panicking deployments.

mjudeikis (approved):
/lgtm
/approve

PR #3857, opened 2026-02-25
Fix {cluster} extract logic for vw

mjudeikis:
## Summary

Still need testing with custom image in the env where I hit this.

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
BREAKING CHANGE: 
Fix {cluster} extract logic for VirtualWorkspaces. Previously, if VirtualWorkspace, used in FrontProxy mapping, had a path `cluster/{cluster}` - it was not resolved, and so WorkspaceAuthorizationConfiguration was not run if used inside FrontProxy, but forwarded to VirtualWorkspace without checking. As a result, if one has misconfigured VirtualWorkspace, it might receive traffic intended for another recipient. 
```

### DIFF
--- a/pkg/proxy/lookup/lookup.go
+++ b/pkg/proxy/lookup/lookup.go
@@ -113,9 +113,24 @@ func newClusterResolveHandler(delegate http.Handler, index proxyindex.Index) htt
 
 func newMappingHandler(delegate http.Handler, index proxyindex.Index) http.HandlerFunc {
 	return func(w http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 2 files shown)


ntnn:
/hold 

We are discussing but I don't think it's a good idea to allow routing more `/services/` URLs through the front proxy. It was a mistake from the start and this would just allow more problems.

ntnn:
/unhold 
/lgtm
/approve

We finished discussing and while this is not a good change it is a needed change.

PR #3859, opened 2026-02-25
fix:  cross-shard APIBinding selector inheritance

cnvergence:
## Summary

When a child workspace is created on a different shard than its parent, the APIBinder initializing controller and DefaultAPIBindingLifecycle controller could not find the parent workspace's APIBinding. As a result, permission claim selectors were not inherited.

Replaced the local-only APIBinding lookup with a global cache lookup. `listAPIBindingsByPath` now queries the cache server's APIBindings informer directly using the kcp.io/path index
Added apis.kcp.io/apibindings CRD to the cache server bootstrap list and to the replication controller.

/kind bug

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3860

## Release Notes

```release-note
NONE
```


mjudeikis (approved):
/lgtm
/approve

PR #3872, opened 2026-03-02
Improve version banner in our docs

xrstf:
## Summary
I got triggered _hard_ by our docs shifting themselves down by a few lines every time you navigate through them and load a new page. Everytime I clicked on a link, began to rea---- ah, sorry, everything just shifted down. Infuriating.

This PR moves the JS code from the page head to the end of the `<body>`, which allows it to immediately get all the necessary DOM elements and inject itself into the page. This significantly reduces the likelihood of a visual tear on pageload.

I also changed the background color a bit, since the contrast was very low and it was hard to read for me. The color was chosen from the `!!! note` blocks' border.

Old:

screenshot_20260302_143305

New:

screenshot_20260302_143313

I also removed the explicit positioning (setting the `top` CSS property) since I could not see/reproduce what this might be used for. Without it, it still looks fine in Chrome and Firefox.

Lastly, since I got bit by that, too, I ran [`txtidy`](codeberg.org/xrstf/txtidy) on the docs and stripped all trailing whitespace.

## What Type of PR Is This?
/kind documentation

## Release Notes
```release-note
NONE
```


docs/content/setup/production/.pages
```diff
@@ -1,8 +1,9 @@
...
   - index.md
-  - overview.md  
+  - overview.md
   - prerequisites.md
   - kcp-dekker.md
   - kcp-vespucci.md
-  - kcp-comer.md
\ No newline at end of file
+  - kcp-comer.md
+  - audit-logging.md
```
ntnn, line 9: audit-logging link is gone

ntnn: got my wires crossed :D

PR #3881, opened 2026-03-04
Add APIResourceSchema VW

mjudeikis:
## Summary

When dealing with projects like kube-bind.io, integrating into kcp, one might need to access the schemas of the consumers. This adds a dedicated VW to get schema objects.

Prerequisite:
1. GlobalServiceAccounts enabled
2. Provider entity has `/ - access` and `get - apibindings` permission in the consumer workspaces.

There are no EndpointSlices for this (we might add them later; for now, we want to test it manually). As this could be considered global VW, since it deals mostly with cache, the same results can be served from any shard (if a sharded deployment is used). 

There is more extensive docs in the docs.go

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Add apiresourceschema virtual workspace for schema access from provider side.
```

### DIFF
--- a/test/integration/workspace/leak_test.go
+++ b/test/integration/workspace/leak_test.go
@@ -94,9 +94,15 @@ var (
 	// operations and aren't indicative of a leak when deleting
 	// a workspace.
 	randomGoroutineSources = []goleak.Option{
+		// HTTP/HTTP2 goroutines - these are connection pool goroutines
+		// that may persist briefly after operations complete
 		goleak.IgnoreCreatedBy("net/http.(*Server).Serve"),
 		goleak.IgnoreCreatedBy("golang.org/x/net/http2.(*Transport).newClientConn"),
 		goleak.IgnoreCreatedBy("golang.org/x/net/http2.(*serverConn).serve"),
+		// gRPC transport goroutines from etcd client - these are connection
+		// management goroutines that may briefly outlive operations
+		goleak.IgnoreTopFunction("google.golang.org/grpc.(*addrConn).resetTransportAndUnlock"),
+		goleak.IgnoreCreatedBy("google.golang.org/grpc.(*acBalancerWrapper).Connect"),
 	}
 )
 


pkg/virtual/apiresourceschema/builder/build.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	realPath := "/"
+	if len(parts) > 1 {
+		realPath = "/" + parts[1]
+	}
+
+	// Handle /clusters/<cluster-path>/... pattern that clients add
+	// We need to strip this part to get to the API path
+	if strings.HasPrefix(realPath, "/clusters/") {
```
gman0, line 181: Would it make sense to enforce the `.../clusters/<Target cluster>` to be equal to `consumerClusterName`? It could be a little confusing if it's just ignored.

pkg/virtual/apiresourceschema/doc.go
```diff
@@ -0,0 +1,234 @@
...
+//	│                         │           │                         │
+//	│  ServiceAccount:        │           │  ClusterRoleBinding:    │
+//	│    provider-operator    │──────────►│    subjects:            │
+//	│                         │ Global SA │    - kind: User         │
+//	│  JWT contains:          │  Format   │      name: system:kcp:  │
+//	│    ClusterNameKey       │           │        serviceaccount:  │
+//	│                         │           │        <provider-cluster│
+//	│                         │           │        >:default:       │
+//	│                         │           │        provider-operator│
+//	└─────────────────────────┘           └─────────────────────────┘
```
ntnn, line 193: Why is foreign SA access relevant? I'd assume the same mechanisms as should be usual apply and foreign SAs don't need special consideration.
And the functionality is for identities with access to cluster xyz to be able to read schemas bound to cluster xyz in a VW on the local shard.
Maybe I'm missing something but I feel mentioning foreign SAs could mislead people

pkg/virtual/apiresourceschema/doc.go
```diff
@@ -0,0 +1,234 @@
...
+//
+//	Test 3: Read-Only Enforcement
+//	  - CREATE  ──────────► Forbidden/MethodNotSupported
+//	  - UPDATE  ──────────► Forbidden/MethodNotSupported
+//	  - DELETE  ──────────► Forbidden/MethodNotSupported
+//
+//	Test 4: Authorization (Cross-Workspace)
+//	  - Provider SA without RBAC ──────────► Forbidden
+//	  - Grant RBAC in consumer workspace
+//	  - Provider SA with RBAC ──────────────► Access granted
```
ntnn: I don't think documenting the test coverage is a good idea. As a TODO list or maybe a "checklist" of sorts in a docstring next to the tests sure but not in a `doc.go`.

pkg/virtual/apiresourceschema/doc.go
```diff
@@ -0,0 +1,234 @@
...
+//
+//
+//	                         Key Constraints
+//	---------------------------------------------------------------------------
+//
+//	- READ-ONLY: Only get, list, watch operations are supported
+//	- SCOPE: Only schemas from bound APIBindings are visible
+//	- DYNAMIC: Schema list updates as bindings change
+//	- CROSS-CLUSTER: Schemas aggregated from multiple provider workspaces
+//	- CLUSTER-SCOPED: APIResourceSchemas are not namespaced
```
ntnn: I think this can be scrapped, or? All of these are explained above.

test/e2e/virtual/apiresourceschema/virtualworkspace_test.go
```diff
@@ -0,0 +1,609 @@
...
+	apifixtures.BindToExport(t.Context(), t, providerPath, group, consumerPath, kcpClusterClient)
+
+	// Wait for binding to be bound
+	kcptestinghelpers.EventuallyCondition(t, func() (conditions.Getter, error) {
+		return kcpClusterClient.Cluster(consumerPath).ApisV1alpha2().APIBindings().Get(t.Context(), group, metav1.GetOptions{})
+	}, kcptestinghelpers.Is(apisv1alpha2.InitialBindingCompleted))
+
+	// Set up the VW URL
+	rootShardCfg := server.RootShardSystemMasterBaseConfig(t)
+	vwURL := fmt.Sprintf("%s/services/%s/%s", rootShardCfg.Host, apiresourceschema.VirtualWorkspaceName, consumerClusterName)
```
ntnn: Shouldn't this fail for multi shard tests? The test should inspect the consumer WS and see which shard it landed on.

test/e2e/virtual/apiresourceschema/virtualworkspace_test.go
```diff
@@ -0,0 +1,609 @@
...
+	}, kcptestinghelpers.Is(apisv1alpha2.InitialBindingCompleted))
+
+	t.Logf("Waiting for APIBinding %s to be bound", group2)
+	kcptestinghelpers.EventuallyCondition(t, func() (conditions.Getter, error) {
+		return kcpClusterClient.Cluster(consumerPath).ApisV1alpha2().APIBindings().Get(t.Context(), group2, metav1.GetOptions{})
+	}, kcptestinghelpers.Is(apisv1alpha2.InitialBindingCompleted))
+
+	// Set up the VW client
+	rootShardCfg := server.RootShardSystemMasterBaseConfig(t)
+	vwURL := fmt.Sprintf("%s/services/%s/%s", rootShardCfg.Host, apiresourceschema.VirtualWorkspaceName, consumerClusterName)
```
ntnn: Same as above with the shard

test/e2e/virtual/apiresourceschema/virtualworkspace_test.go
```diff
@@ -0,0 +1,609 @@
...
+	t.Logf("Testing that delete is denied")
+	err = vwKcpClient.Cluster(consumerClusterName.Path()).ApisV1alpha1().APIResourceSchemas().Delete(t.Context(), schemaName, metav1.DeleteOptions{})
+	require.True(t, apierrors.IsForbidden(err) || apierrors.IsMethodNotSupported(err), "expected Forbidden or MethodNotSupported for delete, got: %v", err)
+
+	t.Logf("Verified that VW is read-only")
+}
+
+// TestAPIResourceSchemaVirtualWorkspaceMultipleBindings tests that when a consumer has multiple
+// APIBindings to different APIExports, all referenced schemas are accessible.
+func TestAPIResourceSchemaVirtualWorkspaceMultipleBindings(t *testing.T) {
```
ntnn, line 359: Isn't this test equivalent to TestAPIResourceSchemaVirtualWorkspaceAccess? Both create multiple providers, bind them to a single consumer and verify that both are available in the VW.

test/e2e/virtual/apiresourceschema/virtualworkspace_test.go
```diff
@@ -0,0 +1,609 @@
...
+// virtual workspace is properly gated by RBAC permissions on APIBindings in the consumer workspace.
+//
+// This test verifies:
+//  1. A ServiceAccount in the provider workspace without permission cannot access the VW
+//  2. After granting permission using the global kcp SA format (system:kcp:serviceaccount:<cluster>:<namespace>:<name>),
+//     the provider's ServiceAccount can access schemas through the consumer's VW
+//
+// This demonstrates cross-workspace ServiceAccount access using the kcp global SA identity format,
+// where ClusterNameKey is extracted from the SA token's JWT claims.
+func TestAPIResourceSchemaVirtualWorkspaceAuthorization(t *testing.T) {
```
ntnn, line 462:
```suggestion
// TestAPIResourceSchemaVirtualWorkspaceAuthorization tests that access to the apiresourceschema
// virtual workspace is properly gated by RBAC permissions on APIBindings in the consumer workspace.
//
// This test verifies:
//  1. A ServiceAccount in the provider workspace without permission cannot access the VW
//  2. After granting permission using the global kcp SA format (system:kcp:serviceaccount:<cluster>:<namespace>:<name>),
//     the provider's ServiceAccount can access schemas through the consumer's VW
//
// This demonstrates cross-workspace ServiceAccount access using the kcp global SA identity format,
// where ClusterNameKey is extracted from the SA token's JWT claims.
func TestAPIResourceSchemaVirtualWorkspaceAuthorization(t *testing.T) {
```

Makes sense - but I'd prefer if we use an SA from neither the provider or consumer. Setting up another workspace isn't (or shouldn't be) that expensive.

pkg/virtual/apiresourceschema/doc.go
mjudeikis: Because in most cases, this will be used/accessed by the ServiceAccount present in the provider's workspace. 
So this way its explisit - you need to have certain permissions in the consumer workspaces to access their schemas. 
It works very well with users, but I wanted to be 100% it will work in the controller-pattern using operators.

pkg/virtual/apiresourceschema/doc.go
mjudeikis: It depedens :D if you will know what is happening after 9 months when you need to check thjis :D

test/e2e/virtual/apiresourceschema/virtualworkspace_test.go
mjudeikis: As stated in the PR description, it operates on cached object, so it should not matter which shard is being called, it should have same data.

test/e2e/virtual/apiresourceschema/virtualworkspace_test.go
mjudeikis: Provider-side SA is the pattern users will be using. And if it works from provider one, it should work from anywhere else

test/e2e/virtual/apiresourceschema/virtualworkspace_test.go
mjudeikis: its bit duplicate, but easier to argue the logs.
This one tests that we can see all schemas (multiple bindings), where access concentrates on access. 

One can argue they can be merged, but this was more for reasoning in logs and failures.

test/e2e/virtual/apiresourceschema/virtualworkspace_test.go
mjudeikis: I dont understand this change diff :/

test/e2e/virtual/apiresourceschema/virtualworkspace_test.go
ntnn: Mhm true; I was thinking of the consumer workspace APIRS VW only being available on the consumers' shard but since it just proxies the request to the cache so to speak the explicit shard doesn't matter.

test/e2e/virtual/apiresourceschema/virtualworkspace_test.go
ntnn: I don't understand it either :D I think I accidentally hit suggest and didn't delete it and now it shows the text twice :D

test/e2e/virtual/apiresourceschema/virtualworkspace_test.go
mjudeikis: I think you are right. issue is with apibindings informer :/

test/e2e/virtual/apiresourceschema/virtualworkspace_test.go
mjudeikis: you were right. changed. Basically it worked because of https://github.com/kcp-dev/kcp/pull/3884 and it should not.

pkg/virtual/apiresourceschema/builder/build.go
mjudeikis: added

PR #3883, opened 2026-03-05
bump otel dependencies (CVE-2026-24051)

xrstf:
## Summary
This is to fix the security advisory https://github.com/kcp-dev/kcp/security/dependabot/121.

There are also two more advisories for the client-go and SDK modules in this repo, but using the same logic as in https://github.com/kcp-dev/api-syncagent/pull/137#issuecomment-3992792429

> I disagree with this change. The SDK should have the lowest compatible version listed, to increase its usefulness to consumers. Forcing everyone who imports our SDK onto a newer, and in this case even specific, Go version is IMHO the wrong approach. Just like the dependencies in the SDK should be the oldest compatible versions the SDK can work with.

... I did not bump dependencies in those two other modules. It's not up to us to dictate how consumers of our SDK/client-go handle their security updates. I.e. I think the other two security alerts we received should be dismissed (and maybe we configure dependabot to ignore those modules).

## What Type of PR Is This?
/kind chore

## Release Notes
```release-note
[CVE-2026-24051] Bump opentelemetry SDK to v1.41.0
```


xrstf:
Okay, I am now convinced that this is not a flake 😅

PR #3884, opened 2026-03-06
feat: Add defaultSelector to PermissionClaim on APIExport for automatic selector inheritance

mjudeikis:
## Summary

This PR adds a `defaultSelector` field to `PermissionClaim` on `APIExport`, allowing API providers to specify default permission claim selectors that are automatically applied when APIBindings are created via WorkspaceType.defaultAPIBindings. Else, same as other permissions - its a suggesion.

### Motivation

This replaces the approach in #3859, which proposed replicating `APIBindings` to the cache server to enable cross-shard selector inheritance. That approach had significant scalability concerns due to high cardinality (O(workspaces × bindings)).

Instead, this solution leverages the fact that `APIExport` is already replicated to the cache server. By adding defaultSelector directly to the `PermissionClaim` on `APIExport`, we achieve the same goal without additional replication overhead.

### Changes

- Added `DefaultSelector` field to `PermissionClaim` in types_apiexport.go
- Updated `defaultapibindinglifecycle` controller to use `DefaultSelector` from `APIExport` when creating `APIBindings`
- Updated `apibinder_initializer` controller with the same logic
- Added v1alpha1 ↔ v1alpha2 conversion support for the new field
- Added E2E test to verify selector inheritance works correctly
- Removed cross-shard APIBinding lookup logic that is no longer needed

### How it works
- API provider sets `defaultSelector` on `PermissionClaim` in their `APIExport`
- When a workspace is created with a `WorkspaceType` that references this `APIExport` in `defaultAPIBindings`
- The controller creates an `APIBinding` with permission claims that use the `defaultSelector` from the `APIExport`
- If no `defaultSelector` is specified, the existing behaviour (`matchAll: true`) is preserved

### Example

```yaml
kind: APIExport
metadata:
  name: my-api
spec:
  permissionClaims:
    - group: ""
      resource: configmaps
      verbs: ["get", "list"]
      defaultSelector:
        labelSelector:
          matchLabels:
            platform-mesh.io/enabled: "true"
```

When an APIBinding is auto-created via WorkspaceType, it will have:

```
permissionClaims:
  - group: ""
    resource: configmaps
    state: Accepted
    selector:
      labelSelector:
        matchLabels:
          platform-mesh.io/enabled: "true"
```

# UPDATE

I added a warrning conditon telling when Claims are different:

```
PermissionClaimsValid: False
Reason: PermissionClaimsMismatch
Severity: Warning
Message: "2 permission claim(s) have mismatched verbs or selectors (showing first 2): secrets. (verbs: accepted [get list], exported [get list watch]), configmaps. (selector differs from defaultSelector)"
```
So its easier to argue things on use side

## What Type of PR Is This?
/kind bug
/kind cleanup
/kind feature
/kind api-change
/kind regression

## Related Issue(s)

Fixes #

## Release Notes

```release-note
action: Add `defaultSelector` field to `PermissionClaim` on `APIExport`. When APIBindings are auto-created via `WorkspaceType.defaultAPIBindings`, the selector is now taken from the APIExport's `defaultSelector` instead of defaulting to `matchAll: true`. Existing APIExports without `defaultSelector` retain the previous `matchAll: true` behavior.
```

### DIFF
--- a/pkg/reconciler/apis/permissionclaimlabel/permissionclaimlabel_reconcile.go
+++ b/pkg/reconciler/apis/permissionclaimlabel/permissionclaimlabel_reconcile.go
@@ -201,9 +201,6 @@ func (c *controller) reconcile(ctx context.Context, apiBinding *apisv1alpha2.API
 		unexpectedOrInvalidErrors = append(unexpectedOrInvalidErrors, fmt.Errorf("unexpected/invalid claim for %s.%s (identity %q)", claim.Resource, claim.Group, claim.IdentityHash))
 	}
 
-	// Detect verb and selector mismatches between accepted and exported claims
-	claimMismatches := detectClaimMismatches(expectedClaims, acceptedClaimsMap, exportedClaimsMap, logger)
-
 	if len(unexpectedOrInvalidErrors) > 0 {
 		i := len(unexpectedOrInvalidErrors)
 		if i > 10 {
@@ -221,7 +218,11 @@ func (c *controller) reconcile(ctx context.Context, apiBinding *apisv1alpha2.API
 			len(errsToDisplay.Errors()),
 			errsToDisplay,
 		)
-	} else if len(claimMismatches) > 0 {
+	}
+
+	// Detect verb and selector mismatches between accepted and exported claims
+	claimMismatches := detectClaimMismatches(expectedClaims, acceptedClaimsMap, exportedClaimsMap, logger)
+	if len(claimMismatches) > 0 {
 		// Build mismatch messages
 		mismatchMsgs := make([]string, 0, len(claimMismatches))
 		for _, m := range claimMismatches {
@@ -251,7 +252,9 @@ func (c *controller) reconcile(ctx context.Context, apiBinding *apisv1alpha2.API
 			i,
 			strings.Join(mismatchMsgs[:i], ", "),
 		)
-	} else {
+	}
+
+	if len(unexpectedOrInvalidErrors) == 0 && len(claimMismatches) == 0 {
 		conditions.MarkTrue(apiBinding, apisv1alpha2.PermissionClaimsValid)
 	}
 


mjudeikis:
I think something still missing

olamilekan000:
Just wondering if the documentation can be updated to describe how the permission claim inheritance is meant to work? @mjudeikis

mjudeikis:
@mjudeikis-bot please create an issue to add documentation about Permissionclaims workspacetypes, managed, unmanaged WT type.

cnvergence (approved):

pkg/reconciler/apis/permissionclaimlabel/permissionclaimlabel_reconcile.go
```diff
@@ -214,6 +221,36 @@ func (c *controller) reconcile(ctx context.Context, apiBinding *apisv1alpha2.API
 			len(errsToDisplay.Errors()),
 			errsToDisplay,
 		)
+	} else if len(claimMismatches) > 0 {
```
ntnn: I think this if/elseif/else would be nicer as a switch now, wouldn't it?

ntnn: Also are `unexpectedOrInvalidErrors` and `claimMismatches` really exclusive to each other?

ntnn (changes requested):

pkg/reconciler/apis/permissionclaimlabel/permissionclaimlabel_reconcile.go
mjudeikis: Might be :) Idea was that one is terminal, another is not. But make sense to set both always. changed

mjudeikis:
/retest
both know flakes which we should address separatly

mjudeikis:
/retest
ci no logs

PR #3889, opened 2026-03-09
Split BindRootAPIs and use errgroup.Group

ntnn:
## Summary

Split from https://github.com/kcp-dev/kcp/pull/3847 because something is going wrong and I want to reduce the scope of that PR.

## What Type of PR Is This?

/kind cleanup

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


ntnn:
/test pull-kcp-test-integration

Let's see if this fails more often. If it does we need to add this leak:

```text
goleak.IgnoreTopFunction("k8s.io/apimachinery/pkg/watch.(*StreamWatcher).receive"),
```

PR #3895, opened 2026-03-10
Add loadtesting framework

SimonTheLeg:
## Summary

**This PR adds the loadtesting framework, considerations for review:** 

* `testing/example_test.go` is a good starting point to see how all different aspects of the framework come together
* Updates the Readme to describe the loadtest itself. Old Readme has been moved to Proposal.md
* For the stats we just wrap around https://github.com/montanaflynn/stats, so we can give stats names for the output. But other than that I found it much better to use a well-tested, zero dependency library instead of us going around implementing our own P99 and so on functions
* The framework is kept as minimal as possible, and takes some inspiration from clusterloader2 kubernetes performance testing https://github.com/kubernetes/perf-tests/tree/master/clusterloader2, especially the overall folder structure and the tuningset terminology. It does modernize some aspects of it by using Go iterators for the tuninsets themselves to control concurrency via yielding, which I think results in a bit cleaner code and error handling
* I went against passing configurable options to the tests via flags and instead went with env variables. The idea behind this was mainly that a) we don't need to have custom TestMains everywhere to ensure centralized flags in setup.go + b) we can run things like go test ./... without all test packages requiring the same flags

## What Type of PR Is This?

/kind feature

## Release Notes

```release-note
NONE
```

### DIFF
--- a/test/load/pkg/tuningset/stepped_load_test.go
+++ b/test/load/pkg/tuningset/stepped_load_test.go
@@ -18,6 +18,7 @@ package tuningset
 
 import (
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"github.com/stretchr/testify/require"
@@ -57,41 +58,35 @@ func TestSteppedLoad(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			ts := NewSteppedLoad(tt.burstSize, tt.stepDelay, tt.count, 0)
+			synctest.Test(t, func(t *testing.T) {
+				ts := NewSteppedLoad(tt.burstSize, tt.stepDelay, tt.count, 0)
 
-			// Record the timestamp of each yield to verify burst grouping.
-			yieldTimes := make([]time.Time, 0, tt.count)
-			for range ts {
-				yieldTimes = append(yieldTimes, time.Now())
-			}
-			require.Len(t, yieldTimes, tt.count)
-
-			// Verify items within the same burst are yielded nearly instantly
-			// and that step delays appear between bursts.
-			const burstTolerance = 10 * time.Millisecond
-			const stepTolerance = 50 * time.Millisecond
-
-			for i := 1; i < tt.count; i++ {
-				gap := yieldTimes[i].Sub(yieldTimes[i-1])
-				isBurstBoundary := i%tt.burstSize == 0
-
-				if isBurstBoundary {
-					// Between bursts: expect ~stepDelay.
-					require.InDelta(t,
-						tt.stepDelay.Milliseconds(),
-						gap.Milliseconds(),
-						float64(stepTolerance.Milliseconds()),
-						"gap between item %d and %d should be ~stepDelay (%v), got %v",
-						i-1, i, tt.stepDelay, gap,
-					)
-				} else {
-					// Within a burst: expect near-zero gap.
-					require.Less(t, gap, burstTolerance,
-						"gap within burst between item %d and %d should be <10ms, got %v",
-						i-1, i, gap,
-					)
+				// Record the timestamp of each yield to verify burst grouping.
+				yieldTimes := make([]time.Time, 0, tt.count)
+				for range ts {
+					yieldTimes = append(yieldTimes, time.Now())
 				}
-			}
+				require.Len(t, yieldTimes, tt.count)
+
+				// Verify items within the same burst are yielded instantly
+				// and that step delays appear between bursts.
+				for i := 1; i < tt.count; i++ {
+					gap := yieldTimes[i].Sub(yieldTimes[i-1])
+					isBurstBoundary := i%tt.burstSize == 0
+
+					if isBurstBoundary {
+						require.Equal(t, tt.stepDelay, gap,
+							"gap between item %d and %d should be exactly stepDelay (%v), got %v",
+							i-1, i, tt.stepDelay, gap,
+						)
+					} else {
+						require.Zero(t, gap,
+							"gap within burst between item %d and %d should be 0, got %v",
+							i-1, i, gap,
+						)
+					}
+				}
+			})
 		})
 	}
 }
@@ -106,7 +101,6 @@ func TestSteppedLoadParallel(t *testing.T) {
 		stepDelay      = 100 * time.Millisecond
 		count          = 15 // 3 bursts, 2 pauses
 		actionDuration = 150 * time.Millisecond
-		tolerance      = 100 * time.Millisecond
 	)
 
 	// With parallel execution inside bursts:
@@ -116,30 +110,26 @@ func TestSteppedLoadParallel(t *testing.T) {
 	// stepDelay pause
 	// burst3 yields instantly, all 5 actions start concurrently
 	//
-	// Total wall time ≈ 2*stepDelay + actionDuration = 200ms + 150ms = 350ms
-	//
-	// If sequential, it would be: count * actionDuration + 2*stepDelay
-	// = 15*150ms + 200ms = 2450ms
+	// Total wall time = 2*stepDelay + actionDuration = 200ms + 150ms = 350ms
 	numPauses := (count - 1) / burstSize
 	expDuration := time.Duration(numPauses)*stepDelay + actionDuration
 
-	ts := NewSteppedLoad(burstSize, stepDelay, count, 0)
+	synctest.Test(t, func(t *testing.T) {
+		ts := NewSteppedLoad(burstSize, stepDelay, count, 0)
 
-	var wg wait.Group
-	start := time.Now()
-	for range ts {
-		wg.Start(func() {
-			time.Sleep(actionDuration)
-		})
-	}
-	wg.Wait()
-	elapsed := time.Since(start)
+		var wg wait.Group
+		start := time.Now()
+		for range ts {
+			wg.Start(func() {
+				time.Sleep(actionDuration)
+			})
+		}
+		wg.Wait()
+		elapsed := time.Since(start)
 
-	require.InDelta(t,
-		expDuration.Milliseconds(),
-		elapsed.Milliseconds(),
-		float64(tolerance.Milliseconds()),
-		"elapsed time %v should be close to parallel estimate %v",
-		elapsed, expDuration,
-	)
+		require.Equal(t, expDuration, elapsed,
+			"elapsed time %v should equal parallel estimate %v",
+			elapsed, expDuration,
+		)
+	})
 }
--- a/test/load/pkg/tuningset/uniform_qps_test.go
+++ b/test/load/pkg/tuningset/uniform_qps_test.go
@@ -18,6 +18,7 @@ package tuningset
 
 import (
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"github.com/stretchr/testify/require"
@@ -31,7 +32,6 @@ func TestUniformQPS(t *testing.T) {
 	t.Parallel()
 
 	const count = 10
-	const tolerance = 100 * time.Millisecond
 
 	tests := []struct {
 		name           string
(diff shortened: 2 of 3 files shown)


test/load/pkg/framework/setup.go
```diff
@@ -0,0 +1,114 @@
...
+		if v == "" {
+			missing = append(missing, KCPShardKubeconfig)
+		} else {
+			cfg.ShardKubeconfig = loadKubeconfig(t, v)
+		}
+	}
+
+	if len(missing) > 0 {
+		t.Fatalf("Require: the following required environment variables are not set: %v", missing)
+	}
```
ntnn: Wouldn't it be short to just `require.NotEmpty` `cfg.FrontProxyKubeconfig`?

```go
	if slices.Contains(caps, KCPFrontProxyKubeconfig) {
		cfg.FrontProxyKubeconfig = os.Getenv(string(KCPFrontProxyKubeconfig))
		require.NotEmpty(t, cfg.FrontProxyKubeconfig, "env var must be set", KCPFrontProxyKubeconfig))
	}
```

This should happen early enough that we don't need to check all settings first before failing.

test/load/pkg/framework/setup.go
```diff
@@ -0,0 +1,114 @@
...
+	return cfg
+}
+
+func loadKubeconfig(t *testing.T, kubeconfigPath string) *rest.Config {
+	t.Helper()
+
+	rawConfig, err := clientcmd.LoadFromFile(kubeconfigPath)
+	if err != nil {
+		t.Fatalf("failed to load kubeconfig from %s: %v", kubeconfigPath, err)
+	}
```
ntnn: Similar here - we are in a test harness, I think it's valid to use `require` here, e.g. `require.NoError(t, err, ...)`.

test/load/pkg/framework/setup.go
SimonTheLeg: Hmm I really like the behavior that it immediately tells you **all** the variables that are missing. I expect over time for the list to grow quite large and when you run these tests as a pod somewhere, I think its convenient if the first error message just tells you everything.

What I could offer ot make it a bit more streamlined is something like this. Let me know if you have a strong preference on it over the other one.

```Go
	cfg := &Config{}
	var missing []Capability

	setters := map[Capability]func(*rest.Config){
		KCPFrontProxyKubeconfig: func(c *rest.Config) { cfg.FrontProxyKubeconfig = c },
		KCPShardKubeconfig:      func(c *rest.Config) { cfg.ShardKubeconfig = c },
	}

	for _, cap := range caps {
		v := os.Getenv(string(cap))
		if v == "" {
			missing = append(missing, cap)
			continue
		}
		setters[cap](loadKubeconfig(t, v))
	}

	require.Empty(t, missing, "the following required environment variables are not set: %v", missing)

```

test/load/pkg/framework/setup.go
SimonTheLeg: agreed on this one. Changed.

SimonTheLeg:
/unhold 
I think the E2E are currently borked by our recent rebase k8s 1.35 update, but previously skipped unit tests for the load testing framework are now correctly executed :)
So ready for review

test/load/pkg/framework/setup.go
ntnn: You could also use a combination of assert and a boolean:

```go
	if slices.Contains(caps, KCPFrontProxyKubeconfig) {
		cfg.FrontProxyKubeconfig = os.Getenv(string(KCPFrontProxyKubeconfig))
		assert.NotEmpty(t, cfg.FrontProxyKubeconfig, "env var must be set", KCPFrontProxyKubeconfig))
		anyMissing = true
	}
	// ...
	require.False(t, anyMissing, "missing required components and/or settings")
```

Keeps it streamlined while staying flexible to check if e.g. additional tooling is present if the configuration requires it.

SimonTheLeg: sorry might be Monday brain: But this would not work for the happy path or?
If I have the env set correctly then assert is fine, but we still would set anyMissing to true, even though it is not missing?

ntnn: Ah true :D I wasn't thinking correctly - but you can check on `testing.T` if it is already failed:
https://pkg.go.dev/testing#T.Failed

```go
	if slices.Contains(caps, KCPFrontProxyKubeconfig) {
		cfg.FrontProxyKubeconfig = os.Getenv(string(KCPFrontProxyKubeconfig))
		assert.NotEmpty(t, cfg.FrontProxyKubeconfig, "env var must be set", KCPFrontProxyKubeconfig))
	}
	// ...
	require.False(t, t.Failed(), "missing required components and/or settings")
```

SimonTheLeg: I have to be honest, this does not look great to me:

```
    /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup_test.go:9:
                Error Trace:    /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup.go:89
                                                        /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup_test.go:9
                Error:          Should NOT be empty, but was <nil>
                Test:           TestBla
                Messages:       env var must be set%!(EXTRA framework.Capability=FRONTPROXY_KUBECONFIG)
    /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup_test.go:9:
                Error Trace:    /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup.go:94
                                                        /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup_test.go:9
                Error:          Should NOT be empty, but was <nil>
                Test:           TestBla
                Messages:       env var must be set%!(EXTRA framework.Capability=SHARD_KUBECONFIG)
    /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup_test.go:9:
                Error Trace:    /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup.go:108
                                                        /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup_test.go:9
                Error:          Should be false
                Test:           TestBla
                Messages:       missing required components and/or settings
```

**Proposal to maybe meet in the middle?**

```Go
	if slices.Contains(caps, KCPFrontProxyKubeconfig) {
		v := os.Getenv(string(KCPFrontProxyKubeconfig))
		cfg.FrontProxyKubeconfig = loadKubeconfig(t, v)
		assert.NotEmpty(t, v, KCPFrontProxyKubeconfig+" env var must be set")
	}
```

I still don't like it very much, but at least output is a bit cleaner

```Go
    /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup_test.go:9:
                Error Trace:    /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup.go:90
                                                        /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup_test.go:9
                Error:          Should NOT be empty, but was
                Test:           TestBla
                Messages:       FRONTPROXY_KUBECONFIG env var must be set
    /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup_test.go:9:
                Error Trace:    /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup.go:96
                                                        /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup_test.go:9
                Error:          Should NOT be empty, but was
                Test:           TestBla
                Messages:       SHARD_KUBECONFIG env var must be set
    /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup_test.go:9:
                Error Trace:    /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup.go:108
                                                        /Users/simonbein/code/github/simontheleg/kcp/test/load/pkg/framework/setup_test.go:9
                Error:          Should be false
                Test:           TestBla
                Messages:       missing required components and/or settings
```

ntnn: That's just an interpolation issue^^ You can append it or use `assert.NotEmptyf` 
Both is fine imho

SimonTheLeg: ah I see. Switched out with NotEmptyf 👍

SimonTheLeg:
I think all comments have been addressed ptal

test/load/pkg/framework/setup.go
```diff
@@ -0,0 +1,125 @@
...
+	// var missing []Capability
+
+	// if slices.Contains(caps, KCPFrontProxyKubeconfig) {
+	// 	v := os.Getenv(string(KCPFrontProxyKubeconfig))
+	// 	if v == "" {
+	// 		missing = append(missing, KCPFrontProxyKubeconfig)
+	// 	} else {
+	// 		cfg.FrontProxyKubeconfig = loadKubeconfig(t, v)
+	// 	}
+	// }
```
ntnn:
```suggestion
```

test/load/pkg/framework/setup.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// if slices.Contains(caps, KCPShardKubeconfig) {
+	// 	v := os.Getenv(string(KCPShardKubeconfig))
+	// 	if v == "" {
+	// 		missing = append(missing, KCPShardKubeconfig)
+	// 	} else {
+	// 		cfg.ShardKubeconfig = loadKubeconfig(t, v)
+	// 	}
+	// }
+
```
ntnn:
```suggestion
```

test/load/pkg/measurement/duration.go
```diff
@@ -0,0 +1,28 @@
...
+
+import "time"
+
+// RecordElapsedDurationMS is a convenience function to handle duration measurements.
+// You can call it in your test using:
+//
+//	defer measurement.RecordElapsedDurationMS(time.Now(), sink)
+func RecordElapsedDurationMS(start time.Time, sink Sink) {
+	duration := time.Since(start)
+	sink.Drop(Measurement{Name: "duration_ms", Value: float64(duration.Milliseconds())})
```
ntnn:
```suggestion
	sink.Drop(Measurement{Name: "duration_ms", Value: duration.Seconds() * 1000})
```

`Duration.Milliseconds` is a precision loss because it truncates to integer instead of keeping the fractional.

Makefile
```diff
@@ -410,7 +410,7 @@ test: ## Run tests
 	for MOD in $$(git ls-files '**/go.mod' | sed 's,/go.mod,,'); do \
 		if [ "$$MOD" != "." ]; then \
 			echo "Testing $$MOD module..."; \
-			(cd $$MOD && $(GO_TEST) -race $(COUNT_ARG) -coverprofile=coverage.txt -covermode=atomic $(TEST_ARGS) $(WHAT)); \
+			(cd $$MOD && $(GO_TEST) -race $(COUNT_ARG) -coverprofile=coverage.txt -covermode=atomic $(TEST_ARGS) $$(go list "$(WHAT)" | grep -v 'test/load/testing')); \
```
ntnn, line 413: Do we really need this? `test/load` is its own go module So I'd think it enters `test/load` and then runs `go test -race ...`
It's just that this looks very hacky =/

test/load/pkg/tuningset/stepped_load_test.go
```diff
@@ -0,0 +1,145 @@
...
+	wg.Wait()
+	elapsed := time.Since(start)
+
+	require.InDelta(t,
+		expDuration.Milliseconds(),
+		elapsed.Milliseconds(),
+		float64(tolerance.Milliseconds()),
+		"elapsed time %v should be close to parallel estimate %v",
+		elapsed, expDuration,
+	)
```
ntnn: I feel like this would flake hard in our CI. Can we try using synctest to not rely on the system clock for these tests?
https://pkg.go.dev/testing/synctest

test/load/pkg/tuningset/uniform_qps_test.go
```diff
@@ -0,0 +1,79 @@
...
+			wg.Wait()
+			elapsed := time.Since(start)
+
+			require.InDelta(t,
+				tt.expDuration.Milliseconds(),
+				elapsed.Milliseconds(),
+				float64(tolerance.Milliseconds()),
+				"elapsed time %v should be close to estimate %v",
+				elapsed, tt.expDuration,
+			)
```
ntnn: Get the same flaky feeling here as in the randomized qps test.

Makefile
SimonTheLeg: the problem is that go test -race would run the load tests themselves, which cannot (and should not) be executed for unit testing

test/load/pkg/measurement/duration.go
SimonTheLeg: oh. Very good spot!

test/load/pkg/tuningset/stepped_load_test.go
SimonTheLeg: ooooh true we are now on Go 1.25 since the rebase and can use this directly. I previously pondered using a clock as an argument to the tuningsets so it could be mocked, but did not like the interface. But with this we'll get it super clean + we can even check for equality since clocks are deterministic + test passes much quicker with the default ticks of the fake clock. Nice! 🥳

SimonTheLeg:
forgot to write: I think all comments have been addressed ptal

PR #3898, opened 2026-03-12
strip debug info in container images

xrstf:
## Summary
This reduces the total size of the kcp container image from 723 to 529 MB (26% reduction). When running many e2e tests for the kcp-operator, I have since begun to preload kcp into kind so I do not have to re-pull 723MB every time. That gave me the thought of maybe also making the images just smaller if we can.

I don't think we need debugging symbols in our production kcp images. Is there someone out there who regularly uses delve to remotely debug a kcp?

## What Type of PR Is This?
/kind cleanup

## Release Notes
```release-note
All kcp binaries in the container images now have their debugging symbols stripped, saving roughly 25% in total image size.
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,7 +56,7 @@ ARG TARGETARCH
 
 RUN --mount=type=cache,target=/root/.cache/go-build \
     --mount=type=cache,target=/go/pkg/mod \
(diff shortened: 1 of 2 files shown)


ntnn (approved):
/lgtm
/approve

PR #3900, opened 2026-03-12
Bookmark events in terminating VW

ntnn:
## Summary

ref: https://github.com/platform-mesh/security-operator/issues/402

When watches are set up on the terminating VW with bookmarks the client logs an error with `awaiting required bookmark event for initial events stream, no events received for 10.00949897s`.
This was because the filter in the storage wrapper only allowed items with a deletion timestamp, which bookmark events don't have.

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/test/e2e/virtual/terminatingworkspaces/virtualworkspace_test.go
+++ b/test/e2e/virtual/terminatingworkspaces/virtualworkspace_test.go
@@ -686,6 +686,91 @@ func TestTerminatingWorkspacesVirtualWorkspaceWatch(t *testing.T) {
 	}
 }
 
+func TestTerminatingWorkspacesVirtualWorkspaceWatchBookmark(t *testing.T) {
+	t.Parallel()
+	framework.Suite(t, "control-plane")
+
+	source := kcptesting.SharedKcpServer(t)
+	wsPath, _ := kcptesting.NewWorkspaceFixture(t, source, core.RootCluster.Path())
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	t.Cleanup(cancelFunc)
+
+	sourceConfig := source.BaseConfig(t)
+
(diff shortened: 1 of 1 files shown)


test/e2e/virtual/terminatingworkspaces/virtualworkspace_test.go
```diff
@@ -689,6 +686,91 @@ func TestTerminatingWorkspacesVirtualWorkspaceWatch(t *testing.T) {
 	}
 }
 
+func TestTerminatingWorkspacesVirtualWorkspaceWatchBookmark(t *testing.T) {
```
SimonTheLeg, line 689: Instead of writing another test, couldn't we just add `AllowWatchBookmarks: true,` to the existing test and then check for bookmark events as well? My current understanding is that these events get auto-created for clients which have specified the bookmark capability - or putting it another way, from client side there is no way to forcefully trigger these events, right?

https://github.com/ntnn/kcp/blob/d6e285fbc3b6776d1049c253d75f35ea0c303d2f/test/e2e/virtual/terminatingworkspaces/virtualworkspace_test.go#L620

ntnn: Correct; I was thinking about that last night as well but I think it makes sense to have a dedicated test for it - on one hand because it's quite important and on the other if we make another VW and look at other tests we have a focus on also specifically checking bookmark events

mjudeikis: I'm starting to lean towards smaller tests. CI is already running a lot, and having smaller tests per feature makes it easier to identify flakes.

Issue #3905, opened 2026-03-13
cleanup: e2e: validate usage of `require.*` assertions in Eventually


gman0:
Some of the e2e tests use `require.*` assertions in their `Eventually()` , pre-maturely ending the test on transient errors.

See e.g. https://github.com/kcp-dev/kcp/pull/3902

Check all occurrences and see which may be changed into retrials.

PR #3907, opened 2026-03-16
e2e: tolerate transient errors in Eventually

gman0:
## Summary

There are a few e2e tests that fail on errors that may be only transient. This PR adds a helper function `TolerateOrFail` depending on the supplied predicates and tolerates/fails accordingly. It's being called where appropriate: as far as I could tell, these are the last places that were left after the other `Eventually()` VS `require.*` cleanups we had.

## What Type of PR Is This?

/kind cleanup

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3905

## Release Notes

```release-note
NONE
```

### DIFF
--- a/test/e2e/virtual/replication/virtualworkspace_test.go
+++ b/test/e2e/virtual/replication/virtualworkspace_test.go
@@ -325,6 +325,7 @@ func TestCachedResourceVirtualWorkspace(t *testing.T) {
 			}
 			return true, ""
 		}, wait.ForeverTestTimeout, time.Millisecond*100, "expected user-1 to watch sheriffs")
+		defer sheriffWatch.Stop()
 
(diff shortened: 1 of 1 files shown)


gman0:
Also, I think our error handling in tests could use some work. E.g. it seems, as time went on, the different [`kcptestinghelpers.Eventually`](https://github.com/kcp-dev/kcp/blob/301a8f749e7b99a0c81f43b37aa5b5e5ff0fc0b4/staging/src/github.com/kcp-dev/sdk/testing/helpers/eventually.go#L35), `require.Eventually`, and  `require.EventuallyWithT` retry-able functions have been used, and it's a bit of a mess. IMHO we should pick one and stick with it.

I'd be really happy if someone more knowledgeable around testing frameworks could give their few cents and say if there are any ideas worth borrowing.

PR #3931, opened 2026-03-20
Try prevent flaking by placing org and ws on same shard

ntnn:
## Summary

`create and clean workspace` has been flaking due to eventually timeouts, while the other subtest hasn't flaked. The other subtest places the org and workspace on the root shard explicitly, this test didn't.

My guess is that the cross-shard operation (plus all the other tests) is just enough delay for the test to fail every now and then. It's worth a try - at worst the two tests will have a similar setup.

## What Type of PR Is This?

/kind flake

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/test/e2e/reconciler/workspacedeletion/controller_test.go
+++ b/test/e2e/reconciler/workspacedeletion/controller_test.go
@@ -62,7 +62,12 @@ func TestWorkspaceDeletion(t *testing.T) {
 		{
 			name: "create and clean workspace",
 			work: func(ctx context.Context, t *testing.T, server runningServer) {
-				orgPath, _ := kcptesting.NewWorkspaceFixture(t, server, core.RootCluster.Path(), kcptesting.WithType(core.RootCluster.Path(), "organization"))
+				// The kcptesting.WithRootShard was added due to the
+				// test flaking in CI, it is not actually required for
+				// the test execution. However we observed that this
+				// subtest was flaking while the next subtest - which is
+				// pinning the workspaces to the root shard - was not.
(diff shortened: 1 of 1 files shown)


test/e2e/reconciler/workspacedeletion/controller_test.go
```diff
@@ -62,7 +62,7 @@ func TestWorkspaceDeletion(t *testing.T) {
 		{
 			name: "create and clean workspace",
 			work: func(ctx context.Context, t *testing.T, server runningServer) {
-				orgPath, _ := kcptesting.NewWorkspaceFixture(t, server, core.RootCluster.Path(), kcptesting.WithType(core.RootCluster.Path(), "organization"))
+				orgPath, _ := kcptesting.NewWorkspaceFixture(t, server, core.RootCluster.Path(), kcptesting.WithRootShard(), kcptesting.WithType(core.RootCluster.Path(), "organization"))
```
gman0, line 70: Maybe add a comment for the rationale behind this change? Something like in the PR's description would suffice.

ntnn:
@gman0 I like the suggestion; I also put in a comment in the second test. I think we could refactor both at some point to not require pinning the shard in either.

ntnn:
/retest

The failures are all related to the APIBindings not being ready yet, see: https://github.com/kcp-dev/kcp/pull/3934

ntnn:
@gman0 @mjudeikis please re-lgtm

ntnn:
/retest

Known flake create_a_workspace_without_an_explicit_type

PR #3938, opened 2026-03-25
Add kcp restart test

### LINKED ISSUE #3918: bug: WatchList breaks restarting a kcp instance with an initialized etcd storage
### Describe the bug

When restarting a kcp instance with WatchList enabled kcp fails to finish bootstrapping.
The error is a bit opaque but effectively the shard resource is missing because the APIBindings are missing, which bind the kcp APIs into the root workspace. That prevents the bootstrap from finishing and kcp never becoming ready.

- [x] Fix the WatchList error
- [x] Add a test verifying that kcp being restarted with the same etcd storage still works

For the test I think the integration tests can be used, they allow explicitly stopping and starting the kcp server.

### Steps To Reproduce

1. Revert https://github.com/kcp-dev/kcp/pull/3917
2. Start kcp
3. Stop kcp
4. Start kcp again with the same `.kcp`
5. Run: `kubectl ws tree ; kubectl get shards ; kubectl get apibindings.apis.kcp.io`

`ws tree` should work, `get shards` will fail (because the shards API is missing) and `get apibindings` will return nothing.

### Expected Behaviour

1. Revert https://github.com/kcp-dev/kcp/pull/3917
2. Start kcp
3. Stop kcp
4. Start kcp again with the same `.kcp`
5. Run: `kubectl ws tree ; kubectl get shards ; kubectl get apibindings.apis.kcp.io`

```text
>k ws tree ; k get shards ; k get apibindings.apis.kcp.io
.
└── root

NAME   REGION   URL                      EXTERNAL URL             AGE
root            https://127.0.0.1:6443   https://127.0.0.1:6443   61s
NAME                 AGE   READY
shards.core.kcp.io   71s   True
tenancy.kcp.io       71s   True
topology.kcp.io      71s   True

```

### Additional Context

_No response_

ntnn:
## Summary

See #3918 - all of our tests are greenfield tests, so we never checked that a kcp instance still works when booting from initialized storage.

The fixtures don't really expose any meaningful way to restart an instance and I feel adding this complexity just for one test is a bit much at the moment.

Instead this is copying most of the setup from the scaffolding to make sure restarting a kcp instance works. If the need arises in other areas where the ability to restart a fixture is required that can still be changed.

## What Type of PR Is This?

/kind cleanup

## Related Issue(s)

Fixes #3918

## Release Notes

```release-note
NONE
```

### DIFF
--- a/test/e2e/server/restart_test.go
+++ b/test/e2e/server/restart_test.go
@@ -1,0 +1,218 @@
+/*
+Copyright 2026 The kcp Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package server
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strconv"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


ntnn:
/retest

That looks like an infra fail:

```text
     testing.go:1369: TempDir RemoveAll cleanup: unlinkat /tmp/TestKcpRestart3485379012/001/etcd-server/member/wal: directory not empty 
```

ntnn:

Let's see. This should work but we might have to increase the waits?

PR #3992, opened 2026-04-03
Fix: APIBinding read-modify-write race via Server-Side Apply

### LINKED ISSUE #3926: bug: APIBinding reconciler crash after first write produces ordering-dependent LogicalCluster recovery state
**Background**

**Describe the bug**

**Expected Behaviour**

**Proposed Fix**

**Additional Context**

**Versions**

swastik959:
## Summary
This PR fixes the read-modify-write race condition that causes LogicalCluster conditions to diverge when the APIBinding reconciler crashes mid-flight.

Previously, controllers were using Merge Patch on stale caches, blindly overwriting each other's status conditions. Furthermore, the APIBinding reconciler was locking the cluster before saving its phase intent, leading to a corrupted intermediate state upon crash recovery.

- Migrated to Server-Side Apply (SSA): Added NewSSACommitter in committer.go to handle status updates using ApplyPatchType instead of MergePatchType.
- Fixed TypeMeta Serialization: Explicitly defined APIVersion and Kind with json tags in the Resource struct to bypass Go's encoding/json dropping inline structs, ensuring the API server accepts the SSA patch.
- Targeted LogicalCluster Updates: Updated updateLogicalCluster in apibinding_controller.go to only patch specific KCP lock annotations using SSA, rather than uploading the entire cluster object and wiping out sibling conditions.
- Crash-Proofed Reconciler: Modified newReconciler in apibinding_reconcile.go to return reconcileStatusStopAndRequeue when transitioning to the Binding phase. This guarantees the intent is committed to etcd before it attempts to lock resources or generate CRDs.

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #3926 

## Release Notes

```release-note
NONE
```
## Testing Done

- Built and verified KCP locally.
- Created a wildwest-provider workspace and applied the cowboys APIExport and schemas.
- Created a wildwest-consumer workspace and applied the APIBinding.
- Verified that the APIBinding successfully reached phase: Bound and that all status conditions (Ready, APIExportValid, InitialBindingCompleted, PermissionClaimsApplied, etc.) merged flawlessly via SSA without overwriting one another.

### DIFF
--- a/pkg/reconciler/committer/committer.go
+++ b/pkg/reconciler/committer/committer.go
@@ -35,8 +35,10 @@ import (
 
 // Resource is a generic wrapper around resources so we can generate patches.
 type Resource[Sp any, St any] struct {
+	APIVersion        string `json:"apiVersion"`
+	Kind              string `json:"kind"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
-	Spec              Sp `json:"spec"`
+	Spec              Sp `json:"spec,omitempty"`
 	Status            St `json:"status,omitempty"`
 }
 
@@ -135,7 +137,7 @@ func generatePatchAndSubResources[Sp any, St any](old, obj *Resource[Sp, St]) ([
 	name := old.Name
 
 	oldForPatch := forPatch(old)
-	// to ensure they appear in the patch as preconditions
+
 	oldForPatch.UID = ""
 	oldForPatch.ResourceVersion = ""
 
@@ -145,7 +147,7 @@ func generatePatchAndSubResources[Sp any, St any](old, obj *Resource[Sp, St]) ([
 	}
 
 	newForPatch := forPatch(obj)
-	// to ensure they appear in the patch as preconditions
+
 	newForPatch.UID = old.UID
 	newForPatch.ResourceVersion = old.ResourceVersion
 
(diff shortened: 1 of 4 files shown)


pkg/reconciler/committer/committer.go
```diff
@@ -166,3 +168,84 @@ func generatePatchAndSubResources[Sp any, St any](old, obj *Resource[Sp, St]) ([
 
 	return patchBytes, subresources, nil
 }
+
+func NewSSACommitter[R runtime.Object, P Patcher[R], Sp any, St any](patcher ClusterPatcher[R, P], fieldManager string) CommitFunc[Sp, St] {
```
mjudeikis, line 172: While I like this version if commiter, it needs tests, for commiter and controller, it changes as it's quite sensitive. 

Will come back to this after holidays (tuesday) as it needs more reading ;)

mjudeikis:
@swastik959 will you clean this?

swastik959:
@mjudeikis sure will do it

mjudeikis:
@swastik959 I would recommend split this into 2 prs - adding a new committer - we would merge this fast + tests, and second to use it.

swastik959:
@mjudeikis sure

mjudeikis:
@swastik959, please split this into 2 PRs. commiter changes + commiter tests, and fix itself (once first merged)

mjudeikis:
replaced by https://github.com/kcp-dev/kcp/pull/4004

PR #3998, opened 2026-04-07
Identity-based cached resources

### LINKED ISSUE #3827: feature: add support for replicating arbitrary resources to the cache server
### Feature Description

Currently, we have a hard-coded list of resources that the cache-server knows about, and there is no way to add other user-defined resources.

> Adding New Resources
> 
> Not implemented at the moment. Our near-term plan is to maintain a list of hard-coded resources that we want to keep in the cache server. In the future, we will use the ReplicationClam which will describe schemas that need to be exposed by the cache server.
>
> https://docs.kcp.io/kcp/main/concepts/sharding/cache-server/#adding-new-resources

There are strong use cases for when we would want arbitrary resources to be replicated and accessible across shards.

### Proposed Solution

First option is CachedResources API: it already handles the use-cases around replication of arbitrary resources, the basics are there. There are two hurdles to overcome:
* Consumption: CachedResources relies on its sister resource CachedObjects on the cache server side, making it not suitable for general k8s clients. We'd need to expose the actual GVK of the referenced resource in the CachedResource.
* Schema: the resource referenced by CachedResource is schema-less. We need to somehow make this work on the client side.
  * We can assume that the owner of the CachedResource is responsible for the resource, including the schema.
  * The CachedResource has `.status.identityHash`, we can use this and pair GVR+hash to distinguish between resources.

Second option is customizing cached-server itself. We'd need a special "claim" resource anyway (similar to what the docs pasted above says). The pros is we have more flexibility over what we need, cons is we already have CachedResources that fall into the same category of use-cases.

### Alternative Solutions

_No response_

### Want to contribute?

### Additional Context

_No response_

gman0:
## Summary

This PR refactors the Replication Virtual Workspace (VW) and the CachedResource replication subsystem to use identity-based resource addressing on the cache server, completely replacing the earlier CachedObject wrapper approach.

Check the e2e tests for usage.

  API changes

  - CachedResourceEndpointSlice gains a required, immutable spec.export field (ExportBindingReference) that links an endpoint slice to its owning APIExport. A new APIExportValid condition reflects whether the referenced export exists and references back to the slice.
  - CachedObject type is removed. The cache server no longer stores user-defined resources as opaque CachedObject wrappers; instead, resources are stored directly under an identity-prefixed GVR (e.g.
  cpuflavors:<identityHash>), matching how built-in resources are handled.

  Replication reconciler

  - The per-CachedResource replication controller now appends the CachedResource's identity hash to the GVR resource name when interacting with the cache server (resource + ":" + identity), so cached
  copies of user-defined resources are addressable by identity just like other bound-CRD resources.
  - GenCachedObjectName and associated CachedObject-specific label/annotation constants are removed; identity-prefixed GVRs carry the necessary disambiguation.
  - The reconciler only watches partial metadata and updates objects in cache when the original RV changes. This works if this is the only reconciler that writes to cache.
    - This may or may not be a premature optimization \o/ 
    - We can switch to real object comparisons -- at the price of watching full objects.

  Replication Virtual Workspace

  - The builder is significantly simplified: shard-level CachedObject lookup and the apidomainkey-based URL parsing are replaced by a straightforward forwarding approach that reads directly from the
  cache server via a dedicated cacheDynamicClusterClient.
  - A new forwarding.go encapsulates the read-only REST storage provider for the Replication VW, wiring each CachedResource's schema and identity into a forwardingregistry-backed handler pointed at the
  cache server.
  - API reconciler controllers (controllers/apireconciler) now own the per-CachedResource API definition lifecycle within the VW, replacing the previous flat list of endpoint-driven definitions.
  - **REMOVED:** Wildcard request are no longer accepted. Let me know if you disagree but IMHO they didn't make much sense. Yes it is nice to be semantically correct and behave like a regular cluster-aware apiserver, on the other hand do we need it? It can be added later if we can find a real use case for it.

  Cache server

  - crd_lister.go is extended to synthesize CRDs for active CachedResources on the fly, making user-defined replicated resources visible to the cache server's API machinery without requiring explicit CRD
   registration.
  - A second partial-metadata DDSIF (CachePartialMetadataDDSIF) is introduced, pointed at the cache server, so the main kcp server can watch cache-side resources for informer-driven controllers without
  using full objects.

  Tests & docs

  - New e2e test (test/e2e/cache/replication_api_cache_test.go) covering the end-to-end CachedResource replication and Replication VW serving path.
  - Existing virtual workspace and cachedresources e2e tests updated to reflect the new export field requirement and identity-based addressing.
  - Docs updated: cached-resources.md expanded with use-case framing, corrected CachedResourceEndpointSlice YAML examples, and the stale "Adding New Resources" section in cache-server.md updated to
  reference the CachedResource API.

^^ PR summary largely AI-generated, but IMHO not super verbose and quite helpful. This is a pretty large PR but I tried to split the different areas of changes into dedicated commits. I highly recommend reviewing commit by commit (though they do have a bit of inter-dependence on each other).

## What Type of PR Is This?

/kind api-change

## Related Issue(s)

Fixes #3827 

## Release Notes

```release-note
- CachedResourceEndpointSlice requires a cross-reference to APIExport it is being exported by
- CachedResourceEndpointSlice are no longer created automatically
- Objects replicated by CachedResources can be retrieved from the cache server API using their real GVR:identity
```


gman0:
I've added DNM because there is a go.mod change that is just for testing. Will drop once https://github.com/kcp-dev/kubernetes/pull/190 and https://github.com/kcp-dev/kubernetes/pull/191 is merged.

pkg/cache/server/crd_lister.go
```diff
@@ -18,29 +18,84 @@ package server
...
-	// since all available CRDs are stored in bootstrap.SystemCRDLogicalCluster
-	// and there is only a single registry per a CRD
-	// there is no need to filter by the given cluster
+func (c *crdClusterLister) Cluster(cluster logicalcluster.Name) kcp.ClusterAwareCRDLister {
 	return &crdLister{
 		crdClusterLister: c,
-		cluster:          bootstrap.SystemCRDLogicalCluster,
+		cluster:          cluster,
 	}
 }
```
mjudeikis, line 100: Did this change? bootstrap.SystemCRDLogicalCluster ?

pkg/cache/server/crd_lister.go
```diff
@@ -54,22 +109,315 @@ type crdLister struct {
...
+	return types.UID(strings.ToLower(base36.EncodeBytes(h[:])))
+}
+
+// synthesizeCRD builds a CRD from a set of CachedResources that share the same identity+GR.
+func (c *crdClusterLister) synthesizeCRDForCachedResources(crs []*cachev1alpha1.CachedResource) (*apiextensionsv1.CustomResourceDefinition, error) {
+	gr := schema.GroupResource{
+		Group:    crs[0].Spec.Group,
+		Resource: crs[0].Spec.Resource,
+	}
+	kind := crs[0].Annotations[cachedresources.AnnotationResourceKind]
```
mjudeikis, line 356: Why only the first element?

mjudeikis (commented):
wow. This PR is HUGE! 
Im not so much worried about cached resource implementation, but more about the cache server changes.

pkg/cache/server/crd_lister.go
```diff
@@ -54,22 +109,315 @@ type crdLister struct {
...
+		matchingIdentity := identity == "" || cr.Status.IdentityHash == identity
+		if matchingIdentity && cr.Spec.Group == group && cr.Spec.Resource == resource {
+			matchingCRs, err := c.listCachedResourcesByIdentityAndGR(cr.Status.IdentityHash, gr)
+			if err != nil {
+				return nil, err
+			}
+			if len(matchingCRs) == 0 {
+				return nil, apierrors.NewNotFound(apiextensionsv1.Resource("customresourcedefinitions"), name)
+			}
+			return c.synthesizeCRDForCachedResources(crs)
```
gman0:
```suggestion
			return c.synthesizeCRDForCachedResources(matchingCRs)
```

pkg/cache/server/crd_lister.go
gman0: So in general this CRD lister is very similar to the one we have in https://github.com/kcp-dev/kcp/blob/main/pkg/server/apiextensions.go , specifically in that with APIBindings, the CRD lister/getter is built based on the APIBindings in the cluster. Similarly, the lister/getter is built using the CachedResources in the cluster.

Cache-server's old CRD lister had a hardcoded cluster name, but now it actually respects the cluster in request and lists from that cluster (but prepends the built-in CRDs first).

pkg/cache/server/crd_lister.go
gman0: All `crs` have the same GVR+identity, so their Kind should be same too. So it doesn't really matter which CR this is picked from.

gman0:
The cache server CRD lister builds one synthetic CRD per all CachedResources that have matching GR+Identity. It gathers all these CachedResources, grabs their Versions and builds a schemaless CRD from that.

Two issues with that:

1. The storage version on that synthetic CRD is chosen as `max(MatchingCachedResources[].Version)`, but in fact it really depends on the CachedResource.Spec.Version that did the replication for that specific cluster. AFAIK this should be okay -- the owner should know the real schema and how to convert.

2. We may have a problem later down the line if, in a single cluster, there are multiple CachedResources with the same GR but different versions. There already is an [admission plugin](https://github.com/kcp-dev/kcp/blob/main/pkg/admission/cachedresource/admission.go) that ensures there can be only one CachedResource per GVR per cluster. We may need to expand that to one per GR. But then schema upgrades can't be done without recreating the CachedResource :/ Depending on the use-case, this can be smaller or a bigger problem.

gman0:
Also I'm still hunting down the bug that caused the test failure.

pkg/reconciler/cache/cachedresources/cachedresources_reconcile_replication.go
```diff
@@ -79,13 +81,17 @@ func (r *replication) reconcile(ctx context.Context, cachedResource *cachev1alph
 			return reconcileStatusStopAndRequeue, nil
 		}
 
-		// Global informer is based on the CachedResource type and we construct index based on the schema labels.
 		controllerCtx, cancel := context.WithCancel(ctx)
 
-		global := r.cacheKcpInformers.Cache().V1alpha1().CachedObjects()
+		global, err := r.globalDiscoveringDynamicKcpInformers.ForResource(gvr)
```
gman0, line 86: TODO: check that the CRD-GVR-source (which the CachePartialMetadataDDSIF uses) stops the shared informer once the CRD goes away.

mjudeikis:
/lgtm
/approve

release was cut, so we can land this and start testing

PR #4002, opened 2026-04-08
Feat: Added SSA commiter

### LINKED ISSUE #3926: bug: APIBinding reconciler crash after first write produces ordering-dependent LogicalCluster recovery state

swastik959:
## Summary
This PR introduces Server-Side Apply (SSA) support to the committer package by adding a new NewSSACommitter utility.

Currently, KCP controllers rely heavily on JSON Merge Patch (types.MergePatchType) for status updates. When multiple controllers update different status conditions on the same object concurrently based on a stale cache read, it results in a read-modify-write race where conditions are blindly overwritten.
## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3926

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/committer/committer.go
+++ b/pkg/reconciler/committer/committer.go
@@ -35,8 +35,10 @@ import (
 
 // Resource is a generic wrapper around resources so we can generate patches.
 type Resource[Sp any, St any] struct {
+	APIVersion        string `json:"apiVersion,omitempty"`
+	Kind              string `json:"kind,omitempty"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
-	Spec              Sp `json:"spec"`
+	Spec              Sp `json:"spec,omitempty"`
 	Status            St `json:"status,omitempty"`
 }
 
@@ -135,7 +137,7 @@ func generatePatchAndSubResources[Sp any, St any](old, obj *Resource[Sp, St]) ([
 	name := old.Name
 
 	oldForPatch := forPatch(old)
-	// to ensure they appear in the patch as preconditions
+
 	oldForPatch.UID = ""
 	oldForPatch.ResourceVersion = ""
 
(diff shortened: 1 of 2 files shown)


pkg/reconciler/committer/committer.go
```diff
@@ -166,3 +168,84 @@ func generatePatchAndSubResources[Sp any, St any](old, obj *Resource[Sp, St]) ([
 
 	return patchBytes, subresources, nil
 }
+
+func NewSSACommitter[R runtime.Object, P Patcher[R], Sp any, St any](patcher ClusterPatcher[R, P], fieldManager string) CommitFunc[Sp, St] {
```
mjudeikis, line 179: The code comment is missing, explaining what this will do. Please make sure it has note about force=true

pkg/reconciler/committer/committer_test.go
```diff
@@ -133,3 +133,119 @@ func TestGeneratePatchAndSubResources(t *testing.T) {
...
+				ObjectMeta: metav1.ObjectMeta{Finalizers: []string{"changed"}},
+				Status:     corev1alpha1.LogicalClusterStatus{Phase: "new"},
+			},
+		},
+	}
+
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			require.Panics(t, func() {
+				// Fixed signature
```
mjudeikis: this comment has no value

pkg/reconciler/committer/committer_test.go
```diff
@@ -133,3 +133,119 @@ func TestGeneratePatchAndSubResources(t *testing.T) {
...
+				_, _, _ = generateSSAPatchAndSubResources(tc.old, tc.new)
+			})
+		})
+	}
+}
+
+func TestGenerateSSAPatchAndSubResources(t *testing.T) {
```
mjudeikis, line 187: I would like to see a more complex test, too. Now all of them are changing single value. 
We need to prove it patches values when multiple exist, while keeping old ones (Culprit of the issue this should fix with follow-up pr. At best would be to mock the status of the logical cluster and use that to test

mjudeikis:
/lgtm
/approve

Thanks for this!

PR #4004, opened 2026-04-08
Fix: APIBinding read-modify-write race via Server-Side Apply

### LINKED ISSUE #3926: bug: APIBinding reconciler crash after first write produces ordering-dependent LogicalCluster recovery state

swastik959:
## Summary

This PR fixes the read-modify-write race condition that causes LogicalCluster and APIBinding conditions to diverge and overwrite each other, especially during crash recovery or concurrent reconciler runs.

Previously, the APIBinding controller relied on JSON Merge Patch and stale informer caches. If the controller crashed mid-flight, or if DefaultAPIBindingLifecycleController updated the cluster concurrently, intermediate conditions would be blindly erased. Furthermore, the reconciler was attempting to lock the cluster and generate CRDs before safely committing its initial Phase: Binding intent to the database.

Building on the infrastructure introduced in the SSA Committer PR, this PR implements the following fixes:

- Migrated APIBinding to SSA: Replaced the legacy MergePatch committer with NewSSACommitter for the APIBinding resource, using an explicit FieldManager.
- Targeted LogicalCluster Locks: Updated updateLogicalCluster to use Server-Side Apply (ApplyPatchType). By explicitly managing the payload, the reconciler now only asserts ownership over its specific locking annotations and conditions, rather than uploading the entire cluster object and wiping out sibling conditions.
- Crash-Proofed State Machine: Modified newReconciler to return reconcileStatusStopAndRequeue immediately after transitioning the object to Phase: Binding. This guarantees the intent is durably committed to etcd before the controller attempts heavy, conflict-prone operations like resource locking or CRD generation.
- Test Updates: Updated apibinding_reconcile_test.go to align with the new state machine expectations (expecting the requeue and appropriately deferring the Ready condition check).   

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #3926 

## Release Notes

```release-note
NONE
```
## Testing Done

- Built and verified KCP locally.
- Created a wildwest-provider workspace and applied the cowboys APIExport and schemas.
- Created a wildwest-consumer workspace and applied the APIBinding.
- Verified that the APIBinding successfully reached phase: Bound and that all status conditions (Ready, APIExportValid, InitialBindingCompleted, PermissionClaimsApplied, etc.) merged flawlessly via SSA without overwriting one another.

### DIFF
--- a/pkg/reconciler/apis/apibinding/apibinding_controller.go
+++ b/pkg/reconciler/apis/apibinding/apibinding_controller.go
@@ -63,7 +63,8 @@ import (
 )
 
 const (
-	ControllerName = "kcp-apibinding"
+	ControllerName    = "kcp-apibinding"
+	LocksFieldManager = "apibinding-reconciler-locks"
 )
 
 var (
@@ -165,8 +166,10 @@ func NewController(
 					Kind:       "LogicalCluster",
 				},
 				ObjectMeta: metav1.ObjectMeta{
-					Name:        lc.Name,
-					Annotations: lc.Annotations,
+					Name: lc.Name,
+					Annotations: map[string]string{
+						ResourceBindingsAnnotationKey: lc.Annotations[ResourceBindingsAnnotationKey],
+					},
 				},
 			}
 			patchBytes, err := json.Marshal(patchObj)
@@ -179,7 +182,7 @@ func NewController(
 				types.ApplyPatchType,
 				patchBytes,
 				metav1.PatchOptions{
-					FieldManager: "apibinding-reconciler-locks",
+					FieldManager: LocksFieldManager,
 					Force:        ptr.To(true),
 				},
 			)
--- a/pkg/reconciler/apis/apibinding/apibinding_controller.go
+++ b/pkg/reconciler/apis/apibinding/apibinding_controller.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
 
 	kcpcache "github.com/kcp-dev/apimachinery/v2/pkg/cache"
 	kcpapiextensionsclientset "github.com/kcp-dev/client-go/apiextensions/client"
@@ -68,11 +69,6 @@ const (
 var (
 	SystemBoundCRDsClusterName = logicalcluster.Name("system:bound-crds")
 )
-
-// boolPtr returns a pointer to a boolean value.
-func boolPtr(b bool) *bool {
-	return &b
-}
 
 // NewController returns a new controller for APIBindings.
 func NewController(
@@ -163,15 +159,17 @@ func NewController(
 			return logicalClusterInformer.Lister().Cluster(name).Get(corev1alpha1.LogicalClusterName)
 		},
 		updateLogicalCluster: func(ctx context.Context, lc *corev1alpha1.LogicalCluster) error {
-			// FIXED: Pass the actual annotations from the LogicalCluster, not an empty map.
-			patchBytes, err := json.Marshal(map[string]interface{}{
-				"apiVersion": "core.kcp.io/v1alpha1",
-				"kind":       "LogicalCluster",
-				"metadata": map[string]interface{}{
-					"name":        lc.Name,
-					"annotations": lc.Annotations,
+			patchObj := &corev1alpha1.LogicalCluster{
+				TypeMeta: metav1.TypeMeta{
+					APIVersion: corev1alpha1.SchemeGroupVersion.String(),
+					Kind:       "LogicalCluster",
 				},
-			})
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        lc.Name,
+					Annotations: lc.Annotations,
+				},
+			}
+			patchBytes, err := json.Marshal(patchObj)
 			if err != nil {
 				return err
 			}
@@ -182,7 +180,7 @@ func NewController(
 				patchBytes,
 				metav1.PatchOptions{
 					FieldManager: "apibinding-reconciler-locks",
-					Force:        boolPtr(true),
+					Force:        ptr.To(true),
 				},
 			)
 			return err
--- a/pkg/reconciler/apis/apibinding/apibinding_controller.go
+++ b/pkg/reconciler/apis/apibinding/apibinding_controller.go
@@ -18,6 +18,7 @@ package apibinding
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"time"
 
@@ -28,6 +29,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -66,6 +68,11 @@ const (
 var (
 	SystemBoundCRDsClusterName = logicalcluster.Name("system:bound-crds")
 )
+
+// boolPtr returns a pointer to a boolean value.
+func boolPtr(b bool) *bool {
+	return &b
+}
 
 // NewController returns a new controller for APIBindings.
 func NewController(
@@ -156,12 +163,35 @@ func NewController(
 			return logicalClusterInformer.Lister().Cluster(name).Get(corev1alpha1.LogicalClusterName)
 		},
 		updateLogicalCluster: func(ctx context.Context, lc *corev1alpha1.LogicalCluster) error {
-			_, err := kcpClusterClient.CoreV1alpha1().LogicalClusters().Cluster(logicalcluster.From(lc).Path()).Update(ctx, lc, metav1.UpdateOptions{})
+			// FIXED: Pass the actual annotations from the LogicalCluster, not an empty map.
+			patchBytes, err := json.Marshal(map[string]interface{}{
+				"apiVersion": "core.kcp.io/v1alpha1",
+				"kind":       "LogicalCluster",
+				"metadata": map[string]interface{}{
+					"name":        lc.Name,
+					"annotations": lc.Annotations,
+				},
+			})
+			if err != nil {
+				return err
+			}
+			_, err = kcpClusterClient.CoreV1alpha1().LogicalClusters().Cluster(logicalcluster.From(lc).Path()).Patch(
+				ctx,
+				lc.Name,
+				types.ApplyPatchType,
+				patchBytes,
+				metav1.PatchOptions{
+					FieldManager: "apibinding-reconciler-locks",
+					Force:        boolPtr(true),
+				},
+			)
 			return err
 		},
 		deletedCRDTracker: newLockedStringSet(),
-		commit:            committer.NewCommitter[*APIBinding, Patcher, *APIBindingSpec, *APIBindingStatus](kcpClusterClient.ApisV1alpha2().APIBindings()),
-	}
+		commit: committer.NewSSACommitter[*APIBinding, Patcher, *APIBindingSpec, *APIBindingStatus](
+			kcpClusterClient.ApisV1alpha2().APIBindings(),
+			ControllerName,
+		)}
 
 	logger := logging.WithReconciler(klog.Background(), ControllerName)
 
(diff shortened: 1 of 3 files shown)


pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -492,8 +522,20 @@ func (c *controller) process(ctx context.Context, key string) (bool, error) {
...
-	newResource := &Resource{ObjectMeta: binding.ObjectMeta, Spec: &binding.Spec, Status: &binding.Status}
+	oldResource := &Resource{
+		APIVersion: "apis.kcp.io/v1alpha2",
+		Kind:       "APIBinding",
+		ObjectMeta: old.ObjectMeta,
+		Spec:       &old.Spec,
+		Status:     &old.Status,
+	}
+	newResource := &Resource{
+		APIVersion: "apis.kcp.io/v1alpha2",
```
mjudeikis: Just checking, how this would work (or won't work) if we're operating on `v1alpha1` ? We still have that api available. Need test for this.

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -156,12 +163,35 @@ func NewController(
 			return logicalClusterInformer.Lister().Cluster(name).Get(corev1alpha1.LogicalClusterName)
 		},
 		updateLogicalCluster: func(ctx context.Context, lc *corev1alpha1.LogicalCluster) error {
-			_, err := kcpClusterClient.CoreV1alpha1().LogicalClusters().Cluster(logicalcluster.From(lc).Path()).Update(ctx, lc, metav1.UpdateOptions{})
+			// FIXED: Pass the actual annotations from the LogicalCluster, not an empty map.
```
mjudeikis: comment with no value

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -67,6 +69,11 @@ var (
 	SystemBoundCRDsClusterName = logicalcluster.Name("system:bound-crds")
 )
 
+// boolPtr returns a pointer to a boolean value.
+func boolPtr(b bool) *bool {
```
mjudeikis: use [k8s](https://pkg.go.dev/k8s.io/utils/ptr)

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -156,12 +163,35 @@ func NewController(
 			return logicalClusterInformer.Lister().Cluster(name).Get(corev1alpha1.LogicalClusterName)
 		},
 		updateLogicalCluster: func(ctx context.Context, lc *corev1alpha1.LogicalCluster) error {
-			_, err := kcpClusterClient.CoreV1alpha1().LogicalClusters().Cluster(logicalcluster.From(lc).Path()).Update(ctx, lc, metav1.UpdateOptions{})
+			// FIXED: Pass the actual annotations from the LogicalCluster, not an empty map.
+			patchBytes, err := json.Marshal(map[string]interface{}{
```
mjudeikis: Now this is slippery slope. If this controller ever starts updating more than annotations in the code, it will start failing silently. we need some better safeguard for this logic

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -156,12 +163,35 @@ func NewController(
...
+				"apiVersion": "core.kcp.io/v1alpha1",
+				"kind":       "LogicalCluster",
```
mjudeikis: Not sure I like that we hardcoding versions, this will make version changing harder. do we have static values of these in sdk?

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -492,8 +522,20 @@ func (c *controller) process(ctx context.Context, key string) (bool, error) {
 	}
 
 	// If the object being reconciled changed as a result, update it.
-	oldResource := &Resource{ObjectMeta: old.ObjectMeta, Spec: &old.Spec, Status: &old.Status}
-	newResource := &Resource{ObjectMeta: binding.ObjectMeta, Spec: &binding.Spec, Status: &binding.Status}
+	oldResource := &Resource{
+		APIVersion: "apis.kcp.io/v1alpha2",
```
mjudeikis: same comment as above with versions

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -134,7 +134,8 @@ func (r *newReconciler) reconcile(ctx context.Context, apiBinding *apisv1alpha2.
 		"Waiting for API(s) to be established",
 	)
 
-	return reconcileStatusContinue, nil
+	// FIXED: Stop and requeue so the intent state commits BEFORE dangerous ops begin
```
mjudeikis: useless comment.

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
```diff
@@ -134,7 +134,8 @@ func (r *newReconciler) reconcile(ctx context.Context, apiBinding *apisv1alpha2.
...
+	return reconcileStatusStopAndRequeue, nil
```
mjudeikis, line 137: So does one exit from this reconciler? If it's always stopped? This looks wrong. Or feels wrong,.

mjudeikis (changes requested):
quite a few comments :)

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
swastik959: The newReconciler only executes when `Phase == ""`. By returning `StopAndRequeue`, we commit the Phase: Binding transition to etcd. The immediate requeue then routes the object into the bindingReconciler on the very next loop. Not sure if there is some other component getting affected though, any lead ?

pkg/reconciler/apis/apibinding/apibinding_controller.go
swastik959: Not sure I guess the code for controller is explicitly designed for interacting v1alpha2 
`type APIBinding = apisv1alpha2.APIBinding` I have an Idea that api server both in KCp and kubernetes is able to convert from v1alpha1 to v1alpha2 or am I missing something ?

mjudeikis:
@swastik959 Some comments are not addressed. If something is not clear, please ask :)

swastik959:
@mjudeikis I have commented on them not sure if they are visible or not 
Screenshot from 2026-04-15 13-46-02
Screenshot from 2026-04-15 13-45-50

mjudeikis:
Its pending, you need to "submit your review" via review tab 
image

Github has 2 modes for comments:
1. Every comment is posted individually
2. Bulk post with full review (preferable) so all at once.

2nd helps with big reviews when you leave a comment and 2 files down, you get this "Oh, this is why this was done this way", and you can delete the comment as it's not yet posted :)

swastik959:
@mjudeikis Thanks, was not aware about it :sweat_smile:

pkg/reconciler/apis/apibinding/apibinding_controller.go
mjudeikis: So the problem with this is that if somebody changes api 2 years later, this will not be cached on compile time. We tend to lose things like this in time :) so use api is preferable.

swastik959: we can Just use 
`apisv1alpha2.SchemeGroupVersion.String()` for that case I believe instead of a simple string

pkg/reconciler/apis/apibinding/apibinding_reconcile.go
swastik959: @mjudeikis any leads on this ?

mjudeikis: This sounds right, just need to read a code to remind myself.

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -156,12 +159,37 @@ func NewController(
...
 		updateLogicalCluster: func(ctx context.Context, lc *corev1alpha1.LogicalCluster) error {
-			_, err := kcpClusterClient.CoreV1alpha1().LogicalClusters().Cluster(logicalcluster.From(lc).Path()).Update(ctx, lc, metav1.UpdateOptions{})
+			patchObj := &corev1alpha1.LogicalCluster{
+				TypeMeta: metav1.TypeMeta{
+					APIVersion: corev1alpha1.SchemeGroupVersion.String(),
+					Kind:       "LogicalCluster",
+				},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:        lc.Name,
+					Annotations: lc.Annotations,
```
mjudeikis, line 169: This patch will claim all annotations. I think it should just do one we care, else it basically takes over the supervision of all annotations.

it should own only one it needs to

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -156,12 +159,37 @@ func NewController(
...
+				return err
+			}
+			_, err = kcpClusterClient.CoreV1alpha1().LogicalClusters().Cluster(logicalcluster.From(lc).Path()).Patch(
+				ctx,
+				lc.Name,
+				types.ApplyPatchType,
+				patchBytes,
+				metav1.PatchOptions{
+					FieldManager: "apibinding-reconciler-locks",
+					Force:        ptr.To(true),
```
mjudeikis, line 183: I wonder if we make this only annotation owned, do we need force here?

mjudeikis:
In addition there are other reconcilers writing LC status:
```
`kcp-logicalcluster` is a merge-patch committer and a real candidate to race on LogicalCluster.Status.
Workspace tenancy controllers under `pkg/reconciler/tenancy/` also write LogicalCluster.Status (phase, initializers, conditions) and should be audited.
```

Basically this fix work only if all places are changed.

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -156,12 +159,37 @@ func NewController(
...
+			if err != nil {
+				return err
+			}
+			_, err = kcpClusterClient.CoreV1alpha1().LogicalClusters().Cluster(logicalcluster.From(lc).Path()).Patch(
+				ctx,
+				lc.Name,
+				types.ApplyPatchType,
+				patchBytes,
+				metav1.PatchOptions{
+					FieldManager: "apibinding-reconciler-locks",
```
mjudeikis, line 182: Move this close to controller name ontop:
```
FieldManager = "apibinding-reconciler-locks"
```

mjudeikis (commented):
Sorry, this takes a while. This is a VERY sensitive write path. So we need to get this right, and it's not straightforward. Every round of reviews is like gizza piramide digging and looking for secret chambers. 

But we are on the right path :)

pkg/reconciler/apis/apibinding/apibinding_controller.go
swastik959: I tested dropping Force: ptr.To(true), but it results in immediate SSA conflicts: Apply failed with 1 conflict: conflict with "kcp" using core.kcp.io/v1alpha1: .metadata.annotations.internal.apis.kcp.io/resource-bindings.

Because the core "kcp" field manager (or mutating webhooks) frequently touches the LogicalCluster annotations, our apibinding-reconciler-locks manager must use Force: true to successfully acquire shared ownership of this specific key during the binding process without throwing conflict errors.

swastik959: @mjudeikis please feel free to review the changes and comment whenever you feel right

mjudeikis: checking again. sorry, was bit swamped

mjudeikis:
last comment :D and we can merge

**Test gap on the new `StopAndRequeue` ordering**

`TestReconcileNew` covers reconcile #1 (Phase=Binding, no Ready). `TestReconcileBinding` covers reconcile #2 with a hand-crafted `Phase=Binding` input. Nothing chains them on the same object.

That means someone could revert [`apibinding_reconcile.go:137`](https://github.com/kcp-dev/kcp/blob/apibinding/commiter/pkg/reconciler/apis/apibinding/apibinding_reconcile.go#L137) back to `reconcileStatusContinue`, or add a `MarkFalse(Ready, ...)` before the early return, and both existing tests still pass. The "Phase commits before conditions" guarantee this PR is built around is unguarded.

Could you add a two-call sequence test? Something like:

```go
// reconcile #1 — fresh binding, no Phase
requeue, err := c.reconcile(ctx, apiBinding)
require.NoError(t, err)
require.True(t, requeue)
require.Equal(t, apisv1alpha2.APIBindingPhaseBinding, apiBinding.Status.Phase)
require.False(t, conditions.Has(apiBinding, conditionsv1alpha1.ReadyCondition))

// reconcile #2 — same object, simulate the requeue
_, err = c.reconcile(ctx, apiBinding)
require.NoError(t, err)
require.True(t, conditions.Has(apiBinding, conditionsv1alpha1.ReadyCondition))
```

Small and mechanical, but it pins the ordering the PR introduces.

swastik959:
@mjudeikis Done

mjudeikis (approved):
/lgtm
/approve

PR #4009, opened 2026-04-09
fix(server): isolate virtual workspace proxy impersonation per request

### LINKED ISSUE #4008: bug: external virtual workspace proxy can leak impersonation between concurrent requests
### Describe the bug

When `kcp` is configured to proxy `/services/...` requests to an external virtual workspace server, `pkg/server/handler.go:WithVirtualWorkspacesProxy` mutates a shared `httputil.ReverseProxy` by replacing its `Transport` on every request with a request-specific impersonating round tripper.

Because the same `ReverseProxy` instance is reused across requests, concurrent proxied requests can race on that shared `Transport` field. In practice, request A can install an `alice` impersonation transport, request B can replace it with `bob`, and then request A can be served with `bob`'s impersonation headers.

That means the external virtual workspace can observe the wrong `Impersonate-*` headers for a request, which breaks correctness and can attribute authorization/audit behavior to the wrong user.

### Steps To Reproduce

1. Run `kcp` with an external virtual workspace server so `/services/*` requests are handled through `WithVirtualWorkspacesProxy` (for example by setting `--shard-virtual-workspace-url` plus the corresponding client cert/key and CA flags).
2. Send two concurrent authenticated requests from different users to the same proxied virtual workspace endpoint.
3. Interleave the requests so one pauses after swapping `proxy.Transport` but before calling `proxy.ServeHTTP`, then let the second request replace the transport and complete.
4. Observe that the first request can reach the external virtual workspace with the second user's `Impersonate-*` headers.

A deterministic unit reproduction can be built by blocking the handler at the `logger.V(4).Info(...)` call, which executes after the transport swap and before `ServeHTTP`.

### Expected Behaviour

Each proxied request should use request-local proxy state so that impersonation headers cannot leak across concurrent requests.

### Additional Context

Root cause is the shared assignment in `pkg/server/handler.go`:

- `proxy.Transport = clientgotransport.NewImpersonatingRoundTripper(...)`

`httputil.ReverseProxy` reads `Transport` when `ServeHTTP` starts, so mutating that shared field in the handler is not concurrency-safe.

This affects deployments that run virtual workspaces out-of-process and rely on the shard to proxy `/services/...` traffic to that external server.

officialasishkumar:
## Summary

Stop mutating a shared `httputil.ReverseProxy` when proxying shard `/services/...` requests to an external virtual workspace server. Each request now uses a request-scoped proxy copy with its own impersonating transport, and `pkg/server/handler_test.go` adds a deterministic regression test for the exact transport-swap race window.

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #4008

## Release Notes

```release-note
Fix external virtual workspace proxying so concurrent `/services/...` requests keep impersonation headers request-scoped.
```

## Usage of AI

used AI in coding assistance and fixing grammatical errors while commenting.

### DIFF
--- a/pkg/server/handler.go
+++ b/pkg/server/handler.go
@@ -89,7 +89,11 @@ func WithVirtualWorkspacesProxy(apiHandler http.Handler, shardVirtualWorkspaceUR
 			return
 		}
 
-		proxy.Transport = clientgotransport.NewImpersonatingRoundTripper(
+		// ReverseProxy instances are shared across requests, so the transport must
+		// be request-scoped to avoid leaking impersonation headers between
+		// concurrent requests.
+		requestScopedProxy := *proxy
+		requestScopedProxy.Transport = clientgotransport.NewImpersonatingRoundTripper(
 			clientgotransport.ImpersonationConfig{
 				UserName: user.GetName(),
 				UID:      user.GetUID(),
@@ -100,7 +104,7 @@ func WithVirtualWorkspacesProxy(apiHandler http.Handler, shardVirtualWorkspaceUR
 		)
 
 		logger.V(4).Info("proxying virtual workspace", "target", req.URL.String())
-		proxy.ServeHTTP(w, req)
+		requestScopedProxy.ServeHTTP(w, req)
 	}
 }
 
--- a/pkg/server/handler_test.go
+++ b/pkg/server/handler_test.go
@@ -1,0 +1,205 @@
+/*
+Copyright 2026 The kcp Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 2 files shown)


mjudeikis:
/ok-to-test 

Nice fix :) thanks for this!

mjudeikis:
@officialasishkumar lint is failing :)

officialasishkumar:
@mjudeikis Fixed the lint issue in aa44e9379 by using `http.NoBody` in the test request. `go test ./pkg/server -run TestWithVirtualWorkspacesProxyUsesRequestScopedImpersonationTransport -count=1` and `golangci-lint run ./pkg/server` both pass locally.

mjudeikis:
@officialasishkumar DCO needs to be signed for all commits. If you use AI assistance, usually this needs to be instructed. This is standard for the entire Kube ecosystem

officialasishkumar:
@mjudeikis updated the description

mjudeikis (approved):
/lgtm
/approve

PR #4011, opened 2026-04-10
🐛 Etcd key poisoning fix

mjudeikis:
## Summary

Fixes an etcd key-corruption bug in `WithLocalProxy`: when a multi-segment workspace path (e.g. root:internal-cluster) could not be resolved against the shard's local index, the fallback stuffed the raw path string into `request.Cluster.Name` and forwarded the request. For sufficiently privileged callers this reached storage, and `NoNamespaceKeyRootFunc `concatenated the path verbatim into the etcd key, producing orphaned rows like `/registry/<group>/<resource>/customresources/root:internal-cluster/...` next to the legitimate hash-keyed rows. These rows are invisible to normal reads (which resolve path → hash), survive workspace deletion, and leak via wildcard partial-metadata lists. The race is triggered by any cold/stale index state: informer not yet synced after restart, wrong-shard routing, scheduling lag.

### Changes:

- `pkg/server/localproxy.go`: the unresolvable-path fallback now returns 404 NotFound for the workspace instead of forwarding with a poisoned context.
- `pkg/server/filters/filters.go`: new defense-in-depth filter WithClusterNameShapeInvariant that rejects any request whose context carries a path-shaped cluster.Name (bare names, wildcards, and system:* still pass). Wired in after `WithLocalProxy` on the main shard chain and after WithClusterScope on the cache server chain.
- Regression tests for both the localproxy fallback and the new filter.
- No behaviour change on any legitimate request path. Existing orphaned rows must be cleaned up out-of-band with etcdctl.

| Client type                                                              | Canonical path still works?                                                  |
|--------------------------------------------------------------------------|------------------------------------------------------------------------------|
| kubectl / client-go going through the front-proxy                        | Yes, unchanged                                                               |
| kubectl / client-go going directly to the correct shard                  | Yes, unchanged                                                               |
| kubectl / client-go going directly to the wrong shard                    | No (now 404 instead of silent corruption — this was always broken, invisibly)|
| Controller / informer watching a live workspace                          | Yes, unchanged                                                               |
| Controller cleaning up resources in an LC whose Workspace is gone        | Should use the hash; if it used the path, it was relying on the bug          |

related to https://kubernetes.slack.com/archives/C021U8WSAFK/p1775749028395829 

## What Type of PR Is This?
/kind bug
/kind regression

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Fix an etcd key-corruption bug where an unresolvable multi-segment workspace path in `/clusters/<path>/...` on a shard could cause resources to be written to etcd under a key segment containing the raw workspace path instead of the logical cluster name, producing orphaned rows invisible to the normal API read path. The shard now returns 404 for unresolvable workspace paths, and a new defense-in-depth filter rejects any request whose context carries a path-shaped cluster name before it can reach storage.

```

### DIFF
--- a/hack/etcd-list-poisoned-keys.sh
+++ b/hack/etcd-list-poisoned-keys.sh
@@ -1,0 +1,187 @@
+#!/usr/bin/env bash
+
+# Copyright 2026 The kcp Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# List etcd keys that look like they were corrupted by the workspace-path
+# poisoning bug: the local proxy used to stuff multi-segment workspace paths
+# (e.g. "root:internal-cluster") verbatim into the cluster-name segment of
+# the etcd key, producing rows that are invisible to the normal read path
+# but still consume space and can leak via wildcard partial-metadata lists.
+#
+# This script is READ-ONLY. It never modifies or deletes keys.
+#
+# Usage:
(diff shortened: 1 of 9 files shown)


ntnn (approved):
/retest

flakes

/approve
/lgtm

mjudeikis:
want to add script to check keys

ntnn (approved):
/lgtm
/approve

PR #4015, opened 2026-04-12
Add more GC tests

ntnn:
## Summary

I have a PR ready that gets the upstream GC to be cluster-aware with not-as-many-as-expected-but-still-a-lot-of modifications.
Since fiddling with such essential things like the GC requires rigorous testing I added a bunch more tests to validate our GC at the moment.

I want to merge the tests first to:
a) see if any of them start to flake in CI 
b) validate their correctness in use
c) show that the modifications aren't influencing the tests

After the next release I'll PR the new GC.

## What Type of PR Is This?

/kind chore

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


ntnn:
```text
 Linting test/load module...
test/load/pkg/stats/stats.go:20:2: could not import github.com/montanaflynn/stats (pkg/stats/stats.go:20:2: github.com/montanaflynn/stats@v0.7.1: Get "http://athens-proxy.athens.svc.cluster.local/github.com/montanaflynn/stats/@v/v0.7.1.zip": dial tcp 10.96.192.241:80: connect: connection refused) (typecheck)
	"github.com/montanaflynn/stats"
	^
1 issues:
* typecheck: 1 
```

infra problem^^

/test pull-kcp-lint

test/e2e/garbagecollector/apis_test.go
```diff
@@ -0,0 +1,191 @@
...
+	kcptestinghelpers "github.com/kcp-dev/sdk/testing/helpers"
+
+	"github.com/kcp-dev/kcp/config/helpers"
+	wildwestv1alpha1 "github.com/kcp-dev/kcp/test/e2e/fixtures/wildwest/apis/wildwest/v1alpha1"
+	wildwestclientset "github.com/kcp-dev/kcp/test/e2e/fixtures/wildwest/client/clientset/versioned/cluster"
+	"github.com/kcp-dev/kcp/test/e2e/framework"
+)
+
+// TestBoundAPICrossWorkspace verifies that objects from the same bound
+// API in different workspaces but same names do not confuse the gc.
```
gman0, line 53: Can you reword this comment? It's a bit unclear.

gman0:
In general there are a few tests that do the same thing, only for different types of resources (built-in, CRD, and bound). At the same time it's not exercising all code paths for them equally.

E.g. `TestBoundAPICrossWorkspace` sort of implies that UID could be ignored when resolving owner references. `TestGarbageCollectorNormalCRDs` does that too. There is no such test for built-in APIs though. Why? If the answer is, "because we are already doing it for CRDs and bound resources" -- then what's special about these two? Why not just CRD?

It would be nice if we had a general skeleton of a specific GC feature that we want to test, and then run them through all types of resources we have. And then have such skeletons for all features we are interested in testing. Deduplicates the code, and covers more.

gman0:
Secondly, I think it would be good to have a test that churns through more (10?) objects. This is to exercise locking is correct.

gman0:
Other than that, this is a nice improvement - but a few opportunities weren't taken IMHO. If none of my comments above make sense, it's good to go too.

ntnn:
> In general there are a few tests that do the same thing, only for different types of resources (built-in, CRD, and bound). At the same time it's not exercising all code paths for them equally.
> E.g. `TestBoundAPICrossWorkspace` sort of implies that UID could be ignored when resolving owner references. `TestGarbageCollectorNormalCRDs` does that too. There is no such test for built-in APIs though. Why? If the answer is, "because we are already doing it for CRDs and bound resources" -- then what's special about these two? Why not just CRD?

That's true, some of the tests are preexisting and I didn't want to touch those but wanted to add some more scenarios (e.g. the one with bound APIs - there's already one that tests basically the same thing but with the same CRD in different workspaces).

But I agree that it's sometimes very duplicated. Some of the duplication comes from the pre-existing test that I didn't want to touch :D 
  
> It would be nice if we had a general skeleton of a specific GC feature that we want to test, and then run them through all types of resources we have. And then have such skeletons for all features we are interested in testing. Deduplicates the code, and covers more.

I think I can see that, making some generic scenarios and then feed the workspaces and types in, that way scenarios can be tested within and across workspaces? At least for scenarios that can be reflected with bound APIs.

> [...] there's already one that tests basically the same thing but with the same CRD in different workspaces).

At the end the resource itself is not that important for the GC and the different tests for CRD and bound APIs are just to ensure the GC doesn't confuse objects from the same API between workspaces. Ofc that shouldn't happen in the first place - but with moving to a cluster-aware GC we have to make sure the objects don't shadow each other in the caches of the GC.

gman0:
Yes I think we're just making the same point :D

For the sake of keeping this PR "small", as it currently is, I'm for merging, but I do want to note that there is dedup and restructuring work to be done, and this PR adds on top of that.

/lgtm

mjudeikis:
/approve
reviews were done. tagging

PR #4030, opened 2026-04-14
fix(apibinding): Avoid updating cached LogicalCluster on no-op reconcile

### LINKED ISSUE #4028: bug: LogicalCluster modified in client cache
### Describe the bug

Found in e2e run https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/4027/pull-kcp-test-e2e/2043985938412998656

### Steps To Reproduce

N/A

### Expected Behaviour

Don't modify objects in cache.

### Additional Context

_No response_

ifdotpy:
## Summary

- avoid updating a `LogicalCluster` when APIBinding reconciliation does not change resource-binding locks
- fix the permission-claims-only APIBinding path used by `logical-clusters`, where `apiExport.Spec.Resources` is empty and reconcile previously issued a no-op `LogicalCluster` update against the informer-cached object
- add a regression test that reproduces the cached `LogicalCluster` mutation by simulating a client update path that normalizes `TypeMeta`

## What Type of PR Is This?

/kind bug
/kind regression
/kind failing-test

## Related Issue(s)

Fixes #4028

## Release Notes

```release-note
NONE
```

## Validation

- `go test ./pkg/reconciler/apis/apibinding`
- `go test ./pkg/reconciler/apis/apibinding -run TestReconcileDoesNotMutateCachedLogicalClusterWhenLocksUnchanged`
- `go test ./test/e2e/apibinding -run TestAPIBindingLogicalCluster -count=5`
- `go test ./test/e2e/apibinding -run TestAPIBindingCRDs -count=3`

### DIFF
--- a/pkg/reconciler/apis/apibinding/apibinding_reconcile_test.go
+++ b/pkg/reconciler/apis/apibinding/apibinding_reconcile_test.go
@@ -193,6 +193,7 @@ func TestReconcileBinding(t *testing.T) {
 		existingAPIBindings       []*apisv1alpha2.APIBinding
 		logicalCluster            *corev1alpha1.LogicalCluster
 		updateLogicalClusterError error
+		mutateTypeMetaOnUpdate    bool
 
 		// reconcile result
 		wantError   bool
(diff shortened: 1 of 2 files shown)


pkg/reconciler/apis/apibinding/apibinding_reconcile_test.go
```diff
@@ -796,6 +796,69 @@ func TestReconcileBinding(t *testing.T) {
 	}
 }
 
+func TestReconcileDoesNotMutateCachedLogicalClusterWhenLocksUnchanged(t *testing.T) {
```
mjudeikis: This test is bit watter'y. We should be testing cases where "we expect it to be mutated, one where we dont".
and should we just extend `TestReconcileBinding` with 2 more test cases?

mjudeikis (changes requested):
I think tests need changing.

PR #4033, opened 2026-04-15
[wip] cached resource flakes

mjudeikis:
## Summary

Working theory that cache server breaks down on lazy initialized resources and it fails in 3s window for waiting for "state to be up to date" 

This is not the fix - this is to prove theory

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


mjudeikis:
fix identified here https://github.com/kcp-dev/kcp/pull/4037

PR #4037, opened 2026-04-16
Fix CachedResources flakes

mjudeikis:
## Summary

In general, failures with cached resources were due to multiple tests using the same GK (not GVR), which caused shared informers to get confused. We were getting errors as the wrong schema was being picked up (kcp logs shows "missing namespace on schema" errors). To make it worse, other GVR running in parallel tests were in namespaces. So we had a namespace vs cluster scope clash. 

I will create a follow-up issue as we have to move more of those to identity-based indexing as its an issue not only for cached resources but in general. 

## What Type of PR Is This?
/kind bug
/kind cleanup

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/cache/cachedresources/cachedresources_reconcile_replication.go
+++ b/pkg/reconciler/cache/cachedresources/cachedresources_reconcile_replication.go
@@ -62,7 +62,6 @@ func (r *replication) reconcile(ctx context.Context, cachedResource *cachev1alph
 		Resource: cachedResource.Spec.Resource,
 	}
 	cluster := logicalcluster.From(cachedResource)
-	cacheGVR := cacheReplicationGVR(cachedResource)
 
 	var resourceLabelSelector labels.Selector
 	if cachedResource.Spec.LabelSelector != nil {
@@ -84,13 +83,9 @@ func (r *replication) reconcile(ctx context.Context, cachedResource *cachev1alph
 
 		controllerCtx, cancel := context.WithCancel(ctx)
 
-		// Replicated objects are stored and addressed on the backing cache API under
-		// resource:identityHash. The global informer must watch that identity-scoped GVR,
-		// otherwise updates can be misclassified as creates and get dropped after an
-		// AlreadyExists response.
-		global, err := r.globalDiscoveringDynamicKcpInformers.ForResource(cacheGVR)
+		global, err := r.globalDiscoveringDynamicKcpInformers.ForResource(gvr)
 		if err != nil {
-			logger.Error(err, "Failed to get global informer for resource", "resource", cacheGVR)
+			logger.Error(err, "Failed to get global informer for resource", "resource", gvr)
 			cancel()
 			return reconcileStatusStopAndRequeue, err
 		}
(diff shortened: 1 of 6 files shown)


pkg/reconciler/cache/cachedresources/cachedresources_reconcile_replication.go
```diff
@@ -83,9 +84,13 @@ func (r *replication) reconcile(ctx context.Context, cachedResource *cachev1alph
 
 		controllerCtx, cancel := context.WithCancel(ctx)
 
-		global, err := r.globalDiscoveringDynamicKcpInformers.ForResource(gvr)
+		// Replicated objects are stored and addressed on the backing cache API under
+		// resource:identityHash. The global informer must watch that identity-scoped GVR,
+		// otherwise updates can be misclassified as creates and get dropped after an
+		// AlreadyExists response.
+		global, err := r.globalDiscoveringDynamicKcpInformers.ForResource(cacheGVR)
```
gman0: That's wrong ^ This is a partial metadata informer, so `:identity` should not be part of gvr.

mjudeikis: But this means there is still an issue, if at the cache server level it's stored with identities, but queried without?

gman0: This is not working without it? If so then there is problem in the crd lister in pkg/cache/server. But on paper this should have worked.

mjudeikis: let me revert this. I been working on cachedResource flake, and got bit sidetracked into VW flake too.

pkg/reconciler/cache/cachedresources/cachedresources_reconcile_replication.go
```diff
@@ -143,16 +138,8 @@ func (r *replication) reconcile(ctx context.Context, cachedResource *cachev1alph
 		go func() {
 			defer cancel()
 
-			// Only require the local informer to sync before starting the controller.
```
mjudeikis, line 146: This is revert of previous "fix"

pkg/reconciler/cache/cachedresources/cachedresources_reconcile_replication.go
gman0: Oops didn't see your msg.

> But this means there is still an issue, if at the cache server level it's stored with identities, but queried without?

Yes that's fine, partial metadata sets up the storage prefix such that it's skipped. This is the same mechanism that we already have with bound resources through exports/bindings.

mjudeikis:
/test all
passes consistently now

gman0:
@xrstf wait, i think this might need a closer look.

/hold

xrstf:
/lgtm cancel

oops, sorry.

gman0:
Will check again today.

ntnn (approved):
/approve
/lgtm cancel

gman0:
Just for the record I think we are still off from the real fix.

> In general, failures with cached resources were due to multiple tests using the same GK (not GVR), which caused shared informers to get confused.

This should not matter. The resources are in different workspaces, so they should get different keys no?

> I will create a follow-up issue as we have to move more of those to identity-based indexing as its an issue not only for cached resources but in general.

^^

PR #4043, opened 2026-04-17
fix wrong initializer being used for nested workspace types

### LINKED ISSUE #4029: spec.initializer: true on a derived WorkspaceType doesn't add an initializer to new instances
### Question / possible bug

When `spec.initializer: true` is set on a WorkspaceType created via `spec.extend.with` (i.e. a derived type), new workspaces of that derived type are NOT created with an initializer marker — they go directly to phase `Ready`.

Setting `spec.initializer: true` on the **base** type (the one being extended) DOES work — new workspaces of any derived type get `status.initializers: ["<base-cluster>:<base-name>"]`.

### Reproduction

```
root:universal           spec.initializer: true   ← built-in
  ↑ extended by
root:account             (no own initializer)
  ↑ extended by
root:orgs:showroom-acc   (no own initializer)
```

Case A — patch `root:account` only:

```sh
kubectl --context=:root apply --server-side -f - <<APPLY
kind: WorkspaceType
metadata:
  name: account
spec:
  initializer: true
APPLY

# Create a workspace under root:orgs:showroom of type showroom-acc
# -> status.initializers: ["root:account"]    works as expected
```

Case B — patch `showroom-acc` only (the leaf, derived type):

```sh
kubectl --context=:root:orgs apply --server-side -f - <<APPLY
kind: WorkspaceType
metadata:
  name: showroom-acc
spec:
  initializer: true
APPLY

# Create a workspace of type showroom-acc
# -> phase=Ready immediately, no initializers   unexpected
```

### Expected behaviour

If `spec.initializer` is honored at all on derived types, setting it should produce an initializer marker named `<derived-cluster>:<derived-name>` on new workspaces of that derived type. This is how we'd expect to scope initializers per-WorkspaceType (e.g. per-org account types in a multi-tenant control plane), so different agents can own different types.

### Why this matters

Without this, agents that want to selectively initialize only specific derived types must rely on watching the parent type globally, which forces them to fire on every account everywhere — losing the per-type / per-org granularity the WorkspaceType extend mechanism is supposed to enable.

### Environment

- Setup: cc-d2 deployment of Platform Mesh
- KCP version: please ask if needed

xrstf:
## What Type of PR Is This?
/kind bug

## Related Issue(s)
Fixes #4029

## Release Notes
```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling_test.go
+++ b/pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling_test.go
@@ -592,3 +592,172 @@ func shardNameToBase36Sha224(name string) string {
 	base36hash := strings.ToLower(base36.EncodeBytes(hash[:]))
 	return base36hash[:8]
 }
+
+// TestLogicalClustersInitializers_DerivedTypeInNestedWorkspace tests the scenario from
+// https://github.com/kcp-dev/kcp/issues/4029 where a derived WorkspaceType in a nested
+// workspace has spec.initializer: true. The initializer name must use the canonical path
+// from the kcp.io/path annotation, not the physical cluster name.
+func TestLogicalClustersInitializers_DerivedTypeInNestedWorkspace(t *testing.T) {
+	// Scenario:
+	// - "base" type lives in root:org (canonical path), physical cluster "abc123"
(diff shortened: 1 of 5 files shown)


embik:
Strong -1 on this, see my comment: https://github.com/kcp-dev/kcp/issues/4029#issuecomment-4268599141

PR #4046, opened 2026-04-17
extend loadtesting framework with reports and workspacetrees

SimonTheLeg:
## Summary

This PR improves the loadtesting framework by adding proper connection pooling, setting a user-agent, and a new report structure. All features can be seen in action in the updated example.

## What Type of PR Is This?

/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


test/load/go.mod
```diff
@@ -3,6 +3,8 @@ module github.com/kcp-dev/kcp/test/load
 go 1.25.0
 
 require (
+	github.com/kcp-dev/logicalcluster/v3 v3.0.5
+	github.com/kcp-dev/sdk v0.31.0
```
xrstf, line 7: Should this have a `replace` directive to use the local SDK?

test/load/pkg/framework/setup.go
```diff
@@ -102,5 +106,32 @@ func loadKubeconfig(t *testing.T, kubeconfigPath string) *rest.Config {
 	restConfig, err := clientcmd.NewNonInteractiveClientConfig(*rawConfig, rawConfig.CurrentContext, nil, nil).ClientConfig()
 	require.NoError(t, err, "failed to create rest.Config from %s", kubeconfigPath)
 
+	// Strip any /clusters/<path> suffix from the Host URL. KCP kubeconfigs
```
xrstf, line 109: _kcp_

test/load/pkg/measurement/report.go
```diff
@@ -0,0 +1,130 @@
...
+import (
+	"fmt"
+	"io"
+	"sort"
+	"strings"
+	"text/tabwriter"
+	"time"
+)
+
+// Parameter is an ordered key-value pair used to describe test metadata
```
xrstf, line 28: A parameter itself is not ordered. 😁

test/load/pkg/tree/flat.go
```diff
@@ -0,0 +1,72 @@
...
+
+// flatTree is a workspace tree where all workspaces are direct children of the
+// root cluster. There is no nesting — every workspace lives at depth 1.
+// We mainly use this for comparison with nested trees.
+type flatTree struct {
+	Count int
+}
+
+// NewFlatTree creates a flat workspace tree with the given total count.
+func NewFlatTree(count int) *flatTree {
```
xrstf, line 36: Would it make sense to take a root cluster path as a parameter, so i could have more than one tree in a given kcp? So instead of having everthing hardcoded in `root`, it could be `root:loadtest1`.

xrstf: Doing that would allow us to combine trees. Why not have a flat tree where each node is a symmetricTree, for example? Or other techniques that re-use the two implementations you provide as building blocks.

test/load/go.mod
SimonTheLeg: done for sdk. For logicalcluster we don't have that in tree, right? It is still its own repository

test/load/pkg/tree/flat.go
SimonTheLeg: oh yes that is a very good point. I changed it

SimonTheLeg:
all comments have been addressed. PR is ready for re-review

xrstf (approved):
/approve

PR #4047, opened 2026-04-20
Improve delete of workspaces

### LINKED ISSUE #4012: bug: workspace object deletion seems to lack behind logicalcluster deletion by considerable margin
### Describe the bug

Especially on larger setups, you can run into situations where even 5-10 minutes after the logicalcluster is already gone, `kubectl get workspace` still reports the workspace as ready and good to go.

Here is an example. We can see in monitoring that all logicalclusters are gone for minutes at this point.

Image

however if we query for them using kubectl the workspace still shows up and shows up as ready

```sh

❯ k get ws
loadtest-999   organization            Ready   https://frontproxy.kcp-loadtests.kcp.shoot.canary.k8s-hana.ondemand.com:443/clusters/root:loadtest-999   35m
```

Workspace (or more precisely logicalcluster)-content cannot be accessed anymore

```
❯ k --server https://frontproxy.kcp-loadtests.kcp.shoot.canary.k8s-hana.ondemand.com:443/clusters/root:loadtest-999 get ns bla
Error from server (Forbidden): forbidden: User "" cannot get path "/clusters/root:loadtest-999/api/v1/namespaces/bla": workspace access not permitted
```

### Steps To Reproduce

1. Setup kcp with metrics
2. create a couple of thousand workspaces
3. delete all the workspaces
4. note how you can observe in the metrics that the logicalclusters are gone but k get workspaces still reports workspaces as ready

### Expected Behaviour

after a logical cluster is gone. The workspace object should disappear swiftly as well.

### Additional Context

_No response_

mjudeikis:
## Summary

Currently, workspace deletion is based on pooling. 
This will wire informer-based deletes and index for logicalClusters.

In addition, reduce pooling to 5 mins from the previous 10.

This should improve the speed at which workspaces are deleted when logicalcluster goes away.

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/4012

## Release Notes

```release-note
Improve workspace deletion after LogicalCluster is deleted
```

### DIFF
--- a/pkg/reconciler/tenancy/workspace/workspace_reconcile_phase.go
+++ b/pkg/reconciler/tenancy/workspace/workspace_reconcile_phase.go
@@ -119,7 +119,7 @@ func (r *phaseReconciler) reconcile(ctx context.Context, workspace *tenancyv1alp
 
 				if !conditions.IsTrue(workspace, tenancyv1alpha1.WorkspaceContentDeleted) {
 					after := time.Since(logicalCluster.CreationTimestamp.Time) / 5
-					if max := time.Minute * 10; after > max {
+					if max := time.Minute * 5; after > max {
 						after = max
 					}
 					cond := conditions.Get(logicalCluster, tenancyv1alpha1.WorkspaceContentDeleted)
--- a/pkg/reconciler/tenancy/workspace/workspace_reconcile_phase.go
+++ b/pkg/reconciler/tenancy/workspace/workspace_reconcile_phase.go
@@ -118,6 +118,14 @@ func (r *phaseReconciler) reconcile(ctx context.Context, workspace *tenancyv1alp
 				}
 
 				if !conditions.IsTrue(workspace, tenancyv1alpha1.WorkspaceContentDeleted) {
+					// Adaptive requeue: poll at 1/5 of the LogicalCluster's age, capped at 5 minutes.
+					// Young clusters get checked quickly (fast feedback when deletion finishes),
+					// older/stuck ones back off to reduce reconciler load.
+					// If the LogicalCluster never finishes deleting, the workspace stays in this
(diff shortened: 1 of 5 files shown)


pkg/reconciler/tenancy/workspace/workspace_reconcile_phase.go
```diff
@@ -119,7 +119,7 @@ func (r *phaseReconciler) reconcile(ctx context.Context, workspace *tenancyv1alp
 
 				if !conditions.IsTrue(workspace, tenancyv1alpha1.WorkspaceContentDeleted) {
 					after := time.Since(logicalCluster.CreationTimestamp.Time) / 5
-					if max := time.Minute * 10; after > max {
+					if max := time.Minute * 5; after > max {
```
ntnn, line 130: Why are we maxing out at five minutes? I feel like we should put in a comment why we chose this over the previous 10m.

ntnn: I'm just not a huge fan of undocumented magic numbers :D

ntnn (commented):
Otherwise looks good on a quick look.

/approve

pkg/reconciler/tenancy/workspace/workspace_reconcile_phase.go
mjudeikis: I don't know the original number, either. It's just some arbitrary time "for workspace delete" to succeed.
10 min per test was WAY too long.

TBH not sure what even todo here :D  Not enough data at scale to make a very informed decision.

mjudeikis: added explanation :)

PR #4048, opened 2026-04-20
Fix stale Scheduling metrics 

### LINKED ISSUE #4010: bug: metrics for fresh kcp show a logicalcluster stuck in scheduling
### Describe the bug

When looking at our kcp_logicalcluster metrics on a fresh kcp, it will show up one logicalcluster per shard as `Scheduling`. This is even hours after kcp has started and no workspace has been created.

Here is an example for a three sharded setup.

Image

Possibly we have a ghost logicalcluster somewhere which never gets ready or our metrics report a mistake.

### Steps To Reproduce

1. Setup kcp and setup prometheus to scrape the metrics endpoint
2. Scrape the metrics (either using the endpoint directly or using something like prometheus scraping). Here is an example for prometheus https://github.com/SimonTheLeg/kcp/blob/level-1-loadtest/test/load/setup/manifests/monitoring.yaml (will be merged to main soon, when I am back as well)

### Expected Behaviour

On an empty kcp that is fully started without any added workspaces by users, no logicalcluster should be in Scheduling

### Additional Context

_No response_

mjudeikis:
## Summary

Fix stale Scheduling metrics. 
metric some from system LC, which was bootstrapped via yaml and does not have the right statuses due to an unhandled condition. 

## What Type of PR Is This?
/kind bug
/kind chore

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/4010
## Release Notes

```release-note
NONE
```

### DIFF
--- a/staging/src/github.com/kcp-dev/cli/pkg/workspace/plugin/tree.go
+++ b/staging/src/github.com/kcp-dev/cli/pkg/workspace/plugin/tree.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"text/tabwriter"
 
 	tea "github.com/charmbracelet/bubbletea"
 	"github.com/spf13/cobra"
@@ -34,6 +35,7 @@ import (
 	"github.com/kcp-dev/logicalcluster/v3"
 	apisv1alpha1 "github.com/kcp-dev/sdk/apis/apis/v1alpha1"
 	apisv1alpha2 "github.com/kcp-dev/sdk/apis/apis/v1alpha2"
+	corev1alpha1 "github.com/kcp-dev/sdk/apis/core/v1alpha1"
 	tenancyv1alpha1 "github.com/kcp-dev/sdk/apis/tenancy/v1alpha1"
 	kcpclientset "github.com/kcp-dev/sdk/client/clientset/versioned/cluster"
 )
@@ -55,6 +57,7 @@ type TreeOptions struct {
 
 	Full        bool
 	Interactive bool
+	Wide        bool
 
 	kcpClusterClient kcpclientset.ClusterInterface
 }
@@ -71,6 +74,7 @@ func (o *TreeOptions) BindFlags(cmd *cobra.Command) {
 	o.Options.BindFlags(cmd)
 	cmd.Flags().BoolVarP(&o.Full, "full", "f", o.Full, "Show full workspace names")
 	cmd.Flags().BoolVarP(&o.Interactive, "interactive", "i", o.Interactive, "Interactive workspace tree browser")
+	cmd.Flags().BoolVar(&o.Wide, "wide", o.Wide, "Show workspace and logical cluster status for each node")
 }
 
 // Complete ensures all dynamically populated fields are initialized.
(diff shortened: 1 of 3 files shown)


staging/src/github.com/kcp-dev/cli/pkg/workspace/plugin/tree.go
```diff
@@ -71,6 +74,7 @@ func (o *TreeOptions) BindFlags(cmd *cobra.Command) {
 	o.Options.BindFlags(cmd)
 	cmd.Flags().BoolVarP(&o.Full, "full", "f", o.Full, "Show full workspace names")
 	cmd.Flags().BoolVarP(&o.Interactive, "interactive", "i", o.Interactive, "Interactive workspace tree browser")
+	cmd.Flags().BoolVar(&o.Wide, "wide", o.Wide, "Show workspace and logical cluster status for each node")
```
mjudeikis, line 77: This add additional mode to include statuses in `tree` mode

staging/src/github.com/kcp-dev/cli/pkg/workspace/plugin/tree.go
```diff
@@ -222,18 +251,9 @@ func (o *TreeOptions) populateInteractiveNodeBubble(ctx context.Context, node *t
...
 
-	shouldLoadChildren := node.expanded
-	if !shouldLoadChildren {
-		if workspace == currentWorkspace {
-			shouldLoadChildren = true
-		} else if currentWorkspace.HasPrefix(workspace) {
-			shouldLoadChildren = true
-		}
-	}
+	shouldLoadChildren := node.expanded || currentWorkspace.HasPrefix(workspace)
```
mjudeikis, line 254: Interactive mode is broken. Fixing it too

hack/run-with-prometheus.sh
```diff
@@ -14,13 +14,54 @@
...
 set -o errexit
 
 cd "$(dirname $0)/.."
 
-PROMETHEUS="$(UGET_PRINT_PATH=relative make --no-print-directory prometheus)"
+# Resolve the prometheus binary path from the Makefile to avoid relying on
+# `make prometheus` stdout, which can be polluted by sub-commands (downloads,
+# recipe tracing) in some environments.
+PROMETHEUS_VER="$(awk -F':=' '/^PROMETHEUS_VER/ {gsub(/ /,"",$2); print $2}' Makefile)"
+PROMETHEUS="$(pwd)/hack/tools/prometheus-${PROMETHEUS_VER}"
```
xrstf: `PROMETHEUS="$(UGET_PRINT_PATH=absolute make --no-print-directory prometheus)"`

is the pattern we use in other projects and what uget was made for. :-) No need for this hackery.

hack/run-with-prometheus.sh
```diff
@@ -36,6 +77,42 @@ while ! curl -s http://localhost:9090/-/ready; do
 done
 echo 'Prometheus is ready!'
 
+# Dev-only: when running a local `kcp start` (no ARTIFACT_DIR set, so not CI/Prow),
+# watch for kcp's cert+kubeconfig to appear and auto-populate a scrape config.
+# In CI, ARTIFACT_DIR is set and e2e tests populate .prometheus-config.yaml via
+# the Go testing framework (staging/.../testing/server/metrics.go) — this watcher
+# would race and clobber those per-server entries, so it stays off.
```
xrstf: Couldn't we simply set ARTIFACT_DIR? When I run tests for kcp locally, I set it anyway because the deeply nested paths in /tmp make it impossible to find anything :D We could set it to `.artifacts` and have `make clean` wipe it, too?

mjudeikis: The problem is that most other /cmd test endpoints are producing their own Prometheus configs. 
So need a better way to distinguish when we produce one and when we generate on in the script./

Will split this bash nonsense to separate pr as I see we're going to be iterating on this for a while :D

ntnn (approved):
/lgtm
/approve

PR #4052, opened 2026-04-21
Update run-with-prometheus to support local dev

mjudeikis:
## Summary

So one can easily run prometheus when developing metrics or anything metrics related

## What Type of PR Is This?
/kind cleanup

## Related Issue(s)

Fixes #

## Release Notes

```release-note
make run-with-prometheus work for local dev
```

### DIFF
--- a/hack/run-with-prometheus.sh
+++ b/hack/run-with-prometheus.sh
@@ -14,13 +14,42 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Usage: ./hack/run-with-prometheus.sh <command> [args...]
+#
+# Starts a local Prometheus (http://localhost:9090) using .prometheus-config.yaml,
+# runs the given command, then tears Prometheus down and exits with the command's
+# exit code. Without arguments the script runs nothing, exits 0, and immediately
+# kills Prometheus — you must pass a command.
+#
+# Examples:
+#   # Single-node kcp (production binary). The dev watcher below auto-populates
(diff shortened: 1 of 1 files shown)


hack/run-with-prometheus.sh
```diff
@@ -14,13 +14,47 @@
...
 
 cd "$(dirname $0)/.."
 
-PROMETHEUS="$(UGET_PRINT_PATH=relative make --no-print-directory prometheus)"
+PROMETHEUS="$(UGET_PRINT_PATH=absolute make --no-print-directory prometheus | tail -n 1)"
+
+if [ ! -x "${PROMETHEUS}" ]; then
+  echo "prometheus binary not found at ${PROMETHEUS}" >&2
+  exit 1
+fi
```
xrstf: Why is this necessary? What didn't work about old approach? `make prometheus` should never succeed if it fails to download Prometheus.

Makefile
```diff
@@ -144,6 +144,7 @@ golangci-lint-version:
 
 .PHONY: prometheus
 prometheus: $(PROMETHEUS)
+	@echo $(PROMETHEUS)
```
xrstf: uget already prints the binary and only the binary path if UGET_PRINT_PATH is set. There's no need to echo something here. In fact, this breaks uget and forces you to have the `tail -n1` in the prometheus script.

mjudeikis:
@xrstf updated

xrstf (approved):
Thanks!

/approve

PR #4074, opened 2026-04-27
Fix mounts panic

mjudeikis:
## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Fix panic for mounts when GVR is not yet synced
```

### DIFF
--- a/pkg/reconciler/tenancy/workspacemounts/workspace_indexes.go
+++ b/pkg/reconciler/tenancy/workspacemounts/workspace_indexes.go
@@ -66,6 +66,8 @@ func newIndexWorkspaceByMountObject(dynamicRESTMapper *dynamicrestmapper.Dynamic
 
 		gv, err := schema.ParseGroupVersion(ws.Spec.Mount.Reference.APIVersion)
 		if err != nil {
+			// Malformed user input — safe to surface as an error; the value
+			// will never become valid by re-indexing the same object.
 			return nil, fmt.Errorf("unable to parse APIVersion of mount reference: %w", err)
 		}
 		gvk := schema.GroupVersionKind{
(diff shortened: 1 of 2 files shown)


pkg/reconciler/tenancy/workspacemounts/workspace_indexes.go
```diff
@@ -66,6 +66,8 @@ func newIndexWorkspaceByMountObject(dynamicRESTMapper *dynamicrestmapper.Dynamic
 
 		gv, err := schema.ParseGroupVersion(ws.Spec.Mount.Reference.APIVersion)
 		if err != nil {
+			// Malformed user input — safe to surface as an error; the value
+			// will never become valid by re-indexing the same object.
```
ntnn:
```suggestion
```

I don't think this comment makes too much sense?

pkg/reconciler/tenancy/workspacemounts/workspace_indexes_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+// TestIndexWorkspaceByMountObject exercises the mount-reference indexer.
+//
+// The unknown-kind case matters because client-go's storeIndex.updateSingleIndex
+// panics on any IndexFunc error (see k8s.io/client-go/tools/cache/thread_safe_store.go),
+// which would tear down the apiserver. A Workspace pointing at a kind whose
+// APIBinding/CRD hasn't reconciled yet is a normal, recoverable state — not a
+// reason to crash the process. A malformed APIVersion, by contrast, is
+// user-input that will never become valid by re-indexing the same object, so
+// reporting it is preferable to silently dropping it.
```
ntnn: Can we move this comment to the specific test case?

pkg/reconciler/tenancy/workspacemounts/workspace_indexes.go
mjudeikis: I think it does. I ran into this issue and was "reasoning to myself" about why one error is ok to return, and the other is not. but will remove :D

pkg/reconciler/tenancy/workspacemounts/workspace_indexes_test.go
mjudeikis: Please no :) Took me a while to get to the bottom of this and this is kinda critical path in mounts. And when reading code, you need to understand why somebody did things how they did, instead of going and hunting for info in the tests :)

ntnn: Yeah but I'd rather have the docs for a specific test right next to it to understand the intricacies of that specific test^^
Also makes it easier to reference the settings in that test^^

For the test body when it looks odd it's usually because a few cases need very specific handling :D

ntnn: That being said - looking at the test body; I don't think these three tests need to be table-driven. 

There's two tests with wantErr==false and wantEmpty==true and one with wantErr=true; the test body is more routing than testing.
We don't have to change that because it's already quite often the case in kcp's tests, just an observation :D

mjudeikis: ah! this comment was ON test :D not on the main file :D right, comment leggit.

ntnn (commented):
/lgtm
/approve

PR #4075, opened 2026-04-27
restored api-conversion implementation

### LINKED ISSUE #3427: feature: re-implement `APIConversions`
### Feature Description

kcp has an API resource called [`APIConversion`](https://docs.kcp.io/kcp/main/reference/crd/apis.kcp.io/apiconversions/). This used to be working as a way for API providers to define API version conversion rules (something that is usually done via webhooks in upstream Kubernetes). The original PR implementing this was #2105, but its functionality got removed with #3017 as it caused issues with the Kubernetes rebase.

We should bring the feature back, as it is critical to API providers that want to evolve their API to another version.

### Proposed Solution

We should re-implement `APIConversions` (it would make a lot of sense to look at the previous implementation). There were a couple of potential concerns raised with the original implementation (also see Slack thread linked below):

> it made rebase harder at the time.
 
>  I think converting lists was hard or impossible.

An implementation should include documentation for this feature and e2e testing. Also look for older issues like #2696 to see potential improvements.

### Alternative Solutions

_No response_

### Want to contribute?

### Additional Context

See [this Slack thread](https://kubernetes.slack.com/archives/C021U8WSAFK/p1747769957631829) for some recent discussion.

olamilekan000:
## Summary
```
changes restores the  APIConversion  implementation
```
---
Screenshot 2026-04-27 at 19 44 51

---
Screenshot 2026-04-27 at 19 45 22

---
Screenshot 2026-04-27 at 17 53 16

---

Screenshot 2026-04-27 at 19 17 27

---

Screenshot 2026-04-27 at 19 18 10

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes [3427](https://github.com/kcp-dev/kcp/issues/3427)

## Release Notes

```release-note
added APIConversion implementation
```


olamilekan000:
While testing the API conversions, the conversion from one api to the other was failing for standard local CRDs because the admission plugin strictly required an `APIResourceSchema` to exist. Since pure local CRDs don't generate those schemas (it has to be generated through apibinding and apiexport), the apiconversion webhook was rejecting them with a NotFound error. To fix this, I added a provision in the admission validator using the dynamic client to check if a matching `CustomResourceDefinition` exists locally instead. If it does, we gracefully bypass the strict schema mapping and let the CEL interpreter handle the validation at runtime.

pkg/conversion/deferred_converter.go
```diff
@@ -0,0 +1,115 @@
...
+
+	return converter.Convert(in, targetGV)
+}
+
+func (f *deferredConverter) getConverter() (conversion.CRConverter, error) {
+	f.lock.Lock()
+	defer f.lock.Unlock()
+
+	if f.delegate != nil {
+		return f.delegate, nil
```
mjudeikis: If we update APIResources post create (but dont update CRD) - Think this will return stale converter?

The problem in plain terms: once the first conversion happens, `f.delegate` is populated with a `*Converter` whose CEL programs were compiled against whatever `APIConversion` existed at that moment. There's no read of the lister on subsequent calls, no resourceVersion check, no event handler — so a kubectl edit apiconversion ... does nothing for already-running converters in this process.

mjudeikis: We need to track APIConversion RV to see if converted is still valid.

pkg/admission/apiconversion/apiconversion_admission.go
```diff
@@ -0,0 +1,173 @@
...
+
+		gvr := schema.GroupVersionResource{Group: "apiextensions.k8s.io", Version: "v1", Resource: "customresourcedefinitions"}
+		_, crdErr := o.dynamicClient.Cluster(cluster.Name.Path()).Resource(gvr).Get(ctx, u.GetName(), metav1.GetOptions{})
+
+		// In kcp, local CustomResourceDefinitions do not generate APIResourceSchemas.
+		// Therefore, restricting this validation entirely to the APIResourceSchema lister breaks the ability
+		// to use pure local CRDs with the CEL engine. This fallback natively inspects the local Virtual Workspace
+		// for a raw CustomResourceDefinition matching the conversion rule.
+		if crdErr == nil {
+			klog.FromContext(ctx).
```
mjudeikis: Maybe this would allow not to skip?

```
if crdErr == nil {
    crdU, _ := o.dynamicClient.Cluster(...).Resource(gvr).Get(...)   // already retrieved above so we have the object already
    crd := &apiextensionsv1.CustomResourceDefinition{}
    if err := runtime.DefaultUnstructuredConverter.FromUnstructured(crdU.Object, crd); err != nil {
        return admission.NewForbidden(a, fmt.Errorf("decoding CRD: %w", err))
    }
    structuralSchemas, err := buildStructuralSchemasFromCRD(crd.Spec.Versions)
    if err != nil {
        return admission.NewForbidden(a, err)
    }
    apiConversion := &apisv1alpha1.APIConversion{}
    if err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, apiConversion); err != nil {
        return fmt.Errorf("failed to convert unstructured to APIConversion: %w", err)
    }
    if _, err := conversion.Compile(apiConversion, structuralSchemas); err != nil {
        return fmt.Errorf("error compiling conversion rules: %w", err)
    }
    return nil
}
```

pkg/server/config.go
```diff
@@ -677,10 +678,12 @@ func NewConfig(ctx context.Context, opts kcpserveroptions.CompletedOptions) (*Co
 	admissionPluginInitializers = append(admissionPluginInitializers, kubePluginInitializer...)
 
 	authInfoResolver := webhook.NewDefaultAuthenticationInfoResolverWrapper(kubeControlPlane.ProxyTransport, kubeControlPlane.Generic.EgressSelector, kubeControlPlane.Generic.LoopbackClientConfig, kubeControlPlane.Generic.TracerProvider)
-	conversionFactory, err := NewCRConverterFactory(serviceResolver, authInfoResolver)
-	if err != nil {
-		return nil, err
-	}
+	_ = authInfoResolver
```
mjudeikis: Why do we do this? It's not used?

pkg/cache/server/config.go
```diff
@@ -248,19 +251,12 @@ func NewConfig(opts *cacheserveroptions.CompletedOptions, optionalLocalShardRest
 				c.ApiExtensionsSharedInformerFactory,
 				c.KcpSharedInformerFactory,
 			),
-			ConversionFactory: &nopCRConversionFactory{},
+			ConversionFactory: kcpconversion.NewCRConverterFactory(
+				c.KcpSharedInformerFactory.Apis().V1alpha1().APIConversions(),
+				5*time.Second,
```
mjudeikis: opts.Extra.ConversionCELTransformationTimeout ?

pkg/conversion/conversion_rules.go
```diff
@@ -0,0 +1,208 @@
...
+
+	return rulesForVersion, kerrors.NewAggregate(errs)
+}
+
+// compileRule takes a single APIConversionRule, compiles the CEL transformation (if any), and converts the rule to a
+// CompiledRule that the converter uses to perform conversions.
+func compileRule(rulePath *field.Path, rule apisv1alpha1.APIConversionRule, schema *structuralschema.Structural) (*CompiledRule, error) {
+	// if rule.Field starts with a ".", such as .spec.fieldName, strip the leading "."
+	fromPath := rule.Field
+	if fromPath[0] == '.' {
```
mjudeikis: This function is called from Compile, and from NewConverter, hence toPath and fromPath never checked for `len`

pkg/conversion/deferred_converter.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	apiConversion, err := f.getAPIConversion(clusterName, conversionName)
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, fmt.Errorf("error checking for APIConversion for CRD %s|%s: %w", clusterNameAnnotation, f.crd.Name, err)
+	}
+	if errors.IsNotFound(err) {
+		switch f.crd.Spec.Conversion.Strategy {
+		case apiextensionsv1.NoneConverter:
+			return conversion.NewNOPConverter(), nil
```
mjudeikis: We should be setting `f.delegate = conversion.NewNOPConverter()` else this code path will be hot on every request

pkg/conversion/widgets-crd.yaml
```diff
@@ -0,0 +1,83 @@
+metadata:
```
mjudeikis, line 1: Is this even used?

mjudeikis (commented):
How much of this is original implementation and how much is your change? I feel like original implementation had few corners cut :/ Tried to leave few comments but I suspect there will be more

olamilekan000:
> How much of this is original implementation and how much is your change? I feel like original implementation had few corners cut :/ Tried to leave few comments but I suspect there will be more

Almost all is from the original PR. I only refactored the CEL flow to use  EnvSet and StoredExpressionsEnv, and also updated the admission plugin with that dynamic client fallback so it also handles local CRDs instead of just looking for APIResourceSchemas.

pkg/server/config.go
ntnn: Shouldn't this also be caught by the linter?

pkg/conversion/widgets-crd.yaml
ntnn: It's used in the unit tests.

pkg/conversion/deferred_converter.go
```diff
@@ -0,0 +1,115 @@
...
+		)
+		return nil, fmt.Errorf("unable to get converter")
+	}
+
+	apiConversion, err := f.getAPIConversion(clusterName, conversionName)
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, fmt.Errorf("error checking for APIConversion for CRD %s|%s: %w", clusterNameAnnotation, f.crd.Name, err)
+	}
+	if errors.IsNotFound(err) {
+		switch f.crd.Spec.Conversion.Strategy {
```
ntnn: .Conversion can be nil, that should be checked here.

ntnn (commented):
I think this skips the `.ConvertToVersion` that is implemented for kcp's native types, or?
The old conversion webhook in pkg/server (that did this conversion) should be refactored into the new conversion factory.

pkg/server/config.go
olamilekan000: I was trying to retain some of the changes from the old PR, hence why I left it that way. I think it can be removed

pkg/conversion/widgets-crd.yaml
olamilekan000: yes

pkg/admission/apiconversion/apiconversion_admission.go
olamilekan000: so we don't need for the validation?

pkg/cache/server/config.go
ntnn: I'd prefer this be a constant with a comment explaining why 5s.

pkg/conversion/conversion_rules.go
```diff
@@ -0,0 +1,208 @@
...
+		return cr, nil
+	}
+
+	env, err := createCELEnv(schemaForField)
+	if err != nil {
+		return nil, field.InternalError(rulePath.Child("field"), fmt.Errorf("error creating CEL environment: %w", err))
+	}
+
+	ast, issues := env.Compile(rule.Transformation)
+	if issues != nil {
```
ntnn, line 135:
```suggestion
	if issues.Err() != nil {
```

issues is never nil iirc

pkg/features/kcp_features.go
```diff
@@ -142,6 +147,9 @@ var defaultVersionedGenericControlPlaneFeatureGates = map[featuregate.Feature]fe
 	WorkspaceAuthentication: {
 		{Version: version.MustParse("1.32"), Default: false, PreRelease: featuregate.Alpha},
 	},
+	APIConversion: {
+		{Version: version.MustParse("1.32"), Default: false, PreRelease: featuregate.Alpha},
```
ntnn, line 176:
```suggestion
		{Version: version.MustParse("1.33"), Default: false, PreRelease: featuregate.Alpha},
```

Since the rebase landed inbetween

pkg/cache/server/config.go
```diff
@@ -235,6 +236,9 @@ func NewConfig(opts *cacheserveroptions.CompletedOptions, optionalLocalShardRest
 		return nil, err
 	}
 
+	// Ensure the cache server natively listens for APIConversions
+	_ = c.KcpSharedInformerFactory.Apis().V1alpha1().APIConversions().Informer()
```
ntnn:
```suggestion
```

The same is a few lines below

pkg/cache/server/options/options.go
```diff
@@ -73,14 +75,15 @@ func (o *CompletedOptions) Validate() []error {
...
-		Logs:             logs.NewOptions(),
+		ServerRunOptions:                   genericoptions.NewServerRunOptions(),
+		Etcd:                               genericoptions.NewEtcdOptions(storagebackend.NewDefaultConfig(kubeoptions.DefaultEtcdPathPrefix, nil)),
+		SecureServing:                      genericoptions.NewSecureServingOptions().WithLoopback(),
+		Authentication:                     NewAuthentication(),
+		Authorization:                      NewAuthorization(),
+		APIEnablement:                      genericoptions.NewAPIEnablementOptions(),
+		EmbeddedEtcd:                       *etcdoptions.NewOptions(rootDir),
+		Logs:                               logs.NewOptions(),
+		ConversionCELTransformationTimeout: time.Second,
```
ntnn: Is the 1s timeout arbitrary or do we have an educated guess on this default? I think a comment explaining why that is the default would be good.

pkg/admission/apiconversion/apiconversion_admission.go
```diff
@@ -0,0 +1,229 @@
...
+
+		gvr := schema.GroupVersionResource{Group: "apiextensions.k8s.io", Version: "v1", Resource: "customresourcedefinitions"}
+		_, crdErr := o.dynamicClient.Cluster(cluster.Name.Path()).Resource(gvr).Get(ctx, u.GetName(), metav1.GetOptions{})
+
+		// In kcp, local CustomResourceDefinitions do not generate APIResourceSchemas.
+		// Therefore, restricting this validation entirely to the APIResourceSchema lister breaks the ability
+		// to use pure local CRDs with the CEL engine. This fallback natively inspects the local Virtual Workspace
+		// for a raw CustomResourceDefinition matching the conversion rule.
+		if crdErr == nil {
+			crdU, err := o.dynamicClient.Cluster(cluster.Name.Path()).Resource(gvr).Get(ctx, u.GetName(), metav1.GetOptions{})
```
ntnn: These two GETs are identical. Should the second one use a different client?

pkg/admission/apiconversion/apiconversion_admission.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// ValidateInitialization ensures the required injected fields are set.
+func (o *apiConversionAdmission) ValidateInitialization() error {
+	if o.getAPIResourceSchema == nil {
+		return fmt.Errorf("getAPIResourceSchema is unset")
+	}
+
+	return nil
+}
```
ntnn, line 169: Missing check the dynamicClient is not nil

pkg/conversion/conversion_factory.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// NewConverter returns the appropriate conversion.Converter based on the CRD. If the CRD identifies as for a "wildcard
+// partial metadata request", the nop converter is used. Otherwise, it returns a CEL-based converter if there is an
+// associated APIConversion, a nop converter if the strategy is "none", or an error otherwise.
+func (f *CRConverterFactory) NewConverter(crd *apiextensionsv1.CustomResourceDefinition) (conversion.CRConverter, error) {
+	if !f.enabled {
+		if crd.Spec.Conversion != nil && crd.Spec.Conversion.Strategy != apiextensionsv1.NoneConverter {
+			return nil, fmt.Errorf("conversion strategy %q is not supported for CRD %s", crd.Spec.Conversion.Strategy, crd.Name)
+		}
```
ntnn, line 88: This errors for webhook converters, or? Wouldn't that also prevent e.g. the webhook converter on CRDs that have it configured?

olamilekan000:
@ntnn I haven't been chanced to properly look at your feedback, and I do not want to respond hastily to it. Once im a bit free, I'll respond do not also hesitate to let me know if it's blocking anything. Thanks

pkg/conversion/conversion_factory.go
olamilekan000: Yes it will prevent them. so, if strategy is anything aside "none" and conversion is disabled, we throw an error.  Or to put it another way, if strategy is Webhook and the gate is off, we throw an error because kcp doesn’t support conversion webhooks

pkg/server/config.go
```diff
@@ -715,11 +716,14 @@ func NewConfig(ctx context.Context, opts kcpserveroptions.CompletedOptions) (*Co
 	c.Apis = kubeControlPlane
 	admissionPluginInitializers = append(admissionPluginInitializers, kubePluginInitializer...)
 
-	authInfoResolver := webhook.NewDefaultAuthenticationInfoResolverWrapper(kubeControlPlane.ProxyTransport, kubeControlPlane.Generic.EgressSelector, kubeControlPlane.Generic.LoopbackClientConfig, kubeControlPlane.Generic.TracerProvider)
-	conversionFactory, err := NewCRConverterFactory(serviceResolver, authInfoResolver)
-	if err != nil {
-		return nil, err
-	}
+	conversionFactory := kcpconversion.NewCRConverterFactory(
```
mjudeikis: This breaks existing our conversion machinery. We need e2e test to close the gap, and we need better handling for this. 

my ai agent reivew bellow:

```
Both [pkg/server/config.go:719](vscode-webview://0pikil8f7g1qhrf2uek89gkocorlns8ks47u870endnhaeslejc6/pkg/server/config.go#L719) and [pkg/cache/server/config.go:262](vscode-webview://0pikil8f7g1qhrf2uek89gkocorlns8ks47u870endnhaeslejc6/pkg/cache/server/config.go#L262) replace the conversion factory with kcpconversion.NewCRConverterFactory. The old factory ([pkg/server/conversion_webhook.go:60-69](vscode-webview://0pikil8f7g1qhrf2uek89gkocorlns8ks47u870endnhaeslejc6/pkg/server/conversion_webhook.go#L60)) special-cased the apis.kcp.io group and ran native scheme-based conversion for it:


if crd.Spec.Group == apis.GroupName {
    return &schemaBasedConverter{...}, nil   // scheme.ConvertToVersion
}
return f.delegate.NewConverter(crd)
That handler was added deliberately (commit dad2aea9e "add conversion handler for core APIs") to convert APIBinding/APIExport between v1alpha1 and v1alpha2 — a non-trivial hand-written conversion (permission-claim reshaping + JSON round-trip annotations, see staging/src/github.com/kcp-dev/sdk/apis/apis/v1alpha2/types_apibinding_conversion.go). Storage version is v1alpha2; both versions are served.

The new factory has no group special-casing. For an apibindings.apis.kcp.io CRD (whose spec.conversion.strategy: None), the deferred converter resolves to NewNOPConverter():

getAPIConversion(...) → NotFound (no user APIConversion for core types)
crd.Spec.Conversion.Strategy == None → [deferred_converter.go:112](vscode-webview://0pikil8f7g1qhrf2uek89gkocorlns8ks47u870endnhaeslejc6/pkg/conversion/deferred_converter.go#L112) returns NOP.
And crucially this happens even with the feature gate off — [conversion_factory.go:62-67](vscode-webview://0pikil8f7g1qhrf2uek89gkocorlns8ks47u870endnhaeslejc6/pkg/conversion/conversion_factory.go#L62) also returns NOP for None-strategy CRDs. So a client requesting an APIBinding/APIExport in a non-storage version gets the object back with only apiVersion relabeled and fields not converted → corrupt/lossy response and a real v1alpha1 API-compatibility break.

This likely escapes existing e2e because tests use the storage version (v1alpha2) and rarely request v1alpha1, so conversion never fires. Recommend adding an e2e that GETs a stored v1alpha2 APIBinding as v1alpha1 (and vice-versa) and asserts permission-claim fields round-trip.

Fix: the new CRConverterFactory.NewConverter must keep the apis.GroupName → scheme-based-converter branch (delegate to the existing schemaBasedConverter) before falling through to the deferred CEL/NOP path.
```

pkg/conversion/deferred_converter.go
```diff
@@ -0,0 +1,130 @@
...
+	if errors.IsNotFound(err) {
+		if f.delegate != nil && f.apiConversionResourceVersion == "" {
+			return f.delegate, nil
+		}
+
+		if f.crd.Spec.Conversion != nil {
+			switch f.crd.Spec.Conversion.Strategy {
+			case apiextensionsv1.NoneConverter:
+				// handled below
+			case apiextensionsv1.WebhookConverter:
```
mjudeikis, line 105: We are dropping this. But do we support this?

pkg/conversion/converter.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error creating CEL interpreter activation: %w", err)
+	}
+
+	evalResult, _, err := rule.Program.ContextEval(ctx, activation)
+	if err != nil {
+		return nil, fmt.Errorf("error executing transformation: %w", err)
+	}
+
+	return evalResult.Value(), nil
```
mjudeikis: This will panic for some reason; this is not JSON payload. (Internally, it uses `runtime.DeepCopyJSONValue `)

mjudeikis (commented):
We are regressing conversion api.

To confirm this we need to add e2e, because tests use the storage version (v1alpha2) and rarely request v1alpha1, so conversion never fires. We need to  add an e2e that GETs a stored v1alpha2 APIBinding as v1alpha1 (and vice-versa) and asserts permission-claim fields round-trip.

pkg/conversion/deferred_converter.go
olamilekan000: No we don't at the momenr

olamilekan000:
@mjudeikis @ntnn can you kindly help review. Thanks

Issue #4016, opened 2026-04-13
Workspace Lifecycle Content Access Enhancement


mjudeikis:
## Summary

Replace annotation-based workspace owner identity with a structured API field, introduce fine-grained RBAC-based workspace content access for initializing and terminating virtual workspace content proxies, add a `Terminating` phase, and add workspace content access to the terminating virtual workspace.

Tracking KEP: https://github.com/kcp-dev/enhancements/blob/main/keps/core/0003-workspace-lifecycle-rbac.md 
https://github.com/kcp-dev/enhancements/pull/11 

## Phases

- [ ] **Phase 1**: Structured owner identity on LogicalCluster — #4017
- [ ] **Phase 2**: Terminating phase and content proxy — #4018
- [ ] **Phase 3**: Declarative RBAC on WorkspaceType — #4019

## Phase Dependencies

```
Phase 1 → Phase 2 → Phase 3
```

Phase 2 depends on Phase 1 (needs `spec.ownerUser` for impersonation). Phase 3 depends on both Phase 1 and Phase 2.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

mjudeikis:
This would fix this https://github.com/kcp-dev/kcp/issues/4038 too

PR #4082, opened 2026-04-28
Add aditional validation for WST

mjudeikis:
## Summary

Enhance WorkspaceType permissions admission checking with more in-depth defence.
This could be done in the controller itself, BUT as we are in the middle of refactoring https://github.com/kcp-dev/kcp/issues/4016 this would make it harder to backport, and its moving target. 

Once this is fixed, I will create a follow-up to do a controller-level check once this one is merged, backported and released. 

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Add more in-depth defence on WorkspaceType default APIBindings
```

### DIFF
--- a/pkg/admission/workspacetype/admission.go
+++ b/pkg/admission/workspacetype/admission.go
@@ -18,6 +18,7 @@ package workspacetype
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 
@@ -26,13 +27,32 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apiserver/pkg/admission"
 	genericapirequest "k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
 
+	kcpkubernetesclientset "github.com/kcp-dev/client-go/kubernetes"
+	"github.com/kcp-dev/logicalcluster/v3"
+	apisv1alpha2 "github.com/kcp-dev/sdk/apis/apis/v1alpha2"
 	"github.com/kcp-dev/sdk/apis/core"
 	tenancyv1alpha1 "github.com/kcp-dev/sdk/apis/tenancy/v1alpha1"
+	kcpinformers "github.com/kcp-dev/sdk/client/informers/externalversions"
+
+	apibindingadmission "github.com/kcp-dev/kcp/pkg/admission/apibinding"
+	kcpinitializers "github.com/kcp-dev/kcp/pkg/admission/initializers"
+	"github.com/kcp-dev/kcp/pkg/authorization/delegated"
+	"github.com/kcp-dev/kcp/pkg/indexers"
 )
 
 // Validate WorkspaceTypes creation and updates for
-//  - "organization" type is only created in root workspace.
+//   - the "root" WorkspaceType can only be created in the root cluster.
+//   - .spec.defaultChildWorkspaceType.path, .spec.limitAllowedChildren.types[*].path,
+//     and .spec.limitAllowedParents.types[*].path must be set when their parent
+//     fields are present.
+//   - the user has the "bind" verb on every APIExport listed in
+//     spec.defaultAPIBindings (newly added entries on update). This prevents
+//     a privilege escalation where unprivileged users would otherwise cause
+//     the default-apibinding-controller to create APIBindings on their behalf
+//     to APIExports they cannot bind directly.
 
 const (
 	PluginName = "tenancy.kcp.io/WorkspaceType"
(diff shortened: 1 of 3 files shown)


test/e2e/apibinding/default_apibinding_admission_test.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+
+		// Admission relies on informers, which can lag — retry until we observe
+		// the deterministic forbidden response.
+		var lastErr error
+		kcptestinghelpers.Eventually(t, func() (bool, string) {
+			_, lastErr = user1KcpClient.Cluster(tenantPath).TenancyV1alpha1().WorkspaceTypes().Create(t.Context(), wt, metav1.CreateOptions{})
+			if lastErr == nil {
+				return false, "expected forbidden error, got nil"
+			}
```
gman0: That doesn't sound correct. Does this really default to `err == nil`?

pkg/admission/workspacetype/admission.go
```diff
@@ -97,5 +132,121 @@ func (o *workspacetype) Validate(ctx context.Context, a admission.Attributes, _
...
+	if len(wt.Spec.DefaultAPIBindings) == 0 {
+		return nil
+	}
+
+	// On Update, only check entries that are newly added compared to the old
+	// object — pre-existing entries were already validated when the WorkspaceType
+	// was created (or, for objects predating this check, are grandfathered in to
+	// avoid breaking legitimate updates by users who never had bind permission).
+	var existing map[tenancyv1alpha1.APIExportReference]struct{}
+	if a.GetOperation() == admission.Update {
```
ntnn: This could be written better. `existing` is always nil unless the operation is update. Only when `existing` has content the default API bindings are checked in the for below.

It would make more sense and be more readable to return early when the operation is not update?

Or is this intended to extend later? Though I was a bit confused reading through it the first time.

test/e2e/apibinding/default_apibinding_admission_test.go
mjudeikis: It's fine, means we got no error, so informers are not synced - we test for error.

pkg/admission/workspacetype/admission.go
mjudeikis: No, I think it's right. There was explicit behaviour I was trying to fix. Basically, avoid breaking existing WSTs. And for this we need this convoluted check. I added better comment.

ntnn (commented):
/lgtm
/approve

pkg/admission/workspacetype/admission.go
```diff
@@ -26,13 +27,29 @@ import (
...
+
+	apibindingadmission "github.com/kcp-dev/kcp/pkg/admission/apibinding"
+	kcpinitializers "github.com/kcp-dev/kcp/pkg/admission/initializers"
+	"github.com/kcp-dev/kcp/pkg/authorization/delegated"
+	"github.com/kcp-dev/kcp/pkg/indexers"
 )
 
 // Validate WorkspaceTypes creation and updates for
-//  - "organization" type is only created in root workspace.
+//   - "organization" type is only created in root workspace.
```
sttts: what does this mean? I know it is preexisting.

test/e2e/apibinding/default_apibinding_admission_test.go
gman0: But how does a not-yet-synced informer cause Create to succeed? If a key is missing in cache, a NotFound error would be returned, and the admission should return some sort of final error, no? If it can succeed, doesn't it defeat the purpose of this APIExport `bind` verb check?

pkg/admission/workspacetype/admission.go
mjudeikis: stale comment. Needs updating

mjudeikis: Updated

test/e2e/apibinding/default_apibinding_admission_test.go
mjudeikis: make sense, updated

ntnn (commented):
/lgtm
/approve

ntnn (commented):

The test failure is unrelated to the changes in this PR

pkg/admission/workspacetype/admission.go
```diff
@@ -97,5 +135,132 @@ func (o *workspacetype) Validate(ctx context.Context, a admission.Attributes, _
...
+			exportClusterName = clusterName
+		case exportPath == core.RootCluster.String():
+			exportClusterName = core.RootCluster
+		default:
+			path := logicalcluster.NewPath(exportPath)
+			export, err := o.getAPIExport(path, exportName)
+			if err != nil {
+				return forbidden
+			}
+			exportClusterName = logicalcluster.From(export)
```
SimonTheLeg, line 213: I think this is not covered by the testing. We have a test which returns early on an error (l 211), but this default success case we don't seem to have unit-tested.

SimonTheLeg (changes requested):

pkg/admission/workspacetype/admission.go
SimonTheLeg: (github line numbers are messing up again 🙄 . What I mean is: only line 213 is untested)

mjudeikis: added

ntnn (approved):
/lgtm
/approve

PR #4085, opened 2026-04-28
Recover workspaces from Unavailable

ntnn:
## Summary

In high-pressure environments a workspace may be marked unavailable because the 10s timeout is not enough for a logical cluster to be propagated and be known to the shard.

When this happens the workspace never recovers as there is nothing updating the status.

To prevent this:

1. The timeout is increased to one minute via a documented const (so ~60 retries instead of 10)
2. Added reconcile that allows the Workspace to recover from the logical cluster not being visible during initialization

## What Type of PR Is This?

/kind bug
/kind flake

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/tenancy/workspace/workspace_reconcile_phase.go
+++ b/pkg/reconciler/tenancy/workspace/workspace_reconcile_phase.go
@@ -39,6 +39,11 @@ type phaseReconciler struct {
 	requeueAfter func(workspace *tenancyv1alpha1.Workspace, after time.Duration)
 }
 
+// logicalClusterTimeout is the time limit for a logical cluster to
+// appear for the shard before marking its corresponding Workspace as
+// Unavailable.
+const logicalClusterTimeout = time.Second
+
 func (r *phaseReconciler) reconcile(ctx context.Context, workspace *tenancyv1alpha1.Workspace) (reconcileStatus, error) {
 	logger := klog.FromContext(ctx).WithValues("reconciler", "phase")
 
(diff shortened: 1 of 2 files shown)


pkg/reconciler/tenancy/workspace/workspace_reconcile_phase.go
```diff
@@ -97,6 +102,27 @@ func (r *phaseReconciler) reconcile(ctx context.Context, workspace *tenancyv1alp
...
+						}
+						if err != nil {
+							return reconcileStatusStopAndRequeue, err
+						}
+						// Only resetting the condition. Terminal condition phase will update the phase and the
+						// next reconcile will continue where the workspace left off.
+						logger.V(3).Info("LogicalCluster reappeared, recovering workspace", "cluster", workspace.Spec.Cluster)
+						conditions.MarkTrue(workspace, tenancyv1alpha1.WorkspaceInitialized)
+						// Immediately request requeueing to recover the workspace faster.
+						r.requeueAfter(workspace, time.Second)
```
gman0, line 122: Do we need to consider any other back-off period behaviors (exponential), to not spam too many retries?

ntnn: The expontential is why I added the requeue - at this point in the reconcile loop the workspace should be in an exponential backoff and this should mark the workspace as operational again.
By forcing a quick requeue I wanted to trigger the next reconcile loop quickly so the initialization can continue.

mjudeikis:
/lgtm
/approve 

lets see how it works or does not :D

PR #4090, opened 2026-04-30
Add informer.GVRLifecycleHandler and use it to drive GC

ntnn:
## Summary

There's a bug in the GC rewrite that I suspect only happens when an API is added and removed quickly and an object from the API owns another object.

The root cause of this is that the monitors might not be in sync or are still syncing while the API was added and deleted. That means the GC never sees the parent object and can never establish the edge between owner and owned object, which makes it impossible for the GC to apply a deletion policy.

This happened because the DDSIF.Subscribe was just triggering a resync, then the GC sync would get the available GVRs at that moment, update the monitors and then wait for the monitors to be synched.
This time can be too long for APIs that are added and removed quickly.

The solution is adding `GVRLifecycleHandler` to the DDSIF and resyncing the monitors with each GVR change explicitly. With each GVR change being processed and triggering a monitor sync the GC shouldn't miss any APIs, regardless of how short they live.

Needs: https://github.com/kcp-dev/kubernetes/pull/193

## What Type of PR Is This?

/kind bug
/kind flake

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


ntnn:
```text
 make imports
[...]]
go: github.com/pivotal-cf/yaml-patch/cmd/yaml-patch imports
	github.com/krishicks/yaml-patch imports
	gopkg.in/yaml.v2: cannot find module providing package gopkg.in/yaml.v2
make[1]: *** [Makefile:224: hack/tools/yaml-patch-v0.0.11] Error 1
make[1]: Leaving directory '/home/prow/go/src/github.com/kcp-dev/kcp'
```

o.O

/test pull-kcp-verify-codegen

mjudeikis (approved):
/lgtm
/approve

PR #4094, opened 2026-05-04
WorkspaceType initializerPermissions / terminatorPermissions

mjudeikis:
## Summary

This adds two new optional fields to `WorkspaceTypeSpec`:
```
initializerPermissions []rbacv1.PolicyRule
terminatorPermissions []rbacv1.PolicyRule
```

These declare RBAC rules that initializer/terminator controllers get when they reach into a workspace through the lifecycle virtual workspace's content proxy.

### How it works:

**Mode 1** — least‑privilege - when permissions set

WST is fetched from the local wildcard informer with a fallback to the cache‑server informer (covers the multi‑shard case where the WST lives on a different shard than the workspace).
The request is evaluated in‑process against the declared PolicyRules using kube's RBAC matcher. The kcp `/clusters/<id>/` prefix is stripped before computing `RequestInfo` so the standard k8s factory can parse it.

If allowed, the request is reverse‑proxied to the shard impersonating the caller, with a synthetic group `system:kcp:initializer:<initializer>` (or `system:kcp:terminator:<terminator>`) prepended.

The workspace content authorizer sees the synthetic group and short‑circuits to `DecisionAllow` (THIS IS VERY IMPORTANT PART!!!!). **Without that bypass the caller — who has no in‑workspace RBAC — would be denied.**

**Mode 2** — owner impersonation fallback:

When permissions are empty, it preserves historical behaviour — proxy impersonates LogicalCluster.Spec.CreatedBy, getting full cluster‑admin in the workspace. 
**THIS WILL BE DEPRECATED LATER ON. WE HAVE WARRNING ADDED TO CLI.**

Spoof prevention: The synthetic groups are only trustworthy because the front‑proxy strips them on ingress via --authentication-drop-groups using the prefix form.

## Why it works

1**. In-process RBAC at the VW (this evaluator)**. The VW evaluates each incoming request against the WST's declared `terminatorPermissions` before forwarding. Out-of-scope requests (e.g. `GET secret` when only `configmaps` is declared) are rejected with 403 here. This is the authoritative scope decision and runs in-process, so it cannot be bypassed by request mutation downstream.

2. **Synthetic-group bridge to the shard**. Once the in-process check passes, the VW forwards with impersonation set to the caller plus a synthetic group `system:kcp:terminator:<type>` (`authorization.TerminatorGroup`). The shard's RBAC **trusts that synthetic group as the bearer of terminator-scoped privileges** — this is how the forwarded request gets actual access to workspace content even though the caller, **by their own identity, has none**. Critically, the synthetic group is only attached after the layer-1 check has approved the request.

3. Front-proxy strips forged synthetic groups. A malicious user cannot skip the VW and assert the synthetic group themselves (e.g. via Impersonate-Group: system:kcp:terminator:...) because the front-proxy is configured with --authentication-drop-groups to strip the system:kcp:terminator:* prefix before authentication reaches the shard. So self-asserted synthetic groups never bind on the shard, and a direct shard request reduces to the user's normal identity (no privileges).

# Risks

This all builds on premises (not new, exists now, just vocalizing it here):

1. No RBAC binding ever grants impersonate on groups to a non-trusted principal.
2. Every connected IdP scrubs or controls the groups claim — no self-service IdPs, no IdPs where users can edit their own group memberships.
3. No issued client cert has` system:kcp: in O=`.

Front‑proxy bypass = full forgery. Any path into a shard that does not go through the front‑proxy (direct shard access, internal traffic, mTLS to a shard from a misconfigured client) lets an attacker self‑assert `system:kcp:initializer:*` and get unconditional access via the bypass at workspace_content_authorizer.go:112. The threat model now critically depends on the drop list being applied on every shard ingress.

Behaviour change is immediate for existing WSTs. Comment says "Changes take effect immediately for all workspaces of this type." A user who edits an existing WST to add restrictive rules will instantly break in‑flight initialization for any workspace using that type. There is no per‑workspace pin or grace period.

In‑process RBAC evaluator is a new code path. `EvaluateLifecyclePermissions` re‑implements `/clusters/<id>` prefix stripping and uses `rbac.RulesAllow` directly. Subtle deviations from server‑side RBAC (e.g., missing nonResourceURLs semantics, label selectors via cel, future RBAC features) would silently allow/deny.

`workspace_reconcile_phase.go` changes the source of truth for the WS phase. Workspaces will now flip to `Terminating/Deleting` based on local terminator length, even before the LC phase reconciler agrees. Any consumer who compares the WS phase to the LC phase will see them disagree during the deletion window.

No validation on the PolicyRule content. A user can declare wildcards verbs: ["*"], resources: ["*"], apiGroups: ["*"], which is functionally identical to Mode 2 but advertised as "least privilege". Worth either documenting or adding admission validation.

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Add workspacetype InitializerPermissions and TerminatorPermissions
```


mjudeikis:
/retest 
not sure if flake or real

docs/content/concepts/authorization/authorizers.md
```diff
@@ -130,8 +130,32 @@ By default, workspaces are only accessible to a subject if they are in `Ready` p
 can be accessed only by subjects who are granted `admin` verb on the `workspaces/content` resource in the
 parent workspace.
 
+The authorizer also permits content access while the workspace is in the `Terminating` and `Deleting` phases so
```
ntnn, line 133: Is `Deleting` correct? I thought `Terminating` should allow access and then the WS moves from `Terminating` to `Deleting` once all terminators are done and no further ws access is allowed.

staging/src/github.com/kcp-dev/sdk/apis/tenancy/v1alpha1/types_workspacetype.go
```diff
@@ -146,6 +147,32 @@ type WorkspaceTypeSpec struct {
...
+	// Rules are evaluated in-process by the VW proxy on each request; no ClusterRole or
+	// ClusterRoleBinding objects are created inside the workspace.
+	//
+	// When empty (the default), the VW content proxy falls back to impersonating the
+	// workspace owner (full cluster-admin), preserving the historical behavior.
+	//
+	// Changes take effect immediately for all workspaces of this type.
+	//
+	// +optional
+	TerminatorPermissions []rbacv1.PolicyRule `json:"terminatorPermissions,omitempty"`
```
ntnn, line 175: Maybe this and remove `.initialzer` and `.terminator` in a future v1alpha2? That would allow us to expand more for the lifecycle controllers if need be.

```go
type WorkspaceTypeSpec struct {
    Initializer *LifecycleSpec `json:"..."`
    Terminator *LifecycleSpec `json:"..."`
}

type LifecycleSpec struct {
    Permissions []rbacv1.PolicyRule `json:"permisisons,omitempty"`
}
```

ntnn: That would also allow to require permissions to be set using spec validation rather than erroring in the code.

pkg/virtual/terminatingworkspaces/builder/proxy.go
```diff
@@ -0,0 +1,54 @@
...
+					req.Header.Del(key)
+				}
+			}
+			req.URL.Scheme = forwardedHost.Scheme
+			req.URL.Host = forwardedHost.Host
+		},
+		Transport: rt,
+	}
+	proxy.ServeHTTP(writer, request)
+}
```
ntnn: This is identical to the one for the init VW, right? Maybe we should put shared stuff in `pkg/virtual/shared` or similar to deduplicate the code. I think the WST lookup is also identical and could just be a shared function.

pkg/authorization/workspace_content_authorizer.go
```diff
@@ -100,6 +100,18 @@ func (a *workspaceContentAuthorizer) Authorize(ctx context.Context, attr authori
...
+	// are injected by the initializing/terminating VW content proxies *after* they
+	// have evaluated the request against the WorkspaceType's
+	// initializer/terminator permissions. They cannot be self-asserted because the
+	// front-proxy drops them on ingress. Their presence is the authoritative
+	// "pre-authorized by VW" marker — delegating further would re-run RBAC against
+	// the caller's identity, which by design has no in-workspace permissions, so
+	// allow directly here.
+	if HasLifecycleGroup(attr.GetUser().GetGroups()) {
+		return authorizer.DecisionAllow, "lifecycle VW pre-authorized access", nil
+	}
```
ntnn: This should happen after the LC check, or? `.getLogicalCluster` checks globally

ntnn: And maybe also only if the phase is actually init/term

ntnn: Yeah, I'd rather check if LC exists and only do the `HasLifecycleGroup->Allow` if the LC is actually known to be in init/term phase.

pkg/virtual/initializingworkspaces/builder/build.go
```diff
@@ -251,12 +262,62 @@ func BuildVirtualWorkspace(
...
+						UserName: caller.GetName(),
+						UID:      caller.GetUID(),
+						// Inject the synthetic group so the workspace content authorizer
+						// trusts this request as pre-authorized by the VW. This will be used for auditing.
+						Groups: append([]string{authorization.InitializerGroup(initializer)}, caller.GetGroups()...),
+						Extra:  extra,
+					}
+					authenticatingTransport, err := rest.TransportFor(thisCfg)
+					if err != nil {
+						http.Error(writer, fmt.Sprintf("could create round-tripper: %v", err), http.StatusInternalServerError)
```
ntnn:
```suggestion
						http.Error(writer, fmt.Sprintf("could not create round-tripper: %v", err), http.StatusInternalServerError)
```

test/e2e/virtual/terminatingworkspaces/virtualworkspace_test.go
```diff
@@ -1084,3 +1084,253 @@ func TestTerminatingWorkspacesVirtualWorkspaceContentAccess(t *testing.T) {
...
+	t.Log("Self-asserted synthetic groups against the front-proxy must be stripped (no privilege escalation)")
+	// Build a config that talks to the front-proxy directly (not the VW), authenticating as
+	// user-2 (no terminate verb) but asserting the synthetic group via Impersonate. The
+	// front-proxy --authentication-drop-groups list strips system:kcp:terminator:*, so the
+	// request must reach the shard *without* the synthetic group and be denied normally.
+	directConfig := rest.AddUserAgent(rest.CopyConfig(sourceConfig), t.Name()+"-direct")
+	directConfig = framework.StaticTokenUserConfig("user-2", directConfig)
+	directConfig.Impersonate = rest.ImpersonationConfig{
+		UserName: "user-2",
+		Groups:   []string{string(terminator)}, // attempt to forge: system:kcp:terminator:<...>
```
ntnn: The group is just the terminator name, or? Shouldn't this be `"system:kcp:terminator:"+string(terminator)"`?

test/e2e/virtual/terminatingworkspaces/virtualworkspace_test.go
```diff
@@ -1084,3 +1084,253 @@ func TestTerminatingWorkspacesVirtualWorkspaceContentAccess(t *testing.T) {
 		require.True(c, errors.IsNotFound(err), "workspace not deleted yet, err: %v", err)
 	}, wait.ForeverTestTimeout, 100*time.Millisecond)
 }
+
```
SimonTheLeg, line 1088: I think we should have the equivalent of this test also for `initializingworkspace/virtualworkspace_test.go`. I know they are both pretty identical, but still there are some nuances and I would feel better if we had it.

docs/content/concepts/authorization/authorizers.md
```diff
@@ -130,8 +130,32 @@ By default, workspaces are only accessible to a subject if they are in `Ready` p
...
+the workspace.
+
 ServiceAccounts declared within a workspace don't have access to content of initializing workspaces.
 
+##### Lifecycle Synthetic Groups
+
+The initializing and terminating virtual workspace content proxies can forward requests with one of two synthetic
+groups attached to the controller's identity:
+
+* `system:kcp:initializer:<initializer-name>` (e.g. `system:kcp:initializer:root:tenant`)
```
SimonTheLeg: hmm I am wondering if we should possibly refer to this as the initializer-path? or just call it initializer? Because name for me sounds to much like the name of the wst, which it is not.

And I think it also caused some confusion on Nelos comment on the test. Because it is indeed the full path.

pkg/proxy/options/authentication.go
```diff
@@ -65,8 +65,19 @@ func NewAuthentication() *Authentication {
...
+		// that lifecycle VW content proxies inject after they have evaluated the
+		// request against the WorkspaceType's initializer/terminator permissions.
+		// Stripping them on ingress prevents clients from self-asserting them and
+		// bypassing the in-process RBAC evaluation.
+		DropGroups: []string{
+			user.SystemPrivilegedGroup,
+			bootstrap.SystemLogicalClusterAdmin,
+			bootstrap.SystemKcpInitializerGroupPrefix + "*",
+			bootstrap.SystemKcpTerminatorGroupPrefix + "*",
+		},
```
SimonTheLeg, line 80: Since we change DropGroups, we should update our filter unit tests inside `pkg/authentication/groups_test`

pkg/virtual/initializingworkspaces/builder/proxy.go
```diff
@@ -0,0 +1,54 @@
...
+	"strings"
+
+	"k8s.io/client-go/transport"
+)
+
+// serveProxy strips any client-supplied auth/impersonation headers from the request
+// and reverse-proxies it to forwardedHost using the supplied (impersonating)
+// transport. Used by the workspace-content sub-workspace handler under both modes
+// of operation (synthetic-group + caller identity, or owner impersonation fallback).
+func serveProxy(writer http.ResponseWriter, request *http.Request, forwardedHost *url.URL, rt http.RoundTripper) {
```
SimonTheLeg: probably we can combine the initializer and terminator proxies, as from what I can see they do the exact same thing.

SimonTheLeg (changes requested):
overall this looks in a pretty good state, only some minor requests from my side. One largely involves around the terminology if we want to call it <initializer-name> in our docs, as it is more of the path of it?

One thing I am thinking if we need it. I wonder if in our E2E tests we should always use one which uses inheritance between workspacetypes. Just to be sure that this does not break anything. I don't think it would, but I feel safer if we did.

test/e2e/virtual/terminatingworkspaces/virtualworkspace_test.go
```diff
@@ -1084,3 +1084,253 @@ func TestTerminatingWorkspacesVirtualWorkspaceContentAccess(t *testing.T) {
...
+	}
+	directKube, err := kcpkubernetesclientset.NewForConfig(directConfig)
+	require.NoError(t, err)
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		_, err := directKube.Cluster(wsClusterName.Path()).CoreV1().ConfigMaps(nsName).List(ctx, metav1.ListOptions{})
+		require.Error(c, err, "self-asserted synthetic group should not grant access")
+		// Either the impersonation is rejected, or the request is denied because the
+		// effective user has no rights inside the workspace.
+		require.True(c, errors.IsForbidden(err) || errors.IsUnauthorized(err) || errors.IsNotFound(err),
+			"expected forbidden/unauthorized/notfound, got: %v", err)
```
SimonTheLeg, line 1320: I cannot quite follow this, why are we checking for IsNotFound?

docs/content/concepts/authorization/authorizers.md
mjudeikis: Deleting allows access now. All hell got loose once I stopped allowing delete access :/ Apparently, a lot of cleanup processes need this, and the test just went all "red on me". So this would be breaking change :/

test/e2e/virtual/terminatingworkspaces/virtualworkspace_test.go
mjudeikis: added comment

staging/src/github.com/kcp-dev/sdk/apis/tenancy/v1alpha1/types_workspacetype.go
ntnn: As discussed we'll fix that in v1alpha2

mjudeikis: Will defer this to v1alpha2 as its too much nesting now :) all api is flat, so lets keep it flat for now

pkg/proxy/options/authentication.go
mjudeikis: updated

ntnn (approved):
/lgtm
/approve

PR #4101, opened 2026-05-05
remove default /services/ mapping in front-proxy during e2e tests

### LINKED ISSUE #3826: bug: Stop encouraging bad architecture by routing `/services/` on front-proxy to the root shard
### Describe the bug

The front-proxy is in many places configured to route `/services/` to the root shard.

E.g. in the kcp-operator: https://github.com/kcp-dev/kcp-operator/blob/24e13a58f10c384b6b9751f81c3c94d97e9bd5e1/internal/resources/frontproxy/configmap.go#L64-L78
Or in the helm-chart everything is routed as a fallback to the root shard: https://github.com/kcp-dev/helm-charts/blob/ff4b694c715b6fa07f114d4564463e07601ff772/charts/kcp/templates/front-proxy-configmap.yaml#L10-L17

The problem with this is that to users new to kcp it looks like the VWs should be available through the front-proxy and leads to e.g. downstream rewriting the root shard VW URL to the external URL - which of course only works until a second shard is added.
It also lead to a PR rewriting the VW URL to the external URL: https://github.com/kcp-dev/kcp/pull/3823

Ideally front-proxy would need no initial mapping, it already has the information for the root shard: https://github.com/kcp-dev/kcp/blob/c749f92ccf4f5b37ab149b6e719d544c64d8147f/pkg/proxy/options/options.go#L62
And should be able to build its mapping for clusters from there.

I did some smoke tests on the weekend and a bunch of tests failed when the `/services/` mapping was gone. I presume that most of these are just the wrong client being used in those tests rather than kcp itself actually failing.

### Steps To Reproduce

1. Deploy kcp in any way
2. Access root shard VWs through fp 

### Expected Behaviour

1. Deploy kcp in any way
2. Fail to access root shard VWs through fp

### Additional Context

_No response_

xrstf:
## Summary
Routing virtual-workspace traffic through the front-proxy is nowadays considered an anti-pattern. To get rid of it, we should first start by making sure none of our own tests rely on that mapping anymore.

The bigger impact will be felt in the kcp-operator, which is usually responsible for creating the front-proxy's ConfigMap. No release note on this PR here because this only affects our tests.

## What Type of PR Is This?
/kind cleanup

## Related Issue(s)
Fixes #3826

## Release Notes
```release-note
NONE
```

### DIFF
--- a/cmd/sharded-test-server/frontproxy.go
+++ b/cmd/sharded-test-server/frontproxy.go
@@ -70,14 +70,6 @@ func startFrontProxy(
 
 	mappings := []types.PathMapping{
 		{
-			Path: "/services/",
-			// TODO: support multiple virtual workspace backend servers
-			Backend:         fmt.Sprintf("https://localhost:%s", vwPort),
-			BackendServerCA: filepath.Join(workDirPath, ".kcp", "serving-ca.crt"),
-			ProxyClientCert: filepath.Join(workDirPath, ".kcp-front-proxy", "requestheader.crt"),
-			ProxyClientKey:  filepath.Join(workDirPath, ".kcp-front-proxy", "requestheader.key"),
-		},
-		{
 			Path:            "/e2e/clusters/{cluster}/",
 			Backend:         "https://localhost:2443",
 			BackendServerCA: filepath.Join(workDirPath, ".kcp", "serving-ca.crt"),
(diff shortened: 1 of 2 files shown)


ntnn (approved):
/lgtm
/approve

PR #4102, opened 2026-05-05
Split annotation for multiple-writers

### LINKED ISSUE #4089: Convert all writers of internal.apis.kcp.io/resource-bindings to SSA so Force: true can be removed
## Background

In the fix for #3926, the APIBinding reconciler was converted to Server-Side Apply when writing the `internal.apis.kcp.io/resource-bindings` annotation on `LogicalCluster` ([apibinding_controller.go:162-190](https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/apis/apibinding/apibinding_controller.go#L162-L190)). The patch uses `Force: true` because two other code paths still write the same annotation via plain `Update`, so SSA without Force would fail with `FieldManagerConflict`.

The two other writers:

1. **`logicalclustercleanup` controller** — [logicalclustercleanup_controller.go:331-333](https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/apis/logicalclustercleanup/logicalclustercleanup_controller.go#L331-L333): rebuilds the annotation for migration, removal of orphaned APIBinding entries, and CRD entry expiry. Uses `Update`.
2. **`crdnooverlappinggvr` admission plugin** — [crdnooverlappinggvr_admission.go:139-147](https://github.com/kcp-dev/kcp/blob/main/pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go#L139-L147): pre-locks a `GroupResource` for a new CRD on admission. Uses `Update` inside `RetryOnConflict`.

Because both use plain `Update`, every write reclaims ownership of the annotation under whatever field manager their client reports. The APIBinding reconciler then needs `Force: true` to take ownership back on its next SSA Apply.

## Why `Force: true` is a real risk, not just cosmetic

`Force: true` resolves field-ownership conflicts but does **not** protect against read-modify-write races. The APIBinding reconciler reads the annotation from its informer cache, computes a new `ResourceBindingsAnnotation` map ([logical_cluster_lock.go:96-145](https://github.com/kcp-dev/kcp/blob/main/pkg/reconciler/apis/apibinding/logical_cluster_lock.go#L96-L145)), and applies the marshaled JSON blob back. Anything another writer added to the live annotation after the cache read but before the Apply lands is silently overwritten.

The most concerning failure mode:

1. The `crdnooverlappinggvr` admission plugin admits CRD `crd1` for GR `foo.bar` and writes a pending lock entry.
2. The APIBinding reconciler is mid-reconcile from a stale informer read taken before the lock landed.
3. The reconciler's SSA Apply lands carrying its pre-lock view of the annotation. The pending lock for `foo.bar` is gone.
4. A second admission for an overlapping GR sees no lock and is admitted.

The admission plugin's locking guarantee is weakened. The cleanup controller eventually reconciles, but by then both CRDs may exist.

## Why this is structurally a single-blob problem

SSA tracks ownership at the **field** level. Annotations are a `map[string]string`, so each annotation key is a separate field, but the *value* of a key is opaque — there is no sub-ownership inside a JSON blob. Three managers cannot co-own different entries inside one JSON-encoded annotation. Whoever writes that key last owns the whole string. That is why moving the other two writers to SSA without restructuring the data model would not help: they would still conflict on the same field.

## Proposed fix

**Option A — split the annotation by writer (smaller change)**

Replace the single `internal.apis.kcp.io/resource-bindings` annotation with one annotation per writer, each with its own field manager:

| Annotation key | Field manager | Owner |
|---|---|---|
| `internal.apis.kcp.io/locks-bindings` | `kcp-apibinding` | apibinding controller (entries for its APIBinding's bound resources) |
| `internal.apis.kcp.io/locks-crds` | `kcp-logicalclustercleanup` | cleanup controller (established CRD entries + expiry) |
| `internal.apis.kcp.io/locks-pending` | `kcp-crdnooverlappinggvr` | admission plugin (pending CRD locks) |

Readers (`GetResourceBindings`, the dynamic rest mapper, the replication authorizer) union the three maps with a defined precedence (`crds` > `bindings` > `pending`). Each writer SSA-applies only its own key, so no field-ownership overlap and `Force: true` is not needed.

Trade-off: today, `WithLockedResources` decides "lock or skip" against the union map in one read-modify-write. With three separate writers, two of them can both lock the same GR concurrently, and conflict resolution moves to read time. For an optimistic locking design this is acceptable — the cleanup controller already arbitrates inconsistencies — but it is a real semantic shift to call out in review.

**Option B — single SSA writer (cleaner end state, larger refactor)**

Make the cleanup controller the only writer of the annotation. The APIBinding reconciler stops writing the annotation and only reconciles its own APIBinding object; cleanup controller already watches APIBindings and recomputes the annotation. The hard case is admission: it must lock synchronously before admitting the CRD, so it cannot wait on an async controller. Under Option B, admission would write a separate `ResourceLock` CR (one per GR), and the cleanup controller would fold those into the annotation.

## Acceptance criteria

- `Force: true` removed from the APIBinding reconciler's SSA Apply on `LogicalCluster`.
- Concurrent writes from the APIBinding reconciler, cleanup controller, and `crdnooverlappinggvr` admission plugin do not produce `FieldManagerConflict` errors and do not silently overwrite each other's entries.
- Existing readers (`GetResourceBindings`, `dynamicrestmapper_dynamic_controller.go`, the replication authorizer content tests) continue to observe the union of all three writers' state.
- Migration path for existing `LogicalCluster` resources carrying the old `internal.apis.kcp.io/resource-bindings` annotation.

## Related

- Follow-up to #3926 (the PR that introduced SSA + `Force: true` for the APIBinding reconciler narrows but does not close the cross-writer race.)

mjudeikis:
## Summary

Currently, we have a lot of write conflicts on logicalcluster as 3 operators write to the same annotations. 
This PR moves all of them to separate annoations so we can use SSA and field ownerships 

## What Type of PR Is This?
/kind cleanup

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/4089

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
+++ b/pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
@@ -18,6 +18,7 @@ package crdnooverlappinggvr
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"strings"
@@ -27,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apiserver/pkg/admission"
 	"k8s.io/apiserver/pkg/endpoints/request"
 	"k8s.io/client-go/util/retry"
@@ -80,7 +82,33 @@ func (p *crdNoOverlappingGVRAdmission) SetKcpInformers(local, global kcpinformer
 // SetKcpClusterClient sets the client for kcp resources. It's part of WantsKcpClusterClient.
 func (p *crdNoOverlappingGVRAdmission) SetKcpClusterClient(c kcpclientset.ClusterInterface) {
 	p.updateLogicalCluster = func(ctx context.Context, logicalCluster *corev1alpha1.LogicalCluster, opts metav1.UpdateOptions) (*corev1alpha1.LogicalCluster, error) {
-		return c.CoreV1alpha1().LogicalClusters().Cluster(logicalcluster.From(logicalCluster).Path()).Update(ctx, logicalCluster, opts)
+		// Use SSA to write only to the pending locks annotation key
+		patchObj := &corev1alpha1.LogicalCluster{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: corev1alpha1.SchemeGroupVersion.String(),
+				Kind:       "LogicalCluster",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name: logicalCluster.Name,
+				Annotations: map[string]string{
+					apibinding.LocksPendingAnnotationKey: logicalCluster.Annotations[apibinding.LocksPendingAnnotationKey],
+				},
+			},
+		}
+		patchBytes, err := json.Marshal(patchObj)
+		if err != nil {
+			return nil, err
+		}
+		return c.CoreV1alpha1().LogicalClusters().Cluster(logicalcluster.From(logicalCluster).Path()).Patch(
+			ctx,
+			logicalCluster.Name,
+			types.ApplyPatchType,
+			patchBytes,
+			metav1.PatchOptions{
+				FieldManager: apibinding.FieldManagerPending,
+				// Force is not needed because each writer uses its own annotation key
+			},
+		)
 	}
 }
 
@@ -136,7 +164,7 @@ func (p *crdNoOverlappingGVRAdmission) Validate(ctx context.Context, a admission
 		}
 
 		var updated *corev1alpha1.LogicalCluster
-		updated, _, skipped, err = apibinding.WithLockedResources(nil, time.Now(), lc, []schema.GroupResource{gr}, apibinding.ExpirableLock{
+		updated, _, skipped, err = apibinding.WithLockedResourcesForPending(nil, time.Now(), lc, []schema.GroupResource{gr}, apibinding.ExpirableLock{
 			Lock:      apibinding.Lock{CRD: true},
 			CRDExpiry: ptr.To(p.now()),
 		})
--- a/pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission_test.go
+++ b/pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission_test.go
@@ -140,7 +140,7 @@ func TestValidate(t *testing.T) {
 		},
 
 		{
-			name: "fails without resource binding annotation on LogicalCluster",
+			name: "succeeds without resource binding annotation on LogicalCluster - creates pending lock",
 			attr: createAttr(&apiextensions.CustomResourceDefinition{
 				ObjectMeta: metav1.ObjectMeta{Name: "test"},
 				Spec: apiextensions.CustomResourceDefinitionSpec{
@@ -150,7 +150,7 @@ func TestValidate(t *testing.T) {
 			}),
 			clusterName:    "root:acme",
 			initialObjects: []runtime.Object{createLogicalCluster("root:acme")},
-			wantErr:        true,
+			wantAnnotation: `{"foo.acme.dev":{"c":true,"e":"2022-01-01T00:00:00Z"}}`,
 		},
 		{
 			name: "fails without LogicalCluster",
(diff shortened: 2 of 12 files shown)


ntnn (approved):
/lgtm
/approve

ntnn:
/retest

flake?

```text
=== FAIL: test/e2e/proxy TestMappingWithClusterContext (31.18s)
    proxy_test.go:49: Shared kcp server will target configuration "/home/prow/go/src/github.com/kcp-dev/kcp/.kcp/admin.kubeconfig"
    proxy_test.go:49: Waiting for readiness for server at https://127.0.0.1:6444/
    proxy_test.go:84: Waiting for tenancy.kcp.io/v1alpha1 to be served
    proxy_test.go:84: Waiting for tenancy.kcp.io/v1alpha1 to be served
    proxy_test.go:84: Waiting for tenancy.kcp.io/v1alpha1 to be served
    proxy_test.go:84: Waiting for topology.kcp.io/v1alpha1 to be served
    proxy_test.go:84: Waiting for topology.kcp.io/v1alpha1 to be served
    proxy_test.go:84: Created root:universal workspace root:e2e-workspace-oidc-cvcjh as /clusters/qukrku7l79v8csda on shard "root"
    mockoidc.go:187: Creating WorkspaceAuthenticationConfguration mockoidc-5456868631789275145...
    authfixtures.go:52: Creating WorkspaceType with-oidc...
    proxy_test.go:103: Creating Workspace...
    proxy_test.go:104: Created root:e2e-workspace-oidc-cvcjh:with-oidc workspace root:e2e-workspace-oidc-cvcjh:team-a as /clusters/14jsb23pvjb2qv6h on shard "shard-1"
    authfixtures.go:71: Creating ClusterRoleBinding...
    proxy_test.go:148: 
        	Error Trace:	/home/prow/go/src/github.com/kcp-dev/kcp/test/e2e/proxy/proxy_test.go:148
        	            				/usr/local/go/src/runtime/asm_arm64.s:1268
        	Error:      	Not equal: 
        	            	expected: "cluster:14jsb23pvjb2qv6h"
        	            	actual  : ""
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-cluster:14jsb23pvjb2qv6h
        	            	+
        	Test:       	TestMappingWithClusterContext
    proxy_test.go:139: 
        	Error Trace:	/home/prow/go/src/github.com/kcp-dev/kcp/test/e2e/proxy/proxy_test.go:139
        	Test:       	TestMappingWithClusterContext
DONE 278 tests, 5 skipped, 1 failure in 373.368s
I0513 08:37:30.503413   22846 shard.go:248] "gathering shard metrics" shard="kcp-0"
I0513 08:37:30.642486   22846 shard.go:272] "writing metrics file" shard="kcp-0" path="/logs/artifacts/kcp/kcp-0-metrics.txt"
I0513 08:37:30.644374   22846 shard.go:277] "wrote metrics file" shard="kcp-0" path="/logs/artifacts/kcp/kcp-0-metrics.txt"
I0513 08:37:30.644415   22846 shard.go:248] "gathering shard metrics" shard="kcp-1"
I0513 08:37:30.775585   22846 shard.go:272] "writing metrics file" shard="kcp-1" path="/logs/artifacts/kcp/kcp-1-metrics.txt"
I0513 08:37:30.777499   22846 shard.go:277] "wrote metrics file" shard="kcp-1" path="/logs/artifacts/kcp/kcp-1-metrics.txt"
make: *** [Makefile:379: test-e2e-sharded-minimal] Error 1
```

sttts:
Isn't this one annotation meant as a lock and splitting it up would defeat it's purpose? Better be extra careful understanding why it exists.

sttts:
Compare https://github.com/kcp-dev/kcp/pull/3251. The conflict was intentional I believe. I don't remember all the details from the time.

pkg/admission/crdnooverlappinggvr/crdnooverlappinggvr_admission.go
```diff
@@ -80,7 +82,33 @@ func (p *crdNoOverlappingGVRAdmission) SetKcpInformers(local, global kcpinformer
...
+				Kind:       "LogicalCluster",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name: logicalCluster.Name,
+				Annotations: map[string]string{
+					apibinding.LocksPendingAnnotationKey: logicalCluster.Annotations[apibinding.LocksPendingAnnotationKey],
+				},
+			},
+		}
+		patchBytes, err := json.Marshal(patchObj)
```
sttts, line 98: I am pretty sure this is wrong. The update was intentionally a consistent write to ensure that the passed `logicalCluster` and the one in etcd are the same resource version. Patching blindly with server side apply defeats the purpose of the whole approach.

mjudeikis:
will re-open later

PR #4122, opened 2026-05-12
Wait for authenticators to be initialized

### LINKED ISSUE #4115: bug: TokenReview oidc authenticator isn't initialized when workspace is initialized
### Describe the bug

In 0.31.1 kcp version it seems that oidc-authenticator which is configured in WorkspaceAuthenticationConfiguration resource is initialized after the first TokenReview is created, because TokenReview in status has an error `invalid bearer token, oidc: authenticator not initialized`, but when the second request comes everything is fine. It doesn't matter how much time has passed after the workspace is initialied, the first TokenReview request gets 401 and the immediately coming second request gets authenticated response. 

### Steps To Reproduce

If's it's sufficient, for reproducing you can run` task local-setup` from this branch https://github.com/platform-mesh/helm-charts/pull/1663.

After the setup is ready, you need to:
1. login into the https://portal.localhost:8443/
2. create an organization
3. switch into the org
4. navigate into acocunts page

It will trigger TokenReview request. First TokenReview will have in status  `invalid bearer token, oidc: authenticator not initialized`. 

Keycloak is configured as an oidc authenticator for workspaces by `WorkspaceAuthenticationConfiguration` resource in` :root:orgs` workspace. It happens during organization initialization and at the moment of tokenReview request the workspace is already initialized

### Expected Behaviour

oidc initializer is initialized at the moment when workspace is initialized and first request returns authenticated.

### Additional Context

_No response_

ntnn:
## Summary

The authenticators do have a healthcheck, however this healthcheck is not exposed and it is not easy to expose without intrusive changes.

An alternative would've been to use the update function returned by kubeAuthConfig.New which _does_ wait until the authenticator's healthcheck are ready, however that would create each authenticator twice and just discard the first iteration.

Instead the authenticator is validated by continually authenticating a dummy token created with the first issuer until the error is no longer "not initialized".

## What Type of PR Is This?

/kind bug

/cherry-pick release-0.31

## Related Issue(s)

Fixes #4115

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/authentication/index.go
+++ b/pkg/authentication/index.go
@@ -18,9 +18,14 @@ package authentication
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 	"fmt"
+	"net/http"
+	"strings"
+	"time"
 
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apiserver/pkg/authentication/authenticator"
 	"k8s.io/klog/v2"
 	kubeauthenticator "k8s.io/kubernetes/pkg/kubeapiserver/authenticator"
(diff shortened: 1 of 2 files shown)


cnvergence (approved):

mjudeikis (approved):
/lgtm
/approve

PR #4125, opened 2026-05-13
Auth init v2

mjudeikis:
## Summary

Attempt to fix the init state. 

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

Issue #4100, opened 2026-05-05
feature: Rewrite front-proxy to be a simple proxy


ntnn:
### Feature Description

Currently the front-proxy doesn't just route user requests to the correct shard - it also handles authn/authz, which the shards trust.
That leads to discrepancies in how authn/authz works depending on where the user connects, leading to issues like #4066.
This could theoretically also lead to security issues where authz is properly handled in front-proxy but not in shards or VW e.g.

Also see #3826 

### Proposed Solution

1. Handle per-workspace authentication with WACs entirely to shards
1.1. WACs must be cached in the cache-server
1.2. Workspace authenticators must be build eagerly in shards (like currently in front-proxy) - but only for logical clusters on that shard
2. Drop all authn/authz handling from front-proxy and move it to the shards, the front-proxy should only proxy requests to shards for clusters it knows
2.1. There's some rewriting happening with Service Accounts, I think that can be dropped entirely.
3. Shards should not implicitly trust any connection - currently they trust the front-proxy through the RequestHeader authenticator, they should always validate instead
4. front-proxy should still allow users to mount a mapping table for global services / virtual workspaces but should not forward anything by itself. That also means that global service / VWs will have to do authn/authz themselves.

Means front-proxy should only proxy clusters it knows from its discovery. When it doesn't know a cluster it proxies to a random shard.
It cannot directly deny the request because that would allow malicious actors to basically port knock for cluster names and observing the response time.
We also can't default to proxying to the root shard because that would allow a similar attack vector and put unnecessary pressure on the root shard.
We'll have to see how this works with TLS etcpp - if this is passthrough an attacker could potentially get information on which shard answered from the response headers.

The front-proxy also can't to TLS termination anymore - when authn/authz moves to the shards certificate-based authn/authz must still work.

### Want to contribute?

- [ ] I would like to work on this issue.

olamilekan000:
This looks interesting and also an opportunity to get my hands dirty with shards once more. @ntnn should I assign it to myself or do you have plans to pick it up?

ntnn:
@olamilekan000 Please do! I wouldn't mind detangling that myself but I wouldn't get to it for a while :D

mjudeikis:
this will be big effort. I suspect a lot of small prs, not few big ones. Take it slow :)

xrstf:
> The front-proxy also can't to TLS termination anymore

How would the proxy then work, if it can't inspect the URL to figure out the target cluster (and thereby, shard)?

[...]

PR #4126, opened 2026-05-13
Terser authentication logic in front-proxy

ntnn:
## Summary

That logic never sat right with me and it seems to cause a logic error.

The original error was fixed with https://github.com/kcp-dev/kcp/pull/4125, which moved the front-proxy back to its old behaviour that relies on essentially a race for the test to function:

The root cause of the error was that an authenticated request came through but the authenticator wasn't yet ready, so the original code here: 
https://github.com/kcp-dev/kcp/blob/5a9d2e2518464cb32e9871e543b5db2b38902ede/pkg/proxy/filters/filters.go#L49-L59

Would return `nil, false` from here https://github.com/kcp-dev/kcp/blob/5a9d2e2518464cb32e9871e543b5db2b38902ede/pkg/authentication/filters.go#L67-L73

Because the handler that should add the workspace authenticator in the cotext silently fails here:
https://github.com/kcp-dev/kcp/blob/5a9d2e2518464cb32e9871e543b5db2b38902ede/pkg/authentication/filters.go#L41-L44

Because of the change that validated authenticators before returning them the authenticator was in the index later than the test assumed, turning this into a race.

Anyhow - this doesn't fix that issue. I just noticed this issue because of that.

This fixes another problem that currently isn't tested - if a request with a static token comes in for a workspace that has per-workspace authentication configured the request will get denied.

I skipped writing a test for this because this will be fixed as part of https://github.com/kcp-dev/kcp/issues/4100

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/proxy/filters/filters.go
+++ b/pkg/proxy/filters/filters.go
@@ -27,57 +27,44 @@ import (
 	authenticatorunion "k8s.io/apiserver/pkg/authentication/request/union"
 	genericapifilters "k8s.io/apiserver/pkg/endpoints/filters"
 	"k8s.io/apiserver/pkg/endpoints/request"
-	"k8s.io/klog/v2"
 
 	"github.com/kcp-dev/kcp/pkg/authentication"
 )
 
-// WithOptionalAuthentication creates a handler that authenticates a request
-// if a ClientCert is presented but passes through to the next handler if one is
-// not. It will also call the authenticator present in the request context, if
-// any, to support per-workspace authentication.
-// When additionalAuthMethods is true we also attempt to authenticate even when
-// no client cert is detected in the request.
-func WithOptionalAuthentication(handler, failed http.Handler, auth authenticator.Request, additionalAuthMethods bool) http.Handler {
+// WithOptionalAuthentication creates a handler that authenticates
+// a request if credentials are presented but passes through to the next
+// handler if one is not.
+func WithOptionalAuthentication(handler, failed http.Handler, auth authenticator.Request) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-		// Do not accidentally modify the variables in the scope of WithOptionalAuthentication().
-		var (
-			authenticator  = auth
-			additionalAuth = additionalAuthMethods
-		)
-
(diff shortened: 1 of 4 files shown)


pkg/proxy/filters/filters.go
```diff
@@ -33,50 +33,45 @@ import (
 )
 
 // WithOptionalAuthentication creates a handler that authenticates a request
-// if a ClientCert is presented but passes through to the next handler if one is
-// not. It will also call the authenticator present in the request context, if
-// any, to support per-workspace authentication.
-// When additionalAuthMethods is true we also attempt to authenticate even when
-// no client cert is detected in the request.
-func WithOptionalAuthentication(handler, failed http.Handler, auth authenticator.Request, additionalAuthMethods bool) http.Handler {
+// if credentials is presented but passes through to the next handler if one is
```
ntnn:
```suggestion
// if credentials are presented but passes through to the next handler if one is
```

pkg/proxy/options/authentication.go
```diff
@@ -72,19 +71,6 @@ func NewAuthentication() *Authentication {
 	return auth
 }
 
-// When configured to enable auth other than ClientCert, this returns true.
-func (c *Authentication) AdditionalAuthEnabled() bool {
```
mjudeikis, line 76: Will be suprised if this will not break anything.

ntnn:
This whole thing is a mess.

Basically the front-proxy tries to authenticate a request but just lets it through if it doesn't succeed unless the workspace has per-workspace auth configured.
Then it actually fails.

This causes a problem in the tests because there the shards have fixed identities with tokens (from the csv file) that the front-proxy doesn't know about. This only works because the `additionalAuth` boolean is false for the front-proxy and is only set for per-workspace auth explicitly. And this in turn doesn't show up in the tests because the per-worspace auth tests only use per-workspace auth and the tests using static tokens don't.

This also hides a bug if a static token is used to access a worksapce with per-workspace auth - this will get declined on the front-proxy simply because per-workspace auth does not recognize the static token.

PR #4127, opened 2026-05-14
Fix ws tree with clusters url

mjudeikis:
## Summary

`ws tree` gets stuck if the URL contains `/clusters/<name>` and it creates an infinite loop. 

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


ntnn (approved):
/lgtm
/approve

Issue #3350, opened 2025-03-25
bug: Creating and then deleting workspaces causes leaked goroutines


julian-hj:
### Describe the bug

baseline pprof from kcp before workspace creation/deletion: 
[pre-ws.pprof.txt](https://github.com/user-attachments/files/19456964/pre-ws.pprof.txt)

pprof after creating and then deleting an org workspace 100 times:
[post-ws.pprof.txt](https://github.com/user-attachments/files/19456977/post-ws.pprof.txt)

go tool rendering of pprof:

![Image](https://github.com/user-attachments/assets/afcf5f69-6938-4683-97e6-946e7e1aa6dd)

### Steps To Reproduce

1. Patch kcp to enable pprof
[addpprof.patch](https://github.com/user-attachments/files/19457006/addpprof.patch)
2. Start kcp in tilt and set the kubeconfig to the kind cluster
3. port forward pprof port: `kubectl port-forward -n kcp-alpha deployment/alpha 6060:6060 &`
4. get a baseline pprof: `http://localhost:6060/debug/pprof/goroutine > pre-ws.pprof`
5. create a simple yaml file for an organization workspace: [org-ws.yaml.txt](https://github.com/user-attachments/files/19457044/org-ws.yaml.txt)
6. target kcp, and then create and delete the same org 100 times in a loop: `for i in {1..100}; do k apply -f org-ws.yaml; sleep 3; k delete -f org-ws.yaml; sleep 3; done`
7. collect pprof again and observe the large number of goroutines.

### Expected Behaviour

Fewer goroutines

### Additional Context

This looks related to #3016 although that issue is long since fixed, and didn't require new workspaces.

As far as I can tell, creating a new workspace causes the creation of listeners that wont get cancelled until the sharedIndexInformer is terminated.

embik:
Hi @julian-hj, thanks for reporting this issue in such detail! We'll take a look at fixing this.

ntnn:
@embik Could you please assign the issue to me? Thanks!

embik:
/assign @ntnn

Thanks!

sanjimoh:
Repro was straightforward with the mentioned steps from @julian-hj . Below are some of my observations from the collected pprofs.

There is a significant spike in Goroutine Count of ~4.6x
- Pre-workspace creation/deletion: 3,392 goroutines
- Post-workspace creation/deletion: 15,592 goroutines

Major Contributors to Goroutine Growth deducted from the pprof:

[...]

PR #4138, opened 2026-05-21
Add client evictor & investigastion docs

### LINKED ISSUE #4071: deleting logical clusters does not free up memory
On large scale installations, larger chunks of memory are being retained after logicalclusters are deleted. The memory also is not fully re-used on the creation of new logicalclusters. This can even lead to OOMs.

Image

The effect can also be observed over multiple days where memory is still in use

Image

Deleting the pods frees up the memory again.

Image

We should investigate where the memory is being held and how to release it properly or re-use it on the second creation run.

mjudeikis:
## Summary

Whats in this PR:

- `pkg/pproflabels/labels.go` — pprof label helper, wired into kubequota controllers
- `staging/.../apimachinery/pkg/client/constructor.go` — Evict + global EvictCluster registry
- `pkg/server/clientcache_evictor.go + pkg/server/server.go` — LogicalCluster delete → EvictCluster
- `Makefile` — fix prometheus target so UGET_PRINT_PATH=absolute works when binary cached (needed for ` ./hack/run-with-prometheus.sh ./bin/kcp start --debug-socket-path=/tmp/kcp-debug.sock` to work locally on mac)
- `docs/content/developers/investigations/memory-leak-attribution.md` — full investigation guide

Impact:

Targeted bug (`clientCache` retention): 213 MB → 3.7 MB cum (98% recovery)
Total leak per workspace: 745 KB → 562 KB (-183 KB)
Goroutine leak class from #3350: already mostly fixed, wiring confirms attribution works

Next investigations identified (separate PRs):

- 200 MB cum on `structToUnstructured` — likely OpenAPI v3 controller's per-cluster spec cache
- 99 MB on j`son.objectInterface` — JSON-decoded objects still pinned by something
- `pkg/index/index.go` `DeleteLogicalCluster` incomplete (5 of 8 per-cluster maps not cleaned)

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/4071 (one of few fixes, more will come)

## Release Notes

```release-note
Add client evictor to release memory on LogicalCluster deletes
```

### DIFF
--- a/pkg/admission/kubequota/kubequota_admission.go
+++ b/pkg/admission/kubequota/kubequota_admission.go
@@ -44,6 +44,7 @@ import (
 	corev1alpha1listers "github.com/kcp-dev/sdk/client/listers/core/v1alpha1"
 
 	"github.com/kcp-dev/kcp/pkg/admission/initializers"
+	"github.com/kcp-dev/kcp/pkg/pproflabels"
 )
 
 // PluginName is the name of this admission plugin.
@@ -162,7 +163,7 @@ func (k *KubeResourceQuota) Validate(ctx context.Context, a admission.Attributes
 }
 
 // getOrCreateDelegate creates a resourcequota.QuotaAdmission plugin for clusterName.
-func (k *KubeResourceQuota) getOrCreateDelegate(clusterName logicalcluster.Name) (*stoppableQuotaAdmission, error) {
+func (k *KubeResourceQuota) getOrCreateDelegate(clusterName logicalcluster.Name) (retDelegate *stoppableQuotaAdmission, retErr error) {
 	k.lock.RLock()
 	delegate := k.delegates[clusterName]
 	k.lock.RUnlock()
@@ -181,41 +182,50 @@ func (k *KubeResourceQuota) getOrCreateDelegate(clusterName logicalcluster.Name)
 
 	// Set up a context that is cancelable and that is bounded by k.serverDone
 	ctx, cancel := context.WithCancel(context.Background())
-	go func() {
-		// Wait for either the context or the server to be done. If it's the server, cancel the context.
-		select {
-		case <-ctx.Done():
-		case <-k.serverDone:
+
+	// Label the per-cluster init for the duration of this function so that the
+	// supervisor goroutine below, plus any processorListener goroutines spawned
+	// by ValidateInitialization() via AddEventHandler on the shared informer,
+	// inherit pprof labels and remain attributable if they leak (#4071, #3350).
+	pproflabels.Cluster(ctx, PluginName, clusterName, func(ctx context.Context) {
+		go func() {
+			// Wait for either the context or the server to be done. If it's the server, cancel the context.
+			select {
+			case <-ctx.Done():
+			case <-k.serverDone:
+				cancel()
+			}
+		}()
+
+		const evaluatorWorkersPerWorkspace = 5
+		quotaAdmission, err := resourcequota.NewResourceQuota(k.userSuppliedConfiguration, evaluatorWorkersPerWorkspace)
+		if err != nil {
 			cancel()
+			retErr = err
+			return
 		}
-	}()
 
-	const evaluatorWorkersPerWorkspace = 5
-	quotaAdmission, err := resourcequota.NewResourceQuota(k.userSuppliedConfiguration, evaluatorWorkersPerWorkspace)
-	if err != nil {
-		cancel()
-		return nil, err
-	}
+		delegate = &stoppableQuotaAdmission{
+			QuotaAdmission: quotaAdmission,
+			stop:           cancel,
+		}
 
-	delegate = &stoppableQuotaAdmission{
-		QuotaAdmission: quotaAdmission,
-		stop:           cancel,
-	}
+		delegate.SetDrainedNotification(ctx.Done())
+		delegate.SetResourceQuotaLister(k.scopingResourceQuotaInformer.Cluster(clusterName).Lister())
+		delegate.SetResourceQuotaInformer(k.scopingResourceQuotaInformer.Cluster(clusterName).Informer())
+		delegate.SetExternalKubeClientSet(k.kubeClusterClient.Cluster(clusterName.Path()))
+		delegate.SetQuotaConfiguration(k.quotaConfiguration)
 
-	delegate.SetDrainedNotification(ctx.Done())
-	delegate.SetResourceQuotaLister(k.scopingResourceQuotaInformer.Cluster(clusterName).Lister())
-	delegate.SetResourceQuotaInformer(k.scopingResourceQuotaInformer.Cluster(clusterName).Informer())
-	delegate.SetExternalKubeClientSet(k.kubeClusterClient.Cluster(clusterName.Path()))
-	delegate.SetQuotaConfiguration(k.quotaConfiguration)
+		if err := delegate.ValidateInitialization(); err != nil {
+			cancel()
+			retErr = err
+			return
+		}
 
-	if err := delegate.ValidateInitialization(); err != nil {
-		cancel()
-		return nil, err
-	}
-
-	k.delegates[clusterName] = delegate
-
-	return delegate, nil
+		k.delegates[clusterName] = delegate
+		retDelegate = delegate
+	})
+	return retDelegate, retErr
 }
 
 type stoppableQuotaAdmission struct {
--- a/pkg/admission/kubequota/kubequota_admission.go
+++ b/pkg/admission/kubequota/kubequota_admission.go
@@ -163,7 +163,11 @@ func (k *KubeResourceQuota) Validate(ctx context.Context, a admission.Attributes
 }
 
 // getOrCreateDelegate creates a resourcequota.QuotaAdmission plugin for clusterName.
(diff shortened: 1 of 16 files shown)


staging/src/github.com/kcp-dev/apimachinery/pkg/client/constructor.go
```diff
@@ -34,18 +34,64 @@ type Constructor[R any] struct {
 type Cache[R any] interface {
 	ClusterOrDie(clusterPath logicalcluster.Path) R
 	Cluster(clusterPath logicalcluster.Path) (R, error)
+	// Evict drops the cached client for clusterPath, if any. Used to release
+	// per-cluster client state (REST clients, codec factories, parsed
+	// schemas) when a logical cluster is deleted. Safe to call concurrently
+	// with Cluster / ClusterOrDie. No-op if the path is not cached.
+	Evict(clusterPath logicalcluster.Path)
+}
```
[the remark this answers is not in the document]
mjudeikis: I would like to make this breaking so downstream clients MUST provide eviction, else it just yet-another-leak

pkg/server/server.go
```diff
@@ -493,6 +493,7 @@ func (s *Server) Run(ctx context.Context) error {
 		go s.CacheKcpSharedInformerFactory.Apis().V1alpha2().APIExports().Informer().Run(hookContext.Done())
 		go s.CacheKcpSharedInformerFactory.Cache().V1alpha1().CachedResources().Informer().Run(hookContext.Done())
 		go s.CacheKcpSharedInformerFactory.Cache().V1alpha1().CachedResourceEndpointSlices().Informer().Run(hookContext.Done())
+		installClientCacheEvictor(ctx, s.KcpSharedInformerFactory.Core().V1alpha1().LogicalClusters())
```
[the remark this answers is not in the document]
mjudeikis: good point

staging/src/github.com/kcp-dev/apimachinery/pkg/client/constructor.go
ntnn, line 184: Can't this then lead to leaked clients?
If a client is being built for a cluster and the client does not exist, so it drops out of .RUnlock, build the client, then acquires the lock and writes the new client. If inbetween an Evict on the same cluster happened the client will be left over.

staging/src/github.com/kcp-dev/apimachinery/pkg/client/constructor.go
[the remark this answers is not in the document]
ntnn: But is this something of interest for downstream? kcp-internal only has this problem because we have the shard-cluster relation and the internal view.
External clients should only see the clusters.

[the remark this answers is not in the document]
ntnn: And since they only see the clusters they won't have this problem.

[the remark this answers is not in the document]
mjudeikis: I mean, the idea was that one could make this external, but I opted out of this. So we get compile errors :)

staging/src/github.com/kcp-dev/apimachinery/pkg/client/constructor.go
mjudeikis: This updated to use map[string]stuff to track it. should be good now @ntnn

staging/src/github.com/kcp-dev/apimachinery/pkg/client/constructor.go
```diff
@@ -34,18 +34,65 @@ type Constructor[R any] struct {
...
+	evictors   []Evictor
+)
+
+// RegisterEvictor adds e to the set of caches notified by EvictCluster.
+// NewCache calls this automatically.
+func RegisterEvictor(e Evictor) {
+	evictorsMu.Lock()
+	defer evictorsMu.Unlock()
+	evictors = append(evictors, e)
+}
```
ntnn: That also means that evictors are never deregistered.
If clients are repeatedly build, each with their own cache, this would lead to them never being GC'd because they are always referenced here.

staging/src/github.com/kcp-dev/apimachinery/pkg/client/constructor.go
```diff
@@ -94,8 +152,22 @@ func (c *clientCache[R]) Cluster(clusterPath logicalcluster.Path) (R, error) {
 	if exists {
 		return cachedClient, nil
 	}
+	if c.evictGen[clusterPath] != genAtStart {
+		// An Evict raced with this build. Hand the freshly-built client to the
+		// caller so its in-flight request can complete, but do not cache it —
+		// the cluster has been signalled as gone.
+		return instance, nil
+	}
```
ntnn: We're only skipping caching if the client is built during an eviction. 
If a client is requested after the cluster has been evicted we'd still be caching the (dead) client and hand it back to the caller.

pkg/server/server.go
```diff
@@ -493,6 +493,7 @@ func (s *Server) Run(ctx context.Context) error {
 		go s.CacheKcpSharedInformerFactory.Apis().V1alpha2().APIExports().Informer().Run(hookContext.Done())
 		go s.CacheKcpSharedInformerFactory.Cache().V1alpha1().CachedResources().Informer().Run(hookContext.Done())
 		go s.CacheKcpSharedInformerFactory.Cache().V1alpha1().CachedResourceEndpointSlices().Informer().Run(hookContext.Done())
+		installClientCacheEvictor(hookCtx, s.KcpSharedInformerFactory.Core().V1alpha1().LogicalClusters())
```
ntnn: As in the other PR, I think this should be installed after the wait below to ensure the informer is synced before kicking off the reconciler.

pkg/reconciler/kubequota/kubequota_controller.go
```diff
@@ -252,78 +253,85 @@ func (c *Controller) process(ctx context.Context, key string) error {
...
-		IgnoredResourcesFunc: quotaConfiguration.IgnoredResources,
-		InformersStarted:     c.informersStarted,
-		Registry:             generic.NewRegistry(quotaConfiguration.Evaluators()),
-	}
-
-	resourceQuotaController, err := resourcequota.NewController(ctx, resourceQuotaControllerOptions)
-	if err != nil {
-		return err
-	}
+func (c *Controller) startQuotaForLogicalCluster(ctx context.Context, clusterName logicalcluster.Name) (retErr error) {
```
ntnn:
```suggestion
func (c *Controller) startQuotaForLogicalCluster(ctx context.Context, clusterName logicalcluster.Name) error {
```

No named returns

pkg/admission/kubequota/kubequota_admission.go
```diff
@@ -162,7 +163,7 @@ func (k *KubeResourceQuota) Validate(ctx context.Context, a admission.Attributes
 }
 
 // getOrCreateDelegate creates a resourcequota.QuotaAdmission plugin for clusterName.
-func (k *KubeResourceQuota) getOrCreateDelegate(clusterName logicalcluster.Name) (*stoppableQuotaAdmission, error) {
+func (k *KubeResourceQuota) getOrCreateDelegate(clusterName logicalcluster.Name) (retDelegate *stoppableQuotaAdmission, retErr error) {
```
ntnn:
```suggestion
func (k *KubeResourceQuota) getOrCreateDelegate(clusterName logicalcluster.Name) (*stoppableQuotaAdmission, error) {
```

No named returns

staging/src/github.com/kcp-dev/apimachinery/pkg/client/constructor.go
mjudeikis: and this is fine, we might be in the deletion flow. Client will just give error and reconciler will retry

ntnn: Ah the then dead client is fine because it's cache is empty?

staging/src/github.com/kcp-dev/apimachinery/pkg/client/constructor.go
```diff
@@ -87,15 +169,35 @@ func (c *clientCache[R]) Cluster(clusterPath logicalcluster.Path) (R, error) {
...
 	return instance, nil
 }
+
+// Evict drops the cached client for clusterPath, if any, and records the
+// path so future Cluster() calls do not re-cache for it.
+func (c *clientCache[R]) Evict(clusterPath logicalcluster.Path) {
+	c.Lock()
+	defer c.Unlock()
+	delete(c.clientsByClusterPath, clusterPath)
+	c.evicted[clusterPath] = struct{}{}
```
ntnn, line 202: Oh... I don't know if this is a good idea. Technically someone could exhaust a shard by creating and deleting workspaces until the shard is restarted. 
Maybe we just eat the possibility of storing dead clients.

ntnn: Or we ttlcache it :D but that'd be a follow up

mjudeikis:
```
100k churned workspaces ≈ ~6MB,
```
good luck doing this, for 6MB you will need 90h of create/delete :D I think this is fine :)

ntnn (approved):
/lgtm
/approve

under duress and protest :D

ntnn:
/test pull-kcp-test-e2e

seems like ci infra failure?

PR #4146, opened 2026-05-22
bump golang.org/x dependencies

xrstf:
## Summary
The Go team has published a new set of x dependencies today that fix a few CVEs. From a cursory look it doesn't seem like we are affected, but I wrote my `go-bump-by-prefix` script and gosh darnit, I will use it in occasions like this.

## What Type of PR Is This?
/kind chore

## Release Notes
```release-note
NONE
```


ntnn (approved):
/lgtm
/approve

Issue #4041, opened 2026-04-17
feature: Extend metrics on usage data


SimonTheLeg:
We want to extend our metrics to cover the following. Also we should generally avoid using the "_count" suffix going forward, as apparently this is not a good practice in prometheus, but for workspaces, I could not come up with anything more creative.

### Metrics to Add:

* `kcp_workspace_count` number of workspaces on a given shard. Should be grouped by their phase (similiar to kcp_logicalcluster_count metric). This metric is important for us to use in conjunction with the  kcp_logicalcluster_count metric since workspaces can be scheduled on different shards than their logicalcluster

* `kcp_apibinding_phase`: Gauge tracking the current count of APIBindings by phase (Unbound, Binding, Bound). Labels: `phase`. Note: empty phase ("") should be reported as "Unbound". Helps identify distribution of bindings across lifecycle stages.

* `kcp_apibinding_condition_status`: Gauge tracking the count of APIBindings with each condition status. Labels: `type` (Ready, APIExportValid, PermissionClaimsValid, InitialResourcesBound), `status` (True, False, Unknown). Primary metric for monitoring overall binding health.

* `kcp_apibinding_ready_duration_ms`: Histogram measuring time from creation to Ready=True condition across all APIBindings in milliseconds. Tracks how long bindings typically take to become fully operational.

- `kcp_apiexport_condition_status`: Gauge tracking the count of APIExports with each condition status. (True, False, Unknown). Primary metric for monitoring overall export health.

- `kcp_apiexport_ready_duration_ms`: Histogram measuring time from creation to Ready=True condition across all APIExports in milliseconds. Tracks how long exports typically take to become fully operational.

### General Notes:
- All metrics aggregate across all APIBindings/Exports in the kcp installation
- No per-APIBinding/Exports labels (name, workspace, etc.) to avoid cardinality explosion

Elbehery:
i am interested to work on this 👍🏽

mirzakopic:
@SimonTheLeg can you please assign this to @Elbehery

ntnn:
@Elbehery you can just write `/assign` - the command is available for everyone

SimonTheLeg:
I think this task is done and can be closed, right @Elbehery?

Elbehery:
Yes 👍🏽

Elbehery:
@SimonTheLeg  I am not authorized to close :) 

Would u kindly close this as completed!

PR #4147, opened 2026-05-22
Add metrics for APIBinding and APIExport usage data

Elbehery:
Implements the metrics proposed in #4041. Adds five new Prometheus metrics to improve
  observability of `APIBinding` and `APIExport` lifecycle and health, following the same
  patterns already used for `kcp_workspace_count` and `kcp_logicalcluster_count`.

  All metrics are system-wide aggregates with no per-resource labels to avoid cardinality
  explosion.

  ### New metrics

  | Metric | Type | Labels | Description |
  |---|---|---|---|
  | `kcp_apibinding_phase` | Gauge | `phase` | Number of APIBindings in each phase (`Binding`, `Bound`, or empty for newly created) |
  | `kcp_apibinding_condition_status` | Gauge | `condition`, `status` | Number of APIBindings with each condition type and status (`True`/`False`/`Unknown`) |
  | `kcp_apibinding_ready_duration_ms` | Histogram | — | Milliseconds from APIBinding creation to reaching the `Bound` phase |
  | `kcp_apiexport_condition_status` | Gauge | `condition`, `status` | Number of APIExports with each condition type and status |
  | `kcp_apiexport_ready_duration_ms` | Histogram | — | Milliseconds from APIExport creation to becoming fully operational (`IdentityValid=True` and `VirtualWorkspaceURLsReady=True`) |

  ### Implementation

  - Phase and condition state is tracked per resource key in a `sync.Mutex`-protected map
    in each controller; on update only the diff is applied to the gauges.
  - Histogram observations are idempotent — each resource contributes exactly one sample
    (guarded by a phase-transition check for bindings, a `readyAPIExports` set for exports)
    so controller restarts do not double-count.
  - Each metric is introduced in its own commit with unit tests covering add, update,
    delete, and concurrency cases.

cc @mjudeikis @ntnn @SimonTheLeg 

/kind feature

```release-note
Add metrics for APIExport and APIBinding usage
```

### DIFF
--- a/pkg/reconciler/apis/apibinding/apibinding_controller.go
+++ b/pkg/reconciler/apis/apibinding/apibinding_controller.go
@@ -628,6 +628,9 @@ func (c *controller) handlePhaseMetricsOnUpdate(oldBinding, newBinding *apisv1al
 		if newPhase != "" {
 			kcpmetrics.IncrementAPIBindingPhase(newPhase)
 		}
+		if newPhase == string(apisv1alpha2.APIBindingPhaseBound) {
+			kcpmetrics.ObserveAPIBindingReadyDuration(newBinding.CreationTimestamp.Time)
+		}
 		c.countedAPIBindings[key] = newPhase
 	}
 }
--- a/pkg/reconciler/apis/apibinding/apibinding_controller.go
+++ b/pkg/reconciler/apis/apibinding/apibinding_controller.go
@@ -198,7 +198,8 @@ func NewController(
 			kcpClusterClient.ApisV1alpha2().APIBindings(),
 			ControllerName,
 		),
-		countedAPIBindings: make(map[string]string),
+		countedAPIBindings:          make(map[string]string),
+		countedAPIBindingConditions: make(map[string]map[string]string),
 	}
 
 	logger := logging.WithReconciler(klog.Background(), ControllerName)
@@ -209,16 +210,20 @@ func NewController(
 			binding := tombstone.Obj[*apisv1alpha2.APIBinding](obj)
 			c.enqueueAPIBinding(binding, logger, "")
 			c.handlePhaseMetricsOnAdd(binding)
+			c.handleConditionMetricsOnAdd(binding)
 		},
 		UpdateFunc: func(oldObj, obj interface{}) {
 			binding := tombstone.Obj[*apisv1alpha2.APIBinding](obj)
+			old := tombstone.Obj[*apisv1alpha2.APIBinding](oldObj)
 			c.enqueueAPIBinding(binding, logger, "")
-			c.handlePhaseMetricsOnUpdate(tombstone.Obj[*apisv1alpha2.APIBinding](oldObj), binding)
+			c.handlePhaseMetricsOnUpdate(old, binding)
+			c.handleConditionMetricsOnUpdate(old, binding)
 		},
 		DeleteFunc: func(obj interface{}) {
 			binding := tombstone.Obj[*apisv1alpha2.APIBinding](obj)
 			c.enqueueAPIBinding(binding, logger, "")
 			c.handlePhaseMetricsOnDelete(binding)
+			c.handleConditionMetricsOnDelete(binding)
 		},
 	})
 
@@ -366,6 +371,8 @@ type controller struct {
 
 	mu                 sync.Mutex
 	countedAPIBindings map[string]string
+	// countedAPIBindingConditions maps binding key -> (conditionType -> status)
+	countedAPIBindingConditions map[string]map[string]string
 }
 
 // enqueueAPIBinding enqueues an APIBinding .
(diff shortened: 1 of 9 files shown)


ntnn (commented):
/ok-to-test

pkg/reconciler/apis/apibinding/apibinding_phase_metrics_test.go
```diff
@@ -0,0 +1,131 @@
...
+		countedAPIBindings:          make(map[string]string),
+		countedAPIBindingConditions: make(map[string]map[string]string),
+	}
+}
+
+func binding(cluster, name string, phase apisv1alpha2.APIBindingPhaseType) *apisv1alpha2.APIBinding {
+	return &apisv1alpha2.APIBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        name,
+			Annotations: map[string]string{"kcp.io/cluster": cluster},
```
ntnn: There's a constant for this in the logicalcluster package: https://github.com/kcp-dev/logicalcluster/blob/7f4518bf96ac7636222bad89f8ffce473fb923df/name.go#L71

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -133,17 +135,28 @@ func NewController(
...
-			c.enqueueAPIExport(newObj.(*apisv1alpha2.APIExport))
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			export := newObj.(*apisv1alpha2.APIExport)
+			c.enqueueAPIExport(export)
+			c.handleConditionMetricsOnUpdate(oldObj.(*apisv1alpha2.APIExport), export)
+			c.handleReadyDurationMetricOnUpdate(export)
 		},
 		DeleteFunc: func(obj interface{}) {
-			c.enqueueAPIExport(obj.(*apisv1alpha2.APIExport))
+			export := obj.(*apisv1alpha2.APIExport)
```
ntnn: This should use the tombstone package; we aren't yet consistent for this across the codebase but anywhere a deleted object is type asserted should use that: https://github.com/kcp-dev/kcp/blob/9f1202fccb565a6c08529b749cae58c48294bd55/pkg/tombstone/tombstone.go#L26-L39

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -381,3 +398,122 @@ func InstallIndexers(apiExportInformer apisv1alpha2informers.APIExportClusterInf
...
+	if err != nil {
+		return
+	}
+
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if _, alreadyObserved := c.readyAPIExports[key]; alreadyObserved {
+		return
+	}
```
ntnn, line 527: This needs a hook on OnAdd I think? On a controller restart the existing APIExports will be replayed through the Add handler and on the next resync this map is missing the preexisting APIExports, so they will be observed again.

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -381,3 +398,122 @@ func InstallIndexers(apiExportInformer apisv1alpha2informers.APIExportClusterInf
...
+
+func isAPIExportReady(export *apisv1alpha2.APIExport) bool {
+	identityValid := false
+	urlsReady := false
+	for _, cond := range export.Status.Conditions {
+		switch cond.Type {
+		case apisv1alpha2.APIExportIdentityValid:
+			identityValid = string(cond.Status) == "True"
+		case apisv1alpha2.APIExportVirtualWorkspaceURLsReady:
+			urlsReady = string(cond.Status) == "True"
```
ntnn:
```suggestion
		case apisv1alpha2.APIExportIdentityValid:
			identityValid = cond.Status == corev1.ConditionTrue
		case apisv1alpha2.APIExportVirtualWorkspaceURLsReady:
			urlsReady = cond.Status == corev1.ConditionTrue
```

pkg/reconciler/apis/apibinding/apibinding_phase_metrics_test.go
Elbehery: fixed 👍🏽

pkg/reconciler/apis/apiexport/apiexport_controller.go
Elbehery: fixed 👍🏽

pkg/reconciler/apis/apiexport/apiexport_controller.go
Elbehery: added `c.handleReadyDurationMetricOnUpdate(export)`

pkg/reconciler/apis/apiexport/apiexport_controller.go
Elbehery: fixed 👍🏽 .. actually apologies I did not use the constant initially :)

Elbehery:
failures are due to race during job tear-down 

```
```

I think we could retry one more time, otherwise `/override` to the rescue 🥳

pkg/reconciler/apis/apibinding/apibinding_controller.go
```diff
@@ -353,6 +368,11 @@ type controller struct {
 
 	deletedCRDTracker *lockedStringSet
 	commit            CommitFunc
+
+	mu                 sync.Mutex
```
ntnn: This lock is protecting both countedAPIBindings and countedAPIBindingConditions?
We don't have another lock but since it only protects the metric maps I'd name it `countedAPIBindingsLock` and clarify that it protects both variables

pkg/reconciler/apis/apiexport/apiexport_controller.go
```diff
@@ -214,6 +229,10 @@ type controller struct {
 	listShards func() ([]*corev1alpha1.Shard, error)
 
 	commit CommitFunc
+
+	mu                         sync.Mutex
+	countedAPIExportConditions map[string]map[string]string
+	readyAPIExports            map[string]struct{}
```
ntnn, line 239: Similar here, the lock should have a descriptive name and a comment to show which variables it protects.

ntnn: Could just call it `metricsLock` or `metricsMu`.

ntnn (commented):
Can't do a full review because github shows me inconsistent stages depending on where I look and when I refresh :D Someimtes it's what I think was the last revision, sometimes the new revision. I'll check tomorrow^^

Elbehery:
> Can't do a full review because github shows me inconsistent stages depending on where I look and when I refresh :D Someimtes it's what I think was the last revision, sometimes the new revision. I'll check tomorrow^^

Schönen Feierabend und Bis Morgen :)

pkg/reconciler/apis/apiexport/apiexport_controller.go
ntnn: I think we need a copy of `...OnUpdate` for the `OnAdd` handle. On restarts the duration of already ready APIExports would be recorded again. So the `.OnAdd` version should just observe the names of already ready APIExports.

pkg/server/metrics/metrics.go
```diff
@@ -39,11 +41,61 @@ var (
...
+	)
+
+	apiExportReadyDurationMs = metrics.NewHistogram(
+		&metrics.HistogramOpts{
+			Name:           "kcp_apiexport_ready_duration_ms",
+			Help:           "Duration in milliseconds from APIExport creation to becoming fully operational (IdentityValid and VirtualWorkspaceURLsReady both True).",
+			StabilityLevel: metrics.ALPHA,
+			Buckets:        []float64{100, 500, 1000, 2500, 5000, 10000, 30000, 60000, 120000, 300000},
+		},
+	)
```
ntnn, line 90: These are missing the shard label

pkg/reconciler/apis/apibinding/apibinding_controller.go
Elbehery: fixed 👍🏽

pkg/reconciler/apis/apiexport/apiexport_controller.go
Elbehery: fixed 👍🏽

pkg/server/metrics/metrics.go
Elbehery: fixed 👍🏽

ntnn (approved):
/lgtm
/approve

Thanks!

PR #4148, opened 2026-05-25
Add watchcache ring ticker to unblock cacher shrink

mjudeikis:
## Summary

Upstream's watchCache only calls resizeCacheLocked from updateCache(event) and gates the shrink path on isCacheFullLocked(). During idle periods after a workload burst, no events flow, so inflated rings never shrink and the shard's memory footprint stays elevated until restart. See `https://github.com/kubernetes/kubernetes/issues/139250`.

Install a kicker that, every 80s (just above the upstream eventFreshDuration of 75s), patches a per-resource "kick object" in the root cluster with a tick timestamp. Each patch produces a distinct etcd revision and therefore one watch event per target resource, giving resizeCacheLocked a chance to fire its shrink branch.

Cost is ~7 metadata patches per shard per 80s. The annotation key is fixed so per-object growth is bounded, and the field manager "kcp-watchcache-kicker" makes the patches trivial to filter in audit logs. Missing target objects (e.g. no APIBindings in a freshly bootstrapped root) are skipped silently.

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Add ticker to trigger cache cleanups
```

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -493,6 +493,7 @@ func (s *Server) Run(ctx context.Context) error {
 		go s.CacheKcpSharedInformerFactory.Apis().V1alpha2().APIExports().Informer().Run(hookContext.Done())
 		go s.CacheKcpSharedInformerFactory.Cache().V1alpha1().CachedResources().Informer().Run(hookContext.Done())
 		go s.CacheKcpSharedInformerFactory.Cache().V1alpha1().CachedResourceEndpointSlices().Informer().Run(hookContext.Done())
+		installWatchCacheKicker(hookCtx, s.KubeClusterClient, s.KcpClusterClient)
 		go s.KcpSharedInformerFactory.Core().V1alpha1().LogicalClusters().Informer().Run(hookContext.Done())
 		go s.KcpSharedInformerFactory.Cache().V1alpha1().CachedResources().Informer().Run(hookContext.Done())
 		go s.KcpSharedInformerFactory.Cache().V1alpha1().CachedResourceEndpointSlices().Informer().Run(hookContext.Done())
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -493,7 +493,6 @@ func (s *Server) Run(ctx context.Context) error {
 		go s.CacheKcpSharedInformerFactory.Apis().V1alpha2().APIExports().Informer().Run(hookContext.Done())
 		go s.CacheKcpSharedInformerFactory.Cache().V1alpha1().CachedResources().Informer().Run(hookContext.Done())
 		go s.CacheKcpSharedInformerFactory.Cache().V1alpha1().CachedResourceEndpointSlices().Informer().Run(hookContext.Done())
-		installWatchCacheKicker(hookCtx, s.KubeClusterClient, s.KcpClusterClient)
 		go s.KcpSharedInformerFactory.Core().V1alpha1().LogicalClusters().Informer().Run(hookContext.Done())
 		go s.KcpSharedInformerFactory.Cache().V1alpha1().CachedResources().Informer().Run(hookContext.Done())
 		go s.KcpSharedInformerFactory.Cache().V1alpha1().CachedResourceEndpointSlices().Informer().Run(hookContext.Done())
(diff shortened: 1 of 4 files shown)


pkg/server/watchcache_kicker.go
```diff
@@ -0,0 +1,178 @@
...
+	// eventFreshDuration (75 s). Ticking at this cadence ensures each kicked
+	// event finds the recent-quarter boundary in any inflated watchCache ring
+	// older than the fresh window — letting upstream's resizeCacheLocked
+	// shrink branch fire.
+	watchCacheKickerInterval = 80 * time.Second
+
+	// watchCacheKickerAnnotation carries a per-tick timestamp on each kicked
+	// object so each patch produces a distinct etcd revision (and therefore
+	// a watch event flowing through the cacher).
+	watchCacheKickerAnnotation = "kcp.io/watchcache-kicker-tick"
```
ntnn: `internal.kcp.io`?

pkg/server/watchcache_kicker.go
```diff
@@ -0,0 +1,178 @@
...
+func kickAll(
+	ctx context.Context,
+	logger klog.Logger,
+	kubeClient kcpkubernetesclientset.ClusterInterface,
+	kcpClient kcpclientset.ClusterInterface,
+) {
+	ts := time.Now().UTC().Format(time.RFC3339Nano)
+	patch := fmt.Appendf(nil, `{"metadata":{"annotations":{%q:%q}}}`, watchCacheKickerAnnotation, ts)
+	opts := metav1.PatchOptions{FieldManager: watchCacheKickerFieldManager}
+	root := core.RootCluster.Path()
```
ntnn: Should this happen in a system workspace so it happens on each shard? If this only happens in the root workspace only the root shard is kicked?

mjudeikis: good point. I think this needs to target system workspace, yes. This is per shard

pkg/server/server.go
```diff
@@ -493,6 +493,7 @@ func (s *Server) Run(ctx context.Context) error {
 		go s.CacheKcpSharedInformerFactory.Apis().V1alpha2().APIExports().Informer().Run(hookContext.Done())
 		go s.CacheKcpSharedInformerFactory.Cache().V1alpha1().CachedResources().Informer().Run(hookContext.Done())
 		go s.CacheKcpSharedInformerFactory.Cache().V1alpha1().CachedResourceEndpointSlices().Informer().Run(hookContext.Done())
+		installWatchCacheKicker(hookCtx, s.KubeClusterClient, s.KcpClusterClient)
```
ntnn: This seems a bit out of place here - can we move that below the block starting the informers? 
Purely for aesthetic reasons.

pkg/server/watchcache_kicker.go
```diff
@@ -0,0 +1,176 @@
...
+			fn: func() error {
+				list, err := kcpClient.Cluster(shard).ApisV1alpha2().APIBindings().List(ctx, metav1.ListOptions{Limit: 1})
+				if err != nil || len(list.Items) == 0 {
+					return err
+				}
+				_, err = kcpClient.Cluster(shard).ApisV1alpha2().APIBindings().
+					Patch(ctx, list.Items[0].Name, types.MergePatchType, patch, opts)
+				return err
+			},
+		},
```
ntnn, line 163: Missing CR/CRES? Or does APIBindings trigger CR/CRES as well?

pkg/server/server.go
ntnn: I think this should even be under the wait below to ensure the informers are up. Otherwise the kicker starts up and could log errors due to informers not running?

pkg/server/watchcache_kicker.go
```diff
@@ -0,0 +1,176 @@
...
+	configshard "github.com/kcp-dev/kcp/config/shard"
+)
+
+const (
+	// watchCacheKickerInterval is just above the upstream apiserver's default
+	// eventFreshDuration (75 s). Ticking at this cadence ensures each kicked
+	// event finds the recent-quarter boundary in any inflated watchCache ring
+	// older than the fresh window — letting upstream's resizeCacheLocked
+	// shrink branch fire.
+	watchCacheKickerInterval = 80 * time.Second
```
ntnn, line 43: We don't have a way to only trigger this if the watch cache has been bursted, or?

mjudeikis: we could - but too much brittle logic. Not worth the effort

pkg/server/watchcache_kicker.go
mjudeikis: Interntional For now its only hotpaths :/

ntnn: Then we should update the comment accordingly or it might trip us.

ntnn:
The more I look at this the more I think this should be a fix in our kube fork instead.

mjudeikis:
> The more I look at this the more I think this should be a fix in our kube fork instead.

In reallity I dont think we need todo anything at all. In normal environments, there will always be something happening in terms of events. So we can just leave it "no fix"

PR #4149, opened 2026-05-25
fix: WorkspaceType admission rejecting forward-referenced APIExports

### LINKED ISSUE #4145: bug: workspace type creation with referenced non existing ApiExport
### Describe the bug

Before (until version 0.31.2) it was possible to create a WorkspaceType which references non existing yet ApiExport in defaultApiBindings section. With 0.31.2 it's not possible even using admin kubeconfig. It might be not a bug as it makes sense to do not create WorkspaceType which references something non existing

### Steps To Reproduce

1. Run sharded test server https://github.com/kcp-dev/kcp/tree/main/cmd/sharded-test-server
2. Using admin kubeconfig try to create workspace type like this and you will get the next error 
```
kind: WorkspaceType
metadata:
  name: provider
spec:
  defaultAPIBindings:
    - export: core.platform-mesh.io
      path: root:platform-mesh-system
```
Error:
`Error from server (Forbidden): error when creating "workspace-type-provider.yaml": workspacetypes.tenancy.kcp.io "provider" is forbidden: unable to create or update WorkspaceType: no permission to bind to export root:platform-mesh-system:core.platform-mesh.io`

If you create platform-mesh-system workspace, you will still face the same issue 

**Also important thing. If you create ApiExport and than reference it in WorkspaceType, the issue will be gone**

### Expected Behaviour

As it was before, WorkspaceType is created without issues

### Additional Context

_No response_

pen-pal:
## Summary

Fixes a regression in 0.31.2 where creating a `WorkspaceType` referencing an APIExport that doesn't exist yet is rejected with `no permission to bind to export ...`, even for admins.

The bind-permission check (added in #4082) treats a NotFound on the APIExport lookup the same as a permission denial. This breaks the common flow of creating a `WorkspaceType` before its APIExports.

Now: if the APIExport isn't found, skip the SAR for that entry. The check still runs whenever the APIExport exists.

## What Type of PR Is This?

/kind bug
/kind regression

## Related Issue(s)

Fixes #4145

## Release Notes

```release-note
Fix a regression in WorkspaceType admission that rejected creates referencing an APIExport that did not yet exist, even for admins. Such forward references are allowed again; the bind-permission check still applies when the APIExport exists.
```

### DIFF
--- a/pkg/admission/workspacetype/admission.go
+++ b/pkg/admission/workspacetype/admission.go
@@ -208,6 +208,15 @@ func (o *workspacetype) checkDefaultAPIBindingsPermissions(ctx context.Context, 
 			path := logicalcluster.NewPath(exportPath)
 			export, err := o.getAPIExport(path, exportName)
 			if err != nil {
+				if apierrors.IsNotFound(err) {
+					// APIExport does not exist (yet). The default-apibinding-
+					// controller will only create a binding once it appears,
+					// and without it we have no cluster to authorize against.
+					// Skipping the entry preserves the declarative ordering
+					// use case (WorkspaceType created ahead of its exports)
+					// that pre-dated this admission check.
+					continue
+				}
 				return forbidden
 			}
 			exportClusterName = logicalcluster.From(export)
--- a/pkg/admission/workspacetype/admission_test.go
+++ b/pkg/admission/workspacetype/admission_test.go
@@ -208,13 +208,15 @@ func TestValidate_DefaultAPIBindings(t *testing.T) {
 			decision: authorizer.DecisionAllow,
 		},
 		{
-			name: "create denied when referenced APIExport cannot be resolved",
+			name: "create allowed when referenced APIExport cannot be resolved",
 			op:   admission.Create,
 			newWT: newWorkspaceType(tenantCluster, "z",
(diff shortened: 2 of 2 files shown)


ntnn:
Hi @pen-pal! Thanks for the contribution!

However I don't think it's a regression - that this was possible previously is purely incidental and wasn't a planned feature (and as far as I can see never documented).
On the contrary - allowing adding APIExports to a WST that do not exist yet and just skip the authz check for them is a security incident in waiting.

We explicitly modeled the WST to allow users to add APIExports they have access to, which then implicitly grants users who can instantiate workspaces of this type the ability to bind these APIExports.
With this change a malicious actor could create a WST with predicted APIExports, wait for them to be created by their target and then bind them (since the WST skips the authz check on the APIExport, as this is guaranteed during the modification of the WST) and can bypass authz on the APIExport that way.
That is basically adding back part of the security issue that the previous PR #4082 closed.

I'd expect users to be able to wait until an APIExport has propagated through the cache server and for CI/CD tools to retry to apply manifests, so they WST creation/update should proceed as soon as the referenced APIExport has propagated.

/hold

mjudeikis:
this will not be fixed as proposed here. we fixed a security issue so its not a regression as such. sorry

PR #4152, opened 2026-05-26
fix(workspacetype): resolve defaultAPIBindings path via LogicalCluster

### LINKED ISSUE #4145: bug: workspace type creation with referenced non existing ApiExport

mjudeikis:
## Summary

The bind-permission check added in 51246a32a ("Add aditional validation for WST") rejected creation of WorkspaceTypes that referenced an APIExport which was not visible to the admitting shard:

```
  workspacetypes.tenancy.kcp.io "X" is forbidden: unable to create or
  update WorkspaceType: no permission to bind to export <path>
```

The lookup served two purposes — checking existence and resolving exportPath into a cluster name for the bind SAR — but it conflated "export is not visible" with "user cannot bind". That regressed two real workflows:

  1. Forward references (#4145). Until 0.31.2 users could create a WorkspaceType whose defaultAPIBindings pointed at an APIExport that did not exist yet. The new check rejected this with a misleading permission error even when admin-issued RBAC granted bind on the named (future) APIExport — and RBAC's resourceNames does match resources that do not exist yet.

  2. Sharded clusters (e2e TestWorkspaceTypesAPIBindingInitialization). When the WorkspaceType is admitted on shard A and the APIExport was just created on shard B, the cache-replicated APIExport may not be visible to shard A's informer yet. Admission failed non-deterministically with the same forbidden error.

Resolve exportPath via the LogicalCluster of the referenced workspace instead of the APIExport object. LogicalClusters are also cache- replicated but are typically established earlier (the workspace must exist for the test/user to reach this code path), and the resolution no longer depends on the APIExport itself existing. The bind SAR is then issued against the resolved cluster with the user's identity, as before, so the permission gate is preserved end-to-end:

```
  - workspace at exportPath not visible    -> forbidden (deterministic
                                              message names the path)
  - workspace visible, user lacks bind     -> forbidden (unchanged)
  - workspace visible, user has bind, but
    APIExport does not exist yet           -> admitted; the default-
                                              apibinding controller
                                              will bind it later
```

Belt-and-suspenders: wrap the parent1 create in the failing e2e in `kcptestinghelpers.Eventually` that retries only on the deterministic "not found or not yet visible" message — any other admission error still surfaces immediately, so a real regression is not masked.

Fixes #4145.

## What Type of PR Is This?

/kind bug

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/admission/workspacetype/admission.go
+++ b/pkg/admission/workspacetype/admission.go
@@ -192,7 +192,10 @@ func (o *workspacetype) checkDefaultAPIBindingsPermissions(ctx context.Context, 
 		exportPath := ref.Path
 		exportName := ref.Export
 
-		// unified forbidden error that does not leak workspace existence
+		// Single forbidden response for both "cannot resolve the workspace"
+		// and "user lacks bind on the export". Distinguishing the two would
+		// let a caller with WorkspaceType create/update permission probe
+		// for the existence of arbitrary workspaces by reading the error.
 		forbidden := admission.NewForbidden(a, fmt.Errorf("unable to create or update WorkspaceType: no permission to bind to export %s",
 			logicalcluster.NewPath(exportPath).Join(exportName).String()))
 
@@ -213,12 +216,7 @@ func (o *workspacetype) checkDefaultAPIBindingsPermissions(ctx context.Context, 
 			path := logicalcluster.NewPath(exportPath)
 			lc, err := o.getLogicalCluster(path)
 			if err != nil {
-				// Distinguish "workspace not visible" from the bind denial
-				// below: the user might have permission once the workspace
-				// is reachable, but right now we cannot evaluate it.
-				return admission.NewForbidden(a, fmt.Errorf(
-					"unable to create or update WorkspaceType: workspace %q referenced in spec.defaultAPIBindings[].path is not found or not yet visible",
-					exportPath))
+				return forbidden
 			}
 			exportClusterName = logicalcluster.From(lc)
 		}
--- a/pkg/admission/workspacetype/admission.go
+++ b/pkg/admission/workspacetype/admission.go
@@ -32,8 +32,8 @@ import (
 
 	kcpkubernetesclientset "github.com/kcp-dev/client-go/kubernetes"
 	"github.com/kcp-dev/logicalcluster/v3"
-	apisv1alpha2 "github.com/kcp-dev/sdk/apis/apis/v1alpha2"
 	"github.com/kcp-dev/sdk/apis/core"
+	corev1alpha1 "github.com/kcp-dev/sdk/apis/core/v1alpha1"
 	tenancyv1alpha1 "github.com/kcp-dev/sdk/apis/tenancy/v1alpha1"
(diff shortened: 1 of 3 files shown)


pkg/admission/workspacetype/admission.go
```diff
@@ -206,11 +211,16 @@ func (o *workspacetype) checkDefaultAPIBindingsPermissions(ctx context.Context,
...
+			lc, err := o.getLogicalCluster(path)
 			if err != nil {
-				return forbidden
+				// Distinguish "workspace not visible" from the bind denial
+				// below: the user might have permission once the workspace
+				// is reachable, but right now we cannot evaluate it.
+				return admission.NewForbidden(a, fmt.Errorf(
+					"unable to create or update WorkspaceType: workspace %q referenced in spec.defaultAPIBindings[].path is not found or not yet visible",
+					exportPath))
 			}
```
ntnn: This woud leak lc/workspace existence, or?
This would allow a malicious actor with permissions to create/update a WST to "port knock" for workspaces.

mjudeikis: Good catch — you're right, this leaked LC/workspace existence to anyone holding WorkspaceType create/update.

Fixed in 103b60992: the LC-not-found branch now returns the same `forbidden` value the bind-denial branch returns, so the two cases are byte-identical to the caller. The error names the `path:export` they supplied themselves, so nothing new is revealed.

Knock-on: the e2e retry in `TestWorkspaceTypesAPIBindingInitialization` is now keyed off the unified `"no permission to bind to export"` message instead of the deterministic-not-visible one. Because that message is also a real bind denial, a sustained perm regression no longer surfaces immediately — it surfaces when `Eventually` deadlines (≈30s). I think that's the right trade here; the unit test in `admission_test.go` still covers the bind-denied path directly. PTAL.

ntnn (approved):
/lgtm
/approve

PR #4153, opened 2026-05-26
fix(virtual): shorten DenyCacheTTL for lifecycle VW authorizers

mjudeikis:
## Summary

The initializing and terminating VW content proxies authorize every incoming request by issuing a `SubjectAccessReview` for the "initialize" (or "terminate") verb on the underlying WorkspaceType. That SAR is served by a `delegated.NewCachingAuthorizer`, which underneath wraps Kubernetes' webhook authorizer and caches *decisions* in an LRU keyed by (user, attributes). The default DenyCacheTTL inherited from delegated.Options is 30 seconds.

That TTL is too long for these VWs, which are explicitly designed around just-in-time RBAC grants:

  1. An operator (or test) grants user-U the "initialize" verb on workspacetypes/X.
  2. user-U immediately makes a request through the VW.
  3. The VW's SAR races the shard's RBAC informer. If the ClusterRoleBinding has not yet been observed, the SAR returns Deny.
  4. That Deny is cached for 30s. Every subsequent request from user-U within that window — including the test's own EventuallyWithT poll loop — hits the cached Deny without ever re-issuing the SAR. By the time the cache expires, the test has also exhausted wait.ForeverTestTimeout (which is also 30s) and fails with "Condition never satisfied".

Manifested as a flake in `TestInitializingWorkspacesVirtualWorkspace`- `InitializerPermissions` on the sharded e2e job, where the cross-shard RBAC propagation window is wide enough that the first SAR sometimes loses the race.

Shorten DenyCacheTTL to 5s on both lifecycle VW authorizers. Long enough to absorb a brief denial loop without hammering the SAR endpoint; short enough that a fresh "initialize"/"terminate" grant becomes effective within one or two poll intervals of typical test/operator workflows. AllowCacheTTL is left at the 5-minute default — revocation responsiveness in a still-initializing/ terminating workspace is acceptable; the workspace itself is short-lived.

## What Type of PR Is This?
/kind bug
/kind failing-test
/kind flake

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/virtual/initializingworkspaces/builder/build.go
+++ b/pkg/virtual/initializingworkspaces/builder/build.go
@@ -24,6 +24,7 @@ import (
 	"net/url"
 	"path"
 	"strings"
+	"time"
 
 	authenticationv1 "k8s.io/api/authentication/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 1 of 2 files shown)


ntnn (approved):
/lgtm
/approve

ntnn:

There's no actual failure in the log?

mjudeikis:
/retest 
looks strange

PR #4158, opened 2026-05-27
[WIP] Add tests verifying stable hashing/encoding of stored values

ntnn:
## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

Issue #3498, opened 2025-07-29
feature: migration of workspaces between shards


embik:
### Feature Description

For a growing kcp instance, migrating logical clusters between shards might eventually become necessary, especially when single shard environments evolve to multi-shard ones. As such, it should be possible to migrate workspaces from one shard to another, either online or at the very least offline.

### Proposed Solution

Unclear right now. Some mechanism to put logical clusters into maintenance, fetch all data stored in the underlying data store, send it to the other shard.

### Want to contribute?

- [ ] I would like to work on this issue.

mjudeikis:
We been chatting about this. There should be some ways to achieve this using etcd client and maybe dedicated management vw. 
1. Shards put to maintenance mode (read only?)
2. Workspace move iniciated via VW
3. Update frontproxy index (how, maybe a simple restart of both shards and fp for now)
4. Shards taken out of maintenance (restart?)

Basically moving from one shard to another. I propose we start with simpliest approach:
1. Shards maintenance mode (Read only)
2. dedicated VW which has access to both shards.
3. Restart all the things to commit. 


Current snip of shard etcd keys:
```
/registry/core/secrets/root/kcp-system/topology.kcp.io
/registry/core/serviceaccounts/2zvtcam3cl4vycy3/default/default
/registry/core/serviceaccounts/2zvtcam3cl4vycy3/kube-system/default
/registry/core/serviceaccounts/30pb2119n7xgh3kq/default/default
/registry/core/serviceaccounts/30pb2119n7xgh3kq/kube-system/default
/registry/core/serviceaccounts/root/default/default
/registry/core/serviceaccounts/root/kcp-system/default
/registry/core/serviceaccounts/system:admin/default/default
/registry/core/serviceaccounts/system:admin/kube-public/default
/registry/core/serviceaccounts/system:admin/kube-system/default
/registry/core/serviceaccounts/system:shard/default/default
/registry/rbac.authorization.k8s.io/clusterrolebindings/2zvtcam3cl4vycy3/workspace-admin
/registry/rbac.authorization.k8s.io/clusterrolebindings/30pb2119n7xgh3kq/workspace-admin
/registry/rbac.authorization.k8s.io/clusterrolebindings/root/system:kcp:authenticated:apiexport:kcp-admin:maximal-permission-policy
/registry/rbac.authorization.k8s.io/clusterrolebindings/root/system:kcp:authenticated:apiexport:tenancy:bind
[...]

PR #4159, opened 2026-05-27
Add base36/63-sha224 helper

ntnn:
## Summary

This is adjacent to #3498.

TL;DR:

1. Moving `pkg/crypto` to `apimachinery/.../pkg/util/crypto`
  Why? To have one `kcpcrypto` 
  Why there? Because the helpers added are used in staging repos as well and this mirrors upstreams apimachinery `pkg/util/<pkg>`
2. Adding `baseX{,sha224}` helpers to `kcpcrypto`
  Why? 
    1. Because we have many different iterations of this across the codebase, some use a library, some `math/big.Int`
    2. Because for migration operators need a library function to do the same hashing/encoding of shard names we do
3. Update all hashing/encoding places to use kcpcrypto
  Why? For consistency.

Prose:

Shards names are stored hashed on the Workspace in an annotation `internal.tenancy.kcp.io/shard`, presumably to not intentionally leak shard names to users.
For migrations we need to provide a library function for operators utilizing the migration primitive to check which shard an LC currently is in and which shard to migrate to.
We currently have many different places where base36, base36sha224 is implemented in different ways, sometimes using the base36 library, sometimes via `math/big.Int`.
Since these are all roughly the same shape I consolidated them into a tested package.
Most of the callsites are just ephemeral. 

These are not:

  - pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling.go:163 
    written to the internal.tenancy.kcp.io/shard annotation on Workspace objects
  - pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling.go:445 
    used as the LogicalCluster name (ws.Spec.Cluster); a durable workspace identifier
  - pkg/index/rewriters/user.go:39 
    HomeClusterName becomes the actual logical cluster name for the user's home workspace
  - staging/src/github.com/kcp-dev/sdk/apis/apis/v1alpha1/permissionclaims/labels.go and v1alpha2
    produces Kubernetes label keys/values written onto API resources

But all of these produce equivalent output so existing installations are unaffected, this is verified by the commit adding stable hashing/encoding unit tests. To show that the behaviour did not change this commit was PR'd separately here: https://github.com/kcp-dev/kcp/pull/4158

## What Type of PR Is This?

/kind chore

## Related Issue(s)

Fixes #

## Release Notes

```release-note
New utility crypto package in apimachinery for base36/63 related encoding
Updated existing code paths to use new apimachinery crypto package for base36/63 encoding
```


mjudeikis (approved):
/lgtm
/approve

ntnn:
/retest

unrelated:

```text
 === FAIL: test/e2e/reconciler/workspacedeletion TestWorkspaceDeletion/nested_workspace_cleanup_when_an_org_workspace_is_deleted (31.39s) 
 ```

PR #4160, opened 2026-05-27
load test allow supplying an external config

SimonTheLeg:
Little framework change to allow supplying external configs to loadtest. This makes configuring optional parameters a lot easier.

## What Type of PR Is This?

/kind chore

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/test/load/testing/params.go
+++ b/test/load/testing/params.go
@@ -1,0 +1,111 @@
+/*
+Copyright 2026 The kcp Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package testing
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/kcp-dev/sdk/apis/core"
+
+	"github.com/kcp-dev/kcp/test/load/pkg/tree"
+)
+
+var testConfig Config
+
+// Config is the top-level configuration structure unmarshalled from the
+// JSON config file passed via --config.
+type Config struct {
+	Params Params `json:"params"`
+}
+
+func defaultConfig() Config {
+	return Config{
+		Params: defaultParams(),
+	}
+}
+
+func defaultParams() Params {
+	return Params{
+		WorkspaceCount:          10000,
+		WorkspaceDepth:          5,
+		WorkspaceTreeType:       "symmetric",
+		CreateWorkspaceQPS:      8.0,
+		CRUDConfigMapQPS:        150,
+		ProviderWorkspacesCount: 1000,
+		ConsumerWorkspacesCount: 9000,
+	}
+}
+
+// Params holds all configurable load test parameters. Values can be
+// overridden by passing --config=<path-to-json> to the test binary.
+// Any field not set in the JSON file retains its default value.
+type Params struct {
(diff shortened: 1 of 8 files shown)


test/load/testing/params.go
```diff
@@ -0,0 +1,111 @@
...
+		ConsumerWorkspacesCount: 9000,
+	}
+}
+
+// Params holds all configurable load test parameters. Values can be
+// overridden by passing --config=<path-to-json> to the test binary.
+// Any field not set in the JSON file retains its default value.
+type Params struct {
+	// WorkspaceCount is the total number of workspaces to create.
+	WorkspaceCount int `json:"workspaceCount"`
```
Elbehery, line 60: Would we need `omitempty` ?

test/load/testing/params.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// WorkspaceDepth controls how deep the workspace tree is.
+	WorkspaceDepth int `json:"workspaceDepth"`
+
+	// WorkspaceTreeType selects the tree layout: "symmetric" or "flat".
```
Elbehery, line 65: as the values are consts, would using `enum` here make sense?

test/load/testing/params.go
```diff
@@ -0,0 +1,111 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/kcp-dev/sdk/apis/core"
+
+	"github.com/kcp-dev/kcp/test/load/pkg/tree"
+)
+
+var testConfig Config
```
Elbehery, line 29: I would remove this var, and return `Config` value from func 

```
func parseConfig(configFile string) (Config, error)
```

test/load/testing/params.go
SimonTheLeg: I don't think I can quite follow: What would be the advantage of having omitempty here?

We are only ever unmarshalling + we are setting defaults anyways via a func

test/load/testing/params.go
SimonTheLeg: Personally, since it is being used in a single switch statement and I don't really see much more than 2 options happening in the future either, I'd keep it as is. I mean it would not offer any protection against what matters, which is the user mispelling it in the json file. And it just requires additional harder to read boilerplate code imo.

Elbehery: lgtm 👍🏽

test/load/testing/params.go
Elbehery: ok, I am giving a review as per best practices :) 

I understand this would not be a demand here 👍🏽

test/load/testing/params.go
SimonTheLeg: I am not sure I can follow the idea: So the caller of parseConfig is the singleton TestMain. It's purpose is to create a single global config state which is used by all tests. Returning the Config from the func, would still require TestMain to make it available for the tests. So then it would need a global store again, the declaration would just be in the main_test file, which I think offers no advantage.

I am also quite sure that this pattern is the idiomatic Go way, I mean for example it's similarly used in net/http
https://github.com/golang/go/blob/master/src/net/http/main_test.go#L23

Elbehery: So I found that `testConfig` is being read only from `main_test.go` 

> It's purpose is to create a single global config state which is used by all tests

exactly, and it is being invoked from main **before** it is being used by all tests.

Therefore I did not found a reason to have a global variable :)

SimonTheLeg: > Therefore I did not found a reason to have a global variable :)

But you still would need one. If you have parseConfig return the config, then you still need to make it accessible to the tests themselves. And then we are back at needing a global variable.
All we would be doing is moving the global variable declaration from the params.go file to main_test.go file.

Elbehery: I could see now ur point, thanks for clarification. 

I thought the Config would be part of the test struct, but we don't have one here

mjudeikis (approved):
/lgtm
/approve

PR #4161, opened 2026-05-27
Set shard hash annotation on LC and propagate to Workspace

ntnn:
## Summary

Currently the shard annotation on Workspaces is set once during LC scheduling and then never reevaluated.

With incoming LC migration the shard-relevant fields must be updated depending on where the LC is scheduled.

To this end the following changes are made

1. `core.kcp.io/shard` is made part of the core API
2. `core.kcp.io/shard` is set on the LC by the LC reconciler
3. Workspace reconciler compares the LC annotation with the Workspace annotation and updates on a drift

The `internal.tenancy.kcp.io/shard` annotation is updated as well and marked deprecated.

Drawback is that when an LC isn't visible to the shard reconciling stops - which I think makes sense.

Also added the hash on the Shard status so it is getting printed when getting shards:

image

## What Type of PR Is This?

/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
LogicalClusters get the `core.kcp.io/shard` annotation with the value being the hashed name of the shard they are scheduled on
Workspaces derive the `core.kcp.io/shard` annotation from its LogicalCluster
```

### DIFF
--- a/pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling.go
+++ b/pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling.go
@@ -55,6 +55,9 @@ import (
 const (
 	// WorkspaceShardHashAnnotationKey keeps track on which shard LogicalCluster must be scheduled. The value
 	// is a base36(sha224) hash of the Shard name.
+	//
+	// Deprecated: use corev1alpha1.LogicalClusterShardAnnotationKey instead.
+	// This key is still updated and read as a fallback for backwards compatibility but will be removed in a future release.
 	WorkspaceShardHashAnnotationKey = "internal.tenancy.kcp.io/shard"
 
 	// workspaceClusterAnnotationKey keeps track of the logical cluster on the shard.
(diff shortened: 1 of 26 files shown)


ntnn:
/retest

nested_workspace_cleanup_when_an_org_workspace_is_deleted

I'm gonna take a look at that flake.

ntnn:
/retest

infra failed

ntnn:
/retest

infra failure

pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling.go
```diff
@@ -54,6 +55,9 @@ import (
 const (
 	// WorkspaceShardHashAnnotationKey keeps track on which shard LogicalCluster must be scheduled. The value
 	// is a base36(sha224) hash of the Shard name.
+	//
+	// Deprecated: use corev1alpha1.LogicalClusterShardAnnotationKey instead.
```
mjudeikis, line 59: create an issue to track this.

staging/src/github.com/kcp-dev/sdk/apis/core/v1alpha1/logicalcluster_types.go
```diff
@@ -70,6 +70,9 @@ const (
 	//
 	// Deprecated: use LogicalClusterInactiveAnnotationKey.
 	LogicalClusterInactiveAnnotationKeyLegacy = "internal.kcp.io/inactive"
+
+	// LogicalClusterShardAnnotationKey is the hashed shard name the LogicalCluster is scheduled on.
```
mjudeikis, line 74: any point why this is hased?

ntnn: I asked that myself and I didn't find anything concrete documented about this. I _presume_ that the intent was to not expose the shard names to users.
Currently it is opaque to the user where an LC is scheduled - they just connect to the FP which then proxies to the correct shard. Similarly operators don't really care because they just connect to all VWs in their APIExportES.

While I don't think it's necessary to not expose the shard name (and it's fairly easy to build a rainbow table since we don't salt the hashes) I also don't think we need to deliberately expose the shard names.

pkg/reconciler/tenancy/workspace/workspace_reconcile_scheduling.go
ntnn: https://github.com/kcp-dev/kcp/issues/4166

ntnn:
/hold

I'm tinkering with exposing the shard names to have a comparison.

ntnn:
/close

Closing, superseded. by https://github.com/kcp-dev/kcp/pull/4171

PR #4164, opened 2026-05-28
evictor wiring server->clients->cache

ntnn:
## Summary

Different solution to https://github.com/kcp-dev/kcp/pull/4138 with wiring instead of a global package state variable.
Based on that PR branch but removed the pprof and docs changes.

I think this makes more sense - `.Cluster` is the start of the lifecycle of a cluster for the client and `.Evict` is the end.
User don't have to use the `.Evict` - but anyone who looks at the API surface will immediately know that clients are not "free" and that they'll have to keep in mind that they may need to lifecycle the clusters at some point.
Additionally when a user runs into OOM issues they'll be spared the investigation _and_ they have the solution in hand.
If we hide this away in package-level state machine users will not know about this _and_ won't be able to find it if they go looking for solutions.

Diff only: https://github.com/kcp-dev/kcp/pull/4163

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/server/clientcache_evictor.go
+++ b/pkg/server/clientcache_evictor.go
@@ -17,6 +17,8 @@ limitations under the License.
 package server
 
 import (
+	"context"
+
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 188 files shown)


ntnn:
/hold

Just a quick go to showcase the approach I'd like in comparison.

PR #4165, opened 2026-05-28
Add rbac to /metrics 

### LINKED ISSUE #4062: bug: accessing shard metrics is handled on workspace level
### Problem Statement

This issue consists of two parts which are intertwined and should be handled together:

1. Currently metrics on a shard can be accessed from any level in the workspace hierarchy (see "Steps to reproduce) using permissions defined inside the workspace itself
2. There is no possibility except for `system:masters` to be able to access metrics from all shards using a single authentication directly on the `<shard-ip>:6443/metrics`, which would be the common URL used by discovering metrics queriers like prometheus

### Steps to reproduce accessing metrics using workspace level permissions

1. Let's assume you are in a workspace `:root:myws`
2. Since it is your workspace, create the following clusterrole and assign it to your user using a binding
    ```yaml
    kind: ClusterRole
    metadata:
      name: metrics-viewer
    rules:
      - nonResourceURLs:
          - /metrics
        verbs:
          - get
    ```
3. You are now able to get metrics of the whole shard using the unique path of your workspace (notice `/clusters/root:myws/metrics`)
    ```bash
    kubectl get --raw /clusters/root:myws/metrics --server="https://<shard-ip>:6443" --insecure-skip-tls-verify
    ```

### Desired final state

* It should not be possible to give yourself permissions in your workspace to access metrics of a shard using the aforementioned url-trick. These requests should return "not implemented"
* It should be possible for a kcp-admin to setup an identity*+role+binding, which allows the identity to access metrics from all shards directly using `<shard-ip>:6443/metrics` top-level path. This should be done inside the root workspace
* document the approach to setup the role+binding!

*identity referring to regular k8s identities like Users or Serviceaccounts

mjudeikis:
## Summary

Reject workspace-scoped /clusters/<ws>/metrics and cache-server /services/cache/shards/<sh>/clusters/<ws>/metrics with 501 Not Implemented. Metrics are shard-wide and have no per-workspace meaning; allowing them via a workspace ClusterRole + binding was a privilege escalation.

We might want to implement this down the line. For now - no. 

Top-level <shard>:6443/metrics is now authorized via root-workspace RBAC: a new bootstrap ClusterRole `system:kcp:metrics-reader` grants GET on `/metrics`, and a kcp-admin binds it once in `:root` for an identity of their choice. The binding is replicated to every shard via the cache server, so a single root binding covers all shards.

Cache server top-level /metrics is now reachable as well (previously, WithShardScope returned 400 for any path without a /shards/ prefix).

Fixes kcp-dev/kcp#4062

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Implement rbac for /metrics
```

### DIFF
--- a/docs/content/setup/production/metrics.md
+++ b/docs/content/setup/production/metrics.md
@@ -28,9 +28,12 @@ https://<shard>:6443/clusters/<workspace>/metrics
 https://<cache-server>:6443/services/cache/shards/<shard>/clusters/<workspace>/metrics
 ```
 
-return `501 Not Implemented`. Metrics are a shard-wide signal and have no
-per-workspace or per-shard scope; the kcp HTTP filter rejects these requests
-before authorization runs.
+return `501 Not Implemented`. Today this is a placeholder: per-workspace and
+per-shard metrics are not yet implemented, and the data the underlying
+`/metrics` handler exposes is shard-wide with no per-workspace or per-shard
+meaning. The kcp HTTP filter rejects these requests before authorization runs
+so that a future implementation can fill in real workspace-scoped metrics
+without changing the URL contract.
 
 ## Authorizing a scraper on a shard
 
--- a/docs/content/setup/production/metrics.md
+++ b/docs/content/setup/production/metrics.md
@@ -1,6 +1,6 @@
 # Scraping kcp metrics
 
-kcp exposes Prometheus metrics on the `/metrics` endpoint of every shard and of
+kcp exposes Prometheus metrics on the `/metrics` endpoint of every shard and on
 the cache server. These are **shard-wide** resources: a single scrape returns
 process-level data for the entire shard, not for any particular workspace.
 
@@ -21,26 +21,49 @@ For the cache server:
 https://<cache-server>:6443/metrics
 ```
 
-Workspace- or shard-scoped variants such as
+## Workspace- and shard-scoped variants are rejected
+
+Requests such as
 
 ```
 https://<shard>:6443/clusters/<workspace>/metrics
 https://<cache-server>:6443/services/cache/shards/<shard>/clusters/<workspace>/metrics
+https://<cache-server>:6443/services/cache/shards/<shard>/metrics
 ```
 
-return `501 Not Implemented`. Today this is a placeholder: per-workspace and
-per-shard metrics are not yet implemented, and the data the underlying
-`/metrics` handler exposes is shard-wide with no per-workspace or per-shard
-meaning. The kcp HTTP filter rejects these requests before authorization runs
-so that a future implementation can fill in real workspace-scoped metrics
-without changing the URL contract.
+return `501 Not Implemented`. The kcp HTTP filter rejects these requests before
+authorization runs, so a workspace-local `ClusterRole` with
+`nonResourceURLs: ["/metrics"]` can no longer grant access to shard-wide data.
+
+!!! warning
+    Earlier kcp releases accidentally allowed a workspace administrator to grant
+    themselves access to shard metrics by creating a `ClusterRole` with
+    `nonResourceURLs: ["/metrics"]` and a binding inside their own workspace.
+    This was a privilege escalation: the data exposed is shard-wide, not
+    workspace content. The path is now rejected at any non-root scope and the
+    only authoritative binding is one created in `:root` (see below).
+
+### What the shard / cache server `/metrics` actually returns
+
+The bare top-level `/metrics` is the standard Prometheus exposition produced by
+the underlying apiserver process. It aggregates counters and histograms across
+every workspace served by the shard (or cached by the cache server) and there
+is no per-workspace breakdown.
+
+### Future per-workspace metrics
+
+Per-workspace and per-shard metrics are not implemented today. The 501 status
+is a deliberate placeholder so that a future implementation can fill in real
+workspace-scoped metrics under the same URL contract without forcing scrapers
+to migrate paths.
 
 ## Authorizing a scraper on a shard
 
 kcp ships a bootstrap `ClusterRole` named `system:kcp:metrics-reader` that grants
 `GET` on `/metrics`. To allow an identity to scrape every shard, create a
-`ClusterRoleBinding` in the `:root` workspace. The binding is replicated to all
-shards via the cache server, so a single binding is enough.
+`ClusterRoleBinding` in the `:root` workspace. The root workspace lives on the
+root shard, and the binding is replicated to all shards via the cache server,
+so a single binding in `:root` is sufficient to cover the entire deployment.
 
 ```yaml
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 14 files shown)


docs/content/setup/production/metrics.md
```diff
@@ -0,0 +1,127 @@
+# Scraping kcp metrics
+
+kcp exposes Prometheus metrics on the `/metrics` endpoint of every shard and of
+the cache server. These are **shard-wide** resources: a single scrape returns
```
ntnn:
```suggestion
kcp exposes Prometheus metrics on the `/metrics` endpoint of every shard and on
the cache server. These are **shard-wide** resources: a single scrape returns
```

docs/content/setup/production/metrics.md
```diff
@@ -0,0 +1,127 @@
...
+https://<shard>:6443/clusters/<workspace>/metrics
+https://<cache-server>:6443/services/cache/shards/<shard>/clusters/<workspace>/metrics
+```
+
+return `501 Not Implemented`. Today this is a placeholder: per-workspace and
+per-shard metrics are not yet implemented, and the data the underlying
+`/metrics` handler exposes is shard-wide with no per-workspace or per-shard
+meaning. The kcp HTTP filter rejects these requests before authorization runs
+so that a future implementation can fill in real workspace-scoped metrics
+without changing the URL contract.
```
ntnn: This feels like it waves three things at once and makes it confusing to read for people who don't already know whats going on.

1. `/metrics` on workspaces and VWs returns 501
2. `/metrics` on the shard/cache server is aggregated over all resources
3. Discussing future implementations for a per-workspace

I think these should be separate topics and not in one blob of text.

docs/content/setup/production/metrics.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Authorizing a scraper on a shard
+
+kcp ships a bootstrap `ClusterRole` named `system:kcp:metrics-reader` that grants
+`GET` on `/metrics`. To allow an identity to scrape every shard, create a
+`ClusterRoleBinding` in the `:root` workspace. The binding is replicated to all
+shards via the cache server, so a single binding is enough.
```
ntnn: Arguably you can only do a single binding because there's only one root workspace on the root shard. Maybe reword the last sentence a bit?

docs/content/setup/production/metrics.md
```diff
@@ -0,0 +1,127 @@
...
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:kcp:metrics-reader
+subjects:
+  - kind: User
+    apiGroup: rbac.authorization.k8s.io
+    name: prometheus
+```
+
+Apply it against the root workspace:
```
ntnn:
```suggestion
Apply it to the root workspace:
```

docs/content/setup/production/metrics.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Note: workspace-local `nonResourceURLs: /metrics` no longer works
+
+Earlier kcp releases accidentally allowed a workspace administrator to grant
+themselves access to shard metrics by creating a `ClusterRole` with
+`nonResourceURLs: ["/metrics"]` and a binding inside their own workspace. This
+was a privilege escalation: the data exposed is shard-wide, not workspace
+content. The path is now rejected at the workspace scope and the only
+authoritative binding is one created in `:root`.
+
```
ntnn: Should we put this into documentation for people setting up metrics now? If someone used this and it broke I'd guess they check at least the existing issues and then see that it was a security issue.

pkg/server/filters/shardlevelpaths.go
```diff
@@ -0,0 +1,62 @@
...
+//     root-workspace privileges.
+func WithShardLevelPaths(handler http.Handler) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		if !shardpaths.Paths.Has(req.URL.Path) {
+			handler.ServeHTTP(w, req)
+			return
+		}
+
+		cluster := request.ClusterFrom(req.Context())
+		if cluster != nil && !cluster.Name.Empty() && cluster.Name != core.RootCluster {
```
ntnn: So `/clusters/root/metrics` is allowed?

pkg/cache/server/handler.go
```diff
@@ -62,10 +65,13 @@ func init() {
...
 // not all paths require to have a valid shard name,
-// as of today the following paths pass through: "/livez", "/readyz", "/healthz".
+// as of today the following paths pass through: "/livez", "/readyz", "/healthz",
+// and any path declared shard-level in pkg/authorization/shardpaths (e.g. /metrics)
+// so that prometheus-style scrapers can hit the cache server directly without
+// constructing a /shards/<sh>/ prefix.
 func WithShardScope(handler http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-		if path := req.URL.Path; path == "/livez" || path == "/readyz" || path == "/healthz" {
+		if path := req.URL.Path; path == "/livez" || path == "/readyz" || path == "/healthz" || shardpaths.Paths.Has(path) {
```
ntnn, line 74: Should `/livez` etcpp also land in shardpaths?

test/e2e/authorizer/shardmetrics_test.go
```diff
@@ -0,0 +1,175 @@
...
+	})
+
+	t.Run("top-level /metrics for system:masters succeeds", func(t *testing.T) {
+		var code int
+		body, err := adminKubeClusterClient.RESTClient().Get().AbsPath("/metrics").Do(ctx).StatusCode(&code).Raw()
+		require.NoError(t, err)
+		require.Equal(t, http.StatusOK, code)
+		require.Contains(t, string(body), "apiserver_")
+	})
+}
```
ntnn, line 175: Can we add some test against the cache-server? Maybe just that `/metrics` refuses?

pkg/cache/server/handler.go
mjudeikis: Maybe. If we evern decide to change we can remove them 🤷

pkg/server/filters/shardlevelpaths.go
mjudeikis: it was. but I think for consistency we remove it

docs/content/setup/production/metrics.md
mjudeikis: Wdym? This is docs page.

ntnn (approved):
/lgtm
/approve

ntnn (approved):
/lgtm
/approve

mjudeikis:
/retest 

quota bug

mjudeikis:
/retest 
OCI is funky again

ntnn (approved):
/lgtm
/approve

PR #4178, opened 2026-05-29
Stable VW URL ordering in WorkspaceType status

ntnn:
## Summary

The WorkspaceType reconciler keep churning the status until the VW URLs are randomly in the right order to stop reconciling.
This makes the built URL list stable so one reconcile is enough.

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/tenancy/workspacetype/workspacetype_controller_reconcile.go
+++ b/pkg/reconciler/tenancy/workspacetype/workspacetype_controller_reconcile.go
@@ -17,13 +17,16 @@ limitations under the License.
 package workspacetype
 
 import (
+	"cmp"
 	"context"
 	"fmt"
 	"net/url"
 	"path"
+	"slices"
 
 	"k8s.io/klog/v2"
 
+	corev1alpha1 "github.com/kcp-dev/sdk/apis/core/v1alpha1"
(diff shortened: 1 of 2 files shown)


mjudeikis (approved):
/lgtm
/approve

mjudeikis:
/hold

just thinking outloud, will this now force every xEndpointSlice author to do this ordered?

mjudeikis:
/lgtm cancel

do we really want this? any way to make this without ordering?

ntnn:
> /hold
> 
> just thinking outloud, will this now force every xEndpointSlice author to do this ordered?

We don't need to - it's just the way this reconciler works. The wrapper in the reconciler compares slices ordered instead of as a set. 
Authors of other xES can do this however they want to



> /lgtm cancel
> 
> do we really want this? any way to make this without ordering?

We'd have to change how the reconciler patches.

I don't think it's a big deal - it was ordered before, it was only unordered because the URLs were put in a map and then moved map->list, which makes the order random.
And since the only reconciler modifying the xES should be this reconciler it doesn't matter for other reconcilers.

ntnn:
/test pull-kcp-test-e2e-shared

looks like a flake

ntnn:

that's one of the new tests, its maybe flaking?

PR #4180, opened 2026-05-31
Add kcptesting.WithDefaultTokenAuthFile

ntnn:
## Summary

Why?

`cmd/test-server/kcp` is importing `test/e2e/framework` for the default token auth file.
This causes e.g. `sharded-test-server` to panic if run outside of the kcp repository.

Instead the default token auth file is an option that can be passed to the server upon init.
Funny enough that pattern already existed for the audit policy file.

I also wanted to try and scrap the `init` alltogether but that would technically be an API change and I want to backport this.

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/cmd/sharded-test-server/main.go
+++ b/cmd/sharded-test-server/main.go
@@ -34,6 +34,7 @@ import (
 	genericapiserver "k8s.io/apiserver/pkg/server"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/retry"
+	"k8s.io/klog/v2"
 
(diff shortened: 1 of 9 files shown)


cmd/sharded-test-server/main.go
```diff
@@ -44,6 +45,7 @@ import (
 )
 
 func main() {
+	klog.InitFlags(flag.CommandLine)
```
ntnn, line 48: `test/e2e/framework` calls this, so sharded-test-server had this implicitly.

mjudeikis (approved):

PR #4189, opened 2026-06-04
Add static tilt setup

mjudeikis:
## Summary

When developing ontop of kcp one might want to integrate kcp into their setup but dont need reloads. 
This adds static tilt setup option where once can use it same as they would use grafana or other tool. 

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/contrib/tilt/front-proxy.static.yaml
+++ b/contrib/tilt/front-proxy.static.yaml
@@ -1,0 +1,45 @@
+apiVersion: operator.kcp.io/v1alpha1
+kind: FrontProxy
+metadata:
+  name: frontproxy
+spec:
+  # no image override: the kcp-operator defaults to its version-matched
+  # upstream kcp image (static install, no locally built images).
+  rootShard:
+    ref:
+      name: root
+  external:
+    hostname: kcp.localhost
+    port: 8443
+  deploymentTemplate:
+    spec:
+      template:
+        spec:
+          hostAliases:
+            - ip: 10.96.2.2
+              hostnames:
+                - kcp.localhost
+                - root.kcp.localhost
+                - theseus.kcp.localhost
+  certificateTemplates:
+    server:
+      spec:
+        dnsNames:
+        - kcp.localhost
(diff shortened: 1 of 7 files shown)


contrib/tilt/front-proxy.static.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: operator.kcp.io/v1alpha1
+kind: FrontProxy
+metadata:
+  name: frontproxy
+spec:
+  # no image override: the kcp-operator defaults to its version-matched
+  # upstream kcp image (static install, no locally built images).
```
ntnn, line 7: We could probably solve this with kustomizations, or? static is the default and dev is an overlay that loads in a specific image?
That way we aren't duplicating each manifest

ntnn: And the tilfile select static or dev kustomization depending on an env var or something

mjudeikis: The problem with all these are you can't include them as is like:
```
kcp_tilt = os.path.join(kcp_dir, 'contrib', 'tilt', 'Tiltfile.static')
if not os.path.exists(kcp_tilt):
    fail('expected upstream kcp Tiltfile at %s — is --kcp-dir pointing at a kcp checkout?' % kcp_tilt)

include(kcp_tilt)
```
Basically, this (for now) will make external usage easier. And once we get some patterns sorted, we can see how we can optimise. I don't want to overoptimize it yet until we get external consumers using this so we can see where it breaks

mjudeikis: Basically, once I prove this works with external contributions and we have playground/downstream of our own - we can look into optimizations

ntnn: Sounds good

ntnn (approved):
/lgtm
/approve

PR #4191, opened 2026-06-05
Add logical cluster migration

ntnn:
## Summary

This PR adds migrating logical clusters between shards. 
The intent of an admin to migrate an LC is expressed using `LogicalClusterMigration.migration.kcp.io`.
The object is replicated to the cache-server and used to synchronize the migration between the origin and destination shard.
The destination shards pulls the data of the migrating LC using the `LogicalClusterDump` type through the migrating VW.
The migrating VW was added only to not put the pressure of the migration on the front-proxy as well. 

More details on the process are in the reconciler package docs.

The happy path works but it still needs some resilience and we'll probably hit some real blockers once we try it on real infrastructure.
To actually use migrations a) the feature gate has to be enabled and b) the APIExport migration.kcp.io has to be bound.

## What Type of PR Is This?

/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Added migrating logical clusters between shards gated by the LogicalClusterMigration feature gate
```


ntnn:
/test pull-kcp-test-e2e

flake

```text
 === FAIL: test/e2e/authorizer TestShardMetricsEndpoint (30.36s) 
 ```

File: staging/src/github.com/kcp-dev/apimachinery/third_party/informers/shared_informer.go [hunk 1]
```diff
@@ -600,6 +634,16 @@ func (s *sharedIndexInformer) handleDeltas(logger klog.Logger, obj interface{},
 func (s *sharedIndexInformer) handleBatchDeltas(logger klog.Logger, deltas []cache.Delta, isInInitialList bool) error {
 	s.blockDeltas.Lock()
 	defer s.blockDeltas.Unlock()
+
+	if s.ignoreFunc != nil {
```
gman0, line 662: Why doesn't `handleBatchDeltas` need to special-case ReplacedAllInfo deltas, similar to how `handleDeltas` does?

pkg/informer/informer.go
```diff
@@ -282,6 +294,26 @@ func (d *DiscoveringDynamicSharedInformerFactory) ClusterWithContext(ctx context
...
+
+	for _, inf := range d.informers {
+		store := inf.Informer().GetIndexer()
+		objs, err := store.ByIndex(kcpcache.ClusterIndexName, kcpcache.ClusterIndexKey(cluster))
+		if err != nil {
+			utilruntime.HandleError(fmt.Errorf("failed to list objects for cluster %v: %w", cluster, err))
+			continue
+		}
+		for _, obj := range objs {
+			if err := store.Delete(obj); err != nil {
```
gman0, line 310: Is it possible the handlers that use this informer are still processing this object? It sounds a bit racy :D I wonder if it could be moved into the controller (the caller, using this informer), purge outside of the `reconcile` func while holding a lock. So that all threads can observe the same events.

staging/src/github.com/kcp-dev/apimachinery/third_party/informers/shared_informer.go
```diff
@@ -342,6 +342,24 @@ func (s *sharedIndexInformer) SetIgnoreFunc(fn func(interface{}) bool) error {
 	return nil
 }
 
+// ForceRelist causes the underlying reflector to perform a full relist
+// on the next list/watch cycle.
```
gman0: ... causes the underlying reflector to perform a full relist IF the said reflector also implements this magical forceRelister interface.

I think more context could be added into the comment. With the single user of this being the LC migration controller, it's not super obvious (without further digging) WHY is this even here.

staging/src/github.com/kcp-dev/apimachinery/third_party/reflector/reflector.go
```diff
@@ -174,6 +175,12 @@ type Reflector struct {
 	isLastSyncResourceVersionUnavailable bool
 	// lastSyncResourceVersionMutex guards read/write access to lastSyncResourceVersion
 	lastSyncResourceVersionMutex sync.RWMutex
+	// activeWatch holds the current watch so ForceRelist can stop it.
+	activeWatch     watch.Interface
```
gman0, line 179: A nit, but could this be `thisWatch`, just `watch` or something like that? :D While reading, I was wondering that if this is THE active watch, I was looking for where is the other, inactive watch.

staging/src/github.com/kcp-dev/apimachinery/third_party/informers/shared_informer.go
ntnn: `ReplacedAllInfo` isn't in the batchable map in the FIFO kube implementation, so it that type can't flow through the `handleBatchDeltas`. It'd make sense to be defensive here and add the implementation there as well but I decided not to do that because I wanted to keep the modifications small-ish. On top I don't think `ReplacedAllInfo` will be batched because it is a batching type by itself.

pkg/informer/informer.go
ntnn: Yesn't :D There is a race but I don't think the transition has to be perfect.

The ignore simply tells the informers to ignore any further events from that cluster, the purge cluster then cleans out the stores to prevent e.g. lists from seeing the (stale) objects; so arguable all handlers are observing the same events. I was thinking of wiring ignoreFunc into filterFunc (filterFunc filters events sent to handlers) but decided against it because filterFunc ignores new watched events, so it ignores all events for all handlers equally, while filterFunc could take hold while events are being fanned out to handlers.

And for cases where an object from a migrating cluster is already in the workqueue of a handler we can't do anything except hope that the handler is implemnented correctly and discards the object once it is gone. At that point I'd bet more on eventual consistency rather than trying to make the split perfect.

staging/src/github.com/kcp-dev/apimachinery/third_party/reflector/reflector.go
ntnn: Just `watch` clashes with the imported package `watch` (even if its a field^^ I try not to clash names) and imho `thisWatch` and `activeWatch` have the same problem, but `activeWatch` is a bit more speaking because that is the actively used watch^^
But we could use `cur(rent)Watch` e.g.

pkg/reconciler/migrating/logicalclustermigration/datacleanup.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	//	<prefix>/<group>/<resource>/customresources/<lc>/...
+	if parts[2] == "customresources" {
+		if len(parts) < 4 {
+			return "", ""
+		}
+		return prefix + parts[0] + "/" + parts[1] + "/customresources/" + parts[3], logicalcluster.Name(parts[3]) + "/"
+	}
+	//	<prefix>/<group>/<resource>/<lc>/...
+	return prefix + parts[0] + "/" + parts[1] + "/" + parts[2], logicalcluster.Name(parts[2]) + "/"
```
gman0: This handles resources originating from CRDs and built-ins, but not identity-based.

We'll need something like what we have in https://github.com/kcp-dev/kubernetes/blob/kcp-1.36.0/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store_kcp.go#L113

I'm wondering though how we can do that from the key alone; we have far less information than the apiextensions-apiserver code above ^^

I think we can assume a few things and say that:

* After all the trimming you're doing, the key we have is:
  ```
  []string{
    "<Group>",
    "<Resource>",
    [ "customresources" or "<Identity>" ,  ]
    "<Cluster>",
    [ "<Namespace>" , ]
    "<Name>",
  }
  ```
* Cluster name can be at the 3rd or 2nd index (0-counting) depending if the resource is built-in or not.
  * Since we don't know which, we should check both against `!= targetLogicalCluster+"/"`. If there is a match, we have a good key.
  * Downside is that we theoretically could have a false-positive match. But the only way to do that is if `<Identity>` matches `targetLogicalCluster`.
    * Consequence of that would be that an APIExport owner could theoretically forge such etcd keys (or rather those who bind to it) that they would be deleted even if the cluster is not matching. IMHO this is not a HUGE deal (and btw we should add salt to the identity hash function :D).

pkg/server/migrationdump/server.go
```diff
@@ -0,0 +1,231 @@
...
+
+		if !resp.More {
+			return entries, nil
+		}
+		key = string(resp.Kvs[len(resp.Kvs)-1].Key) + "\x00"
+	}
+}
+
+// belongsToLogicalCluster reports whether an etcd key belongs to the given logical cluster.
+func belongsToLogicalCluster(prefix, key string, target logicalcluster.Name) bool {
```
gman0: The same problem as in https://github.com/kcp-dev/kcp/pull/4191/changes#r3373299200 

As a side note, could this be refactored into a single place?

pkg/server/migrationdump/server.go
```diff
@@ -0,0 +1,231 @@
...
+
+// scanEtcdEntries returns every etcd key/value pair belonging to the given
+// logical cluster.
+//
+// TODO: stream entries as NDJSON so the response doesn't have to be buffered
+// in memory.
+//
+// TODO: paginate via spec.Continue / status.Continue so a single dump call
+// doesn't have to hold all keys in memory.
+func scanEtcdEntries(ctx context.Context, etcdClient *clientv3.Client, storagePrefix string, target logicalcluster.Name) ([]migrationv1alpha1.EtcdEntry, error) {
```
gman0, line 168: Worth refactoring with `scanEtcdEntries` pkg/reconciler/migrating/logicalclustermigration/datacleanup.go ?

pkg/server/migrationdump/server.go
```diff
@@ -0,0 +1,231 @@
...
+		writeError(w, r, apierrors.NewBadRequest("no cluster in context"))
+		return
+	}
+
+	user, ok := genericapirequest.UserFrom(ctx)
+	if !ok || user == nil {
+		writeError(w, r, apierrors.NewUnauthorized("no user info"))
+		return
+	}
+	if !slices.Contains(user.GetGroups(), bootstrappolicy.SystemExternalLogicalClusterAdmin) {
```
gman0, line 106: Worth documenting.

pkg/server/filters/migrationdump.go
```diff
@@ -0,0 +1,36 @@
...
+limitations under the License.
+*/
+
+package filters
+
+import (
+	"net/http"
+)
+
+const migrationDumpHandlerPath = "/apis/migration.kcp.io/v1alpha1/logicalclusterdumps"
```
gman0, line 23: Maybe this `migrationDumpHandlerPath` const could moved to somewhere else, along with the one at pkg/virtual/migratingworkspaces/doc.go ? Like - if, as a first-time-reader, I list all references to this const, it would be nice see that this is used in both of those places; makes it a bit easier to orient oneself too :)

pkg/virtual/migratingworkspaces/builder/build.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/kcp-dev/logicalcluster/v3"
+	"github.com/kcp-dev/virtual-workspace-framework/framework"
+	"github.com/kcp-dev/virtual-workspace-framework/pkg/handler"
+	"github.com/kcp-dev/virtual-workspace-framework/pkg/rootapiserver"
+
+	bootstrappolicy "github.com/kcp-dev/kcp/pkg/authorization/bootstrap"
+	"github.com/kcp-dev/kcp/pkg/virtual/migratingworkspaces"
+)
+
+func BuildVirtualWorkspace(
```
gman0, line 43: Worth adding a comment what's the purpose of this VW; what you've summarized in the PR description is a good starting point. Otherwise whoever reads this in a year's time will try to delete this :D

test/integration/logicalclustermigration/ignore_cluster_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/kcp-dev/logicalcluster/v3"
+	"github.com/kcp-dev/sdk/apis/core"
+	kcptesting "github.com/kcp-dev/sdk/testing"
+
+	"github.com/kcp-dev/kcp/test/integration/framework"
+)
+
+var configMapGVR = schema.GroupVersionResource{Group: "", Version: "v1", Resource: "configmaps"}
```
gman0: This global is guaranteed to get overwritten by some unfortunate person with something wrong. Move to the func?

pkg/server/filters/migrationdump.go
ntnn: I did that and produced an import cycle :D
But I didn't want to make a separate package just for this constant, but it would be better yeah. I was thinking of maybe placing it in a helper package in the API.

test/integration/logicalclustermigration/ignore_cluster_test.go
ntnn: That's true - I think I only did this because the linter complained about the same variable definition :D

pkg/reconciler/migrating/logicalclustermigration/doc.go
```diff
@@ -0,0 +1,49 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package logicalclustermigration implements migrating logical clusters between shards.
+//
+// The migration process is roughly:
+//
+// - LogicalClusterMigration is created
```
mjudeikis: Where it is created? What data we need to provide ?

pkg/reconciler/migration/logicalclustermigration/doc.go
```diff
@@ -0,0 +1,49 @@
...
+//
+// Origin shard (reconcileOriginCleanup):
+//
+//   - Purges all objects, including the LC from etcd
+//
+// Destination shard (reconcileDestinationFinalize):
+//
+//   - Updates the LC to remove the annotation
+//   - Resyncs all informers so they pick up the LC's objects
+//
```
mjudeikis, line 51: Did you have any tests with reversing these? Basically, purging before we start on other shards will prevent us from "reverting". If something went wrong, we can't rollback. 

And if cluster is in "removed" state, we should be able to start it on remote shard and do some validation?

pkg/server/filters/migratinglogicalcluster.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// WithBlockMigratingLogicalClusters rejects requests to logical clusters that are currently being migrated.
+//
+// This is very similar to WithBlockInactiveLogicalClusters, however the
+// migration requires that no client except other shards can access the
+// logical cluster, otherwise operators running with admin rights might
+// modify objects after they were migrated, producing an inconsistent
+// state.
+func WithBlockMigratingLogicalClusters(handler http.Handler, isMigrating func(logicalcluster.Name) bool) http.HandlerFunc {
```
mjudeikis, line 38: Do we have something like this in front-proxy index handler? I think we might need to manage state in there somehow too?

mjudeikis (commented):
Now 2 questions:
1. Do we have frontproxy routing logic changed in frontproxy? Frontproxy stores logicalclusters and their associated shards. So we need to handle it in there? Did I missed it?
2. Does this implies now that shards MUST be able to reach origin/destination VW urls for migration to happen? If so we should document this as this is West-East traffic. 

In general, we should document traffic we send West-East and North-South.

pkg/reconciler/migration/logicalclustermigration/reconciler.go
```diff
@@ -0,0 +1,277 @@
...
+			return false, err
+		}
+	}
+
+	// Cancel per-cluster context to stop all requests to it
+	// Just passing .Path is fine since lcName is the logical cluster id.
+	c.cancelLogicalClusterConnections(lcName.Path(), errors.New("logical cluster is being migrated"))
+
+	// Purge informer stores for this cluster
+	c.ddsif.PurgeCluster(lcName)
```
gman0, line 151: What about purging this cluster from the other local informers?

pkg/reconciler/migrating/logicalclustermigration/datacleanup.go
ntnn: I added pkg/etcd to combine the etcd key handling into one package and I think the parsing is better now as well :D

pkg/server/migrationdump/server.go
ntnn: Where should we document this? I could see adding this on the type in the SDK

pkg/server/filters/migrationdump.go
ntnn: I'd keep that as a followup

pkg/virtual/migratingworkspaces/builder/build.go
ntnn: I already added a few lines in the doc.go that explains that its sole purpose is to bypass the front-proxy; I'd hope that if someone tries to remove it they read that too and see the migration e2e tests fail :D

pkg/reconciler/migrating/logicalclustermigration/doc.go
ntnn: Anywhere, the LCMigration is propagated to cache and the shards only react to the object from the cache. I'll add that.

pkg/reconciler/migration/logicalclustermigration/doc.go
ntnn: No, at the moment its only the happy path. I want to work on resilience after merging because it's already very big.

I think we can purge after we enabled the LC on the other shard, yes.

ntnn: But I'd take that as a followup

pkg/server/filters/migratinglogicalcluster.go
ntnn: I don't think so - the shard rejects traffic for the LC, so even if the front-proxy routes traffic for the LC to the shard it is getting denied. Once the LC is active in the destination shard the front-proxy updates the mapping and routes traffic for the LC to the new shard.

Arguably there will be some things that may need adjusting, but the e2e tests cover the general use cases with watches and gets.

pkg/reconciler/migration/logicalclustermigration/reconciler.go
ntnn: We'll have to fix that as we get to them; I think there will be a bunch of things breaking once we try migrating an LC with shared APIs (either exporting or binding or both)

ntnn:

looks like flakes, nothing this PR touches

ntnn:



```text

=== Failed
=== FAIL: test/e2e/logicalclustermigration TestFullMigration/PreMigrateInformer (30.01s)
    migration_test.go:241: 
        	Error Trace:	/home/prow/go/src/github.com/kcp-dev/kcp/test/e2e/logicalclustermigration/migration_test.go:244
        	            				/usr/local/go/src/runtime/asm_arm64.s:1447
        	Error:      	Should be true
        	Messages:   	informer should have picked up post-migration-informer-cm
    migration_test.go:241: 
        	Error Trace:	/home/prow/go/src/github.com/kcp-dev/kcp/test/e2e/logicalclustermigration/migration_test.go:241
        	Test:       	TestFullMigration/PreMigrateInformer
        	Messages:   	waiting for informer to pick up post-migration object
I0615 09:04:16.684343   72843 reflector.go:434] "Stopping reflector" type="*v1.ConfigMap" resyncPeriod="0s" reflector="k8s.io/client-go@v0.0.0-20260602065202-e006560fc76a/tools/cache/reflector.go:343"
```

pkg/server/filters/migrationdump.go
ntnn: followup: https://github.com/kcp-dev/kcp/issues/4204

maxpain:
wow

PR #4193, opened 2026-06-05
Add registerControllerWithoutLeaderElection for per-replica controllers

mjudeikis:
## Summary

Controllers that only maintain per-process, in-memory state which a replica's own API server depends on (and never write to storage) must run on every replica, not just the elected leader. Previously the only such case — the DynamicRESTMapper builtin/dynamic controllers — was wired through a bespoke post-start hook guarded by a sync.Once.

Generalize that into the controller framework: add a controllersWithoutLeaderElection registry with registerControllerWithoutLeaderElection(), started once per process from the kcp-start-controllers post-start hook via startControllersWithoutLeaderElection(), independent of the leader-election machinery. This mirrors registerController()/ startControllers() for the leader-elected set.

Convert the DynamicRESTMapper controllers to register through it, removing the one-off hook and the sync.Once/cached-error fields. installDynamicRESTMapper is idempotent (installControllers can run again on leadership re-acquisition): it returns early if the controllers are already registered, so their informer event handlers are not added twice.

No functional change: the DynamicRESTMapper controllers still run on every replica exactly once.

## What Type of PR Is This?
/kind chore

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Refactor controller registration pattern for non-leader-election controllers
```

### DIFF
--- a/pkg/server/controllers.go
+++ b/pkg/server/controllers.go
@@ -32,7 +32,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	pluginvalidatingadmissionpolicy "k8s.io/apiserver/pkg/admission/plugin/policy/validating"
 	"k8s.io/apiserver/pkg/cel/openapi/resolver"
-	genericapiserver "k8s.io/apiserver/pkg/server"
 	"k8s.io/client-go/discovery"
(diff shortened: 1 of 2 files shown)


ntnn (commented):
/lgtm
/approve

PR #3515, opened 2025-08-06
Maintain a minimum and a build go version

ntnn:
## Summary

See https://github.com/kcp-dev/kcp/pull/3511#pullrequestreview-3090640629

Updates the script to maintain the minimum and build version in `go.mod`.

## What Type of PR Is This?

/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

PR #4194, opened 2026-06-05
remove misleading note about Go in getting started guide

xrstf:
## Summary
`IGNORE_GO_VERSION` has been removed in #3515, so no need to set it anymore. And when people set up their machines to work on kcp, we do not need to inform them about the exact rules on how to _update_ Go in kcp. These rules are also already covered by our verify-go-version script, so anyone accidentally doing the wrong thing would be caught anway.

## What Type of PR Is This?
/kind cleanup
/kind documentation

## Release Notes
```release-note
NONE
```


ntnn (approved):
/lgtm
/approve

PR #4202, opened 2026-06-15
Gate informers for MAP/MAPB on their feature gate

ntnn:
## Summary

MAP/MAPB was added during the kube rebase but is applied unconditionally, which could lead to errors if downstream explicitly disables them.

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/admission/mutatingadmissionpolicy/mutating_admission_policy.go
+++ b/pkg/admission/mutatingadmissionpolicy/mutating_admission_policy.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apiserver/pkg/admission/plugin/policy/mutating"
 	"k8s.io/apiserver/pkg/authorization/authorizer"
 	genericapirequest "k8s.io/apiserver/pkg/endpoints/request"
+	"k8s.io/apiserver/pkg/features"
 	"k8s.io/client-go/discovery/cached/memory"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/informers"
@@ -49,6 +50,7 @@ import (
 
 	"github.com/kcp-dev/kcp/pkg/admission/initializers"
 	"github.com/kcp-dev/kcp/pkg/admission/kubequota"
+	kcpfeatures "github.com/kcp-dev/kcp/pkg/features"
 )
 
 const PluginName = "KCPMutatingAdmissionPolicy"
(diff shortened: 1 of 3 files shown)


pkg/admission/mutatingadmissionpolicy/mutating_admission_policy.go
```diff
@@ -159,15 +156,15 @@ func (k *KubeMutatingAdmissionPolicy) SetAuthorizer(authz authorizer.Authorizer)
...
-func (k *KubeMutatingAdmissionPolicy) InspectFeatureGates(featureGates featuregate.FeatureGate) {
-	k.featureGates = featureGates
-}
-
 func (k *KubeMutatingAdmissionPolicy) ValidateInitialization() error {
 	return nil
 }
 
 func (k *KubeMutatingAdmissionPolicy) Admit(ctx context.Context, a admission.Attributes, o admission.ObjectInterfaces) error {
+	if !kcpfeatures.DefaultFeatureGate.Enabled(features.MutatingAdmissionPolicy) {
```
gman0: Could this be moved to the `Register` func instead? That way the plugin will not be enabled at all, instead of just skipping Admit.

pkg/informer/informer.go
```diff
@@ -978,7 +980,12 @@ var builtInInformableTypes map[schema.GroupVersionResource]GVRPartialMetadata =
 func (s *crdGVRSource) GVRs() map[schema.GroupVersionResource]GVRPartialMetadata {
 	crdGVRs := s.crdIndexer.ListIndexFuncValues(byGroupVersionResourceIndex)
 	result := make(map[schema.GroupVersionResource]GVRPartialMetadata, len(crdGVRs)+len(builtInInformableTypes))
+	mapEnabled := kcpfeatures.DefaultFeatureGate.Enabled(features.MutatingAdmissionPolicy)
```
gman0: `map` is a bit overloaded term, which in this case turns out is also an acronym. Full name pls?

pkg/admission/mutatingadmissionpolicy/mutating_admission_policy.go
ntnn: Good point, done

pkg/informer/informer.go
ntnn: True, renamed to `mutatingAdmissionPolicyEnabled`

gman0:
Actually, I wonder if we actually need to also gate the informers & replicaiton at all? Is it not enough to just gate the admission plugin and reconciler? AFAIK normally we only gate code that actually does stuff.

ntnn:
> Actually, I wonder if we actually need to also gate the informers & replicaiton at all? Is it not enough to just gate the admission plugin and reconciler?

I wasn't super sure either but better safe than sorry :shrug:
For replication I'm fairly certain we need that because if we disable MAP/MAPB the raw list would still list and start the informers and block the startup since those informers never sync if the cache-server doesn't offer those APIs

mjudeikis:
/retest 
`PreMigrateInformer` looks migration related?

ntnn:
That also had a flake in the other PR. I'll have to check why that is flaking. I'd guess its just a test implementation bug.

PR #4211, opened 2026-06-16
loadtest: configure CR sizing and complexity

SimonTheLeg:
## Summary

Adds the ability to specify sizes, complexity and number of bindings per workspace to mimic real-world use-cases better.

Specifically this allows creation of synthetic RessourceSchemas and their corresponding Exports and Bindings as well as objects.

This PR is chunked in two distinc commits to make reviewing it a bit easier, as it is quite large.

## What Type of PR Is This?

/kind feature

## Release Notes

```release-note
NONE
```

### DIFF
--- a/test/load/config.example.json
+++ b/test/load/config.example.json
@@ -8,6 +8,7 @@
     "crudConfigMapQPS": 150,
     "providerWorkspacesCount": 10,
     "consumerWorkspacesCount": 90,
+    "bindingsPerConsumer": 1,
     "createAPIExportQPS": 2.0,
     "createAPIBindingQPS": 2.0,
     "crudSharedAPIQPS": 150
--- a/test/load/config.example.json
+++ b/test/load/config.example.json
@@ -11,6 +11,9 @@
     "bindingsPerConsumer": 1,
     "createAPIExportQPS": 2.0,
     "createAPIBindingQPS": 2.0,
-    "crudSharedAPIQPS": 150
+    "crudSharedAPIQPS": 150,
+    "crLeafFields": 160,
+    "crListItems": 30,
+    "crTargetSizeBytes": 6000
   }
 }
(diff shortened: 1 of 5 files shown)


test/load/config.example.json
xrstf, line 2: This right here, this is the indication that JSON was not the best choice of data formats 😁

test/load/pkg/crgen/crgen_test.go
```diff
@@ -0,0 +1,292 @@
...
+			d := NewDummy(tt.leafFields, 0, 0)
+			ars := d.GenerateAPIResourceSchema("1")
+
+			require.Equal(t, schemaObjectName("1"), ars.Name)
+			require.Equal(t, "loadtest.kcp.io", ars.Spec.Group)
+			require.Equal(t, apiextensionsv1.NamespaceScoped, ars.Spec.Scope)
+			require.Equal(t, kindName("1"), ars.Spec.Names.Kind)
+			require.Equal(t, listKindName("1"), ars.Spec.Names.ListKind)
+			require.Equal(t, pluralResource("1"), ars.Spec.Names.Plural)
+			require.Equal(t, singularResource("1"), ars.Spec.Names.Singular)
```
xrstf: Are these all the same "1"? Can we make it a const?

test/load/pkg/crgen/crgen.go
```diff
@@ -0,0 +1,299 @@
...
+	paddedData := currentData + strings.Repeat("x", deficit)
+	_ = unstructured.SetNestedField(obj.Object, paddedData, "spec", "data")
+}
+
+// These derive deterministic resource names from a string id. All load test
+// resources follow the same naming convention so that providers and consumers
+// can reference each other by id alone.
+
+func pluralResource(id string) string {
+	return fmt.Sprintf("loadtestresources%s", id)
```
xrstf: Does it make sense to have the ID as the last element? Shouldn't it be "loadtest%dresources"?

test/load/config.example.json
SimonTheLeg: Yes!

xrstf (approved):
/approve

PR #4219, opened 2026-06-20
Handle cluster migrations gracefully in front-proxy

ntnn:
## Summary

Fixes the flake with the pre-migration watch.

## What Type of PR Is This?

/kind bug
/kind flake

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/index/index_test.go
+++ b/pkg/index/index_test.go
@@ -17,9 +17,12 @@ limitations under the License.
 package index
 
 import (
+	"context"
 	"testing"
+	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/kcp-dev/logicalcluster/v3"
 	corev1alpha1 "github.com/kcp-dev/sdk/apis/core/v1alpha1"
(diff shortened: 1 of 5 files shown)


ntnn:
/retest

 time=2026-06-20T21:50:58.066Z level=ERROR source=manager.go:259 msg="error reloading target set" component="scrape manager" err="invalid config id:kcp-shared-TestUserWithWarrants-1781992249"

pkg/proxy/lookup/lookup.go
```diff
@@ -87,6 +91,18 @@ func newClusterResolveHandler(delegate http.Handler, index proxyindex.Index) htt
 			return
 		}
 
+		// If the cluster was recently migrated and the request is not a fresh list or watch
+		// return a 410 to force the client to do a relist.
+		//
+		// This is only required for watches.
+		// Lists with a lower RV will get appropriate objects from the shard.
+		// Lists with a higher RV will get a 504 from the shard.
+		if index.RecentlyMigrated(result.Cluster) && isInProgressWatch(req) {
```
gman0, line 100: Hmm, what about clients that HAVE done a list before, already know a good RV on the new shard, and want to watch from there on; all while this 90s period is ticking? Maybe the error message could be a bit more helpful, like: `...; watch with resourceVersion temporarily unavailable`, or something along those lines.

This could be called a pretty hard nit-picking, but a random user, looking at logs of their client might not know what is going on.

pkg/index/index_test.go
```diff
@@ -696,3 +699,115 @@ func newLogicalCluster(cluster string) *corev1alpha1.LogicalCluster {
...
+
+	select {
+	case <-ctx.Done():
+	case <-time.After(wait.ForeverTestTimeout):
+		t.Fatal("expected in-flight watch context to be cancelled on shard change")
+	}
+
+	// A watch opened *after* the migration must be live, not stuck in a sticky
+	// cancelled state. This guards against using a sticky Cancel instead of a
+	// one-shot Delete, which would permanently kill all watches for the cluster.
```
gman0, line 791: Huh? Sounds like the agent had a bit more context than what's in the comment here :P

pkg/proxy/lookup/lookup.go
ntnn: Lists shouldn't be affected, they are just passed through to the shard and as mentioned in the comment they'll either have a lower RV than the new shard, in which case they'll get a diff to the current RV of the new shard.
If they have a higher RV than the new shard the shard will respond with a 504, causing the client to do a relist.

ntnn: The original implementation also rejected lists but the more I read the code for reflectors the less it made sense to handle lists instead of letting them through^^

pkg/index/index_test.go
ntnn: :D Sortof yes, the comments in the original code (from where I copied the context handling) talked about the sticky cancellation and thats what it ran with when building the tests :D

pkg/proxy/lookup/lookup.go
gman0: @ntnn, what I meant is that the user does List, knows the correct RV, and does a subsequent Watch starting from that point in time.

It's fine for Watch to reject - I don't disagree with that. Just the log msg is hiding the real reason for the rejection and would (at least for me) read as something unrelated being wrong, whereas in fact it's only a temporary thing.

pkg/index/index_test.go
gman0: Yeah but can you pls expand on that thought INSIDE the comment? :D Otherwise nobody knows what that means. Or maybe it's just that I'm uneducated and this is common knowledge?

mjudeikis (commented):
/lgtm
/approve

gman0:
@mjudeikis were the comments I sent resolved, or the review overall was deemed without merit and it was decided that this can go ahead?

mjudeikis:
> @mjudeikis were the comments I sent resolved, or the review overall was deemed without merit and it was decided that this can go ahead?

Sorry, I think I tagged this by mistake while going via all backlist
/approve cancel

pkg/proxy/lookup/lookup.go
ntnn: I updated the error message, I think its clearer now

pkg/index/index_test.go
ntnn: True :D Updated with a comment on how the contextmanager handles the context

mjudeikis:
/lgtm
/approve

I need this as its keep flaking in CI

ntnn:
/retest

That looks like infra errors?

mjudeikis:
/retest
Ci infra

PR #4233, opened 2026-06-27
Release CRD resource lock when the CRD is deleted

### LINKED ISSUE #3799: bug: APIBinding lock on crd deletion is not removed from logicalCluster annotation
### Describe the bug

When adding and deleting a crd in a workspace the GRV lock does not gets cleaned up, so one cannot create an APIBinding for the same apigroup (refer to slack thread https://kubernetes.slack.com/archives/C021U8WSAFK/p1769013090663529).
I suspect there is an issue that the expiry property is not written in the logicalcluster annotation: https://github.com/kcp-dev/kcp/blob/release-0.29/pkg/reconciler/apis/apibinding/logical_cluster_lock.go

### Steps To Reproduce

1. Create an example CRD

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: networks.foo.bar.com
spec:
  group: foo.bar.com
  versions:
    - name: v1alpha1
      served: true
      storage: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                foo:
                  type: string
                bar:
                  type: integer
  scope: Namespaced
  names:
    plural: networks
    singular: network
    kind: Network
```

2. Delete the created CRD
3. Take a look at the logicalCluster annotation:

```bash
kubectl get logicalcluster cluster -o jsonpath='{.metadata.annotations.internal\.apis\.kcp\.io\/resource-bindings}'
```

There is an entry ```"networks.foo.bar":{"c":true}``` showing the binded crd although it is deleted and there is no "e" property set.

4. Even after waiting the suggested 30s the entry is not cleaned up.

### Expected Behaviour

When a CRD is deleted the annotation is cleaned up (maybe with a delay allowed by CRDExpiry) so one can create an APIBinding for the same apigroup.

### Additional Context

_No response_

BergCyrill:
## Summary

The logicalclustercleanup controller clears a CRD locks expiry while the CRD exists, but the removal branch only fired for locks with a non-nil, passed expiry. So once an established CRD was deleted, its resource-bindings lock was never released, blocking new APIBindings for the same group-resource.

Release the lock when its CRD no longer exists and it has no expiry set.

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #3799

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/reconciler/apis/logicalclustercleanup/logicalclustercleanup_controller.go
+++ b/pkg/reconciler/apis/logicalclustercleanup/logicalclustercleanup_controller.go
@@ -308,9 +308,11 @@ func (c *controller) process(ctx context.Context, key string) error {
 			continue
 		}
 
-		// CRD doesn't exist.
-		if b.CRDExpiry != nil && time.Now().After(b.CRDExpiry.Time) {
-			logger.V(4).Info("removing expired CRD binding of non-existing CRD", "crd", gr)
+		// CRD doesn't exist (anymore): release the lock immediately if it was
+		// held without an expiry (the CRD was deleted), or once an existing
+		// expiry has passed.
+		if b.CRDExpiry == nil || time.Now().After(b.CRDExpiry.Time) {
+			logger.V(4).Info("removing CRD binding of non-existing CRD", "crd", gr)
 			delete(rbs, gr)
 		}
 	}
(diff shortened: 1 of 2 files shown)


mjudeikis (approved):
/lgtm
/approve

mjudeikis:
/retest 
this might be read failure/regression?

PR #4234, opened 2026-06-29
handle deletion of workspace while in migration

### LINKED ISSUE #4206: feature: Defensive logical cluster migration
### Feature Description

The initial logical cluster migration is somewhat fragile.

There are a couple cases that are currently not considered like what happens when a workspace for an LC is being deleted during the migration?
What happens if there is a cluster ID clash (shouldn't happen but a check is cheap)?
What happens if a shard is shutdown or losses connectivity?
etcpp

### Proposed Solution

Add test cases

### Alternative Solutions

_No response_

### Want to contribute?

### Additional Context

_No response_

SimonTheLeg:
## Summary

This PR introduces a new phase Aborting, which gets used when a Logicalcluster receives a DeletionTimestamp while it also has an ongoing migration. 

This then triggers two independent cleanup processed, which report their status as a condition. OriginCleaned already existed, this PR adds DestinationCleaned, which describes the progress of all data deleted in destination LC. Lastly condition Aborted is introduced to indicate that aborting has finished successfully. I think it makes sense to have this condition separatedly from Failed, as a migration can fail for tons of factors outside of the users control (e.g. network is down), but the aborted state in a way does provide a success indicator.

## What Type of PR Is This?

/kind feature

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/4206

## Release Notes

```release-note
NONE
```

### DIFF
--- a/config/crds/migration.kcp.io_logicalclustermigrations.yaml
+++ b/config/crds/migration.kcp.io_logicalclustermigrations.yaml
@@ -150,6 +150,7 @@ spec:
                 - Migrating
                 - OriginCleanup
                 - DestinationFinalize
+                - Aborting
                 - Completed
                 - Failed
                 type: string
--- a/config/root-phase0/apiexport-migration.kcp.io.yaml
+++ b/config/root-phase0/apiexport-migration.kcp.io.yaml
@@ -11,7 +11,7 @@ spec:
       crd: {}
   - group: migration.kcp.io
     name: logicalclustermigrations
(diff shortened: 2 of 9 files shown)


SimonTheLeg:
/retest

For me locally all of them pass without any issues, even in a multi-shard setup

SimonTheLeg:
/retest 

full log has disappeared 🙄

mjudeikis (approved):
/lgtm
/approve

PR #4238, opened 2026-06-30
Restore mounted workspace access through the front-proxy

xmudrii:
## Summary

Fixes a regression where requests to mounted workspaces fail with `Unauthorized` in sharded deployments. The regression was introduced by the fix for GHSA-c8w2-fgvx-vhv4, which made the front-proxy unconditionally strip inbound `X-Remote-*` identity headers when a request is not authenticated.

A mounted-workspace request crosses the front-proxy twice: the source shard's local proxy forwards
the request to the destination workspace's canonical URL, which is the front-proxy itself. That second hop previously carried no client certificate, so the front-proxy could not authenticate it and (correctly, per the CVE fix) cleared the forwarded identity headers — leaving the destination shard with no identity and returning 401.

This restores mounts without weakening the CVE fix, by authenticating the second hop the standard Kubernetes "trusted front proxy" way instead of trusting unauthenticated headers:

- The shard's local proxy now presents a requestheader-CA-signed client certificate when forwarding back to the front-proxy (new `--mount-proxy-client-cert-file` / `--mount-proxy-client-key-file` shard flags; the transport falls back to the previous behavior when unset).
- The front-proxy now runs its requestheader authenticator and unions the requestheader CA into its serving client-CA bundle, so it advertises that CA in the TLS handshake (otherwise the proxy never presents its cert) and authenticates the *original* user from the forwarded headers.
- The header-scrubbing logic from the CVE fix (`ClearAuthHeaders` / `SetAuthHeaders`) is left untouched: inbound `X-Remote-*` headers are still always cleared and re-stamped only from an authenticated identity. Trust is granted solely to certificates signed by the requestheader CA — which shards already trust — and is bounded further by `--requestheader-allowed-names`.

The change is opt-in and backward compatible: with the new flags unset, behavior is unchanged.

The `sharded-test-server` dev harness is updated to mint the mount-proxy certificate and pass the corresponding flags to the shards and the front-proxy.

## What Type of PR Is This?

/kind regression

## Release Notes

```release-note
Fixed mounted workspace access failing with `Unauthorized` in sharded deployments, a regression introduced by the fix for GHSA-c8w2-fgvx-vhv4. Enabling mounts now requires configuring the front-proxy with `--requestheader-client-ca-file` and each shard with `--mount-proxy-client-cert-file`/`--mount-proxy-client-key-file`.
```

/assign @xrstf @mjudeikis

### DIFF
--- a/cmd/sharded-test-server/frontproxy.go
+++ b/cmd/sharded-test-server/frontproxy.go
@@ -133,6 +133,19 @@ func startFrontProxy(
 		fmt.Sprintf("--root-kubeconfig=%s", filepath.Join(workDirPath, ".kcp", "root.kubeconfig")),
 		fmt.Sprintf("--shards-kubeconfig=%s", filepath.Join(workDirPath, ".kcp-front-proxy", "shards.kubeconfig")),
 		fmt.Sprintf("--client-ca-file=%s", filepath.Join(workDirPath, ".kcp", "client-ca.crt")),
+		// Trust requestheader-CA-signed clients (e.g. a shard's local proxy re-entering
+		// the front-proxy while forwarding a mounted workspace) to assert identity via
+		// X-Remote-* headers. Without this the front-proxy cannot authenticate the
+		// forwarded identity on that hop and clears the headers, breaking mounts.
+		fmt.Sprintf("--requestheader-client-ca-file=%s", filepath.Join(workDirPath, ".kcp", "requestheader-ca.crt")),
+		"--requestheader-username-headers=X-Remote-User",
+		"--requestheader-group-headers=X-Remote-Group",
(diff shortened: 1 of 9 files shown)


xrstf:
> The change is opt-in and backward compatible: with the new flags unset, behavior is unchanged.

i.e. without the flags, kcp is broken?

Are these flags then something the kcp-operator has to know about (i.e. "if kcp >= 0.32.3, then add those flags"?)

xrstf (approved):
/approve

xmudrii:
> i.e. without the flags, kcp is broken?
>
> Are these flags then something the kcp-operator has to know about (i.e. "if kcp >= 0.32.3, then add those flags"?)

Yes, exactly, we would need to update Helm charts and operator to use those flags after this PR is merged and the release is out.

PR #4252, opened 2026-07-08
Fix readiness in tilt lib

mjudeikis:
## Summary

Fix how readiness is computed in tilt lib.
Change gorup used in kubeconfig

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


gman0:
/lgtm
/approve

https://docs.tilt.dev/api.html#api.k8s_resource

> By default, Tilt will wait for pods to be ready if it thinks a resource has pods.

Maybe we can do better, by somehow telling Tilt these are NOT pod-having resources. Good enough for now I think.

Issue #4235, opened 2026-06-29
feature: Poor man quota limit


mjudeikis:
### Feature Description

We need a way to enforce (hard) object count limit on workspaces.
One can just create gazzilion of object and take down the shard itself. 

Would be good to have some sane default on the workspace, where user can update it later

### Proposed Solution

TBD

### Want to contribute?

- [ ] I would like to work on this issue.

mjudeikis:
done as initial cut

PR #4259, opened 2026-07-09
Add LogicalCluster level quota

mjudeikis:
## Summary

Implements #4235: a hard limit on the total number of objects in a logical cluster (across all resource types) — something ResourceQuota cannot express. Protects a shard from a single workspace, creating unbounded objects.

### How?

- A per-shard scanner periodically counts objects per logical cluster via a single keys-only etcd scan (`--logical-cluster-object-count-scan-interval`, default 60s). A new validating admission plugin (`core.kcp.io/LogicalClusterObjectCountLimit`) tracks deltas between scans (+1 create / −1 delete) and rejects creates with `403 Forbidden` once the limit is reached.
- Limit resolution: `core.kcp.io/max-total-objects` annotation on the LogicalCluster (admin-only, protected by `ReservedMetadata`) takes precedence; otherwise the shard-wide `--logical-cluster-total-object-limit` flag applies (default 0 = disabled). Annotation <= 0 opts a logical cluster out. The default never applies to r`oot/system:*` clusters so shard bootstrap can't be blocked.
- Deletes, updates, subresources, and events are never blocked; enforcement only starts once the logical cluster is Ready. Counts are eventually consistent by design ("poor man's quota") — drift self-corrects each scan.
- Metrics `kcp_logicalcluster_object_count{shard,cluster}` / `kcp_logicalcluster_object_limit` are published only for clusters at ≥90% of their limit to keep cardinality bounded.

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Add LogicalCluster quota limited --logical-cluster-object-count-scan-interval
```

### DIFF
--- a/docs/content/concepts/workspaces/object-count-limit.md
+++ b/docs/content/concepts/workspaces/object-count-limit.md
@@ -1,0 +1,65 @@
+---
+description: >
+  Enforce a hard limit on the total number of objects in a workspace.
+---
+
+# Total Object Count Limit
+
+kcp can enforce a hard limit on the total number of objects (across all resource types)
+in a logical cluster. This protects a shard from a single workspace creating so many
+objects that it degrades or takes down the shard. It complements the standard Kubernetes
+`ResourceQuota` support, which can only limit counts per resource type
+(`count/<resource>.<group>`), not the total across all types.
+
+## Configuration
+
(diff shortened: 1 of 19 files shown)


xrstf (approved):
/approve
/lgtm

PR #4266, opened 2026-07-13
drop usage of deprecated/insecure square/go-jose in favor of go-jose/go-jose

xrstf:
## Summary
The old library has known security issues (like CVE-2024-28180) that will never be fixed in the old code.

## What Type of PR Is This?
/kind chore

## Release Notes
```release-note
NONE
```


ntnn (approved):
/lgtm
/approve

PR #4244, opened 2026-07-03
Paginate LogicalClusterDump to avoid whole-cluster etcd transfers

iakmc:
## Summary
- Paginate `LogicalClusterDump` so migrations no longer transfer an entire logical cluster's etcd data in a single blocking request/response.
- Add `spec.continue`, `spec.limit`, and `spec.maxBytes` to `LogicalClusterDump`, and `status.continue` for the returned page's resume token.
- Add `status.entriesCopied` and `status.dumpContinue` to `LogicalClusterMigration` so copy progress is visible and a shard restart resumes from the last completed page instead of re-copying everything.
- `reconcileMigrating` now copies one page per reconcile and requeues until the origin reports no more pages, instead of blocking on the full copy in one call.
- Addressed review: fixed the `DataCopied` condition getting stuck at `False` after a transient error even once later pages succeed, cache the origin HTTP client per migration instead of rebuilding it on every page, and moved the page byte-size default next to `ScanPageSize` in `pkg/etcd`.

## Test plan
- [x] `go build ./...`
- [x] `go test ./pkg/reconciler/migration/... ./pkg/server/migrationdump/... ./pkg/etcd/...`
- [x] New unit tests for page-boundary scanning (limit, byte budget, oversized entry, continue-token resume, cluster filtering)
- [x] New unit tests for reconciler requeue/resume-after-restart behavior, and for the fixed stuck-condition bug
- [x] `make test-e2e-sharded-minimal WHAT=./test/e2e/logicalclustermigration/... TEST_ARGS="-run TestFullMigration -v"` — verifies `status.entriesCopied`/`status.dumpContinue` on a real 2-shard migration

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #4207

## Release Notes

```release-note
LogicalClusterDump is now paginated. Origin shard sends the data in pages (limited by entry count and total size), not the whole logical cluster in one response. LogicalClusterMigration status now shows entriesCopied and dumpContinue, so if destination shard restarts, it continues the copy from where it stopped instead of starting again.
```

PR #4272, opened 2026-07-14
Share and refcount origin migration client across concurrent migrations

### LINKED ISSUE #4267: feature: reuse origin migration client across concurrent migrations from the same shard
### Feature Description

Follow-up from [#4244](https://github.com/kcp-dev/kcp/pull/4244#discussion_r3569560727) (LogicalClusterDump pagination). In datacopy.go we cache one HTTP client per (origin shard, logical cluster) so a migration reuses the same connection instead of opening a new one for every page. Right now the client gets closed as soon as that one migration finishes.

Problem is the cache is per logical cluster, not per shard, so if two migrations run from the same origin shard at the same time they don't even share a client. And when one migration finishes we just close its client without checking if another migration from the same shard is still running. So with a few migrations happening from shard A to shard B around the same time, we keep opening and closing clients for shard A instead of reusing one.

### Proposed Solution

Key the cache by origin shard only, not shard+LC, so migrations from the same shard share a client. Add a simple refcount, +1 when a migration starts using a shard's client, -1 when it's done, only actually close it when it hits 0.

### Alternative Solutions

_No response_

### Want to contribute?

### Additional Context

_No response_

iakmc:
## Summary
- Key the destination shard's cached origin migration client by origin shard only (was `(originShard, logicalCluster)`), so concurrent migrations from the same origin shard share one client/HTTP transport instead of each building their own.
- Refcount each shard's cached client (`acquireOriginClient`/`releaseOriginClient`) so it's only torn down once the last concurrent migration using it finishes, instead of evicting per-migration.
- Follow-up from review on #4244.

## Test plan
- [x] `go build ./...`
- [x] `go test ./pkg/reconciler/migration/... ./pkg/server/migrationdump/... ./pkg/etcd/...`
- [x] New unit tests (datacopy_test.go) that directly check the sharing/refcount logic: logical clusters on the same shard share one client, different shards get different clients, calling acquire again for the same logical cluster doesn't double-count it, the shared client stays alive as long as at least one migration is still using it, and releasing something that was never acquired just does nothing instead of breaking.
- [x] New e2e test `TestConcurrentMigrationsFromSameOriginShard` - runs 3 migrations from the same origin shard at once and checks they all finish correctly with no data mixing between them. Passed locally on a real 2-shard cluster.
- [x] Also tested it live by hand: ran 3 migrations concurrently on a local cluster with a temporary delay added to make sure they would actually overlap, and checked the logs - only one client got created and one got torn down for all 3, not one each. Removed the temporary delay after.

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #4267

## Release Notes

```release-note
No user-facing behavior changes here — this only affects internal HTTP client reuse during migrations.
```

### DIFF
--- a/pkg/reconciler/migration/logicalclustermigration/controller.go
+++ b/pkg/reconciler/migration/logicalclustermigration/controller.go
@@ -43,7 +43,6 @@ import (
 	apisv1alpha2 "github.com/kcp-dev/sdk/apis/apis/v1alpha2"
 	corev1alpha1 "github.com/kcp-dev/sdk/apis/core/v1alpha1"
 	migrationv1alpha1 "github.com/kcp-dev/sdk/apis/migration/v1alpha1"
-	kcpsdkclient "github.com/kcp-dev/sdk/client/clientset/versioned"
 	kcpclientset "github.com/kcp-dev/sdk/client/clientset/versioned/cluster"
 	migrationv1alpha1client "github.com/kcp-dev/sdk/client/clientset/versioned/typed/migration/v1alpha1"
 	apisv1alpha1informers "github.com/kcp-dev/sdk/client/informers/externalversions/apis/v1alpha1"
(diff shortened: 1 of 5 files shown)


ntnn (approved):
/lgtm
/approve

PR #4275, opened 2026-07-15
[2/2] Add spec.deletionPolicy=WaitForSuccessor to APIBinding

mjudeikis:
## Summary

Builds on `APIBinding` resource adoption. Adds an opt-in deletion policy that holds the binding's finalizer when a bound resource has instances but no successor to adopt them, instead of deleting the instances.

  `spec.deletionPolicy: Delete | WaitForSuccessor   # default: Delete`

This implements kep: https://github.com/kcp-dev/enhancements/pull/14 0004

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Adds spec.deletionPolicy: Delete | WaitForSuccessor for better handling of the resources, served by apibindings
```


pkg/reconciler/apis/apibindingdeletion/apibinding_adoption.go
```diff
@@ -17,28 +17,54 @@ limitations under the License.
...
+	// APIExport serving the same group/resource with the same APIResourceSchema
+	// (by UID) and the same identity. That binding takes the instances over;
+	// they are served again as soon as it binds.
+	retainedAdopted retentionReason = "Adopted"
+
+	// retainedWaiting means spec.deletionPolicy is WaitForSuccessor and no
+	// successor exists yet. Instances stay in storage, covered by the deleting
+	// binding, and the binding's finalizer is held until a successor appears
+	// (or the policy is changed back to Delete).
+	retainedWaiting retentionReason = "WaitingForSuccessor"
```
SimonTheLeg, line 42: I think we should export these constants and the retentionReason type, so we in other packages or third parties can use them

pkg/reconciler/apis/apibindingdeletion/apibinding_adoption.go
```diff
@@ -17,28 +17,54 @@ limitations under the License.
...
+// WaitForSuccessor degrades to Delete when the LogicalCluster itself is
+// deleting (or already gone): workspace teardown removes bindings as part of
+// removing all content, and waiting for a successor that can never come would
+// deadlock the workspace in Terminating.
 //
-// Lookups go through (possibly lagging) informers. A just-created successor
-// can be missed on one pass; the deleting binding is requeued and re-evaluated
-// while its instances remain, so a transient miss only delays the handover.
-func (c *Controller) adoptedResources(clusterName logicalcluster.Name, apibinding *apisv1alpha2.APIBinding) (map[schema.GroupResource]string, error) {
+// Successor verification is same-storage only: the successor's APIExport must
```
SimonTheLeg, line 62: Does that mean on a standard multi-shard setup, a user would run into a `1 - (1 / number of shards)` probability of failing unless they specify which shard their logicalcluster goes to? Or am I reading this wrong?

If that is the case, I am wondering if we need some sort of protection so users don't shoot themselves in the foot on this one.

pkg/reconciler/apis/apibindingdeletion/apibinding_adoption.go
mjudeikis: This is internal to this controller for now. Once we get second consumer - we can do that. I dont want to expose before there is even a user. Its like functions... Dont export if you dont need to.

pkg/reconciler/apis/apibindingdeletion/apibinding_adoption.go
mjudeikis: Not quite. APIBindings are per shards and always on the same shard. So this is ok

PR #4280, opened 2026-07-19
Add TokenReview support for delegated auth in apiexport virtual workspace

### LINKED ISSUE #4279: feature: delegated authn/authz for APIExport providers (TokenReview/SubjectAccessReview against consumer workspaces)
### Feature Description

An APIExport-based service provider often terminates its own endpoints outside of kcp's request path: reverse tunnels, agent registration endpoints, webhooks, per-tenant proxies, MCP/gateway surfaces. Callers of those endpoints present credentials minted **in consumer workspaces** (typically ServiceAccount tokens the provider's own controllers provisioned there via permission claims, or end-user tokens). The provider then needs the standard Kubernetes auth-delegator pattern against the consumer workspace:

1. `POST tokenreviews.authentication.k8s.io` — authenticate the token (must run in the workspace that issued it), and
2. `POST subjectaccessreviews.authorization.k8s.io` — authorize the resolved identity against that workspace's RBAC.

Today a workspace-scoped provider cannot do either:

- **SA identity scoping.** The provider's own SA token authenticates only in its home logical cluster. Requests to `/clusters/<consumer>/...` resolve to an identity with no workspace access, and kcp hides the workspace — every call fails with `the server could not find the requested resource`, including the TokenReview/SAR POSTs. (Debugging aside: this 404 is very opaque for this failure mode — it reads like a missing API rather than "your credential does not work here".)
- **No sanctioned channel.** The APIExport virtual workspace serves bound resources and claimed stored resources, but not the authn/authz review APIs. And claiming them via PermissionClaims was explicitly ruled out in #1697 (correctly — generic claims on non-CRUD APIs are the wrong shape).

So the provider↔consumer relationship that APIBinding establishes gives the provider data-plane access to claimed resources, but no way to do delegated authentication/authorization for the consumers it serves.

### Workarounds today (and why they're unsatisfying)

1. **Run the provider with an admin / front-proxy credential.** Works everywhere, and is what dev setups tend to quietly do — which hides the gap until production, where handing a service provider cross-tenant admin is exactly what a least-privilege story should avoid.
2. **Self-provisioned auth-delegator grant.** Claim `clusterroles`/`clusterrolebindings` in the APIExport, and on each engaged cluster have the provider create a binding for its globally-qualified SA (`system:kcp:serviceaccount:<home>:<ns>:<name>`, per the GlobalServiceAccount aliasing) to a role granting `create` on `tokenreviews` + `subjectaccessreviews`, then call the review APIs on the native `/clusters/<consumer>` path. This works, but requires claiming *RBAC write* in consumer workspaces just to obtain two review APIs, every provider must reimplement the bootstrap dance, and the consumer's accepted claim ("clusterrolebindings: all") doesn't express the actual intent ("provider may validate credentials against my workspace").
3. **Side-step delegated auth entirely** — e.g. the provider compares presented tokens byte-for-byte against credentials it itself issued (readable via its claimed `secrets`). Fine for provider-issued agent credentials, but it removes consumer-workspace RBAC as the source of truth and doesn't cover end-user tokens at all.

### Proposed improvements

Ordered by how much they'd solve:

1. **Serve TokenReview / SubjectAccessReview on the APIExport virtual workspace, scoped to the engaged cluster.** `POST /services/apiexport/<cluster>/<export>/clusters/<consumer>/apis/authentication.k8s.io/v1/tokenreviews` (and the SAR equivalent) — the VW forwards to the consumer workspace's authenticator/authorizer. The APIBinding already encodes "this provider serves this workspace"; delegated auth for the callers a provider serves is arguably an intrinsic part of that relationship rather than an extra privilege.
2. **A typed, explicit claim instead of generic claimability.** Rather than reopening #1697, add a first-class field (e.g. `authDelegation: true` on APIExport, accepted on APIBinding like permission claims). The consumer sees exactly what they're granting — "provider may authenticate tokens and run access reviews against this workspace" — and can reject/revoke it. Option 1 gated on option 2's consent bit would be the complete shape.
3. **DX: make the foreign-credential 404 diagnosable.** `the server could not find the requested resource` for "your SA token is scoped to another workspace" sends people hunting missing APIBindings/CRDs. Even a documented note, or a distinguishable error for authenticated-but-out-of-scope identities on native paths, would save a lot of debugging time.

### Motivating case

Hit in production in: an edges agent (single APIExport, agents in consumer workspaces) terminates agent tunnels and needs to validate agent SA tokens minted in the consumer workspace. Dev worked (front-proxy admin kubeconfig = workaround 1), prod failed with the opaque 404 (`token review: the server could not find the requested resource (post tokenreviews.authentication.k8s.io)`); we currently ship a mix of workarounds 2 and 3.

### Want to contribute?

### Additional Context

- #1697 — ruled out generic PermissionClaims on SAR and other non-CRUD APIs; this proposal deliberately asks for an explicit, consented mechanism instead.
- #4105 — same architectural direction for WorkspaceType init/termination: typed claims honored by the authorizer, instead of privileged side-channels.
- GlobalServiceAccount identity aliasing (`system:kcp:serviceaccount:<cluster>:<ns>:<name>`) is what makes workaround 2 possible today; this issue is about not requiring RBAC-write claims to get there.

mjudeikis:
## Summary

Enables delegated authentication (TokenReview) through the apiexport virtual workspace.

- Add `TokenReview` to the built-in APIs served by the apiexport virtual workspace.
- Make `tokenreviews` claimable via permission claims.
- Add e2e test covering auth delegation.

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/4279

## Release Notes

```release-note
Add TokenReview support for delegated auth in apiexport virtual workspace
```

### DIFF
--- a/pkg/permissionclaim/permissionclaim_labeler.go
+++ b/pkg/permissionclaim/permissionclaim_labeler.go
@@ -47,7 +47,7 @@ var NonPersistedResourcesClaimable = map[schema.GroupResource]bool{
 	authorizationv1.SchemeGroupVersion.WithResource("selfsubjectrulesreviews").GroupResource():   false,
 	authorizationv1.SchemeGroupVersion.WithResource("subjectaccessreviews").GroupResource():      true,
 	authenticationv1.SchemeGroupVersion.WithResource("selfsubjectreviews").GroupResource():       false,
-	authenticationv1.SchemeGroupVersion.WithResource("tokenreviews").GroupResource():             false,
+	authenticationv1.SchemeGroupVersion.WithResource("tokenreviews").GroupResource():             true,
 }
 
 // Labeler calculates labels to apply to all instances of a cluster-group-resource based on permission claims.
--- a/pkg/virtual/apiexport/schemas/builtin/builtin.go
+++ b/pkg/virtual/apiexport/schemas/builtin/builtin.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	authenticationv1 "k8s.io/api/authentication/v1"
 	authorizationv1 "k8s.io/api/authorization/v1"
 	certificatesv1 "k8s.io/api/certificates/v1"
 	coordinationv1 "k8s.io/api/coordination/v1"
@@ -332,5 +333,15 @@ var BuiltInAPIs = []internalapis.InternalAPI{
 		GroupVersion:  schema.GroupVersion{Group: "authorization.k8s.io", Version: "v1"},
 		Instance:      &authorizationv1.LocalSubjectAccessReview{},
 		ResourceScope: apiextensionsv1.NamespaceScoped,
+	},
+	{
+		Names: apiextensionsv1.CustomResourceDefinitionNames{
+			Plural:   "tokenreviews",
+			Singular: "tokenreview",
+			Kind:     "TokenReview",
+		},
+		GroupVersion:  schema.GroupVersion{Group: "authentication.k8s.io", Version: "v1"},
+		Instance:      &authenticationv1.TokenReview{},
+		ResourceScope: apiextensionsv1.ClusterScoped,
 	},
 }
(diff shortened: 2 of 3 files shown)


ntnn (approved):
/lgtm
/approve

PR #4285, opened 2026-07-23
pkg/virtual: use a TTLCache for impersonating roundtrippers

### LINKED ISSUE #4084: feature: Cache http configs and transport in terminating VW
### Feature Description

https://github.com/kcp-dev/kcp/pull/4077 implemented the proxy for terminating VW.

Deletion of resources can take hours or even days depending on the process behind the CRs. E.g. if a database must get a full backup (as opposed to an incremental one) before being deleted and the database is quite large the CR can be kept in deletion for hours on end. 

An http.Transport is (iirc) quite heavy, so re-creating transports for every request can be very resource intensive.

See: https://github.com/kcp-dev/kcp/pull/4077#discussion_r3154406909

Also check if this needs to be implemented in the initializing VW.

### Proposed Solution

Use `shardlookup.TTLCache` to a) synchronize parallel requests to create the same transport and b) cache the transport.
When the workspace moves from terminating to deletion the transport can be evicted from the cache.

### Alternative Solutions

_No response_

### Want to contribute?

### Additional Context

_No response_

neolit123:
## Summary

```
Initializing and terminating VWs construct a new http.Transport
for every rest.ImpersonationConfig for a given request.
This is taxing on both the CPU and memory, so adding some
caching is desired.

The TTLCache is perfect for this since it supports both
eviction callbacks and concurrent enqueueing.

- Implement TransportCache in proxy.go.
- Add unit tests.

Benchmarks are omitted but as a short summary, for a 100
identical rest.ImpersonationConfig, the new approach is 5% faster
and 3x less memory to obtain an http.Transport.
```

## What Type of PR Is This?

/kind feature

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/4084

## Release Notes

```release-note
feature: cache impersonating HTTP transports in the initializing and terminating virtual workspaces as a performance improvement.
```

### DIFF
--- a/pkg/virtual/shared/proxy.go
+++ b/pkg/virtual/shared/proxy.go
@@ -21,12 +21,18 @@ limitations under the License.
 package shared
 
 import (
+	"context"
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"sort"
 	"strings"
 
+	utilnet "k8s.io/apimachinery/pkg/util/net"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/transport"
+
+	"github.com/kcp-dev/kcp/pkg/shardlookup"
 )
 
 // ServeProxy strips any client-supplied auth/impersonation headers from the
@@ -56,4 +62,107 @@ func ServeProxy(writer http.ResponseWriter, request *http.Request, forwardedHost
 		Transport: rt,
 	}
 	proxy.ServeHTTP(writer, request)
+}
+
+type idleConnectionsCloser interface {
+	CloseIdleConnections()
+}
+
+// closeIdleConnections closes the idle connections held by rt's underlying
+// http.Transport and reports whether it found one. The round-trippers built by
+// rest.TransportFor do not implement the function, so unwrap them until
+// an http.Transport is found.
+func closeIdleConnections(rt http.RoundTripper) bool {
+	for {
+		if closer, ok := rt.(idleConnectionsCloser); ok {
+			closer.CloseIdleConnections()
+			return true
+		}
+		wrapper, ok := rt.(utilnet.RoundTripperWrapper)
+		if !ok {
+			return false
+		}
+		rt = wrapper.WrappedRoundTripper()
+		if rt == nil {
+			return false
+		}
+	}
+}
+
+// TransportCache caches impersonating round-trippers so that long-lived
+// initializing and terminating workspaces do not rebuild an http.Transport
+// on every request. TTLCache allows for concurrent requests on the same
+// key to be enqueued. Idle connections are closed once an entry expires.
+type TransportCache struct {
+	cfg   *rest.Config
+	cache *shardlookup.TTLCache[http.RoundTripper]
+}
+
+// NewTransportCache creates a TransportCache that derives transports from cfg.
+func NewTransportCache(cfg *rest.Config) *TransportCache {
+	c := &TransportCache{
+		cfg:   cfg,
+		cache: shardlookup.NewTTLCache[http.RoundTripper](),
+	}
+	c.cache.OnEviction(func(rt http.RoundTripper) {
+		_ = closeIdleConnections(rt)
+	})
+	return c
+}
+
+// StartWithContext starts the background eviction goroutine and stops it when
+// ctx is cancelled.
+func (c *TransportCache) StartWithContext(ctx context.Context) {
+	c.cache.StartWithContext(ctx)
+}
+
+// TransportFor returns a cached or new round-tripper for the given impersonation config.
+func (c *TransportCache) TransportFor(imp rest.ImpersonationConfig) (http.RoundTripper, error) {
+	return c.cache.Get(impersonationKey(imp), func() (http.RoundTripper, error) {
+		cfg := rest.CopyConfig(c.cfg)
+		cfg.Impersonate = imp
+		return rest.TransportFor(cfg)
+	})
+}
+
+const impersonationKeySeparator = "\x00"
+
+// impersonationKey returns a string that uniquely identifies the given
+// impersonation config. \x00 (NUL) separator is used as it cannot be in
+// any of the values. This is faster and less memory than doing something
+// like: sorting, marshaling to JSON, and hashing.
+func impersonationKey(imp rest.ImpersonationConfig) string {
+	var b strings.Builder
+	b.WriteString("u=")
+	b.WriteString(imp.UserName)
+	b.WriteString(impersonationKeySeparator)
+	b.WriteString("uid=")
+	b.WriteString(imp.UID)
+
+	groups := append([]string(nil), imp.Groups...)
+	sort.Strings(groups)
+	for _, g := range groups {
+		b.WriteString(impersonationKeySeparator)
+		b.WriteString("g=")
+		b.WriteString(g)
+	}
+
+	keys := make([]string, 0, len(imp.Extra))
+	for k := range imp.Extra {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	for _, k := range keys {
+		vals := append([]string(nil), imp.Extra[k]...)
+		sort.Strings(vals)
+		for _, v := range vals {
+			b.WriteString(impersonationKeySeparator)
+			b.WriteString("e=")
+			b.WriteString(k)
+			b.WriteString("=")
+			b.WriteString(v)
+		}
+	}
+
+	return b.String()
 }
--- a/pkg/virtual/initializingworkspaces/builder/build.go
+++ b/pkg/virtual/initializingworkspaces/builder/build.go
@@ -225,6 +225,8 @@ func BuildVirtualWorkspace(
 			}
 		}),
 		HandlerFactory: handler.HandlerFactory(func(rootAPIServerConfig genericapiserver.CompletedConfig) (http.Handler, error) {
+			transportCache := virtualshared.NewTransportCache(cfg)
+
 			if err := rootAPIServerConfig.AddPostStartHook(workspaceContentName, func(hookContext genericapiserver.PostStartHookContext) error {
 				defer close(workspaceContentReadyCh)
 
@@ -238,6 +240,8 @@ func BuildVirtualWorkspace(
 						return nil
 					}
 				}
+
+				transportCache.StartWithContext(hookContext)
 
 				return nil
 			}); err != nil {
@@ -304,20 +308,18 @@ func BuildVirtualWorkspace(
 						return
 					}
 
-					thisCfg := rest.CopyConfig(cfg)
 					extra := map[string][]string{}
 					for k, v := range caller.GetExtra() {
 						extra[k] = v
 					}
-					thisCfg.Impersonate = rest.ImpersonationConfig{
+					authenticatingTransport, err := transportCache.TransportFor(rest.ImpersonationConfig{
 						UserName: caller.GetName(),
 						UID:      caller.GetUID(),
 						// Inject the synthetic group so the workspace content authorizer
 						// trusts this request as pre-authorized by the VW. This will be used for auditing.
 						Groups: append([]string{authorization.InitializerGroup(initializer)}, caller.GetGroups()...),
 						Extra:  extra,
-					}
-					authenticatingTransport, err := rest.TransportFor(thisCfg)
+					})
 					if err != nil {
 						http.Error(writer, fmt.Sprintf("could not create round-tripper: %v", err), http.StatusInternalServerError)
 						return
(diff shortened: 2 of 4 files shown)


pkg/virtual/shared/proxy.go
```diff
@@ -57,3 +63,106 @@ func ServeProxy(writer http.ResponseWriter, request *http.Request, forwardedHost
...
+			return false
+		}
+	}
+}
+
+// TransportCache caches impersonating round-trippers so that long-lived
+// initializing and terminating workspaces do not rebuild an http.Transport
+// on every request. TTLCache allows for concurrent requests on the same
+// key to be enqueued. Idle connections are closed once an entry expires.
+type TransportCache struct {
```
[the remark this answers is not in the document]
neolit123: file was missing. now commited.

pkg/virtual/shared/proxy.go
```diff
@@ -57,3 +63,106 @@ func ServeProxy(writer http.ResponseWriter, request *http.Request, forwardedHost
...
 }
+
+type idleConnectionsCloser interface {
+	CloseIdleConnections()
+}
+
+// closeIdleConnections closes the idle connections held by rt's underlying
+// http.Transport and reports whether it found one. The round-trippers built by
+// rest.TransportFor do not implement the function, so unwrap them until there
+// an http.Transport is found.
```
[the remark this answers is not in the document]
neolit123: fixed in https://github.com/kcp-dev/kcp/compare/5293c1f41df0843edf82e93e582edc1e343d7767..067fe113b93e0948e7a4068c10d97b0118fae44d

neolit123:
```
=== SKIP: test/e2e/virtual/replication TestCachedResourceVirtualWorkspace (0.00s)
    virtualworkspace_test.go:61: TestCachedResourceVirtualWorkspace is flaking (ref https://github.com/kcp-dev/kcp/issues/4026)
=== Failed
=== FAIL: test/e2e/quota TestClusterScopedQuota (41.70s)
    quota_test.go:581: Shared kcp server will target configuration "/home/prow/go/src/github.com/kcp-dev/kcp/.kcp/admin.kubeconfig"
    quota_test.go:581: Waiting for readiness for server at https://127.0.0.1:6444/
    quota_test.go:591: Waiting for tenancy.kcp.io/v1alpha1 to be served
    quota_test.go:591: Waiting for tenancy.kcp.io/v1alpha1 to be served
    quota_test.go:591: Waiting for tenancy.kcp.io/v1alpha1 to be served
```
https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/4285/pull-kcp-test-e2e-sharded/2080308338167386112

/retest

pkg/virtual/shared/proxy.go
```diff
@@ -57,3 +63,106 @@ func ServeProxy(writer http.ResponseWriter, request *http.Request, forwardedHost
...
+
+// NewTransportCache creates a TransportCache that derives transports from cfg.
+func NewTransportCache(cfg *rest.Config) *TransportCache {
+	c := &TransportCache{
+		cfg:   cfg,
+		cache: shardlookup.NewTTLCache[http.RoundTripper](),
+	}
+	c.cache.OnEviction(func(rt http.RoundTripper) {
+		_ = closeIdleConnections(rt)
+	})
```
mjudeikis, line 109: I think this eviction hook is actively harmful here: for this config (no custom `Dial`/`Proxy`/`TLS.GetCert`), `rest.TransportFor` does **not** build a fresh `http.Transport` per call — client-go's global `tlsTransportCache` already dedupes the underlying `http.Transport`, and the impersonation/bearer wrappers are layered on top of that shared instance.

So `closeIdleConnections` unwraps down to that *shared, process-wide* transport and closes its idle pool. Every time any identity's entry expires (every ~60s under steady traffic, since the TTL cache disables touch-on-hit), the idle connections serving **all** identities — and any other client built from an equivalent config (`kubeClusterClient`, `dynamicClusterClient`) — get torn down, forcing fresh TLS handshakes. That likely negates or inverts the perf win this PR is after.

Suggestion: drop the `OnEviction` → `CloseIdleConnections` hook entirely. The wrapper is GC-able and the underlying transport is owned by client-go's cache, not this cache.

Minor related note: evicting a cached *failure* entry calls `closeIdleConnections(nil)` — safe as written (both type assertions fail on a nil interface), but subtle; an explicit nil guard would make it obvious if the hook stays.

pkg/virtual/shared/proxy.go
```diff
@@ -57,3 +63,106 @@ func ServeProxy(writer http.ResponseWriter, request *http.Request, forwardedHost
...
+		wrapper, ok := rt.(utilnet.RoundTripperWrapper)
+		if !ok {
+			return false
+		}
+		rt = wrapper.WrappedRoundTripper()
+		if rt == nil {
+			return false
+		}
+	}
+}
```
mjudeikis, line 90: If eviction cleanup is genuinely needed (see comment below on whether it should exist at all), `utilnet.CloseIdleConnectionsFor` in `k8s.io/apimachinery/pkg/util/net` already implements exactly this unwrap loop — better to reuse it than to reimplement (the `bool` return only exists for the tests).

pkg/virtual/shared/proxy.go
```diff
@@ -57,3 +63,106 @@ func ServeProxy(writer http.ResponseWriter, request *http.Request, forwardedHost
...
+		vals := append([]string(nil), imp.Extra[k]...)
+		sort.Strings(vals)
+		for _, v := range vals {
+			b.WriteString(impersonationKeySeparator)
+			b.WriteString("e=")
+			b.WriteString(k)
+			b.WriteString("=")
+			b.WriteString(v)
+		}
+	}
```
mjudeikis, line 165: This encoding is ambiguous when extra keys/values contain `=`: `{"a": ["b=c"]}` and `{"a=b": ["c"]}` both serialize to `e=a=b=c`. The test suite covers the nil-value case (which doesn't collide) but not this one. Extras in kcp carry scopes/warrants where `=` is plausible (base64/JWT-ish values).

Since this key hands out *impersonating* transports, a collision means one identity silently reuses another identity's transport — an impersonation-confusion primitive, so I'd treat this as a must-fix even if exploitability today is low.

Fix is cheap: separate key and value with the NUL separator too, e.g. `e=<k>\x00<v>` (or distinct `ek=`/`ev=` fields).

pkg/virtual/shared/proxy.go
```diff
@@ -57,3 +63,106 @@ func ServeProxy(writer http.ResponseWriter, request *http.Request, forwardedHost
...
+		return rest.TransportFor(cfg)
+	})
+}
+
+const impersonationKeySeparator = "\x00"
+
+// impersonationKey returns a string that uniquely identifies the given
+// impersonation config. \x00 (NUL) separator is used as it cannot be in
+// any of the values. This is faster and less memory than doing something
+// like: sorting, marshaling to JSON, and hashing.
```
mjudeikis, line 133: "\x00 cannot be in any of the values" is asserted but not enforced. HTTP headers can't carry NUL, but usernames can originate elsewhere (e.g. client cert subjects) where NUL is technically possible: `UserName: "a\x00uid=b"` with empty UID collides with `UserName: "a", UID: "b\x00uid="`.

Belt-and-braces options: length-prefix each field, or fall back to an uncached transport when any field contains the separator.

pkg/virtual/shared/proxy.go
```diff
@@ -57,3 +63,106 @@ func ServeProxy(writer http.ResponseWriter, request *http.Request, forwardedHost
...
+type TransportCache struct {
+	cfg   *rest.Config
+	cache *shardlookup.TTLCache[http.RoundTripper]
+}
+
+// NewTransportCache creates a TransportCache that derives transports from cfg.
+func NewTransportCache(cfg *rest.Config) *TransportCache {
+	c := &TransportCache{
+		cfg:   cfg,
+		cache: shardlookup.NewTTLCache[http.RoundTripper](),
```
mjudeikis, line 105: Two notes on inheriting the `shardlookup` defaults here:

- The TTL cache also caches **errors** (10s failure TTL), so a transient `TransportFor` failure now makes that identity fail for up to 10s. Probably acceptable — these errors are config-derived and rare — but it's a behavior change worth calling out.
- The 1-minute `SuccessTTL` was tuned for shard lookups; for transports a longer TTL via `NewTTLCacheWithOptions` would reduce churn, especially since touch-on-hit is disabled so even hot identities are rebuilt every minute.

pkg/virtual/shared/proxy.go
mjudeikis, line 1: Nit: the transport cache has nothing to do with proxying — a separate `transport_cache.go`/`transport_cache_test.go` would keep the package tidier.

mjudeikis (commented):
Thanks — good direction, clean code, and solid tests for what they cover. Two substantive concerns before this merges, details inline:

1. **The `OnEviction` → `CloseIdleConnections` hook likely hurts more than it helps.** For this config, client-go's global TLS transport cache already dedupes the underlying `http.Transport`, so the eviction callback closes the idle pool of a transport shared by *all* identities (and other clients in the process) every time any entry expires. I'd drop it.

2. **The cache key encoding is ambiguous** (`=` in extra keys/values, and the NUL assumption is unenforced for usernames). Since collisions here conflate impersonated identities, the encoding should be made unambiguous.

Both are small, contained fixes. Also a couple of minor notes on TTL defaults / error caching and file placement inline.

mjudeikis:
Following up on my review with a more fundamental question: **do we need this cache at all?**

I checked client-go v0.36.0 (what kcp is on): `transport.New` already dedupes the underlying `http.Transport` via the global `tlsTransportCache` in [`transport/cache.go`](https://github.com/kubernetes/client-go/blob/v0.36.0/transport/cache.go). The cache key is built only from TLS options, dial, and proxy — impersonation and bearer tokens are deliberately *not* part of the key; they're applied as thin `RoundTripper` wrappers around the shared transport.

The VW config here has no custom `Dial`/`Proxy`/`GetCert`, so it's cacheable. That means pre-PR, `rest.TransportFor(thisCfg)` per request costs: one `rest.CopyConfig`, a cache-key computation + map lookup, and two small wrapper allocations (bearer token + impersonation). No new `http.Transport`, no new connections, no TLS handshakes — the connection pool was already shared across all impersonated identities.

So the premise "constructs a new `http.Transport` for every `rest.ImpersonationConfig`" doesn't hold for this config. A fresh transport per call only happens for *uncacheable* configs (custom `Dial`, `Proxy`, or `GetCert`), which isn't the case here. The benchmark numbers in the description (5% faster, 3x fewer allocs) are consistent with skipping wrapper allocations — if transport construction were actually being avoided, the difference would be dramatic.

Given that, the TTL cache buys very little while adding a security-sensitive cache key (see the collision comments in my review) and the eviction/`CloseIdleConnections` interaction with the shared transport.

@neolit123 was there a profile behind #710 showing transport construction as a real cost? If so, it'd be worth checking whether the config in that scenario was uncacheable — fixing *that* would address the root cause. Otherwise I'd lean towards closing this and revisiting if the wrapper allocations ever show up in a profile.

(One footnote: v0.36 moved the transport cache to weak pointers, so an unreferenced transport can be GC'd and rebuilt, but under steady traffic it stays alive — doesn't change the conclusion.)

neolit123:
thanks for the reviews @mjudeikis 
there are some good findings, especially the potential NUL exploits.

> Following up on my review with a more fundamental question: do we need this cache at all?

i'd leave that to @ntnn to comment on, who created the ticket.
- https://github.com/kcp-dev/kcp/issues/4084

> @neolit123 was there a profile behind https://github.com/kcp-dev/kcp/pull/710 showing transport construction

it's #4084, looks like i wrote #710 by mistake.

ntnn:
> I checked client-go v0.36.0 (what kcp is on): transport.New already dedupes the underlying http.Transport via the global tlsTransportCache in [transport/cache.go](https://github.com/kubernetes/client-go/blob/v0.36.0/transport/cache.go?rgh-link-date=2026-08-11T11%3A09%3A40Z). The cache key is built only from TLS options, dial, and proxy — impersonation and bearer tokens are deliberately not part of the key; they're applied as thin RoundTripper wrappers around the shared transport.

Ah I didn't know that, that is quite handy. I haven't compared in detail now but the key diff here is that we are caching by impersonation; if there is no concern between the transport and the impersonation config we can just rely on the upstream cache.

mjudeikis:
I think we dont know. If this starts showing in perff dumps - we can revisit. Now we are fixing smth we dont know its broken

neolit123:
discussed on slack to close this for the time being:
https://kubernetes.slack.com/archives/C09C7UP1VLM/p1789375012683889?thread_ts=1789372030.520739&cid=C09C7UP1VLM

PR #4288, opened 2026-07-24
Add global cluster-aware validating admission webhook

mjudeikis:
## Summary

Introduce a compiled-in admission plugin, `apis.kcp.io/GlobalValidatingWebhook`, that applies a single `ValidatingWebhookConfiguration` - supplied to the shard at startup via `--admission-control-config-file` - to requests in EVERY logical cluster, without a per-workspace WebhookConfiguration object.

This is to admission what `--authorization-webhook-config-file` is to authorization: one endpoint, wired at bootstrap, applied globally. Unlike the authorization webhook it is an admission webhook, so the callout carries the full object body + operation and **can deny**. The request's logical cluster is conveyed to the endpoint on the object's `kcp.io/cluster` annotation (`create/update` on the object, delete on the old object). Parsed config is run through API defaulting so nil namespace/object selectors do not silently match nothing.

This allows to start setting global policies on all instance. 

## What Type of PR Is This?

/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Add global/system-wide ValidatingWebhookConfiguration
```

### DIFF
--- a/pkg/admission/globalwebhook/plugin.go
+++ b/pkg/admission/globalwebhook/plugin.go
@@ -1,0 +1,217 @@
+/*
+Copyright 2026 The kcp Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package globalwebhook implements a GLOBAL, cluster-aware validating admission
+// webhook: a single ValidatingWebhookConfiguration, supplied to the shard at
+// startup (via --admission-control-config-file, this plugin's config section),
+// that fires for requests in EVERY logical cluster - without a per-workspace
+// WebhookConfiguration object, and tamper-proof against workspace admins.
+//
+// This is to admission what --authorization-webhook-config-file is to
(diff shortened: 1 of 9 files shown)


pkg/admission/globalwebhook/plugin.go
```diff
@@ -0,0 +1,217 @@
...
+	if obj, ok := attr.GetObject().(metav1.Object); ok && obj != nil {
+		if undo := validatingwebhook.SetClusterAnnotation(obj, clusterName); undo != nil {
+			defer undo()
+		}
+	}
+	if old, ok := attr.GetOldObject().(metav1.Object); ok && old != nil {
+		if undo := validatingwebhook.SetClusterAnnotation(old, clusterName); undo != nil {
+			defer undo()
+		}
+	}
```
xrstf, line 189: Is this typical Kube logic? I thought in general one would make a deep copy for temporary changes.

pkg/admission/globalwebhook/plugin.go
```diff
@@ -0,0 +1,217 @@
...
+		return fmt.Errorf("globalwebhook: create validating admission webhook: %w", err)
+	}
+	plugin.SetExternalKubeClientSet(p.kubeClusterClient.Cluster(clusterName.Path()))
+	plugin.SetNamespaceInformer(p.localKubeSharedInformerFactory.Core().V1().Namespaces().Cluster(clusterName))
+	plugin.SetAPISource(p.source)
+	if err := plugin.ValidateInitialization(); err != nil {
+		return fmt.Errorf("globalwebhook: validate initialization: %w", err)
+	}
+
+	// Convey the logical cluster to the external webhook via the kcp.io/cluster
```
xrstf, line 176: Beelzebub's lawyer here, wondering: If this is advertised as a way to implement _policies_, what will be the likelihood of the webhook basing its logic on workspace paths? Should we include the path as well? Else we basically force many webhooks to build up their own, front-proxy-style cache.

PR #4289, opened 2026-07-24
display a warning banner for old versions

SimonTheLeg:
## Summary

This change displays a banner on old releases to warn users that they are viewing an old release. Motivation was that we received community feedback, where a user searched for "kcp cluster management" on Google, which led to an old version back when we had this functionality, leading to some confusion.

## What it looks like now

unreleased (aka main)
grafik

latest release
grafik

old version
grafik

## Implementation Hints

I know this looks a little bit like much code to just display another message. However the revamp makes sense. Unlike previously with main, which can be directly obtained from url path, we now need to have the whole list of available versions, to figure out if we are the latest one. With this approach we now just go ahead and fetch it directly from mike's versions.json. This is much better than the alternative, where we would need to async wait for the version dropdown menue to render and then parse it out of the DOM.

## Next steps

I will need to cherrypick this to all release branches once, so it gets deployed to those versions. Good news is for future versions this is not necessary anymore and will work out of the box, since versions.json is global.

## What Type of PR Is This?

/kind documentation

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Display out of date banner on old documentation versions
```

### DIFF
--- a/docs/overrides/scripts/version_banner.js
+++ b/docs/overrides/scripts/version_banner.js
@@ -1,72 +1,40 @@
 (function () {
   "use strict";
 
-  const UNRELEASED_VERSION = "main";
+  const path = window.location.pathname;
+  const match = path.match(/^(.*?\/)(main|v\d+\.\d+[^\/]*)(\/|$)/);
+  if (!match) return;
 
-  function getCurrentVersion() {
-    const path = window.location.pathname;
-    const versionMatch = path.match(/\/(main|v\d+\.\d+[^\/]*)/);
-    return versionMatch ? versionMatch[1] : null;
-  }
+  const root = match[1];
+  const currentVersion = match[2];
 
-  function getAvailableVersions() {
-    const versionList = document.querySelector("ul.md-version__list");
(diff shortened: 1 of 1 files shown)


mjudeikis (commented):
/lgtm
/approve

SimonTheLeg:
alright let's start the first test baloon, before porting this to all releases:

/cherrypick release-0.32
/cherrypick release-0.31

SimonTheLeg:
let the train begin 🚄 

/cherrypick release-0.30
/cherrypick release-0.29
/cherrypick release-0.28
/cherrypick release-0.27
/cherrypick release-0.26
/cherrypick release-0.25
/cherrypick release-0.24
/cherrypick release-0.23
/cherrypick release-0.22
/cherrypick release-0.20
/cherrypick release-0.11
/cherrypick release-0.10

SimonTheLeg:
alright, the issue seems to be that we don't have this script below 0.31. I'll create the manual cherry-picks for the other versions. We definitely need those, some Google results still link to 0.10 versions

PR #4294, opened 2026-07-28
Add implementation of Access Virtual Workspace to contrib

### LINKED ISSUE #3839: feature: Implement Out-of-Tree "Access" Virtual Workspace
### Feature Description

We need to develop a standalone Virtual Workspace (VW) that acts as an access aggregator for kcp (and potentially other backends). The goal is to provide a "Discovery" endpoint that allows users or MCP (Model-Context Protocol) servers to query: "What workspaces/clusters do I have access to?"

This follows the SelfClusterAccessReview (SCAR) pattern, functioning similarly to a /me endpoint—authenticated but not requiring explicit authorization for the request itself, as it only returns resources the caller already owns or can see.

### Proposed Solution

- The Virtual Workspace (VW): A singleton service (accessible via FrontProxy) that lives in an external repository.
- Opt-in Indexing: Workspaces are only "indexed" by this service if they have an APIBinding to a specific system APIExport. This allows users to selectively pick which workspaces are discoverable.
- The RBAC Graph: A controller will watch ClusterRoleBindings and RoleBindings across shards to build an in-memory mapping of Subject (User/Group) -> LogicalCluster.
- One should be able to plug external index provider (index should be interface based) so if user using external webhook authorizer, it should be possible to hook in same functionality

Pluggable Access Providers: While the initial focus is kcp-native RBAC, the internal logic should be interfaced to allow for backends like OpenFGA or external ReBAC systems.

Technical Specification
1. The SCAR API
The API returns a list of accessible endpoints formatted for MCP consumption.

```
type SelfClusterAccessReview struct {
    metav1.TypeMeta   `json:",inline"`
    metav1.ObjectMeta `json:"metadata,omitempty"`
    Status            SelfClusterAccessReviewStatus `json:"status,omitempty"`
}

type SelfClusterAccessReviewStatus struct {
    // Clusters the subject has at least 'view' access to
    Clusters []AccessEndpointSlice `json:"clusters"`
}

type AccessEndpointSlice struct {
    ClusterName string `json:"clusterName"`
    Endpoint    string `json:"endpoint"`
}
```

Authentication: Must support Bearer Tokens and Certificates.

Filtering: When a request hits /services/access-virtual-workspace, the VW identifies the caller (User/Groups) and filters the global in-memory map to return only relevant LogicalClusters.

Indexing Logic: The indexer must account for logicalcluster, user, and group.

Proposed Roadmap / Action Items
[ ] Schema Definition: Formalize the SelfClusterAccessReview CRD/Structs.
[ ] Access Provider Interface: Define a Go interface to allow swapping between kcp-native RBAC and OpenFGA.
[ ] In-Memory Indexer: Build the controller to watch RBAC artifacts across shards and maintain the Subject-to-Cluster graph.
[ ] VW Server: Implement the REST handler for the SCAR endpoint.
[ ] kcp Integration: Wire up the APIExport/APIBinding machinery for the "Opt-in" indexing feature.
[ ] MCP Server Integration: Update kubernetes-mcp-server to consume this endpoint for dynamic inventory.

References
[Discussion: MCP Server Cluster Discovery](https://github.com/kcp-dev/kcp/issues)
[PR: Related MCP Changes](https://github.com/kcp-dev/kcp/issues)

### Alternative Solutions

_No response_

### Want to contribute?

### Additional Context

_No response_

cnvergence:
## Summary

This PR is introducing the SCAR API and its RBAC provider, a virtual workspace that answers "which workspaces does this user have access to?" in a single call - a SelfClusterAccessReview (SCAR), instead of N SelfSubjectAccessReviews. 

A follow-up commit in this PR adds the MCP virtual workspace at /services/mcp in the same binary, calling the provider in-process. After that: an external AccessProvider for non-RBAC authorizers, then ClusterRole rule evaluation and warrants/scopes support.

What this introduces:

- SelfClusterAccessReview types.
- the in-memory permission map.
- the AccessProvider seam, so deployments with a non-RBAC authorizer can populate the same graph.
- watches ClusterRoleBindings and RoleBindings across shards via multicluster-runtime and the apiexport provider.
- SCAR as a create-only REST storage on a fixed-group-version virtual workspace.
- serving and wiring: TLS, delegated authentication (front-proxy request header certs, bearer-token TokenReview, client certs), per-VW authorisation, and a /debug/graph snapshot endpoint.
- cmd/access-vw, cmd/scar-to-kubeconfig, config/ (APIExport, example APIBinding, deployment manifests), and docs.

Some open questions:

Should this be placed in an in-tree contrib module, or a separate repo?
Would you want to push a contrib image?

## What Type of PR Is This?

/kind feature

## Related Issue(s)

Fixes https://github.com/kcp-dev/kcp/issues/3839

## Release Notes

```release-note
Added contrib/access-vw, a virtual workspace serving the SelfClusterAccessReview API for
permission-aware workspace discovery.
```


cnvergence:
Will move into https://github.com/kcp-dev/contrib-access-virtual-workspace

PR #4298, opened 2026-07-31
Fix restMapper for core kcp resources

mjudeikis:
## What Type of PR Is This?
/kind bug

We should fix this api incosistency, but as api is in the wild - we need this hack

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Fix dynamic-restmapper when dealing with multi-version apis, where different object might not exists in v2 but does in v1
```


ntnn (approved):
/lgtm
/approve

neolit123 (commented):
might be worth adding a UT?

mjudeikis:
> UT

What is UT?

ntnn:
Man... that is ugly. I'm thinking if it wouldn't be better to just fix the APIs now. If we copy the v1alpha1 kinds up to v1alpha2 the bug should also be fixed and we don't break kube convention?

ntnn (approved):
/lgtm

neolit123:
> > UT
> 
> What is UT?

unit tests. i see that you added some in commit 2.

PR #4301, opened 2026-08-03
add documentation for the system workspaces

### LINKED ISSUE #3570: docs: document system workspaces (like shard-local `system:admin`)
We should write documentation on the special system workspaces of kcp (is there more than `system:admin` right now? I'm not aware). What they do, what is special about them, and how to access them.

`system:admin` for example is important in RBAC, because it powers the global authorizer, so every RBAC that is in that workspace applies to every logical cluster on the shard.

Another note here: It might be time to rename the concept from "system workspace" to "system (logical) cluster". There is no `Workspace` object for `system:admin` and I suspect the naming is a leftover from before the reshuffling of terminology.

xrstf:
## Summary
Fixes #3570

## What Type of PR Is This?
/kind documentation

## Release Notes
```release-note
NONE
```

### DIFF
--- a/docs/content/concepts/workspaces/.pages
+++ b/docs/content/concepts/workspaces/.pages
@@ -2,6 +2,7 @@ nav:
     - index.md
     - workspace-types.md
     - workspace-types-best-practices.md
+    - system-workspaces.md
     - virtual-workspaces.md
     - workspace-initialization.md
     - workspace-termination.md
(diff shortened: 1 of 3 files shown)


neolit123 (commented):
/retest
/lgtm
/hold

i read  trough the doc and LGTM as far as i know about these WS, but also explains some things that i didn't know.

just drop the hold if you think one review is enough.

(EDIT: erm, i guess an approver is required too)

ntnn (approved):
/lgtm
/approve
/hold cancel

PR #4302, opened 2026-08-04
Forward the caller's identity to virtual workspaces

mjudeikis:
## Summary

The shard proxies virtual resources to the virtual workspace over a connection that authenticates as the shard, so the VW authorized the shard rather than the user. Stamp the request-header identity on the forwarded request instead, clearing any inbound copies first.

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Forward caller identity to virtual-workspaces
```

### DIFF
--- a/pkg/server/virtualresources/server.go
+++ b/pkg/server/virtualresources/server.go
@@ -47,6 +47,7 @@ import (
 	"github.com/kcp-dev/kcp/pkg/cache/client/shard"
 	"github.com/kcp-dev/kcp/pkg/endpointslice"
 	"github.com/kcp-dev/kcp/pkg/indexers"
+	"github.com/kcp-dev/kcp/pkg/proxy/authheaders"
 	"github.com/kcp-dev/kcp/pkg/reconciler/dynamicrestmapper"
 	kcpfilters "github.com/kcp-dev/kcp/pkg/server/filters"
 )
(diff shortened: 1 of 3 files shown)


pkg/server/virtualresources/server.go
```diff
@@ -380,8 +381,37 @@ func newVirtualResourceHandler(cfg *rest.Config, vwURL, clusterNameOrWildcard st
 		return nil, err
 	}
 
-	proxy := httputil.NewSingleHostReverseProxy(scopedURL)
-	proxy.Transport = tr
+	proxy := &httputil.ReverseProxy{
+		Transport: tr,
+		Rewrite: func(r *httputil.ProxyRequest) {
+			r.SetURL(scopedURL)
+			r.SetXForwarded()
```
ntnn: https://cs.opensource.google/go/go/+/refs/tags/go1.26.5:src/net/http/httputil/reverseproxy.go;l=82

I had a look as this feels a bit fishy to me. To me this looks like we are taking the forwarded-for values from the incoming request. So the VW would get a request authorized by the shard, with the users identity but the forwarded-for comes from the user?

So if the VW decides anything based on this that is an injection point. I feel like we should scrub that? The user identity should be enough.

ntnn: Specifically this

> //   - The X-Forwarded-Host header is set to the host name requested
> //     by the client.

mjudeikis: Let me test. This been train of fixes and networking debugging. Might be left-over from prior test

pkg/server/virtualresources/server.go
```diff
@@ -380,8 +381,36 @@ func newVirtualResourceHandler(cfg *rest.Config, vwURL, clusterNameOrWildcard st
...
-	proxy := httputil.NewSingleHostReverseProxy(scopedURL)
-	proxy.Transport = tr
+	proxy := &httputil.ReverseProxy{
+		Transport: tr,
+		Rewrite: func(r *httputil.ProxyRequest) {
+			r.SetURL(scopedURL)
+			// SetURL clears the outbound Host along with pointing the request at
+			// the virtual workspace; keep the host the caller asked for, which is
+			// what a single-host reverse proxy forwards.
+			r.Out.Host = r.In.Host
```
ntnn: Maybe I'm being a bit dense here - but isn't the whole ide a of the VR handler to route to the VW URL of the VR?
If we forward to the host the user requested aren't we essentially forwarding to the shard itself?

as in `r.In.Host` is the shard and `r.Out.Host` must be the VW, so `r.SetURL(scopedURL)` looks correct to me

mjudeikis: gone

ntnn (approved):
/lgtm
/approve

mjudeikis:
Open

mjudeikis:
I messed up with my github config. need revert and will reopen

PR #4303, opened 2026-08-04
Add VirtualWorkspace hop limit

mjudeikis:
## Summary

A shard forwards virtual resources to a virtual workspace, and a virtual workspace may serve them by delegating back to the shard. Chained that way the two can also form a cycle, which nothing static distinguishes from a legitimate chain. Carry a hop count on each leg and refuse to go deeper than a real chain needs.

Ran into a case where, if misconfigured, one can take down all shard by creating funky traffic patterns

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Add hop limit for Virtual workspaces
```


xrstf (approved):
/approve
/lgtm

PR #4306, opened 2026-08-04
Remove un-used code

mjudeikis:
## Summary

I think we dont need this code, dont we?

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/config/crds/apis.kcp.io_apiexports.yaml
+++ b/config/crds/apis.kcp.io_apiexports.yaml
@@ -544,10 +544,6 @@ spec:
                             a projection of the referenced resource inside the workspaces that
                             bind to the APIExport.
                           properties:
-                            identityHash:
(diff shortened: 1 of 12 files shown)


mjudeikis:
@gman0 This field is kinda unused and required. Making any of the external endpoint-slices is bit hard to use as they need to come up with some non-existing value

gman0:
This then breaks the verbs provider, because there could be multiple exports with the same GR, but different identity.

https://github.com/kcp-dev/kcp/blob/46d2a1ca20d26d03a214aea3f1d033120e3a5154/pkg/server/aggregatingcrdversiondiscovery/verbs_provider.go#L59-L89

I'll need to run through the code again but maybe this can be fixed :thinking:

mjudeikis:
might be... So we need to fgix this in different place. got ya

PR #4313, opened 2026-08-04
cli/pkg/base/options.go: don't disable any CLI options

### LINKED ISSUE #4312: feature: allow CLI users to pass auth flags like --as and --as-group
### Feature Description

this is useful for impersonation from the CLI for various testing scenarios.

```
kubectl ws tree --as alice --as-group foobar
```

i'd just enable all of these:
https://github.com/kcp-dev/kcp/blob/5371a2a2ad3c5e4ca600ad39e05111ddaaee24fa/staging/src/github.com/kcp-dev/cli/pkg/base/options.go#L71-L79

`--cluster` is a bit confusing in the case of kcp, but there is a `--workspace` already too.

### Proposed Solution

enable all the default flags

### Alternative Solutions

_No response_

### Want to contribute?

### Additional Context

_No response_

neolit123:
## Summary

This allows passing the additional flags:

--as
--as-group
--client-certificate
--client-key
--cluster
--request-timeout

The only conflating one might be --cluster in the context of kcp, but there is already a --workspace flag.

e.g.

```
k ws tree --as alice
```

## What Type of PR Is This?

/kind feature

## Related Issue(s)

Fixes #4312

## Release Notes

```release-note
Enable the following flags for all CLI commands similarly to how kubectl supports them: --as, --as-group, --client-certificate, --client-key, --cluster, --request-timeout.
```

### DIFF
--- a/staging/src/github.com/kcp-dev/cli/pkg/base/options.go
+++ b/staging/src/github.com/kcp-dev/cli/pkg/base/options.go
@@ -68,16 +68,7 @@ func (o *Options) BindFlags(cmd *cobra.Command) {
 		)
 	}
 
-	// We add only a subset of kubeconfig-related flags to the plugin.
-	// All those with LongName == "" will be ignored.
 	kubectlConfigOverrideFlags := clientcmd.RecommendedConfigOverrideFlags("")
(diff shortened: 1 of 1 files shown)


mjudeikis (commented):
/lgtm
/approve

PR #4321, opened 2026-08-09
e2e: fix second EventuallyWithT loop in TestShardMetricsEndpoint to retry properly

### LINKED ISSUE #3933: bug: e2e: flaky RBAC admittance
We have a few places in e2e where tests rely on a user being admitted an RBAC permission, but the RBAC cache hasn't caught up yet, failing the test.

pujitha24:
## Summary

Fixes the second `require.EventuallyWithT` polling loop in `TestShardMetricsEndpoint` (ClusterRoleBinding creation retry), which still asserted against the outer `*testing.T` instead of the tick-local `*assert.CollectT`. This caused the loop to abort after a single failed attempt instead of retrying, mirroring a bug already fixed in the first loop in the same test.

## What Type of PR Is This?

/kind flake

## Related Issue(s)

Fixes #3933

## Release Notes

```release-note
NONE
```

---
AI assistance: this change was drafted with Claude Code.

### DIFF
--- a/test/e2e/authorizer/shardmetrics_test.go
+++ b/test/e2e/authorizer/shardmetrics_test.go
@@ -95,7 +95,7 @@ func TestShardMetricsEndpoint(t *testing.T) {
 				Name:     "user-1",
 			}},
 		}, metav1.CreateOptions{})
-		require.NoError(t, err)
+		require.NoError(c, err)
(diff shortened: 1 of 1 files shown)


neolit123 (commented):
/lgtm
/ok-to-test

PR #4326, opened 2026-08-11
Remove identityHash from virtual resource storage definition

### LINKED ISSUE #4324: feature: strip identity from virtual storage definition
### Feature Description

Virtual resource definition in APIExports contains identity:

```yaml
kind: APIExport
metadata:
  name: virtualmachines
spec:
  resources:
  - name: instances
    group: machines.svm.io
    schema: today.instances.machines.svm.io
    storage:
      virtual:
        reference:
          apiGroup: cache.kcp.io
          kind: ClusterCachedResourceEndpointSlice
          name: instances
        identityHash: 676a1a997f7507fdfcfd1fcf35920efdff1297cd215cad4f486220c30a5c7090
```

Bad, because:

* Assumes that *every* virtual resource has an identity
* Makes identity rotation is harder than it has to be for resources that *do* have identities (e.g. CachedResource)

The identity itself is used only in `pkg/server/aggregatingcrdversiondiscovery`.

### Proposed Solution

Kick it out of the resource storage spec. We could use `<Cluster>/<Name>/<APIGroup>/<Kind>` as a key (indexes and `apis.kcp.io/schema-storage` annotation). Let's see if that fits.

### Alternative Solutions

_No response_

### Want to contribute?

### Additional Context

_No response_

gman0:
## Summary

This PR removes `APIExport.spec.resources[].storage.virtual.identityHash` field.

Instead, we pivot on "virtual resource fingerprint", i.e. `<APIExport's identity hash>|<Endpoint slice name>/<Endpoint slice Kind>.<Endpoint slice group>`.

Since this is currently used only by CachedResource machinery (still behind `CacheAPIs` gate, AND v1alpha1), I'd lean towards not needing any conversion.

## Related Issue(s)

Fixes #4324 

## Release Notes

```release-note
BREAKING CHANGE: APIExport's resource storage `virtual` no longer has `identityHash` field
```


pkg/server/apiextensions.go
```diff
@@ -358,7 +358,7 @@ func tryDecorateCRDWithSchemaStorage(in *apiextensionsv1.CustomResourceDefinitio
 	out := shallowCopyCRDAndDeepCopyAnnotations(in)
 
 	if resourceStorage.Virtual != nil {
-		out.Annotations[apisv1alpha1.AnnotationSchemaStorageKey] = fmt.Sprintf("virtual:%s", resourceStorage.Virtual.IdentityHash)
```
mjudeikis, line 361: I think we need to fingerprint this with `logicalcluster.From(o)` As this will make unique. if 2 APIExports using same identity this will clash

staging/src/github.com/kcp-dev/sdk/apis/apis/v1alpha2/types_apiexport.go
```diff
@@ -200,13 +202,15 @@ type ResourceSchemaStorageVirtual struct {
...
-	// IdentityHash is the identity of the virtual resource.
-	IdentityHash string `json:"identityHash"`
-
 	// Resource selector TBD.
 	// We are not sure if it belongs here.
 }
 
+func (virtual *ResourceSchemaStorageVirtual) Fingerprint(o *APIExport) string {
+	gk := schema.GroupKind{Group: ptr.Deref(virtual.Reference.APIGroup, ""), Kind: virtual.Reference.Kind}
+	return fmt.Sprintf("%s|%s/%s", o.Status.IdentityHash, virtual.Reference.Name, gk)
```
mjudeikis: basically here, add LC as part of fingerprint?

pkg/server/apiextensions.go
gman0: That should be fine: the same identity means the same owner, which we assume uses the same schema (there is this assumption all over the place, e.g. [here](https://github.com/kcp-dev/kcp/blob/14d84849506417ced097a53a799aebd44ddac276/pkg/virtual/apiexport/controllers/apireconciler/apiexport_apireconciler_reconcile.go#L150-L158)).

But true, it (probably) could be based on LC + apiexport names - we are just building an index from this.

mjudeikis: The problem I can think of is:
 1. We publish a quick start
 2. people take it, iterate, and we get to the point where the same identity, same references, and the same GK get seeded into the cache server from different tenants.

pkg/server/aggregatingcrdversiondiscovery/verbs_provider.go
```diff
@@ -61,31 +63,28 @@ func (f *storageAwareResourceVerbsProviderFactory) newResourceVerbsProvider(ctx
 	if strings.HasPrefix(crd.Annotations[apisv1alpha1.AnnotationSchemaStorageKey], boundCRDVirtualStorageAnnotationPrefix) {
 		// Resources with virtual storage need the ResourceSchemaStorageVirtual from their parent APIExport.
 
-		vrIdentity := strings.TrimPrefix(crd.Annotations[apisv1alpha1.AnnotationSchemaStorageKey], boundCRDVirtualStorageAnnotationPrefix)
-		apiExportIdentity := crd.Annotations[apisv1alpha1.AnnotationAPIIdentityKey]
-
-		apiExports, err := f.getAPIExportsByVirtualResourceIdentity(vrIdentity)
+		fingerprint := strings.TrimPrefix(crd.Annotations[apisv1alpha1.AnnotationSchemaStorageKey], boundCRDVirtualStorageAnnotationPrefix)
+		apiExports, err := f.getAPIExportsByVirtualResourceFingerprint(fingerprint)
```
gman0, line 67: Just a note: with the latest index key changes, there should be exactly one export.

PR #4328, opened 2026-08-17
Add names selection to ClusterCachedResource

mjudeikis:
## Summary

A `ClusterCachedResource` can now narrow publishing to resources called exactly the names in `spec.names`, on top of the existing label selector. This is for callers that want one object in particular cached rather than a whole kind.

Also tighten virtual resource fingerprint handling: the fingerprint carries the owning APIExport, so the index no longer gates on the identity hash and discovery errors out on ambiguous matches instead of sorting for a deterministic pick.

## What Type of PR Is This?
/kind api-change

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Add names selector into ClusterCachedResource
```

### DIFF
--- a/pkg/reconciler/cache/clustercachedresources/replication/selection.go
+++ b/pkg/reconciler/cache/clustercachedresources/replication/selection.go
@@ -1,0 +1,85 @@
+/*
+Copyright 2026 The kcp Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package replication
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	cachev1alpha1 "github.com/kcp-dev/sdk/apis/cache/v1alpha1"
+)
+
+// Selection is what a ClusterCachedResource says it publishes: everything of
+// its kind, or some subset picked out by labels, by name, or by both.
+type Selection struct {
+	Labels labels.Selector
+	Names  sets.Set[string]
+}
+
+// SelectionFor reads the selection off a ClusterCachedResource. A resource that
+// sets neither publishes its whole kind.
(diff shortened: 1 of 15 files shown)


pkg/server/aggregatingcrdversiondiscovery/verbs_provider.go
```diff
@@ -71,14 +69,13 @@ func (f *storageAwareResourceVerbsProviderFactory) newResourceVerbsProvider(ctx
...
-				cmp.Compare(logicalcluster.From(a), logicalcluster.From(b)),
-				cmp.Compare(a.Name, b.Name),
-			)
-		})
+		// A fingerprint carries the owning APIExport's logical cluster and name, so it
+		// identifies exactly one APIExport. More than one match means the index is
+		// keyed by something weaker than we think, and picking either one would serve
+		// another workspace's virtual workspace URL under this one's discovery.
+		if len(apiExports) > 1 {
+			return nil, fmt.Errorf("virtual resource fingerprint %q matches %d APIExports, expected exactly one", fingerprint, len(apiExports))
```
gman0, line 77: Yeah, why not.

pkg/reconciler/cache/clustercachedresources/replication/selection.go
gman0, line 1: A unit test for this one pls.

pkg/reconciler/cache/clustercachedresources/replication/selection.go
```diff
@@ -0,0 +1,85 @@
...
+func (s Selection) ListOptions() metav1.ListOptions {
+	opts := metav1.ListOptions{}
+	if s.Labels != nil {
+		opts.LabelSelector = s.Labels.String()
+	}
+	return opts
+}
+
+// Filter drops the objects a selection does not publish.
+func (s Selection) Filter(items []unstructured.Unstructured) []unstructured.Unstructured {
```
gman0, line 74: Either rename `Filter`  to `FilterForMatchingName`, or filter `items` with `s.Matches`. Otherwise the method name is a bit lying about what it actually does.

gman0:
Add an e2e test pls as well.

pkg/reconciler/cache/clustercachedresources/replication/selection.go
```diff
@@ -0,0 +1,85 @@
...
+	if s.Names.Len() == 0 {
+		return items
+	}
+	out := items[:0]
```
gman0: Why construct the slice like this?

mjudeikis: just reusing same memory. But changed.

pkg/reconciler/cache/clustercachedresources/replication/selection.go
mjudeikis: done

pkg/reconciler/cache/clustercachedresources/replication/selection.go
mjudeikis: added

gman0:
e2e will come as a follow up

/lgtm
/approve

PR #4322, opened 2026-08-10
fix: skip DynamicRESTMapper CRD mapping until CRD is Established

pujitha24:
## Summary

Fixes a bug in the `DynamicRESTMapper`'s `DynamicTypesController` where a CRD observed before it reached the `Established` condition could have its `Status.AcceptedNames` (Kind/Singular/Plural) read while still empty, causing an incorrect/empty type mapping to be stored for that CRD. This happened because the `apis.kcp.io/CRDNoOverlappingGVR` admission plugin writes a resource-bindings lock annotation onto the `LogicalCluster` at CRD Create time, before the CRD has been reconciled to `Established`. `gatherGVKRsForCRD` now applies the same "not ready yet, skip" guard (`apiextensionshelpers.IsCRDConditionTrue(crd, Established)`) that `enqueueCRDUpdate` already uses, so a not-yet-established CRD is skipped instead of stored with garbage data.

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #3896

## Release Notes

```release-note
Fixed a bug where the `DynamicRESTMapper` could store an incorrect/empty type mapping for a CRD observed before it reached the `Established` condition.
```

---
AI assistance: this change was drafted with Claude Code.

PR #4331, opened 2026-08-18
Retry APIExport references that are not resolvable yet

mjudeikis:
## Summary

Stay with me on this.... Wasted half day on this :D  

The apiexport-reference controller resolved `storage.virtual.reference`
through the `DynamicRESTMapper` and treated a no-match as a settled answer:
it skipped the reference, returned nil, and forgot the key. Nothing
re-enqueues it -- the controller watches APIExports and
`ClusterCachedResources`, not CRDs -- so the APIExport stayed without its
`ClusterCachedResource` until something happened to touch it again.
    
That mapper is per-replica in-memory state, fed asynchronously by its own
controller, and since #4322 (fixing another issue) it withholds a CRD until it is Established.
A no-match therefore means "this replica does not know that kind yet",
which is true of every kind for a few seconds after it is created. An
APIExport applied in the same breath as the CRD it references loses,
permanently: no `ClusterCachedResource`, so the referenced object never
reaches the cache server, so every shard fails discovery of the virtual
resource with "failed to retrieve virtual workspace URL" from then on.
    
`wantedFor` now reports unresolved references instead of dropping them, and
reconcile returns an error for them so the key requeues. The rate limiter
starts at milliseconds, so the bootstrap case converges about as fast as
an event would have delivered it, while a reference to a kind that really
does not exist settles into a slow retry -- and says so, where before it
left no trace anywhere.
    
Pruning is now skipped while any reference is unresolved. The wanted set
is partial in that state, and deleting against it would drop a
`ClusterCachedResource` the APIExport still asks for, draining and
refilling the cache for a kind that never stopped being referenced.
References that did resolve are still applied, so one unsettled reference
not hold up the rest.

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Retry APIExport references that are not resolvable yet
```

### DIFF
--- a/pkg/reconciler/cache/apiexportreference/apiexportreference_controller.go
+++ b/pkg/reconciler/cache/apiexportreference/apiexportreference_controller.go
@@ -329,12 +329,6 @@ func (c *controller) reconcile(ctx context.Context, rawKey string) error {
 	// Anything this APIExport owns but no longer asks for goes. Deleting the
 	// ClusterCachedResource is what removes the copies: its finalizer drains
 	// them while the kind is still served.
-	//
-	// Only when every reference resolved, though. With one still unresolved the
-	// wanted set is a partial answer, and pruning against it would delete a
-	// ClusterCachedResource the APIExport does in fact still ask for -- then
-	// recreate it on the next pass, draining and refilling the cache for a kind
-	// that never stopped being referenced.
 	if len(unresolved) == 0 {
 		for _, ccr := range existing {
 			if _, keep := wanted[ccr.Name]; keep {
@@ -355,10 +349,6 @@ func (c *controller) reconcile(ctx context.Context, rawKey string) error {
 	// fields this controller owns, so an existing ClusterCachedResource keeps
 	// whatever others put on it -- the clustercachedresources controller
 	// defaults spec.identity in place, and that must survive this write.
-	//
-	// References that did resolve are applied even while others have not: an
-	// APIExport that mixes a settled reference with a brand-new one should not
-	// have the settled one held hostage.
 	for _, want := range wanted {
 		if err := c.applyResource(ctx, cluster, want); err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


pkg/reconciler/cache/apiexportreference/apiexportreference_controller.go
```diff
@@ -324,55 +329,85 @@ func (c *controller) reconcile(ctx context.Context, rawKey string) error {
...
-			"clusterCachedResource", ccr.Name)
-		if err := c.deleteResource(ctx, cluster, ccr.Name); err != nil && !apierrors.IsNotFound(err) {
-			return err
+	//
+	// Only when every reference resolved, though. With one still unresolved the
+	// wanted set is a partial answer, and pruning against it would delete a
+	// ClusterCachedResource the APIExport does in fact still ask for -- then
+	// recreate it on the next pass, draining and refilling the cache for a kind
+	// that never stopped being referenced.
+	if len(unresolved) == 0 {
```
mjudeikis, line 332: Lack of this was basically deleting cachedResources and so making this not working due to rest mapper record being absent

pkg/reconciler/cache/apiexportreference/apiexportreference_controller.go
```diff
@@ -324,55 +329,85 @@ func (c *controller) reconcile(ctx context.Context, rawKey string) error {
...
-		klog.FromContext(ctx).V(2).Info("dropping a ClusterCachedResource nothing references anymore",
-			"clusterCachedResource", ccr.Name)
-		if err := c.deleteResource(ctx, cluster, ccr.Name); err != nil && !apierrors.IsNotFound(err) {
-			return err
+	//
+	// Only when every reference resolved, though. With one still unresolved the
+	// wanted set is a partial answer, and pruning against it would delete a
+	// ClusterCachedResource the APIExport does in fact still ask for -- then
+	// recreate it on the next pass, draining and refilling the cache for a kind
+	// that never stopped being referenced.
```
gman0: The worst AI-generated comment I've seen today so far 😱 Can you simplify please? The code itself is pretty self-explanatory actually.

pkg/reconciler/cache/apiexportreference/apiexportreference_controller.go
```diff
@@ -324,55 +329,85 @@ func (c *controller) reconcile(ctx context.Context, rawKey string) error {
...
+	if len(unresolved) == 0 {
+		for _, ccr := range existing {
+			if _, keep := wanted[ccr.Name]; keep {
+				continue
+			}
+			if !ccr.DeletionTimestamp.IsZero() {
+				continue
+			}
```
gman0, line 339: I'm wondering if we could move this `!ccr.DeletionTimestamp.IsZero()` up? If the export is in Deleting phase, the CCRs would in any case be GC'd, no?

gman0:
In general way too many comments for handling a simple retry-until-exists :D

pkg/reconciler/cache/apiexportreference/apiexportreference_controller.go
gman0: Ah I see, this needs to be deleted eagerly when the export changes... sorry, this can stay as-is, then.

PR #4333, opened 2026-08-19
pkg/server: disable the openapi endpoint for system:bound-crds

### LINKED ISSUE #4332: bug: system:bound-crds's openapi/v3 fails on duplicate bound GVK within a shard
### Describe the bug

if a shard has some providers exporting the same GVK (actual schema doesn't matter) and there are consumers bound to them, these GVKs are 'indexed' in the system:bound-crds special WS. that works fine until someone tries do an openapi/v3 request on the WS. since it contains the same GVK multiple times the extension api server "merger" just errors out.

> {"ts":"2026-08-18T10:43:47.430Z","caller":"handler3/handler.go:214","msg":"Error serving discovery: failed to merge specs: OpenAPI V3 merge schema conflict on somegroup.v1.SomeKind"}

https://github.com/kcp-dev/kubernetes/blob/e006560fc76a/staging/src/k8s.io/apiextensions-apiserver/pkg/controller/openapi/builder/merge.go#L142

instead of trying to patch upstream, it makes more sense to me to just stop serving openapi from this WS.

it appears that if a MutatingAdmissionPolicies exists somewhere on the shard it will result on an umbrella openapi/v3 request on all WS even the system ones. i don't think that's necessarily desired by design. maybe at least we can skip it on this particular system WS. 
 
https://github.com/kcp-dev/kubernetes/blob/e006560fc76a/staging/src/k8s.io/apiserver/pkg/admission/plugin/policy/mutating/patch/typeconverter.go#L80-L85

https://github.com/kcp-dev/kcp/blob/04fcc9232/pkg/admission/mutatingadmissionpolicy/mutating_admission_policy.go#L408

### Steps To Reproduce

- create multiple provider WS and export the same GVK from them
- create multiple consumer WS and bind to the same exported GVK
- observe the errors in the logs every 5 seconds as long as you have a MAP somewhere in the shard

### Expected Behaviour

don't show openapi GVK merge conflicts from a system WS that is likely not designed to serve openapi

### Additional Context

NONE

neolit123:
## Summary

This system WS holds all bound CRDs across the shard, so merging their openapi fails on duplicate GVKs. The schema diff doens't matter as the upstream ext apisever "merger" just errors on matching GVK keys. Serve an empty openapi doc from this WS to avoid merge errors.

Also in MAP admission, skip this WS so no callers even trigger an openapi fetch on this WS. It seems MAP admission is the only internal kcp mechanism that does it (?). External admin users could have a use case for this but if there are duplicates they will just get an error and no actual merged schemas.

If you think there are better ways to solve this, LMK, i didn't want to patch upstream to relax the merge

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #4332

## Release Notes

```release-note
bug: stop serving OpenAPI v3 from the `system:bound-crds` workspace as it may contain duplicated GVKs sourced from different consumers.
```

### DIFF
--- a/pkg/admission/mutatingadmissionpolicy/mutating_admission_policy.go
+++ b/pkg/admission/mutatingadmissionpolicy/mutating_admission_policy.go
@@ -61,6 +61,8 @@ import (
 
 const PluginName = "KCPMutatingAdmissionPolicy"
 
+const systemBoundCRDsClusterName = logicalcluster.Name("system:bound-crds")
+
 func Register(plugins *admission.Plugins) {
 	plugins.Register(PluginName,
 		func(config io.Reader) (admission.Interface, error) {
@@ -205,6 +207,10 @@ func (k *KubeMutatingAdmissionPolicy) Admit(ctx context.Context, a admission.Att
 	cluster, err := genericapirequest.ValidClusterFrom(ctx)
 	if err != nil {
 		return err
+	}
+
+	if cluster.Name == systemBoundCRDsClusterName {
+		return nil
 	}
 
 	sourceCluster, err := k.getSourceClusterForGroupResource(cluster.Name, a.GetResource().GroupResource())
--- a/pkg/server/openapiv3/servicecache.go
+++ b/pkg/server/openapiv3/servicecache.go
@@ -56,6 +56,8 @@ const (
 
 	// TODO(sttts): move to central place in kube.
 	BoundAnnotationKey = "apis.kcp.io/bound-crd"
+
+	systemBoundCRDsClusterName = logicalcluster.Name("system:bound-crds")
 )
 
 // WithOpenAPIv3 returns a handler that serves OpenAPI v3 specs for CRDs, not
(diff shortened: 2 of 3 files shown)


gman0:
Hmm, VAP doesn't trigger that as well?

gman0:
But other than that, yeah makes sense. Thanks!

Can't approve but, at at least:

/lgtm

neolit123:
> Hmm, VAP doesn't trigger that as well?

i couldn't find other references. on the ticket i mentioned only the one single place (MAP) that i found.

ntnn (approved):
/approve

neolit123:
thanks for the reviews!

FTR, i think that this owners file might be a bit wrong to flag the admission package as `kind/api-change`
https://github.com/kcp-dev/kcp/blob/main/pkg/admission/OWNERS

neolit123:
> thanks for the reviews!
> 
> FTR, i think that this owners file might be a bit wrong to flag the admission package as `kind/api-change` https://github.com/kcp-dev/kcp/blob/main/pkg/admission/OWNERS

/assign @mjudeikis

mjudeikis:
/approve 
admission is api :)

Issue #4336, opened 2026-08-21
epic: Root shard de-sharding


mjudeikis:
### Demo Objective

Goal of this epic is decouple root shard as failure point in architecture. 

1. Decouple where shard objects are stored https://github.com/kcp-dev/kcp/issues/4335
2. Add quotas and limits for shards https://github.com/kcp-dev/kcp/issues/3472
3. Add heartbeat to the shards so if shards goes offline - we mark them as `NotReady` so we stop scheduling to it. Like kubelet does with nodes. Ticket - TBD. 
4. Add object pruning when shard objects is deleted - ticket TBD
5. Add shard aware scheduling https://github.com/kcp-dev/kcp/issues/4256
6. Implement identity rotation from KEP-5 https://github.com/kcp-dev/enhancements/blob/c62b9141c1b49df82b0be1d519ec74a11a3a7996/keps/core/0005-apiexport-identity-rotation.md so we can do 7
7. Decouple identities from root shard https://github.com/kcp-dev/kcp/issues/4347 

### Action Items

- [x] Scope of the current demo is necessary to fit in the prototype boundaries
- [x] Contribute to the final demo script and recording

PR #4337, opened 2026-08-21
WIP: Move shards object ownership to shards

mjudeikis:
## Summary

This PR moves shard object authoritative data into the shards themselves. And the root shard only sees a shallow copy.
This is part of a bigger effort to reduce root shard dependency https://github.com/kcp-dev/kcp/issues/4336

Not yet sure if this gonna work in the form it is now :) 

IMPORTANT: This removed ability to make shards `unschedulable` via annotation. Need to see if this still right place for it. Because shards are "owning their own objects", we need two way communication for this. 

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Move shard authoritive data into shards itself
```

### DIFF
--- a/cmd/sharded-test-server/main.go
+++ b/cmd/sharded-test-server/main.go
@@ -26,12 +26,16 @@ import (
 	"strings"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	kuser "k8s.io/apiserver/pkg/authentication/user"
 	genericapiserver "k8s.io/apiserver/pkg/server"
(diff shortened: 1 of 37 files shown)


mjudeikis:
Pending item is:
How we model operations now - scheduling, not scheduling, limits in this model

I think this need small write-up before we process further.

gman0:
Just FYI, I'm already using this in cache cross-sync work :D 

If there are multiple caches, the shards who belong to the non-root cache join the system (i.e. appear on root shard) only once the non-root cache syncs back to the root cache.

`shard_gamma` -> `cache_1` -> (cache cross-sync) -> `cache_root` -> `shard_root`

mjudeikis:
closing in favour of https://github.com/kcp-dev/kcp/pull/4349

PR #4345, opened 2026-08-27
hack: don't let artifact tar warnings fail otherwise-green CI runs

### LINKED ISSUE #4343: flake: `tar: kcp/audit.log: file changed as we read it`
### Describe the bug

Sometime the post-steps in CI cause a failure, probably because not all processes are finished before CI starts packing up:

```text
 Command terminated with 0
```

https://public-prow.kcp.k8c.io/view/s3/prow-public-data/pr-logs/pull/kcp-dev_kcp/4342/pull-kcp-test-e2e-shared/2092889102822150144

### Steps To Reproduce

-

### Expected Behaviour

-

### Additional Context

_No response_

pujitha24:
## Summary

CI post-steps in `hack/run-with-prow.sh` sometimes fail while compressing
artifacts (`tar: kcp/audit.log: file changed as we read it`), even though the
test run itself passed. This happens because `cmd/test-server`'s `start()`
returned as soon as it saw the shutdown signal, relying on a deferred
`cancel()` to stop the shard asynchronously — so the wrapping process (and
the artifact-compression step) could exit before the shard had actually
terminated and flushed its log/audit files.

This PR fixes that at the root: `start()` now explicitly cancels the shard
and blocks on its termination channel before returning, so the process is
confirmed exited first. An earlier version of this PR worked around the
symptom in `hack/run-with-prow.sh` instead; that's been reverted in favor of
this fix.

The equivalent change for `cmd/sharded-test-server` is bigger (it needs
`startFrontProxy` to expose its own termination channel) and is left for a
follow-up PR.

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #4343

## Release Notes

```release-note
NONE
```

### DIFF
--- a/cmd/test-server/main.go
+++ b/cmd/test-server/main.go
@@ -150,5 +150,13 @@ func start(shardFlags []string, workDirPath, logDirPath string, quiet bool) erro
 
 	s.GatherMetrics(metricsCtx)
 
+	// Stop the shard and wait for its process to actually exit before we
+	// return, instead of relying on the deferred cancel() above, which
+	// only asynchronously triggers termination. Without this, main() can
+	// return - and the process running it can exit - while the shard is
+	// still shutting down and flushing its log/audit files.
+	cancel()
+	<-errCh
+
 	return nil
 }
--- a/hack/run-with-prow.sh
+++ b/hack/run-with-prow.sh
@@ -34,7 +34,18 @@ echo "Command terminated with ${EXIT_CODE}"
 
 echo 'Compressing build artifacts...'
 cd "$ARTIFACTS"
+set +o errexit
 tar cjf artifacts.tar.bz2 --remove-files *
+TAR_EXIT_CODE=${?}
+set -o errexit
+if [[ "${TAR_EXIT_CODE}" -ne 0 ]]; then
+	# Processes started by the test run (e.g. kcp servers) can still be
+	# shutting down and writing to their log/audit files when we get here,
+	# which makes tar report a non-fatal "file changed as we read it"
+	# warning (or a failed rmdir of the directory it lives in) as a fatal
+	# error. Don't let that mask the actual test result.
+	echo "warning: compressing artifacts exited with code ${TAR_EXIT_CODE}, some files may still be present uncompressed in ${ARTIFACTS}"
+fi
 echo 'Done compressing files.'
 
 exit "${EXIT_CODE}"
--- a/hack/run-with-prow.sh
+++ b/hack/run-with-prow.sh
@@ -34,18 +34,7 @@ echo "Command terminated with ${EXIT_CODE}"
 
 echo 'Compressing build artifacts...'
 cd "$ARTIFACTS"
-set +o errexit
 tar cjf artifacts.tar.bz2 --remove-files *
-TAR_EXIT_CODE=${?}
-set -o errexit
-if [[ "${TAR_EXIT_CODE}" -ne 0 ]]; then
-	# Processes started by the test run (e.g. kcp servers) can still be
-	# shutting down and writing to their log/audit files when we get here,
-	# which makes tar report a non-fatal "file changed as we read it"
-	# warning (or a failed rmdir of the directory it lives in) as a fatal
-	# error. Don't let that mask the actual test result.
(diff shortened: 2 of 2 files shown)


gman0:
Why can't we just wait for all processes to exit?

pujitha24:
Good question, so I actually went and checked. The command run-with-prow.sh runs directly (`"${@}"`, plain `go test`) already waits properly for its own kcp servers: `sdk/testing/server/fixture.go` sends SIGTERM and blocks on `cmd.Wait()` before the test binary returns, so those are fully reaped by the time this script gets to the tar step.

The race comes from the shared/sharded flows (`test-e2e-shared-minimal` etc.). There, kcp/cache-server/front-proxy are children of `cmd/test-server/main.go` and `cmd/sharded-test-server/main.go`, and those `start()` functions cancel the child processes and return right after `GatherMetrics`, without ever waiting on the `terminatedCh`/error channels the shard already exposes. So the wrapper binary (and the Makefile's `wait $PID`) can exit before the kcp process it just killed has actually finished dying and flushing audit.log — this script has no handle on those grandchild processes to wait on itself.

So "wait for all processes to exit" is the right fix, but it belongs in those two `main()`s, not here. That's a real change to shutdown ordering across two binaries and I don't want to guess at timeouts/ordering without being able to validate it against real CI. Want me to take a shot at it as a follow-up, or would you rather fold it into this PR?

mjudeikis:
I would like to see how this could looks like in the main go code. How big is the change

pujitha24:
Looked into this concretely. For `cmd/test-server/main.go` it's small — `start()` already has `errCh` from `WaitForReady`, it just never blocks on it after `GatherMetrics`:

```go
s.GatherMetrics(metricsCtx)

cancel()   // stop the shard
<-errCh    // wait for it to actually exit before we return
return nil
```

That's ~4 lines.

`cmd/sharded-test-server/main.go` is bigger. Shards, the cache server, and virtual workspaces already expose a `terminatedCh` (currently only consumed to detect *premature* failure, not reused at shutdown), so those just need to be waited on again after `shutdownCtx.Done()`. But `startFrontProxy` in `frontproxy.go` creates its `terminatedCh` locally and only uses it internally for the readiness-wait loop — it never returns it. To wait for front-proxy on shutdown too, `startFrontProxy`'s signature has to change to return that channel, plus the call site in `main.go`. So realistically this touches `main.go`, `frontproxy.go`, and the shutdown `select` block — rough guess ~30-50 lines across those files, plus a real CI run to make sure it doesn't introduce a hang if a process doesn't exit cleanly.

I haven't written that patch yet since it's a shutdown-ordering change I want to validate against real CI runs rather than guess at, so I'd rather keep it out of this PR (which is just the tar-warning wrapper) and open it separately. Happy to go do that now if that's preferred over merging this as-is.

mjudeikis:
So can we do this better? :) over trying to hack bash?

pujitha24:
Yeah, fair — the tar wrapper is a band-aid, not a fix. I'd rather land the real thing than keep this hack around.

Of the two pieces I scoped out above, the `cmd/test-server/main.go` side is small and I'm confident in it (call `cancel()` explicitly, then block on the `errCh` from `WaitForReady` before returning, instead of letting the deferred `cancel()` race with process exit). The `cmd/sharded-test-server` side is bigger — it needs `startFrontProxy` to return its `terminatedCh` too — and I don't want to merge shutdown-ordering changes there without a real CI run backing it up.

Want me to swap this PR for the `test-server/main.go` fix now and follow up with the sharded-test-server side once I can validate it, or would you rather I close this and open the real fix as a fresh PR?

mjudeikis:
First, I would like you to respond to me in your own words and not with an AI proxy :) I think some human contact would still be great in this age of AI :) If I wanted AI to do these nits, I could do it on my own. There is a reason we prefer human contact :) 

Regards the request - repurpose same pr please

pujitha24:
Fair, sorry about that — I'll keep it plain. And yeah, already done on the repurpose: this PR is now just the cmd/test-server/main.go fix, cancel() then wait on errCh before start() returns, and I pulled the hack/run-with-prow.sh workaround back out in the same commit. So it's the real fix now, not the band-aid. Sharded-test-server is still a separate follow-up since I want a CI run behind that one before touching its shutdown ordering.

PR #4349, opened 2026-09-01
Add global admin virtualWorkspace

mjudeikis:
## Summary

This is different version of https://github.com/kcp-dev/kcp/pull/4337
This is part of a bigger effort to reduce root shard dependency https://github.com/kcp-dev/kcp/issues/4336

In general, this is the biggest part of the work to remove root shard as special. Most of the responsibilities are moving into the cache server as a storage/persistent layer and are served by the shards themselves. 

 In this pr:
1. add new special "fake" virtual workspace `:admin` where kcp-admin can edit and do things at the platform level. It gets details to cache server via every shard VirtualWorkspace and so propagates data into every shard
2. Moves shard registration into this per-shard space
3. Makes `:root:Shards` just a view of shards, but it's temporary. We will deprecate this.
4. Add `--peer-shard-kubeconfig` into frontproxy, so FP knows where to find the initial list of "peers". 

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Wire in `:admin` workspace
Move shard registration to be "per shard" and away from root shard.
```


docs/content/concepts/sharding/shards.md
```diff
@@ -69,6 +69,52 @@ right scalability and availability properties.
...
+## The Admin Workspace (`:admin`)
+
+Every shard serves the *Admin workspace* at `/services/admin`: an
+installation-wide administrative view, reachable identically through any
+shard. Its first resource is an aggregated view of all `Shard` objects. The
+view is backed by the cache server, where every shard's `Shard` object is
+replicated, so all shards serve an identical view in a single
+`resourceVersion` space: consumers can fail over between shards and resume
+watches with the same resource version. Over time the Admin workspace is the
+home for further installation-wide admin surfaces.
```
gman0: This is only true if all shards belong to the same cache. If the client fails-over to a shard on a different cache, the RV space is different.

pkg/proxy/index/index_controller.go
```diff
@@ -85,32 +85,43 @@ func NewController(
...
-
-			c.stopShard(shard.Name)
+			c.onShardDelete(obj.(*corev1alpha1.Shard))
 		},
 	})
 
 	return c
 }
 
+func (c *Controller) onShardAdd(ctx context.Context, shard *corev1alpha1.Shard) {
```
mjudeikis, line 104: This is just cleanup ^

test/e2e/apibinding/apibinding_permissionclaims_sharded_test.go
```diff
@@ -77,7 +77,7 @@ func TestAPIBindingPermissionClaimsAppliedAcrossShards(t *testing.T) {
 	require.NoError(t, err, "failed to construct dynamic cluster client")
 
 	t.Logf("Listing shards")
-	shards, err := kcpClusterClient.Cluster(core.RootCluster.Path()).CoreV1alpha1().Shards().List(t.Context(), metav1.ListOptions{})
+	shards, err := kcptesting.ListShards(t.Context(), cfg)
```
mjudeikis, line 80: This is change in all test, changing wher we list authoritative shard list

mjudeikis:
TODO: I think this breaks on upgrade. We need upgrade test and this need to be fixed

PR #4359, opened 2026-09-10
Fix slate LC routing

mjudeikis:
## Summary

During cluster migration we have bit of situation that front-proxy gets double events for same logical cluster on different shards. We need to explistlly handle copy object and to know where to route things :) 

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Better handle LC migration routing in front-proxy
```


ntnn:
MHmm... A cluster that is being migrated shouldn't be handled by front-proxy at all until the migration is done.

/hold

mjudeikis:
Im on the line... There is still routing happening as shard  responds with "its not available"... And the problem is that we now get 2 events from 2 different shards for same LC.

mjudeikis:
I think informers will still emit events? Movign this conv to slack.

mjudeikis:
/retest
flake https://github.com/kcp-dev/kcp/pull/4362

PR #4361, opened 2026-09-11
[1/5] Admin workspace 

mjudeikis:
## Summary

Make shard lookups root-agnostic, add Schedulable condition and --shard-labels

Preparation for serving Shards from outside the root workspace:

- `indexers.ShardByName` looks a Shard up by name across all logical clusters of a lister (preferring the shard-owned `system:shard` copy) and replaces the hard-coded root-cluster lookups in the workspace scheduler, the endpoint-slice URL controllers, migration data copy and the partition selector.
- The shard controller sets a `Schedulable` condition mirroring the unschedulable (cordon) annotation, so applying a cordon is acknowledged on the object. Printed as a column on Shards.
- kcp start gains `--shard-labels`; labels are part of the shard's own configuration and applied on registration. The sharded-test-server passes region labels this way instead of patching Shards in root.

**ALL 5 prs are based on top of each other. So I will address reviews in follow-ups with expcetion of 5th :)** 

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Add --shard-labels to shard flags
Add condition `Schedulable` to shards
```


pkg/server/server.go
```diff
@@ -623,13 +623,17 @@ func (s *Server) Run(ctx context.Context) error {
 		s.CacheKcpSharedInformerFactory.WaitForCacheSync(hookCtx.Done())
 
 		// create or update shard
+		labels := map[string]string{
+			"name": s.Options.Extra.ShardName,
+		}
+		for k, v := range s.Options.Extra.ShardLabels {
+			labels[k] = v
+		}
```
ntnn, line 631: We should probably hard override the "name" label.

ntnn (commented):
Otherwise ok I think

/lgtm
/approve

PR #4363, opened 2026-09-11
Re-check map entry under write lock

ntnn:
## Summary

1. Re-check lc->shard entries under write lock
2. Handle flake in migration test

## What Type of PR Is This?

/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```

### DIFF
--- a/pkg/index/index.go
+++ b/pkg/index/index.go
@@ -257,6 +257,12 @@ func (c *State) UpsertLogicalCluster(shard string, logicalCluster *corev1alpha1.
 		c.lock.Lock()
 		defer c.lock.Unlock()
 
+		// Re-read under write lock, a concurrent upsert may have won the race.
+		got = c.clusterShards[clusterName]
+		if got == shard {
+			return
+		}
+
 		// If got is not empty then the logical cluster was migrated from shard `got` to shard `shard`.
 		// Record the timestamp and delete the context from the manager.
 		// The timestamp is recorded so clients with a watch are getting a 410 sent back to trigger a full relist.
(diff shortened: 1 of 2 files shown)


ntnn:

```text
 === FAIL: test/e2e/garbagecollector TestGarbageCollectorCRDDeletionCascading (43.71s) 
 ```

File: pkg/index/index.go [hunk 1]
```diff
@@ -253,29 +253,76 @@ func (c *State) UpsertLogicalCluster(shard string, logicalCluster *corev1alpha1.
...
-			c.clusterContexts.Delete(clusterName, fmt.Errorf("logical cluster %s migrated from shard %s to shard %s", clusterName, got, shard))
-		}
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	// Re-read under write lock, a concurrent upsert may have won the race.
+	got = c.clusterShards[clusterName]
+	if got == shard {
+		return
+	}
```
mjudeikis, line 267: I think we need one more gate here? 
```
	if got != "" && logicalCluster.Annotations["internal.kcp.io/migrating"] != "" {
	    return
	}
```

ntnn: Wouldn't hurt, true

ntnn: I'm thinking FP should ignore any event of an LC with the migrating annotation set except deletion - as long as the LC is migrating it shouldn't be updated in the FP

ntnn:
/retest

```text
tar: kcp/audit-1.log: file changed as we read it
```

ntnn:
/retest

GC and workspacetype flakes

mjudeikis:
/lgtm
/approve 

its basically different shape of my pr :D

PR #4366, opened 2026-09-13
Add recover flow for workspace.

mjudeikis:
## Summary

Fixing some workspace conditioning. Added big comment why we need this... Bit hard to find this in production. But when it happens its ugly. 

## What Type of PR Is This?
/kind bug

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


mjudeikis:
/retest

informer flake?

ntnn (approved):
/lgtm
/approve

I think I hit this a few times but never checked the root cause.
I think that could also be the cause for the "wait until APIs are served" in our kcptesting suite?

mjudeikis:
> I think I hit this a few times but never checked the root cause. I think that could also be the cause for the "wait until APIs are served" in our kcptesting suite?

Could be. We merge and see :D

PR #4367, opened 2026-09-14
Prevent returning APIBindings before its backing informer is ready

mjudeikis:
## Summary

I noticed some flakes where permissionClaim-backed resources are not available. I think this might be a race condition here between the informer, backing GVRs of the permissionClaims and the reconciler itself. 

## What Type of PR Is This?
/kind flake

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


ntnn (approved):
/lgtm
/approve

PR #4368, opened 2026-09-14
fix: avoid panic on unauthenticated user when checking for system user

MuneebAijaz:
## What Type of PR Is This?

Fixed a panic in the per-workspace authentication filter that could occur when a delegated authenticator returned no user.

Error logs from pod 

```
  "apiserver panic'd" URI="/clusters/root:test-abc/openapi/v3" auditID=""
  runtime error: invalid memory address or nil pointer dereference
  kcp/pkg/authentication.(*eagerIndex).Lookup.wrapWithSecurityFilters.ForbidSystemUsernames.func1
    /workspace/pkg/authentication/usernames.go:37
    <- GroupFilter -> ExtraFilter -> withClusterScope (authenticators.go:78)
```

## Related Issue(s)

Fixes #

## Release Notes

```release-note
NONE
```


pkg/authentication/usernames.go
```diff
@@ -33,10 +33,9 @@ import (
 func ForbidSystemUsernames(delegate authenticator.Request) authenticator.Request {
 	return authenticator.RequestFunc(func(req *http.Request) (*authenticator.Response, bool, error) {
 		result, authenticated, err := delegate.AuthenticateRequest(req)
-		if err == nil {
```
mjudeikis, line 36: I think this might need bit wider extension. This will allow strange cases where user could be authenticated but no user set (your scenario, you hit)

```
func ForbidSystemUsernames(delegate authenticator.Request) authenticator.Request {
	return authenticator.RequestFunc(func(req *http.Request) (*authenticator.Response, bool, error) {
		result, authenticated, err := delegate.AuthenticateRequest(req)
		if err != nil || !authenticated {
			return result, authenticated, err
		}

		// An authenticated result must carry a user.
		if result == nil || result.User == nil {
			return nil, false, errors.New("authenticated request without user")
		}

		if strings.HasPrefix(result.User.GetName(), "system:") {
			return nil, false, errors.New("system usernames are not admitted")
		}

		return result, authenticated, err
	})
}

```

pkg/authentication/usernames_test.go
```diff
@@ -0,0 +1,193 @@
...
+			case tc.wantName != "":
+				require.NotNil(t, resp)
+				require.NotNil(t, resp.User)
+				require.Equal(t, tc.wantName, resp.User.GetName())
+			case tc.response != nil && tc.response.User == nil && tc.err == nil:
+				// The delegate's user-less response is forwarded as-is.
+				require.Equal(t, tc.response, resp)
+			default:
+				require.Nil(t, resp)
+			}
```
mjudeikis, line 173:
```
if tc.wantName != "" {
				require.NotNil(t, resp)
				require.NotNil(t, resp.User)
				require.Equal(t, tc.wantName, resp.User.GetName())
			} else {
				require.Nil(t, resp)
			}
```
basically somth like this

mjudeikis:
/retest
CI bug?

PR #4371, opened 2026-09-16
[3/5] Admin workspace. Cordon shards through the Admin workspace

mjudeikis:
## Summary

The Admin workspace becomes writable for exactly one purpose: the allow-listed operational annotations (currently the unschedulable annotation used for cordoning). A write is validated against the `allow-list`, applied to the cache copy of the target shard, and the replication controller applies cache-owned annotations from the cache to the authoritative local object, from where the shard's controller picks it up and sets the Schedulable condition. No component connects to another shard directly, and intent for an unavailable shard parks in the cache until it reconnects.

Direct create/update/delete of Shard objects by non-system users is now denied by admission with a pointer at the Admin workspace: shards register themselves and own their objects.

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Allow cordoning of shards via admin virtual workspace 
```


ntnn (approved):
/lgtm
/approve

PR #4373, opened 2026-09-16
[4/5] Admin workspace. Discover shards in the front-proxy through the Admin workspace

mjudeikis:
## Summary

The front-proxy no longer watches Shards in the root workspace. It reads them through `/services/admin` of its peer shards: `--shard-peer-kubeconfig` (repeatable) seeds the peer set, falling back to the `--root-kubeconfig` server, and every shard observed through the view joins the failover set at runtime while decommissioned shards leave it. Requests are distributed round-robin with per-request failover and a cooldown for failed peers; since all peers serve the identical cache-backed view, a broken watch resumes against another peer with the same resourceVersion.

`--root-kubeconfig` is deprecated: it remains only for APIExport identity resolution until identities are resolvable without the root shard.

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
add --shard-peer-kubeconfig and make front-proxy to auto-discover shard using admin virtual workspace
```


pkg/proxy/index/index_controller_test.go
```diff
@@ -0,0 +1,118 @@
...
+	// restart triggered by the update; simulate that.
+	c.state.UpsertLogicalCluster("alpha", logicalClusterObj("myws"))
+
+	result, found = c.state.LookupURL(logicalcluster.NewPath("myws"))
+	if !found {
+		t.Fatal("shard base URL was lost after a baseURL change; workspaces on the shard are unroutable")
+	}
+	if result.URL != "https://new/clusters/myws" {
+		t.Errorf("expected the new base URL to be used, got %q", result.URL)
+	}
```
ntnn: aren't we using require/assert?

pkg/proxy/options/options.go
```diff
@@ -59,7 +66,11 @@ func (o *Options) AddFlags(fss *cliflag.NamedFlagSets) {
 	fs.StringVar(&o.MappingFile, "mapping-file", o.MappingFile, "Config file mapping paths to backends")
 	fs.StringVar(&o.RootDirectory, "root-directory", o.RootDirectory, "Root directory.")
 	fs.StringVar(&o.RootKubeconfig, "root-kubeconfig", o.RootKubeconfig, "The path to the kubeconfig of the root shard.")
+	if err := fs.MarkDeprecated("root-kubeconfig", "use --shard-peer-kubeconfig for shard discovery; --root-kubeconfig is still required for APIExport identity resolution and will be removed once identities are resolvable without the root shard."); err != nil {
```
ntnn, line 69: Mhm... I think we should only start marking --root-kubeconfig as deprecated once it actually is deprecated.

pkg/proxy/peers.go
```diff
@@ -0,0 +1,254 @@
...
+	defer p.mu.Unlock()
+	for _, seed := range p.seeds {
+		if seed.String() == u.String() {
+			delete(p.dynamic, name)
+			return nil
+		}
+	}
+	p.dynamic[name] = u
+	return nil
+}
```
ntnn, line 95: This reads weird. I'd expect that during startup we try getting the current ShardList from any of the seeds (retrying with another one until one usccess) and use that to fill dynamic, so seeds as a member becomes obsolete. Also what if a Shard's URL genuinely changes but the front-proxy still has the old URI in the seed?

ntnn (changes requested):

pkg/proxy/index/index_controller_test.go
mjudeikis: updated. got distracted :D

pkg/proxy/options/options.go
mjudeikis: I think this is fine? It still works; we have a new way of doing things. And we gonna keep it alive for at least 2 releases.

pkg/proxy/peers.go
mjudeikis: reworked this, yes this was/is bit odd

ntnn (approved):
/lgtm
/approve

The pick peer mechanism is nice!

PR #4374, opened 2026-09-17
[5/5] Admin workspace. Move shard registration out of the root workspace

mjudeikis:
## Summary

Shards now register their `Shard` object in the shard-local `system:shard`
logical cluster instead of the root workspace, so shard startup no longer
depends on the root shard. The replication controller lets `system:shard`
through to the cache server, and the Admin workspace serves the
shard-owned copies. The shard status controller runs on every shard and
maintains only its own authoritative object.
    
For a familiar UX the shard hosting the root workspace runs a
`shardrepresentation` controller mirroring the cache server's Shard
objects into root as read-only representations. They are marked with an
annotation and excluded from replication back to the cache, so global
informers see every shard exactly once.
    
During mixed-version windows a shard exists twice in the cache (legacy
root copy and shard-owned copy). The Admin workspace collapses both on
the list and the watch path: the shard-owned copy wins, events for a
losing copy are translated into the winner's state, and DELETED is only
emitted once no copy is left, so consumers keyed by name (e.g. the
front-proxy index) never see a shard vanish while it is alive.

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Move shard registration out of the root workspace
```


mjudeikis:
/retest
Gc flake. will make logging more verbose as its not clear from artifacts

pkg/reconciler/cache/replication/replication_controller.go
```diff
@@ -185,6 +191,28 @@ func IsNoSystemClusterName(obj interface{}) bool {
 	return true
 }
 
+// isNoSystemClusterNameExceptSystemShard is like IsNoSystemClusterName, but
+// additionally lets objects in the shard-local system:shard logical cluster
+// through. Used for resources whose authoritative copy is shard-owned, like
+// Shards.
+func isNoSystemClusterNameExceptSystemShard(obj interface{}) bool {
```
ntnn, line 198: This can just delegate to `IsNoSystemClusterName` after checking `system:shard`, or?

ntnn (approved):
/lgtm
/approve

pkg/reconciler/cache/replication/replication_controller.go
mjudeikis: Code does not feel better with this change

PR #4376, opened 2026-09-17
1/3 APIExport identity rotation. Retry permission claim labeling on a fixed delay while informers sync

mjudeikis:
## Summary

A binding whose claimed resources have no synced dynamic informer yet is not failing: the informer just needs a moment to list. Retry it after a fixed delay instead of counting a failure, so bursts of unrelated updates do not stack per-item exponential backoff and park the binding for up to the 1000s cap after the informer has synced.

This becomes super flaky once identity migration starts happening. 

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Add controller retry on apibinding resource dynamic informer
```


xrstf (approved):
/approve

PR #4378, opened 2026-09-18
Shard soft/hard limits for workspaces

mjudeikis:
## Summary

Add soft/hard limits to shards and make shards to be aware of these limits.

Workspace when there is no capacity:
```
kind: Workspace
...
spec:
  type:
    name: universal
    path: root
status:
  conditions:
  - lastTransitionTime: "2026-09-18T07:08:05Z"
    message: No available shards to schedule the workspace
    reason: Unschedulable
    severity: Error
    status: "False"
    type: WorkspaceScheduled
  phase: Unavailable
  ```

Shard:
```
- apiVersion: core.kcp.io/v1alpha1
  kind: Shard
...
  spec:
    baseURL: https://127.0.0.1:6446
    externalURL: https://127.0.0.1:6443
    resourceLimits:
      hard:
        workspaces: "20"
      soft:
        workspaces: "10"
    virtualWorkspaceURL: https://127.0.0.1:6446
  status:
    conditions:
    - lastTransitionTime: "2026-09-18T06:59:38Z"
      message: 'soft/hard: workspaces=10/20'
      status: "True"
      type: ResourceLimitsApplied
    - lastTransitionTime: "2026-09-18T06:58:28Z"
      status: "True"
      type: Schedulable
    used:
      workspaces: "20"
```

## What Type of PR Is This?
/kind feature

## Related Issue(s)

Fixes #

## Release Notes

```release-note
Add shard soft/hard limits
```


olamilekan000 (approved):

olamilekan000:
This is going to need a follow-up implementation in the operator, right?

mjudeikis:
> This is going to need a follow-up implementation in the operator, right?

yes

olamilekan000:
> > This is going to need a follow-up implementation in the operator, right?
> 
> yes

I'll create an issue for it and probably work on it

pkg/virtual/admin/builder/forwarding.go
```diff
@@ -268,39 +316,23 @@ func withShardsView(cacheDynamicClusterClient kcpdynamic.ClusterInterface) forwa
...
+			// maps are left in place: the cache bookkeeping annotations below
+			// route the write to the owning shard.
+			for i, path := range MutableFields {
+				if !present[i] {
+					unstructured.RemoveNestedField(raw.Object, path...)
+					continue
+				}
+				if err := unstructured.SetNestedField(raw.Object, runtime.DeepCopyJSONValue(values[i]), path...); err != nil {
+					return nil, false, err
 				}
```
ntnn, line 399: Is this type safe? The cache-server does no validation iirc and just trusts the writes. 
Since here we are not doing any validation a user could write whatever into the fields and that would persist in the cache-server and then break in the cache->local sync when the shard tries to pull updates?

mjudeikis: Its kinda as safe as crd validation. I can see if I can bring typed types here to make it bit "safer"

ntnn: The VWs don't do schema validation iirc, that only happens in the real storage they are proxying to - and the cache server doesn't do schema validation.
We could also hook validation in the VWS.

ntnn (approved):
/lgtm
/approve

PR #4384, opened 2026-09-22
Return NotFound to WatchList clients when a shard does not serve the resource

waldemar-kindler:
## Summary

A controller consuming an APIExport through its virtual workspace never syncs its informers, and never reconciles anything, if it starts before any workspace has bound that export. There is no error anywhere — it simply sits there.

`DefaultDynamicDelegatedStoreFuncs`' watcher answers a `NotFound` from the delegate with an empty watch that yields no events and stays open, added in 8d385ddcc so that a wildcard consumer aggregating across shards is not wedged by a shard that holds none of these objects.

A WatchList client cannot use that stream. It treats the watch as its initial list and waits for the server to close the list with an `initial-events-end` bookmark before considering its store synced. An empty watch sends no bookmark and never errors, so nothing makes it relist — the comment's *"objects that later appear on this shard are picked up on the next relist"* never happens, because there is no next relist:

```
reflector.go:327: awaiting required bookmark event for initial events stream,
                  no events received for 10.004121875s      (every 10s, forever)
```

This is easy to hit in a normal deployment: a controller starts, the `APIExportEndpointSlice` is still empty, the first binding appears, the controller connects and watches a resource this shard does not serve yet, and from then on its informers never sync.

## The change

Surface the `NotFound` to WatchList clients, which retry and pick the resource up once the shard serves it. Ordinary watches are untouched and keep the empty watch.

## Alternative, and a question for reviewers

The fuller fix would be for the empty watch to emit the `initial-events-end` bookmark itself, so a WatchList client can sync against a shard that holds nothing. That needs a resourceVersion to report as synced-at, and I did not want to invent one — if there is a right value to use here, that is the better fix and I am happy to rework this.

## How it was found, and how to reproduce

Bisected from a downstream controller ([fleet](https://edp.buildth.ing/DevFW-CICD/fleet)) whose integration suite passes on v0.32.3 and fails on v0.33.0. The variable is *when* the controller starts, each run in its own process:

| controller starts | v0.32.3 | v0.33.0 |
| --- | --- | --- |
| after a workspace already binds the export | ok | ok |
| before any binding exists (empty endpoint slice) | ok | **hangs, informers never sync** |

Reproducing it needs only that ordering: publish an APIExport, start a controller against its virtual workspace with nothing bound yet, then create a workspace that binds it. On v0.32.3 the controller engages; on v0.33.0 it logs the reflector warning above every ten seconds until it gives up.

Everything else was ruled out by measurement along the way: the client's sdk/apimachinery version (fails with both 2.32.3 and the matching 2.33.0), multicluster-provider including its unreleased main, kcp main, single-shard vs sharded topology, and every hand-made LIST/WATCH shape — including a direct WatchList request, which returns its bookmark correctly when the resource *is* served.

## Testing

- New unit test covers both paths: a plain watch still gets the empty watch, a WatchList request gets the `NotFound`. It fails without the change (`expected a NotFound, got <nil>`).
- `pkg/forwardingregistry` package tests pass.
- Built v0.33.0 with this change and ran the downstream suite that first surfaced it: the previously hanging ordering passes, and the full suite passes.


mjudeikis:
Thanks @waldemar-kindler for this. Where did you see this? feel link is not working for me.  Can you just double-check this still failing on `main`, as we recently merged a few fixes in a similar code path to this. Is this already fixed, or is it still an issue?

waldemar-kindler:
Tried it with today's main branch.

Sorry for the internal link. Project is not published yet.

mjudeikis:
Ok, let see what CI says.

mjudeikis:
Next time please fill the PR template as CI will not allow to merge it without it

waldemar-kindler:
> Next time please fill the PR template as CI will not allow to merge it without it

I'm so sorry. Thanks for letting me know. Will do next time and thanks for the super quick response!

